Issue #75, opened 2019-12-11
Add Box to ThemeProvider


FalseLobster:
The Fabric Web folks have a Box in their ThemeProvider to apply the background color to the region covered by the ThemeProvider.  It seems like we might want to do this as well with a View.  See ThemeProvider and Box here:

https://github.com/microsoft/fluent-ui-react-staging/tree/master/packages/react-theming/src/components

rurikoaraki:
This seems to be an outdated issue. I'm not sure if this is still wanted/needed.

Going to close this in a few days if there's no objection. Feel free to reopen if you disagree

PR #80, opened 2019-12-17
✨ Add Box to ThemeProvider

FalseLobster:
This adds a 'Box' to ThemeProvider, mostly getting it out for review in the unlikely event someone wants to land this.

- The web is doing this too (see link in issue #75) on native it's just useful to have someone automatically set the background for you because that's one thing that might change when theme changes.
- Maybe I should've written the HOC using compose?
- re-exports themed-stylesheet from theming-react-native
- I kept theming-react-native with pure ReactJS support although I don't know if that actually makes sense any more.

### DIFF
--- a/apps/playground/App.tsx
+++ b/apps/playground/App.tsx
@@ -25,7 +25,7 @@ const mockThemingModuleHelper = createMockThemingModuleHelper(mockThemingModule,
 const customThemeRegistry = createPlatformThemeRegistry('TaskPane', mockThemingModuleHelper);
 // default theme
 customThemeRegistry.setTheme({});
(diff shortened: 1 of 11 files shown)


FalseLobster:
Ok @JasonVMo @tkamireh - I added a useBoxStyle hook, but the ThemeProvider can't actually use it because it's the one that's setting the ThemeContext.  Looking at how this gets used in App.tsx, it still feels like there's room for improvement because consumers likely want control over the Box.  I think the web folks made ThemeProvider a full-blown composed component so the Box slot could be replaced/tweaked, but hopefully this feels like a step in the right direction.

FalseLobster:
This got broken by repo rename.  I'll send PR separately

Issue #82, opened 2020-01-15
Move react-native-uifabric here (win32 controls)


FalseLobster:
We have a few win32 components in another repo.  This issue just tracks moving them here.

JasonVMo:
This work is done now.  See the commit referenced above.

PR #83, opened 2020-01-23
Migrate `react-native-uifabric` package from `sdx-platform` repository

dotnetjunky:
Summary of my changes:

- Copied all TypeScript files as-is, with only minimal changes at a few places to make ESLint happy. 
- Copied build scripts from other packages in this repository. 
- Added a `metroPack` Just script to use as bundling task for `react-native-uifabric` package, because the default webpack doesn't work with react-native-win32. 
- However, I have not been able to get bundling script to work just yet. So I'm leaving it empty for now. I want to merge what I have now to avoid possible merge conflict with other people's changes.

The net result is that the new package is building TS files and running Jests tests; but it doesn't have the final bundle file. :-)


dotnetjunky:
Linking to {dotnetjunky}/{fluentui-react-native}#82

packages/react-native-uifabric/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "just": "just-scripts",
+    "clean": "just-scripts clean",
+    "code-style": "just-scripts code-style",
+    "lint": "just-scripts eslint",
+    "start": "just-scripts dev",
+    "start-test": "just-scripts jest-watch",
+    "update-snapshots": "just-scripts jest -u",
+    "verify-api": "just-scripts verify-api-extractor",
+    "update-api": "just-scripts update-api-extractor",
+    "run-win32": "rex-win32 --bundle RNTester.win32 --component RNTesterApp --basePath ./dist/win32/dev",
```
tkamireh: Are the rex-win32 dependencies correct?

dotnetjunky: no, Rex is not working yet, because bundling is not implemented. I'll fix it when I got bundling task working.

tkamireh:
Let's call it experiments instead of experimentals. Gonna take one more quick pass over things, but looks good to me.

experimentals/react-native-uifabric/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "enzyme-to-json": "^3.1.1",
+    "office-ui-fabric-react": "^7.28.2",
+    "react": "16.8.6",
+    "react-dom": "16.8.6",
+    "react-native": "https://github.com/microsoft/react-native/archive/v0.60.0-microsoft.35.tar.gz",
+    "react-native-windows": "0.60.0-vnext.115",
+    "react-test-renderer": "16.8.6"
+  },
+  "peerDependencies": {
+    "react": "16.8.6",
```
acoates-ms: uifabric shouldn't have such specific peer dependencies.  -- We should really start doing semver here.
We should be ok with any 0.60 release of react-native.

experimentals/react-native-uifabric/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "react-native-uifabric",
```
acoates-ms: Is this win32 specific?  Shouldn't the name be something win32 specific while we are at it?

packages/themed-stylesheet/package.json
```diff
@@ -28,7 +28,7 @@
     "@types/react-native": "0.60.19",
     "@uifabricshared/build-native": "^0.1.1",
     "@uifabricshared/eslint-config-rules": "^0.1.1",
-    "react-native": "^0.59.0",
+    "react-native": "^0.60.0",
```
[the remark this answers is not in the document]
acoates-ms: Does this package use react-native?  I dont see it as either a dep or peerdep?   The answer to those questions would determine if we need to bump more things, and what kind of change type this change should be.

experimentals/react-native-uifabric/package.json
dotnetjunky: At this time, I'm just copying the existing package over as is. The package is marked as "private", so we can easily rename the package later, until we publish it.

packages/themed-stylesheet/package.json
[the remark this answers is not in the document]
JasonVMo: If I remember right this package does use StyleSheet from react native.  I don't know that it needs the version bump, that API has been stable for some time.  Do all our peer dependencies in RN need to be aligned to make the build system happy or can each package specify its minimum required version?

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "packages": ["packages/*", "scripts", "apps/*"],
+  "packages": ["packages/*", "scripts", "apps/*", "experimentals/*"],
```
JasonVMo: I agree with other comments.  This should be called experiments rather than experimentals, that matches the pattern in fabric-web.

packages/themed-stylesheet/package.json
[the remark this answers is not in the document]
acoates-ms: We probably want the dev deps aligned so that you dont have to download 100 versions of RN to work on the repo.  But the peerdeps should specify the full range of what the code would actually work with.   (Or what we support).  Which for now should probably be `>= 0.59 < 0.62` or something like that.

experimentals/react-native-uifabric/package.json
JasonVMo: I would leave this right now for consistency.  Ideally much of the code here will be shareable between platforms but I think the process should be:

1. get it checked in, mostly as-is
2. get metro working correctly for bundling
3. figure out the right directory structure / package names and migrate into that form

I think we will want to come up with new package names before we make this version the source of truth.

experimentals/react-native-uifabric/package.json
JasonVMo: I opened Issue #85 to track this as I think this needs to be done more broadly.  For this particular case will "react-native": "^0.60" resolve correctly against the tarball we are referencing for react-native-windows?

packages/themed-stylesheet/package.json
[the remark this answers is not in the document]
dotnetjunky: @acoates-ms Yes it does. It has this code. 

`import { ViewStyle, TextStyle, ImageStyle, StyleSheet } from 'react-native';`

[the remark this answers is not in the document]
dotnetjunky: Without this change, I'm getting build errors due to the fact that there're different versions of React-Native in the repo.

lerna.json
dotnetjunky: I don't know why I typed `experimentals`. I do prefer `experiments`.

experimentals/react-native-uifabric/package.json
dotnetjunky: I'm just copying this thing over from `sdx-platform`.

dotnetjunky: I'll change to ^16.0.0

packages/themed-stylesheet/package.json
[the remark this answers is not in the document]
dotnetjunky: @acoates-ms @JasonVMo This is a dev dep, so I think we should be good with this change?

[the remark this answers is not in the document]
dotnetjunky: I tried reverting the change and the build still succeeded. So I'm reverting it now.

PR #86, opened 2020-01-24
Make the metro bundle task to use for react-native-uifabric

dotnetjunky:
This PR implements a `Just` task, called metroPack, in order to produce the bundle for the `react-native-uifabric` package, and probably any packages that depends on react-native in the future.


experiments/react-native-uifabric/metro.config.js
```diff
@@ -0,0 +1,54 @@
...
+const uifabricsharedPackages = Object.keys(allPackages).reduce(
+  (prev, key) => (allPackages[key].packagePath.startsWith('packages') ? [...prev, key] : prev),
+  []
+);
+
+const additionalWatchFolders = uifabricsharedPackages.map(package => pathToSymLinkedPackage(package));
+
+const extraNodeModules = {
+  'react-native': path.resolve(__dirname, '../../node_modules/@office-iss/react-native-win32')
+};
```
acoates-ms: Can we just inline this?  seems odd to have it separate.

experiments/react-native-uifabric/package.json
```diff
@@ -7,7 +7,7 @@
   "typings": "lib/index.d.ts",
   "private": true,
   "scripts": {
-    "build": "just-scripts build --useMetro",
+    "build": "just-scripts build --useMetro --platform win32",
```
acoates-ms: Do we really want to build all the bundles during build?

The fluent repo split things out into:
build
lint
test
bundle
as top level commands.  That way you dont have to sit there waiting for everything to bundle whenever you do a build.

experiments/react-native-uifabric/metro.config.js
dotnetjunky: ok.

experiments/react-native-uifabric/package.json
dotnetjunky: OK, just pushed a new commit, which split the bundle task out of the build task.

acoates-ms:
Should probably add to the CI loop a task that builds the bundles, to verify that they do not get broken.

PR #87, opened 2020-01-28
Make RNTester loads successfully in web debugger

dotnetjunky:
With this change, ReactTest/RNTester can launch with the web debugger. To start the debugger server, run `yarn start:tester`, and then run `yan run-win32-web` to launch ReactTester.


tkamireh:
Can you add change files? `yarn change` from root. If you want you could add the step to check for change files as part of the ci loops. I think the script is `yarn checkchange` or something like that.

acoates-ms:
It doesn't look like any CI is running at all?

dotnetjunky:
Hmm, I'm not sure why the `jest` task fails when running as part of the `npm run build` script, but if I run `npm run jest` directly, it passes. This might just be an issue with the Just script.

azure-pipelines.yml
```diff
@@ -26,6 +26,5 @@ steps:
     displayName: 'npm run build'
 
   - script: |
-      yarn verify-api
-      yarn check-for-changed-files
```
FalseLobster: Can we leave the check-for-changed files and open an issue to reintroduce verify-api at least?

tkamireh: I don't understand why either of these were removed. What am I missing?

dotnetjunky: @FalseLobster Added the check-for-changed-files back.
@tkamireh The verify-api task doesn't exist and always fail.

tkamireh: ah I see, you could add a line to scripts/just.config.js (it should just be following the update-api task definition)

PR #88, opened 2020-01-28
Fabric Native Button change onPress to onClick

lenahong:


experiments/react-native-uifabric/src/components/Button/Button.types.ts
```diff
@@ -99,6 +99,10 @@ export interface IButtonProps extends IPressableProps {
    * A RefObject to access the IButton interface. Use this to access the public methods and properties of the component.
    */
   componentRef?: React.RefObject<IFocusable>;
+  /**
+   * A callback to call on button clock event
```
dotnetjunky: typo. 'click', not 'clock'.

PR #91, opened 2020-02-04
Fix build/test for themed-stylesheet

JasonVMo:
I'm breaking the build system changes into smaller steps to keep things running.  This first step fixes the current build.  The issue in themed-stylesheet is that the inclusion of react-native-uifabric brought in a funky react-native version that was having mismatches.

Since react-native-win32 is now unforked I moved all dependencies to react-native: "^0.60.0" rather than using the tarball, then made react-native-win32 an explicit dependency.

I also moved the various test related dependencies into the scripts package to have those all in one place so they can be better version aligned.


scripts/just.config.js
```diff
@@ -41,38 +40,35 @@ module.exports = function preset() {
...
   task('code-style', series('prettier', 'eslint'));
-  task('update-api', series('clean', 'copy', 'sass', 'ts', 'update-api-extractor'));
-  task('dev', series('clean', 'copy', 'sass', 'webpack-dev-server'));
+  task('update-api', series('clean', 'copy', 'ts', 'update-api-extractor'));
+  task('dev', series('clean', 'copy', 'webpack-dev-server'));
 
   task('build:node-lib', series('clean', 'copy', series(condition('validate', () => !argv().min), 'ts:commonjs-only'))).cached();
 
-  task('bundle', argv().useMetro ? 'metroPack' : 'webpack').cached();
+  task('bundle', series(condition('metroPack', () => argv().useMetro), condition('webpack', () => !argv().useMetro))).cached();
```
dotnetjunky: why this change?

scripts/just.config.js
```diff
@@ -41,38 +40,35 @@ module.exports = function preset() {
...
   task('jest-watch', jestWatch);
-  task('sass', sass);
   task('ts:commonjs', ts.commonjs);
   task('ts:esm', ts.esm);
-  task('ts:amd', ts.amd);
   task('eslint', eslint);
   task('ts:commonjs-only', ts.commonjsOnly);
   task('webpack', webpack);
-  task('metroPack', metroPackTask(argv()['bundleName']));
+  task('metroPack', () => metroPackTask(argv()['bundleName']));
```
dotnetjunky: what's the difference here?

scripts/just.config.js
JasonVMo: It turns out that using argv() directly in just.config.js (not inside a function) wreaks havoc with the just tasks.  It is somewhat indeterminate what happens but generally it does things like causes tasks to disappear.  I talked to Ken about this and since just doesn't have a ternary operator he recommended this series with two conditions type of approach.

scripts/just.config.js
JasonVMo: Basically don't run argv directly in the just config, run it when the task is being executed.

PR #95, opened 2020-02-06
Add Box to ThemeProvider

FalseLobster:
This is just a continuation of PR #80 which got into a weird state after the repo rename.

I (hopefully) addressed Tariq & Jason's feedback about the use of faulty withBox HOC, although I do wonder if we should take the same approach as Web where the Box is implemented with compose.  I thought my approach here was just "easier" because we don't have some of the same requirements as web, but looking at how I use it in App.tsx it's easy for me to imagine it adding an extra view to the hierarchy that Apps would otherwise want to adjust the props of.

Good enough for now, IMO.


packages/theming-react-native/src/index.ts
```diff
@@ -3,6 +3,6 @@ export * from './Theme.types';
 export { IThemeLayerProps, ThemeLayer } from './ThemeLayer';
 export * from './platform';
 export * from './ThemeProvider';
-export * from './ThemeProvider.types';
 export * from './ThemeContext';
 export * from './NativeModule';
+export * from '@uifabricshared/themed-stylesheet';
```
tkamireh: Does this actually need to export everything from themed-stylesheet?

packages/theming-react-native/src/ThemeProvider/ThemeProvider.tsx
```diff
@@ -22,7 +24,13 @@ export const ThemeProvider: React.FunctionComponent<IThemeProviderProps> = (prop
     };
   }, [registryToUse, themeName, setThemeState]);
 
-  const themeProvider = <ThemeContext.Provider value={theme}>{children}</ThemeContext.Provider>;
+  const boxStyle = getBoxStyle(theme);
```
tkamireh: Why not use the hook, and not export `getBoxStyle`?

packages/theming-react-native/src/index.ts
FalseLobster: It's a good question.  Early feedback with some consumers of @theming-react-native bemoaned having to import from a second package and I thought we were leaning toward having theming-react-native being a "friendly" reexport for the theming concepts in the repo--- I think the theming ramp is effectively already reexported from here.

packages/theming-react-native/src/ThemeProvider/ThemeProvider.tsx
FalseLobster: I can't use the hook here.  The ThemeProvider is what's setting the ThemeContext and useTheme here is before the context has been set, so useBoxStyle runs on an undefined theme object.  This is part of the reason I had Box as a separate component initially, because rendering <Box> underneath <ThemeContext.Provider> guaranteed a ThemeContext was initialized by the time the Box was rendered.

tkamireh: ah right, I didn't read carefully
I was concerned with exporting both the hook and the non-hook out of `index`, that I didn't notice that the context wasn't set up yet.

PPatBoyd:
j.w. we want this still, yeah?  Anything blocking or just low-pri?

FalseLobster:
> j.w. we want this still, yeah? Anything blocking or just low-pri?

Yea, I've just struggled to find time to do any coding work of my own :\

PR #97, opened 2020-02-11
New docs pages

PPatBoyd:
The fabric website is undergoing a change to be a gatsby website based on public/private .md/.mdx files; we can remove our old docs files and our dependency on office-ui-fabric-react in this new model.  I've added basic files with docs from the deleted files, but will expand on each according to current status in our repo.


scripts/typescript-compatibility-build.yml
```diff
@@ -123,7 +123,7 @@ jobs:
 
       # Build sample app consuming Office UI Fabric React to test TypeScript version compatibility
       - script: |
-          node common/scripts/install-run-rush.js test --to $(TypeScriptVersion)-office-ui-fabric-react-test
+          node common/scripts/install-run-rush.js test --to $(TypeScriptVersion)-fluentui-react-native-test
```
tkamireh: Do we actually run this compatability build pipeline anywhere? It's probably a good thing to think about, but if we're not using it anywhere let's maybe get rid of it?

experiments/react-native-uifabric/src/components/Pressable/Pressable.mdx
```diff
@@ -0,0 +1,3 @@
+# Pressable
```
tkamireh: Is there an issue open to track filling this (and other controls) in?

experiments/react-native-uifabric/package.json
```diff
@@ -30,21 +30,32 @@
     "@uifabricshared/immutable-merge": "0.2.2",
     "@uifabricshared/theming-ramp": "0.5.2",
     "@uifabricshared/theming-react-native": "0.3.2",
-    "tslib": "^1.9.3"
+    "tslib": "^1.10.0"
```
tkamireh: Any reason in particular for this bump? I'm guessing it's fine and there shouldn't be any fallout, but just want to make sure it's intended

PPatBoyd: I took all of the dependencies yarn brought in for oufr (see yarn.lock) as direct dependencies, including this bump.  I also had to add dev dependencies for oufr's devdependencies, which was a long, long list of packages we didn't need for the few we did need to build (e.g. yargs).

experiments/react-native-uifabric/src/components/Pressable/Pressable.mdx
PPatBoyd: I'll add one per-file :)

scripts/typescript-compatibility-build.yml
PPatBoyd: Let's add a backlog item to do so in our build-maintenance efforts; I don't want to add it to this change

experiments/react-native-uifabric/src/components/Stack/StackItem/StackItem.mdx
```diff
@@ -0,0 +1,3 @@
+# StackItem
```
FalseLobster: Maybe for Stack and StackItem include "Not Yet Implemented" to avoid giving the impression that they work and we just haven't written docs for them yet?

dotnetjunky: Or `Partially Implemented`? I think it works in certain cases.

experiments/react-native-uifabric/package.json
```diff
@@ -30,21 +30,32 @@
...
-    "react-native": "^0.60.0"
+    "react-native": "^0.60.0",
+    "es6-map": "^0.1.5",
+    "es6-weak-map": "^2.0.2",
+    "@uifabric/foundation": "^7.5.3",
+    "@uifabric/icons": "^7.3.2",
+    "@uifabric/merge-styles": "^7.8.3",
+    "@uifabric/react-hooks": "^7.0.3",
+    "@uifabric/set-version": "^7.0.3",
+    "@uifabric/styling": "^7.10.2",
```
dotnetjunky: what are these dependencies? Do we need them?

PPatBoyd: removing w/ abandoned oufr dependency removal (jasonmo has done it separately)

experiments/react-native-uifabric/src/components/Stack/StackItem/StackItem.mdx
PPatBoyd: Added Not Yet Implemented to keep it simple

docs/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "fluentui-docs",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next",
```
dotnetjunky: What is the `next` command?

docs/pages/Controls/Button.mdx
```diff
@@ -0,0 +1,20 @@
+# Button Documentation
```
dotnetjunky: nitpick: In the experiments package, we name the folder 'components'. Can we use the same name here rather than `Controls`?

docs/package.json
PPatBoyd: it's a low-config react framework to serve up the mdx pages, simpler than gatsby.  It's present so the repo can serve docs reasonable for local dev (and sample code can be build-checked via jsx in mdx), mdx is navigable and readable on github, and polished publish by the gatsby fluent-site

PR #105, opened 2020-02-13
Implement PersonaCoin component

dotnetjunky:
- Implement the PersonaCoin component
- Added a test page in RNTester for PersonaCoin. It consists of two sub pages: `StandardUsage` and `CustomizeUsage`.
    - `StandardUsage`  Demonstrate usage of `PersonaCoin` by setting the public props.
    - `CustomizeUsage`  Demonstrate usage of `PersonaCoin` through customization tokens.


experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.types.ts
```diff
@@ -0,0 +1,93 @@
+import { IViewWin32Props } from '@office-iss/react-native-win32';
+import { ImageProps, ViewProps, ImageURISource, TextProps } from 'react-native';
+import { IBackgroundColorTokens, IForegroundColorTokens } from 'src/tokens';
+import { IRenderData } from '@uifabricshared/foundation-composable';
+
+export const personaCoinName = 'RNFPersonaCoin';
+
+export enum PersonaSize {
```
JasonVMo: We should get rid of all enums in our interface files.  Instead they should be discrete unions so:

`export type PersonaSize = 'size4' | 'size8' | 'size24' | ...`

This has the advantage of actually being strings so elsewhere where you have switch statements they can instead become direct mappings.  I __think__ you can then do something like:

    interface IMyTable {
      [key: keyof PersonaSize]: { param1: string, param2: string }
    };

The type can be direct indexed using the string value and typescript will complain if you change the PersonaSize type.  (I think)  You can test it, if it doesn't then you might have to define it as a mapped type so:

   type IMyObject = { [K in keyof PersonaSize]: { param1: string, param2: string } };

Either way we shouldn't use TS enums, they are pure overhead and the fact that they become numbers means that mismatched versions will cause mismatched indexing into enums if a new value is added.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.types.ts
```diff
@@ -0,0 +1,93 @@
...
+  size32,
+  size40,
+  size48,
+  size56,
+  size72,
+  size100,
+  size120
+}
+
+export enum PersonaCoinColor {
```
JasonVMo: get rid of the enum

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.types.ts
```diff
@@ -0,0 +1,93 @@
...
+  lightRed,
+  gold,
+  burgundy,
+  warmGray,
+  coolGray,
+  cyan,
+  rust
+}
+
+export enum PersonaPresence {
```
JasonVMo: Ditto

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
```diff
@@ -0,0 +1,97 @@
+import { PersonaSize, PersonaCoinColor, PersonaPresence } from './PersonaCoin.types';
+import { ImageURISource } from 'react-native';
+
+interface Dictionary<T> {
+  [key: string]: T;
+}
+
+const presenceIconCache: Dictionary<ImageURISource> = {
```
JasonVMo: Long term we need a better strategy than having these embedded images

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+export type PersonaSizeConfig = {
+  physicalCoinSize: number;
+  iconSize: number;
+  initialsFontSize: number;
+};
+
+export function getSizeConfig(size: PersonaSize): PersonaSizeConfig {
+  switch (size) {
```
JasonVMo: See comment about getting rid of enums.  This should be a direct indexed const object like:

    const sizeValues = {
      size8: { physicalCoinSize: 8, iconSize: 0, initialsFontSize: 4 },
      size24: ...
    }
    const values = sizeValues[size];

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
```diff
@@ -0,0 +1,97 @@
...
+  blocked: {uri: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAMAAADXqc3KAAAAe1BMVEUAAADpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABv////pABv83uHqCiTsJz7rGjP7zNHvRVj/+vv+8vP6x83sIzv2mqXsKD/7zNL5v8b3oqz2maTuP1P82Nz819vtKUDrGjLrGTK2dyGvAAAAEXRSTlMA8dSzeQj64mseEhHkJMpq5T3coEAAAADSSURBVCjPhdHZFoIgFAVQGQRLGzhQiaLNw/9/YZRk4IvncZ+17uJys7mUGyYFoWxTJpxzghDC878vKFBftLX6UgN0MXqF/qRCTg5VaPI1ro36Re+MWQ/TOPrEXw7840uCNvb9oQVZ+qJAnbpSNQpfMHQTVx2YLyT0xJWG9IWAnbiyEL4gaCKPCopb7P9RDOeJqzNYeG7i3+eGBVMPC2Yczpn9cfTmAT58IoV5Rn4FDSdZbY1rxzk9qpXHofkcqtO2uXfhUHOn9SkLJoWQrCizmbwBa5IbuX8ugpIAAAAASUVORK5CYII='},
+  busy: {uri: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAMAAADXqc3KAAAANlBMVEUAAADpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvpABvsgIdoAAAAEXRSTlMA8eLUEQj6s2sheXrlyrJqHBqb3nMAAAB/SURBVCjPhZFRDsQgCEQFEdS223L/yy7JummtGN/nvASUCSvoFGSIclIXpwragC3deYn6IJZ/fqF24NXmHPri+E2rOlCDQTAKIBNZHbKJjyfEBHoCTbAn2ATMRPRHzZfPn+t/cH4SIw3rY2p17H2+l0VRRtr8ag3KgswomcKCL1NmGBp9lqsZAAAAAElFTkSuQmCC'}
+};
+
+export type PersonaSizeConfig = {
+  physicalCoinSize: number;
```
JasonVMo: Small note but js minification can't minify after the dot.  Thus having long names inside objects tends not to be minified.  It's always a balance but just something to be aware of.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+export function getPresenceIconSource(presence: PersonaPresence): ImageURISource {
+  return presenceIconCache[PersonaPresence[presence]];
+}
+
+function personaCoinColorToHexCode(personaCoinColor: PersonaCoinColor): string {
+  switch (personaCoinColor) {
+    case PersonaCoinColor.lightBlue:
```
JasonVMo: Same comment as above, use discrete unions of strings and direct indexing into an object rather than a case statement.

JasonVMo: Also should this table be in the tokens?  Such that a user can override one of the colors if they want?

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.tsx
```diff
@@ -0,0 +1,169 @@
...
+      iconStyle.height = iconSize;
+    }
+  }
+
+  return {
+    slotProps: mergeSettings<IPersonaCoinType['slotProps']>(useStyling(props), {
+      root: { ...rest, style: sizeStyle },
+      initials: {
+        children: initials,
+        style: initialsStyle
```
JasonVMo: This is a tricky one, you are passing a style object created inside this function into the settings, this will then get merged with the cached styles but will create a new object on each render pass.  As a result this control will re-render on each pass, even if it doesn't need to.  Granted the entry props for the persona block a lot of that but it still adds expense.

It might be better to use style functions in the styles section to produce the styles based on tokens.  This way the results will be cached based on token entry values.  It also means they will be shared across persona instances.  Come talk to me and I can explain how that might work.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.types.ts
dotnetjunky: I can change it, but are we OK with diverging from the web Persona component, which uses enum?

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
dotnetjunky: Yup, I'll switch to using Icon/SVG when it is available.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
dotnetjunky: OK, I renamed it to `physicalSize` to make it a tad shorter.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
dotnetjunky: I don't think we want to allow that. 'lightBlue' should show as light blue, instead of a different color. If they want to change color, they can provide the `coinColor` token.

experiments/react-native-uifabric/src/components/PersonaCoin/PersonaCoin.helpers.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function convertCoinColor(coinColor: PersonaCoinColor): string {
+  return colorTable[coinColor];
+}
+
+export function calculateEffectiveSizes(tokens: IPersonaCoinTokens): PersonaSizeConfig {
+  const { size, coinSize, iconSize, initialsSize } = tokens;
+
+  if (size) {
+    return sizeTable[size];
```
JasonVMo: It is probably best to apply the size value, then patch with coinSize, iconSize, and initialsSize as overrides.  So this could then become something like:

    const sizeEntry = size && sizeTable[size] || sizeTable['size40'];
    return {
      physicalSize: coinSize || sizeEntry.coinSize,
      etc...
    };

JasonVMo: To follow up a bit I think this is a good general practice for our APIs, for instance a fontVariant property might set fontFace, fontSize, and fontWeight.  It should be valid for a user to specify something like: fontVariant="header", fontWeight=500 to basically override one property specifically.

PR #109, opened 2020-02-20
Repath experiments/tester and experiments/react-native-uifabric

JasonVMo:
This change moves:
experiments/react-native-uifabric -> packages/experiments/react-native-uifabric
experiments/tester -> apps/win32

This also updates the lerna config to include the docs project and removes experiments/* from the lerna config and package.json at the root.


JasonVMo:
I'm withdrawing this for right now, we can just remove the components one by one from react-native-uifabric

PR #111, opened 2020-02-21
Write Persona component

dotnetjunky:
This component includes the `PersonaCoin` component and 4 optional Texts.


experiments/tester/src/RNTester/TestComponents/Common/Slider.tsx
```diff
@@ -0,0 +1,169 @@
+import * as React from 'react';
```
dotnetjunky: Ignore this file. I have sent out a separate PR for this change. Including it here to pass the CI build.

experiments/tester/src/RNTester/TestComponents/Common/AlignmentPicker.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
```
dotnetjunky: Ignore this file. I have sent out a separate PR for this change. Including it here to pass the CI build.

experiments/tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
```diff
@@ -1,76 +1,19 @@
 import * as React from 'react';
```
dotnetjunky: Ignore this file. I have sent out a separate PR for this change. Including it here to pass the CI build.

experiments/react-native-uifabric/src/components/Persona/Persona.helpers.ts
```diff
@@ -0,0 +1,66 @@
+import { PersonaSize } from '../PersonaCoin';
+import { ITextTokens } from '../../tokens';
+
+const textFontTable: { [key in PersonaSize]: ITextTokens } = {
```
tkamireh: why not make `{ [key in PersonaSize]: ITextTokens }` a type, say `IPersonaFontTable`?

experiments/react-native-uifabric/src/components/Persona/Persona.helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+  size40: { fontSize: 'medium' },
+  size48: { fontSize: 'medium' },
+  size56: { fontSize: 'xLarge' },
+  size72: { fontSize: 'xLarge' },
+  size100: { fontSize: 'xLarge' },
+  size120: { fontSize: 'xLarge' }
+};
+
+const secondaryFontTable: { [key in PersonaSize]: ITextTokens } = {
+  size8: { fontSize: 0 },
```
tkamireh: What's the advantage to setting fontSize to 0 as opposed to null guarding in the caller?

experiments/react-native-uifabric/src/components/Persona/Persona.settings.ts
```diff
@@ -0,0 +1,12 @@
+import { IComposeSettings } from '@uifabricshared/foundation-compose';
+import { IPersonaType, personaName } from './Persona.types';
+
+export const settings: IComposeSettings<IPersonaType> = [
+  {
+    tokens: {
+      horizontalGap: 16,
```
tkamireh: Should these values scale with the size of the persona?

experiments/react-native-uifabric/src/components/Persona/Persona.helpers.ts
dotnetjunky: will do that.

experiments/react-native-uifabric/src/components/Persona/Persona.types.ts
```diff
@@ -0,0 +1,46 @@
+import { IPersonaCoinProps, IPersonaCoinTokens } from '../PersonaCoin';
+import { ViewProps, TextProps } from 'react-native';
+import { IRenderData } from '@uifabricshared/foundation-composable';
+import { ITextTokens } from '../../../src/tokens/TextTokens';
+
+export const personaName = 'RNFPersona';
+
+export interface IPersonaState {
+  text?: string;
```
tkamireh: The state here is actually the existence of text so these should be bools, right? Though maybe we don't have a good way to retain the text values outside of `renderData.state`. Is it possible to merge the text into the slot directly, it's a child, which is just another prop. You're probably doing it the best way expressible right now, but I don't know how I feel about keeping text strings in state.

experiments/react-native-uifabric/src/components/Persona/Persona.settings.ts
dotnetjunky: yes, it should.

experiments/react-native-uifabric/src/components/Persona/Persona.types.ts
dotnetjunky: The reason I do this is so that I can omit the slot altogether if the text is empty. 
` {!!text && <Slots.text>{text}</Slots.text>}`
If it insert the text directly into the slot, the slot is always rendered.

tkamireh: I think the actual problem I have is that the word `state` means so many different things and I kind of wish renderData split things into state and info or something and that state actually reflected state in the `react` state and info got used for the things we need to keep around for rendering.

tkamireh: @dotnetjunky yup, I understand 😊. I just have a problem with the word state being used everywhere. I think what you have is totally reasonable, it would be cool though if you the text value was merged into the slot directly and all you needed to do was keep a bool around to tell you whether to render that slot. But that's a problem for another time

dotnetjunky: I agree. `state` here should have been called `data` or something like that.

experiments/react-native-uifabric/src/components/Persona/Persona.helpers.ts
dotnetjunky: So I use this value to decide whether I should hide this text element. If a `fontSize` is undefined/null, it means the Text is rendered as the default font size.

experiments/react-native-uifabric/src/components/Persona/Persona.settings.ts
dotnetjunky: Added style function to calculate the gap based on persona size.

PR #112, opened 2020-02-21
Adding V1 RadioGroup Component

samuelfreiberg:
Adding the RadioGroup & RadioButton component. It also integrates with the test app.


experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+    backgroundColor: 'black',
+    height: 10,
+    width: 10,
+    left: 4,
+    top: 4
+  }
+};
+
+export const focusedStyle = {
+  borderColor: 'rgba(128, 128, 128, 1)'
```
dotnetjunky: use theme color?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const info = React.useContext(RadioGroupContext);
+
+    const pressable = useAsPressable({
+      ...rest
+    });
+
+    const state: IRadioButtonState = {
+      ...pressable.state,
+      selected: info.selectedKey == userProps.buttonKey ? true : false,
+      disabled: disabled ? disabled : false
```
dotnetjunky: this can be written as: `disabled: disabled || false`

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+    // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
+    const info = React.useContext(RadioGroupContext);
+
+    const pressable = useAsPressable({
+      ...rest
+    });
+
+    const state: IRadioButtonState = {
+      ...pressable.state,
+      selected: info.selectedKey == userProps.buttonKey ? true : false,
```
dotnetjunky: this can be written as: `selected: info.selectedKey === userProps.buttonKey`

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const styleProps = useStyling(userProps, (override: string) => state[override] || userProps[override]);
+
+    // This function is called every time a RadioButton gains focus. It does two things:
+    // 1) Calls pressable's onFocus in order to keep track of our state's focus variable. It is dependent on pressable's
+    //    focus variable. Without this, it wouldn't stay updated because we're overriding it's onFocus below for the rootProps.
+    // 2) Selects the currently focused button by calling the RadioGroup's callback function.
+    const onFocusChange = (ev: NativeSyntheticEvent<{}>) => {
+      pressable.props.onFocus && pressable.props.onFocus(ev);
+      // This check is necessary because this func gets called even when a button loses focus (not sure why?) which then calls the client's onChange multiple times
+      if (!state.selected) info.onButtonSelect(buttonKey);
```
dotnetjunky: do you have the prettier plugin installed? This should be wrapped in a { } block and the statement goes to the next line.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+    //When the option is disabled, there's no built-in disable functionality for React-Native. So here's the logic I implemented for it.
+    const rootProps = state.disabled
+      ? { rest, ...accessibilityTraits }
+      : {
+          ...pressable.props,
+          ...accessibilityTraits,
+          rest,
+          onFocus: onFocusChange
+        };
+    const buttonProps = state.disabled ? { rest, style: { borderColor: 'grey' } } : rest;
```
dotnetjunky: So any calculations of styles should come from the style functions that are passed to the `styles` prop below, so that they can be cached. We should avoid constructing style objects in this function.

dotnetjunky: This comment applies to all other places in this function and the RadioGroup control.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+  render: (Slots: ISlots<IRadioButtonSlotProps>, renderData: IRadioButtonRenderData, ...children: React.ReactNode[]) => {
+    return (
+      <RadioGroupContext.Consumer>
+        {context => (
+          <Slots.root>
+            <Slots.button>
+              <Slots.innerCircle />
+            </Slots.button>
+            <Slots.content />
+            {children}
```
dotnetjunky: do we actually want to put `children` here?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+  settings: settings,
+  slots: {
+    root: View,
+    button: { slotType: View, filter: filterViewProps },
+    innerCircle: { slotType: View, filter: filterViewProps },
+    content: Text
+  },
+  styles: {
+    root: [],
+    button: [],
```
dotnetjunky: Referring to my comments above, all the style calculations should happen here.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
```diff
@@ -0,0 +1,64 @@
...
+import { IPressableState, IPressableProps } from '../Pressable';
+
+export const radioButtonName = 'RadioButton';
+
+// The state of the Radio Button. It extends IPressableState to keep track of press, focus, and hover
+export interface IRadioButtonState extends IPressableState {
+  /*
+   ** Whether or not the RadioButton is currently checked
+   */
+  selected: boolean;
```
dotnetjunky: make this property optional?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /*
+   ** Whether or not the option is disabled
+   */
+  disabled: boolean;
```
dotnetjunky: make this property optional?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface IRadioButtonProps extends IPressableProps {
+  /*
+   ** The text string for the option
+   */
+  content: string;
+
+  /*
+   ** A unique key-identifier for each option
+   */
+  buttonKey: string;
```
dotnetjunky: `key` is sufficent.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.settings.ts
```diff
@@ -0,0 +1,26 @@
...
+    root: {
+      accessible: true,
+      accessibilityRole: 'radio',
+      style: {
+        display: 'flex',
+        alignItems: 'flex-start',
+        flexDirection: 'column',
+        alignSelf: 'flex-start'
+      }
+    } as ViewProps,
```
dotnetjunky: why do you need `as ViewProps`?

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.settings.ts
```diff
@@ -0,0 +1,26 @@
...
+export const settings: IComposeSettings<IRadioGroupType> = [
+  {
+    root: {
+      accessible: true,
+      accessibilityRole: 'radio',
+      style: {
+        display: 'flex',
+        alignItems: 'flex-start',
+        flexDirection: 'column',
+        alignSelf: 'flex-start'
```
dotnetjunky: why is `alignSelf` needed?

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
```diff
@@ -0,0 +1,23 @@
...
+// import { RadioGroup } from '../../components';
+
+export const RadioGroupTest: React.FunctionComponent<{}> = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
+    return;
+  };
+
+  return (
+    <View>
```
dotnetjunky: You don't need this <View>

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { RadioButton, RadioGroup } from 'react-native-uifabric';
+// import { RadioButton } from '../../components';
+// import { RadioGroup } from '../../components';
```
dotnetjunky: delete commented code

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  onChange: (key: string) => void;
+}
+
+export interface IRadioGroupTokens extends ITextTokens {}
+
+export interface IRadioGroupSlotProps {
+  root: React.PropsWithRef<IViewWin32Props>;
+  label: ITextProps;
+  radioButtonContainer: ViewProps;
```
dotnetjunky: I think `container` is sufficient. This is a radio group, it's obvious this is a container of radio buttons.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export type IRadioGroupRenderData = IRenderData<IRadioGroupSlotProps, IRadioGroupState>;
+
+export interface IRadioGroupType {
+  props: IRadioGroupProps;
+  //tokens:
```
dotnetjunky: Doesn't it have any tokens?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+      style: {
+        display: 'flex',
+        alignItems: 'flex-start',
+        flexDirection: 'row',
+        alignSelf: 'flex-start',
+        minHeight: 20,
+        marginTop: 0,
+        position: 'relative'
+      }
+    } as ViewProps,
```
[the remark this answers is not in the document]
samuelfreiberg: Took this from @lenahong. Will follow-up with her and see if it's necessary

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+      style: {
+        backgroundColor: 'transparent',
+        width: 20,
+        height: 20,
+        top: 0,
+        left: 0,
+        borderWidth: 1,
+        borderStyle: 'solid',
+        borderColor: 'black',
+        borderRadius: 50,
```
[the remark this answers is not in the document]
samuelfreiberg: 50% which creates a circle.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+        marginTop: 4,
+        marginRight: 6,
+        marginBottom: 6,
+        marginLeft: 6,
+        position: 'relative'
```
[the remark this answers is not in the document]
samuelfreiberg: Doesn't look like it's necessary, removing it.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.settings.ts
samuelfreiberg: Took this from Button.settings.ts, but it looks like it's not necessary here. Removing it

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
samuelfreiberg: Removed the View

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
samuelfreiberg: Removed

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: It will, but not necessary for V1.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+    button: {
+      style: {
+        backgroundColor: 'transparent',
+        width: 20,
+        height: 20,
+        top: 0,
+        left: 0,
+        borderWidth: 1,
+        borderStyle: 'solid',
+        borderColor: 'black',
```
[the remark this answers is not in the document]
samuelfreiberg: Yes probably, I'm not familiar with themes and how they're defined so I'll link up with @FalseLobster

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
dotnetjunky: then it should be `50%`

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { RadioButton, RadioGroup } from 'react-native-uifabric';
+// import { RadioButton } from '../../components';
```
dotnetjunky: delete commented code

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
```diff
@@ -0,0 +1,23 @@
...
+// import { RadioGroup } from '../../components';
+
+export const RadioGroupTest: React.FunctionComponent<{}> = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
+    return;
```
dotnetjunky: Perhaps try calling the `alert()` function so that in the test app, we know it's working.

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { RadioButton, RadioGroup } from 'react-native-uifabric';
+// import { RadioButton } from '../../components';
+// import { RadioGroup } from '../../components';
+
+export const RadioGroupTest: React.FunctionComponent<{}> = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
```
dotnetjunky: You should create a proper function here so that React doesn't re-render if not necessary.

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+//import { IKeyboardEvent } from '@office-iss/react-native-win32';
```
dotnetjunky: delete

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+      setSelectedKey(key);
+      userCallback && userCallback(key);
+    },
+    [setSelectedKey, userCallback]
+  );
+
+  return { onChange, selectedKey };
+}
+
+// This hook will be used to capture onKeyDown
```
dotnetjunky: Is this block needed? If not, delete it.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,59 @@
...
+   ** Updates the selected button and calls the client’s onChange callback
+   */
+  onChange: (key: string) => void;
+}
+
+export interface IRadioGroupProps {
+  /*
+   ** Descriptive label for the RadioGroup. This will be displayed as the title of the radio group to the user
+   */
+  label: string;
```
dotnetjunky: Should this prop be optional? If it's not defined, we just show the buttons

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /*
+   ** The key of the RadioButton that will initially be selected
+   */
+  defaultSelectedKey: string;
```
dotnetjunky: I think this prop should be optional. If it's not defined, then none of the radio button is selected.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /*
+   **  OPTIONAL: An aria label for narrator. If not provided, it will be set to the label of the radio group
+   */
+  ariaLabel?: string;
+
+  /*
+   ** Callback for receiving a notification when the choice has been changed
+   */
+  onChange: (key: string) => void;
```
dotnetjunky: This prop should be optional.

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
samuelfreiberg: Deleted

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: Perhaps yes, will re-visit in V2

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: Changed to optional

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: Changed

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+      pressed: {}
+    }
+  },
+  radioButtonName
+];
+
+export const selectedStyle = {
+  style: {
+    position: 'relative',
+    borderRadius: 50,
```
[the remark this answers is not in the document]
lenahong: borderRadius should be in %

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+  {
+    root: {
+      accessible: true,
+      acceptsKeyboardFocus: true,
+      accessibilityRole: 'radio',
+      style: {
+        display: 'flex',
+        alignItems: 'flex-start',
+        flexDirection: 'row',
+        alignSelf: 'flex-start',
```
tkamireh: we should probably be more selective with when we set alignSelf. alignItems on the container in the group should be sufficient.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
lenahong: This can be done in settings using _overrides, no?

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
lenahong: use React.useCallback so you can specify when to re-render

tkamireh: `useCallback`?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Removed

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+// This hook updates the selected button and calls the client's onChange callback with the new button
+export function useAsRadioGroupSelection(defaultSelectedKey: string, userCallback?: onChangeCallback) {
+  const [selectedKey, setSelectedKey] = React.useState(defaultSelectedKey);
+
+  const onChange = React.useCallback(
+    (key: string) => {
+      setSelectedKey(key);
+      userCallback && userCallback(key);
+    },
+    [setSelectedKey, userCallback]
```
tkamireh: not that it makes a huge difference, but memoize on selectedKey instead

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
samuelfreiberg: Removed

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { compose, IUseComposeStyling } from '@uifabricshared/foundation-compose';
+import { ISlots, withSlots } from '@uifabricshared/foundation-composable';
+import { filterViewProps } from '../../utilities/RenderHelpers';
+import { settings } from './RadioGroup.settings';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { foregroundColorTokens, textTokens } from '../../tokens';
+import { useAsRadioGroupSelection } from '../../hooks';
+
+// Creates context to allow communication between its children components (RadioButtons)
+export const RadioGroupContext = React.createContext({
```
tkamireh: You can add a type here. `IRadioGroupContext` or something and pass it in as a generic to createContext.

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
samuelfreiberg: Replacing userCallback with selectedKey?

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,98 @@
...
+      label: { children: label },
+      radioButtonContainer: rest
+    });
+
+    return { slotProps, state };
+  },
+
+  render: (Slots: ISlots<IRadioGroupSlotProps>, renderData: IRadioGroupRenderData, ...children: React.ReactNode[]) => {
+    // This is necessary because renderData can be undefined, which causes problems in the return call.
+    let key = 'Undefined';
```
tkamireh: Can you elaborate, when is `renderData` undefined? I would think that as long as you don't do any crazy casting that you should end up with a valid renderData, with valid keys. Just use the actual `undefined` or `null` values and guard against that.

samuelfreiberg: You're correct, renderData should never be undefined. But when I simply to try to assign, for example,  key = renderData.state.selectedKey, I get a _Type 'string | undefined' is not assignable to type string_ error message.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Changed

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
samuelfreiberg: Something like this? 
  const onChange = React.useCallback((key: string) => {
    return;
  }, []);

experiments/react-native-uifabric/src/hooks/RadioGroup.hooks.ts
tkamireh: no, I'm talking about `setSelectedKey`, memoizing on that feels more correct.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
tkamireh: That seems strange? state is an optional on renderData (I'd expect an error on state being potentially undefined), but selectedKey is non-optional.

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
dotnetjunky: If the callback never changes, creating a normal function is the simplest solution. No need to call `useCallback`:

```
function onChange(key: string): void {
}
```

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
dotnetjunky: At the top of the `render` function, just check for `renderData.state`. If it's undefined, return null. Then you don't have to worry about it for the rest of `render`.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,98 @@
...
+    let onChangeCall = (key: string) => {
+      return;
+    };
+    if (renderData.state != undefined) {
+      key = renderData.state.selectedKey;
+      onChangeCall = renderData.state.onChange;
```
dotnetjunky: After you have checked for `renderData.state`, you can just do: 
```
const { key, onChange } = renderData.state;
```

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Fixed. Added theming

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+        position: 'relative'
+      }
+    },
+    innerCircle: {},
+    content: {
+      style: {
+        fontSize: 12,
+        marginTop: 3,
+        borderStyle: 'dashed',
+        borderColor: 'rgba(128, 128, 128, 0)',
```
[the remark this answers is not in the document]
samuelfreiberg: Fixed

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: Talked with Alicia, she agreed. Changed

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Both 50% and '50%' cause an error. 50 seems to work.

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
samuelfreiberg: Changed to the above

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
dotnetjunky: This is weird. @lenahong Is '50%' supposed to work? 50 shouldn't work. We need to create a bug for this.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
samuelfreiberg: No better theme option available for our usage. Leaving as is for V1. Still works with different theme right now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changed

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: I do have the Prettier plugin installed. Just fixed it though

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: For now, yes. In V2 I am going to change it up a bit to allow for Pickers, etc. to be children

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
lenahong: I just checked back on my borderRadius work and we decided not to support percentage for borderRadius because React Native doesn't. The reason that 50 worked is because any number beyond the maxinumum borderRadius is cut off at what the 50% of the width/height would be, in this case 10. If we want to eventually expose the width/height token for customizing radioButton size, I suggest creating a helper function that does the 50% calculation for the token processor.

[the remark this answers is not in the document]
dotnetjunky: I agree with Lena's suggestion. That is also the approach that I took for the Persona control. Basically, you calculate the width or height and divide by 2 to set the border radius. 

Setting it to 50 here is not the right solution.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
dotnetjunky: It's not fixed. The body of the `if` statement should be wrapped in { }

experiments/tester/src/RNTester/TestComponents/RadioGroupTest.tsx
samuelfreiberg: I tested it out, but it tends to get annoying when testing other features such as narrator. I'm going to leave it out for now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
samuelfreiberg: This is state, not a prop. It's always going to be either true or false

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
samuelfreiberg: This is state, not a prop. It's always going to be either true or false

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Just committed the fix

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Replaced this using _overrides in my newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Replaced this using _overrides in my newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: I agree this is not the right solution. However, tokens were in the plan for V2. Currently for V1, we don't support tokens. Do you think we can submit as is for V1, and then this sprint I'll work on getting the tokens working?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Will be addressed in V2

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -0,0 +1,86 @@
...
+    left: 4,
+    top: 4
+  }
+};
+
+export const hoveredStyle = {
+  style: {
+    position: 'relative',
+    opacity: 0.5,
+    borderRadius: 50,
```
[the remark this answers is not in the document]
samuelfreiberg: Will be addressed in V2

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Updating with constants in settings file for the time being.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Updating with constants in settings file for the time being.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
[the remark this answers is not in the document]
samuelfreiberg: Will be addressed in V2, opening a task for it. In the meantime I added constants at the top of the settings file to communicate intent better

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+  displayName: radioButtonName,
+
+  usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
+    const { content, buttonKey, disabled, ariaLabel, ...rest } = userProps;
+
+    // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
+    const info = React.useContext(RadioGroupContext);
+
+    const pressable = useAsPressable({
+      ...rest
```
lenahong: nitpick: useAsPressable(rest); should work

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    const state: IRadioButtonState = {
+      ...pressable.state,
+      selected: info.selectedKey === userProps.buttonKey,
+      disabled: disabled || false
```
lenahong: I don't think "|| false" is necessary

samuelfreiberg: disabled is an optional prop, so if they don't feed it in, it will be undefined. However I want state.disabled to be a boolean

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+      onAccessibilityAction: (event: { nativeEvent: { actionName: any } }) => {
+        switch (event.nativeEvent.actionName) {
+          case 'Select':
+            info.onButtonSelect(buttonKey);
+        }
+      }
+    };
+
+    //When the option is disabled, there's no built-in disable functionality for React-Native. So here's the logic I implemented for it.
+    const rootProps = state.disabled
```
lenahong: I don't think you should do this. All props should be passed into root. For instance, if disabled prop is set to true on the fly, it would need the pressable props and onFocus

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+      : {
+          ...pressable.props,
+          ...accessibilityTraits,
+          rest,
+          onFocus: onFocusChange
+        };
+
+    const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
+      root: rootProps,
+      button: rest,
```
lenahong: Do you need all the props to pass down to button and innerCircle slots?

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const ariaRoles = {
+      accessibilityRole: 'radiogroup',
+      accessibilityLabel: ariaLabel ? ariaLabel : label
+    };
+
+    const slotProps = mergeSettings<IRadioGroupSlotProps>(styleProps, {
+      root: { rest, ...ariaRoles },
+      label: { children: label },
+      container: rest
```
lenahong: Does container need all RadioGroup props?

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+  displayName: radioGroupName,
+
+  usePrepareProps: (userProps: IRadioGroupProps, useStyling: IUseComposeStyling<IRadioGroupType>) => {
+    const { label, ariaLabel, ...rest } = userProps;
+
+    // This hook updates the Selected Button and calls the customer's onClick function. This gets called after a button is pressed.
+    const data = useAsRadioGroupSelection(userProps.defaultSelectedKey ? userProps.defaultSelectedKey : '', userProps.onChange);
+
+    const state: IRadioGroupState = {
+      selectedKey: data.selectedKey,
```
FalseLobster: Maybe it's not a big deal, but I'd consider putting these inside an 'info' or 'context' subobject like Button does, or changing the override lookup that useStyling is doing.  As it is, if you added a "selectedKey" or "onChange" override in your settings those values would always get applied.  Unlikely to happen, but a little weird.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: You're right, this is unnecessary. Changing it now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (renderData.state == undefined) {
+      return null;
+    }
+
+    return (
+      <RadioGroupContext.Provider
+        // Passes in the selected key and a hook function to update the newly selected button and call the client's onChange callback
+        value={{
+          selectedKey: renderData.state.selectedKey,
+          onButtonSelect: renderData.state.onChange
```
FalseLobster: I'm curious why you called on thing onButtonSelect and another onChange? It'd be kinda nice to just say value={renderData.state} or value={renderData.state.context}

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Looks like no. Changing it now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Nope, changing it now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+      >
+        <Slots.root>
+          <Slots.label />
+          <Slots.container>{children}</Slots.container>
+        </Slots.root>
+      </RadioGroupContext.Provider>
+    );
+  },
+
+  settings: settings,
```
FalseLobster: Very nit, but I think you can just say `settings,` without the `settings:`

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+        info.onButtonSelect(buttonKey);
+      }
+    };
+
+    const accessibilityTraits = {
+      accessibilityRole: 'radio',
+      accessibilityLabel: ariaLabel ? ariaLabel : content,
+      // Why are we able to do this? The radio button that's focused is always going to be the one selected. Therefore, unless it's disabled,
+      // we can always set state=selected for the narrator, because every time the narrator is focused on a button, that's the one selected
+      accessibilityStates: state.disabled ? ['disabled'] : ['selected'],
```
FalseLobster: You should not report the state as 'selected' just because that's how Narrator appears to behave because every AT gets the same info and we need to report the correct info because we don't know what they do with it.

Moreover, this is still a bug in Narrator, because Narrator can query the selection state of any item before it gets actual focus and reads anything aloud and this is an invalid state.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+      innerCircle: rest,
+      content: { children: content }
+    });
+
+    return { slotProps, state };
+  },
+
+  render: (Slots: ISlots<IRadioButtonSlotProps>, renderData: IRadioButtonRenderData, ...children: React.ReactNode[]) => {
+    return (
+      <RadioGroupContext.Consumer>
```
FalseLobster: You don't need this.  You're using the hook in usePrepareProps.  This is the old way to consume context.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    const accessibilityTraits = {
+      accessibilityRole: 'radio',
+      accessibilityLabel: ariaLabel ? ariaLabel : content,
+      // Why are we able to do this? The radio button that's focused is always going to be the one selected. Therefore, unless it's disabled,
+      // we can always set state=selected for the narrator, because every time the narrator is focused on a button, that's the one selected
+      accessibilityStates: state.disabled ? ['disabled'] : ['selected'],
+      accessibilityActions: [{ name: 'Select', label: 'Select a RadioButton' }],
+      onAccessibilityAction: (event: { nativeEvent: { actionName: any } }) => {
```
FalseLobster: React.useCallback here to memoize this

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Just so I wouldn't get confused with the userProps.onChange. But you're right, I'll change it now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    // Grab the styling information from the userProps, referencing the state as well as the props.
+    const styleProps = useStyling(userProps, (override: string) => state[override] || userProps[override]);
+
+    // This function is called every time a RadioButton gains focus. It does two things:
+    // 1) Calls pressable's onFocus in order to keep track of our state's focus variable. It is dependent on pressable's
+    //    focus variable. Without this, it wouldn't stay updated because we're overriding it's onFocus below for the rootProps.
+    // 2) Selects the currently focused button by calling the RadioGroup's callback function.
+    const onFocusChange = (ev: NativeSyntheticEvent<{}>) => {
```
FalseLobster: This should probably also be a React.useCallback memoized function

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
dotnetjunky: This is not addressed?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changing

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Should it be dependent on [state.focused]?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
samuelfreiberg: Changing it to "key" is interacting strangely with something, it causes the control to select all the items at once which is completely wrong

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Weird behavior when just doing [state.focused], leaving out of commit for now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
samuelfreiberg: I can look into this weird behavior more in V2

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: Fixed in my newest commit

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Fixed in my newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    // This function is called every time a RadioButton gains focus. It does two things:
+    // 1) Calls pressable's onFocus in order to keep track of our state's focus variable. It is dependent on pressable's
+    //    focus variable. Without this, it wouldn't stay updated because we're overriding it's onFocus below for the rootProps.
+    // 2) Selects the currently focused button by calling the RadioGroup's callback function.
+    const onFocusChange = React.useCallback(
+      (ev: NativeSyntheticEvent<{}>) => {
+        pressable.props.onFocus && pressable.props.onFocus(ev);
+        // This check is necessary because this func gets called even when a button loses focus (not sure why?) which then calls the client's onChange multiple times
+        if (!state.info.selected) {
```
lenahong: instead of checking for selected state, try

if (pressable.props.onFocus && pressable.props.onFocus(ev)) 
{
  info.onChange(buttonKey)
}

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+    // Grab the styling information from the userProps, referencing the state as well as the props.
+    const styleProps = useStyling(userProps, (override: string) => state.info[override] || userProps[override]);
+
+    // This function is called every time a RadioButton gains focus. It does two things:
+    // 1) Calls pressable's onFocus in order to keep track of our state's focus variable. It is dependent on pressable's
+    //    focus variable. Without this, it wouldn't stay updated because we're overriding it's onFocus below for the rootProps.
+    // 2) Selects the currently focused button by calling the RadioGroup's callback function.
+    const onFocusChange = React.useCallback(
+      (ev: NativeSyntheticEvent<{}>) => {
+        pressable.props.onFocus && pressable.props.onFocus(ev);
```
lenahong: you can remove this line then

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
lenahong: wait, actually why don't you check for the focused state in pressable.state

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+        // This check is necessary because this func gets called even when a button loses focus (not sure why?) which then calls the client's onChange multiple times
+        if (!state.info.selected) {
+          info.onChange(buttonKey);
+        }
+      },
+      [pressable.props]
```
lenahong: [state, pressable.props]

samuelfreiberg: Added

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -28,60 +28,57 @@ export const RadioButton = compose<IRadioButtonType>({
...
-    const pressable = useAsPressable({
-      ...rest
-    });
+    const pressable = useAsPressable(rest);
 
     const state: IRadioButtonState = {
-      ...pressable.state,
-      selected: info.selectedKey === userProps.buttonKey,
-      disabled: disabled || false
+      info: {
```
FalseLobster: I think you misunderstood my comment--- I suggested making this change in RadioGroup, not in RadioButton.  I.e.

... in RadioButton
```
const state: IRadioButtonState = {
   ...pressable.state,
   selected: info.selectedKey === userProps.buttonKey,
   disabled: disabled || false
}
```
... in RadioGroup
```
const state: IRadioGroupState = {
      context (or info): { 
           selectedKey: data.selectedKey,
           onChange: data.onChange
       }
    };
```
The reason for this is because of how the line below works:
`useStyling(userProps, (override: string) => state[override] || userProps[override]);`

What this does is it looks at the `state` object and says, "Hey, do I have a 'disabled' state?" and if so, applies the __overrides from 'disabled' in the settings file.  I.e.

```
disabled: {
        tokens: {
          borderColor: 'buttonBorderDisabled',
          color: 'disabledBodyText',
          backgroundColor: 'background'
        }
      }
```

since no __override corresponds to `onChange` or `selectedKey` it doesn't make sense in Radio**Group** to pass that into the override state function.  Your just putting into the `state` to smuggle additional details into the render function for the purpose of setting the RadioGroup Context (hence why I'm proposing calling it `context` instead of `info` like Button).

**Technically** this does mean if you wrote an __override called `context` in the settings file, like

`__precedence: [ context, otherstuff... ]`
`__overrides: { context: { tokens: { borderColor: 'red' } } }`

that would get applied, but I think it conveys the code's intent better

FalseLobster: Realistically this isn't a huge detail, I felt liked I owed you some further clarification, though :)

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
+      root: {
+        rest,
+        ...pressable.props,
+        onFocus: onFocusChange,
+        accessibilityRole: 'radio',
+        accessibilityLabel: ariaLabel ? ariaLabel : content,
+        accessibilityStates: accessibilityStates,
+        accessibilityActions: [{ name: 'Select', label: 'Select a RadioButton' }],
```
dotnetjunky: The label prop shouldn't be hardcoded, should it? Eventually it need to be localized; so for now I think we should at least move it to settings?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Ok this makes sense, I wasn't too sure what the (override: string) => state[override] || userProps[override]) did but this clears it up for me. I'll look into changing this now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changed in my newest iteration.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changed in my newest iteration

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: This was changed in one of my last iterations

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+        onAccessibilityAction: React.useCallback((event: { nativeEvent: { actionName: any } }) => {
+          switch (event.nativeEvent.actionName) {
+            case 'Select':
+              info.onChange(buttonKey);
```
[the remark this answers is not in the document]
samuelfreiberg: Adding now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /*
+   ** OPTIONAL: Whether or not the radio button is selectable
```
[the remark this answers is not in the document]
samuelfreiberg: Removing

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { compose, IUseComposeStyling } from '@uifabricshared/foundation-compose';
+import { ISlots, withSlots } from '@uifabricshared/foundation-composable';
+import { filterViewProps } from '../../utilities/RenderHelpers';
+import { settings } from './RadioGroup.settings';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { foregroundColorTokens, textTokens } from '../../tokens';
+import { useAsRadioGroupSelection } from '../../hooks';
+
+export const RadioGroupContext = React.createContext<IRadioGroupContext>({
+  selectedKey: '',
```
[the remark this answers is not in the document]
samuelfreiberg: You're right, changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onChange: (key: string) => {
```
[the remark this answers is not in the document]
samuelfreiberg: Changing

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+});
+
+export const RadioGroup = compose<IRadioGroupType>({
+  displayName: radioGroupName,
+
+  usePrepareProps: (userProps: IRadioGroupProps, useStyling: IUseComposeStyling<IRadioGroupType>) => {
+    const { label, ariaLabel, ...rest } = userProps;
+
+    // This hook updates the Selected Button and calls the customer's onClick function. This gets called after a button is pressed.
+    const data = useAsRadioGroupSelection(userProps.defaultSelectedKey ? userProps.defaultSelectedKey : '', userProps.onChange);
```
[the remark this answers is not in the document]
samuelfreiberg: Changing with null

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const state: IRadioGroupState = {
+      selectedKey: data.selectedKey,
+      onChange: data.onChange
+    };
+
+    const styleProps = useStyling(userProps, (override: string) => state[override] || userProps[override]);
+
+    const ariaRoles = {
+      accessibilityRole: 'radiogroup',
+      accessibilityLabel: ariaLabel ? ariaLabel : label
```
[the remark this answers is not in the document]
samuelfreiberg: Changing, thanks for the tip

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+          }
+        }, [])
+      },
+      content: { children: content }
+    });
+
+    return { slotProps, state };
+  },
+
+  render: (Slots: ISlots<IRadioButtonSlotProps>, renderData: IRadioButtonRenderData, ...children: React.ReactNode[]) => {
```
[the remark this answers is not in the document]
samuelfreiberg: So I can use in the useStyling _(override: string) => state[override] || userProps[override])_ and as an easy point of reference. What would you suggest?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
+      root: {
+        rest,
+        ...pressable.props,
+        onFocus: onFocusChange,
+        accessibilityRole: 'radio',
+        accessibilityLabel: ariaLabel ? ariaLabel : content,
+        accessibilityStates: accessibilityStates,
+        accessibilityActions: [{ name: 'Select', label: 'Select a RadioButton' }],
+        onAccessibilityAction: React.useCallback((event: { nativeEvent: { actionName: any } }) => {
```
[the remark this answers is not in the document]
samuelfreiberg: Not sure I understand, do you mean before creating the root object?

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.types.ts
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration, resolving now.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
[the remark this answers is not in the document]
dotnetjunky: I mean you should assign the result of `React.useCallback` to a local variable at the top of the function, then assign it here.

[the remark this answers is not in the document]
samuelfreiberg: Got it, changed in my newest iteration. Resolving thread now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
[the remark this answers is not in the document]
dotnetjunky: In that case, you can create a local object inside `usePrepareProps()`. No need to pass the state to the `render` function.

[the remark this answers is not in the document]
samuelfreiberg: Changed in newest iteration. Removed state from renderData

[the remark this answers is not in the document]
samuelfreiberg: I believe I correctly did what you suggested, but can you look at my newest iteration and let me know if that's what you thought?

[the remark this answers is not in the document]
samuelfreiberg: Just updated again and removed state interface

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const pressable = useAsPressable(rest);
+
+    // Used when creating accessibility properties in mergeSettings below
+    const onAccessibilityAction = React.useCallback((event: { nativeEvent: { actionName: any } }) => {
+      switch (event.nativeEvent.actionName) {
+        case 'Select':
+          info.onChange && info.onChange(buttonKey);
+          break;
+      }
+    }, []);
```
dotnetjunky: Here you have to add `info` and `buttonKey` as a dependencies, because you use them inside the callback.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+    // 2) Selects the currently focused button by calling the RadioGroup's callback function.
+    const onFocusChange = React.useCallback(
+      (ev: NativeSyntheticEvent<{}>) => {
+        pressable.props.onFocus && pressable.props.onFocus(ev);
+        // This check is necessary because this func gets called even when a button loses focus (not sure why?) which then calls the client's onChange multiple times
+        if (!state.selected) {
+          info.onChange && info.onChange(buttonKey);
+        }
+      },
+      [state, pressable.props]
```
dotnetjunky: add `info` and `buttonKey` to the list of depedencies.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Yes you're right, changing now

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Adding in now

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,63 @@
...
+  root: React.PropsWithRef<IViewWin32Props>;
+  label: ITextProps;
+  container: ViewProps;
+}
+
+export type IRadioGroupRenderData = IRenderData<IRadioGroupSlotProps, IRadioGroupState>;
+
+export interface IRadioGroupType {
+  props: IRadioGroupProps;
+  //tokens:
```
dotnetjunky: nit: delete comment.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changed in newest iteration, resolving thread

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.tsx
samuelfreiberg: Changed in newest iteration, resolving thread

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
samuelfreiberg: Changed in newest iteration, resolving thread

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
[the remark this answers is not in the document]
dotnetjunky: This is not resolved. If the prop `onChange` is optional, you don't need to provide an empty implementation here.

PR #113, opened 2020-02-22
Fix direct debugger

dotnetjunky:
This change will make direct debugger work with the Fabric Tester app. Also updated the README file to include instructions on how to do direct debugging.

### DIFF
--- a/experiments/tester/README.md
+++ b/experiments/tester/README.md
@@ -22,7 +22,30 @@
     yarn run-win32
 ```
 
-## Debug `RNTester` app
+## Debug `RNTester` app with direct debugging
+
+1. Follow the same step #1 as above.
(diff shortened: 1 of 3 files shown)


experiments/tester/README.md
```diff
@@ -22,7 +22,30 @@
     yarn run-win32
 ```
 
-## Debug `RNTester` app
+## Debug `RNTester` app with direct debugging
+
+1. Follow the same step #1 as above.
+2. Follow the same step #2 as above.
+
+3. Build the RNTester bundle with dev option. This will ensure source map is included in the bundle.
```
tkamireh: Add the command inline. `bundle-dev` right?

tkamireh: lol ignore me I just saw it

PR #115, opened 2020-02-27
Create text component package + supporting fixes

JasonVMo:
This change ended up being a bit bigger than expected.  The test in themed-stylesheet is currently disabled as the cannot find module 'warnOnce' error popped up again.  I will address that in a subsequent PR but getting this one in so that the file moves are less disruptive.  That caveat aside here are the changes:

## Created component package for text

This moves the text component into its own package under /packages/components/text.  This is designed to be a package that can be invoked on any platform.  In truth it is probably not there yet as it is mostly keeping things intact.  I wanted to get the overall infrastructure in place first.

Note that the text component is imported and re-exported from react-native-uifabric to keep that interface the same.

## Created utility packages for hooks, adapters, and tokens

This removes the corresponding functionality from the react-native-uifabric package and puts them into three separate packages.  They are as follows:

- __adapters__ - this publishes an IViewProps, ITextProps, and IImageProps interface that varies depending on the platform used.  It also publishes explicit versions that can be used if so desired.  The filter functions are implemented in here as well, they should now multiplex to filter properties correctly for whatever platform they are invoked on.
- __interactive-hooks__ - this includes the various pressable hooks with split implementations such that they should mostly no-op on android or iOS but work on win32.  Note that pressable needs an implementation for Android/iOS but that is NYI for the moment.  I'll open an issue for that.
- __tokens__ - mostly a straight up move of the token folder, just to put it in a common place where independent packages can consume it.

## Did some react-native package cleanup

This was somewhat trying to address the jest failure, but while I tabled the full fix this is still useful.  It aligns the various react/react-native versions we are using, then makes sure to nohoist the metro and RN components from expo such that they don't change the base versions.  Our metro version was .51 from expo but that build of RN actually expects .54.

I validated the builds, bundles and running the win32 tester and did a hand check for matching functionality.  In the next phase of work (looking at jest and other configs) I will look at re-enabling the expo config.


JasonVMo:
Added Issue #116 to track the missing useAsPressable implementation.

FalseLobster:
Nice, exciting to see! 🎉

dotnetjunky:
While you're at it, can you also look into the source map issue? When I run web debugger for the tester app, I couldn't debug .tsx files under `react-native-uifabric`. They show up as .js files.

PR #118, opened 2020-03-04
Added token functionality for RadioGroup/RadioButton

samuelfreiberg:
Added token functionality for RadioGroup and RadioButton

### DIFF
--- a/experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
+++ b/experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
@@ -78,9 +78,6 @@ export const settings: IComposeSettings<IRadioButtonType> = [
           style: {
             borderColor: 'rgba(128, 128, 128, 1)'
           }
-        },
-        tokens: {
-          color: 'menuItemTextHovered'
         }
       },
       focused: {
@@ -100,11 +97,6 @@ export const settings: IComposeSettings<IRadioButtonType> = [
           style: {
             opacity: 0.5
           }
-        },
-        tokens: {
-          color: 'menuItemTextHovered',
-          borderColor: 'menuItemTextHovered',
-          backgroundColor: 'menuItemTextHovered'
         }
       },
       disabled: {
--- a/experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
+++ b/experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
@@ -66,7 +66,7 @@ export const settings: IComposeSettings<IRadioButtonType> = [
       }
     },
     // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out'
-    _precedence: ['disabled', 'hovered', 'focused', 'selected'],
+    _precedence: ['disabled', 'hovered', 'selected'],
     _overrides: {
       selected: {
         innerCircle: {
@@ -78,18 +78,6 @@ export const settings: IComposeSettings<IRadioButtonType> = [
           style: {
             borderColor: 'rgba(128, 128, 128, 1)'
           }
-        }
-      },
-      focused: {
-        content: {
-          style: {
-            borderColor: 'rgba(128, 128, 128, 1)'
-          }
-        },
-        tokens: {
-          borderColor: 'menuItemTextPressed',
-          color: 'menuItemTextHovered',
-          backgroundColor: 'menuItemTextPressed'
         }
       },
       hovered: {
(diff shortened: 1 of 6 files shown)


experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -82,7 +82,7 @@ export const RadioGroup = compose<IRadioGroupType>({
   },
   styles: {
     root: [],
-    label: [foregroundColorTokens, textTokens],
```
PPatBoyd: doesn't the label still need text tokens?

PPatBoyd: if the test page doesn't currently offer text customization, it seems worth adding a button to show it works; doesn't need to be exhaustive though.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -66,7 +66,7 @@ export const settings: IComposeSettings<IRadioButtonType> = [
       }
     },
     // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out'
-    _precedence: ['disabled', 'hovered', 'focused', 'selected'],
```
PPatBoyd: focused is mentioned in the comment above, but moreover; are we removing 'focused' as a state because it's blocking/breaking, or just because it doesn't work atm?  If it's benignly inactive, I would leave it in and just track the bug(s) that make it work again

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
samuelfreiberg: I talked with EJ and we thought we'd just allow the text color to change for now, which is why I got rid of textTokens. We'll probably add it in for V2.

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
samuelfreiberg: Right now, this specific code doesn't break anything, however, it doesn't do anything either because of a change I had to make in my last PR that deals with Pressable not doing exactly what I was hoping :). In addition, this will actually need to change because of the token work I did. I removed this code to clean up the file a bit. I do have a task open to fix the underlying problem in which I'll add this back in.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.tsx
PPatBoyd: I believe as written if someone theme changes font family or anything else, this label won't respect the theme.  I don't expect people to make radio button-specific stylizations often, but we have to assign the text props coming from the theme which may be a higher-level customization.

samuelfreiberg: Yah you're right, I just added back textTokens in my latest iteration

samuelfreiberg: And I just tested in the test app, customizing the fontFamily, for instance, does change the label.

experiments/react-native-uifabric/src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -45,7 +45,7 @@ export interface IRadioGroupProps {
   onChange?: (key: string) => void;
 }
 
-export interface IRadioGroupTokens extends ITextTokens {}
+export interface IRadioGroupTokens extends IForegroundColorTokens, ITextTokens {}
```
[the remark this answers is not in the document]
samuelfreiberg: Foreground contains the color token. So having this here allows the client to change the label color if they want to

experiments/react-native-uifabric/src/components/RadioGroup/RadioButton.settings.ts
```diff
@@ -78,33 +78,13 @@ export const settings: IComposeSettings<IRadioButtonType> = [
...
         }
       },
       hovered: {
         innerCircle: {
           style: {
             opacity: 0.5
           }
-        },
-        tokens: {
-          color: 'menuItemTextHovered',
```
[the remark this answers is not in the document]
samuelfreiberg: Yes, this was an unnecessary piece of code. I already initialized these tokens at the top of the file, so it inherits it already. The colors do style correctly for different states, I've tested it out :)

PR #119, opened 2020-03-06
🎨 Create host settings in theme object for direct access to native colors

FalseLobster:
this adds `getHostSettingsWin32(theme: ITheme)` to `theming-react-native`.  This allows the theme object to carry platform & host specific settings for clients who really need it under 

```
const theme: ITheme = {
    colors: // ...
    typography: //...
    components: //...
    host: IHostSettingsWin32 (or whatever platform)
}

type IHostSettingsWin32 = {
    palette: IOfficePalette;
}
```

The `getHostSettingsWin32` is just a way to extract the value in a strongly typed fashion (`IHostSettingsWin32` unless invoked on a theme without the win32 host settings)

These means customers are fully unblocked with direct access to the semantically named native colors at the discussions around the fluentui token sets continues to evolve.  I also spruces up the fluentui testpage for theming:

image

### DIFF
--- a/change/@uifabricshared-theming-react-native-2020-03-05-17-33-52-hostcolors.json
+++ b/change/@uifabricshared-theming-react-native-2020-03-05-17-33-52-hostcolors.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "🎨 Create host settings in theme object for direct access to native colors",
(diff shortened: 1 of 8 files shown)


dotnetjunky:
On the test page, you could use the <Picker> component for selecting theme. It makes the page shorter.

Issue #117, opened 2020-03-03
We should use WeakMaps for the theme settings cache instead of Symbol


FalseLobster:
Symbols seem copy-able in the babel transpilation down to ES5. Currently I'm deleting the old theme settings cache in resolvePartialTheme

@JasonVMo FYI.

FalseLobster:
Fixed

PR #120, opened 2020-03-06
🎨 Adds support for authoring JS theme definitions that are aware of the host theme setting

FalseLobster:
This adds `name: string` to `ITheme` so that in writing a functional theme definition a user can write an `IProcessTheme<ITheme, IPartialTheme>` that responds to the host theme setting directly.  See the new `caterpillarTheme` for how it responds to Black vs. Colorful.

This also addresses #117 which was preventing child themes from rendering correctly because the caches from the parent side were being copied into the fully resolved child themes.


experiments/tester/package.json
```diff
@@ -16,7 +16,8 @@
     "bundle-dev": "just-scripts bundle --useMetro --bundleName rnTester.dev",
     "start": "react-native start --projectRoot ./src",
     "run-win32": "rex-win32 --bundle RNTester --component RNTesterApp --windowTitle \"FluentUI Tester\" --basePath ./dist --pluginProps",
-    "run-win32-web": "rex-win32 --bundle RNTester --component RNTesterApp --basePath ./dist --useWebDebugger --windowTitle \"FluentUI Tester\" --useLiveReload --pluginProps"
+    "run-win32-web": "rex-win32 --bundle RNTester --component RNTesterApp --basePath ./dist --useWebDebugger --windowTitle \"FluentUI Tester\" --useLiveReload --pluginProps",
+    "run-win32-devmain": "rex-win32 --bundle RNTester --component RNTesterApp --basePath ./dist --pluginProps --useDevMain"
```
dotnetjunky: `--pluginProps` need to be last so that we can pipe extra parameters to it when running `yarn` command.

dotnetjunky: Can  you also add the `windowTitle` parameter?

dotnetjunky: I feel like we shouldn't add an Office-specific command to an open source project, but I understand RNW repo is also doing the same.

FalseLobster: Is --windowTitle on devmain or the tenant?

dotnetjunky: It shouldn't matter. the local copy of @rex-win32 package supports this prop.

dotnetjunky:
Is it a requirement to use emoji in PR title now?

experiments/tester/package.json
FalseLobster: Yes, but I needed to make changes to devmain code for this PR to work because I'm developing the native module concurrently.  When I had --windowTitle it wouldn't run with my C++ changes

FalseLobster:
> Is it a requirement to use emoji in PR title now?

Woah, you're saying this like that's a problem or something 😜

experiments/tester/package.json
dotnetjunky: I see. What error did you see? I checked in native change to windowTitle to devmain in this CL: Change 20186530 by REDMOND\kinhln@kinhln16_main on 2020/01/31 10:54:03. It should be available in devmain.

dotnetjunky: Make sure the `--pluginProps` is last.

FalseLobster: I just did a forward sync with my devmain code so I'll let you know.  I'm not opposed to just checking in the --windowTitle part and just assuming it works, though, tbh, I just needed it not there at the time to test my code.

FalseLobster: I'll address both of these suggestions in the next push

PR #122, opened 2020-03-07
⚗ Jazz up theme test page a little

FalseLobster:
- adds `host` to `ITheme`
- makes the Theme test page more configurable (I used radiobutton, not Picker.  sorry Luan! I would have but I don't think I have the tenant changes yet).

Now it looks like this:

image

### DIFF
--- a/experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
+++ b/experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
@@ -1,26 +1,11 @@
 import * as React from 'react';
 import * as ReactNative from 'react-native';
-import { getHostSettingsWin32, ThemeProvider, useTheme, IThemeDefinition } from '@uifabricshared/theming-react-native';
+import { getHostSettingsWin32, ThemeProvider, useTheme, IThemeDefinition, ThemingModuleHelper } from '@uifabricshared/theming-react-native';
 import { themedStyleSheet } from '@uifabricshared/themed-stylesheet';
 import { commonTestStyles } from '../styles';
 import { Button, PrimaryButton, Separator, StealthButton, Text, RadioGroup, RadioButton } from 'react-native-uifabric';
 import { ITheme, IPartialTheme } from '@uifabricshared/theming-ramp';
 import { customRegistry } from '../CustomThemes';
-
-const Panel: React.FunctionComponent = () => {
-  const [disabled, setDisabled] = React.useState(false);
-  const onClick = React.useCallback(() => setDisabled(!disabled), [disabled, setDisabled]);
-  const theme = useTheme();
-  return (
-    <ReactNative.View style={[commonTestStyles.viewStyle, commonTestStyles.stackStyle, { backgroundColor: theme.colors.background }]}>
-      <PrimaryButton onClick={onClick} content="Primary Button" disabled={disabled} />
-      <Button onClick={onClick} content="Default Button" disabled={disabled} />
-      <StealthButton onClick={onClick} content="Stealth Button" disabled={disabled} />
-      <Text>This is a text element</Text>
-      <Button onClick={onClick} content="This button has longer text" disabled={disabled} />
-    </ReactNative.View>
-  );
-};
 
 let brand = 'Office';
 
@@ -31,6 +16,7 @@ const brandColors = {
   Outlook: ['#CCE3F5', '#B3D6F2', '#69AFE5', '#2488D8', '#0078D7', '#106EBE', '#1664A7', '#135995']
 };
 
+// This IProcessTheme takes the parent theme and shims in the brand colors selected in the RadioGroup
 const fakeBrandTheme: IThemeDefinition = (theme: ITheme): IPartialTheme => {
   if (brand === 'Office') return {};
 
@@ -53,7 +39,12 @@ const fakeBrandTheme: IThemeDefinition = (theme: ITheme): IPartialTheme => {
   return brandedTheme;
 };
 
-customRegistry.setTheme(fakeBrandTheme, 'BrandedTheme');
+// this applies the shim to the default theme
+customRegistry.setTheme(fakeBrandTheme, 'Default');
+// this registers platform white colors
+customRegistry.setTheme(ThemingModuleHelper.getPlatformThemeDefinition('WhiteColors'), 'RealWhiteColors');
+// this applies the shim to the white colors theme
+customRegistry.setTheme(fakeBrandTheme, 'WhiteColors', 'RealWhiteColors');
 
 const getThemedStyles = themedStyleSheet((theme: ITheme) => ({
   swatch: {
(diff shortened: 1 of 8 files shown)


dotnetjunky:
The Picker component is already available to use in the test app, because the @rex-win32 package already has the native changes. That's why I can use it the PersonaCoin test page.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -31,6 +16,7 @@ const brandColors = {
   Outlook: ['#CCE3F5', '#B3D6F2', '#69AFE5', '#2488D8', '#0078D7', '#106EBE', '#1664A7', '#135995']
 };
 
+// This IProcessTheme takes the parent theme and shims in the brand colors selected in the RadioGroup
 const fakeBrandTheme: IThemeDefinition = (theme: ITheme): IPartialTheme => {
   if (brand === 'Office') return {};
```
dotnetjunky: Do you have the Prettier plugin installed? It should have reformatted this code and wrapper the body inside a { } block.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -138,22 +157,39 @@ const ThemeTestInner: React.FunctionComponent = () => {
...
     brand = app;
-    customRegistry.setTheme(fakeBrandTheme, 'BrandedTheme');
+    // Invalidate the DAG children of the shimmed brand colors
+    customRegistry.setTheme(fakeBrandTheme, 'Default');
+    customRegistry.setTheme(fakeBrandTheme, 'WhiteColors', 'RealWhiteColors');
   }, []);
+
+  const [theme, setTheme] = React.useState('Default');
+  const onThemeChange = React.useCallback(
+    (theme: string) => {
```
dotnetjunky: use can pass `setTheme` directly to `useCallback`:

`const onThemeChange = React.useCallback(setTheme, [setTheme]);`

dotnetjunky: Whing brings out the question, why do you need this at all? You can use `setTheme` directly.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -53,7 +39,12 @@ const fakeBrandTheme: IThemeDefinition = (theme: ITheme): IPartialTheme => {
   return brandedTheme;
 };
 
-customRegistry.setTheme(fakeBrandTheme, 'BrandedTheme');
+// this applies the shim to the default theme
```
dotnetjunky: what does `shim` mean?

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -75,17 +66,43 @@ const getThemedStyles = themedStyleSheet((theme: ITheme) => ({
...
+  },
+  pickerContainer: {
+    flexDirection: 'row',
+    justifyContent: 'space-evenly'
   }
 });
 
+const Panel: React.FunctionComponent = () => {
+  const [disabled, setDisabled] = React.useState(false);
+  const onClick = React.useCallback(() => setDisabled(!disabled), [disabled, setDisabled]);
```
dotnetjunky: I don't think `onClick` is worth caching if it depends on `disabled`. In all the usage below, if `disabled` changes, it will cause the buttons to re-render anyways.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -138,22 +157,39 @@ const ThemeTestInner: React.FunctionComponent = () => {
...
-          <RadioButton key={app} buttonKey={app} content={app} />
-        ))}
-      </RadioGroup>
-      <Text style={themedStyles.extraLargeStandardEmphasis}>Platform Theme</Text>
+      <Text style={themedStyles.extraLargeStandardEmphasis}>Configure Theme</Text>
       <Separator />
-      <Panel />
-      <Text style={themedStyles.extraLargeStandardEmphasis}>Custom JS Theme</Text>
+      <ReactNative.View style={styles.pickerContainer}>
+        <RadioGroup label="Pick App Colors" onChange={onAppChange} defaultSelectedKey="Office">
```
dotnetjunky: As mentioned above, I think `<Picker>` is a better choice.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
```diff
@@ -138,22 +157,39 @@ const ThemeTestInner: React.FunctionComponent = () => {
...
+          <RadioButton buttonKey="Office" content="Office" />
+          {Object.keys(brandColors).map((app: string) => (
+            <RadioButton key={app} buttonKey={app} content={app} />
+          ))}
+        </RadioGroup>
+        <Separator vertical />
+        <RadioGroup label="Pick Theme" onChange={onThemeChange} defaultSelectedKey="Default">
```
dotnetjunky: Use `<Picker>`

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: I'm pretty sure this is how prettier formatted this line of code.  I didn't author it that way.

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: https://en.wikipedia.org/wiki/Shim_(computing)

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: Maybe next time!  I agree it's more compact. But it's good to exercise the RadioGroup code, too. Look how funky it renders!  If I didn't use RadioGroup we might never notice :)

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: Yea, but I'm not gonna change it now, sorry.  :)

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: Hmm, maybe it just doesn't enforce if conditions to have { }. I'll adjust

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
FalseLobster: It can still avoid memory churn if the Panel re-renders for another reason, though, right?

i.e. if useTheme() changes, the stuff will rerender, but the onClick lambda expression won't necessarily be recreated

experiments/tester/src/RNTester/TestComponents/ThemeTest.tsx
dotnetjunky: That's totally fine, but I think the test page for RadioGroup should have more examples so that we don't have to discover issues with RadioGroup in another test page.

PR #123, opened 2020-03-09
Split components

JasonVMo:
As discussed in the meeting today this moves all components (with the exception of RadioGroup) into the packages/component/<componentName> directories with each component having its own package.

react-native-uifabric then gathers all of these components and re-exports them so the actual API for that should be pretty much unchanged.

The packages themselves should be pretty small and straightforward.  Generally the pattern for creating components is easy to follow just:

- directory should be /packages/components/ComponentName
- copy the root directory files from another component
- in package.json, update the name of the published package to be @fluentui-native/component-name (until we change the naming scheme)
- put your source files in a subfolder called src with an index.ts file

That's basically it.  If you want it published as part of react-native-uifabric then add an additional entry to the index file there.

I did a full build and bundle pass and validated that the run-win32 command still works and produces usable controls.


dotnetjunky:
It would be nice if we have some script that can generate the skeleton of a component package. We run the script, providing the name of the component and it generates all the required files. This will be particularly helpful when we write more and more packages.

JasonVMo:
> It would be nice if we have some script that can generate the skeleton of a component package. We run the script, providing the name of the component and it generates all the required files. This will be particularly helpful when we write more and more packages.

I created Issue #124 to track this.

PR #126, opened 2020-03-10
Add placeholder variants for Text control

ksiler:
This change adds `variants` to `ITypography` so that users can specify a font variant for fluent-native text. These are placeholder variants since adding the new typeramp for win32 is still in the works.


packages/utils/tokens/src/text-tokens.ts
```diff
@@ -1,15 +1,34 @@
...
-import { IOperationSet } from '@uifabricshared/foundation-tokens';
+import { TextStyle, TextProps } from 'react-native';
 import { ITheme } from '@uifabricshared/theming-ramp';
+import { styleFunction } from '@uifabricshared/foundation-tokens';
 
 export interface ITextTokens {
   fontFamily?: TextStyle['fontFamily'] | string;
   fontSize?: TextStyle['fontSize'] | string;
   fontWeight?: TextStyle['fontWeight'] | string;
+  fontVariant?: string;
```
JasonVMo: Should this be instead something like fontVariant?: keyof ITypography['varaints']

JasonVMo: Also if we are adding the variant to the props then it might be good to create ITextVariantToken or something that can be included in the props interface and in the ITextTokens interface.

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
```diff
@@ -1,46 +1,60 @@
...
+    families: {
+      primary: 'Segoe UI',
+      secondary: 'System',
+      cursive: 'System',
+      monospace: 'System',
+      sansSerif: 'System',
+      serif: 'System'
+    },
+    variants: {
+      smallStandard: { face: 'Segoe UI', size: 11, weight: '400' },
```
JasonVMo: I think we spoke about having these defined as something like:
 smallStandard: { face: 'primary', size: 'small', weight: 'normal' }
That way updates to the base ramps will update the variants as well.

packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
```diff
@@ -1,80 +1,149 @@
...
+      ...palette
+    },
+    typography: {
+      ...module.typography,
+      variants: {
+        smallStandard: {
+          face: module.typography.families.primary,
+          size: module.typography.sizes.small,
+          weight: module.typography.weights.normal
+        },
```
JasonVMo: See the comment above.  If the values in variants refer to the names of the font ramps, pulling in module.typography will automatically update these values so this section won't be needed.

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
ksiler: I'd like to optimize it that way but I'm unsure how. I can't reference those values in the way you have it written. Not all of those values are strings. For instance, size is expected to be a number and weight is expected to be a string but that of '100' through '900' in increments of 100. How can I reference the values by the way they're defined above?

packages/utils/tokens/src/text-tokens.ts
```diff
@@ -1,15 +1,39 @@
...
+export type ITextTokens = ITextStyleTokens & ITextVariantTokens;
+
+export function _buildTextStyles(tokens: ITextTokens, theme: ITheme): TextProps {
+  if (tokens.fontFamily || tokens.fontSize || tokens.fontWeight || tokens.fontVariant) {
+    return {
+      style: {
+        fontFamily: tokens.fontFamily
+          ? theme.typography.families[tokens.fontFamily] || tokens.fontFamily
+          : theme.typography.variants[tokens.fontVariant].face,
+        fontSize: tokens.fontSize
```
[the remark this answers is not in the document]
FalseLobster: Sorry, destruc*tur*ing not destructing.  It's when you do `const { a, b c } = foo;`

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
JasonVMo: There are a couple of parts of it.  First the values reference the name of the key in the typeramp rather than the values.  Then you export a function that resolves types against a theme that people can use in their style processing.  The function would probably look something like:

```
export function resolveTextTokens(theme: ITheme, tokens: ITextTokens): ITextTokens {
  const typography = theme.typography;
  // grab a variant if set, could provide a default value if we want but that means values always get set
  const variant = tokens.variant && variants[tokens.variant];
  const fontSize = tokens.fontSize || (variant && (typography.sizes[variant.size] || variant.size));
  // repeat for others

  // use spread trick to avoid setting fontSize: undefined, unnecesary if there is always a default variant
  return {
    ...(fontSize && { fontSize }) // may need an || {}, can't remember
    ...(fontFamily && { fontFamily })
    ...(fontWeight && { fontWeight })
 };
}
```

This basically returns a style fragment that can be mixed in for style creation.

packages/utils/tokens/src/text-tokens.ts
ksiler: I added keyof ITypography['variants'] to fontVariant and created ITextVariantTokens so it can be included in ITextTokens and in the props interface

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
ksiler: Thanks for the clarification - I forgot about adding the case where variant needs to be looked up by key. I updated the _buildTextStyles function to include this case. It looks a little different than what you wrote here but I think it achieves the same result. Let me know if I can optimize it or make it a little easier to read. The order of precedence for resolving text tokens is now 1. token key 2. token concrete 3. variant key 4. variant concrete.

packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
ksiler: Updated this by removing this section.

packages/components/text/src/Text.types.ts
```diff
@@ -1,20 +1,21 @@
...
-export type ITextType<TBase = INativeTextProps> = {
-  props: ITextProps<TBase>;
-  slotProps: {
-    root: TBase;
-  };
-};
+import { ITextTokens, ITextVariantTokens, IForegroundColorTokens, IColorTokens } from '@fluentui-native/tokens';
+import { ITextProps as INativeTextProps } from '@fluentui-native/adapters';
+
+export const textName = 'RNFText';
```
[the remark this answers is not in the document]
ksiler: Is there a good reason for this? All other controls follow an "RNF[control]" format.

packages/utils/tokens/src/text-tokens.ts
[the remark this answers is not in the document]
ksiler: Updated this function to include some destructuring.

packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
```diff
@@ -1,80 +1,82 @@
-import { ITheme, IPartialPalette, IColorRamp, resolvePartialTheme } from '@uifabricshared/theming-ramp';
```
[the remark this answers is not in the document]
ksiler: I'm not entirely sure :(

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
```diff
@@ -4,12 +4,13 @@ import { getStockWebPalette, ITypography, ISpacing } from '@uifabricshared/themi
...
+      small: 10,
       medium: 12,
       large: 14,
-      largePlus: 18,
-      extraLarge: 24,
-      huge: 32
+      largePlus: 16,
+      extraLarge: 20,
+      huge: 28,
+      giant: 42
```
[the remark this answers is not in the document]
ksiler: Otherwise, would I have to define new sizes for the new type ramp? Like "fluentSmall", "fluentMedium", "fluentLarge", etc?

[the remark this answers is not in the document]
ksiler: I just remembered these aren't px but typographical points - I'll change them to the appropriate values (e.g. medium should be '9' instead of '12' to reflect 9 tp === 12 px) . I'm also going to make these sizes reflect the sizes in the classic type ramp since we haven't added the new type ramp yet. We can decide how to handle naming for the new type ramp values later?

PR #132, opened 2020-03-19
FluentUI Callout

PPatBoyd:
Also made organizing changes to share styles in test pages


packages/components/Callout/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "clean": "just-scripts clean",
+    "update-snapshots": "just-scripts jest -u",
+    "verify-api": "just-scripts verify-api-extractor",
+    "update-api": "just-scripts update-api-extractor"
+  },
+  "dependencies": {
+    "@uifabricshared/foundation-compose": "^1.5.9",
+    "@fluentui-react-native/adapters": "^0.1.0",
+    "@fluentui-react-native/interactive-hooks": "^0.1.0",
+    "@fluentui-react-native/tokens": "^0.1.0"
```
samuelfreiberg: Please make sure the dependencies are right. Some of these look wrong, every dependency on something published out of this repo should be on the latest version. You should see warnings in the console during bundle that mention this. e.g. tokens, adapters, interactive-hooks all look out of date. You can hover on the text in vs code to see the latest version.
- Taken from what Alicia commented on my PR regarding Checkbox. Applies here as well

packages/components/Callout/src/Callout.tsx
```diff
@@ -0,0 +1,30 @@
+import { requireNativeComponent } from 'react-native';
+import { IUseComposeStyling } from '@uifabricshared/foundation-compose';
+import { calloutName, ICalloutProps, ICalloutSlotProps, ICalloutType } from './Callout.types';
+import { useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { composable } from '@uifabricshared/foundation-composable';
+
+const RCTCallout = requireNativeComponent('RCTCallout');
+
+export const Callout = composable<ICalloutType>({
```
samuelfreiberg: Just because I'm curious, why do some of our components use "compose" and others use "composable"?

packages/components/Callout/package.json
PPatBoyd: checked, all good :)

packages/components/Callout/src/Callout.tsx
PPatBoyd: if I understand correctly (and this is a good point that should be documented if it isn't, I didn't find it) the difference is 'compose' creates a HOC [using composable internally] where composable creates a functional component.  @jasonvmo any elaboration would be useful in distinguishing here :)

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setState({ ...state, showCustomizedCallout: !state.showCustomizedCallout });
+  }, [state, setState]);
+
+  const toggleShowCustomizedCalloutInner = React.useCallback(() => {
+    setState({ ...state, showCustomizedCallout: !state.showCustomizedCallout });
+  }, [state, setState]);
+  const myRect: ScreenRect = { screenX: 10, screenY: 10, width: 100, height: 100 };
+  return (
+    <View>
+      <TextWin32 ref={stdBtnRef} style={fabricTesterStyles.testSection}>
```
FalseLobster: I don't see a reason we can't use the core Text component.  Considering that Jason is working on making this run on other platforms, I think we should default to the x-plat primitives whenever possible

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {state.showStandardCallout && (
+        <Callout anchorRect={myRect}>
+          <View style={{ height: 200, width: 400 }}>
+            <Button content="test button please ignore" onClick={toggleShowStandardCalloutInner} />
+          </View>
+        </Callout>
+      )}
+
+      <TextWin32 ref={custBtnRef} style={fabricTesterStyles.testSection}>
+        Customized Usage
```
FalseLobster: I'd love to see some more detail in this test page!  I know this is probably just a V1 check in though.

packages/components/Callout/src/Callout.tsx
FalseLobster: I would say the main difference is that composable is sort of a base-class component that doesn't care about what theming/styling system is used, but provides a mechanism (useStyling) to inject one.  It just asks you to define the Slots and your hook logic in usePrepareProps. You can also provide a useStyling implementation (you shouldn't, see compose below) and the render function if you have >1 slot.  It's useful for FTZ because it has no actual visuals and is just a direct render of a purely behavioral native component.  

It allows for "old school" functional components to be described in their individual parts so that they can be recomposed without violating the rule of hooks and allows for eliding unnecessary vdom nodes.  Our philosophy is a little different from the web, here, FYI.

Compose's **primary** purpose is the injection of our idea of what the useStyling implementation should be into the composable pattern.  So to this end it's useful to look at useStying to see what it does:

```ts
// get the theme value from the context (or the default theme if it is not set)
  const theme = React.useContext(ThemeContext) || getTheme();
  // get the cache for this component from the theme
  const cache = getComponentCache(theme);

  // resolve the array of settings for these options
  lookupOverride = lookupOverride || props;
  const { settings, key } = getThemedSettings<IComponentSettings<IWithTokens<TSlotProps, TTokens>>, ITheme>(
    options.settings,
    theme,
    cache,
    baseKey,
    lookupOverride,
    _getSettingsFromTheme as any
  );

  // finish by processing the tokens and turning IComponentSettings into ISlotProps (this removes things like _overrides)
  return returnAsSlotProps(
    processTokens<TSlotProps, TTokens, ITheme>((props as unknown) as TTokens, theme, settings as any, options.resolvedTokens, key, cache)
  ) as IWithTokens<TSlotProps, TTokens>;
```

So it 
1. Is aware of our Theme object, because it calls looks at the ThemeContext.  This is a big thing.
2. It checks the theme's cache for resolved component slot props before doing any work, if none found....
3. Gets your settings object/function, processes it in response to the theme & applies overrides that you specify in your state 
4. resolves your tokens

So I would expect any component that wants to participate in our Theming & Styling system to have a composed version.

Whether Callout needs both a composed version & composable version, or just one or the other depends on the scenarios we expect it being used in.

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
PPatBoyd: Same!  I'm intending to expand as I add props

packages/components/Callout/src/Callout.tsx
FalseLobster: In short,
- Composable - Purely behavioral, but provides extensibility point for a styling system to be provided, allows for recomposition without overhead.
- Compose - Injects our theming system into Compose

FalseLobster: One of our "big ideas" was that the web team would take our implementation of compose and just write a useStyling implementation that makes more sense in ReactJS because it uses CSS, but they chose another route :)

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from 'react';
+import { ScreenRect, Text, View } from 'react-native';
+import { Button, Callout, Separator } from '@fluentui/react-native';
+import { fabricTesterStyles } from '../Common/styles';
+
+export const CalloutTest: React.FunctionComponent<{}> = () => {
+  const [state, setState] = React.useState({
```
dotnetjunky: It will make the code cleaner or easier to read if you use separate states for each callout. 

```
const [showStandardCallout, setShowStandardCallout] = React.useState(false);
const [showCustomizedCallout, setShowCustomizedCallout] = React.useState(false);
```

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <View>
+      <Text ref={stdBtnRef} style={fabricTesterStyles.testSection}>
+        Standard Usage
+      </Text>
+      <Separator />
+      <Button content="Press for Callout" onClick={toggleShowStandardCallout} />
+
+      {state.showStandardCallout && (
+        <Callout anchorRect={myRect}>
+          <View style={{ height: 200, width: 400 }}>
```
dotnetjunky: Can you set width and height on the Callout directly and avoid the View?

packages/components/Callout/src/Callout.tsx
```diff
@@ -0,0 +1,30 @@
...
+  displayName: calloutName,
+  usePrepareProps: (props: ICalloutProps, useStyling: IUseComposeStyling<ICalloutType>) => {
+    const { componentRef, ...rest } = props;
+    const buttonRef = useViewCommandFocus(componentRef);
```
dotnetjunky: why is it named `buttonRef`?

packages/components/Callout/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../../tsconfig.json",
```
dotnetjunky: Can we use a package name syntax rather than relative path?

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
PPatBoyd: Nope; per API review we're having the callout size to contents, and not providing direct w/h props

packages/components/Callout/tsconfig.json
PPatBoyd: I'll check this out; I was going off of another package's tsconfig for reference

experiments/tester/src/RNTester/TestComponents/Common/styles.ts
```diff
@@ -1,11 +1,116 @@
 import { StyleSheet } from 'react-native';
+import { IStackProps } from '@fluentui-react-native/stack';
 
-export const styles = StyleSheet.create({
+export const commonTestStyles = StyleSheet.create({
```
FalseLobster: I feel like these should probably use themedStyleSheet like the styles I wrote for the theme test page... it's part of the reason I didn't use sectionHeader like Luan did

packages/components/Callout/package.json
FalseLobster: @PPatBoyd your statement may no longer be true when you pull from master.  I know adapters is for sure at 1.1, for example

packages/components/Callout/src/Callout.types.ts
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import * as ReactNative from 'react-native';
+import { IRenderData } from '@uifabricshared/foundation-composable';
+import { IBackgroundColorTokens, IBorderTokens } from '@fluentui-react-native/tokens';
+import { IFocusable } from '@fluentui-react-native/interactive-hooks';
+
+export const calloutName = 'RNFCallout';
+
```
FalseLobster: I think we should simply call these "Callout"

packages/components/Callout/src/Callout.types.ts
```diff
@@ -0,0 +1,58 @@
...
+  | 'bottomCenter'
+  | 'bottomRightEdge';
+
+export interface ICalloutTokens extends IBackgroundColorTokens, IBorderTokens {
+  anchorRect?: ReactNative.ScreenRect;
+  directionalHint?: DirectionalHint;
+  gapSpace?: number;
+  minPadding?: number;
+}
+
```
FalseLobster: It's a little curious to me that your using Tokens & Composable.  Are you adding a compose version later?

experiments/tester/src/RNTester/TestComponents/Common/styles.ts
dotnetjunky: Alicia, can the styles you wrote be shared in some way? It'd be nice to have a consistent look and feel across all test pages.

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const stdBtnRef = React.useRef<Text>(null);
+
+  const custBtnRef = React.useRef<Text>(null);
+
+  const toggleShowStandardCallout = React.useCallback(() => {
+    setShowStandardCallout(!showStandardCallout);
+  }, [showStandardCallout, setShowStandardCallout]);
+
+  const toggleShowStandardCalloutInner = React.useCallback(() => {
```
dotnetjunky: This is exactly the same as `toggleShowStandardCallout`. What am I missing?

experiments/tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }, [showCustomizedCallout, setShowCustomizedCallout]);
+
+  const toggleShowCustomizedCalloutInner = React.useCallback(() => {
+    setShowCustomizedCallout(!showCustomizedCallout);
+  }, [showCustomizedCallout, setShowCustomizedCallout]);
+
+  const myRect: ScreenRect = { screenX: 10, screenY: 10, width: 100, height: 100 };
+  return (
+    <View>
+      <Text ref={stdBtnRef} style={fabricTesterStyles.testSection}>
```
dotnetjunky: I'm still not clear what is this ref for. Do you need to access `stdBtnRef` somewhere? It doesn't look like you do.

packages/components/Callout/api-extractor.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../../api-extractor.json"
```
dotnetjunky: We should use the package name syntax, like the `.eslintrc.js` file above.

packages/components/Callout/src/Callout.types.ts
```diff
@@ -0,0 +1,58 @@
...
+  | 'topRightEdge'
+  | 'rightTopEdge'
+  | 'rightCenter'
+  | 'rightBottomEdge'
+  | 'bottonLeftEdge'
+  | 'bottomAutoEdge'
+  | 'bottomCenter'
+  | 'bottomRightEdge';
+
+export interface ICalloutTokens extends IBackgroundColorTokens, IBorderTokens {
```
dotnetjunky: Callout doesn't have UI, why do you extend from IBackgroundColorTokens and IBorderTokens?

PPatBoyd: The beak needs a concept of background, border, even if we push it to child views to mind the main content colors

experiments/tester/src/RNTester/TestComponents/Common/styles.ts
PPatBoyd: ^ was my intention too, share nicely styled test pages for consistent LNF.  Mind if I push unifying the styles to a new work item?

FalseLobster: Yea, that's fine.  I almost did it myself but I felt like it was gonna snowball.  No reason we can't share the themed styles I wrote

PR #133, opened 2020-03-20
Add windows as a platform to repo - part 1

JasonVMo:
I'm breaking this into two changes to keep the changes manageable.  The parts that are most pertinent are:

- move the tester from experiments/test to apps/fluent-tester
- rename the tester package
- change the scripts bin name from just-scripts to fluent-scripts, this helps disambiguate things as yarn got confused at one point.
- add the windows, ios, and android files for react-native init to the fluent-tester package.  __Note that these need a bunch of coercing to work and aren't functional yet__
- added the capability of having different metro configs per platform to our metro script

There's still a bunch more work to do but given the file moves I wanted these changes in so the final changes are less disruptive.  Hopefully files within tester/src will register as moves correctly such that merges are not too painful.  They are effectively unchanged.


apps/fluent-tester/.flowconfig
```diff
@@ -0,0 +1,99 @@
...
+module.system.haste.paths.blacklist=<PROJECT_ROOT>/node_modules/react-native/Libraries/react-native/react-native-implementation.js
+module.system.haste.paths.blacklist=<PROJECT_ROOT>/node_modules/react-native/Libraries/Animated/src/polyfills/.*
+
+munge_underscores=true
+
+module.name_mapper='^[./a-zA-Z0-9$_-]+\.\(bmp\|gif\|jpg\|jpeg\|png\|psd\|svg\|webp\|m4v\|mov\|mp4\|mpeg\|mpg\|webm\|aac\|aiff\|caf\|m4a\|mp3\|wav\|html\|pdf\)$' -> 'RelativeImageStub'
+
+suppress_type=$FlowIssue
+suppress_type=$FlowFixMe
+suppress_type=$FlowFixMeProps
```
FalseLobster: What's the flowconfig for?

apps/fluent-tester/metro.config.js
```diff
@@ -0,0 +1,45 @@
...
+  ],
+  resolver: {
+    extraNodeModules: {
+      // Redirect react-native to react-native-windows
+      'react-native': rnwPath,
+      'react-native-windows': rnwPath
+    },
+    // Include the macos platform in addition to the defaults because the fork includes macos, but doesn't declare it
+    platforms: ['ios', 'android', 'windesktop', 'windows', 'win32', 'web', 'macos'],
+    providesModuleNodeModules: ['react-native-windows'],
```
dotnetjunky: will this work for win32? Currently it is set to @office-iss/react-native-win32

JasonVMo: I made a change so that if a bundle includes a platform it will try to load `metro.config.${platform}.js` before `metro.config.js`.  As a result in fluent-tester it will reroute to metro.config.win32.js

apps/fluent-tester/.flowconfig
JasonVMo: It was generated by the react-native init process.

JasonVMo:
I'm going to resubmit this.  I keep hitting one thing that exposes issues in other things and this change has gotten bigger.

PR #136, opened 2020-03-24
Add new fluent typeramp to fluent native text with variants

ksiler:


packages/components/Link/src/Link.settings.ts
```diff
@@ -7,8 +7,8 @@ export const settings: IComposeSettings<ILinkType> = [
   {
     tokens: {
       fontFamily: 'primary',
-      fontSize: 'medium',
-      fontWeight: 'medium',
+      fontSize: 'secondary',
+      fontWeight: 'regular',
       color: 'link'
```
FalseLobster: Should this use a variant?

packages/components/text/src/Text.settings.ts
```diff
@@ -5,9 +5,7 @@ import { IComposeSettings } from '@uifabricshared/foundation-compose';
 export const settings: IComposeSettings<ITextType> = [
   {
     tokens: {
-      fontFamily: 'primary',
-      fontSize: 'medium',
-      fontWeight: 'medium',
+      fontVariant: 'secondaryStandard',
```
FalseLobster: I would expect body to be the default for all the controls

packages/components/text/src/__tests__/__snapshots__/Text.test.win32.tsx.snap
```diff
@@ -5,9 +5,7 @@ exports[`Text all props 1`] = `
   style={
     Object {
       "color": "pink",
-      "fontFamily": "Segoe UI",
-      "fontSize": 12,
-      "fontWeight": "500",
+      "fontVariant": "bodyStandard",
```
FalseLobster: I would expect fontVariant to be stripped out in the snapshot

packages/framework/theming-ramp/src/Typography.types.ts
```diff
@@ -85,8 +83,7 @@ export type FontWeightValue = '100' | '200' | '300' | '400' | '500' | '600' | '7
  */
 export interface IFontWeights {
   light: FontWeightValue;
-  semiLight: FontWeightValue;
-  medium: FontWeightValue;
+  regular: FontWeightValue;
   semiBold: FontWeightValue;
   bold: FontWeightValue;
 }
```
FalseLobster: I'm wondering if all these font weights still make sense in the new type ramp

packages/utils/tokens/src/text-tokens.ts
```diff
@@ -1,15 +1,34 @@
...
+
+export interface ITextVariantTokens {
+  fontVariant?: keyof ITypography['variants'] | string;
+}
+
+export interface ITextStyleTokens {
+  fontFamily?: keyof ITypography['families'] | TextStyle['fontFamily'] | string;
+  fontSize?: keyof ITypography['sizes'] | TextStyle['fontSize'] | string;
+  fontWeight?: keyof ITypography['weights'] | TextStyle['fontWeight'] | string;
+}
```
FalseLobster: I'm not sure we want | string on all of these types?  Unless we want to do something similar to the colors object where we also make it a dictionary so that variants could be extended for a subtheme

FalseLobster: which might actually be a good idea

packages/framework/theming-ramp/src/Typography.types.ts
ksiler: Good point. Perhaps I should just keep 'regular' and 'semibold'? Also, should it be 'semibold' or 'semiBold'? I feel that 'semibold' is easier to type.

packages/components/text/src/Text.settings.ts
ksiler: Make sense given the name but this would increase the size from what it is currently (12px -> 14px).

packages/experiments/experimental-web-controls/src/Link/Link.settings.ts
```diff
@@ -5,8 +5,8 @@ export const settings: IComposeSettings<ILinkType> = [
   {
     tokens: {
       fontFamily: 'primary',
-      fontSize: 'medium',
-      fontWeight: 'medium',
+      fontSize: 'secondary',
+      fontWeight: 'regular',
```
ksiler: This may have been changed by mistake - do I want to edit these values in experimental-web-controls?

packages/components/text/src/Text.settings.ts
ksiler: Is that what we want?

packages/utils/tokens/src/text-tokens.ts
ksiler: So, should these be kept as is? Or is there additional work needed to make it behave like the colors object?

packages/components/Link/src/Link.settings.ts
ksiler: Probably. Updated this to use fontVariant in recent commits.

packages/framework/theming-ramp/src/Typography.types.ts
ksiler: Updated fontWeights to 'regular' and 'semiBold' in recent commits.

packages/components/text/src/__tests__/__snapshots__/Text.test.win32.tsx.snap
ksiler: Updated the snapshot to list fontFamily, fontSize, and fontWeight in recent commits.

experiments/tester/src/RNTester/TestComponents/Text/TextTest.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Separator } from '@fluentui-react-native/separator';
+import { StandardUsage } from './StandardUsage';
+import { CustomizeUsage } from './CustomizeUsage';
+import { styles } from './styles';
+import { TextWin32 } from '@office-iss/react-native-win32';
+
+export const TextTest: React.FunctionComponent<{}> = () => {
+  return (
+    <View>
+      <TextWin32 style={styles.section}>Standard Usage</TextWin32>
```
tkamireh: Does this need to be `TextWin32`. I think it can just be normal `Text`, right?

ksiler: Yeah that should be fine. I'll change it.

packages/components/text/src/Text.settings.ts
JasonVMo: Probably not.  Let's leave it as is until we get official redlines.

packages/experiments/experimental-web-controls/src/Link/Link.settings.ts
JasonVMo: I don't think so, I think experimental web-controls will probably be removed at some point.  I'd like to get react-native-web running, and have the tester be able to render there, at that point there will be no point to the experimental web controls.

If they are becoming a maintenance nightmare maybe we should remove them earlier.

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
```diff
@@ -28,6 +23,21 @@ function _defaultTypography(): ITypography {
...
+      bodyStandard: { face: 'primary', size: 'body', weight: 'regular' },
+      bodySemibold: { face: 'primary', size: 'body', weight: 'semiBold' },
+      subheaderStandard: { face: 'primary', size: 'subheader', weight: 'regular' },
+      subheaderSemibold: { face: 'primary', size: 'subheader', weight: 'semiBold' },
+      headerStandard: { face: 'primary', size: 'header', weight: 'regular' },
+      headerSemibold: { face: 'primary', size: 'header', weight: 'semiBold' },
+      heroStandard: { face: 'primary', size: 'hero', weight: 'regular' },
+      heroSemibold: { face: 'primary', size: 'hero', weight: 'semiBold' },
+      heroLargeStandard: { face: 'primary', size: 'heroLarge', weight: 'regular' },
+      heroLargeSemibold: { face: 'primary', size: 'heroLarge', weight: 'semiBold' }
```
JasonVMo: Given the discussion earlier about control text size, I wonder if we will want to add some extra variant values for things like controlText.  This allows a common reference point and allows people to author it directly.  Just a thought.  (but not something to address for this review)

packages/utils/tokens/src/text-tokens.ts
JasonVMo: Actually isn't the string part redundant?  I would expect that it picks up string from TextStyle['fontFamily'] and string | number from TextStyle['fontSize'].  Not blocking but good to test/update at some point.

PR #137, opened 2020-03-25
Add a mac project

HeyImChris:
- This adds the first macOS xcodeproj. 
- Update gitignore for a bunch of generated files
- Use xcconfig files for Debug and Release flavors. I noticed in more mature projects we also have "common" and "framework" xcconfigs to name a few. We don't need those yet, but when we do we can add them.
- Update xcconfigs and some other files to cooperate with cocoapod generated workspaces

This is just a blank project that launches a window when run. Later the plan is to pull in dependencies, add tests, and update CI for them, but first let's get a simple project in and verify we can build it in our CI loops.

This is a swift project to reflect what most of our consumers will be using.

#### Why not create one project with both macOS and iOS?
Because React-Native for macOS and iOS have build products with the same name (e.g. React.xcodeproj), we need separate projects to consume the iOS version and the macOS version.  This is the same reason we do this in React-Native-Community modules as well. If we ever renamed the macOS products to React_macOS.xcodeproj, for instance, then we could combine these into one.

#### How do I use this?
(same yarn commands as usual)
- `cd apps/mac`
- `pod install`
- launch/run the generated workspace


apps/mac/FluentUIReactNative-macOSUITests/FluentUIReactNative_macOSUITests.m
```diff
@@ -0,0 +1,48 @@
...
+//  FluentUIReactNative_macOSUITests.m
+//  FluentUIReactNative-macOSUITests
+//
+//  Created by Chris Hogan on 3/24/20.
+//  Copyright © 2020 Microsoft. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+
+@interface FluentUIReactNative_macOSUITests : XCTestCase
```
anandrajeswaran: Can we add tests in a separate change? Or at least not add dummy tests?

But also they probably should be swift?

.gitignore
```diff
@@ -11,6 +11,17 @@ pids
 *.seed
 *.pid.lock
 
+# Xcode
+*.xcuserstate
+**/.DS_Store
+
+# CocoaPods
+/apps/mac/Pods/*
```
anandrajeswaran: probably want macos as the directory name for consistency?

apps/mac/FluentUIReactNative-macOS/Info.plist
```diff
@@ -0,0 +1,36 @@
...
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIconFile</key>
+	<string></string>
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
anandrajeswaran: I'm missing where this is actually specified, but what's the domain we're using? Usually has some impact on device deployability

apps/mac/FluentUIReactNative-macOS/Info.plist
```diff
@@ -0,0 +1,36 @@
...
+	<key>CFBundlePackageType</key>
+	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSMinimumSystemVersion</key>
+	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>Copyright © 2020 Microsoft. All rights reserved.</string>
```
anandrajeswaran: Per other comments, we should scrub this copyright and make sure we get the right legalese

apps/mac/FluentUIReactNative-macOS.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,624 @@
...
+				INFOPLIST_FILE = "FluentUIReactNative-macOSTests/Info.plist";
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@loader_path/../Frameworks",
+				);
+				MACOSX_DEPLOYMENT_TARGET = 10.14;
+				PRODUCT_BUNDLE_IDENTIFIER = "Microsoft.FluentUIReactNative-macOSTests";
+				PRODUCT_NAME = "$(TARGET_NAME)";
+				TEST_HOST = "$(BUILT_PRODUCTS_DIR)/FluentUIReactNative-macOS.app/Contents/MacOS/FluentUIReactNative-macOS";
```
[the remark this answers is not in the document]
HeyImChris: This is the default just by creating an xcode proj and running pod install. Let's get a simple case in and we can always come back and add xcconfigs later

apps/mac/FluentUIReactNative-macOS/Info.plist
HeyImChris: This is the same copyright we use in Polyester, but with the year updated. Why do you think it's wrong?

anandrajeswaran: This is public GitHub. Polyester is internal and not strictly licensed that I know of

apps/mac/FluentUIReactNative-macOS/Info.plist
HeyImChris: By domain you mean? I haven't touched this file manually

apps/mac/FluentUIReactNative-macOS/AppDelegate.m
```diff
@@ -0,0 +1,27 @@
+//
+//  AppDelegate.m
+//  FluentUIReactNative-macOS
+//
+//  Copyright © 2020 Microsoft. All rights reserved.
+//
```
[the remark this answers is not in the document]
HeyImChris: I'm reusing the Polyester copyright info, but if that's not good enough here do you happen to have a reference to the MIT license?

apps/mac/FluentUIReactNative-macOS.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
anandrajeswaran: Seems like a simple thing to do now to avoid offloading the cost to the next poor person who tries to create a build setting?

apps/mac/FluentUIReactNative-macOS/Info.plist
anandrajeswaran: yeah we probably should though. Looks like the domain is in the build settings that are inlined in the pbxproj. We should choose it intentionally and not just accept the default

apps/mac/FluentUIReactNative-macOS/AppDelegate.m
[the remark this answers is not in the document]
anandrajeswaran: @markavitale do you have any guidance to reference?

apps/mac/FluentUIReactNative-macOSUITests/FluentUIReactNative_macOSUITests.m
HeyImChris: The test templates I have now make adding them that much easier when I get to that stage of this project. Don't see a benefit to deleting something I'll need later creating more work for myself?

apps/mac/FluentUIReactNative-macOS/AppDelegate.m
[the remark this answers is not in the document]
HeyImChris: The entire repository is tagged with MIT License and I'm not seeing other files in here with licensing information. Let me know what you mean and I'm happy to update this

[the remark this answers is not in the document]
markavitale: Really only guidance is that this is what's used in https://github.com/OfficeDev/ui-fabric-ios (also Apple code licensed under MIT). We aligned on that for all files in Fluent UI Apple.
```
//
//  Copyright (c) Microsoft Corporation. All rights reserved.
//  Licensed under the MIT License.
//
```

apps/mac/FluentUIReactNative-macOS/Info.plist
markavitale: FWIW, fabric-ios just removed the human readable copyright from the info plist completely.

apps/mac/FluentUIReactNative-macOSUITests/FluentUIReactNative_macOSUITests.m
anandrajeswaran: We're not ever going to add our UITests in a monolithic FluentUIReactNative_macOSUITests.m I hope - we would instead add more specific tests in better named files. And ideally we would do it in swift.

Plus the templates get autocreated if you create a target, so this is trivially recreatable. Figured just removing now would be easier than trying to come up with the name of one suite of tests we want so removing it seemed like the best way to scope down this change without adding known incorrect code.

HeyImChris: I'll recreate this all in swift. I won't add tests.

apps/mac/FluentUIReactNative-macOS/Info.plist
HeyImChris: I noticed fabric-macos has the copyright info still in this file

https://office.visualstudio.com/ISS/_git/fabric-macos?path=%2FOfficeUIFabricTestApp%2FOfficeUIFabricTestApp-Info.plist

HeyImChris: I'll go ahead and remove it though since fabric-ios has been in the public for longer, so more likely their legalese is correct.

markavitale: Yeah fabric-macos is not open source or public, so not a good example to follow.

apps/mac/FluentUIReactNative-macOS/Info.plist
HeyImChris: Not sure what you mean by the domain, but I tried to make the swift-conversion iteration of this mirror the initial fabric-macos change you sent out last year. The Info.plist _should_ be the same as it was in that PR.

apps/macOS/FluentUIReactNativeTester_macOS.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,394 @@
...
+			sourceTree = "<group>";
+		};
+/* End PBXVariantGroup section */
+
+/* Begin XCBuildConfiguration section */
+		9F8DD79D242C033700C5877B /* Debug */ = {
+			isa = XCBuildConfiguration;
+			baseConfigurationReference = 9F8DD7A3242C0FFE00C5877B /* FluentUIReactNative_debug.xcconfig */;
+			buildSettings = {
+				ALWAYS_SEARCH_USER_PATHS = NO;
```
anandrajeswaran: Don't know how to reactivate a comment thread, but can we move these to an xcconfig now that you've created one. Build settings really need specific review/intention, but that's difficult to do when it's duplicated so many places and clearly not intentional yet

apps/macOS/FluentUIReactNativeTester_macOS.xcodeproj/project.pbxproj
```diff
@@ -17,14 +17,15 @@
...
 		3764647238B3CE87869651A7 /* libPods-FluentUIReactNativeTester_macOS.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "libPods-FluentUIReactNativeTester_macOS.a"; sourceTree = BUILT_PRODUCTS_DIR; };
 		3D3F8D374CE12B9C46ED1519 /* Pods-FluentUIReactNativeTester_macOS.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-FluentUIReactNativeTester_macOS.release.xcconfig"; path = "Target Support Files/Pods-FluentUIReactNativeTester_macOS/Pods-FluentUIReactNativeTester_macOS.release.xcconfig"; sourceTree = "<group>"; };
-		9F8DD791242C033500C5877B /* FluentUIReactNativeTester_macOS.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = FluentUIReactNativeTester_macOS.app; sourceTree = BUILT_PRODUCTS_DIR; };
+		9F8DD791242C033500C5877B /* .app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = .app; sourceTree = BUILT_PRODUCTS_DIR; };
```
anandrajeswaran: This looks inadvertently broken?

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_debug.xcconfig
```diff
@@ -3,7 +3,22 @@
...
+CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES_AGGRESSIVE
+CLANG_WARN_BOOL_CONVERSION = YES
+CLANG_WARN_CONSTANT_CONVERSION = YES
+CLANG_WARN_INFINITE_RECURSION = YES
+CLANG_WARN_INT_CONVERSION = YES
+CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE
+CLANG_WARN_UNREACHABLE_CODE = YES
+CLANG_WARN__DUPLICATE_METHOD_MATCH = YES
+GCC_WARN_UNUSED_FUNCTION = YES
+GCC_WARN_UNUSED_VARIABLE = YES
```
anandrajeswaran: Why wouldn't any of these also apply to release? Do they stop working with optimization?

Particularly interesting is unused function/variable warnings, which could easily be distinct in release vs. debug with debug only code

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_debug.xcconfig
```diff
@@ -3,7 +3,22 @@
...
+SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
+CLANG_ANALYZER_NONNULL = YES
+CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES_AGGRESSIVE
+CLANG_WARN_BOOL_CONVERSION = YES
+CLANG_WARN_CONSTANT_CONVERSION = YES
+CLANG_WARN_INFINITE_RECURSION = YES
+CLANG_WARN_INT_CONVERSION = YES
+CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE
+CLANG_WARN_UNREACHABLE_CODE = YES
+CLANG_WARN__DUPLICATE_METHOD_MATCH = YES
```
anandrajeswaran: extra underscore?

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_common.xcconfig
```diff
@@ -0,0 +1,14 @@
...
+//
+
+PRODUCT_MODULE_NAME = FluentUIReactNativeTester
+SWIFT_MODULE_NAME = FluentUIReactNativeTester
+
+ALWAYS_SEARCH_USER_PATHS = NO
+CLANG_ENABLE_MODULES = YES
+
+// Versions
+MACOSX_DEPLOYMENT_TARGET = 10.14
```
anandrajeswaran: Our devmain apps need to still support 10.13. Suspect we'll want to match that 'N-2' support on macOS, 'N-1' support on iOS (N = current OS version)

anandrajeswaran: Maybe can handle in a follow when we figure out how to deal with ios: I think this is the only platform-specific config, right? Whereas all the other stuff applies to both macOS and iOS and thus should be shared with iOS?

Again, probably best as a "file away and remember" instead of architecting too strictly now before we know what the iOS/macOS split looks like unless you can think of something both cheap and flexible

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_debug.xcconfig
HeyImChris: Debug should catch all of these issues so release can just build and run without extra checks running in the background

HeyImChris: If we expect to have debug or release macros then it makes sense to move some of these to common. I wasn't planning on that, but I suppose it's low harm even if we don't

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_debug.xcconfig
HeyImChris: https://xcodebuildsettings.com/ says it needs double underscore

apps/macOS/FluentUIReactNativeTester_macOS/Config/FluentUIReactNative_common.xcconfig
HeyImChris: I'll think about it but agree it shouldn't stop this from going in

apps/macOS/Podfile
```diff
@@ -0,0 +1,5 @@
+target 'FluentUIReactNativeTester_macOS' do
+  platform :osx, '10.14'
```
anandrajeswaran: mismatch with MACOSX_DEPLOYMENT_TARGET?

anandrajeswaran: Or is this the active development target? If so, should be 10.15?

PR #141, opened 2020-03-29
Add macos debug/release build validation to CI

HeyImChris:
### Problem
Before checking in meaningful code, we need the pipeline stood up to validate changes.

### Fix
- Add to the existing publishing script 1) cleaning of derived data, 2) cocoapods installation, 3) project pod installation, 4) xcode version installation, 5) debug and release builds of our project.

For #5 above, we can use a template in `apple-xcode-build.yml` to reuse debug and release build processes.

### Open Questions/Work
1) Do we want to rename `azure-pipelines.yml` to something more descriptive like `azure-pipelines-apple.yml`?
2) Still need to add tests to the projects and then update the CI
3) When we add iOS, I'll probably add another layer of abstraction on this so macOS/iOS can call the same script and pass in platforms as parameters.


HeyImChris:
> From my standpoint this looks good. I do think a natural next step might be figuring out how to split the CI loop to parallelize onto multiple machines as it is up to 9 minutes now.
> 
> Looking at the overall time it looks like the yarn build step takes a really long time, I think primarily because jest is really slow when there is no caching happening. If that step is not a prerequisite for bundling then we can potentially run the build loop on an ubuntu image in parallel.

That sounds great! We should definitely iterate on this to optimize parallel runs and yarn build is something we should be able to do so with. Other areas we could parallelize are iOS vs. Mac (once we add iOS) and debug vs. release. With no tests and basically no executing code it's not a problem now but I could see it being one in the future if it's all running serially.

PR #145, opened 2020-04-02
Bundling, jest improvements, adding windows tester

JasonVMo:
So this is a good point to check this in to avoid brutal merges in the future.  There are a couple of components of this change:

## Add @fluentui-react-native/build-tools package

This is built out of packages/utils/build-tools right now.  Eventually I'd like to move it out of the repository but I want to further vet and test it before that happens.  This takes a number of the repository tools from scripts, converts them to typescript, and makes them eventually available in other repos (where there is a lot of similar code).  Included in here:

### metroTask implementation
This changes the bundle description format in package.json to make it more of a unified approach with per platform forks.  See apps/fluent-tester/package.json for an example.  Usage in apps/fluent-tester looks like:

- `yarn bundle` - bundles all versions for that folder.  Note that this shows a haste error on windows but doesn't seem to block anything.
- `yarn bundle --platform win32` - bundle just the win32 version, or any other version
- `yarn bundle --dev` - bundle in development mode
- `yarn bundle --server --platform win32` - start a metro server, configured for a given platform

### updated metro configurations

Metro configuration is tricky for out of tree platforms.  Base metro is designed to bundle out of one RN version/location for all platforms.  This works for ios/android/web but not windows/win32/macos.  To make it work it needs to know the platform target, there is some trickery in here to either target a specific platform or allow the task to drive the platform bits.

This also now bundles directly from TS files rather than JS files.  This should make live reloading work without an extra tsc -watch process.

You can also have a metro.config.win32.js to allow for some win32 specific settings.  This works for any platform.

### updated jest react native configurations

The issue with jest (well one issue) was that babel-jest wasn't always picking up the babel settings for the metro-react-native-babel-preset.  Inlining this made it more reliable.  I enabled tests for Separator and Text just as proofs of concept.  At some point it might be good to enable jest to make multiple passes for each platform.  E.g. having jest.config.js and jest.config.win32.js and so on.  In jest.config.js files the platform can be specified if you want to test react-native.

### various other utilities

A bunch of utilities around getting enabled react native platform versions and things of this nature were added.  This also removes some code from scripts.

## Experiments/tester -> apps/fluent-tester

This moves the tester to fluent-tester and adds the windows specific bits to allow a windows project to be run out of the same project.

- currently bundling for win32 and windows both work from the same tester code
- the windows bits will build (or at least did) via the VS project in the windows directory
- currently the windows project doesn't seem to pick up the bundle correctly.  I routed the windows bundling to the same location as the default but need to investigate more how to do this
- __someone else who is interested can definitely take a look at this__

The yarn run-win32 workflow should still work

## Made the packages platform agnostic

We had a lot of unnecessary win32 code in various places.  Once I fixed the bundling this raised a bunch of issues.  There are a number of files across the repo that have been made platform agnostic.  In general if you are bundling for react-native-windows you can't import from react-native-win32, even if it is just types.

## Reworked interactive-hooks

I pulled in the source code from react-native 0.63, in particular Facebook's pressable implementation.  This has a component called pressable and a hook called pressability.  I brought in the pressability code, converted it to typescript, then got it up and running.  I then reworked the hooks package to be based off of this package.  Some notes:

- this __seems__ to work, it bears more testing
- other platforms should 'just work' though I have some concerns that not all native APIs are correctly exposed on view props and the filters in utils/adapters might strip them out
- I kept the pressability interface almost exactly as-is, the goal being to switch to the FB implementation once we are on 0.63


dotnetjunky:
Which files should we focus on reviewing? There are 314 changed files, which is intimidating. :)

JasonVMo:
> Which files should we focus on reviewing? There are 314 changed files, which is intimidating. :)

Realistically I would focus on things under:
/packages/utils/build-tools
/packages/utils/interactive-hooks

For the rest, a lot of them are just template additions that aren't very interesting.

.gitignore
```diff
@@ -80,4 +80,7 @@ typings/
 .next
 
 # document generation directory
-.docz
\ No newline at end of file
+.docz
+
```
FalseLobster, line 84: I don't have the latest changes from your branch but at one point it seemed like the Windows bundle of the test app wasn't being ignored by  git

packages/utils/interactive-hooks/src/Pressability/CoreEventTypes.ts
```diff
@@ -0,0 +1,168 @@
...
+    clientY: number;
+    pageX: number;
+    pageY: number;
+    timestamp: number;
+  }>
+>;
+
+export type KeyPressEvent = SyntheticEvent<
+  Readonly<{
+    key: string;
```
FalseLobster, line 164: Did you also get KeyPressEvent from Facebook?

.gitignore
JasonVMo: Yeah I moved the bundle location to be under dist to get it ignored for right now

apps/fluent-tester/.gitignore
```diff
@@ -0,0 +1,69 @@
...
+# screenshots whenever they are needed.
+# For more information about the recommended setup visit:
+# https://docs.fastlane.tools/best-practices/source-control/
+
+*/fastlane/report.xml
+*/fastlane/Preview.html
+*/fastlane/screenshots
+
+# Bundle artifact
+*.jsbundle
```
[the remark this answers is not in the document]
JasonVMo: We probably want .bundle to be ignored as well, that seems to be the extension used

packages/utils/interactive-hooks/src/Pressability/CoreEventTypes.ts
JasonVMo: Yes, this is the internal event they use.

PR #149, opened 2020-04-06
Add accessibility documentation

ksiler:
This adds a page for accessibility under docs.


packages/components/Link/src/Link.settings.ts
```diff
@@ -1,45 +1,45 @@
-import { linkName, ILinkType } from './Link.types';
-import { IComposeSettings } from '@uifabricshared/foundation-compose';
-// import { IViewWin32Props } from '@office-iss/react-native-win32';
-import { IViewProps } from '@fluentui-react-native/adapters';
-
-export const settings: IComposeSettings<ILinkType> = [
```
samuelfreiberg, line 6: Why are all these files being touched in this PR, isn't it just adding the a11y doc? (Correct me if I'm wrong please :) )

ksiler: You're right - this should only be adding a page for accessibility. I have no idea why it's including these changes :( I'll look into it

samuelfreiberg: Don't worry I have these problems too! I would just go to your master branch, git diff against upstream master and make sure it's the same (pull if necessary). Then create a new branch from there and add this a11y page

PR #152, opened 2020-04-06
Update immutable merge API

JasonVMo:
As I was looking at the theming stuff I ran into some issues with configurability and chaining of immutable merge.  Looking at it more it seemed like the API was:

- confusing and hard to understand
- not particularly precise.  By that it registered style to a style handler but that was a simple key match that would happen at any level.

As a result I reworked the API of the merge package to (hopefully) be more usable.  The API now works on pattern matching of names or types in a hierarchy.  So to match a collection of props, that might each have a style set and to make arrays append rather than merge would look like:

```ts
immutableMergeCore({
  // treat tokens as special, don't resolve styles or deeply merge, just merge and don't recurse
  tokens: 0,

  // any other object treat as follows
  object: {
    // any array types should use this function to merge
    array: (...arrays: any[]) => { return [].concat(...arrays); },

    // styles use a style resolver to merge
    style: mergeStyle
  }
}, slotProps1, slotProps2, slotProps3);
```

There are more configuration options as well such as calling `immutableMergeCore(2, obj1, obj2)` to recurse twice.  I wrote up a bunch of documentation on it.

It may be worth considering renaming the package to something other than @uifabric.  Maybe switch it to @fluentui-react-native/immutable-merge?

This also enables the button snapshot test.


packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -1,43 +1,120 @@
...
-  return t && typeof t === 'object' && !Array.isArray(t) && (!onlyReturnNonEmptyObjects || Object.getOwnPropertyNames(t).length > 0);
+export type RecursionHandler = BuiltinRecursionHandlers | CustomRecursionHandler;
+
+/**
+ * configuration object for the merge, key names are matched with a few exceptions:
+ * - object: matches non-array object types
+ * - array: matches array types
+ * - [key: string]: matches anything by name
+ */
+export interface MergeOptions {
```
tkamireh, line 37: Are we not fans of the `I` prefix on interfaces and types anymore?

packages/framework/immutable-merge/README.md
```diff
@@ -3,82 +3,183 @@
 This package provides a relatively concise routine to handle merging multiple objects together with the following characteristics:
 
 - No modifications will be made to any object
-- The resulting object will have the minimum number of updates. If only one value is updated three levels deep, only that value and the chain of containing objects will be recreated.
+- Minimal updates. If only one value is updated three levels deep, only that value and the chain of containing objects will be recreated.
 - Empty objects or undefined objects will be ignored and not cause a new branch to be created.
 - Recursion is controllable in a variety of ways
 
-## IMergeOptions
+Note that this does not provide a **strict** immutable package on its own. It also doesn't operate or return only readonly objects. This is to provide flexibility. It could easily be wrapped in various ways to provide that type of functionality, but it is provided in a more flexible form to be useful in other scenarios as well.
```
tkamireh, line 10: When you say it doesn't strictly operate on readonly/immutable objects, is that a consequence of the depth of the merge (some property of some property of some property might get mutated across the result and one of the initial objects after the merge)?

packages/framework/immutable-merge/README.md
```diff
@@ -3,82 +3,183 @@
...
 
-The array of objects or undefined values (internally treated as anything falsy) should be processed, returning a single object or undefined.
+- all values at a given level will overwrite, with the last writer winning
+- if a key does not exist for an object it is ignored
+- if a key does exist, even if it is `undefined` it will replace the previous value
+- only non-array objects will recurse, arrays will be replaced rather than appended
+- keys which exist and have a value of `undefined` will be deleted
 
-The key and options parameters are provided as conveniences in the case that a single handler needs to differentiate different branches or know how deep it is in the tree. In many cases these can be ignored.
+The peculiar pattern of deleting keys which end up as undefined is the only way to delete a key without violating the core principles. An example of key deletion might look like:
```
tkamireh, line 28: If this is a common enough scenario, do we just want to add it as option in `MergeOptions`. That is allowing specification of keys to delete or not include in the final resultant object.

packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -51,76 +128,78 @@ export type IMergeRecursionHandler = (options: IMergeOptions, ...objs: any[]) =>
  * be `{}` - no recursion, `{ depth: -1 }` - recurse infinitely
  * @param objs - an array of objects to merge together
  */
-export function immutableMergeCore(options: IMergeOptions, ...objs: any[]): any {
-  const setToMerge = objs.filter(v => _isNonArrayObject(v, true));
-  const { depth = 0, recurse } = options;
-  const processSingle = options.processSingles && setToMerge.length === 1;
+function immutableMergeWorker(mergeOptions: RecursionOption | MergeOptions, singleMode: boolean, ...objs: any[]): any {
```
tkamireh, line 131: I remember there being a discussion on why it was very difficult to type parameterize some of these functions, requiring the use of `any`, but I don't quite remember the details. Can you expand on that a bit?

tkamireh, line 131: You may want to add to the comment an @ param for singleMode

packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -51,76 +128,78 @@ export type IMergeRecursionHandler = (options: IMergeOptions, ...objs: any[]) =>
...
+  const setToMerge = objs.filter(v => v && getEntityType(v) === 'object' && Object.getOwnPropertyNames(v).length > 0);
```
tkamireh, line 132: Why do you need the `getOwnPropertyNames().length` check? It seems like it could be problematic depending on how an object is constructed.

packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -51,76 +128,78 @@ export type IMergeRecursionHandler = (options: IMergeOptions, ...objs: any[]) =>
...
           delete result[key];
         }
       }
     }
+
     // in the single processing case return the original if nothing changed, otherwise return result
-    return processSingle && !hasChanged ? setToMerge[0] : result;
+    return result || processSet;
   }
   return setToMerge.length > 0 ? setToMerge[0] : undefined;
```
tkamireh, line 169: Should you warn or something if you get here. Processing a single with a set > 1 breaks the api contract right?

packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -51,76 +128,78 @@ export type IMergeRecursionHandler = (options: IMergeOptions, ...objs: any[]) =>
...
-
-            // either recurse to this function or call the custom handler if specified
-            const handler: IMergeRecursionHandler = customHandler || immutableMergeCore;
-            result[key] = handler(nextOptions, ...collectedObjects);
-            hasChanged = hasChanged || result[key] !== originalVal;
+    let result = processSingle ? undefined : Object.assign({}, ...setToMerge);
+    const processSet = result || setToMerge[0];
+
+    for (const key in processSet) {
+      if (processSet.hasOwnProperty(key)) {
```
tkamireh, line 143: Similar concerns to getOwnPropertyNames. This probably works for all of our use cases I do wonder about people doing things directly via prototypal inheritance and whether the results end up being wrong.

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: It looks like Fluent/Fabric is moving away from I prefixes and will be changing everything.

packages/framework/immutable-merge/README.md
JasonVMo: I primarily mean that it doesn't have any enforcement.  The types are not required to be readonly and there aren't ways of hiding properties under functions like some libraries out there.  It just follows the principles above.

packages/framework/immutable-merge/README.md
JasonVMo: MergeOptions is really only pattern matching via index signatures.  I could add a special name, maybe to change behavior for undefined but that is getting weird.

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: I don't remember many of the issues, the public facing functions are not typed to any as they expect sets of object and have a type to coerce similarity between inputs (if someone is using inference)

Realistically though, the internal worker function use of any is somewhat accurate as we are walking and type checking against various object types, including filtering sets of inputs to only merge ones that are object based.

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: The goal is to omit empty objects.  Basically if you call immutableMerge({}, a, {}, {}) it should return a, not { ...a }.  Do you think Object.keys().length would be better?

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: Not necessarily.  The process single mode can still do multiple objects, it is just that in cases where there is a single branch it will keep processing rather than bailing early.  The signature is still ...objs: (T | undefined)[] on that function.

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: I believe this is the correct way to do it.  I just looked at the polyfill for Object.assign and it has the following:
```js
__assign = Object.assign || function(t) {
        for (var s, i = 1, n = arguments.length; i < n; i++) {
            s = arguments[i];
            for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
                t[p] = s[p];
        }
        return t;
    };
```

TypeScript also has a lint error for unprotected key usages within in statements which is why I started adding these when doing the theming stuff.  Given the standard polyfill I think this is correct, it is simply that if your object can't be assigned (e.g. if it is a class) then it can't be merged.

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: Good catch, the comment was out of date.

JasonVMo: I just checked again on the any vs object core type.  I changed the internal type to a generic and it looks like the inferencing chains correctly.

packages/framework/immutable-merge/README.md
tkamireh: Yeah I agree, it would be weird to do without making MergeOptions a heavier object with a full schema. The pattern is probably fine.

packages/framework/immutable-merge/src/Merge.ts
tkamireh: Given how the Object.assign polyfill, maybe the problems I'm thinking about are exceedingly rare. I think this is fine.

PR #156, opened 2020-04-09
Svg

warren-ms:
This adds dependencies and a test page for react-native-svg support. Win32 support for react-native-svg primitives are already checked into netui so this will light up support here.

Metro config files only allow for a single transformer to be specified so I had to create a transformer called transform-selector to select between react-native-transformer and the regular metro transformer depending on file extension. React-native-transformer transforms the xml from a svg file to javascript react-native calls at bundle-time so it doesn't have to be done at run-time.

A test page has been added to RNTester with examples for inline svgs, remotely retrieved svgs, and bundled svgs.


packages/utils/build-tools/src/configureMetro.ts
```diff
@@ -7,6 +7,7 @@ import { getRNVersion, getAllPlatforms, getAllReactNativePaths, PlatformValue, f
 import { getPackageInfo, findGitRoot } from 'just-repo-utils';
 import blacklist from 'metro-config-60/src/defaults/blacklist';
 import { mergeConfigs } from './mergeConfigs';
+import { getDefaultConfig } from 'metro-config-60';
```
JasonVMo, line 10: I need to find the right way to make this version agnostic.  It may need a plugin module similar to haul for 0.60, 0.61, etc.  (obviously not a comment on this change)

packages/utils/build-tools/src/configureMetro.ts
```diff
@@ -66,29 +67,30 @@ export function addPlatformMetroConfig(platform: PlatformValue, base: any = {}):
...
-      ...getPackageInfo()
-        .dependencies()
-        .paths()
-    ],
+    watchFolders: [path.resolve(findGitRoot(), 'node_modules'), ...getPackageInfo().dependencies().paths()],
     resolver: {
-      resolverMainFields: ['react-native', 'browser', 'main']
+      resolverMainFields: ['react-native', 'browser', 'main'],
+      assetExts: assetExts.filter((ext) => ext !== 'svg'),
+      sourceExts: [...sourceExts, 'svg'],
```
JasonVMo, line 81: remove trailing ,.  Do you have prettier settings turned on for vs-code?

packages/utils/build-tools/src/configureMetro.ts
```diff
@@ -66,29 +67,30 @@ export function addPlatformMetroConfig(platform: PlatformValue, base: any = {}):
...
+      babelTransformerPath: require.resolve('./transform-selector'),
 
       getTransformOptions: async () => ({
         transform: {
           experimentalImportSupport: false,
-          inlineRequires: false
-        }
-      })
+          inlineRequires: false,
+        },
```
JasonVMo, line 90: remove trailing ','

packages/utils/build-tools/src/configureMetro.ts
```diff
@@ -66,29 +67,30 @@ export function addPlatformMetroConfig(platform: PlatformValue, base: any = {}):
...
+      }),
     },
-    resetCache: false
+    resetCache: false,
```
JasonVMo, line 93: ditto

packages/utils/build-tools/src/configureMetro.ts
```diff
@@ -66,29 +67,30 @@ export function addPlatformMetroConfig(platform: PlatformValue, base: any = {}):
...
 
       getTransformOptions: async () => ({
         transform: {
           experimentalImportSupport: false,
-          inlineRequires: false
-        }
-      })
+          inlineRequires: false,
+        },
+      }),
```
JasonVMo, line 91: ditto

packages/utils/build-tools/src/transform-selector.ts
```diff
@@ -0,0 +1,10 @@
+import upstreamTransformer = require('metro-react-native-babel-transformer');
+import svgTransformer = require('react-native-svg-transformer');
```
JasonVMo, line 2: This is kind of a strange mix of TS and JS module usage here.  I'm surprised you aren't getting lint/build errors here during the typescript build.  Try changing these to:
```ts
import * as upstreamTransformer from 'metro-react-native-babel-transformer';
import * as svgTransformer from 'react-native-svg-transformer';
```

packages/utils/build-tools/src/transform-selector.ts
```diff
@@ -0,0 +1,10 @@
...
+
+module.exports.transform = function({ src, filename, options }) {
```
JasonVMo, line 4: You should be able to convert this to:
```ts
export function transform({ src, filename, options }) {
```
rather than using the common JS style module syntax.

packages/utils/build-tools/src/configureMetro.ts
warren-ms: Yes I do have it on. Are we opposed to its opinions? 😆

I couldn't find a way to turn it off, clicking on formatting in the status bar doesn't do anything and an extension to disable it that I downloaded doesn't work very well. How is everyone else doing it?

warren-ms: I see prettier is just an extension that is installed by default. I just uninstalled it.

warren-ms: Ok something else is formatting it upon save. Prettier is uninstalled, formatOnSave is disabled but it still formats whenever I ctrl+s.

warren-ms: Oh I found it. The checked in .vscode/settings.json has this line set:

  "editor.formatOnSave": true,

It's overriding vscode global settings.

packages/utils/build-tools/src/transform-selector.ts
warren-ms: using `export function transform({ src, filename, options }) {`
results in this error when bundling:

`@fluentui/react-native: [3:35:44 PM] x Error detected while running 'metro'
@fluentui/react-native: [3:35:44 PM] x ------------------------------------
@fluentui/react-native: [3:35:44 PM] x TypeError: transformer.transform is not a function
@fluentui/react-native:     at D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:231:31
@fluentui/react-native:     at Generator.next (<anonymous>)
@fluentui/react-native:     at asyncGeneratorStep (D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:75:24)
@fluentui/react-native:     at _next (D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:95:9)
@fluentui/react-native:     at D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:100:7
@fluentui/react-native:     at new Promise (<anonymous>)
@fluentui/react-native:     at D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:92:12
@fluentui/react-native:     at JsTransformer.transform (D:\repos\fluentui-react-native-fork\node_modules\metro\src\JSTransformer\worker.js:380:7)
@fluentui/react-native:     at Object.<anonymous> (D:\repos\fluentui-react-native-fork\node_modules\metro\src\DeltaBundler\Worker.js:102:38)`

warren-ms: This seems to have just been some kind of weird build issue. It works now my last yarn build/bundle.

PR #157, opened 2020-04-09
Just makes the JS code more resilient to avoid RedBoxing on macos

FalseLobster:
This should pass in `undefined` for the second argument of ThemingModuleHelpers.ts createThemingModuleHelper which should be handled gracefully on macos.


change/@uifabricshared-theming-react-native-2020-04-09-11-38-21-fixmactheming.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "🐛 Make ThemingModule not crash on Mac",
```
tom-un, line 3: nit: its really *all* platforms will RedBox currently unless the Theming native module is linked.

PR #160, opened 2020-04-13
Adding RadioGroup/Checkbox Snapshot Tests

samuelfreiberg:
Adding RadioGroup and Checkbox snapshot testing and configuring each with Jest testing framework .


samuelfreiberg:
> Checkbox

@JasonVMo I believe I was following the naming convention from other Component test files, such as Separator. Separator's test file also has the .win32. Should we remove this for all?

FalseLobster:
> I was following the naming convention from other Component test files

I think each platform needs a different snapshot output for diffing purposes, at the time we really only had one platform implementation of these, so I likely wrote the tests this way out of laziness.  If you look at what we did in Polyester there's

Foo.test.win32.ts
Foo.test.android.ts
Foo.test.iOS.ts
etc...

all of those are just `import * from Foo.test.shared` which has the actual code.  Eventually we should move the JSX into a shared file and just make the platform specific ones import from there.

PPatBoyd:
change files needed (failed PR build)

PR #161, opened 2020-04-14
Stand up the iOS FluentUI-React-Native project

HeyImChris:
Use CLI to run `npx react-native init FluentUIReactNative-iOS` and get the boiler plate for our iOS test app.

- Adding the test project in its own directory because the CLI tools have a dependency on the node_modules location. To not break Windows by moving them in the fluent-tester directory, I'll keep the iOS build here (at least for now)
- Planning to move the macos project into the ios directory (which will get renamed to `fluent-tester_apple`) and house the Apple projects
- Lower cased the `iOS` and `macOS` directories to align with how React-Native cases their platform directories
- Remove pbxproj build settings and add them to custom xcconfigs for Debug and Release
    - Opted to just use Debug.xcconfig and Release.xcconfig as names because it's getting really long and it's obvious what project they're apart of due to their location
    - Hook up xcconfigs to generated pod config files so we can use a workspace (which is required by react-native)
- To allow Swift, we need to have any .swift file so we can have a bridge. I don't fully understand _why_ we need it, but we fail to find swift libraries without it.
     - Right now this is just using obj-c since that's the default RN gives us, but I plan to change it to swift when I add the first test once I get this bigger change in
- Update git ignore

Next:
iOS
- Update the CI to use this project for iOS
- Make tests run in Swift
- Update RN dependency to Microsoft's RN?

macOS
- Move the macOS project to this directory and update it to depend on Microsoft RN
- Update macOS CI for the new project location

Screen Shot 2020-04-13 at 5 15 52 PM


apps/ios/ios/File.swift
```diff
@@ -0,0 +1 @@
+// Necessary to allow Swift framework bridging
```
anandrajeswaran, line 1: Dummy.swift is generally the name we've used for this

apps/ios/ios/FluentUIReactNativeTester_iOS-Bridging-Header.h
```diff
@@ -0,0 +1,3 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
```
anandrajeswaran, line 2: Do we actually need a bridging header?

apps/ios/ios/FluentUIReactNativeTester_iOS/Info.plist
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleDisplayName</key>
+	<string>Hello App Display Name</string>
```
anandrajeswaran, line 8: Update

apps/ios/ios/FluentUIReactNativeTester_iOS/Info.plist
```diff
@@ -0,0 +1,57 @@
...
+	</dict>
+	<key>NSLocationWhenInUseUsageDescription</key>
+	<string></string>
+	<key>UILaunchStoryboardName</key>
+	<string>LaunchScreen</string>
+	<key>UIRequiredDeviceCapabilities</key>
+	<array>
+		<string>armv7</string>
+	</array>
+	<key>UISupportedInterfaceOrientations</key>
```
anandrajeswaran, line 48: We should support UpsideDown too - particularly interesting for iPad since that orientation support is (oddly) required to support multi-window

apps/ios/ios/FluentUIReactNativeTester_iOS/Info.plist
```diff
@@ -0,0 +1,57 @@
...
+	<array>
+		<string>UIInterfaceOrientationPortrait</string>
+		<string>UIInterfaceOrientationLandscapeLeft</string>
+		<string>UIInterfaceOrientationLandscapeRight</string>
+	</array>
+	<key>UIViewControllerBasedStatusBarAppearance</key>
+	<false/>
```
anandrajeswaran, line 55: Really, no? I would think we want this

apps/ios/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "scripts": {
+    "android": "react-native run-android",
+    "ios": "react-native run-ios",
+    "start": "react-native start",
+    "test": "jest",
+    "lint": "eslint ."
+  },
+  "dependencies": {
+    "react": "16.11.0",
+    "react-native": "0.62.2"
```
anandrajeswaran, line 14: Don't we want this to be 0.60 or so since our fork of react-native is what it should be consuming and that's on 0.60?

apps/ios/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "FluentUIReactNativeTester_iOS",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "android": "react-native run-android",
```
anandrajeswaran, line 6: can platforms not named ios be removed from here?

apps/ios/package.json
anandrajeswaran: Which leads to my dumb question - where do we specify we want react-native from the Microsoft fork?

apps/ios/ios/FluentUIReactNativeTester_iOS/Info.plist
markavitale: It makes more sense now that there is no home button and iPads are just symmetrical panes of glass

.gitignore
```diff
@@ -16,8 +16,10 @@ pids
 **/.DS_Store
 
 # CocoaPods
-/apps/macOS/Pods/*
-/apps/macOS/Podfile.lock
+/apps/macos/Pods/*
+/apps/macos/Podfile.lock
+/apps/ios/ios/Pods/*
+/apps/ios/ios/Podfile.lock
```
markavitale, line 22: Many projects I've seen check in their Podfile.lock much like a package-lock.json so that the versioning stays consistent between builds and between machines unless you intentionally make an update call. Is it a react native pattern we're following to ignore the lock files?

apps/ios/ios/FluentUIReactNativeTester_iOS-Bridging-Header.h
markavitale: For an app target, bridging headers are safe, but agreed there's not a ton of value to an empty one unless we have immediate plans to put headers into it.

apps/ios/ios/FluentUIReactNativeTester_iOS/Info.plist
HeyImChris: Nice catch. Yes is what we want here (i.e. delete it and let the default value reign supreme)

.gitignore
HeyImChris: Not that I know of. Just a general rule to not check in generated files, but agree that it'll be more reliable checking this in

markavitale: Yeah I think package manager lock files are usually an exception to that rule.

apps/ios/package.json
HeyImChris: That'll have to come later (added it to the change notes). I haven't hooked it up yet which is great for iOS, but obviously is a requirement for macOS. A good question is do we even _want_ fluentui-rn for iOS to pull in our fork. Ideally our fork would only ship mac anyway, so long term we'd want to depend just on fb's implementation

anandrajeswaran: Started a conversation offline, but I think we definitely do to keep macOS and iOS aligned and because I'm concerned deforking entirely will increase our merge cost a trivial amount (assuming we resolve the major diffs between repos which we need to do anyway) at the expense of reducing the speed at which we can identify a fix in react-native needed in devmain and get it to devmain

HeyImChris: Yea I'm on board with that. Will plan to consume our fork of RN for both ios and macos

PR #162, opened 2020-04-15
Move macOS and iOS test apps to a shared folder

HeyImChris:
House iOS and macOS test apps in the same directory

iOS and macOS will have a good amount of shared code, so although we need different projects (due to React-Native package naming restrictions), we can still have them be in sibling folders.

This change is a rename of the outermost `ios` folder to `apple` and a shift of the macos project to be a sibling of the `ios` one. MacOS still doesn't have the React-Native dependency which comes next.


apps/apple/ios/FluentUIReactNativeTester_iOS.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
anandrajeswaran, line 1: We should probably .gitignore this file?

HeyImChris: It should be already with `**/project.xcworkspace/contents.xcworkspacedata`.  Something might be off with the formatting though

PR #163, opened 2020-04-16
Fix package publishing

JasonVMo:
Internally our typescript files expose the typescript index.ts as the entry point for the package.  This makes debugging and bundling work directly against the typescript files but is actually incorrect for published packages.

Ken made some changes to beachball to support a prepublish hook, this adds a section in package.json called "onPublish" which will apply items within as part of the publish step.

Then the various packages had onPublish sections added to reference the eventual paths.  I ended up taking this approach rather than doing an automatic conversion because it is more obvious what is happening to people looking at the package.json files.


yarn.lock
```diff
@@ -7607,6 +7607,21 @@ beachball@^1.11.7:
     semver "^6.1.1"
     yargs-parser "^13.1.0"
 
+beachball@^1.29.0:
```
acoates-ms, line 7610: Why didn't this replace the one above in the lock file?

JasonVMo: Also going to remove unnecessary beachball sections from package.json files

PR #166, opened 2020-04-19
iOS/macOS with CI and RN dependencies

HeyImChris:
What this does-
- Fix iOS dependency build errors
  - Reverted custom build settings
- Rename projects to align with FluentUI-Apple
- Add macOS React-Native dependencies
- Update CI to run on macOS
- Update CI to run on iOS
  - For all our CI, we now use matrices to run builds in parallel which should speed up build times
  - Paused this with this PR-- Because React-Native requires a packager to be running to run the project, we need a custom build step (launchPackager) to do this. This must load the command into an executable and then gain executable permissions to discretely run to launch a new terminal instance, allowing the app to bundle.
  - The diagram of how a packager is run is `azure-pipelines.yml` -> `xcodebuild_wrapper` -> `launchPackager.command` -> `launchPackager.sh`
  - Open to renaming `xcodebuild_wrapper` as it's a little misleading. Kept it as is so it's aligned with FluentUI-Apple just with the additional pods and bundler complexities.
- gitignore Pods

- As mentioned in an earlier PR, iOS/macOS have to live separately from Windows due to the node_modules location dependency that Windows doesn't seem to have. They live in their own directory under "apple/FluentUITester" for now.

Validation
- Manually ran the xcodebuild scripts and they worked locally
- pods install correctly
- ios/macos build/run


.ado/ado-packager-setup_macos.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
anandrajeswaran, line 1: Can eliminate a good amount of duplication by having ado-packager-setup_ios/macos call into a common help with a "platform" argument

.ado/azure-pipelines.yml
```diff
@@ -7,69 +7,84 @@ pool:
...
+      matrix:
+        ios:
+          relative_directory: 'ios'
+          xcode_sdx: 'ios'
+          xcode_configuration: 'Debug'
+          xcode_workspacePath: 'apps/apple/FluentUITester/ios/FluentUITester.xcworkspace'
+          xcode_scheme: 'FluentUITester' #todo add debug/release schemes here
+          xcode_actions: 'build'
+          xcode_useXcpretty: true
+          xcode_destination: 'platform=iOS Simulator,OS=latest,name=iPhone 8'
```
anandrajeswaran, line 29: Why iPhone 8?

apps/apple/FluentUITester/ios/FluentUITester.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,614 @@
...
+				SWIFT_VERSION = 5.0;
+				VERSIONING_SYSTEM = "apple-generic";
+			};
+			name = Debug;
+		};
+		13B07F951A680F5B00A75B9A /* Release */ = {
+			isa = XCBuildConfiguration;
+			baseConfigurationReference = B7FF3B9B45E58F40D42487A2 /* Pods-FluentUITester.release.xcconfig */;
+			buildSettings = {
+				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
```
anandrajeswaran, line 451: I think you already mentioned that the plan is to delete this before merging?

apps/apple/FluentUITester/ios/FluentUITester.xcworkspace/contents.xcworkspacedata
```diff
@@ -2,7 +2,7 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:FluentUIReactNativeTester_iOS.xcodeproj">
```
anandrajeswaran, line 5: I thought this file (contents.xcworkspacedata) was .gitignorable?

apps/apple/FluentUITester/ios/FluentUITesterTests/FluentUITesterTests.m
```diff
@@ -0,0 +1,65 @@
...
+    [[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.1]];
+    [[NSRunLoop mainRunLoop] runMode:NSRunLoopCommonModes beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.1]];
+
+    foundElement = [self findSubviewInView:vc.view matching:^BOOL(UIView *view) {
+      if ([view.accessibilityLabel isEqualToString:TEXT_TO_LOOK_FOR]) {
+        return YES;
+      }
+      return NO;
+    }];
+  }
```
anandrajeswaran, line 54: Is this copied from elsewhere? We can simplify greatly using an XCUITest that just launches and checks for the existence of an element with TEXT_TO_LOOK_FOR. This doesn't seem well suited for an XCTest and also has a few anti-patterns we don't want to persist (redundant method invocation, timing dependent tests, dot notation, early returns, etc.)

apps/apple/FluentUITester/macos/FluentUITester-macOS/AppDelegate.m
```diff
@@ -0,0 +1,32 @@
...
+#import <React/RCTBridge.h>
+#import <React/RCTBundleURLProvider.h>
+
+@interface AppDelegate () <RCTBridgeDelegate>
+
+@end
+
+@implementation AppDelegate
+
+- (void)awakeFromNib {
```
anandrajeswaran, line 12: No need for awakeFromNib - just do this in applicationDidFinishLaunching? Important to be precise with this as we'd expect many clients to copy it

apps/apple/FluentUITester/macos/FluentUITester-macOS/AppDelegate.m
```diff
@@ -0,0 +1,32 @@
...
+  [super awakeFromNib];
+
+  _bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:nil];
+}
+
+- (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
+  // Insert code here to initialize your application
+}
+
+- (void)applicationWillTerminate:(NSNotification *)aNotification {
```
anandrajeswaran, line 22: can we just remove altogether? Or at least remove boiler-plate comment?

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,22 @@
+#import "ViewController.h"
+#import "AppDelegate.h"
+
+#import <React/RCTRootView.h>
+
+@implementation ViewController
+
+- (void)viewDidLoad {
```
anandrajeswaran, line 8: Looks more like a loadView than a viewDidLoad? clients frequently get this wrong so we shouldn't provide examples to encourage them. Most recently, deck view got this wrong to absolutely disastrous results

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,22 @@
...
+#import "AppDelegate.h"
+
+#import <React/RCTRootView.h>
+
+@implementation ViewController
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  RCTBridge *bridge = [((AppDelegate *)[NSApp delegate])bridge];
```
anandrajeswaran, line 11: Can we explicitly pass the bridge around instead of assuming time of app delegate? This is the sort of thing that makes it difficult to keep up with Apple updates

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,22 @@
...
+  RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
+
+  NSView *view = [self view];
+
+  [view addSubview:rootView];
+  [rootView setBackgroundColor:[NSColor windowBackgroundColor]];
+  [rootView setFrame:[view bounds]];
+  [rootView setAutoresizingMask:(NSViewMinXMargin | NSViewMinXMargin | NSViewMinYMargin | NSViewMaxYMargin | NSViewWidthSizable | NSViewHeightSizable)];
```
anandrajeswaran, line 19: Should just use setView instead of adding subview so we don't need this? But if we do, let's not propagate usage of manual layout since no clients should be

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,22 @@
...
+- (void)viewDidLoad {
+  [super viewDidLoad];
+
+  RCTBridge *bridge = [((AppDelegate *)[NSApp delegate])bridge];
+  RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
+
+  NSView *view = [self view];
+
+  [view addSubview:rootView];
+  [rootView setBackgroundColor:[NSColor windowBackgroundColor]];
```
anandrajeswaran, line 17: Is this actually necessary? Do we expect every client to do this?

apps/apple/FluentUITester/macos/FluentUITester-macOS/main.m
```diff
@@ -0,0 +1,5 @@
+#import <Cocoa/Cocoa.h>
```
anandrajeswaran, line 1: Minor: AppKit/Appkit.h to avoid unnecessary CoreData include

.ado/ado-packager-setup_macos.sh
HeyImChris: Yea, I was still learning all the syntactic sugar to get it running. For this PR I'm pausing on the CI packager to unblock getting the proj's in

.ado/azure-pipelines.yml
HeyImChris: It's what react native tests against and I wanted to be consistent

apps/apple/FluentUITester/ios/FluentUITester.xcodeproj/project.pbxproj
HeyImChris: Yes it's deleted now- hadn't gotten to it until the latest commit

.ado/azure-pipelines.yml
anandrajeswaran: Is there actually any advantage to consistency in this case? We're not testing react-native and probably want to intentionally choose the device we think is best to test fluentui-react-native?

apps/apple/FluentUITester/ios/FluentUITester.xcworkspace/contents.xcworkspacedata
HeyImChris: I thought it could be but it's in the .gitignore and not being parsed out

apps/apple/FluentUITester/ios/FluentUITesterTests/FluentUITesterTests.m
HeyImChris: I'll remove this completely for now until we get tests running

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
HeyImChris: We should type check it first, but I'm not sure where you're saying we should get it from if not here. I'd rather not stray too far from how RN hooks this up though, at least at first

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
HeyImChris: I don't see any reason it can't just be the root view itself

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
HeyImChris: We need this set or the view's transparent. I don't think clients mirroring our test app and setting that is a big issue?

apps/apple/FluentUITester/macos/FluentUITester-macOS/AppDelegate.h
```diff
@@ -0,0 +1,9 @@
+#import <Cocoa/Cocoa.h>
```
anandrajeswaran, line 1: minor, AppKit/AppKit.h?

apps/apple/FluentUITester/macos/FluentUITester-macOS/Config/Common.xcconfig
```diff
@@ -23,8 +24,8 @@ GCC_WARN_UNUSED_FUNCTION = YES
 GCC_WARN_UNUSED_VARIABLE = YES
 SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
 
-INFOPLIST_FILE = FluentUIReactNativeTester_macOS/Info.plist
+INFOPLIST_FILE = FluentUITester-macos/Info.plist
 
-// Versions
-MACOSX_DEPLOYMENT_TARGET = 10.13
+SDKROOT = macosx;
+MACOSX_DEPLOYMENT_TARGET = 10.14;
```
anandrajeswaran, line 30: why this change? I thought this was minOS, which is still 10.13?

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.h
```diff
@@ -0,0 +1,5 @@
+#import <Cocoa/Cocoa.h>
```
anandrajeswaran, line 1: AppKit/AppKit.h

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.h
```diff
@@ -0,0 +1,5 @@
+#import <Cocoa/Cocoa.h>
+
+@interface ViewController : NSViewController
```
anandrajeswaran, line 3: Probably want a 3-letter prefix if this is indeed objC

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,21 @@
...
+- (void)loadView {
+  id<NSApplicationDelegate> appDelegate = [NSApp delegate];
+  if ([appDelegate isKindOfClass:[AppDelegate class]]) {
+    RCTBridge *bridge = [((AppDelegate *)appDelegate) bridge];
+    RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
+
+    [self setView:rootView];
+    [rootView addConstraint:[NSLayoutConstraint constraintWithItem:rootView attribute:NSLayoutAttributeWidth relatedBy:NSLayoutRelationEqual toItem:nil attribute:NSLayoutAttributeNotAnAttribute multiplier:1.0 constant:500]];
+    [rootView addConstraint:[NSLayoutConstraint constraintWithItem:rootView attribute:NSLayoutAttributeHeight relatedBy:NSLayoutRelationEqual toItem:nil attribute:NSLayoutAttributeNotAnAttribute multiplier:1.0 constant:700]];
+    [rootView setAutoresizingMask:(NSViewMinXMargin | NSViewMinXMargin | NSViewMinYMargin | NSViewMaxYMargin | NSViewWidthSizable | NSViewHeightSizable)];
```
anandrajeswaran, line 17: should not need autoresizingMask?

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,21 @@
...
+@implementation ViewController
+
+- (void)loadView {
+  id<NSApplicationDelegate> appDelegate = [NSApp delegate];
+  if ([appDelegate isKindOfClass:[AppDelegate class]]) {
+    RCTBridge *bridge = [((AppDelegate *)appDelegate) bridge];
+    RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
+
+    [self setView:rootView];
+    [rootView addConstraint:[NSLayoutConstraint constraintWithItem:rootView attribute:NSLayoutAttributeWidth relatedBy:NSLayoutRelationEqual toItem:nil attribute:NSLayoutAttributeNotAnAttribute multiplier:1.0 constant:500]];
```
anandrajeswaran, line 15: recommended to use the anchor-based constraint APIs for more readability - [[rootView widthAnchor] constraintEqualToAnchor:500]. But should also split out 500 as a magic constant

anandrajeswaran: Actually a view shouldn't define its own width since this is the NSViewController's view

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
anandrajeswaran: This doesn't seem like it has anything to do with the RN implementation. We should pass an explicit reference to the bridge from the AppDelegate (does it need to be initialized there?) to the ViewController

apps/apple/FluentUITester/macos/FluentUITester-macOS/Config/Common.xcconfig
HeyImChris: Didn't mean to change it- 10.13 is correct

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
HeyImChris: Given the rootview has no superview, what view would the constraints be set on then?

HeyImChris: Maybe that's a reason to set this as a subview like it was originally

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
HeyImChris: By that I meant this is how React-Native recommends consumers create/manage their content via a bridge. We can stray from it, but we shouldn't without a good reason

HeyImChris: Without rewriting how RN works, we need the app delegate to conform to an RCT protocol. We still will want to cast the value to be safe, but to simplify things I moved the bridge retrieval logic into the view controller

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
```diff
@@ -0,0 +1,17 @@
+#import "FRNViewController.h"
+#import "FRNAppDelegate.h"
+
+#import <React/RCTRootView.h>
+
+@implementation FRNViewController
+
+- (void)loadView {
+  // The delegate must conform to RCTBridgeDelegate to retrieve the JS source file for the packager
+  id<RCTBridgeDelegate> delegate = (id<RCTBridgeDelegate>)[NSApp delegate];
```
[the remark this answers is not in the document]
HeyImChris: I'm with you, yea we can move it all over to the view controller

PR #167, opened 2020-04-20
Update to new version of beachball with working prepublish hook

JasonVMo:
This change has three main parts:

### Bump beachball version

This moves to a version of beachball that has an updated prepublish hook.  See https://github.com/microsoft/beachball/pull/326 for the fix and some of the discussion around this.

### Update prepublish implementation

This change uses the implementation that I added as part of the beachball end to end unit tests.  This works per platform and effectively applies anything in an onPublish object within package.json up to the root level.

### Manually added change files

I added change files for the published packages in the repo to make sure republish is triggered.

### DIFF
--- a/beachball.config.js
+++ b/beachball.config.js
@@ -1,14 +1,16 @@
+const fs = require('fs');
+const path = require('path');
+
 module.exports = {
   disallowedChangeTypes: ['major'],
   hooks: {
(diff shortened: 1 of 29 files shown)


JasonVMo:
I also tested this change locally with verdaccio to make sure the publish step is working correctly.

PR #168, opened 2020-04-20
Add variants and fluent type ramp to text with fix for type error

ksiler:
Until we bump rex-win32, fluentTypography is undefined resulting in "TypeError: Unable to get property 'families' of undefined or null reference". To fix this, I allowed fluentTypography to be an empty object so that typography will fallback to the platform defaults when resolving a partial theme. This gets rid of the error but all font weights will render as 'normal' until rex-win32 is bumped: 

image

This is due to the previous behavior in native where the only weights supported were 'normal' and '700' (aka 'bold'). Since the new fluent type ramp only has 'regular' (aka 'normal') or 'semibold', all 'semibold' fonts render as 'normal' and all 'bold' fonts also render as 'normal'.

This also renames 'fontVariant' to 'variant' to align with the Fluent UI Text control on the web.


docs/content/windows/components/text.mdx
```diff
@@ -1,3 +1,18 @@
-# Text
-
-### Not Yet Implemented
```
PPatBoyd, line 3: Can skip this, I'm deleting this file soon

packages/components/text/src/Text.types.ts
```diff
@@ -1,20 +1,21 @@
-import { ITextTokens, IForegroundColorTokens } from '@fluentui-react-native/tokens';
-import { ITextProps as INativeTextProps } from '@fluentui-react-native/adapters';
-
-export const textName = 'RNFText';
-
-/**
- * Properties for fabric native text field, these extend the default props for text
- */
-export type ITextProps<TBase = INativeTextProps> = TBase &
-  ITextTokens &
```
PPatBoyd, line 10: should the text tokens stay part of the props?

packages/components/text/src/Text.types.ts
```diff
@@ -1,20 +1,21 @@
...
+ */
+export type ITextProps<TBase = INativeTextProps> = TBase &
+  ITextVariantTokens &
+  IForegroundColorTokens & {
+    disabled?: boolean;
+  };
+
+export type ITextType<TBase = INativeTextProps> = {
+  props: ITextProps<TBase>;
+  tokens: ITextTokens & IColorTokens;
```
PPatBoyd, line 17: not sure why these two tokens are here and those part of ITextProps are different; what am I missing? :)

packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
```diff
@@ -1,86 +1,86 @@
...
+  const palette = translatePalette(module, cache, id);
+  return {
+    colors: {
+      brand: createColorRamp({ values: module.ramps.App }),
+      neutrals: createColorRamp({ values: module.ramps.FluentGrays }),
+      warning: createColorRamp({ values: module.ramps.Sepias }),
+      neutrals2: createColorRamp({ values: module.ramps.ClassicGrays }),
+      ...palette
+    },
+    typography: module.fluentTypography || {},
```
PPatBoyd, line 53: do you have a task tracking this || being removed w/ the appropriate rex bump?

ksiler: No but I suppose I should make one. Thanks for the reminder!

ksiler: Link to task here: https://office.visualstudio.com/OC/_workitems/edit/4107153

packages/components/text/src/Text.types.ts
ksiler: I asked this in an earlier conversation on Teams. We decided to only have variant as a prop to start. This way it pushes people to customize text using variants rather than setting family/weight/size directly. We can always add those props later if we decide to do so.

packages/components/text/src/Text.types.ts
ksiler: Seems to me like we're only enabling a subset of tokens to be props on Text to push people to customize text just with those props and not the full set of tokens.

ksiler:
I've noticed the same but haven't gotten to the bottom of it. I'm using Visual Studio Code. It seems to me that the ends are different but I'm not sure how to make it consistent. If you highlight the before and after text from my changes you can see an extra space at the end of the file which is removed in my changes.

PR #179, opened 2020-04-23
Remove docs from yarn and lerna workspace

amitse:
As the documentation would be moved to a different repo, we no longer need these yarn workspace and lerna package for docs


JasonVMo:
Thanks for the contribution!

PR #180, opened 2020-04-24
Add a git attributes file for line ending settings

JasonVMo:
This just adds a .gitattributes file to the root of the repo to put line endings into auto mode.  The git install options are confusing and this enforces the auto mode.  I think this is the root of some of the recent PRs not diffing files correctly.


ksiler:
Do we need to renormalize all line endings in the repo with `git add --renormalize` to fix the files that were checked in with Windows-style line endings?

JasonVMo:
I renormalized all the files with the new setting, the changes are just in the line endings.

PR #181, opened 2020-04-24
Update RN to 0.62 and make app folders platform specific

JasonVMo:
This is a big change with a couple of parts.  I'm still working through errors on the windows bundle (right now it looks like I'm picking up an old metro config) but that is at runtime, not build or bundle time.  This change does the following:

## Update most RN versions to 0.62 in the project

The apple directory and the win32 versions are still unchanged but all files that target RN directly will use 0.62.  This also bumps the react versions and a bunch of other utility libraries.  This will make the changeover when win32 is ready to be updated pretty seamless.

## Pull in experimental 0.62 version of RNW

This pulls in from master to get the 0.62 version, but more importantly it sets up using the experimental pre-built binary.  Now the run-windows command should work.  I believe this is tied to "yarn windows" in that folder.  It redboxes right now but is getting close.

## Split out app specific testers

The folder apps/fluent-tester has become a source provider but does not provide bundles.  

- Instead this is targeted by both apps/win32 and apps/windows.
- The "yarn bundle" command works in each folder to produce bundles
- The windows one is actually routing through the official CLI which has some long term maintanence advantages, particularly enabling the run-windows command

## Build tools were updated to work with projects that bundle using the CLI

This has the same basic syntax for usage, it just passes in a --cli option to opt into that mode.  When win32 is bumped to 0.62 it will switch to --cli mode and we can delete a lot of the code for dealing with haste.

## Removed expo, added apps/mobile

The expo clients had their own copies of RN binaries and required a lot of hoisting to get working.  This instead creates a mobile focused project that right now just builds the stock starter app but does build and bundle correctly without any unnecessary hoisting.

## Cleaned up unused packages, tried to dedupe our versions

Overall this should shrink the number of packages we require substantially.  Yarn times were up to 4 minutes in CI loops I saw.  Since this runs in every parallel loop and in publish it was a contributor to our publish return times being around 30 minutes.

## Remaining work

- Unblock windows bundle - this looks like cleaning up a redbox error for hmrShow or something like that, the web says it is a wrong config, we'll see
- (done) I noticed some different jest versions, I'll do a pass to make sure those are aligned as well.


apps/windows/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "lint": "eslint .",
+    "windows": "react-native run-windows",
+    "bundle": "fluentui-scripts metro --cli"
+  },
+  "dependencies": {
+    "@fluentui-react-native/tester": "^0.0.1",
+    "@uifabricshared/theming-react-native": "0.7.0",
+    "react": "~16.11.0",
+    "react-native": "0.62.2",
+    "react-native-windows": "^0.0.0-master.45"
```
NickGerleman, line 17: As stated, this will also bring in 0.63, 0.64, etc as we merge into our master.

Current plan is to release 0.62.0-preview.1 on Monday if you can wait.

JasonVMo: Because the experience of trying to build UWP was so horrible for people on the 0.60 versions if we can get folks unblocked I'd probably check this in and then just switch off master once the preview version is up.

PR #187, opened 2020-04-28
Apple should always use the system's font

HeyImChris:
Clients redbox using ios/macos themes because the Segoe UI font doesn't exist on macos/ios.

Apps on ios and mac should use the OS recommended fonts. This has the benefit of automatically updating when new OS's come out and also keeping with the platform experience our users expect.

Perform a platform check and conditionally update our dict to contain the new apple values. 

Another approach I considered was to delete the `families` entry altogether when on Apple, but that proved to be a bigger investment as it trips up our TS type checking when the dictionary can optionally have a key-value mapping.

FontFamiy is a style prop on the RN Text component so this just provides a themed manner to set that value.

### DIFF
--- a/packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
+++ b/packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
@@ -1,20 +1,22 @@
 import { ITheme } from './Theme.types';
-import { getStockWebPalette, ITypography, ISpacing } from '@uifabricshared/theming-ramp';
+import { getStockWebPalette, ITypography, ISpacing, FontWeightValue, FontSize, IFontSizes, IFontVariants } from '@uifabricshared/theming-ramp';
+import { Platform } from 'react-native';
 
 function _defaultTypography(): ITypography {
-  return {
+
+  const defaultsDict = {
     sizes: {
-      caption: 8,
-      secondary: 9,
-      body: 11,
-      subheader: 12,
-      header: 15,
-      hero: 21,
-      heroLarge: 32
-    },
+      caption: 8 as FontSize,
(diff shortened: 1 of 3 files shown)


packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
```diff
@@ -38,8 +40,23 @@ function _defaultTypography(): ITypography {
...
   };
+
+  const familiesDictApple = {
+    primary: 'System',
+    secondary: 'System',
+    cursive: 'System',
+    monospace: 'System',
+    sansSerif: 'System',
+    serif: 'System'
+  }
```
anandrajeswaran, line 53: Move inside platform block so other platforms don't pay cost?

packages/framework/theming-react-native/src/BaselinePlatformDefaults.ts
```diff
@@ -38,8 +40,22 @@ function _defaultTypography(): ITypography {
...
+
+  if (Platform.OS === 'macos' || Platform.OS === 'ios') {
+    const familiesDictApple = {
+      primary: 'System',
+      secondary: 'System',
+      cursive: 'System',
+      monospace: 'System',
+      sansSerif: 'System',
+      serif: 'System'
+    }
```
anandrajeswaran, line 54: Should this be "as IVariants"?

HeyImChris: I see why you say that because the win version is casted. First thoughts are to comply with the transpiler instead of explicitly casting everything which could end up with redundancies. I'm open to discuss it more online if you think we _should_ always cast in these scenarios.

PR #191, opened 2020-05-01
use findNodeHandle to avoid serializing refobject

PPatBoyd:


packages/components/Callout/src/Callout.tsx
```diff
@@ -12,12 +12,14 @@ const RCTCallout = requireNativeComponent('RCTCallout');
 export const Callout = compose<ICalloutType>({
   displayName: calloutName,
   usePrepareProps: (props: ICalloutProps, useStyling: IUseComposeStyling<ICalloutType>) => {
-    const { componentRef, ...rest } = props;
+    const { componentRef, target, ...rest } = props;
     const calloutRef = useViewCommandFocus(componentRef);
+    const targetNativeTag = findNodeHandle(target.current);
```
FalseLobster, line 17: I'm sort of curious if this is a race condition. target.current getting updated won't trigger a rerender of callout so if Callout renders before target gets filled in we might be able to run into problems?

PPatBoyd: I'll look into this and see what can happen here, I'm not clear on the finer details of the DOM-to-bridge interactions.  Transient UI being anchored to something that doesn't exist yet feels somewhat not-a-scenario, straightforward to work around around user-side if it's an actual issue.  I'm not quite sure what I'd do to avoid the issue control-side, but will find out more soon :)

PPatBoyd: it goes through usePrepareProps twice and ends up moving the callout immediately after show; not ideal, but not the worst given problems we can have if we start holding live elements around without hosting them.

PR #197, opened 2020-05-08
Remove backgroundColor from focused _override setting

lenahong:
remove backgroundColor from focused _override setting to let hovered setting take precedence

### DIFF
--- a/packages/components/Button/src/Button.settings.ts
+++ b/packages/components/Button/src/Button.settings.ts
@@ -8,7 +8,8 @@ export const settings: IComposeSettings<IButtonType> = [
       backgroundColor: 'buttonBackground',
       color: 'buttonText',
       borderColor: 'buttonBorder',
-      borderWidth: 1
+      borderWidth: 1,
+      borderRadius: 2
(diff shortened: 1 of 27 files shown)


packages/components/Button/src/Button.settings.ts
```diff
@@ -57,7 +58,6 @@ export const settings: IComposeSettings<IButtonType> = [
       focused: {
         tokens: {
           borderColor: 'buttonBorderFocused',
-          backgroundColor: 'buttonBackgroundHovered',
```
FalseLobster, line 60: Why does this even change any behavior?  This is set to the same value as the hovered state.

lenahong: I looked at the Figma file for WinUI Button redlines, and focusing is not supposed to have the hovered background color? Removing this setting only allows for the hover background during focus, when I hover over a already focused button.

FalseLobster: oh got it, thanks for explaining :)

PR #199, opened 2020-05-11
Move win32 RN to 0.62 and pull in updated tester

JasonVMo:
This moves the react-native version for win32 to 0.62.2 and pulls in the associated tester for that version.  Specific changes include:

- bump RN-win32 version to most recent preview version (preview.3)
- pull in matching tester version based on preview.3
- earlier previews had a websocket bug that blocked live reload scenarios
- tweaked a few jest/babel versions to ensure we were consistent in the repo
- switched the commands for bundling/start to use the community CLI now that we are on 0.62

I'll clean up some of the build code as a secondary change after this.  Note that this code somewhat makes the live reload scenarios work.  I tested:

- running the normal run-win32 sequence
- running run-win32-web + yarn start which successfully reloads when ReloadUI is chosen
- running run-win32-devmain from devmain enlistment

True live reload doesn't seem to fully work, you still need to choose the ReloadUI option to prompt a reload, but it is connecting to the bundle server correctly now.  This appears to be a limitation of the current host implementation in native but I need to dig in more.  Generally though the scenarios that used to work now work again.


apps/win32/package.json
```diff
@@ -31,23 +31,22 @@
     "react-native-svg": "^11.0.0"
   },
   "devDependencies": {
-    "@office-iss/react-native-win32": "0.27.3",
-    "@office-iss/rex-win32": "0.0.0-devmain.12823.10000",
+    "@office-iss/react-native-win32": "0.62.0-preview.3",
```
NickGerleman, line 34: FYI we're on preview 4 now

JasonVMo: Currently this is set to match the rex-win32 version that we are pulling in.  Once we get the publishing pipeline well established on the native side I'd like to figure out how to establish regular version bumping in this repo.

tkamireh:
Just to be sure, the removal of the react-native-win32 dependencies that didn't get updated doesn't break anything, right? It's a little strange that we have that dependency in a bunch of places it wasn't required, maybe people were just copying dependencies when creating new packages.

JasonVMo:
> Just to be sure, the removal of the react-native-win32 dependencies that didn't get updated doesn't break anything, right? It's a little strange that we have that dependency in a bunch of places it wasn't required, maybe people were just copying dependencies when creating new packages.

I think that's right.  I removed any package references that didn't have an accompanying import { foo } from '@office-iss/react-native-win32' statement.

PR #204, opened 2020-05-13
Sanajmi/apple test app

Saadnajmi:
## PROBLEM

We want to stand up the FluentUITester app for iOS and macOS. Currently, the existing code in `apps/apple/FluentUITester` doesn't work very, and has some dependency conflicts. While we don't have the full FluentUITester app running with this PR, we let's at least get a starter app with the React Native Launch screen running on both platforms.

## FIX

- Off to the side somewhere else, create a new generated template app following the instructions [here](https://microsoft.github.io/react-native-windows/docs/rnm-getting-started), with React Native version 0.61 (the latest we support on macOS).
- Run `npx react-native upgrade 0.61.0` at `apps/apple/FluentUITester` (This won't do everything we need, so we have some additional steps to do).
- Manually update the `package.json` to use react-native-macos 0.61 as well, so we don't have mismatched React Native and React Native macOS versions.
- The Podfiles still need to be updated. From your off-to-the-side generated project, copy the dependencies of the iOS and macOS Podfile into the ones at `apps/apple/FluentUITester/<platform>/Podfile`.
- Copy the Autoresize mask code from the template app's view controller into FRMViewController, to fix the 1px issue.

## PRODUCT IMPACT

We have a base test app using React Native 0.61 for both iOS and macOS.

## VERIFICATION

From the root, ran the following commands:
```
yarn && yarn build
cd apps/apple/FluentUITester
yarn
pod install --project-directory=ios/
pod install --project-directory=macos/
npx react-native run-ios
npx react-native run-macos
# In a separate terminal
yarn start:macos
```
This launched both test apps properly.
Also made sure that Debug and Release configurations from xcode built and ran successfully.

Apple Test Apps


anandrajeswaran:
I know we've talked about it a bunch, but is there still a need for the "apple" folder? Should this just be in apps/FluentTester/ios or apps/FluentTester/macos? Wondering if that lends itself more naturally to being able to share test code with other platforms

apps/apple/FluentUITester/.flowconfig
```diff
@@ -1,4 +1,6 @@
 [ignore]
+; We fork some components by platform
```
anandrajeswaran, line 2: Shouldn't these just not exist?

apps/apple/FluentUITester/ios/FluentUITester.xcodeproj/project.pbxproj
```diff
@@ -283,29 +252,139 @@
...
+				INFOPLIST_FILE = FluentUITester/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
+				OTHER_LDFLAGS = (
+					"$(inherited)",
+					"-ObjC",
+					"-lc++",
+				);
+				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)";
+				PRODUCT_NAME = FluentUITester;
+				VERSIONING_SYSTEM = "apple-generic";
```
anandrajeswaran, line 269: Should avoid any build settings in the pbxproj. These add considerable complexity as they are duplicated per target. This is what needs to be broken out to xcconfigs

anandrajeswaran: Can go to the build settings pane in the xcodeproj, CMD+A, delete to easily do this.

apps/apple/FluentUITester/ios/FluentUITester.xcodeproj/xcshareddata/xcschemes/FluentUITester.xcscheme
```diff
@@ -1,11 +1,25 @@
...
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "83CBBA2D1A601D0E00E9B192"
+               BuildableName = "libReact.a"
```
anandrajeswaran, line 18: I'm I understanding, but is this not a scheme just to build React.xcodeproj? Why would we need that here?

apps/apple/FluentUITester/ios/FluentUITester/AppDelegate.m
```diff
@@ -48,7 +32,11 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
 - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 {
+#if DEBUG
   return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index" fallbackResource:nil];
+#else
+  return [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
+#endif
```
anandrajeswaran, line 39: Is this common to be different between release and debug? If so, why wouldn't that be a react-native implementation detail?

apps/apple/FluentUITester/ios/FluentUITester/Info.plist
```diff
@@ -50,7 +50,6 @@
 		<string>UIInterfaceOrientationPortrait</string>
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
-		<string>UIInterfaceOrientationPortraitUpsideDown</string>
```
anandrajeswaran, line 53: Need to add this back or we lose the ability for this app to do multi-window

apps/apple/FluentUITester/ios/FluentUITester/Info.plist
```diff
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>$(PRODUCT_NAME)</string>
```
anandrajeswaran, line 8: Curious why this change?

apps/apple/FluentUITester/macos/Common.xcconfig
```diff
@@ -6,7 +6,8 @@
 PRODUCT_MODULE_NAME = FluentUITester
 SWIFT_MODULE_NAME = FluentUITester
 PRODUCT_BUNDLE_IDENTIFIER = Microsoft.FluentUITester-macOS
-PRODUCT_NAME = FluentUITester
+PRODUCT_NAME = FluentUITester-macOS
+SDKROOT = iphoneos
```
anandrajeswaran, line 10: SDKROOT for FluentUITester-macOS shouldn't be iphoneos?

apps/apple/FluentUITester/macos/Common.xcconfig
```diff
@@ -24,9 +25,10 @@ GCC_TREAT_WARNINGS_AS_ERRORS = YES
...
 SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
+LD_RUNPATH_SEARCH_PATHS = $(inherited) @executable_path/Frameworks
 
-INFOPLIST_FILE = FluentUITester-macos/Info.plist
+INFOPLIST_FILE = FluentUITester/Info.plist
 
-SDKROOT = macosx;
-MACOSX_DEPLOYMENT_TARGET = 10.13;
+// Versions
+IPHONEOS_DEPLOYMENT_TARGET = 12.0;
```
anandrajeswaran, line 33: Shouldn't be in macOS xcconfig?

apps/apple/FluentUITester/macos/Debug.xcconfig
```diff
@@ -3,10 +3,10 @@
 //  Licensed under the MIT License.
 //
 
-#include "Pods/Target Support Files/Pods-Shared-FluentUITester-macOS/Pods-Shared-FluentUITester-macOS.debug.xcconfig"
+#include "Pods/Target Support Files/Pods-FluentUITester/Pods-FluentUITester.debug.xcconfig"
```
anandrajeswaran, line 6: This may be a root cause - need the macOS xcconfig?

apps/apple/FluentUITester/macos/FluentUITester-macOS/AppDelegate.m
```diff
@@ -0,0 +1,32 @@
...
+
+@end
+
+@implementation AppDelegate
+
+- (void)awakeFromNib {
+  [super awakeFromNib];
+
+  _bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:nil];
+}
```
anandrajeswaran, line 16: Why in awakeFromNib?

apps/apple/FluentUITester/macos/FluentUITester-macOS/AppDelegate.m
```diff
@@ -0,0 +1,32 @@
+#import "AppDelegate.h"
```
anandrajeswaran, line 1: Looks like we've lost Apple's recommended 3-letter prefix

anandrajeswaran: Throughout the change

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
```diff
@@ -0,0 +1,22 @@
...
+
+  RCTBridge *bridge = [((AppDelegate *)[NSApp delegate])bridge];
+  RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
+
+  NSView *view = [self view];
+
+  [view addSubview:rootView];
+  [rootView setBackgroundColor:[NSColor windowBackgroundColor]];
+  [rootView setFrame:[view bounds]];
+  [rootView setAutoresizingMask:(NSViewMinXMargin | NSViewMinXMargin | NSViewMinYMargin | NSViewMaxYMargin | NSViewWidthSizable | NSViewHeightSizable)];
```
anandrajeswaran, line 19: This may have fixed the 1px issue, but none of this code should be necessary. It is perfectly legal for a client of react-native to use an RCTRootView in autolayout and seemingly it works for iOS

apps/apple/FluentUITester/macos/FluentUITester-macOS/main.m
```diff
@@ -1,4 +1,4 @@
-#import <AppKit/AppKit.h>
+#import <Cocoa/Cocoa.h>
```
anandrajeswaran, line 1: Should still be AppKit to avoid unnecessary includes

apps/apple/FluentUITester/.flowconfig
Saadnajmi: The entire file or the reference to android? The reference to android can probably dissappear

Saadnajmi:
Judging from the # of comments about the generated obj-c files and pbxproj, I'm going to go back and try to copy over those files from the original repo, and use them, instead of using the autogenerated ones.

apps/apple/FluentUITester/macos/FluentUITester-macOS/ViewController.m
Saadnajmi: Not sure I'll be able to solve this issue in this code review, but we could fix up the autogenerated template code at the react-native-macos level in a separate CR, and then propogate the change to here?

apps/apple/FluentUITester/package.json
```diff
@@ -10,9 +10,9 @@
     "start:macos": "node node_modules/react-native-macos/local-cli/cli.js start --use-react-native-macos"
   },
   "dependencies": {
-    "react": "16.11.0",
-    "react-native": "0.60.0",
-    "react-native-macos": "^0.60.0-0"
+    "react": "16.9.0",
+    "react-native": "0.61.5",
```
HeyImChris, line 14: Should probably keep these in sync

apps/apple/FluentUITester/package.json
```diff
@@ -10,9 +10,9 @@
     "start:macos": "node node_modules/react-native-macos/local-cli/cli.js start --use-react-native-macos"
   },
   "dependencies": {
-    "react": "16.11.0",
-    "react-native": "0.60.0",
-    "react-native-macos": "^0.60.0-0"
+    "react": "16.9.0",
```
Saadnajmi, line 13: Technically, this is moving down. Might as well bring this back up to 16.11.0 if we can?

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
```diff
@@ -11,12 +11,18 @@ @interface FRNViewController () <RCTBridgeDelegate>
...
   RCTBridge *bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:nil];
   RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
 
-  [self setView:rootView];
+  NSView *view = [self view];
+
+  [view addSubview:rootView];
+  [rootView setBackgroundColor:[NSColor windowBackgroundColor]];
+  [rootView setFrame:[view bounds]];
+  [rootView setAutoresizingMask:(NSViewMinXMargin | NSViewMinXMargin | NSViewMinYMargin | NSViewMaxYMargin | NSViewWidthSizable | NSViewHeightSizable)];
```
anandrajeswaran, line 25: Should be loadView. viewDidLoad is additional construction after loading from a xib. loadView is when you're making the hierarchy yourself.

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
```diff
@@ -11,12 +11,18 @@ @interface FRNViewController () <RCTBridgeDelegate>
...
+  [super viewDidLoad];
+  
   RCTBridge *bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:nil];
   RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge moduleName:@"FluentUITester" initialProperties:nil];
 
-  [self setView:rootView];
+  NSView *view = [self view];
+
+  [view addSubview:rootView];
+  [rootView setBackgroundColor:[NSColor windowBackgroundColor]];
```
anandrajeswaran, line 23: This definitely shouldn't be needed?

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
anandrajeswaran, line 25: Per Chris's comment, can we revert this code which shouldn't be needed and solve the 1px issue in a separate commit?

apps/apple/FluentUITester/package.json
anandrajeswaran: Maybe, consistency with what react-native generates is probably safer though

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
Saadnajmi: putting it in loadView gave me an infinite loop of trying to initialize the view. I'll see if I can fic that up

anandrajeswaran: Don't call[self view]. that's what attempts to lazily load the view. loadView should construct the hierarchy and then call [self setView:]

Saadnajmi: Separate commit or PR? doing so would mean we don't have a usable macOS test app, and that would be a blocker till we either re-apply this fix or come up with a better one.

apps/apple/FluentUITester/package.json
Saadnajmi: Wil keep as is for now then

apps/apple/FluentUITester/macos/FluentUITester-macOS/FRNViewController.m
anandrajeswaran: Ideally separate PR - this one seems to be about fixing up install of the test apps so making a separate change to its layout/view hierarchy is surprising

apps/apple/FluentUITester/macos/Podfile
```diff
@@ -1,35 +1,41 @@
...
+  pod 'Yoga', :path => '../node_modules/react-native-macos/ReactCommon/yoga'
+
+  pod 'boost-for-react-native', :podspec => '../node_modules/react-native-macos/third-party-podspecs/boost-for-react-native.podspec'
+  pod 'DoubleConversion', :podspec => '../node_modules/react-native-macos/third-party-podspecs/DoubleConversion.podspec'
+  pod 'glog', :podspec => '../node_modules/react-native-macos/third-party-podspecs/glog.podspec'
+  pod 'Folly', :podspec => '../node_modules/react-native-macos/third-party-podspecs/Folly.podspec'
 
   target 'FluentUITester-macOS' do
     platform :macos, '10.15'
+    use_native_modules!
```
anandrajeswaran, line 38: You mentioned this in the PR notes, but should we add a comment explaining this?

apps/apple/FluentUITester/macos/Podfile
```diff
@@ -1,35 +1,41 @@
-# require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 
 abstract_target 'Shared' do
-  # use_native_modules!
```
HeyImChris, line 4: Can macos still not find native_modules even with the include above succeeding?

apps/apple/FluentUITester/macos/Podfile
```diff
@@ -1,35 +1,41 @@
...
+  pod 'ReactCommon/turbomodule/core', :path => '../node_modules/react-native-macos/ReactCommon'
+  pod 'Yoga', :path => '../node_modules/react-native-macos/ReactCommon/yoga'
+
+  pod 'boost-for-react-native', :podspec => '../node_modules/react-native-macos/third-party-podspecs/boost-for-react-native.podspec'
+  pod 'DoubleConversion', :podspec => '../node_modules/react-native-macos/third-party-podspecs/DoubleConversion.podspec'
+  pod 'glog', :podspec => '../node_modules/react-native-macos/third-party-podspecs/glog.podspec'
+  pod 'Folly', :podspec => '../node_modules/react-native-macos/third-party-podspecs/Folly.podspec'
 
   target 'FluentUITester-macOS' do
     platform :macos, '10.15'
```
HeyImChris, line 37: Should be 10.14

apps/apple/FluentUITester/macos/Podfile
HeyImChris: use_native_modules helps with autolinking. It _should_ work for macos if the include at the top is finding the dep correctly

HeyImChris:
> > I know we've talked about it a bunch, but is there still a need for the "apple" folder? Should this just be in apps/FluentTester/ios or apps/FluentTester/macos? Wondering if that lends itself more naturally to being able to share test code with other platforms
> 
> @anandrajeswaran I couldn't reply to your comment but I can quote it here.
> 
> The project is structured such that each platform's test app is at `apps/<platform>/`, and they all share the same source code for the master detail app, which is located at `apps/fluent-tester`.
> There are two exceptions to this rule: `apps/apple`, which contains the macOS and iOS implementation based on react-native-macos, and `apps/mobile`, which contains the android and... iOS implementations based on react-native. There is definitely some duplication there, which I think should be resolved, and would educate whether we get rid of the `apps/apple` folder. I need to talk to @HeyImChris or @tom-un on where we want the iOS app / whether it should be based on react-native, or react-native-macos.

We should be okay to split them up. The reason they weren't is because they were having dependency issues but I'm looking into fixing those now so we won't need this workaround

apps/apple/FluentUITester/macos/Podfile
Saadnajmi: The `require relative` and `use_native_modules` are now uncommented, and seem to be working fine =)

apps/apple/FluentUITester/macos/Podfile
Saadnajmi: It's explained in the React-Native docs and is standard template code. Admittedly I didn't know what it was, but I don' think it needs a comment here

apps/apple/FluentUITester/macos/Podfile
Saadnajmi: Derp, sorry I missed that. Just updated this

PR #208, opened 2020-05-18
Move iOS/macOS into lerna monorepo ecosystem and update CI

HeyImChris:
## Problem
Lerna isn't aware of our FluentUI-React-Native Apple projects. This means we need to run extra steps just to build our software and we have tons of duplicated information in our node_modules.

## Fix
1) Move all apple project settings up one level and delete the FluentUITester folder
2) Update package.json and lerna.json at root to make them aware of the ios/macos projects
3) Stop hoisting babel to root. Not all project deps support monorepo hoisting and babel appears to be one of them.
4) Update our CI to remove unnecessary commands which simplifies the developer story for anyone working in the repo

## Followup
The README needs to get updated and now that we're part of the lerna monorepo, it's a great time to do that. After this gets in and is running correctly, I'll follow up with README updates where applicable to get the build steps written down/in a better state.

## Validation
iOS builds/runs
macOS state hasn't changed (currently there's a 1px width bug that I'll fix in a different PR)


apps/apple/ios/Podfile
```diff
@@ -1,5 +1,5 @@
 platform :ios, '12.0'
-require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+require_relative '../../../node_modules/@react-native-community/cli-platform-ios/native_modules'
```
HeyImChris, line 2: There were two possible solutions here, but this one avoids duplicating node_modules if we can just redirect it to root's version.

.ado/templates/apple-xcode-build.yml
```diff
@@ -10,19 +10,10 @@ parameters:
...
-      yarn
-    env:
-      platform_directory: ${{ parameters.relative_directory }}
-    displayName: "yarn"
-
   # Install any pods necessary for the project
   - bash: |
-      echo "Running: cd apps/apple/FluentUITester/$platform_directory && pod install" &&
-      cd apps/apple/FluentUITester/$platform_directory &&
+      echo "Running: cd apps/apple/$platform_directory && pod install" &&
```
Saadnajmi, line 15: Side note: we could also run `pod install --project-directory=$platform_directory/` if we want to save one command

apps/apple/package.json
```diff
@@ -26,5 +26,8 @@
   },
   "jest": {
     "preset": "react-native"
+  },
+  "workspaces": {
+    "nohoist": ["react-native", "react-native/**", "@babel/**"]
```
Saadnajmi, line 31: Not part of the review, but is this something we should be doing for all the test app platforms?

HeyImChris: Win32/Windows may not need to- I haven't checked but they have a completely separate fork of react-native which may have different requirements. iOS/macOS need it though

.ado/templates/apple-xcode-build.yml
anandrajeswaran: That makes a follow-up change to remove the apple directory easier too?

apps/apple/package.json
anandrajeswaran: Can we understand this before going this step? I'm pretty sure other platforms have a dependency on babel. If it's a problem in our fork of react-native we should be sure upgrading will fix it or otherwise just fix it. but adding this allowance for it here sounds non-preferred from what I can read of yarn

anandrajeswaran: And this feels like the sort of thing where if we say we'll do it as a follow-up it may just never happen. So probably at the very least need a comment explaining this and how a future contributor should treat this line (add no more, look to remove, why it wasn't removed to begin with)

.ado/templates/apple-xcode-build.yml
HeyImChris: Anand are you saying you want to split ios and macos into separate folders with separate package.json's?

anandrajeswaran: I thought that's what you and Saad recommended in side Teams chat?

apps/apple/package.json
HeyImChris: Windows does too but it their version of RN is very different from ios/android/macos. RN expects babel to be at a specific location but yarn automatically pulls it up to root. That's why we need to nohoist it to get it back to where react native is expecting that dependency to be. This is a common problem using yarn in monorepos and nohoist is the tool designed exactly for it

anandrajeswaran: If windows has this fixed in their version of react-native, does that mean we can too?

.ado/templates/apple-xcode-build.yml
HeyImChris: Okay cool yea I think it'll be better organized that way and give us parity with other platforms

apps/apple/package.json
HeyImChris: I can try being more specific. I think we only need the runtime component from within babel

HeyImChris: Comments aren't allowed in json

HeyImChris: Maybe... I don't think it's a big enough issue to merit spending much time on right now when nohoist is designed for this exact problem and recommended to use in react-native monorepos

.ado/templates/apple-xcode-build.yml
Saadnajmi: Yeah, that's the approach I thought we landed on too. I figured we would do that change after we had Lerna working, but doing them both at the same time is also great.
Side note: We can have a separate PR to delete `apps/mobile/ios` after this goes in?

.ado/templates/apple-xcode-build.yml
```diff
@@ -12,9 +12,8 @@ parameters:
 steps:
   # Install any pods necessary for the project
   - bash: |
-      echo "Running: cd apps/apple/$platform_directory && pod install" &&
-      cd apps/apple/$platform_directory &&
-      pod install
+      echo "Running: pod install --project-directory=$platform_directory/" &&
+      pod install --project-directory=$platform_directory/
```
Saadnajmi, line 16: Looking at the CI build failure, it looks like you need the full path and this should be 
`pod install --project-directory=apps/apple/FluentUITester/$platform_directory`

apps/apple/package.json
anandrajeswaran: But shouldn't autolinking in react-native have removed this need? The worry is that there's still something deeply wrong with our react-native fork and we'll have to keep working around it

HeyImChris: It's likely not a deep problem with our fork of react-native. We don't do anything unusual with babel so it's more likely an artifact of trying to consume it in a monorepo with tons of other packages

.ado/templates/apple-xcode-build.yml
HeyImChris: Will follow up with that fix since it's broken CI twice and can be easily worked on in a followup PR

PR #210, opened 2020-05-19
Add a default minimum window size to the macOS test app

Saadnajmi:
## PROBLEM

The FluentUI React Native test app for macOS has an issue where the window only has a width of 1 pixel.. basically making it unusable. This is due to the fact that RCTRootView defaults it's intrinsicContentSize to CGSizeZero if no more information is added.

## FIX

Simply add a mininum size to the window in the storyboard file. 
It's not clear to me that there is an actual preferred minimum size here, given how flexibly yoga can layout the React-Native view. Xcode suggested a minimum of 480x270, so I wont with that.

Side note:
Looking at Facebook's [React-Native repo](https://github.com/facebook/react-native/commit/432868b0c0c591c245a9cdb91201c64aa5bd70c6), I see a property "minimumSize" that is added. When that makes it to React Native macOS, we can switch to using that implementation instead.

## PRODUCT IMPACT

The macOS Fluent-Tester app is now visible

## VERIFICATION

In the `apps/apple/FluentUITester` directory, ran the following commands
```
yarn 
pod install --project-directory=macos/
npx react-native run-macos
yarn start:macos #Shouldn't need to do this last one, but run-macos doesn't seem to run the packaging server for me so here we are
```

The test app displays visibly.


apps/apple/FluentUITester/macos/FluentUITester-macOS/Base.lproj/Main.storyboard
```diff
@@ -690,6 +690,7 @@
                         <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
                         <rect key="contentRect" x="196" y="240" width="480" height="270"/>
                         <rect key="screenRect" x="0.0" y="0.0" width="1680" height="1027"/>
+                        <value key="minSize" type="size" width="480" height="270"/>
```
anandrajeswaran, line 693: Should we just add this from the AppDelegate instead to keep configuration wherever possible in code for easier maintenance? Or since some window stuff is already here we make all static window stuff here?

Saadnajmi: Vote to keep it here as FRNAppDelegate is quite empty at the moment, and the rest of the static window code here as well.

PR #216, opened 2020-05-20
remove node warning

dotnetjunky:
Remove the check for node v10 because we can use v12 now.

### DIFF
--- a/scripts/use-yarn-please.js
+++ b/scripts/use-yarn-please.js
@@ -41,23 +41,6 @@ if (path.basename(npmPath) !== 'yarn.js') {
   process.exit(1);
 }
 
-checkNodeVersion();
-
-function checkNodeVersion() {
-  const nodeVersion = process.versions.node;
-  if (!nodeVersion) {
-    return;
-  }
-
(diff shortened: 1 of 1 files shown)


scripts/use-yarn-please.js
```diff
@@ -41,23 +41,6 @@ if (path.basename(npmPath) !== 'yarn.js') {
...
-checkNodeVersion();
-
-function checkNodeVersion() {
-  const nodeVersion = process.versions.node;
-  if (!nodeVersion) {
-    return;
-  }
-
-  const nodeMajorVersion = nodeVersion.split('.', 1);
-  if (nodeMajorVersion.length > 0 && parseInt(nodeMajorVersion[0]) >= 11) {
```
tkamireh, line 53: I'd just update the version to 13, for anyone using the most current version. Thoughts?

samuelfreiberg: I agree with Tariq, checking against 13 seems like the best thing to do

dotnetjunky: Is there any known issue with version > 13?

Issue #194, opened 2020-05-07
Incorrect instructions for Fluent-Tester


NikoAri:
https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester

Current instructions:
Then go into apps\fluent-tester folder:
    cd apps\fluent-tester
Build the FluentUI Tester bundle:
    yarn bundle
Launch the FluentUI Tester app:
    yarn run-win32

But they should be cd apps\fluent-tester\win32

dotnetjunky:
Feel free to submit a PR :) This is open source.

NikoAri:
Thanks, will submit PR with doc fixes...

ejlayne:
I've submitted a PR that addresses this. https://github.com/microsoft/fluentui-react-native/pull/473 Closing, but please let us know if you have any other issues or feedback.

PR #217, opened 2020-05-20
Update Fluent Tester readme

ejlayne:
Edit Fluent Tester instructions for running on Win32, per Nikolai's issue. https://github.com/microsoft/fluentui-react-native/issues/194

### DIFF
--- a/apps/fluent-tester/README.md
+++ b/apps/fluent-tester/README.md
@@ -4,10 +4,10 @@
 
 1. Make sure you have followed the Getting Started instructions [here](../../README.md) to install packages and build the entire repository.
 
-2. Then go into `apps\fluent-tester` folder:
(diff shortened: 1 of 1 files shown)


FalseLobster:
You know, I just realized these instructions are inaccurate for Windows.  Maybe this should be in the win32 specific subdir for now?

ejlayne:
It would make sense to add instructions for each platform if they are different, and then just have this readme link off to the platform specific instructions. @FalseLobster any objections?

FalseLobster:
> It would make sense to add instructions for each platform if they are different, and then just have this readme link off to the platform specific instructions. @FalseLobster any objections?

No objections here.  This readme.md is already replicated in the win32 directory, so maybe make this change over there and link from here to there?

ejlayne:
Closing this PR - will open separate PR for the right set of changes.

PR #218, opened 2020-05-20
Use forwardRef to handle componentRef prop 

lenahong:
Existing ComponentRef gives us an object with a focus method and not an actual reference to the component.

useViewCommandFocus calls helper function setAndForwardRef which will handle the passed in ref as a forwardedRef.

Added a findNodeHandle call inside ButtonFocusTest to test that ref.current gives us an actual reference to the component.


apps/fluent-tester/src/RNTester/TestComponents/Button/ButtonFocusTest.tsx
```diff
@@ -13,7 +14,10 @@ export const ButtonFocusTest: React.FunctionComponent<{}> = () => {
   const onFocus = React.useCallback(() => {
     setState({ focused: !state.focused });
     if (buttonRef.current && !state.focused) {
+      const node = findNodeHandle(buttonRef.current);
```
PPatBoyd, line 17: Would be nice if we can make this an actual test w/ Sammy's E2E changes.  I'd remove the console.log but it's nbd

lenahong: Good Idea! I'll create a work item for it for separate check in.

PR #222, opened 2020-05-20
Update fontSize prop to use CSS pixels instead of typographic points

ksiler:
As currently defined, the fontSize prop is specified in typographic points. This change defines the fontSize prop in DIPs/CSS px to align with react-native on every other platform and values coming in from design. Once this change (and associated native changes) are in, text that is setting the fontSize prop directly will appear 75% of their original size. The simple fix is to specify fontSize in pixels instead of points.

In addition, some controls were updated to use the text control from fluentui-react-native so that they could use the full set of text tokens and the temporary fix from https://github.com/microsoft/fluentui-react-native/pull/168 was removed.

Website Documentation changes: https://github.com/microsoft/fluentui/pull/13276


apps/fluent-tester/src/RNTester/TestComponents/Text/CustomizeUsage.tsx
```diff
@@ -23,7 +22,7 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
   const Wingdings = Text.customize({ tokens: { variant: 'captionStandard', fontFamily: 'Wingdings' } });
 
   return (
-    <View style={styles.root}>
```
tkamireh, line 26: Any particular reason we're getting rid of the style?

change/@fluentui-react-native-text-2020-05-20-11-46-47-textupdates.json
```diff
@@ -0,0 +1,8 @@
+{
```
tkamireh, line 1: Any idea why there are two change files for text?

apps/fluent-tester/src/RNTester/TestComponents/Text/CustomizeUsage.tsx
ksiler: The root from common styles defines the flexDirection as "row" when the previous flow was "column". Removing the styles returns the layout to "column".

change/@fluentui-react-native-text-2020-05-20-11-46-47-textupdates.json
ksiler: No idea :/

apps/fluent-tester/src/RNTester/TestComponents/Text/CustomizeUsage.tsx
tkamireh: Do the margin values not make a visual difference?

change/@fluentui-react-native-text-2020-05-20-11-46-47-textupdates.json
tkamireh: I don't think it should matter, hopefully beachball is smart enough to resolve them together

apps/fluent-tester/src/RNTester/TestComponents/Text/CustomizeUsage.tsx
ksiler: It does a bit but it doesn't effect the example too much. I can add it back in if you want. I just made this change to make the Text examples more consistent with the other examples which use the styles from common/styles.

tkamireh: oh no it's fine, consistency is great.

PR #224, opened 2020-05-21
Callout improve findNodeHandle usage

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Description of changes

findNodeHandle is called regardless of the validity of the target refobject, and for valid refobjects before they've been resolved by the DOM.  Moving inside useLayoutEffect supports more potential scenarios with ref-resolution concerns.

Also enables the default args jest snapshot test.

### Verification

Manually tested with invalid ref objects that cannot be resolved by findNodeHandle.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/src/Callout.tsx
```diff
@@ -14,14 +15,23 @@ export const Callout = compose<ICalloutType>({
...
     const slotProps = mergeSettings<ICalloutSlotProps>(useStyling(props), {
       root: {
+        ...{
         ref: calloutRef,
-        target: targetNativeTag,
         ...rest
-      }
+      },
+      ...((targetNativeTag !== null) ? {target: targetNativeTag} : {})
+    }
```
FalseLobster, line 34: Can we go back to the way you had these spreads before?

`root { ref: calloutRef, target: targetNativeTag, ...rest }`

It looks a little more natural to me and uses less spreads and objects.  I think you'd only need `const [targetNativeTag, setTargetNativeTag] = React.useState<number>(**undefined**);`

FalseLobster: Might need to make it `useState<number | undefined>`

packages/components/Callout/src/Callout.tsx
```diff
@@ -14,14 +15,23 @@ export const Callout = compose<ICalloutType>({
...
     const calloutRef = useViewCommandFocus(componentRef);
-    const targetNativeTag = findNodeHandle(target.current);
+    const [targetNativeTag, setTargetNativeTag] = React.useState<number>(null);
+
+    React.useLayoutEffect(() => {
+      if ((target !== undefined && target?.current !== null))
+      {
+        setTargetNativeTag(findNodeHandle(target.current));
+      }
+    }, [target]);
```
FalseLobster, line 25: Personally, I'm a fan of the more succinct JS syntax:

`if (target && target.current)`

This makes sure that `target` and `target.current` are both not falsy, which excludes both null and undefined.

Worth noting that `undefined !== null` is a true statement, so you will call findNodeHandle with an `undefined` value in `target.current` and `target` isn't guaranteed to be non-null after the first undefined check before the `&&`. I'm sure that function checks for undefined, but I'd probably avoid the entire if statement to begin with.  A `null` target might crash as is.

Another common pattern is just `target && target.current && setTargetNativeTag(findNodeHandle(target.current));`

Part of me wonders if this would be beneficial in a shared hook... but I suppose we could always move it when we need it :)

PR #226, opened 2020-05-21
Move projects up a dir and opt apple platforms into lerna/yarn

HeyImChris:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Lerna isn't aware of our FluentUI-React-Native Apple projects. This means we need to run extra steps just to build our software and we have tons of duplicated information in our node_modules.

1) Move all apple project settings up one level and delete the FluentUITester folder
2) Update package.json and lerna.json at root to make them aware of the ios/macos projects
3) Stop hoisting babel to root. Not all project deps support monorepo hoisting and babel appears to be one of them.
4) Update our CI to remove unnecessary commands which simplifies the developer story for anyone working in the repo

The README needs to get updated and now that we're part of the lerna monorepo, it's a great time to do that. After this gets in and is running correctly, I'll follow up with README updates where applicable to get the build steps written down/in a better state.

### Verification

Should be no product changes here, so validation is that iOS/macOS still build/run. CI loops passing means we have the correct generated files and E2E windows tests are unbroken.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/apple/package.json
```diff
@@ -21,10 +21,12 @@
...
     "jest": "^24.9.0",
-    "metro-react-native-babel-preset": "^0.58.0",
-    "react-test-renderer": "~16.11.0"
+    "metro-react-native-babel-preset": "^0.58.0"
   },
   "jest": {
     "preset": "react-native"
+  },
+  "workspaces": {
+    "nohoist": ["react-native", "react-native/**", "@babel/runtime/**"]
```
JasonVMo, line 30: You should be able to work around the hoisting of these, but you can do this as a separate step.  We are bundling win32 and windows with react native and react-native-windows hoisted to root.

HeyImChris: Thanks, was hitting some errors without doing so but I'm sure there's a good workaround. Will followup with you on how win does it

apps/apple/package.json
```diff
@@ -26,7 +28,13 @@
   "jest": {
     "preset": "react-native"
   },
-  "workspaces": {
-    "nohoist": ["react-native", "react-native/**", "@babel/runtime/**"]
+  "metroBundles": {
+    "targets": [
+      "win32"
```
HeyImChris, line 33: ios or macos

apps/apple/package.json
```diff
@@ -7,10 +7,13 @@
...
     "test": "jest",
     "lint": "eslint .",
-    "bundle": "fluentui-scripts metro -platform macos --dev --cli",
+    "bundle": "fluentui-scripts metro -platform macos --dev",
     "start:macos": "node ../../node_modules/react-native-macos/local-cli/cli.js start --use-react-native-macos"
   },
   "dependencies": {
+    "@fluentui-react-native/tester": "^0.0.1",
+    "@uifabricshared/theming-react-native": "0.7.28",
+    "@uifabricshared/themed-stylesheet": "0.3.29",
```
anandrajeswaran, line 16: if parity brings in dependencies we may not need, is parity a good thing?

anandrajeswaran: Would we rather hold off on parity and handle this in a separate change?

PR #234, opened 2020-05-27
Support enter key for calling onClick on Button

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32
- [ ] windows
- [ ] android

### Description of changes

onKeyDown recognizes the enter key press to call the onClick event for Button.

### Verification

Manually tested with Buttons recognizing an enter key press when focus and calling onClick event.

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-adapters-2020-05-26-21-34-00-lehon-keyboardClick.json
+++ b/change/@fluentui-react-native-adapters-2020-05-26-21-34-00-lehon-keyboardClick.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Support enter key for calling onClick on Button",
+  "packageName": "@fluentui-react-native/adapters",
+  "email": "lehon@microsoft.com",
+  "dependentChangeType": "none",
+  "date": "2020-05-27T04:33:58.656Z"
+}
--- a/packages/components/Button/src/Button.tsx
+++ b/packages/components/Button/src/Button.tsx
@@ -27,7 +27,7 @@ export const Button = compose<IButtonType>({
     const pressable = useAsPressable({ ...rest, onPress: onClick });
     const onKeyDown = React.useCallback(
       e => {
-        if (onClick && e.nativeEvent.key === 'Enter') {
+        if (onClick && (e.nativeEvent.key === 'Enter' || e.nativeEvent.key === ' ')) {
           onClick();
           e.stopPropagation()
         }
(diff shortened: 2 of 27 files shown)


packages/components/Button/src/Button.tsx
```diff
@@ -25,6 +25,14 @@ export const Button = compose<IButtonType>({
     } = userProps;
     // attach the pressable state handlers
     const pressable = useAsPressable({ ...rest, onPress: onClick });
+    const onKeyDown = React.useCallback(
+      e => {
+        if (onClick && e.nativeEvent.key === 'Enter') {
+          onClick();
```
dotnetjunky, line 31: If you have handled the event, you need to call `e.stopPropagation()` to stop it from bubbling.

change/@fluentui-react-native-adapters-2020-05-26-21-34-00-lehon-keyboardClick.json
```diff
@@ -0,0 +1,8 @@
+{
```
dotnetjunky, line 1: Why do we need to add change file to unrelated packages like this one? @JasonVMo

dotnetjunky:
You change looks like it impact more than win32. Are you sure it only applies to win32?

lenahong:
> You change looks like it impact more than win32. Are you sure it only applies to win32?



> You change looks like it impact more than win32. Are you sure it only applies to win32?

You're right, but I wasn't sure which platforms support keyboard press events since I'm not able to test on any other than win32 and uwp. On UWP, we don't support keyboard focus yet

packages/components/Button/src/Button.tsx
```diff
@@ -25,6 +25,15 @@ export const Button = compose<IButtonType>({
     } = userProps;
     // attach the pressable state handlers
     const pressable = useAsPressable({ ...rest, onPress: onClick });
+    const onKeyDown = React.useCallback(
+      e => {
+        if (onClick && e.nativeEvent.key === 'Enter') {
```
NikoAri, line 30: >Enter [](start = 46, length = 5)

What about space key???

lenahong: Added!

FalseLobster: Actually, usually buttons only support space, not enter.  We should check to make sure this is the expected design.

FalseLobster: Or maybe that's only true for other widgets like Checkboxes?  Disregard :)

lenahong: I checked buttons on Office apps, and they supported enter!

PR #236, opened 2020-05-29
Enable yarn build and yarn bundle for iOS

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Yarn bundle has not been enabled for iOS yet. Let's copy over the bundle command from the win32 test app package.json (along with a few others) to enable this. Specifically:
- `build`
- `clean`
- `lint` (We needed to update the eslint config file to match the win32 test app)
- `just`
- `codestyle`
- `bundle`
- `bundle-dev`

Let's also remove the following commands, as they fail in our current test app:
- `ios` (Renaming the ios/ subdirectory to src/ means react-native run-ios no longer works)
- `test` (There is not a jest.config file in the ios test app, so I don't think we run any actual tests)

While here, let's update some of the Apple PR yml file to remove paths that no longer exist, and simplify some commands.

Also, for some extra parity with the win32/windows test app, let's rename our test app from "Fluent**UI**Tester-ios" to "FluentTester-ios"

### Verification

`yarn && yarn build` at root runs as expected

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3510,6 +3510,20 @@
   resolved "https://registry.yarnpkg.com/@uifabric/prettier-rules/-/prettier-rules-7.0.4.tgz#03082a6cc4b586104738a24892c56e0f1d3df1b6"
   integrity sha512-PK60wLAjPrG2l93r1l/VY2BmnC1la0oxIXYBuln5DOAFJcSXMea9RxoNjuqhhqZGZo6aLjO7lW070MeVOvkABQ==
 
+"@uifabricshared/themed-stylesheet@0.3.29":
+  version "0.3.29"
+  resolved "https://registry.yarnpkg.com/@uifabricshared/themed-stylesheet/-/themed-stylesheet-0.3.29.tgz#19d41328e714589d77fe2ce05c143fc27fc15cf3"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -3510,20 +3510,6 @@
...
-
-"@uifabricshared/theming-react-native@0.7.28":
-  version "0.7.28"
-  resolved "https://registry.yarnpkg.com/@uifabricshared/theming-react-native/-/theming-react-native-0.7.28.tgz#3dee7ce9f852dbbdb36ee10653e612bbaabaf03b"
-  integrity sha512-25wGC1UQOYp2VqyPD5I/V3H9Mzlx3/3gAEQ4J1Uw1xoXxp0YaNaxVjBU9UygmOM3LooyBdmcwrPuIDOkhC+GrA==
-  dependencies:
-    "@uifabricshared/foundation-settings" "^0.5.28"
-    "@uifabricshared/theme-registry" "^0.3.28"
-    "@uifabricshared/theming-ramp" "^0.9.27"
-
```
Saadnajmi, line 3526: Need to undo these changes

PR #240, opened 2020-06-03
Add setInitialFocus prop

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Description of changes

Add setInitialFocus property to the Callout component.  By default, the Callout does not take focus.  with setInitialFocus=true, the Callout attempts to set focus on the first focusable content inside the Callout.

### Verification

Manual verification

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/src/Callout.types.ts
+++ b/packages/components/Callout/src/Callout.types.ts
@@ -98,6 +98,13 @@ export interface ICalloutProps extends ICalloutTokens {
   onShow?: () => void;
 
   /**
+   * If true then the callout will attempt to focus the first focusable element that it contains.
+   * If it doesn't find an element, no focus will be set and the method will return false.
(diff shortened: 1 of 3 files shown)


packages/components/Callout/src/Callout.types.ts
```diff
@@ -97,6 +97,13 @@ export interface ICalloutProps extends ICalloutTokens {
    */
   onShow?: () => void;
 
+  /**
+   * If true then the callout will attempt to focus the first focusable element that it contains.
+   * If it doesn't find an element, no focus will be set and the method will return false.
```
dotnetjunky, line 102: "...no focus will be set and the method will return false."

There's no method here. :)

dotnetjunky:
There's no change files.

PR #241, opened 2020-06-04
support keyboardFocus for Link and add focus visuals

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32
- [X] windows
- [ ] android

### Description of changes

Link.types
- Root slot type is changed to ViewWin32 which includes keyboard and accessibility props.
 
Link
- Added onKeyDown prop that will recognize Enter and Space keyboard events to trigger Link's onPress callback event.

Link.settings
- Set dotted border on focused Links
- Set align and display setting for correct component sizing 

### Verification

Link now accepts a keyboard tab focus and was tested by using keyboard to navigate focus on the links on RNTester's LinkTest. Links are now clickable using Enter and Space keyboard events.

linkFocus

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


packages/components/Link/src/Link.tsx
```diff
@@ -32,6 +30,15 @@ export function useAsLink(userProps: IWithLinkOptions<IViewProps>): ILinkHooks {
   );
 
   const pressable = useAsPressable({ onPress: linkOnPress, ...rest });
+  const onKeyDown = React.useCallback(
```
samuelfreiberg, line 33: Are we sure it should be onKeyDown instead of onKeyUp? I feel more often, the "action" happens on key up. Maybe I'm wrong for link though

packages/components/Link/src/Link.settings.ts
```diff
@@ -41,6 +47,13 @@ export const settings: IComposeSettings<ILinkType> = [
         tokens: {
           color: 'link'
         }
+      },
+      focused: {
+        content: {
+          style: {
+            borderColor: 'rgba(128, 128, 128, 1)',
```
NikoAri, line 54: >128, 128, 128 [](start = 31, length = 13)

Will this work across different Office themes?

change/@fluentui-react-native-radio-group-2020-06-04-14-39-18-link-focus.json
```diff
@@ -0,0 +1,8 @@
+{
```
tkamireh, line 1: Any idea what happened with all these change jsons? I'm guessing you only need the @fluentui-react-native-link one.

packages/components/Link/src/Link.tsx
lenahong: Thanks for pointing that out! I think I implemented the same for Button. I'll make the fixes

change/@fluentui-react-native-radio-group-2020-06-04-14-39-18-link-focus.json
lenahong: when I run "yarn change" it asks me if there are changes in all these packages. Even when I choose "none" for change type, these json files get created. Any way I could avoid that?

packages/components/Link/src/Link.settings.ts
lenahong: @NikoAri I couldn't find the Office Theme color name for link focus. Do you know what it is?

NikoAri: Please try contacting Alicia or Jason for theming related info

---
In reply to: [435718872](https://github.com/microsoft/fluentui-react-native/pull/241#discussion_r435718872) [](ancestors = 435718872)

PR #242, opened 2020-06-05
Add run-ios support to FluentTester-ios

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Two problems are being addresssed here:
1. The iOS app has a non-standard `src/` directory instead of an `ios/` for its iOS specific files. 
1. For whatever reason, `react-native run-ios` is not launching the packager in our monorepo

To fix the first issue, we edit the `react-native.config.js` to specify the location of `FluentTester.xcworkspace`. We also need to modify `yarn ios` to add the argument `--project path src`

To fix the second issue, we modify the `yarn ios` command to run `run-ios` in the background, and the packaging server in the foreground. This is so if a developer is running this command locally, he doesn't have any lingering background processes after running this command, since `run-ios` should terminate on it's own.

### Verification

Ran `yarn ios` at `apps/ios`. Observed that the simulator launched, and the test app built and ran succesfully. 
Typed `Ctrl-c` in the terminal to quit the packaging server, and forced quit the simulator
Ran `ps` in the terminal to make sure no processes lingered.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "FluentTester-ios",
   "version": "0.0.1",
   "private": true,
-  "main": "src/index.tsx",
+  "main": "index.tsx",
   "module": "src/index.tsx",
```
HeyImChris, line 6: can presumably change this too since index.tsx just references src/index.tsx

apps/ios/package.json
```diff
@@ -13,7 +13,8 @@
     "lint": "fluentui-scripts eslint",
     "start": "react-native start",
     "bundle": "fluentui-scripts metro --cli",
-    "bundle-dev": "fluentui-scripts metro --dev --cli"
+    "bundle-dev": "fluentui-scripts metro --dev --cli",
+    "ios": "react-native run-ios --project-path src & react-native start"
```
HeyImChris, line 17: This looks like it runs it in the same terminal instance. For it to fix CI, it'll need to kick it off in a different window. The way CI runs is it has one terminal window open and runs each command sequentially so launching the packager would then not let any other commands execute

apps/ios/react-native.config.js
```diff
@@ -15,19 +15,10 @@
  */
 'use strict';
 
-const macSwitch = '--use-react-native-macos';
```
HeyImChris, line 18: Are you sure this is safe to delete?

apps/ios/package.json
Saadnajmi: This may be an unintentional change, I'll make sure it's necessary or not.

apps/ios/package.json
Saadnajmi: Noted. I could also have the packager run in the background instead of run-ios, but I think that is a less preferable approach. I'll look for a way to launch a new window

apps/ios/react-native.config.js
Saadnajmi: We aren't using any of the forks for react-native in the iOS app, so yes?

apps/ios/package.json
HeyImChris: I actually think it's the right idea but we probably need to make it index.js instead of .tsx.

apps/ios/react-native.config.js
HeyImChris: True, it's really the macOS app we may need it

apps/ios/package.json
HeyImChris: Opening up a new terminal window was the complication we fixed in react-native-macos I explained over teams. That's why it got a bit convoluted to run the packager

apps/ios/react-native.config.js
anandrajeswaran: Shouldn't we be very much be using the fork for the iOS app so fluentui-react-native is using the same version as other Microsoft experiences using react-native? Or am I missing something?

Saadnajmi: When we hooked up the iOS app to the lerna system and moved it one level up out of `apps/apple/`, we switched from using the macOS fork to the facebook master version. Given that iOS is not an out of tree platform at the moment, this feels the more correct apporach to me. Technically, `react-native-mac`, `react-native-win32`, and `react-native-windows` all can build iOS apps, and are all "Microsoft Experiences". I didn't make the decision to use `react-native` over `react-native-macos`, but it feels like the more correct approach to me.

HeyImChris: An advantage of using react-native-macos is that we can more quickly iterate if we need a one-off bug fix. I picture the ideal world as us depending directly on react-native (master) and we have the ability to make changes in that repo quickly

apps/ios/launchPackager.command
```diff
@@ -0,0 +1,19 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Set terminal title
+echo -en "\\033]0;Metro\\a"
+clear
+
+THIS_DIR=$(cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOURCE[0]}")")" && pwd)
+
+# shellcheck source=/dev/null
```
HeyImChris, line 13: Can probably delete this comment?

HeyImChris: Actually is shellcheck clarifying potential errors like how we comment above lines to disable linting errors? Would be great to understand what that's for

Saadnajmi: This is copied from the RN version... so I have no idea what this line is supposed to mean :D
I can trim it down to the necessary bits

apps/ios/src/FluentUITester.xcodeproj/project.pbxproj
```diff
@@ -250,7 +250,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
```
Saadnajmi, line 253: Will update this script to do the previous check to see if a packager is already running

apps/ios/launchPackager.command
Saadnajmi: Decided to keep this in just so we don't have too many differences between our custom file and the react-native one. Also for education purposes if we ever decide to make a proper upstream fix where we don't need a custom commmand file anymore

PR #245, opened 2020-06-08
Adding UWP E2E Testing to CI Pipeline

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32
- [X] windows
- [ ] android

### Description of changes

- Created a template for UWP E2E Testing and added it to the pipeline
- Created a "setup-repo" template that runs all necessary commands to build+bundle repo. Enables code-reuse across all jobs.
- Upgraded Node from v10 to v12 in Windows+Apple pipeline jobs
- Upgraded UWP app configs from React Native v0.0.0-master.45 to v0.0.0-master.47

### Verification

Tested CI Loop, forgot to take screenshots, but manually verified. 

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/windows/windows/NuGet.Config
```diff
@@ -6,5 +6,6 @@
   <packageSources>
     <add key="Nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="react-native" value="https://pkgs.dev.azure.com/ms/_packaging/react-native/nuget/v3/index.json" />
+    <add key="react-native-public" value="https://pkgs.dev.azure.com/ms/react-native/_packaging/react-native-public/nuget/v3/index.json" />
```
tkamireh, line 9: Just curious, what does this feed have that the feed above it doesn't?

.ado/azure-pipelines.yml
```diff
@@ -87,24 +85,14 @@ jobs:
 
       - task: NodeTool@0
         inputs:
-          versionSpec: "10.x"
+          versionSpec: "12.x"
```
tkamireh, line 88: Should the NodeTool task be added to the setup template? If you're going to do it every time, seems like a good idea. Alternatively, I think there's a way to request agents that already have some set of software installed, might be worth looking into.

.ado/templates/e2e-testing-uwp.yml
```diff
@@ -0,0 +1,42 @@
+steps:
+  - task: DownloadSecureFile@1
+    name: UwpCertificate
+    inputs:
```
tkamireh, line 4: Interesting, you don't need to specify some path to search against?

apps/windows/windows/NuGet.Config
samuelfreiberg: I'm not sure exactly. I was having trouble connecting to our private NuGet feed on ADO pipeline. I think this feed publishes all versions of react-native and doesn't need authentication. 

From Danny Van Velzen: I'll quickly create a new feed react-native-public and publish all versions of react-native in there. That one should not need any authentication.

.ado/azure-pipelines.yml
samuelfreiberg: Agents already come with a set of software installed, see [this link ](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md)for windows-2019. With this, I wanted to specify we wanted the 12. version.

Also, I thought about adding it to the "setup-repo" template, but I thought it falls more under a "setup-environment" template, since it isn't dealing with our repo directly. I couldn't think of any other command to add to the "setup-environment" template so I left it out. Thoughts? 

I am open to moving it to "setup-repo" if you think that's the better route.

.ado/templates/e2e-testing-uwp.yml
samuelfreiberg: I added it to our Pipeline's "Secure Files" library in ADO. From there, I can just use the name I added it under to download on the agent machine.

tkamireh: Nice!

.ado/azure-pipelines.yml
tkamireh: Yeah I forgot that demands are only applicable to self-hosted agents. I don't think it matters too much whether it's a part of a setup-repo or setup-environment (it's harder to make this a generally useful template). The argument could be made that it's a pre-req to having a functioning repository and should be in the setup-repo template.

Again I don't think it really matters where it goes, just that if it's something every job needs it should go into one of the templates.

samuelfreiberg: I added the Node+NuGet installs to setup-repo template in the newest commit.

.ado/templates/e2e-testing-uwp.yml
```diff
@@ -0,0 +1,42 @@
...
+    displayName: "Download Azure Credentials Plugin for NuGet"
+
+  - task: PowerShell@2
+    inputs:
+      filePath: $(AzureCredentials.secureFilePath)
+      arguments: "-AddNetfx -Force"
+    displayName: "Install Azure Credentials Plugin for NuGet"
+
+  - script: |
+      yarn windows
```
tkamireh, line 25: Might be good to add a comment here explaining why you need to go through this process of starting up the app and killing it, but leaving the server running.

samuelfreiberg: Updated comment to be more specific in newest commit.

.ado/templates/setup-repo.yml
```diff
@@ -0,0 +1,23 @@
+# This template install dependencies, builds, and bundles this repo
+steps:
+  - task: NodeTool@0
+    inputs:
+      versionSpec: "12.x"
+    displayName: "Install Node.js"
```
HeyImChris, line 6: lower case "Install"?

.ado/templates/setup-repo.yml
```diff
@@ -0,0 +1,23 @@
...
+
+  - task: NuGetToolInstaller@0
+    inputs:
+      versionSpec: "5.6.0"
+    displayName: "Install Nuget"
```
HeyImChris, line 11: minor again- lowercase the display name

.ado/templates/setup-repo.yml
samuelfreiberg: In regards to my comment in "azure-pipelines.yml", I wanted to lowercase only the scripts that run within our repo to make it easier to differentiate the steps when viewing results. In this case, NodeTool finds, downloads, and caches a version of Node and adds it to PATH on the agent machine. Since it's not a script within our repo, I thought I would leave it uppercase. Does that make sense? 

I know it's a little weird, but I was having trouble following the CI results in ADO and this just made it a bit easier.

HeyImChris: Correct me if I'm wrong, but it sounds like you're lowercasing build steps with repo specific scripts/content? I'm probably still not totally understanding but from the screenshot below, would we then want to lowercase "Kill FluentTester Process" or other repo specific steps like that then?

Fwiw this is super minor, but let's get on the same page for how we format our display names cross-plat.

Screen Shot 2020-06-09 at 2 28 02 PM

samuelfreiberg: "Correct me if I'm wrong, but it sounds like you're lowercasing build steps with repo specific scripts/**content**?"

Not the content part. The benchmark I was working off was:

Lowercase the displayName **iff** this script is executed **within** the repo.In other words,  is this script in a package.json within our repo? If so, lowercase. If not, uppercase

For the "Kill FluentTester", that is a PowerShell command to stop a process, not a specific command within the repo. This was the most helpful way for me looking at the results in Azure-pipelines. If you feel differently, I'm open to hear your thoughts.

PR #263, opened 2020-06-11
Add onRestoreFocus callback and improve test page capabilities

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Description of changes

Add onRestoreFocus callback that defers focus setting responsibilities to JavaScript.  Event provides 'containsFocus' parameter that is true when the Callout contained focus during dismissal.

### Verification

Manual verification with test page improvements.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/src/Callout.types.ts
```diff
@@ -92,6 +101,18 @@ export interface ICalloutProps extends ICalloutTokens {
...
+   * Callback invoked during callout dismissal; if set, focus will not be restored by the callout and onRestoreFocus must
+   * result in focus being moved to the appropriate focusable target.
+   *
+   * The callee should carefully consider their scenarios to avoid dropping focus, or inappropriately
+   * moving focus from another component.  Focus is not guaranteed to have entered the React-Native surface at all, and
+   * this callback is most appropriate for components strictly controlling focus.
+   *
+   * restoreFocusEvent.nativeEvent.containsFocus is true if the Callout had focus while being dismissed.
+   */
+  onRestoreFocus?: (restoreFocusEvent: OnRestoreFocusEvent) => void;
```
tkamireh, line 114: I'm not sure I actually understand the purpose here (on first glance seems excessive) - having talked offline it sounded like this was to match a FluentUI (web) API, right? When would the corresponding native event get dispatched (I'm wondering whether it would be after callout teardown -- potentially too late and getting you in weird timing situations)? Is `onDismiss` too early/late for consumers to do any imperative focusing of their own + tracking focused state should be relatively easy for wrapping components.

packages/components/Callout/src/Callout.types.ts
```diff
@@ -26,6 +26,15 @@ export type DirectionalHint =
   | 'bottomCenter'
   | 'bottomRightEdge';
 
+export interface OnRestoreFocusEvent {
```
tkamireh, line 29: nbd, I'd consider just `RestoreFocusEvent`

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -68,22 +87,44 @@ export const CalloutTest: React.FunctionComponent<{}> = () => {
...
           target={anchorRef}
           onDismiss={onDismissStandardCallout}
           onShow={onShowStandardCallout}
+          onRestoreFocus={customRestoreFocus ? onRestoreFocusStandardCallout : undefined}
           accessibilityLabel="Standard Callout"
-          setInitialFocus
+          setInitialFocus={shouldSetInitialFocus}
         >
-          <View style={{ height: 200, width: 400 }}>
+          <View style={{ padding: 20, borderWidth: 2, borderColor: 'black' }}>
```
tkamireh, line 127: height and width props no longer necessary?

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -94,7 +135,7 @@ export const CalloutTest: React.FunctionComponent<{}> = () => {
 
       <View ref={custBtnRef} style={{ flexDirection: 'row' }}>
         <Button content="Press for Callout" onClick={toggleShowCustomizedCallout} />
-        <Text>
+        <Text selectable={true}>
```
tkamireh, line 138: any reason for this? Guessing some possibly related testing?

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -60,6 +68,17 @@ export const CalloutTest: React.FunctionComponent<{}> = () => {
     setShowCustomizedCallout(false);
   }, [setIsCustomizedCalloutVisible, setShowCustomizedCallout]);
 
+  const onRestoreFocusStandardCallout = React.useCallback(
+    (restoreFocusEvent: OnRestoreFocusEvent) => {
+      if (restoreFocusEvent.nativeEvent.containsFocus && decoyBtn1Ref && decoyBtn1Ref.current) {
```
tkamireh, line 73: good opportunity for optional chaining :)

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
PPatBoyd: yeah just some padding around the button, looks nicer

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
PPatBoyd: ahh no reason for this here, was trying to test behavior for another issue -- thanks!

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
PPatBoyd: on containsFocus or decoybtn?  decoybtn has other fun with undefined vs null that makes optional chaining suck (see findNodeHandle PR I just squashed)

packages/components/Callout/src/Callout.types.ts
PPatBoyd: so this replaces the deprecated prop doNotRestoreFocus : boolean , but makes the same native-side adjustment of not attempting to place focus back where it was (note: focus will still end up 'somewhere' but the behavior is different).  When intentionally managing focus for complex situations, this is useful, and yeah I think it's a bit redundant w/ onDismiss but onDismiss w/ 'containsFocus' info is a little bit "why is this a thing", so two callbacks are relatively fine on top of adding the native-side behavior.

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
tkamireh: why not both? I think the whole if/else if block could be something like
```
restoreFocus?.nativeEvent?.containsFocus  ? 
   decoyBtn1Ref?.current?.focus?.()  :
   decoyBtn2Ref?.current?.focus?.();
```
It's probably too hard to read, but the syntax + semantics here should work (undefined and null are both nullish, and that's what matters for optional chaining). Why is optional chaining on decoyBtn bad?

packages/components/Callout/src/Callout.types.ts
tkamireh: Yeah I suppose it's just a matter of how much tracking you want to offload to the wrapping component. Still seems a little weird to me, but I will not get in the way of API alignment

PPatBoyd: it's more enabling the offloading; this is a niche-use API, the default should be appropriate for the vast majority of uses

apps/fluent-tester/src/RNTester/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -69,11 +69,11 @@ export const CalloutTest: React.FunctionComponent<{}> = () => {
...
 
   const onRestoreFocusStandardCallout = React.useCallback(
-    (restoreFocusEvent: OnRestoreFocusEvent) => {
-      if (restoreFocusEvent.nativeEvent.containsFocus && decoyBtn1Ref && decoyBtn1Ref.current) {
-        decoyBtn1Ref.current.focus();
-      } else if (!restoreFocusEvent.nativeEvent.containsFocus && decoyBtn2Ref && decoyBtn2Ref.current) {
-        decoyBtn2Ref.current.focus();
+    (restoreFocusEvent: RestoreFocusEvent) => {
+      if (restoreFocusEvent?.nativeEvent?.containsFocus) {
+        decoyBtn1Ref?.current?.focus();
```
tkamireh, line 74: you can make `focus()` into `focus?.()` - slightly safer

PR #279, opened 2020-06-16
Add memo-cache package, consumption in framework, and input style caching

JasonVMo:
## Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android

## Description of changes

There are three main parts of this change:

### Addition of @fluentui-react-native/memo-cache package

his package implements a hierarchical memoization cache using an API pattern that mimics the react.js useMemo hook. It also provides an implementation of traditional JavaScript memoization built using the react patterned utility.

Efforts were made to keep this as compact as possible. See the README.md file for a longer description.

### Switch framework packages to use memo-cache

Previously the caching was extremely ad-hoc, sometimes using WeakMap, sometimes using raw objects, merging keys in a fairly unsophisticated way. This switches the framework to use the new caching package.

In many cases APIs previously were passed a cache object and a key prefix, the staged nature of the caching package allows for just passing the cache implementation function which bundles it all together.

### Add memoized style merging for slots

This isn't done at every level of style merging, but at the point where we are about to render a slot in the framework, memoize the merge results to maintain object identity. In practice the change is as follows:

**Old Behavior**

- `<Text prop1="foo">Hello</Text>` - would internally create a style which would be the __same__ object from render to render.
- `<Text prop1="foo" style={myStyle}>Hello</Text>` - would merge myStyle with the internal style every time resulting in a new object.

In the second case, with an injected style, it means that the internal shallow props compare will always come up different. This is the primary issue today with `Stack` and why it is not yet included in the core package. In essence `Stack` usually injects a style on its children.

**New Behavior**

As long as myStyle has the same object identity, the merge result will also have the same object identity. This allows shallow props compares to work and unblocks fixing the `Stack` component. The remaining work there is ensuring that the object identities for styles passed to children are consistent from render to render.

## Verification

- Extensive unit testing in the memo-cache package
- Ran through win32 components in the tester, side-by-side with the unchanged versions
- Some debugger validation to ensure styles are found in the cache

### Pull request checklist

- [x] Automated Tests
- [ ] N/A - Keyboard Accessibility
- [ ] N/A - Voiceover
- [ ] N/A - Internationalization and Right-to-left Layouts


packages/framework/memo-cache/CHANGELOG.md
```diff
@@ -0,0 +1,128 @@
+# Change Log - @uifabricshared/immutable-merge
```
tkamireh, line 1: Looks like the wrong changelog 😊, I think the CHANGELOG.json has the same problem.

packages/framework/memo-cache/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### getMemoCache
+
+To get an instance of the memo cache to work with, a caller starts by calling `getMemoCache`.
+
+```ts
+export function getMemoCache<T = any>(key?: object): GetMemoValue<T>;
+```
+
+This function takes an optional parameter `key` which can be an object to use as a base cache reference.
```
tkamireh, line 58: Perhaps a more descriptive param name would be `globalCache`. Also if it serves as a base cache reference, is there some type guarantees it needs to make (haven't gotten very far through this review yet)

packages/framework/memo-cache/README.md
```diff
@@ -0,0 +1,141 @@
...
+| `factory` | This is typically a function, often a simple closure, which returns a value. This value will be cached for a given set of keys. Subsequent calls will just return the value without executing the function. This can also be a value type in which case that will be returned directly. |
+| `keys`    | Variable parameter list, used as the keys for caching                                                                                                                                                                                                                                   |
+
+The return result is a tuple with two elements containing:
+
+- The result of the factory function, or the value of factory if it is not a function
+- A function for caching which is local to the previous keys queried
+
+This recursive calling pattern can be used to create branches piece by piece in component execution. See the examples below for more detail.
```
tkamireh, line 48: It took me a second to understand what this meant. It's not immediately obvious how and why it's useful (after considering it I see how it is). I thought about it for a second, and maybe the right way of framing is that this allows for a natural pipelining of caches/memoization and that it's keys that are the branching mechanism.

packages/framework/memo-cache/README.md
```diff
@@ -0,0 +1,141 @@
...
+// get a unique cache for this component
+const myComponentCache = getMemoCache();
+
+// component is a function that takes props
+export const MyComponent = (props: IMyComponentProps) => {
+  const theme = useContext(ThemeContext);
+  const newProps = { ...props };
+
+  // get the style, cached against the theme so it will only be called once, note that because
+  const [style, nextCache] = myComponentCache(() => {
```
tkamireh, line 127: to drive home the point about keys being the branching mechanism maybe call it `themeLocalCache` or something similar.

packages/framework/memo-cache/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+| Param     | Description                                                                                                                                                                                                                                                                             |
+| --------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `factory` | This is typically a function, often a simple closure, which returns a value. This value will be cached for a given set of keys. Subsequent calls will just return the value without executing the function. This can also be a value type in which case that will be returned directly. |
+| `keys`    | Variable parameter list, used as the keys for caching                                                                                                                                                                                                                                   |
```
tkamireh, line 41: might be good to note that the ordering of keys is important here, at least as I've understood the reduce code in `getCacheEntry`

packages/framework/memo-cache/src/getCacheEntry.ts
```diff
@@ -0,0 +1,30 @@
...
+  noargs?: CacheEntry<TGet>;
+  str?: { [key: string]: CacheEntry<TGet> };
+  obj?: WeakMap<object, TGet>;
+};
+
+function ensureAndReturn(entry: CacheEntry<any>, key: keyof CacheEntry<any>): CacheEntry<any> | { [key: string]: CacheEntry<any> } {
+  return (entry[key] = entry[key] || {});
+}
+
+function jumpToCacheEntry(entry: CacheEntry<any>, val: any): CacheEntry<any> {
```
tkamireh, line 13: This code is fairly internal and hidden, so it shouldn't matter too much, but it could benefit from comments or at least less reliance on chaining assignment operators. It is somewhat difficult to read as is.

packages/framework/memo-cache/src/getMemoCache.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const _baseEntry: CacheEntry<any> = {};
+
+function getMemoValueWorker<T, TGet = any>(entry: CacheEntry<any>, factory: T | ValueFactory<T>, keys: any[]): [T, GetMemoValue<TGet>] {
+  const foundEntry = getCacheEntry(entry, keys);
+  // check the key being set, not the value to disambiguate an undefined factory result/value from never having run the factory
+  if (!foundEntry.hasOwnProperty('value') && factory !== undefined && factory !== null) {
+    foundEntry.value = typeof factory === 'function' ? (factory as ValueFactory<T>)() : factory;
+  }
+  return [foundEntry.value, (fact: TGet | ValueFactory<TGet>, args: any[]) => getMemoValueWorker<TGet>(foundEntry, fact, args)];
```
tkamireh, line 14: What's the desired behavior when `factory = null`? I see the tests check for value being `undefined`, but it's not particularly crazy to say that `null` is a valid value and that if I pass `null` in as my factory I expect to get `null` back and not `undefined`, right? That said, it is a strange case.

JasonVMo: The reason for it is to just make it a no-op in the case where you just want to retrieve the sub-cache.  Originally it didn't support void functions (really only useful for the memoize helper) and when I added void function support and memoize support (one of the last changes) I didn't update it. Frankly I don't think this is a real pattern and it should probably just be the hasOwnProperty check.

packages/framework/memo-cache/README.md
tkamireh: Looks like it doesn't need any type gurantees

packages/framework/memo-cache/CHANGELOG.md
JasonVMo: Doh! Obviously started this package by just copying immutable merge and modifying it :)

packages/framework/memo-cache/README.md
JasonVMo: Hmm...maybe weakKey or globalKey or something like that.  It is simply an identify reference and can be any object (including arrays and functions) and will be weakly held.

packages/framework/memo-cache/README.md
JasonVMo: Good idea

packages/framework/memo-cache/src/getCacheEntry.ts
JasonVMo: Good point, put a bunch of comments in.

packages/framework/memo-cache/src/getMemoCache.ts
JasonVMo: Now passing null returns null and undefined returns undefined correctly with the change.

PR #282, opened 2020-06-17
Web Tester

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android
- [x] web

### Description of changes

Still a work in progress. This PR is for a new project configuration, including Webpack, for the FluentUI web platform tester with successfully being able to render the FluentUI Checkbox component. 

### Verification
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/web/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@types/node": "^12.0.0",
+    "@types/react": "^16.9.0",
+    "@types/react-dom": "^16.9.0",
+    "react": "^16.13.1",
+    "react-art": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "react-native": "0.62.2",
+    "react-native-web": "^0.12.3",
+    "react-scripts": "3.4.1",
+    "typescript": "~3.7.2",
```
JasonVMo, line 20: You should see if you can line up the versions for supporting packages like typescript with the ones already in the repo. Look in /scripts/package.json for the canonical versions of many of these. Right now it looks like we have typescript: "3.8.3" and "webpack": "^4.29.5". The webpack version should resolve to match 4.42.0 (which you have here) but I would suggest updating the one in scripts to match this version.

So try changing this TS version to match the version in scripts, then change scripts webpack to match this one.

apps/web/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "webpack": "4.42.0"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
```
JasonVMo, line 27: react-scripts are a good way to get started, longer term we probably want to eject as typically the automatic configuration will eventually start working against you. No rush on this, just pointing out that just about everyone ends up ejecting at some point so don't be afraid to consider that as an option in the future if you get stuck.

apps/web/package.json
```diff
@@ -0,0 +1,47 @@
...
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "devDependencies": {
+    "@types/react-native": "0.62.2"
```
JasonVMo, line 45: Should probably be "^0.62.2". 

If RN was at >= 1.0.0 the ^ would give flexibility to bump the middle number, given that it is < 1.0.0 the ^ gives flexibility to bump the last digit or patch version.  The types package patch digit doesn't quite follow the RN patch digit and generally we use the latest version that matches the primary react-native version, in this case 62, thus we want 0.62.x

apps/web/README.md
```diff
@@ -0,0 +1,61 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
tkamireh, line 1: I would get rid of this top line, the process we have set up is pretty unique right now and fairly disconnected from what you get with create-react-app.

apps/web/README.md
```diff
@@ -0,0 +1,61 @@
...
+4. Launch the FluentUI Web Tester app:
+
+```
+    yarn web
+```
+5. Go to `localhost:8080` in your browser of choice and you will see the FluentUI Checkbox component show up.
+
+## Available Scripts
+
+In the project directory, you can run:
```
tkamireh, line 22: Is this part of the README still valid, I thought these scripts didn't work anymore?

apps/web/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "web",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@fluentui-react-native/checkbox": "0.3.50",
+    "@testing-library/jest-dom": "^4.2.4",
```
tkamireh, line 7: What happens if you get rid of the @testing-library dependencies? They may end up being unneccessary. Lets opt for a minimal set of dependencies.

apps/web/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "typescript": "3.8.3",
+    "webpack": "4.42.0"
+  },
+  "scripts": {
+    "web": "webpack-dev-server"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "browserslist": {
```
tkamireh, line 28: Can you remind me what these production and development sections are for?

apps/web/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "react-native": "0.62.2",
+    "react-native-web": "^0.12.3",
+    "typescript": "3.8.3",
+    "webpack": "4.42.0"
+  },
+  "scripts": {
+    "web": "webpack-dev-server"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
```
tkamireh, line 26: We should in the near future look at adding a build script that will actually run eslint on the code in this project. When we do that we can try to configure things in the same way that they've been configured across other projects as well.

apps/web/src/index.tsx
```diff
@@ -0,0 +1,29 @@
+// 'use strict';
+
+// import { customRegistry, FabricTester, IFabricTesterProps } from '@fluentui-react-native/tester';
```
tkamireh, line 3: Let's delete these commented lines :)

apps/web/web/index.html
```diff
@@ -0,0 +1,42 @@
+<html lang="en">
+
+<head>
+  <meta charset="utf-8" />
+  <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta name="theme-color" content="#000000" />
+  <meta name="description" content="Web site created using create-react-app" />
+  <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
tkamireh, line 9: Remind me do we actually want these pngs displayed? If not we can delete the png files from this as well.

packages/utils/adapters/package.json
```diff
@@ -26,7 +26,8 @@
     "@uifabricshared/eslint-config-rules": "^0.1.1",
     "react-native": "0.62.2",
     "react-native-web": "^0.12.2",
-    "react-native-windows": "0.0.0-master.47"
+    "react-native-windows": "0.0.0-master.47",
+    "react-dom": "^16.13.1"
```
tkamireh, line 30: So as far as this hack goes, have you tried adding the react-dom dep as a devdep in the root package.json. I'm not sure whether that's the recommended fix or not, but is likely cleaner than having it here if it in fact works.

scripts/package.json
```diff
@@ -60,7 +60,7 @@
     "just-scripts": "^0.38.0",
     "prettier": "~1.17.0",
     "typescript": "3.8.3",
-    "webpack": "^4.29.5",
+    "webpack": "4.42.0",
```
tkamireh, line 63: Can you make this ^4.42.0 instead? Generally we should try to avoid specific version pinning when possible, let me know if it doesn't work :).

apps/web/src/setupTests.ts
```diff
@@ -0,0 +1,5 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom/extend-expect';
```
tkamireh, line 5: Do you need this now, I'd probably get rid of this file if it doesn't break things.

apps/web/src/serviceWorker.ts
```diff
@@ -0,0 +1,149 @@
+// This optional code is used to register a service worker.
```
tkamireh, line 1: So I'm not too familiar, but this isn't getting spun as a prod web-app, right (it would be kind of cool to do this in the future). This may be another one of those things that we can get rid of for now.

apps/web/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
tkamireh, line 1: Can we get rid of this file?

tkamireh:
I left a good number of comments, most of them should be trivial to address 😃. I think once those are addressed we should feel good to merge this in!

apps/web/package.json
gulnazsayed: This was part of the create-react-app initial configuration but it seems to work fine without these production and development sections now, so I'll take them out.

tkamireh:
You'll want to run `yarn change` to generate change files, without them CI will fail

change/@uifabricshared-theming-react-native-2020-07-02-16-00-55-web-tester.json
```diff
@@ -0,0 +1,8 @@
+{
```
tkamireh, line 1: I'm surprised that it generated changes for packages you didn't touch. Might be a bad carry-over from something else.

PR #283, opened 2020-06-17
Remove stale yarn.lock files in ios/macOS

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Previously, the ios and macos test apps were not fully integrated into the lerna monorepo system, and thus had their own yarn.lock files. This prevented the macos test app from upgrading to `0.61.41`, even though package.json should have resolved to "latest 0.xx release above 0.61.0". Removing the yarn lock file at `apps/macos` fixed that. We might as well remove the ios one as well.

### Verification

`yarn && yarn build` at root work fine

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
@HeyImChris Every platform has a `.gitignore` / `.gitattributes`, so I'll leave those as is. I'm pretty sure those are still needed, because I'm not sure the main `.gitignore` at root will cover all the cases

PR #302, opened 2020-06-23
Migrate iOS test app to react-native-test-app

tido64:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Migrates iOS test app to `react-native-test-app`.

### Verification

| Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-23 at 19 52 18 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-23 at 19 52 46 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

cc @FalseLobster, @HeyImChris


apps/ios/react-native.config.js
```diff
@@ -18,7 +18,7 @@
 module.exports = {
   project: {
     ios: {
-      project: "apps/ios/src/FluentUITester.xcworkspace"
```
HeyImChris, line 21: Out of curiosity, what happened to cause it to need to be a relative path now?

apps/ios/src/Podfile
```diff
@@ -1,80 +1,41 @@
...
-  end
-end
+workspace 'FluentUITester.xcworkspace'
 
-target 'FluentUITester' do
-  # YOUR REACT NATIVE PATH (from the iOS folder)
-  rn_path = '../../../node_modules/react-native'
+bundle_script = %{
+export NODE_BINARY=node
+#{__dir__}/../../../node_modules/react-native/scripts/react-native-xcode.sh
```
HeyImChris, line 7: Save this off into an rn_path var since we use it later below too?

apps/ios/src/index.tsx
```diff
@@ -1,15 +1,17 @@
...
+import { AppRegistry, SafeAreaView } from 'react-native';
 import { ThemeProvider } from '@uifabricshared/theming-react-native';
 import { FabricTester, IFabricTesterProps, customRegistry } from '@fluentui-react-native/tester';
 
 const FluentTester: React.FunctionComponent<IFabricTesterProps> = props => {
   return (
-    <ThemeProvider registry={customRegistry}>
-      <FabricTester {...props} />
-    </ThemeProvider>
+    <SafeAreaView>
```
HeyImChris, line 10: Thanks for adding this!

apps/ios/react-native.config.js
tido64: I don't think it's ever been relative? In `react-native-test-app`, we've always used it like this. I actually got an error here, which is why I had to change it in the first place.

apps/ios/src/index.tsx
tido64: I'll extract this into a separate PR since it's useful regardless.

tido64: Here: #303. I ran `yarn change` but it didn't do anything. Is that normal?

HeyImChris: Yea that's fine. We don't verify change files like how you would if you're using rush in, say, sdx-platform. I don't run yarn change for any changes in this repository

tido64: Your [README](https://github.com/microsoft/fluentui-react-native#beachball) suggests otherwise :stuck_out_tongue:

HeyImChris: Ah yes sorry for the confusion- that applies to Windows. The Apple documentation is lacking right now

HeyImChris:
Windows PR failure is a known issue right now and this PR only affects Apple files and the yarn.lock file. The Apple validation verifies yarn.lock is valid, so pushing this in now to get more progress on the iOS side.

Saadnajmi:
!!! =D 
Excited to see this go in!

PR #308, opened 2020-06-30
Update flurn readme.md for expanding component documentation

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Documentation update for updating public website documentation.

### Verification

N/A

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,16 @@ export default HelloWorldApp;
 
 Our component documentation is hosted on the [FluentUI documentation](https://developer.microsoft.com/fluentui).
 
+#### Expanding Component documentation
+
+The FluentUI website is built out of the [FluentUI repository](https://github.com/microsoft/fluentui/tree/master/apps/fabric-website).  React-Native components and controls are documented in a 'cross' (cross-platform) directory in each component page directory, e.g. [Button 'cross' directory](https://github.com/microsoft/fluentui/tree/master/apps/fabric-website/src/pages/Controls/ButtonPage/docs/cross).  The FluentUI website can be run locally to verify changes, and should reflect the current state of controls that have established the *v1* set of properties on any one platform.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,16 @@ export default HelloWorldApp;
 
 Our component documentation is hosted on the [FluentUI documentation](https://developer.microsoft.com/fluentui).
 
+#### Expanding Component documentation
+
+The FluentUI website is built out of the [FluentUI repository](https://github.com/microsoft/fluentui/tree/master/apps/fabric-website).  React-Native components and controls are documented in a 'cross' (cross-platform) directory in each component page directory, e.g. [Button 'cross' directory](https://github.com/microsoft/fluentui/tree/master/apps/fabric-website/src/pages/Controls/ButtonPage/docs/cross).  The FluentUI website can be run locally to verify changes, and should reflect the current state of controls that have established the *v1* set of properties on any one platform.
+
+Since the FluentUI React Native controls are cross-platform, but represented by a single page, it's important to distinguish platform differences and limitations.  Examples include:
```
HeyImChris, line 72: Nit- Do we have guidance for how one should format cross-platform diffs to keep all documentation consistent across controls?

For instance, list the available platforms followed by the API descriptions (e.g. "iOS-only: [whatever iOS APIs there are] \n Win32/macOS-only: [some desktop specific APIs], etc....". If there's no guidance here then documentation might reveal an inconsistent story across components with every engineer applying their own version of what they think is clear.

JasonVMo: That's a great point but I think that should be documented in the fabric-website repo rather than this one, that way it ends up local to the source.

PR #316, opened 2020-07-06
FabricTester Web

gulnazsayed:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [ ] android
- [x] web

### Description of changes

Implemented FabricTester using React Native Web. Changed imports for most component tests so that they would work with React Native Web and added props to feed in component tests for each platform individually. Most component tests are working the way they are supposed to, but there are a few like Separator that don't quite work yet.

### Verification

| After                                      |
|--------------------------------------------|
| image 
image 
image
image
image
image
image
image
image
image
image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/RNTester/FabricTester.tsx
```diff
@@ -1,11 +1,11 @@
+import { Separator, StealthButton, Text } from '@fluentui/react-native';
+import { useTheme } from '@uifabricshared/theming-react-native';
```
JasonVMo, line 2: I'm curious why the reordering of the import statements? It isn't necessarily wrong, I'm just curious whether there were build issues or whether it was reordered to follow an alphabetical approach.

apps/fluent-tester/src/RNTester/TestComponents/index.ts
```diff
@@ -1,88 +1,7 @@
...
-import { PersonaTest, HOMEPAGE_PERSONA_BUTTON } from './Persona';
-import { PersonaCoinTest, HOMEPAGE_PERSONACOIN_BUTTON } from './PersonaCoin';
-import { PressableTest, HOMEPAGE_PRESSABLE_BUTTON } from './Pressable';
-import { RadioGroupTest, HOMEPAGE_RADIOGROUP_BUTTON } from './RadioGroup';
-import { SeparatorTest, HOMEPAGE_SEPARATOR_BUTTON } from './Separator';
-import { SvgTest, HOMEPAGE_SVG_BUTTON } from './Svg';
-import { TextTest, HOMEPAGE_TEXT_BUTTON } from './Text';
-import { ThemeTest, HOMEPAGE_THEME_BUTTON } from './Theme';
 
 export type TestDescription = {
```
JasonVMo, line 3: I would consider figuring out how to add a platform mask as a pattern to the various entries, either as an include list or exclude list. Something like:

```ts
{
  name: 'Button Test',
  component: ButtonFocusTest,
  testPage: HOMEPAGE_BUTTON_BUTTON,
  platforms: ['android', 'ios', 'macos', 'web', 'win32', 'windows']
}
```

Then instead of passing in the set of test descriptions you can pass in a platform setting to the RNTester where it can do filtering. 

The approach you use here is correct and safe, but is a bit tougher to maintain. In particular the authoring of the tests and the references to the tests span multiple packages, and when a test is added or updated it needs to be updated in multiple places.

tkamireh: One of the problems we've run into is related to 'bad' imports calling into things that don't exist in react-native-web. When a component imports (directly or indirectly) something like `Callout` or `FocusTrapZone` that calls `requireNativeComponent` we hit runtime type errors as this isn't something that is supplied by react-native-web (the bundler aliases rn with rn-web). I can imagine a couple solutions to this, but being really granular about the imports and where they come from seems to work. Runtime filtering per platform likely results in the problems we've been seeing. @gulnazsayed do you have any additional thoughts on this? I know you're looking at cleaning up some imports in the testComponents directory to work.

We should look into per platform bundling strategies to make stubbing things easier and runtime filtering becomes with that in place.

apps/fluent-tester/src/RNTester/FabricTester.tsx
gulnazsayed: VSCode reorders them alphabetically every time I save a file. Would you prefer I turn off this setting so it limits the number of changes made to the files?

tkamireh:
Please fill out description and provide screenshots.

apps/fluent-tester/src/RNTester/FabricTester.tsx
```diff
@@ -19,14 +21,16 @@ const EmptyComponent: React.FunctionComponent = () => {
   return <Text style={fabricTesterStyles.noTest}>Select a component from the left.</Text>;
 };
 
-// sort tests alphabetically by name
-const sortedTestComponents = allTestComponents.sort((a, b) => a.name.localeCompare(b.name));
-
 export interface IFabricTesterProps {
   initialTest?: string;
+  enabledTest: TestDescription[];
```
tkamireh, line 26: nit `enabledTests`

apps/fluent-tester/src/RNTester/RNTester.tsx
```diff
@@ -1,15 +1,91 @@
 'use strict';
```
tkamireh, line 1: Do we know how this component gets used? If it doesn't need changes we can ignore it, or better yet if it isn't used, lets delete it.

apps/web/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
tkamireh, line 1: If the svg tests aren't working on web, let's get rid of the test on web and delete this file + the new svg-inline-loader dep

apps/web/src/index.tsx
```diff
@@ -1,9 +1,88 @@
...
+    component: RadioGroupTest,
+    testPage: HOMEPAGE_RADIOGROUP_BUTTON
+  },
+  {
+    name: 'Persona Test',
+    component: PersonaTest,
+    testPage: HOMEPAGE_PERSONA_BUTTON
+  },
+  {
+    name: 'Svg Test',
```
tkamireh, line 72: if this doesn't work, let's get rid of it.

apps/web/src/index.tsx
```diff
@@ -1,9 +1,88 @@
...
+import { HOMEPAGE_PERSONA_BUTTON, PersonaTest } from '../../fluent-tester/src/RNTester/TestComponents/Persona';
+import { HOMEPAGE_PERSONACOIN_BUTTON, PersonaCoinTest } from '../../fluent-tester/src/RNTester/TestComponents/PersonaCoin';
+import { HOMEPAGE_PRESSABLE_BUTTON, PressableTest } from '../../fluent-tester/src/RNTester/TestComponents/Pressable';
+import { HOMEPAGE_RADIOGROUP_BUTTON, RadioGroupTest } from '../../fluent-tester/src/RNTester/TestComponents/RadioGroup';
+import { HOMEPAGE_SEPARATOR_BUTTON, SeparatorTest } from '../../fluent-tester/src/RNTester/TestComponents/Separator';
+import { HOMEPAGE_SVG_BUTTON, SvgTest } from '../../fluent-tester/src/RNTester/TestComponents/Svg';
+import { HOMEPAGE_TEXT_BUTTON, TextTest } from '../../fluent-tester/src/RNTester/TestComponents/Text';
+import { HOMEPAGE_THEME_BUTTON, ThemeTest } from '../../fluent-tester/src/RNTester/TestComponents/Theme';
+
+const tests: TestDescription[] = [
```
tkamireh, line 20: consider putting these in a separate file, makes things easier to read.

apps/web/src/index.tsx
```diff
@@ -1,9 +1,88 @@
...
+const FluentTester: React.FunctionComponent<IFabricTesterProps> = props => {
+  return (
+    <ThemeProvider registry={customRegistry}>
+      <FabricTester enabledTest={tests} {...props} />
+    </ThemeProvider>
+  );
+};
 
-AppRegistry.registerComponent('App', () => App);
+AppRegistry.registerComponent('FluentTester', () => FluentTester);
```
tkamireh, line 86: don't forget to re-add the `runApplication` step

PR #317, opened 2020-07-06
Add copies of Button, Text, and Stack to packages/experimental

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android

### Description of changes

As I'm working on some of the potential changes I'm adding them as a parallel experimental framework. 

- To test that in practice I'm adding some copies of Text, Button, and Stack. 
- This allows them to be iterated on without affecting production usage.
- When I submit the first PR for the experimental stuff, having copies already present will make the changes easier to see.

These are not published and not used right now so this doesn't affect any code.

### Verification

No validation outside of builds right now since the changes don't impact any runtime code.


Saadnajmi:
Apologies if the wording isn't correct:
Would this increase the overall size of the framework, when added as a dependency to another project? Or does having a separate package for "experimental" mean it's not included unless we explicitly import it?

JasonVMo:
> Apologies if the wording isn't correct:
> Would this increase the overall size of the framework, when added as a dependency to another project? Or does having a separate package for "experimental" mean it's not included unless we explicitly import it?

It's a good question. Right now these are private and not included in any package so this change doesn't impact any runtime code or bundles. 

This sets it up so that we can create parallel experimental versions of controls to prototype and compare vs. the standard ones, then switch over in a staged manner. I have some changes I will push for feedback soon that create a pared down version of the framework that is a bit simpler and reduces our bundle size, but I want to be able to demonstrate:

- current framework size vs. proposed (I recently put it the packages/libraries/framework package with bundle analyzer to enable this)
- what an existing control would look like using the new framework (which is why I'm submitting this one)

packages/experimental/Button/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "scripts": {
+    "build": "fluentui-scripts build",
+    "just": "fluentui-scripts",
+    "clean": "fluentui-scripts clean",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "verify-api": "fluentui-scripts verify-api-extractor",
+    "update-api": "fluentui-scripts update-api-extractor"
+  },
+  "dependencies": {
+    "@uifabricshared/foundation-compose": "^1.6.70",
```
tkamireh, line 22: will this dep change to the experimental version?

tkamireh:
lgtm, assuming this is mostly for diffing control impls with the experimental packages

packages/experimental/Button/package.json
JasonVMo: Yes, that will be part of the later change.

PR #318, opened 2020-07-06
Move react-native-windows to 0.62

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [x] windows
- [ ] android

### Description of changes

This moves react-native-windows onto the official 0.62 branch for RNW. This required two changes:

- change the version references, pretty self-explanatory
- there was a temporary workaround for out of tree platforms using a custom resolver in metro.config.js. A better fix was pushed upstream and the 0.62 branch takes advantage of this so I removed the bad reference.

### Verification

Right now just building the changes and running through automated stuff. Once RNW pushes 0.62.1 we should be able to pick up the fix to unblock our CI loops.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


NickGerleman:
IIRC there might be some other template changes to our vcxproj. You might see some weird failures if you try to use new community modules, but if this is working as is this seems good.

apps/windows/metro.config.js
```diff
@@ -11,9 +11,6 @@ const {getWatchFolders} = require('@uifabricshared/build-native');
 module.exports = {
   watchFolders: getWatchFolders(),
   resolver: {
-    resolveRequest: require('react-native-windows/metro-react-native-platform').reactNativePlatformResolver(
-      {windows: 'react-native-windows'},
-    ),
```
Saadnajmi, line 16: For my own understanding, is there no more custom code needed to support out of tree platforms 0.62 onwards? Asking for when we move the macOS branch to 0.62 and I need to make similar changes.

JasonVMo: 0.61 removed the haste map which greatly simplified things, then fairly recently Coates made a change to how the base metro configs were set up that made out of tree platforms far easier to author. Mac should be able to take advantage of that once it moves.

JasonVMo:
> IIRC there might be some other template changes to our vcxproj. You might see some weird failures if you try to use new community modules, but if this is working as is this seems good.

What is the right upgrade strategy for RNW consumers? Should I rerun the init sequence? Is there an upgrade command? The version we were pinned to was also based on RN 0.62, how do I pick up the other part of the changes?

apps/windows/metro.config.js
JasonVMo: For more clarity here, I think that as long as you are referencing the latest version of the @react-native-community/cli project and possibly metro you should get the appropriate changes to make this easy.

PR #319, opened 2020-07-06
Rework CI pipelines to be faster

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [ ] android

### Description of changes

Our CI loops are running really slowly right now, currently a big chunk of this time is the process of running yarn install, yarn build, then yarn bundle. In many cases each of these steps is taking 4-6 minutes.  Looking more deeply at this those steps are doing:

- `yarn`: doing the yarn install, not much we can do except for de-dupe dependencies, doing that separately
- `yarn build`: this currently does typescript builds, linting and jest tests
- `yarn bundle`: this currently bundles all platforms, even if they aren't relevant

This PR does the following:

- Create a new PR loop for building and bundling, this will run the full build loop and bundle all platforms and packages.
- Make all other PR loops use yarn build --min which skips jest/linting and is much faster
- Create dedicated win32 and windows loops
- Each of these loops will only bundle the appropriate platform
- Change the apple loops to only bundle macos/ios

This should hopefully yield faster CI turnaround times. Conceivably even more of the building could be skipped but this change is a more conservative start.

### Verification

This change only affects CI loops.

### DIFF
--- a/.ado/azure-pipelines.yml
+++ b/.ado/azure-pipelines.yml
@@ -123,11 +123,6 @@ jobs:
         workingDirectory: apps/ios
         displayName: "yarn bundle ios"
 
-      - script: |
-          yarn bundle
-        workingDirectory: apps/macos
-        displayName: "yarn bundle macos"
-
       # Select proper XCode version
       - template: templates/apple-xcode-select.yml
 
--- a/.ado/azure-pipelines.yml
+++ b/.ado/azure-pipelines.yml
@@ -5,6 +5,38 @@ variables:
   CI: true
 
 jobs:
+  # Dedicated task to build and bundle JS code, including jest tests, snapshot testing, and linting, because these things can be super
+  # time consuming they don't need to run on every CI pass, instead do a dedicated JS loop to make the platform specific tests start quicker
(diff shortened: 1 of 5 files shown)


JasonVMo:
I initially had a bundle command for macos in the apple loops but it doesn't look like that command is used in the apps/macos directory

HeyImChris:
> I initially had a bundle command for macos in the apple loops but it doesn't look like that command is used in the apps/macOS directory

Thanks for doing this Looking at the builds for this PR, any idea why the iOS CI is still taking 13+ min to run `yarn` whereas windows and win32 are closer to 7?

JasonVMo:
> > I initially had a bundle command for macos in the apple loops but it doesn't look like that command is used in the apps/macOS directory
> 
> Thanks for doing this Looking at the builds for this PR, any idea why the iOS CI is still taking 13+ min to run `yarn` whereas windows and win32 are closer to 7?

It looks like they were all having network issues. The mac one eventually timed out. Nothing in this change should impact what yarn is doing.

tkamireh:
lgtm

PR #321, opened 2020-07-08
Lehon contextmenu

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32
- [X] windows
- [ ] android

### Description of changes

Adding a new component ContextualMenu, currently only supporting single nested menu.

### Verification

ContextualMenuTest was added to our RNTester.

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -0,0 +1,53 @@
+import { IRenderData } from '@uifabricshared/foundation-composable';
+import { ICalloutProps, ICalloutTokens } from '@fluentui-react-native/callout';
+
+export const contextualMenuName = 'ContexualMenu';
+
+/*
+export interface IContextualMenuContext {
+  /*
```
samuelfreiberg, line 8: I know this is an early version, but do we want to have large chunks of commented out code left here? Maybe remove it from the PR but keep it locally?

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { useAsPressable, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+
+export const ContextualMenuItem = compose<IContextualMenuItemType>({
+  displayName: contextualMenuItemName,
+  usePrepareProps: (userProps: IContextualMenuItemProps, useStyling: IUseComposeStyling<IContextualMenuItemType>) => {
+    const {
+      icon,
+      text,
+      accessibilityLabel = userProps.text,
```
samuelfreiberg, line 20: The user should have the option to give an alternate accessibilityLabel other than the text of the item. See Checkbox/RadioButton for example. They have an "ariaLabel" prop, and then set: "accessibilityLabel: ariaLabel ?? content"

packages/components/ContextualMenu/src/ContextualMenuItem.types.ts
```diff
@@ -0,0 +1,75 @@
...
+import { IFocusable, IPressableState } from '@fluentui-react-native/interactive-hooks';
+import { IViewWin32Props } from '@office-iss/react-native-win32';
+
+export const contextualMenuItemName = 'ContextualMenuItem';
+
+export interface IContextualMenuItemState extends IPressableState {
+  /**
+   * Whether the menu item is clicked
+   */
+  // clicked?: boolean;
```
samuelfreiberg, line 16: Same question as above, not sure about leaving the commented code. Maybe someone else can chime in?

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
ksiler: I would agree. I can't really think of an example where a user wouldn't want the accessibilityLabel to be the same as the text other than if they didn't provide text and just an icon (kind of an odd decision but it could be done, right?). In that case, you'd want to allow the user to describe the icon. I also think it's good to allow users to customize the text if they need to. 

On another note, is there a reason why we used the name "ariaLabel" on Checkbox/RadioButton? Could it not have been "accessibilityLabel"?

samuelfreiberg: That's what web calls it and what it's more widely known as, I guess I forgot to change it back though. Next time I work on either of those I'll change it :)

packages/components/ContextualMenu/src/ContextualMenuItem.types.ts
JasonVMo: I would generally remove it before checkin unless there is value in leaving it in. An example might be comments that are indicative of a future direction or a proposal. In that case I might add some context around why the commented code is present. Otherwise I would remove it.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
ksiler: Thanks Sammy!

lenahong: The current implementation actually lets users set accessibilityLabel themselves. When they are not set, it just defaults to Text.

packages/components/ContextualMenu/src/ContextualMenuItem.types.ts
lenahong: I'll remove them. Thanks!

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {renderData!.state.icon && <Slots.icon source={renderData.slotProps!.icon.source} />}
+          {renderData!.state.content && <Slots.content />}
+          {children}
+        </Slots.stack>
+      </Slots.root>
+    );
+  },
+  slots: {
+    root: View,
+    stack: { slotType: View, filter: filterViewProps },
```
JasonVMo, line 80: you can probably remove the filters here. It seems like the main scenario the filters would be useful would be when props or tokens are being copied via spread like what happens with root above, where it's desireable to strip off the ones that don't apply to the inner element. (I also know that I probably started the rampant use of filters in the initial button implementation but in most cases they may not be necessary.

packages/components/ContextualMenu/src/ContextualMenuItem.types.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Title (tooltip) text displayed when hovering over an item.
+   */
+  title?: string;
+  dismissMenu?: (e?: any, dismissAll?: boolean) => void;
+
+}
+
+export interface IContextualMenuItemSlotProps {
+  root: React.PropsWithRef<IViewWin32Props>;
```
JasonVMo, line 57: Why IViewWin32Props vs just ViewProps? Does the implementation need to be forked between platforms?

apps/fluent-tester/src/RNTester/TestComponents/ContextualMenu/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './ContextualMenuTest';
+export * from './consts';
+
```
tkamireh, line 3: nit extra line :)

change/@fluentui-react-native-2020-07-07-19-11-38-lehon-contextmenu.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
tkamireh, line 2: I would say at the very least this is a patch. You've added to what the metapackage exports. I think in an ideal world this is a minor version bump.

packages/components/ContextualMenu/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "dependencies": {
+    "@uifabricshared/foundation-compose": "^1.6.51",
+    "@fluentui-react-native/adapters": ">=0.3.13 <1.0.0",
+    "@fluentui-react-native/interactive-hooks": ">=0.4.20 <1.0.0",
+    "@fluentui-react-native/text": ">=0.5.49 <1.0.0",
+    "@fluentui-react-native/tokens": ">=0.4.48 <1.0.0"
+  },
+  "devDependencies": {
+    "@office-iss/react-native-win32": "0.62.0-preview.3",
```
tkamireh, line 28: If you are able to get rid of the `IViewWin32*` usage (I think that's the only usage), don't forget to get rid of it here :)

packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface IContextualMenuProps extends Omit<ICalloutProps, 'setInitialFocus'> {
+  /*
+  * Whether to set initial focus on the contextual menu container, as opposed to the first menu item.
+  */
+  shouldFocusOnContainer?: boolean;
+
+  /*
+  * Whether tofocus on the menu when mounted
+  */
+  shouldFocusOnMount?: boolean;
```
tkamireh, line 20: So this is effectively a replacement for `setInititalFocus`? Am I right in guessing that shouldFocusOnMount is what fluent web currently uses?

packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+export interface IContextualMenuProps extends Omit<ICalloutProps, 'setInitialFocus'> {
+  /*
+  * Whether to set initial focus on the contextual menu container, as opposed to the first menu item.
+  */
+  shouldFocusOnContainer?: boolean;
+
+  /*
+  * Whether tofocus on the menu when mounted
```
tkamireh, line 18: nit `to focus`

packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -0,0 +1,33 @@
...
+  */
+  shouldFocusOnMount?: boolean;
+}
+
+export type IContextualMenuSlotProps = {
+  root: ICalloutProps;
+};
+
+export type IContextualMenuRenderData = IRenderData<IContextualMenuSlotProps>;
```
tkamireh, line 27: I may have missed something, but I don't see usage of this type anywhere, we can air on the side of at least not exporting, but preferably get rid of it.

packages/components/ContextualMenu/src/ContextualMenuItem.settings.ts
```diff
@@ -0,0 +1,66 @@
+import { contextualMenuItemName, IContextualMenuItemType } from './ContextualMenuItem.types';
+import { IComposeSettings } from '@uifabricshared/foundation-compose';
+
+export const settings: IComposeSettings<IContextualMenuItemType> = [
+  {
+    tokens: {
+      backgroundColor: 'buttonBackground',
+      color: 'buttonText',
+      borderColor: 'buttonBorder',
+      borderWidth: 1,
```
tkamireh, line 10: So I think these are reasonable, but I am curious whether talking to PM and design you felt like there was a possibility that things like borderWidth and borderRadius should be theme lookups instead. I could imagine some UIs having themes where these values need to change.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
tkamireh: https://wesbos.com/destructuring-default-values in case anyone is curious about default values when destructuring

packages/components/ContextualMenu/src/ContextualMenu.types.ts
lenahong: Yes shouldFocusOnMount is what fluent web currently uses. This might actually be worth a discussion. Callout supports setInitialFocus currently, and I'm now thinking we should setting on one for both ContextualMenu and Callout.

lenahong: jk I talked to Jon and we should keep them separate for each component

packages/components/ContextualMenu/src/ContextualMenuItem.settings.ts
lenahong: I agree with you. I'll remove borderWidth and borderRadius from settings

packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -0,0 +1,53 @@
...
+  onItemClick?: (key: string) => boolean | void;
+}
+
+
+export interface IContextualMenuState {
+  context: IContextualMenuContext;
+}
+*/
+
+export type IContextualMenuTokens = ICalloutTokens;
```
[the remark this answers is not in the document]
lenahong: I'll make the changes for ContextualMenu in this pr, and create a separate one for the other components?

packages/components/ContextualMenu/src/ContextualMenuItem.types.ts
lenahong: Nope, ViewProps works fine. I'll change it

packages/components/ContextualMenu/src/ContextualMenu.types.ts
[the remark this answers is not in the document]
JasonVMo: Let's definitely handle the other components separately. There are a few discussions to be had about the patterns that the components should follow. For right now I would say just update things as you come across them, the general guidance is to try to make our coding patterns to be fairly standard for TS.

packages/components/ContextualMenu/src/ContextualMenu.types.ts
PPatBoyd: @lenahong what's the intended difference?

lenahong: He says that ContextualMenu actually passes shouldFocusOnMount to Callout's setInitialFocus and they should get unified eventually, but for now they're different.

PR #325, opened 2020-07-09
Enhance E2E Reporting + CI Reliability

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32
- [X] windows
- [ ] android

### Description of changes

• Created template to publish build artifacts for E2E testing
• Made one E2E template handle both Win32+UWP testing
• Added individual tests to ensure app is booted correctly, ensuring that if this fails, no subsequent tests will be run
• Added common headings on each test page. Previously, some pages had headings and some didn’t, and it looked a bit messy.

### Verification

Manual

### Pull request checklist

- [X] Automated Tests
- [ X Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/templates/e2e-testing-uwp.yml
```diff
@@ -40,7 +40,7 @@ steps:
 
   # Wait for app to launch. A workaround to avoid WinAppDriver error: Failed to locate opened application window with appId
   - powershell: |
-      Start-Sleep -Seconds 60
+      Start-Sleep -Seconds 90
```
JasonVMo, line 43: Ugh, this would be more complicated but is there a way to get some sort of positive handoff?  Generally any sleep type call with a fixed number tends to lead to intermittency.  Note that I'm not asking for a fix for this review, just something to consider for the future.

samuelfreiberg: This was from RNW's CI that I copied over. We may not actually need it because of how I implemented the steps. I'm going to remove this task and observe the affects. I believe it'll be fine without it, thanks for pointing this out.

PR #328, opened 2020-07-13
Fix macOS test app

tido64:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Fix macOS test app failing to bundle JS, and modify CI to make sure it doesn't regress.

### Verification

| n/a | image |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/macos/react-native.config.js
+++ b/apps/macos/react-native.config.js
@@ -15,19 +15,8 @@
  */
 'use strict';
 
-const macSwitch = '--use-react-native-macos';
-const windowsSwitch = '--use-react-native-windows';
+const path = require('path');
 
-if (process.argv.includes(macSwitch)) {
-  process.argv = process.argv.filter((arg) => arg !== macSwitch);
-  process.argv.push('--config=metro.config.macos.js');
-  module.exports = {
-    reactNativePath: require.resolve('react-native-macos'),
-  };
-} else if (process.argv.includes(windowsSwitch)) {
-  process.argv = process.argv.filter((arg) => arg !== windowsSwitch);
-  process.argv.push('--config=metro.config.windows.js');
-  module.exports = {
-    reactNativePath: 'node_modules/react-native-windows',
-  };
-}
+module.exports = {
+  reactNativePath: path.dirname(require.resolve('react-native-macos/package.json')),
+};
--- a/.ado/azure-pipelines.yml
+++ b/.ado/azure-pipelines.yml
@@ -120,8 +120,8 @@ jobs:
 
       - script: |
           yarn bundle
-        workingDirectory: apps/ios
-        displayName: "yarn bundle ios"
+        workingDirectory: apps/$(relative_directory)
+        displayName: "yarn bundle $(relative_directory)"
 
       # Select proper XCode version
       - template: templates/apple-xcode-select.yml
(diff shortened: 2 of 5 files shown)


apps/macos/metro.config.macos.js
```diff
@@ -7,7 +7,7 @@ const path = require('path');
 const blacklist = require('metro-config/src/defaults/blacklist');
 const {getWatchFolders} = require('@uifabricshared/build-native');
 
-const rnmPath = path.resolve(__dirname, 'node_modules/react-native-macos');
+const rnmPath = path.dirname(require.resolve('react-native-macos/package.json'));
```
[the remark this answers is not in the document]
Saadnajmi: Actually, why do we need to use dirname at all? Why can't we just do `require.resolve('react-native-macos')`

apps/macos/react-native.config.js
```diff
@@ -15,14 +15,16 @@
...
 
 if (process.argv.includes(macSwitch)) {
   process.argv = process.argv.filter((arg) => arg !== macSwitch);
   process.argv.push('--config=metro.config.macos.js');
   module.exports = {
-    reactNativePath: require.resolve('react-native-macos'),
+    reactNativePath: path.dirname(require.resolve('react-native-macos/package.json')),
   };
 } else if (process.argv.includes(windowsSwitch)) {
   process.argv = process.argv.filter((arg) => arg !== windowsSwitch);
```
Saadnajmi, line 30: For curiosity sake: Any idea why we don't need to follow the same pattern for `react-native windows`

apps/macos/metro.config.macos.js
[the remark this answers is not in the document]
tido64: > Was this the main reason `react-native` wasn't properly resolving to `react-native-macos` ?

Yes, `react-native-macos` was actually installed at `../../node_modules/react-native-macos` and not at `./node_modules/react-native-macos`. But we should use `require.resolve` to find the path rather than hard coding it.

> Actually, why do we need to use dirname at all? Why can't we just do `require.resolve('react-native-macos')`

`require.resolve('react-native-macos')` resolves to whatever the `main` field in `package.json` points to. In [this case](https://github.com/microsoft/react-native-macos/blob/45aa5b30b1cfedb07d889da634b164ce75613f90/package.json#L17), it would point to `/path/to/react-native-macos/Libraries/react-native/react-native-implementation.js`. There are no guarantees for what `main` can point to so it's much safer to use `package.json`.

apps/macos/react-native.config.js
tido64: We should be, but since this is the macOS test app, it doesn't really matter. We should really just remove the `windows` case and clean up the file a bit to avoid confusion. I can do that later after I figure out how to make the bundle step pass on CI.

tido64: I think this is a success :smile:

image

FYI @HeyImChris

Saadnajmi: Gotcha. Yeah we don't need that switch, but if we dont' remove it with this CR, we probably will when we move to 0.62 since we won't need the switched anymore, so I wouldn't worry about that. If you're done with work and need an owner to merge it in, I'll happily do so now.

tido64: Please go ahead and merge it. Thanks 🙏

PR #331, opened 2020-07-14
Migrate macOS test app to react-native-test-app

tido64:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Migrates macOS test app to `react-native-test-app`.

### Verification

| image | Screenshot 2020-07-14 at 10 09 49 |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


tido64:
cc @HeyImChris, @Saadnajmi

apps/macos/app.json
```diff
@@ -1,4 +1,7 @@
 {
   "name": "FluentTester-macos",
-  "displayName": "FluentTester-macos"
+  "displayName": "FluentTester-macos",
+  "components": [
```
HeyImChris, line 4: What is this doing?

apps/macos/package.json
```diff
@@ -19,8 +19,9 @@
     "@fluentui-react-native/tester": "^0.0.1",
     "@uifabricshared/theming-react-native": "0.7.67",
     "@uifabricshared/themed-stylesheet": "0.3.67",
-    "react": "16.11.0",
-    "react-native-macos": "^0.61.0-0"
+    "react-native": "0.61.5",
```
HeyImChris, line 22: Shouldn't need react-native if we're pulling in react-native-macos?

apps/macos/app.json
tido64: `components`? It's to declare the apps that you registered with `AppRegistry`. The list is used to populate the React menu so you can switch between them. There's only one app in the FluentUI test app so it's not very useful here. You can see it in use here:

react-native-test-app-macos

apps/macos/react-native.config.js
```diff
@@ -18,5 +18,10 @@
 const path = require('path');
 
 module.exports = {
-  reactNativePath: path.dirname(require.resolve('react-native-macos/package.json')),
+  project: {
+    ios: {
+      project: 'src/FluentUITester.xcworkspace'
+    }
+  },
```
Saadnajmi, line 25: Why is this iOS over here?

tido64: It's a bug in `@react-native-community/cli`. I've opened an issue in https://github.com/react-native-community/cli/issues/1054.

Saadnajmi: So we still need this for the macOS test app for autolinking, even though we don't have an iOS project here?

tido64: Yes, I haven't looked much into it but I believe macOS autolinking reuses much of the iOS code.

PR #345, opened 2020-07-20
Replaced Button Test and Checkbox Test with Test Template

gulnazsayed:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [ ] android
- [x] web

### Description of changes

Replaced the Button Test and Checkbox Test with a Test template React Native component. Added status props for now, will figure out where to render it on the page later because I need to confirm the status verbiage. Also fixed the manifest.json and favicon errors.

### Verification

| image image | image image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/RNTester/TestComponents/Test.tsx
```diff
@@ -0,0 +1,61 @@
...
+export type TestSection = {
+  name: string;
+  testID: string;
+  component: React.FunctionComponent<{}>;
+};
+
+export interface TestProps {
+  name: string;
+  description: string;
+  winStatus: string; // status: production, beta, experimental
```
tkamireh, line 14: consider putting all the statuses in a single object. And instead of using a string type use a discrete string union if you know that there are only a few valid options. something like
``` 
type PlatformStatus = 'production' | 'beta' | 'experimental'
interface PlatformStatuses {
  winStatus: PlatformStatus
  // fill in the rest
}
```

PR #357, opened 2020-07-28
Switch to using Lage as a build sequencer

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android

### Description of changes

This switches the repo (mostly) to using Lage as a task running instead of Lerna. This is something being worked on by Ken and the Flywheel team and integrates some of the work the folks in Oslo have been doing on smart caching. This should:

- Speed up builds
- Add incremental build support, further speeding up builds.

Previously we were using just for sequencing the build command, this has been split apart to allow lage to sequence (and cache) linting, testing, building and bundling.  

**The top level commands to build have changed slightly.** They are now as follows:

| Old | New | Effect |
| --- | --- | --- |
| yarn build --min | yarn build | just do a sequenced build from the root |
| yarn build | yarn test | build, lint, and run any tests for the package |
| yarn bundle | yarn bundle | same effect though the new command is smarter, knowing how to execute without having run build ahead of time |
| yarn build & yarn bundle | yarn full-build | build, lint, test, bundle everything, this can overlap sequencing to speed things up |

See the Lage [github project](https://github.com/microsoft/lage) and [documentation site](https://microsoft.github.io/lage/) to learn more.

**Commands within a directory are mostly unchanged.** The exception is that explicit test commands and lint commands have been added where they were missing. To run tests in a directory simply run yarn test, to build yarn build, etc.

Root README has been updated

### Verification

This should be a build only change. The primary impact will be the automated tests and CI loops.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


package.json
```diff
@@ -12,26 +12,29 @@
...
+    "build-tools": "lage build-tools",
+    "buildci": "lage buildci",
+    "bump-versions": "beachball bump",
     "clean-all": "node ./scripts/clean-all.js",
-    "start": "cd apps && cd demo && npm start",
     "docs": "yarn workspace fluent-rn-website start",
-    "build": "lerna run build --stream -- --production --lint --color",
-    "bundle": "lerna run bundle --stream -- --production --lint --color",
+    "bundle": "lage bundle",
     "clean": "lerna run clean --stream",
```
dotnetjunky, line 22: Why is the command still using `lerna`?

README.md
```diff
@@ -92,7 +94,16 @@ Our FluentUI framework documentation is found in this repository alongside the i
 
 ### Yarn + Lerna
```
dotnetjunky, line 95: This should read "Yarn + Lage"

package.json
```diff
@@ -12,26 +12,29 @@
...
-    "bump-versions": "beachball bump",
-    "lint": "lerna run lint",
+    "test": "lage test",
     "verify-api": "lerna run verify-api",
     "update-api": "lerna run update-api"
   },
   "devDependencies": {
     "beachball": "^1.30.2",
+    "lage": "^0.16.1",
     "lerna": "^3.16.4",
```
dotnetjunky, line 38: Remove this?

package.json
JasonVMo: I haven't removed lerna, some of the magic build stuff intuits repo structure using the lerna config but this is a good catch. Honestly it only uses the config so I could actually replace all the commands.

JasonVMo: Fixed

lage.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  pipeline: {
+    ['build-tools']: ['^build-tools'],
```
tkamireh, line 3: Is the `['attribute']: [val]` syntax necessary - can it not be like the others?

JasonVMo: Because this is just a .js file, dashes are not allowed in the keys of objects unless they are declared in string form. They work fine, but have to be declared in this manner.

PR #372, opened 2020-08-04
Shimmer view component on iOS

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

We want to wrap the FluentUI Apple "Shimmer View" on iOS for use in React Native, by making a native module. 
This PR aims to do so. It is the first Native FluentUI Apple control we are wrapping, so it sets some trends (naming, etc) we may want to revisit in the future. 

### Verification

```
# from project root
yarn && yarn build
cd apps/ios
cd src; pod install; cd ..
yarn ios
```

Screen Shot 2020-09-08 at 5 36 52 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/src/Podfile
```diff
@@ -27,6 +27,10 @@ fi
 use_flipper!(false)
 use_test_app! do |target|
   target.app do
+
+    pod 'FURNShimmer', :path => '../../../packages/components/Shimmer/FURNShimmer.podspec'
+
```
Saadnajmi, line 32: @tido64 In the end, this was required, as the native files weren't imported without manually adding this

packages/components/Shimmer/FURNShimmer.podspec
```diff
@@ -0,0 +1,17 @@
+Pod::Spec.new do |s|
+  s.name             = 'FURNShimmer'
```
Saadnajmi, line 2: Naming is going to be a lot of fun here. I just chose this to get us started

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { requireNativeComponent } from 'react-native';
+
+export const FRNShimmer = requireNativeComponent('FRNShimmerView')
+
+export class Shimmer extends React.Component<{}> {
+
+  public render() {
+    return (
+      <FRNShimmer />
```
Saadnajmi, line 10: Another fun naming thing. 
<Shimmer /> is a wrapped <FRNShimmer /> ?

apps/ios/src/Podfile
tido64: Does adding `"@fluentui-react-native/shimmer": "*"` to `/apps/ios/package.json` help? I suspect what's happening is that rn-cli is unable to find the `.podspec` because the package isn't installed anywhere. If you look at `Podfile.lock`, all of the paths are under `node_modules`.

apps/fluent-tester/package.json
```diff
@@ -29,6 +29,7 @@
     "@fluentui-react-native/pressable": ">=0.3.76 <1.0.0",
     "@fluentui-react-native/focus-trap-zone": ">=0.4.0 <1.0.0",
     "@fluentui-react-native/link": ">=0.8.0 <1.0.0",
+    "@fluentui-react-native/shimmer": ">=0.1.0 <1.0.0",
     "@fluentui-react-native/radio-group": ">=0.3.76 <1.0.0",
     "@fluentui-react-native/persona-coin": ">=0.3.76 <1.0.0",
     "@fluentui-react-native/persona": ">=0.5.73 <1.0.0",
```
tido64, line 35: Does this actually pull code from npm? I was under the impression that we should be using `*` so that `npm`/`yarn` would get the packages from the monorepo instead.
```suggestion
    "@fluentui-react-native/pressable": "*",
    "@fluentui-react-native/focus-trap-zone": "*",
    "@fluentui-react-native/link": "*",
    "@fluentui-react-native/shimmer": "*",
    "@fluentui-react-native/radio-group": "*",
    "@fluentui-react-native/persona-coin": "*",
    "@fluentui-react-native/persona": "*",
```

apps/ios/src/Podfile.lock
```diff
@@ -375,6 +385,6 @@ SPEC CHECKSUMS:
   SwiftLint: 22ccbbe3b8008684be5955693bab135e0ed6a447
   Yoga: 3ebccbdd559724312790e7742142d062476b698e
 
-PODFILE CHECKSUM: 0d1a14cb0cdb7ed5c30de302b0bc9bec0c76c253
+PODFILE CHECKSUM: f15e85913f4e293f4d23673f91b7adf34b07cf8c
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.9.1
```
tido64, line 390: You should update your CocoaPods 😄

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { requireNativeComponent } from 'react-native';
+
+export const FRNShimmer = requireNativeComponent('FRNShimmerView')
+
+export class Shimmer extends React.Component<{}> {
```
tido64, line 6: Should this just be a function component? I don't see why it should be a class.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { requireNativeComponent } from 'react-native';
+
+export const FRNShimmer = requireNativeComponent('FRNShimmerView')
```
tido64, line 4: This shouldn't be exported.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: This is a WIP, and needs to be written out to follow our guide for building a new component (https://github.com/microsoft/fluentui-react-native/tree/master/packages/components#building-a-component)

Right now, this is very barebones though

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: Nice catch

apps/ios/src/Podfile
Saadnajmi: Changing just shimmer to * in the package.json for `apps/ios/` and `apps/fluent-tester` did not work.

To answer your other question, I don't think shimmer is published to the npm registry yet. I didn't do any work to explicitly do that. I'll have to check for the other components.

apps/fluent-tester/src/RNTester/TestComponents/Shimmer/ShimmerTest.tsx
```diff
@@ -0,0 +1,45 @@
...
+            labelCornerRadius: 10,
+            usesTextHeightForLabels: false,
+            labelHeight: 20
+          }}
+          shimmerAppearance={{
+            alpha: 0,
+            width: 180,
+            angle: 0,
+            speed: 400,
+            delay: 0.1
```
Saadnajmi, line 31: I should probably remove the props from the test. Or at least have one with and one without

packages/components/Shimmer/FURNShimmer.podspec
Saadnajmi: I still don't like this name. I wonder if podspecs can have names like "FluentUI-ReactNative/Shimmer"?

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+		let usesTextHeightForLabels = appearance["usesTextHeightForLabels"] as? NSNumber ?? NSNumber(booleanLiteral: oldAppearance.usesTextHeightForLabels)
+		let labelHeight = appearance["labelHeight"] as? NSNumber ?? NSNumber(floatLiteral: Double(oldAppearance.labelHeight))
+
+		shimmerView.appearance = ShimmerViewAppearance(
+			tintColor: tintColor,
+			cornerRadius: CGFloat(truncating: cornerRadius),
+			labelCornerRadius: CGFloat(truncating: labelCornerRadius),
+			usesTextHeightForLabels: Bool(truncating: usesTextHeightForLabels),
+			labelHeight: CGFloat(truncating: labelHeight)
+		)
```
Saadnajmi, line 122: ShimmerView has a struct with all the properties, rather than each property exposed individually. In the future, a possible refactor to the native view would be to separate all the properties, and deprecate the objects "ShimmerAppearance" and "ShimmerViewAppearance"

.gitignore
```diff
@@ -20,8 +20,7 @@ pids
 **/.DS_Store
 
 # CocoaPods
-/apps/macos/src/Pods/*
-/apps/ios/src/Pods/*
```
HeyImChris, line 24: Since this is generated, we usually want that to be ignored so people aren't spammed with files when they go to push changes

packages/components/Shimmer/ios/MSFShimmerViewManager.m
```diff
@@ -0,0 +1,9 @@
+#import <React/RCTViewManager.h>
+#import <React/RCTConvert.h>
+
+@interface RCT_EXTERN_MODULE(MSFShimmerViewManager, RCTViewManager)
+
+RCT_EXPORT_VIEW_PROPERTY(appearance, NSDictionary)
```
HeyImChris, line 6: Can you comment what this is and how it's used? Same for shimmerAppearance

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+		super.init(frame: frame)
+	}
+	
+	required init?(coder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+	
+// MARK: - UIView
+	
+    /// Shimmer, by default, only consider top-level views when shimmering. This is
```
HeyImChris, line 48: Nit- consider**s**

packages/components/Shimmer/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@fluentui-react-native/shimmer",
+  "version": "0.6.31",
```
HeyImChris, line 3: 0.0.1?

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
+import Foundation
+import FluentUI
+
+
+@objc(MSFShimmerContainerView)
+class ShimmerContainerView: UIView {
```
HeyImChris, line 6: I'm not sure I quite understand the reason we need this file to have all this logic. I was under the impression that all this logic should live in FluentUI-Apple and this class would simply extend that control and just pass the values through to that.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,105 @@
+/** @jsx withSlots */
+import {
+  compose,
+  UseSlots,
+  buildProps,
+  mergeProps,
+  withSlots
+} from '@fluentui-react-native/experimental-framework';
```
HeyImChris, line 8: Just a heads up, Morse is renaming this so you might hit merge issues with the name

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,105 @@
...
+  usesTextHeightForLabels: boolean;
+
+  /**
+   * If greater than 0, a fixed height to use for all UILabels. This will take precedence
+   * over `usesTextHeightForLabels`. Set to less than 0 to disable.
+   */
+  labelHeight: number;
+}
+
+export interface IShimmerAppearance {
```
HeyImChris, line 47: Does setting all of these work? I probably missed it but didn't see where we propagate these to the swift FUA control

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,105 @@
...
+} from '@fluentui-react-native/experimental-framework';
+import { requireNativeComponent } from 'react-native';
+import { IViewProps } from '@fluentui-react-native/adapters';
+import * as React from 'react';
+
+const shimmerName = "Shimmer"
+
+const NativeShimmerView = requireNativeComponent('MSFShimmerView');
+
+export interface IShimmerViewAppearance {
```
HeyImChris, line 17: Can you document the difference between `IShimmerViewAppearance` and `IShimmerAppearance`?

.gitignore
Saadnajmi: The change I'm making is to ignore the more generic `**/Pods/*`, which maps to any pods subdirectory , as opposed to the more specific ones only for the mac and iOS test apps

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: Yeah, they're documented inline at the prop level, but I can move the documentation here

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: updateShimmerAppearance() inside ShimmerContainerView.swift

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: The FUA Control only wants to Shimmer the top level subviews. I could never actually get it to shimmer with any test code I wrote. This code is lifted from fabric-react-native and converted to swift, comments and everything. 
It's inverting the logic to shimmer the leaf node / children subviews, which I guess works better with React Native's View heirarchy of nested children views.

packages/components/Shimmer/ios/MSFShimmerViewManager.m
Saadnajmi: So, duplicate the documentation from FUA / from Shimmer.tsx ? Since the same property is documented on both ends

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi, line 6: Of note to @anandrajeswaran / @markavitale / @harrieshin , this file is almost a straight convert of an objective-c file (from the Fabric-React-Native shimmer control), so there might be some swiftiness I missed that could make it cleaner

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,105 @@
...
+  mergeProps,
+  withSlots
+} from '@fluentui-react-native/experimental-framework';
+import { requireNativeComponent } from 'react-native';
+import { IViewProps } from '@fluentui-react-native/adapters';
+import * as React from 'react';
+
+const shimmerName = "Shimmer"
+
+const NativeShimmerView = requireNativeComponent('MSFShimmerView');
```
Saadnajmi, line 15: I like "MSF" as the prefix the most

apps/fluent-tester/src/RNTester/TestComponents/Shimmer/ShimmerTest.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <View>
+      <Text style={commonStyles.section} testID={SHIMMER_TESTPAGE}>
+        Shimmer Test Page
+      </Text>
+      <Separator />
+      <Stack style={stackStyle}>
+        <Shimmer
+          style={{ flex: 1, alignItems: 'center' }}
+          appearance={{
+            tintColor: processColor('#FF0000'),
```
harrieshin, line 20: what determine this color?

apps/ios/src/Podfile.lock
```diff
@@ -18,7 +18,11 @@ PODS:
     - boost-for-react-native
     - DoubleConversion
     - glog
+  - FURNShimmer (0.1.0):
+    - MicrosoftFluentUI (~> 0.1.0)
```
harrieshin, line 22: shouldn't we grab the latest fluentui version

packages/components/Shimmer/FURNShimmer.podspec
```diff
@@ -0,0 +1,17 @@
...
+  s.author           = { "Microsoft" => "fluentuinativeowners@microsoft.com"}
+  s.source       = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version = "5.0"
+  s.module_name = 'FURNShimmer'
+
+  s.ios.deployment_target = "11.0"
+  s.ios.source_files   = "ios/*.{swift,h,m}"
+
+  s.dependency 'React'
+  s.dependency 'MicrosoftFluentUI', '~> 0.1.0'
```
harrieshin, line 16: lets use the latest

packages/components/Shimmer/ios/ShimmerContainerView.swift
harrieshin: then extend the FU-A to make it easier for everyone?

packages/components/Shimmer/FURNShimmer.podspec
Saadnajmi: From https://guides.cocoapods.org/using/the-podfile.html

====
In addition to the logic operators CocoaPods has an optimistic operator ~>:

'~> 0.1.2' Version 0.1.2 and the versions up to 0.2, not including 0.2 and higher
'~> 0.1' Version 0.1 and the versions up to 1.0, not including 1.0 and higher
'~> 0' Version 0 and the versions up to 1.0, not including 1.0 and higher
=====

So this should be taking our latest, but will not take the 1.0.0 release whenever that happens, I believe

apps/fluent-tester/src/RNTester/TestComponents/Shimmer/ShimmerTest.tsx
Saadnajmi: Random color used to test that the tintColor prop did indeed set. I suppose it would be better if this came from some theme somewhere instead

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: I'm not sure it would provide value to FUA / it goes against how it was orginally implemented?? I'm still a little unclear on usage of Shimmer in production is, besides the demoes in our test app

packages/components/Shimmer/package.json
Saadnajmi: Right as I was reading this comment, I happened to have this stackoverflow open. It seems 0.1.0 might be the better start 

https://stackoverflow.com/questions/52236193/what-is-the-proper-first-version-0-1-0-or-0-0-1-or-anything-else

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
+import Foundation
```
harrieshin, line 1: nit: curious if we can remove

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+import FluentUI
+
+
+@objc(MSFShimmerContainerView)
+class ShimmerContainerView: UIView {
+
+	@objc var appearance: NSDictionary = NSDictionary() {
```
harrieshin, line 8: do we need @objc?

harrieshin, line 8: each property should have documentation

harrieshin: Nit: I don't know if appearance and shimmerAppearance variable names are clear. maybe it will help with more comments. but different updateshimmerViewappearnce and updateshimmerappearance needs more clarity.

Saadnajmi: Even if they are documented in the ShimmerView itself, and on the react native side? I just didn't document them in this containerView thing which to me is an implementation detail

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+	override public func layoutSubviews() {
+		super.layoutSubviews()
+
+		// Shimmer, by default, only considers top-level views when shimmering. This is
+		// incompatible with Flexbox, which wants you to wrap everything in more views to achieve
+		// a specific layout. The solution is to get a list of all the leaves of the subview tree
+		// and shimmer those by creating "mirror" views to mirror the RCTViews in a flat heirarchy in
+		// the shimmer container.
+		mirrors.forEach { $0.removeFromSuperview() }
+		mirrors = []
```
harrieshin, line 31: is this necessary?

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+		var leaves = [UIView]()
+		containerLeaves(view: self, leaves: &leaves)
```
harrieshin, line 34: this api is weird. you already have access to self in the function because you are your own view, is it just the matter of fact to update the leaves array?

harrieshin, line 34: nit; can we find a better name for this recursive function? searchLeaves(in view:UIView, output: inout [UIview]) ?

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+		for leaf in leaves {
+			guard let superview = leaf.superview, leaf != currentShimmerView else {
```
harrieshin, line 37: based on your other recursive function can this be ever true?

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: Yep!

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: Yes? We're getting rid of the old mirrors (from the array and view haeirarchy), and placing new ones

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: Probably not, but still good to have a guard to protect against future changes?

packages/components/Shimmer/ios/MSFShimmerViewManager.m
Saadnajmi: I'm just gonna add comments everywhere. Prefer overcomment to undercomment I guess

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: I'll prefer to overdocument over underdocument, and add more comments here

packages/components/Shimmer/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "src/index.ts",
+  "module": "src/index.ts",
+  "typings": "lib/index.d.ts",
+  "onPublish": {
+    "main": "lib-commonjs/index.js",
+    "module": "lib/index.js"
+  },
+  "scripts": {
+    "build": "fluentui-scripts build",
+    "just": "fluentui-scripts",
```
JasonVMo, line 14: I'm curious whether this should also have "lint" and "test" scripts, at the very least the linting one seems like a good idea. For jest you can look at how some of the other components are doing snapshot testing and emulate that.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+import * as React from 'react';
+
+const shimmerName = 'Shimmer';
+
+const NativeShimmerView = requireNativeComponent('MSFShimmerView');
+
+/**
+ * Appearance of the views that are shimmered
+ */
+export interface IShimmerViewAppearance {
```
JasonVMo, line 14: This isn't obvious but fluent-web has changed its coding guidance to drop the I prefix for types, we haven't converted everything over yet but we've tried to make new code use `ShimmerViewAppearance` rather than `IShimmerViewAppearance`

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Color to tint the shimmer boxes. Defaults to the Fabric default color.
+   */
+  tintColor: number;
```
JasonVMo, line 18: It's strange to have these all be required, typically it makes this hard to use as people will often not care about the other properties. If you can I would make these all optional and think of these as overrides for the default values.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+   */
+  speed: number;
+
+  /**
+   * Delay between the end of a shimmering animation and the beginning of the next one, in seconds
+   */
+  delay: number;
+}
+
+export interface ShimmerTokens {}
```
JasonVMo, line 67: Should these values actually be set in tokens? Do we want each shimmer to be individually customizable or do we expect people to mostly leave them as is?

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * ViewProps props, based off of the standard react-native ViewProps with some new extensions
+ */
+export type ShimmerProps<TBase = IViewProps> = TBase & {
+  /**
+   * Appearance of the views that are shimmered. If not passed, the default will be used.
+   */
+  appearance?: IShimmerViewAppearance;
```
JasonVMo, line 76: It's not atypical, though used occasionally to have sub-objects in props. They are easy to get wrong in usage, e.g.
```tsx
<Shimmer appearance={{angle: 3}} />
```
creates a new appearance object on every render pass causing react to think props have changed even when they haven't.

JasonVMo: Why not put the properties in the tokens directly?

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Appearance of the shimmer itself (the animation appearance)
+   */
+  shimmerAppearance?: IShimmerAppearance;
+};
+
+interface ShimmerType {
+  props: ShimmerProps;
+  slotProps: { root: ShimmerProps };
```
JasonVMo, line 86: Note it seems like the ShimmerProps interface is the internal interface for the native component, which may make sense. The externally facing interface can be one that is more standard for react/react-native.

JasonVMo: This means that ShimmerProps might be renamed as NativeShimmerProps (and set in slotProps), then have ShimmerProps be the outward facing interface to RN consumers. Then your component can transform them as appropriate.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: I expect people won't be customizing shimmer past the default.

These should be tokens, in my opinion. I was having difficulty with turning them into tokens, so I wanted to start with having them as props. 

Another thing about tokens: the native iOS view sets the defaults for each of these if no appearance / shimmerAppearance is specified. We *could* export each of those from Native -> React Native and set those when creating the control with compose. That was another thing I wasn't sure about, since I didn't see a pattern to follow.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: Thanks! I was confused about that as well.

packages/components/Shimmer/ios/ShimmerContainerView.swift
HeyImChris: I guess the thing here is that now the FUA control differs from the FURN control and I don't think we want that?

packages/components/Shimmer/ios/ShimmerContainerView.swift
```diff
@@ -0,0 +1,150 @@
+import FluentUI
+
+
+@objc(MSFShimmerContainerView)
+class ShimmerContainerView: UIView {
```
Saadnajmi, line 5: This file shoulg go away, as per #414 
@harrieshin @HeyImChris

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: Shimmer.customize works now. We now have a token :D

apps/fluent-tester/package.json
JasonVMo: It's more about the externally facing behavior, internally it just uses local versions. The >=0.1.0 <1.0.0 syntax emulates the behavior of ^1.x.y for packages that haven't been major released yet. The specific ranges aid the bundler in making the right decision for customers, these numbers get updated by beachball during publish. As for switching to * I'm less clear on what that would do, but I'm hesitant to make the change without more thinking on what will happen.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * ViewProps props, based off of the standard react-native ViewProps with some new extensions
+ */
+export type NativeShimmerViewProps<TBase = IViewProps> = TBase & {
+  /**
+   * Appearance of the views that are shimmered. If not passed, the default will be used.
+   */
+  appearance?: ShimmerViewAppearance;
```
JasonVMo, line 75: A few questions here for you:

- do we want these to be settable via props or do we want people to use customization to set them?
- should they be flat to match tokens?

Some points to understand what is happening as you have it written.

- If they set tokens via customize, or just use normally, the props passed to the native component will have two objects, appearance and shimmerAppearance. These will be created once and will have consistent object references.
- If react does a shallow compare to see if props have changed it will see they are the same object and not think it needs to do any remounts.
- If someone sets these as props, these objects will be not be merged and will fully replace what is in tokens. This is because mergeProps only does a deep merge on the style property.
- These will have the same pitfalls as setting styles, inline declarations will cause new objects to be created. Style has lint rules typically associated with it to catch them but these will be up to users to use correctly.

It would work better to do one of the following:

- remove these from props and just tell people to use customize to change them. This ensures they get good caching and shouldn't require any other adjustments.
- make the props flat, similar to what was done in tokens, then set tokenProps to true in the framework, this will patch your token values with any values set in props and the tokens will be present by the time your buildProps routine runs. You will then need to use the array parameter for buildProps to declare what values it should cache on.
- alternatively you can expose a subset in props that you expect people to modify more commonly. In this case you would follow the steps above except tokenProps would be set to just have that list of props that is both a prop and a token.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,138 @@
...
+      speed: 350,
+      delay: 0.4
+    }),
+    shimmerName,
+  ],
+  slotProps: {
+    root: buildProps<NativeShimmerViewProps, ShimmerToken>(
+      (tokens: ShimmerToken) => ({
+        appearance: {
+          tintColor: tokens.tintColor,
```
JasonVMo, line 115: Since you are using explicit set syntax to map to the sub-objects the names don't have to exactly match. E.g. you could rename tintColor to viewTintColor if you wanted. Just something to consider. It looks like you have good descriptions on the properties in the interface, you can just design what a flat interface would look like without being constrained to the native interface exactly.

packages/components/Shimmer/src/Shimmer.tsx
JasonVMo: One other point with the current behavior, if someone only wants to change one property, like corner radius, they still have to set all of the values. This is harder to use and requires them to look up the right constants to copy. The downside here is that if we change those their code won't update to match.

Saadnajmi: As of right now, I expect all of the members of appearance and shimmerAppearance to only be set once at creation of the shimmer (making it more suited as tokens than props). I have those two as not flat to match the API on the native API: the iOS (and eventually android and macOS) controlc each have two structs called appearance and shimmerAppearance that 1:1 map with what I exposed in JS.

I thought that by putting appearance & shimmerAppearance in NativeShimmerViewProps, clients would not be able to manually set those as props, and would only be able to use the token. By your 1st suggestion, do you mean to remove the NativeShimmerViewProps from line 132?

The 2nd suggestion is probably the change we want in the native API long term anyway. The non flat list of properties is weird IMO.

For having to set all the properties, would that be fixed by adding "?"'s to each token in lines 98-107?

Saadnajmi:
To give an update here, I made a lot of native changes in FluentUI Apple to reduce complexity here. 
The pending PR is : https://github.com/microsoft/fluentui-apple/pull/219

Once that goes in, I'll update this PR. That will include @JasonVMo 's suggestion of having a flat list of properties, and @HeyImChris 's suggestion of not having a ShimmerContainerView.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: I made changes to the native side that should address your comments by taking your 2nd suggestion. The Native ShimmerView now has a flat list of properties it takes in, and the React-Native tokenprops match 1:1 with them. We don't need to explicitly set all of them to stylize the shimmerview, and the default / stylized shimmer views work as expected.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: I ended up just naming tintColor viewTintColor, and renaming a lot of the native properties with what I feel are better names. It's a flat interface now anyway, so this should be resolved

packages/components/Shimmer/ios/ShimmerContainerView.swift
Saadnajmi: Used that name in FUA

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Corner radius on each view.
+   */
+  cornerRadius?: number;
+};
+
+/**
+ * Shimmer Token is a flat list of all the properties we can set on the native view
+ */
+export type ShimmerToken = NativeShimmerViewProps;
```
JasonVMo, line 55: I would reverse these and declare ShimmerTokens above, then have the props include the base reference. The way you have it set up here the tokens include all ViewProps as well.

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,98 @@
...
+      shimmerAlpha: 0.4,
+      shimmerWidth: 180,
+      shimmerAngle: -(Math.PI / 45.0),
+      shimmerSpeed: 350,
+      shimmerDelay: 0.4,
+      viewTintColor: 'rgb(241,241,241)',
+      cornerRadius: 4,
+    },
+    shimmerName,
+  ],
```
JasonVMo, line 76: You should add either `tokenProps: true` which will just copy all props into tokens, or `tokenProps: ['shimmerAlpha', 'shimmerWidth', // and so on. The second approach will copy just those properties into tokens, at which point you could likely change your buildProps to just spread the tokens.  Easiest way would be to do something like the following:
```ts
const tokenProps: (keyof ShimmerTokens)[] = [/* list all the tokens here */];

// in the body of the component then add
tokenProps,

// then buildProps becomes:
  root: buildProps(tokens => ({...tokens}), tokenProps)
```

JasonVMo (changes requested):
Generally looks good, just a couple of quick fixes.

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: I'm gonna keep this example around somewhere when I inevitably have to write a guide for converting more FluentUI-Apple controls..

packages/components/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,89 @@
...
+  tokenProps: true,
```
JasonVMo, line 78: You can actually take this line out or change it to be false. It will work either way, this just adds a little bit of extra overhead as it will do some unnecessary prop filtering.

JasonVMo (approved):
One quick boolean update that would be good to do but I'm signing off on the change. No need to re-review if you update it. Thanks for all the work you put into this one!

packages/components/Shimmer/src/Shimmer.tsx
Saadnajmi: Oh, thanks! I didn't realize I left that in there, and thanks for the sign off at this time!

PR #380, opened 2020-08-06
Remove dist from excluded files in tester-win32 and add script to download dependencies for tester

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [x] windows
- [ ] android

### Description of changes

Removes the dist folder from being excluded under @fluentui-react-native/tester-win32 to ensure the bundle gets published. This also adds a script that installs dependencies for the FURN tester (still a work in progress).

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/win32/.gitignore
```diff
@@ -67,3 +67,7 @@ buck-out/
 
 # CocoaPods
 /ios/Pods/
+
+
+# Build targets
+!dist
```
FalseLobster, line 73: I think we want this in the .npmignore not the .gitignore.

If we put it here, we'll end up checking in the bundle, when we just want to make sure it's included as part of package-publish.

ksiler: Ah, true. Thanks for catching that!

ksiler: I copied the contents from .gitignore over to .npmignore because I want to be sure it doesn't publish something we don't want but I'm unsure if I need to include anything from the .gitignore at the root of the repo.

ksiler:
I just updated it to remove the files that don't exist. Are there any files that should be excluded as part of the package that I didn't exclude?

apps/win32/.npmignore
```diff
@@ -0,0 +1,4 @@
+yarn-error.log
+
+# Build targets
+!dist
```
dotnetjunky, line 4: Do you want to ignore the `src` files? You can try running `npm pack --dry-run` to see what files would be included in the npm files. Then add the files that you don't want to pack to this file.

ksiler: I'm not sure if I do. Is there a standard here or is it just preference? When I try to ignore `src` it doesn't ignore the `src` folder in the root of the package but instead the `src` folders under `lib` and `lib-commonjs`

Issue #228, opened 2020-05-22
FocusTrapZone and Callout - Attempted import error: 'requireNativeComponent' is not exported from 'react-native'


frmpiccolo:
![fluentui-react-native_issue_ev01](https://user-images.githubusercontent.com/12400108/82711924-267f7400-9c5d-11ea-8534-101d6c8f6165.png)
![fluentui-react-native_issue_ev02](https://user-images.githubusercontent.com/12400108/82711932-28493780-9c5d-11ea-99ac-8c18c30a29d9.png)

I´m trying to include fluentui/react-native into brand new react-native project but it´s not working. 
Message: "Attempted import error: 'requireNativeComponent' is not exported from 'react-native'" (from @fluentui-react-native/focus-trap-zone/lib/FocusTrapZone.js and @fluentui-react-native/callout/lib/Callout.js). 

Environment information: 
- Windows 10 running WSL (Ubuntu 18.04) 
- Node v12.16.2
- react ~16.9.0
- react-dom ~16.9.0
- react-native ~0.61.5 
- react-native-web ~0.11.7
- expo ~37.0.3
- @fluentui/react-native ^0.15.33

Steps to reproduce: 
1. create-react-native-app (Default new app)
2. cd <appname_folder>
3. npm i @fluentui/react-native 
4. Edit App.js
5. Add "import { Checkbox } from '@fluentui/react-native';" into App.js and save it
6. yarn web

JasonVMo:
Good catch and good issue here. Overall we are still rounding out the various platforms and win32 is getting the most attention. That being said we did recently start standing up react-native-web and the recent PR should hopefully address this (along with some other scenarios).

Also we should start paying attention to github issues more in the future, currently more of our tracking is internal but that is changing over time.

frmpiccolo:
> Good catch and good issue here. Overall we are still rounding out the various platforms and win32 is getting the most attention. That being said we did recently start standing up react-native-web and the recent PR should hopefully address this (along with some other scenarios).
> 
> Also we should start paying attention to github issues more in the future, currently more of our tracking is internal but that is changing over time.

Hi Jason, 
Thanks! 
We'll try again after this PR. 
Good job over there and be all safe!

JasonVMo:
I'm going to reopen this as I may be withdrawing or reworking the PR and don't want to lose sight of this. The current PR both delay loads and creates a common cache to address versioning issues.

[...]

PR #389, opened 2020-08-10
Add deferred loading and caching wrapper for native components

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android

### Description of changes

There are several parts of this change

#### `@fluentui-react-native/native-component` package

This package adds a cached wrapper around `requireNativeComponent`. This addresses two potential gotchas with our previous usage.

**Unnecessary loads**

The naive way to write a component would look something like:

```tsx
const MyNativeComponent = requireNativeComponent('MyNativeComponent');

export const MyComponent = props => {
  const propsInner = doSomeStuff(props);
  return <MyNativeComponent {...propsInner} />;
};
```

The issue here is that the call to `requireNativeComponent` happens when the module is loaded, not when the component is used. This can result in unnecessary overhead as well as causing issues if the native code isn't present. This should also address Issue #228 

**Multiple instances**

For larger projects where different components may be imported directly, it can be easy to end up with slightly different versions of the code present. While this is a performance issue, and can cause subtle behavioral issues if globals are used, in the case of native components this can cause multiple calls to register the same native component. This will cause the view manager to be registered twice resulting in runtime issues.

The package then adds the following:

- `queryNativeComponent` wrapper which provides a safe alternative to `requireNativeComponent`.
- `NativeComponent` and `lateBindComponent` utilities which make it a bit easier to consume the `queryNativeComponent` API

#### Consume the new module

In particular `Callout` and `FocusTrapZone` have switched over to the new paradigm and the frameworks (both old and experimental) support things declared with `NativeComponent`.

#### Move `memo-cache` and `immutable-merge` to 1.0.0

These packages rarely change and feel pretty final in terms of API. Because there is value in keeping the value of the new package consistent it made sense to start it at 1.x and because it depends on memo-cache it made sense to bump that to 1.0 as well.

As part of this I'm renaming `@uifabricshared/immutable-merge` to be `@fluentui-react-native/immutable-merge` as we should be moving away from @uifabricshared as a scope.

### Verification

Automated tests plus manual run through of FocusTrapZone and Callout on Win32 to ensure they still work correctly.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/framework/native-component/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@fluentui-react-native/native-component",
+  "version": "1.0.0",
+  "description": "Layered memoization style cache helper",
```
dotnetjunky, line 4: Is this description accurate?

JasonVMo: Good catch.

Saadnajmi:
I imagine I should wait for this to get merged before creating a new Native component for FURN (iOS specifically)?

packages/components/Callout/src/Callout.tsx
```diff
@@ -34,7 +33,7 @@ export const Callout = compose<ICalloutType>({
   },
   settings: settings,
   slots: {
-    root: RCTCallout
+    root: NativeComponent('RCTCallout')
```
FalseLobster, line 36: It's probably worth checking that the changes you're making here by wrapping the native components with a functional component don't interfere with the setAndForwardRef stuff @PPatBoyd had to do to get useViewCommandFocus working.

JasonVMo: In truth they get updated to directly use the host via the `lateBindComponent` call that happens internally. The component wrappers are only point in time wrappers.

JasonVMo:
> I imagine I should wait for this to get merged before creating a new Native component for FURN (iOS specifically)?

Actually I'm probably going to withdraw this after discussing it a bit more with Coates.

JasonVMo:
I'm going to do this a different way, instead looking into a general deferred loading approach for components.

PR #390, opened 2020-08-11
Added FluentUITester theme selector and fixed layout

gulnazsayed:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [ ] android
- [x] web

### Description of changes

Added three pickers to toggle between platform, app, and theme to change the look of the whole tester. Theme change only works for app selection for now. Fixed some of the layout for the web tester so that it looks better when you toggle between test pages. 

### Verification

| image image | image image image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/RNTester/FabricTester.tsx
```diff
@@ -25,6 +25,83 @@ export interface IFabricTesterProps {
...
+          <Picker
+            selectedValue={selectedPlatform}
+            style={fabricTesterStyles.dropdown}
+            onValueChange={(platformValue) => setSelectedPlatform(platformValue)}
+          >
+            <Picker.Item label="Win32" value="win32" />
+            <Picker.Item label="UWP" value="uwp" />
+            <Picker.Item label="iOS" value="ios" />
+            <Picker.Item label="macOS" value="mac" />
+            <Picker.Item label="Android" value="android" />
```
Saadnajmi, line 69: Is this changing the available tests to what is there for each platform? How would that work if a platform has native components?

JasonVMo: I believe this will allow us to import styling to make the components look like another platform, particularly for web where it will be ideal to preview ios/android/windows/win32/macos.

gulnazsayed: ^ What Jason said! Also, in terms of theme change, so far I was able to get header colors to change when you toggle the App category. Eventually, we will want to be able to toggle styling between the platform, app, and theme so I set up the dropdown mechanism for that but all of the options don't have styling attached to it yet.

apps/fluent-tester/src/RNTester/TestComponents/Test.tsx
```diff
@@ -68,9 +69,23 @@ const styles = StyleSheet.create({
 });
 
 export const Test = (props: TestProps) => {
+
+  const themeColor = () => {
```
tkamireh, line 73: So as part of the TODO/next steps we had mentioned filing as in issue I would two things related to this
1. We should convert the tester to use FluentUI Text and work through the bugs we had been seeing
2. Those Text components and anything else shouldn't need to have a theme color layered on style, in theory they'd respond to the theme changes as long as the entire tester is wrapped in a themeprovider

apps/fluent-tester/src/RNTester/FabricTester.tsx
```diff
@@ -25,6 +25,86 @@ export interface IFabricTesterProps {
...
+  return (
+    <View style={fabricTesterStyles.header}>
+      <Text style={[fabricTesterStyles.testHeader, themeColor()]} testID={BASE_TESTPAGE}>
+        ⚛ FluentUI Tests
+      </Text>
+
+      <View style={fabricTesterStyles.pickerRoot}>
+        <View style={fabricTesterStyles.picker}>
+          <Text style={fabricTesterStyles.pickerLabel}>Platform:  </Text>
+          <Picker
```
tkamireh, line 63: Platform emulated themes would be another one of those things to add to the TODO/Next steps issue :)

PR #391, opened 2020-08-11
Add missing dependencies to package.json files

statm:
There are phantom dependencies that are used but not declared in package.json. For example, `@uifabricshared/foundation-settings` is used in [`Button`](https://github.com/microsoft/fluentui-react-native/blob/master/packages/components/Button/src/Button.tsx#L11) but not showing up in its [package.json](https://github.com/microsoft/fluentui-react-native/blob/master/packages/components/Button/package.json).

A run of:
`lerna exec --stream --no-bail -- depcheck ./ --ignores="eslint" --ignore-patterns="*eslint*,*.d.ts,*.test.*"`
under repo root reveals a few dozens of those phantom dependencies. This PR fixes some of them.

### DIFF
--- a/packages/experimental/use-slots/package.json
+++ b/packages/experimental/use-slots/package.json
@@ -34,8 +34,7 @@
     "@fluentui-react-native/memo-cache": ">=0.2.10 <1.0.0",
     "@fluentui-react-native/use-styling": ">=0.2.1 <1.0.0",
     "@uifabricshared/foundation-settings": ">=0.6.1 <1.0.0",
-    "@uifabricshared/immutable-merge": ">=0.5.67 <1.0.0",
-    "react": "16.11.0"
+    "@uifabricshared/immutable-merge": ">=0.5.67 <1.0.0"
   },
   "devDependencies": {
     "@types/jest": "^19.2.2",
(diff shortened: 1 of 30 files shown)


tkamireh:
nice catch 😀! Perhaps that depcheck should be a part of our workflow/ci?

packages/experimental/use-slots/package.json
```diff
@@ -33,7 +33,9 @@
   "dependencies": {
     "@fluentui-react-native/memo-cache": ">=0.2.10 <1.0.0",
     "@fluentui-react-native/use-styling": ">=0.2.1 <1.0.0",
-    "@uifabricshared/immutable-merge": ">=0.5.67 <1.0.0"
+    "@uifabricshared/foundation-settings": ">=0.6.1 <1.0.0",
+    "@uifabricshared/immutable-merge": ">=0.5.67 <1.0.0",
+    "react": "16.11.0"
```
acoates-ms, line 38: This should be a peerDep.  And should match the other peerDep range for react (since this dep will break people using RN 0.60)  (Same for other additions of react as a dep)

statm: Yes. Moved these into peerDeps and set to `^16.8.0`.

statm:
> nice catch 😀! Perhaps that depcheck should be a part of our workflow/ci?

Yes, but there are some 'missing' deps where I'm not very sure that they should be added or not. I'll circle back in a few days and get it started.

PR #395, opened 2020-08-13
Set "WhiteColors" as palette id for host palette if no id is provided

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [x] android

### Description of changes

Previously, if no themeID was provided when creating a theme registry (i.e. `...createPlatformThemeRegistry();` ) then no host palette colors would be loaded for a theme. This change sets "WhiteColors" as the default themeID when it's not provided for the host palette.

### Verification

Manually tested using the theme test page.

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
```diff
@@ -52,7 +52,7 @@ export function translateOfficeTheme(module: IOfficeThemingModule, cache: Palett
     },
     typography: module.fluentTypography,
     host: {
-      palette: cache[id]
+      palette: cache[id || 'WhiteColors']
```
FalseLobster, line 55: I'd probably create a helper function out of line 36.  i.e. `const getPaletteCacheKey = (id?: string) => { ... }`

It doesn't hurt to include the debug palette and not hard code the 'WhiteColors' default in a second place.

PR #396, opened 2020-08-13
Improving Vertical Alignment of Checkbox's Text

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32
- [X] windows
- [X] android

### Description of changes

Before, the contents of the Root (Checkbox+Text) were vertically aligned to the top of the container (flex-start). They should have been vertically aligned to the center.

However, just doing that is not enough. The height of the checkbox and text are not always the same (also, the text has an invisible 1px border for the focusRect on load, making the FlexBox even bigger), and because of this, I had to align the Text to the bottom of the container AND create a small margin on top. The FlexBox of text is nearly the height of the
 container itself, making the marginTop necessary.

This change has little effect on the default fontSize, but becomes more prominent when the fontSize shrinks.

### Verification

I've tested every fontSize from 6 to 14.

Before:
image

After:

image

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Checkbox/src/Checkbox.settings.ts
```diff
@@ -46,9 +45,10 @@ export const settings: IComposeSettings<ICheckboxType> = [
     content: {
       variant: 'bodyStandard',
       style: {
-        marginTop: 1,
         borderStyle: 'dotted',
-        borderWidth: 1
+        borderWidth: 1,
+        alignSelf: 'baseline',
+        marginTop: 3
```
NikoAri, line 51: >3 [](start = 19, length = 1)

Just curious why is this needed?

ksiler: +1, does `alignSelf: 'baseline'` have any effect? I tried testing it out but I don't see a difference between auto (default) and baseline.

samuelfreiberg: You may be right, in my initial testing I thought it had an effect, but testing it again now, it doesn't look like it. I will remove it

PR #400, opened 2020-08-17
Rename @fluentui-react-native/experimental-framework to @fluentui-react-native/framework

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

All changes are impacting experimental packages, so all platforms, but none for production scenarios.

### Description of changes

Very soon we want to switch over to using the experimental framework. In preparation for that it is getting a rename to be called @fluentui-react-native/framework. Note that there was a package in the repo with this name but it contained the old framework and was unpublished and unused, it only existed for purposes of running the bundle analyzer.

So this change:

- deletes the files under packages/libraries/framework
- repaths packages/experimental/experimental-framework to packages/experimental/framework
- renames @fluentui-react-native/experimental-framework to @fluentui-react-native/framework
- updates references to that framework

It's just a step in the direction of changing over.

### Verification

Automated tests

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Text/src/Text.tsx
```diff
@@ -73,11 +73,11 @@ export const Text = compose<TextType>({
...
           color: tokens.color,
-          ...fontStyles.from(tokens, theme)
-        }
+          ...fontStyles.from(tokens, theme),
+        },
       }),
-      ['color', ...fontStyles.keys]
-    )
+      ['color', ...fontStyles.keys],
+    ),
```
HeyImChris, line 80: Did you run prettier? I see a lot of extraneous commas throughout

JasonVMo: We were getting inconsistency on this setting, after a quick poll in the teams channel we decided to make trailing commas default and I recently updated our general settings to use that. As a result we'll see a bunch of changes like this as people edit files.  Long term though it should sort itself out.

I'm tempted to do a full pass but it will generate a lot of noise for people with changes in progress.

PR #406, opened 2020-08-20
Support shouldFocusOnMount and shouldFocusOnContainer props

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32
- [X] windows
- [ ] android

### Description of changes

- Add a container slot to ContextualMenu to support initial focus on container
- shouldFocusOnMount maps to Callout slot's setInitialFocus, with default value true and initial focus on first menu item
- shouldFocusOnContainer overrides initial focus from first menu item to container

### Verification
shouldfocusOnMount and shouldFocusOnContainer toggle switches were added on RNTester to observe initial focus behavior on button click that will launch the menu.

Focus events were tested with Narrator to check that appropriate container or first menu item was focused on mount, and narrator only recognized the container when shouldFocusOnContainer was set. 

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/packages/components/ContextualMenu/src/ContextualMenu.tsx
+++ b/packages/components/ContextualMenu/src/ContextualMenu.tsx
@@ -1,5 +1,6 @@
 /** @jsx withSlots */
 import * as React from 'react';
+import { View } from 'react-native';
 import {
   contextualMenuName,
   ContextualMenuProps,
@@ -30,7 +31,12 @@ export const CMContext = React.createContext<ContextualMenuContext>({
 export const ContextualMenu = compose<ContextualMenuType>({
   displayName: contextualMenuName,
   usePrepareProps: (userProps: ContextualMenuProps, useStyling: IUseComposeStyling<ContextualMenuType>) => {
-    const { setShowMenu, ...rest } = userProps;
+    const {
+      setShowMenu,
+      shouldFocusOnMount = true,
+      shouldFocusOnContainer = false,
+      ...rest
+    } = userProps;
 
     // This hook updates the Selected Button and calls the customer's onClick function. This gets called after a button is pressed.
     const data = useSelectedKey(null, userProps.onItemClick);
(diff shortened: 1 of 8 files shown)


packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -52,24 +58,32 @@ export const ContextualMenu = compose<ContextualMenuType>({
 
     const slotProps = mergeSettings<ContextualMenuSlotProps>(styleProps, {
       root: {
-        ...rest
+        ...rest,
+        setInitialFocus: shouldFocusOnMount
+      },
+      container: {
+        accessible: shouldFocusOnContainer,
+        ... { acceptsKeyboardFocus: shouldFocusOnContainer },
```
ksiler, line 66: What effect does `... { acceptsKeyboardFocus: shouldFocusOnContainer }` have over `acceptsKeyboardFocus: shouldFocusOnContainer` ?

packages/components/ContextualMenu/src/ContextualMenuItem.settings.ts
```diff
@@ -11,6 +11,7 @@ export const settings: IComposeSettings<ContextualMenuItemType> = [
     root: {
       accessible: true,
       accessibilityRole: 'menuitem',
+      ...{ acceptsKeyboardFocus: true },
```
ksiler, line 14: I have the same question here. Trying to better my understanding 😊

packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -30,7 +31,12 @@ export const CMContext = React.createContext<ContextualMenuContext>({
 export const ContextualMenu = compose<ContextualMenuType>({
   displayName: contextualMenuName,
   usePrepareProps: (userProps: ContextualMenuProps, useStyling: IUseComposeStyling<ContextualMenuType>) => {
-    const { setShowMenu, ...rest } = userProps;
+    const {
+      setShowMenu,
+      shouldFocusOnMount = true,
```
PPatBoyd, line 36: Neat syntax, if I understand correctly the point is "use shouldFocusOnMount if provided in userProps, otherwise default to shouldFocusOnMount=True" ?

packages/components/ContextualMenu/src/ContextualMenu.tsx
PPatBoyd: IIRC the spread operator lets us "sneak" the property through past type checking against ViewProps, if ViewProps doesn't have the property.  @lenahong maybe a comment for what issue we're addressing by adding it?

packages/components/ContextualMenu/src/ContextualMenuItem.settings.ts
```diff
@@ -11,6 +11,7 @@ export const settings: IComposeSettings<ContextualMenuItemType> = [
     root: {
```
PPatBoyd, line 11: should there be an empty default style for the container, like content/icon (in expanded diff)?

PPatBoyd: nm I misread, this is ContextualMenuItem.settings :)

packages/components/ContextualMenu/src/ContextualMenu.tsx
lenahong: I guess type checking does not occur for this code bit, because I tried changing to just `acceptsKeyboardFocus: shouldFocusOnContainer` and it seems to build and work fine.

packages/components/ContextualMenu/src/ContextualMenuItem.settings.ts
lenahong: Type checking does occur here and I'm trying to "sneak" the prop through

packages/components/ContextualMenu/src/ContextualMenu.tsx
ksiler: I see, thanks for the explanation!

PR #407, opened 2020-08-20
Check for private packages

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

This adds a task to scripts that will check for dependency tree violations that might break publishing.

- In particular we have broken partners by having published packages inadvertently depend on private packages
- This will now run as the depcheck for the scripts package (scanning everything)
- As such it is integrated into our CI loops and will block any PRs that have this error

Note that lage will attempt to cache this incorrectly locally, if you are trying to resolve an error of this sort you will likely need to run `yarn depcheck --no-cache` to force it to run.

### Verification

- I tested this locally by running through builds and the depcheck command. 
- I then changed various packages to be marked as private and made sure the check caught them

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/scripts/src/tasks/checkPublishingTask.ts
+++ b/scripts/src/tasks/checkPublishingTask.ts
@@ -1,22 +1,46 @@
-import { TaskFunction } from 'just-task';
+import { TaskFunction, logger } from 'just-task';
+import { getPackageInfos, findGitRoot } from 'workspace-tools';
+
+export type DependencyType = 'dependencies' | 'devDependencies' | 'peerDependencies';
 
 export interface CheckPublishingOptions {
   /**
-   * By default this will only check dependencies, to check devDependencies as well set this to true
+   * An array of fields to check for private internal dependencies, by default this is just dependencies
    */
-  devDependencies?: boolean;
-
-  /**
-   * Opt in to checking peer dependencies as well
-   */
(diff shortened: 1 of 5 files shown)


scripts/src/tasks/checkPublishingTask.ts
```diff
@@ -0,0 +1,46 @@
...
+    logger.info('Starting scan for publishing errors');
+    try {
+      Object.keys(packageInfos).forEach(pkg => {
+        if (!packageInfos[pkg].private) {
+          logger.verbose(`Scanning published package ${pkg} for private dependenies`);
+          dependencyTypes.forEach(dependencyType => {
+            const deps = packageInfos[pkg][dependencyType];
+            Object.keys(deps || {}).forEach(dep => {
+              if (packageInfos[dep] && packageInfos[dep].private) {
+                const errorMsg = `${pkg} has a ${dependencyType} on private package ${dep}`;
```
FalseLobster, line 32: Would this return a false positive for `"private": "false"` ?

JasonVMo: I don't believe so since that would still be falsy when parsed.

PR #408, opened 2020-08-20
Removed apps/mobile folder

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

Under the apps folder, we used to have the `android` and `ios` test app located as a symbol package.
We have since moved the iOS test app out to `apps/ios` and use `react-native-test-app` to handle it's creation / native code.

The android test app under `apps/mobile/android` never compiled to begin with (at least on my machine), so it's not very useful to keep around. I''m currently working with the `react-native-test-app` team to see if we can get a compiling android test app, which would allow win32/UWP engineers to have a mobile platform to test off of (since iOS is unable to build on windows)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
I'm going to close this because CI keeps failing on eslint for files I never touched, despite `yarn buildci` passing locally. 
May revisit when we have a working android test app

PR #418, opened 2020-08-30
Add iconStrokeWidth and iconStrokeColor to PersonaCoin

tom-un:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android

### Description of changes

PersonCoin renders the presence icon with no border.  This change adds an `iconStrokeWidth` and `iconStrokeColor` token allowing a stroke to be rendered around the presence icon.

Additionally, the `.vscode/settings.json` file had `editor.formatOnSave": true` which caused Visual Studio Code to automatically format files on Save, but the formatting specified in the settings does not match the apparent coding styles of this repo.

### Verification

The change was tested using the iOS test app.

| before | after |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-  "editor.formatOnSave": true,
   "editor.detectIndentation": false,
   "editor.trimAutoWhitespace": true,
   "editor.insertSpaces": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "editor.formatOnSave": true,
   "editor.detectIndentation": false,
   "editor.trimAutoWhitespace": true,
   "editor.insertSpaces": true,
--- a/apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
+++ b/apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
@@ -14,7 +14,7 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
   const [physicalSize, setPhysicalSize] = React.useState<number>(80);
   const [iconSize, setIconSize] = React.useState<number>(24);
   const [iconStrokeWidth, setIconStrokeWidth] = React.useState<number>(2);
-  const [iconStrokeColor, setIconStrokeColor] = React.useState<string>('white');
+  const [iconStrokeColor, setIconStrokeColor] = React.useState<string>(undefined);
   const [initialsSize, setInitialsSize] = React.useState<number>(14);
   const [horizontalAlignment, setHorizontalAlignment] = React.useState<IconAlignment>();
   const [verticalAlignment, setVerticalAlignment] = React.useState<IconAlignment>();
--- a/apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
+++ b/apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
@@ -13,6 +13,8 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
   const [textColor, setTextColor] = React.useState<string>();
   const [physicalSize, setPhysicalSize] = React.useState<number>(80);
   const [iconSize, setIconSize] = React.useState<number>(24);
+  const [iconStrokeWidth, setIconStrokeWidth] = React.useState<number>(2);
+  const [iconStrokeColor, setIconStrokeColor] = React.useState<string>('white');
   const [initialsSize, setInitialsSize] = React.useState<number>(14);
   const [horizontalAlignment, setHorizontalAlignment] = React.useState<IconAlignment>();
   const [verticalAlignment, setVerticalAlignment] = React.useState<IconAlignment>();
@@ -37,6 +39,12 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
   }
   if (iconSize) {
     tokens.iconSize = iconSize;
+  }
+  if (iconStrokeWidth) {
+    tokens.iconStrokeWidth = iconStrokeWidth;
+  }
+  if (iconStrokeColor) {
+    tokens.iconStrokeColor = iconStrokeColor;
   }
   if (initialsSize) {
     tokens.initialsSize = initialsSize;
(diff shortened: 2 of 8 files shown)


packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
```diff
@@ -7,14 +7,21 @@ import { calculateEffectiveSizes } from './PersonaCoin.helpers';
 const _iconKeyProps: (keyof IPersonaCoinTokens)[] = ['iconSize', 'size', 'coinSize'];
 
 function _buildIconStyles(tokenProps: IPersonaCoinTokens /*, theme: ITheme */): ImageProps {
-  const { iconSize } = calculateEffectiveSizes(tokenProps);
+  const { iconSize, iconStrokeWidth } = calculateEffectiveSizes(tokenProps);
+  const iconSizeAdjusted = iconSize + iconStrokeWidth * 2;
+  const iconStrokeColor = tokenProps.iconStrokeColor || 'white';
```
dotnetjunky, line 12: The `'white` should come from theme colors.

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
```diff
@@ -7,14 +7,21 @@ import { calculateEffectiveSizes } from './PersonaCoin.helpers';
...
 
   return {
     source: {},
     style: {
       position: 'absolute',
-      width: iconSize,
-      height: iconSize
+      width: iconSizeAdjusted,
+      height: iconSizeAdjusted,
+      bottom: -iconStrokeWidth,
```
dotnetjunky, line 20: does negative values of bottom work in win32?

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
tom-un: I changed it to `theme.colors.background`

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
tom-un: I don't know, but if it didn't I think it would be a win32 bug.  It works on all other platforms including web.

dotnetjunky: Could you test it on win32 and log a bug if it doesn't work? we don't want this bug to go unnoticed.

.vscode/settings.json
```diff
@@ -1,5 +1,4 @@
 {
-  "editor.formatOnSave": true,
```
dotnetjunky, line 2: We do want to have this flag set. Otherwise, people will get surprise CI build failures.

apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
```diff
@@ -74,6 +82,15 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
           }}
         />
 
+        <TextInput
+          style={[commonStyles.textBox, textBoxBorderStyle]}
+          placeholder="Icon stroke color"
+          blurOnSubmit={true}
+          onSubmitEditing={(e) => {
+            setIconStrokeColor(e.nativeEvent.text);
```
dotnetjunky, line 90: if the text is empty, you may want to set it as `undefined` to verify that setting `undefined` stroke color is working.

Saadnajmi:
@tom-un I imagine we'll have to implement the same property for FUA when we wrap this control?

.vscode/settings.json
tom-un: How is this property used during CI?   With it on, hitting save in vs code causes the whole file to get reformatted in a way that is inconsistent with all the other files in the repo.

dotnetjunky: We run a `lint` rule as part of CI. If someone make changes but don't format it according to the lint rules, the build fails. 

I suspect the reason that VS formatting doesn't match the other files is that we turned on this `formatOnSave` only recently and many files were already checked in before that.

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
tom-un: I do not have access to a Windows machine.

.vscode/settings.json
tom-un: Interesting.  The `lint` succeeds with the current formatting or the settings.json formatting.   I reverted the change

apps/fluent-tester/src/RNTester/TestComponents/PersonaCoin/CustomizeUsage.tsx
tom-un: I changed the default of the start var to `undefined`.

tom-un:
> @tom-un I imagine we'll have to implement the same property for FUA when we wrap this control?

@Saadnajmi : I think that when FUA and the other Fluent native's implement the Persona Coin, they should make the stroke be part of a mask on the avatar image so that it shows through whatever is behind the control.

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
Saadnajmi: I can try checking this out on my win32 machine later today

Saadnajmi: Seems to be working fine?

Screen Shot 2020-08-31 at 3 26 02 PM

Saadnajmi: Side note, I couldn't drag the sliders in the "Customize Usage" part. not sure if that's an existing bug

.vscode/settings.json
JasonVMo: For a while we had a few settings undefined which was causing different editors to save in different settings. I believe we've now defined those values such that everyone should be using the same settings. This means that as people edit the files in the repo they will be converted to be consistent one by one.

Generally the formatter doesn't run as part of a CI unless it gets run as part of something else. If that happens it could break CI in that it will prompt that there are unsubmitted changes. (This would be like running yarn causing the lock file to be updated, this would trigger the same check)

packages/components/PersonaCoin/src/PersonaCoin.tokens.icon.ts
dotnetjunky: Yes, I think it was broken a while ago. :(

PR #420, opened 2020-08-31
Update macOS test app to use 0.62, bump react-native-test-app version across apple platforms

### LINKED ISSUE #401: Resizing FluentTester on macOS looks broken
### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [ ] iOS
    - **Xcode version**:
  - [x] macOS
    - **Xcode version**: 11.4.1
  - [ ] win32
  - [ ] windows
  - [ ] android

Resizing the macOS test app causes a chain of resizing that does not at all feel natural and renders the test app unusable for a few seconds

macos_fluent_tester

#### Actual behavior:

Resizing works smoothly

#### Expected behavior:

Resizing is weird as in the attached video

Requested priority: Normal

Products/applications affected: (if applicable)

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

React Native macOS 0.62 is merged, along with an updated react-native-test-app to support it.
Let's bump the macOS and iOS test app versions to use the new RN version.

While here, let's update the macOS podfile (which was missing some custom scripts) and remove the .xcworkspace files from the repo, which are generated anyway.

This also happens to fix #401 

### Verification

Built the iOS and macOS test apps.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


yarn.lock
```diff
@@ -9062,21 +9062,11 @@ header-case@^1.0.0:
     no-case "^2.2.0"
     upper-case "^1.1.3"
 
-hermes-engine@^0.2.1:
-  version "0.2.1"
-  resolved "https://registry.yarnpkg.com/hermes-engine/-/hermes-engine-0.2.1.tgz#25c0f1ff852512a92cb5c5cc47cf967e1e722ea2"
-  integrity sha512-eNHUQHuadDMJARpaqvlCZoK/Nitpj6oywq3vQ3wCwEsww5morX34mW5PmKWQTO7aU0ck0hgulxR+EVDlXygGxQ==
-
```
Saadnajmi, line 9069: Do we want to be using hermes for macOS? The test app supports that with the latest version I believe

HeyImChris: No leave it disabled. There's a lot of APEX pushback against Hermes right now. Until we have clear data on how it's better than JavascriptCore, we shouldn't use it. Even then we might be better off just asking Apple to fix whatever deficiencies JSC has instead of using Hermes

harrieshin: we don't want to increase app size right now

HeyImChris:
Can you verify the generated files you're deleting are in the .gitignore so it doesn't flood devs' work environments with modified files when they build?

yarn.lock
Saadnajmi: OK, There was no change to enable hermes, so we should still not be using it

Saadnajmi:
> Can you verify the generated files you're deleting are in the .gitignore so it doesn't flood devs' work environments with modified files when they build?

Good catch. That's probably how the generated files made it in to begin with

PR #421, opened 2020-08-31
Removed apps/mobile folder, update eslint dependencies

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [x] android

### Description of changes

Under the apps folder, we used to have the android and ios test app located as a symbol package.
We have since moved the iOS test app out to apps/ios and use react-native-test-app to handle it's creation / native code.
The android test app will be moved to react-native-test-app separately.

Let's get rid of the `apps/mobile` directory entirely. Doing so changes some eslint comfigs somehow, so I needed to update eslint dependencies in a few package.json files. Furthermore, I had to remove the eslint rule to do prop-types validation, which we weren't using anyway. I also removed the react/prop-types dependency since we don't use it.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
So this one again fails checks, but passes `yarn buildci` locally, and I'm not sure what's the difference?

Saadnajmi:
Finally figured out what the proper eslint changes I needed to make were to get this PR working. Updated the PR description with that info. If you look at the yarn lock, you can see many older vesrions of eslint/it's dependencies go away

JasonVMo (approved):

PR #442, opened 2020-09-09
RNIcon

warren-ms:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Description of changes
Along with an change in netui this adds RN Icon

### Verification
There is a test page added.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/package.json
```diff
@@ -20,9 +20,10 @@
     "bundle": "fluentui-scripts bundle"
   },
   "dependencies": {
-    "@fluentui/react-native": ">=0.15.89 <1.0.0",
+    "@fluentui/react-native": ">=0.15.90 <1.0.0",
     "@fluentui-react-native/stack": ">=0.2.8 <1.0.0",
     "@fluentui-react-native/interactive-hooks": ">=0.5.1 <1.0.0",
+    "@fluentui-react-native/text": ">=0.6.8 <1.0.0",
```
JasonVMo, line 26: Lots of extra changes showing up in this review, is the merge baseline messed up in some way?  Also I would expect text to be included in @fluentui/react-native so it is unnecessary to add it as a dependency.

apps/fluent-tester/react-native.config.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
```
JasonVMo, line 1: Are we sure this file is needed here? The fluent-tester project isn't an actual app, instead it contains the test code used by the various apps.

apps/fluent-tester/src/RNTester/TestComponents/Icon/IconTest.tsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text } from '@fluentui-react-native/text';
```
JasonVMo, line 4: This can be imported from '@fluentui/react-native' instead.

apps/win32/package.json
```diff
@@ -27,11 +27,12 @@
...
   },
   "dependencies": {
-    "@fluentui-react-native/tester": "^0.4.1",
+    "@fluentui-react-native/tester": "^0.5.1",
     "@uifabricshared/theming-react-native": "0.7.76",
     "react": "16.11.0",
     "react-native": "0.62.2",
-    "react-native-svg": "^11.0.0"
+    "react-native-svg": "^11.0.0",
+    "react-native-windows": "^0.62.1"
```
JasonVMo, line 35: Why the rnw dependency? This project uses @office-iss/react-native-win32

apps/win32/react-native.config.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
```
JasonVMo, line 1: I didn't think win32 used the CLI bundling so I didn't think it needed a react-native.config.js file. (I could very well be wrong, since bundling is half magic anyways)

I would expect reactNativePath to be set to reference react-native-win32 however in this project.

packages/components/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react';
+import {IIconProps, ISvgIconProps, IFontIconProps} from './Icon.types';
+import * as ReactNative from 'react-native';
+import { Text } from '@fluentui-react-native/text';
+import {SvgUri} from 'react-native-svg'
+import * as assetRegistry from 'react-native/Libraries/Image/AssetRegistry';
+import { stagedComponent, useTheme, mergeProps, getMemoCache } from '@fluentui-react-native/framework';
+
+function RenderRasterImage(iconProps: IIconProps) {
```
JasonVMo, line 9: Functions should be camelCased

packages/components/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { width, height } = iconProps;
+  return (
+    <ReactNative.Image
+      source = {iconProps.rasterImageSource.src}
+      style = {{width: width, height: height}}
+    />
+  )
+}
+
+function RenderFontIcon(iconProps: IIconProps) {
```
JasonVMo, line 19: same, should be renderFontIcon

packages/components/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,94 @@
...
+    style.fontFamily = `${style.fontFamily}#${asset.httpServerLocation}/${asset.name}.${asset.type}`;
+  }
+
+  const char = String.fromCharCode(fontSource.codepoint);
+  return(
+    <Text style={style}>{char}</Text>
+  );
+}
+
+function RenderSvg(iconProps: IIconProps) {
```
JasonVMo, line 34: should be renderSvg

packages/components/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,94 @@
...
+    return null;
+  }
+}
+
+const memoCache = getMemoCache<IIconProps>();
+
+export const Icon = stagedComponent((props: IIconProps) => {
+
+  return (rest: IIconProps) => {
+    const theme = useTheme();
```
JasonVMo, line 70: useTheme is a hook, so this line should be moved outside of the closure.

packages/components/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const color = props.color || theme.colors.buttonText ;
+
+    const style = memoCache({
+      color: color
+    }, [theme, color])[0];
```
JasonVMo, line 75: If color is the only input, theme can be removed. Also should be a function, so:
`const style = memoCache(() => ({ color }), [color])`

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
+import { SvgProps } from 'react-native-svg';
+export const iconName = 'Icon';
+
+export interface IFontIconProps
```
JasonVMo, line 4: Typically props inherit from one of the base primitives (ViewProps, TextProps or ImageProps) then have their own properties added.

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
...
+{
+  uri?: string;
+  src?: React.FC<SvgProps>;
+  viewBox?: string;
+  color?: string;
+}
+
+export interface IRasterImageIconProps
+{
+  src: object;
```
JasonVMo, line 23: Should this be typed to match ImageProps (src: ImageProps['src']? Or just be ImageProps?

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export interface IIconProps
+{
+  ariaLabel?: string;
```
JasonVMo, line 28: Typically the base inheritance, likely from ImageProps will pick up the a11y props.

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
...
+  fontSource?: IFontIconProps;
```
JasonVMo, line 29: These interfaces aren't really suitable for an icon control, it is too hard to use. Usually one would expect some sort of name -> values lookup to happen as part of the icon. If we want this interface I might call this object `IconBase` or something that denotes it is a non-opinionated component.

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
...
+  svgSource?: ISvgIconProps;
+  rasterImageSource?: IRasterImageIconProps;
+  width?: number;
```
JasonVMo, line 32: I would expect width/height values to be set by some enum, similar to how persona coin does it. I would then expect those values to be transferred into the right values for svg/raster/font as appropriate.

packages/components/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,35 @@
...
+  height?: number;
+  color?: string;
```
JasonVMo, line 34: Same with color, I would expect that if someone sets this value it will be moved to the right field in the sub-objects. I wouldn't expect users to have to set those sub-objects.

JasonVMo (changes requested):

packages/components/Icon/src/Icon.types.ts
warren-ms: FluentUI Icon and ReactNative Image both allow direct specification of the width and height via numbers though. Shouldn't that be the expectation?

packages/components/Icon/src/Icon.types.ts
warren-ms: Oh yes color shouldn't be in IIconProps but only in IFontIconProps and ISvgIconProps since only those are recolorable and RasterIcons are not.

apps/fluent-tester/package.json
warren-ms: I had a bunch of extra changes from dependbot for some reason. I've rebased to where master should be now. That should remove a bunch of the files changed.

apps/fluent-tester/src/RNTester/TestComponents/Icon/IconTest.tsx
warren-ms: Ok importing from there instead.

apps/win32/package.json
warren-ms: Without this line one of the CI checks fail:

ERR! ERROR DETECTED IN @fluentui-react-native/tester-win32 depcheck
ERR! started
ERR! hash: 08d7269c5dfeb123def1002650ff1e81b1ccf88d, cache hit? false
ERR! Running /tmp/yarn--1599700209945-0.3433538225809185/yarn run depcheck
ERR! warning package.json: No license field
ERR! $ fluentui-scripts depcheck
ERR! [1:20:42 AM] ■ started 'depcheck'
ERR! [1:20:42 AM] ▲ Unused devDependencies
ERR! [1:20:42 AM] ▲ -- @react-native-community/eslint-config
ERR! [1:20:42 AM] ▲ -- @types/jasmine
ERR! [1:20:42 AM] ▲ -- metro-react-native-babel-preset
ERR! [1:20:42 AM] ▲ -- tsconfig-paths
ERR! [1:20:42 AM] ▲ -- allure-commandline
ERR! [1:20:42 AM] ▲ -- appium
ERR! [1:20:42 AM] ▲ -- webdriverio
ERR! [1:20:42 AM] ▲ -- @wdio/allure-reporter
ERR! [1:20:42 AM] ▲ -- @wdio/appium-service
ERR! [1:20:42 AM] ▲ -- @wdio/cli
ERR! [1:20:42 AM] ▲ -- @wdio/jasmine-framework
ERR! [1:20:42 AM] ▲ -- @wdio/local-runner
ERR! [1:20:42 AM] ▲ -- @wdio/spec-reporter
ERR! [1:20:42 AM] ▲ -- @wdio/sync
ERR! [1:20:42 AM] x Missing dependency on react-native-windows

packages/components/Icon/src/Icon.tsx
warren-ms: Done

packages/components/Icon/src/Icon.tsx
warren-ms: Done

packages/components/Icon/src/Icon.tsx
warren-ms: Done

packages/components/Icon/src/Icon.tsx
warren-ms: Done

packages/components/Icon/src/Icon.tsx
warren-ms: Done

packages/components/Icon/src/Icon.types.ts
warren-ms: > Usually one would expect some sort of name -> values lookup to happen as part of the icon.

Can you elaborate on what you mean by this? Icons referenced by name is for the second mile stone. This milestone is about getting font icon drawing mechanism.

apps/fluent-tester/react-native.config.js
warren-ms: I think I had a asset bundling problem that I attributed to this but it seems to work without this file now so removing it.

apps/win32/package.json
JasonVMo: The line below in the log is actually the relevant one. It's coming from the usage in react-native.config.js having a reference to react-native-windows. The unused devDependencies are just warnings.

JasonVMo: If that file goes away or references react-native-win32 instead the depcheck error will go away. You can also run yarn depcheck at the root to test it before CI, though I would recommend running yarn depcheck --no-cache just to be sure.

warren-ms: Changed the refence to react-native-win32 and it looks ok now.

packages/components/Icon/src/Icon.types.ts
warren-ms: When we talked about this in the call yesterday you mentioned putting properties of all three types into a single object so a single object can be passed through slots. But in this model, couldnt the single object be an IIconProps object?

packages/experimental/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface ISvgIconProps
+{
+  uri?: string;
+  src?: React.FC<SvgProps>;
+  viewBox?: string;
+}
+
+export interface IRasterImageIconProps
+{
+  src: object;
```
JasonVMo, line 22: This can be `src: ImageProps['source']`

packages/experimental/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export interface IIconProps extends AccessibilityProps
+{
+  fontSource?: IFontIconProps;
+  svgSource?: ISvgIconProps;
+  rasterImageSource?: IRasterImageIconProps;
+  width?: number;
```
JasonVMo, line 30: Can we add a more standard sizing pattern, maybe something like what is available for PersonaCoin, or some sort of set of 'small' | 'medium' | 'large' etc? Possibly move the PersonaCoin mapping to a new shared location. Specific width/height values could be set via style but likely shouldn't be the primary way to set this.

packages/experimental/Icon/src/Icon.types.ts
```diff
@@ -0,0 +1,33 @@
...
+  src?: React.FC<SvgProps>;
+  viewBox?: string;
+}
+
+export interface IRasterImageIconProps
+{
+  src: object;
+}
+
+export interface IIconProps extends AccessibilityProps
```
JasonVMo, line 25: This interface should have a style prop as well. You should be able to do something like:

```ts
style?: StyleProp<ImageStyle | TextStyle>
```

then the style should be passed along to the text or image component as appropriate.

JasonVMo (changes requested):
With the move of this to experimental, I think we can iterate on the interface changes post-checkin. I would make the functional changes and make sure it is not in fluentui/react-native until the interface goes through more revisions.

packages/components/Icon/src/Icon.types.ts
JasonVMo: Makes sense if we have prior art here.

packages/experimental/Icon/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "build": "fluentui-scripts build",
+    "just": "fluentui-scripts",
+    "clean": "fluentui-scripts clean",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "verify-api": "fluentui-scripts verify-api-extractor",
+    "update-api": "fluentui-scripts update-api-extractor"
+  },
+  "dependencies": {
+    "@fluentui-react-native/framework": "0.2.10",
+    "@uifabricshared/foundation-compose": "^1.6.73",
```
JasonVMo, line 22: I think foundation-compose can be removed

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+
+export const Icon = stagedComponent((props: IconProps) => {
+  const theme = useTheme();
+
+  return (rest: IconProps) => {
+    const color = props.color || theme.colors.buttonText;
+
+    const style = {
+      color: color
```
JasonVMo, line 89: This should be renamed to baseProps or something rather than style since style has a particular meaning.

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,109 @@
...
+const rasterImageStyleCache = getMemoCache<ReactNative.ImageStyle>();
+
+function renderRasterImage(iconProps: IconProps) {
+  const { width, height } = iconProps;
+  const style = rasterImageStyleCache({
+    width: width,
+    height:height
+  }, [width, height])[0];
+
+  return (
```
JasonVMo, line 18: You probably want to forward the style from icon props here, which means you'll have to merge it with your style. I believe the merge-props package has an exported mergeStyles (which is what mergeProps uses internally), so:

```ts
const style = mergeStyles(props.style, rasterImageStyleCache({ width, height }, [width, height])[0]);
```

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,109 @@
...
+      fontFamily: fontSource.fontFamily,
+      codepoint: fontSource.codepoint,
+      fontSize: fontSource.fontSize,
+      color: iconProps.color
+  }, [iconProps.color, fontSource.fontSrcFile, fontSource.fontFamily, fontSource.codepoint, fontSource.codepoint])[0];
+
+  if (style.fontSrcFile != undefined) {
+    const asset = assetRegistry.getAssetByID(style.fontSrcFile);
+    style.fontFamily = `${style.fontFamily}#${asset.httpServerLocation}/${asset.name}.${asset.type}`;
+  }
```
JasonVMo, line 42: Same comment about style merging and forwarding it to the internal text entry

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -0,0 +1,109 @@
...
+  )
+}
+
+const fontStyleMemoCache = getMemoCache();
+
+function renderFontIcon(iconProps: IconProps) {
+  const fontSource: FontIconProps = iconProps.fontSource;
+
+  const style = fontStyleMemoCache({
+      fontSrcFile: fontSource.fontSrcFile,
```
JasonVMo, line 32: You can change this to a function, then bring the if style.fontSrcFile stuff inside the function, that way you only update the fontFamily the first time you calculate the style.

JasonVMo (changes requested):

JasonVMo (approved):

PPatBoyd:
still needed / needs update?

PR #453, opened 2020-09-10
Update react-native-win32 version, enable logbox

acoates-ms:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Description of changes

Bumping the version of react-native-win32 used.

apps/win32/metro.config.js:
Updated the metro config to remove some unneeded code around reactNativePlatformResolver, but then added a work around where metro wont load assets from hoisted packages.  (This allows metro to load the logbox assets used by the new error/warning ui).

apps/win32/pacakge.json:
Modified the launch command of rex-win32 to use fastrefresh.  Now many changes will hot module reload instead of requiring an instance reload.

apps/win32/src/index.tsx:
Added code to enable LogBox.  This is off by default in 0.62 (it will be the default in 0.63).  But it should be ready for use, and is better than anything else we have in win32.

packages/utils/adapters/src/adapters.win32.ts:
Added support for a couple of new properties that were added to react-native-win32 for more complete accessibility support.

### Verification

Manually booted win32 tester app, and tested that fast refresh and logbox seem to work ok.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/win32/metro.config.js
```diff
@@ -9,42 +9,46 @@ const blacklist = require('metro-config/src/defaults/blacklist');
...
+    watchFolders: getWatchFolders(),
+    resolver: {
+      assetExts: assetExts.filter((ext) => ext !== 'svg'),
+      sourceExts: [...sourceExts, 'svg'],
+      blacklistRE: blacklist([
+        // This stops "react-native run-windows" from causing the metro server to crash if its already running
+        new RegExp(`${path.resolve(__dirname, 'windows').replace(/[/\\]/g, '/')}.*`),
+      ]),
+    },
+    // Metro doesn't currently handle assets coming from hoisted packages within a monorepo.  This is the current workaround people use
```
tkamireh, line 26: Just making sure, we don't know what release (if any) is going to "fix" this issue, correct?

tkamireh (approved):

tkamireh:
@samuelfreiberg hey this PR is failing on win32 e2e tests, I've reran it once to check whether it was flaky. Looks like from the pngs test pages just aren't rendering. Any sense from the outputs whether these are legitimate failures. The metro config changed slightly, that's one thing that seems like it could be suspicious, the other could be failures rendering the log box? @acoates-ms ideas?

acoates-ms:
Any idea what I have to do locally to run the e2etests?  I wasn't able to run them locally.

apps/win32/metro.config.js
acoates-ms: There is no current fix in progress for this.  Basically metro/the assets get generated using a url that contains ../.., which isn't really valid as part of a url.

There was some discussion to change the url to have the path to asset as a param on the url instead of part of the url itself.  But I didn't see any progress on that happening.

tkamireh:
> Any idea what I have to do locally to run the e2etests? I wasn't able to run them locally.

https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/E2E#debugging-e2e-failures-locally hopefully the instructions in there are up to date

acoates-ms:
Hmm.. they seem to pass locally.

samuelfreiberg:
Taking a look now, if it's a false negative, re-running just the Win32 PR should (hopefully fix it). I'll keep an eye on this and report back in a bit

FalseLobster:
> Looks like from the pngs test pages just aren't rendering.

So I also re-ran the tests in the vain hope that they might just work, but I might have gotten slightly different results from the .pngs based on what Tariq's saying.

Button-Test-Page

This sounds a lot like AB#4357944 which I suspected was due to compositor changes.  Hopefully it's not because our CI runs headless, that would be a harder problem to solve, but if it reproes in Office automation I suspect that isn't the case, right?

apps/win32/metro.config.js
JasonVMo: Interesting. Is the location we need consistent with regards to the RN path? Because then we could transform the URLs from relative to absolute paths where we dynamically look up the RN-win32 path. Not blocking, just a thought of whether that would be more robust.

JasonVMo (approved):

apps/win32/metro.config.js
acoates-ms: The URL to metro has to be relative, since its in the form:
`http://localhost:8081/asset/<url>`

The discussion in the community was that it should be changed to:
`http://localhost:8081/asset?url=<url>`
Then the ..'s would still be a valid url.

PR #462, opened 2020-09-18
Spinner (Activity Indicator) component on iOS

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Description of changes

There was a request from @tom-un to port the FluentUI Apple ActivityIndicator control to FluentUI React Native. This satisfies that request.

Some background:
A "spinner" on iOS is normally a UIActivityIndicator. In FluentUI Apple, we have a stylized one that we call ActivityIndicator (Swift) / MSFActivityIndicator (obj-c).

React-Native has a base control "ActivityIndicator" already for iOS and Android that wraps UIActivityIndicator on iOS. This control wraps MSFActivityIndicator from FluentUI Apple, but has the same API surface (for the most part) as the base React-Native ActivityIndicator control

I chose to name the component "Spinner" because that seemed the more cross-platform Fluent name (it's what we call the control on web and our FluentUI documentation website).

### Verification

Screen Shot 2020-09-16 at 11 04 26 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Spinner/ios/SpinnerViewManager.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import FluentUI
+
+@objc(MSFSpinnerViewManager)
+class SpinnerViewManager: RCTViewManager {
+	override func view()->UIView! {
+		let activityIndicatorView = ActivityIndicatorView(size: .small)
+		activityIndicatorView.startAnimating()
```
harrieshin, line 8: Is it by design the animation starts when it is created? because in native, the client decides when to start and stop animation.

packages/components/Spinner/ios/MSFSpinnerViewManager.m
```diff
@@ -0,0 +1,42 @@
...
+@end
+
+@interface RCT_EXTERN_MODULE(MSFSpinnerViewManager, RCTViewManager)
+
+RCT_EXPORT_VIEW_PROPERTY(hidesWhenStopped, BOOL);
+
+RCT_EXPORT_VIEW_PROPERTY(color, UIColor);
+RCT_EXPORT_VIEW_PROPERTY(size, MSFActivityIndicatorViewSize);
+
+RCT_CUSTOM_VIEW_PROPERTY(animating, BOOL, MSFActivityIndicatorView)
```
harrieshin, line 30: it seems weird that changing the value is going to start or stop the animation. or am I interpreting wrong?

Saadnajmi: It does. I copied this implementation from fabric-react-native. I think that was done to match the API of the base react-native activity-indicator https://reactnative.dev/docs/activityindicator

packages/components/Spinner/ios/SpinnerViewManager.swift
Saadnajmi: Also copied from fabric-react-native. Though they actually wrapped the startAnimating method in RN, and call it in "componentDidMount, which I think is like viewDidLoad

Saadnajmi: Can remove and just let client set the prop themselves

packages/components/Spinner/src/Spinner.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { compose, UseSlots, buildProps, mergeProps, withSlots } from '@fluentui-react-native/framework';
+import { NativeModules, ViewProps } from 'react-native';
+import * as React from 'react';
+import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
+
+const spinnerName = 'MSFSpinnerView';
+
+const NativeSpinnerView = ensureNativeComponent(spinnerName);
+
+export type Size = 'xSmall' | 'small' | 'medium' | 'large' | 'xLarge';
```
JasonVMo, line 11: It feels like we want to standardize size enums. PersonaCoin has size8, size24, etc. I'm not sure if that or this is the right schema but it's worth starting a discussion in the teams channel about how to make all of these consistent. It seems like the icon component should have the same sizing logic as well.

packages/components/Spinner/src/Spinner.tsx
```diff
@@ -0,0 +1,72 @@
...
+  ],
+  slotProps: {
+    root: buildProps((tokens) => ({ ...tokens }), []),
+  },
+  slots: { root: NativeSpinnerView },
+  render: (props: NativeSpinnerViewProps, useSlots: UseSlots<SpinnerType>) => {
+    const size = ExportedNativeConstants.sizes[props.size || 'small'];
+    const Root = useSlots(props).root;
+    return (rest: NativeSpinnerViewProps, children: React.ReactNode) => (
+      <Root {...mergeProps(props, rest)} style={[props.style, { height: size, width: size }]}>
```
JasonVMo, line 67: This is creating a new style object on every render which will then cause the produced style to be new every time.

Use the memo-cache package to get around this. Because the only variable is size, you only need to cache on the basis of size.  So the code would look something like this:
```tsx
// outside of the spinner component
import { getMemoCache } from '@fluentui-react-native/framework';
const cache = getMemoCache();

// inside render
const sizeProps = cache(() => ({ style: { height: size, width: size }}, [size]);

// root render then becomes:
<Root {...mergeProps(sizeProps, props, rest)>
```

Alternatively you could add the size value to the tokens then do the following:

```ts
// add size to tokens above
// add this inside compose:
tokensThatAreAlsoProps: ['size']
// then add it to the root buildProps call
slotProps: {
  root: buildProps(
    (tokens) => ({ color: tokens.color, style: { height: tokens.size, width: tokens.size } }), 
    ['color', 'size']
  )
```

This second approach does the same thing but allows customization of the default value for size.

JasonVMo (changes requested):

packages/components/Spinner/src/Spinner.tsx
harrieshin: we've purposely don't use any numbers in the enum because designers seems to change the ramp size frequently enough that it doesn't withstand very long. similar to how apple UX enum also don't have numbers because it could change os to os.

harrieshin:
my novice question about the test app. do we have to include stylized version of avatar? if Designers saw that that is kind of not the recommended toolkit design ... it really should be the gray style that consumers should use.

packages/components/Shimmer/FluentUIReactNativeShimmer.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.author           = { "Microsoft" => "fluentuinativeowners@microsoft.com"}
   s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
   s.swift_version    = "5"
-  s.module_name      = 'FluentUIReactNative'
+  s.module_name      = 'FluentUIReactNativeShimmer'
```
HeyImChris, line 14: Probably makes sense for this to be in a separate PR as this is for the Shimmer control, not the Spinner control

packages/components/Spinner/ios/MSFSpinnerViewManager.m
```diff
@@ -0,0 +1,42 @@
...
+	@"xLarge": @(MSFActivityIndicatorViewSizeXLarge),
+
+}), MSFActivityIndicatorViewSizeSmall, integerValue);
+
+@end
+
+@interface RCT_EXTERN_MODULE(MSFSpinnerViewManager, RCTViewManager)
+
+RCT_EXPORT_VIEW_PROPERTY(hidesWhenStopped, BOOL);
+
```
HeyImChris, line 26: Nit- Can probably delete the extraneous line in between the hidesWhenStopped export and the color expo

packages/components/Spinner/src/Spinner.tsx
Saadnajmi: A size enum sounds great! I don't know if it should be the same for the icon component, but that sounds reasonable.

Fwiw, FluentUI Apple, FluentUI Android, and FluentUI Web  use the same sizes (except apple has the extra size xLarge), so I prefer the strings. Optionally a nice additional API would be if we could specify the size to be any number in points.

https://developer.microsoft.com/en-us/fluentui#/controls/ios/spinner
https://developer.microsoft.com/en-us/fluentui#/controls/android/spinner
https://developer.microsoft.com/en-us/fluentui#/controls/web/spinner

packages/components/Spinner/src/Spinner.tsx
Saadnajmi: Size doesn't feel like it's worthy of being a token for me, as I feel different parts of an app will want to size their Spinner differently. Would it make sense to have tokens differ between different parts of an app, AKA the Spinner in this stack inherits token 1, while the spinner in this other stack inherits token 2?


I now remember when that memo cache was moved to 1.0.0, I'll look into using that. Thanks for the explanation :D

Saadnajmi:
> my novice question about the test app. do we have to include stylized version of avatar? if Designers saw that that is kind of not the recommended toolkit design ... it really should be the gray style that consumers should use.

Right... it might be better for me to use a different shade of grey. The only token I defined for this "Spinner" was color, and I made some of the ones in the test app not animate either.

packages/components/Shimmer/FluentUIReactNativeShimmer.podspec
Saadnajmi: Good point. I didn't realize two things can't have the same module name hence this change. I can make a separate Shimmer PR to change some things and merge that in first

apps/ios/src/IOSTests.tsx
```diff
@@ -48,6 +48,13 @@ import {
   ShimmerTest,
 } from '../../fluent-tester/src/RNTester/TestComponents/Shimmer';
 import {
+<<<<<<< HEAD
+  HOMEPAGE_SPINNER_BUTTON,
+  SpinnerTest,
+} from '../../fluent-tester/src/RNTester/TestComponents/Spinner';
+import {
+=======
+>>>>>>> 901b3a534df10a1d471c815349651500d0237575
```
Saadnajmi, line 57: Oof, I'll need to fix that ._.

packages/components/Spinner/src/Spinner.tsx
Saadnajmi: I went with your second approach! I looked at documentation for the web shimmer that said "Don't use more than one spinner in a window" which then told me this lends itself to be a token.

packages/components/Spinner/src/Spinner.tsx
Saadnajmi: I'm following this up inside our Teams Channel (and also the Fluent Community Team). I'll keep this change, and filed #475 to follow up

packages/components/Spinner/src/Spinner.tsx
Saadnajmi: er, now it's correct. Sorry about that.

Saadnajmi:
Another thought, until we have the Android Spinner from FluentUI-Android wrapped, how would we feel about having this control default to https://reactnative.dev/docs/activityindicator for all other platforms? That would be great for testing a different implementation with the same API, and @tom-un mentioned that fabric-react-native does this exactly for its Activityindicator component. Thoughts @JasonVMo @harrieshin ?

Saadnajmi:
Since this is my oldest review, and I kinda want to close it, I have a few thoughts about how to move it forward:

1. Keep it iOS specific, as some other components have been. 
    - This does not fulfill the ask we had for a cross platform component, and is not much better than React Native core's `ActivityIndicator`
1. Add an Android implementation, and try my hand at creating an Android Native Module
    - This would at least allow us to have a usable mobile component, which may be enough for a 1st iteration. Given that `react-native-macos` and `react-native-windows` both probably implement `ActivityIndicator`, this is still arguably less useful. 
1. Keep the native iOS implementation, and have the component fallback to React Native Activity Indicator for other platforms
    - What's the point of having a FluentUI component if we'll just fallback to the core component anyway
1. Convert this to a JS implementation (perhaps lifting code from `FluentUI React`), and have that be the default implementation, _maybe_ allowing for native implementations for a specific platform
    - This is my favorite choice, because I would like to learn how to write a pure JS component, especially for something as simple as a spinner. The animations will be the fun part to figure out.

Thoughts?

Saadnajmi:
Closing because this is old and I'd like to revisit with a different implementation later (perhaps a JS implementation?)

PR #464, opened 2020-09-18
Add FluentTester readme for ios

ejlayne:
### Platforms Impacted
- [x ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding instructions for how to launch the FluentTester on iOS, following Saad's instructions after recent updates. Will update the main Fluent Tester readme to point to this page.

### Verification

Ran through the documentation as I was writing it.

### Pull request checklist

- [ ] Automated Tests
- [ x] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -1,0 +1,26 @@
+`FluentUI Tester` is the test app that we use to test our FluentUI components during development.
+
+## Launch `FluentUI Tester` app on iOS
+
+Prereq: FluentUI Tester on iOS can only run on a Mac.
+
+1. Make sure you have followed the Getting Started instructions [here](../../README.md) to install packages and build the entire Fluent UI React Native repository. E.g. from the root of the repo:
+```
+    yarn && yarn build
+```
+
+2. Then go into `apps/ios/src` folder and run pod install:
+
+```
+    cd apps/ios/src
+    pod install
+```
+Note: if you get the error: "CocoaPods could not find compatible versions for pod "MicrosoftFluentUI"," you will need to run `pod repo update` and then run `pod install` again.
+
+3. Return to the ios directory and run yarn ios to launch the FluentUI Tester app:
+
+```
+    cd ..
+    yarn ios
+```
+Note: the first time you yarn ios, you receive an error and have to run "FluentUITester.xcworkspace" directly from Xcode. The workspace can be found in the apps/ios/src folder. After running the workspace the first time from Xcode, you will be able to `yarn ios` from the CLI.
--- a/apps/README.md
+++ b/apps/README.md
@@ -1,15 +1,15 @@
-`FluentUI Tester` is the test app that we use to test our FluentUI components during development.
+`FluentUI Tester` is the test app that we use to test our FluentUI components during development. It uses the [react-native-test app](https://github.com/microsoft/react-native-test-app) under the covers, and loads the FluentUI bundle.
 
 ## Launch `FluentUI Tester` app on iOS
 
 Prereq: FluentUI Tester on iOS can only run on a Mac.
 
-1. Make sure you have followed the Getting Started instructions [here](../../README.md) to install packages and build the entire Fluent UI React Native repository. E.g. from the root of the repo:
+1. Make sure you have followed the [Getting Started](../../README.md) instructions to install packages and build the entire FluentUI React Native repository. I.e. from the root of the repo:
 ```
     yarn && yarn build
 ```
 
-2. Then go into `apps/ios/src` folder and run pod install:
+2. Then go into `apps/ios/src` folder and run pod install to pull in the project-level Cocoapod dependencies defined in the podfile, and to generate a valid xcworkspace:
 
 ```
     cd apps/ios/src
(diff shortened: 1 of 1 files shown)


apps/README.md
```diff
@@ -0,0 +1,26 @@
+`FluentUI Tester` is the test app that we use to test our FluentUI components during development.
```
Saadnajmi, line 1: Could we make a mention to the fact that it uses [react-native-test-app](https://github.com/microsoft/react-native-test-app) to generate the test app? Just so that developers can refer to documentation there as well. Also once again many thanks to @tido64 and the rest of his team for helping with this :D

apps/README.md
```diff
@@ -0,0 +1,26 @@
...
+    cd apps/ios/src
+    pod install
+```
+Note: if you get the error: "CocoaPods could not find compatible versions for pod "MicrosoftFluentUI"," you will need to run `pod repo update` and then run `pod install` again.
+
+3. Return to the ios directory and run yarn ios to launch the FluentUI Tester app:
+
+```
+    cd ..
+    yarn ios
```
Saadnajmi, line 24: Could also do 
```
cd apps/ios
pod install --project-directory=src/
yarn ios
```
to avoid the extra commands. It's technically more keystrokes, but is the more correct command I feel

apps/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+Note: the first time you yarn ios, you receive an error and have to run "FluentUITester.xcworkspace" directly from Xcode. The workspace can be found in the apps/ios/src folder. After running the workspace the first time from Xcode, you will be able to `yarn ios` from the CLI.
```
Saadnajmi, line 26: Troubleshooting:
- If you have the packager didn't launch in a separate terminal, you can run `yarn start` from `apps/ios` to launch it separately
- If you want to do direct debugging via xcode, after the `pod install`, you can launch `src/FluentUITester.xcworkspace` and build/run the scheme "ReactTestApp"
- If you want to have a clean rebuild of the generated iOS project, you can do the following:
```
cd apps/ios/
rm src/FluentUITester.xcworkspace
rm -r src/Pods/
pod install --project-directory=src.
```

apps/README.md
Saadnajmi: Also, I thought it was fluent-tester, we should probably make a PR to standardize that name, I'm pretty sure I still see Fabric Tester in code :D

Saadnajmi (changes requested):

Saadnajmi:
macOS has essentially the same instructions, so it would be nice to also include that markdown with this one

apps/README.md
```diff
@@ -0,0 +1,26 @@
+`FluentUI Tester` is the test app that we use to test our FluentUI components during development.
+
+## Launch `FluentUI Tester` app on iOS
+
+Prereq: FluentUI Tester on iOS can only run on a Mac.
+
+1. Make sure you have followed the Getting Started instructions [here](../../README.md) to install packages and build the entire Fluent UI React Native repository. E.g. from the root of the repo:
```
HeyImChris, line 7: To make this a little less verbose can we change it to:

"Make sure you followed the [Getting Started](../../README.md) instructions to install"....

apps/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+    yarn && yarn build
+```
+
+2. Then go into `apps/ios/src` folder and run pod install:
```
HeyImChris, line 12: Should probably explain why we need this (just like how you explained why we need to run yarn and yarn build in #1 above). For instance mentioning that pod install pulls in the project-level Cocoapod dependencies defined in the podfile and generates a valid `xcworkspace` one can use

apps/README.md
HeyImChris: Both are definitely good to me...  I do like the simplicity of what EJ has though

apps/README.md
HeyImChris: Re troubleshoooting- `git clean -xfd` would be a sure-fire way to wipe away all differences between your local repository and the one you cloned from upstream

HeyImChris (changes requested):

apps/README.md
Saadnajmi: Doing it the latter way seems to let cocoapods pick up some extra configs from react-native.config or whatever. 
Both work now, but if we add more custom stuff we miiight have to do it the latter more keystrokes way.

Regardless, for now both work so I'm ok with either. It'd be ok to leave as is

apps/README.md
Saadnajmi: That command can also remove uncommited changes you have so it's also dangerous :D 
I dunno that I would want to mention it in the documentation for setting up the test app without caveats (could cause more troubles while troubleshooting)

HeyImChris: Hah yes it can come with a warning on what it does. It’s also all documented in git

apps/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+    cd apps/ios/src
+    pod install
+```
+Note: if you get the error: "CocoaPods could not find compatible versions for pod "MicrosoftFluentUI"," you will need to run `pod repo update` and then run `pod install` again.
```
tido64, line 18: You can simplify this some what by just running `pod install --repo-update`

apps/README.md
ejlayne: Discussed offline, we're going to keep the name "FluentUI Tester", since the component libraries are part of the FluentUI family.

apps/README.md
ejlayne: Left as is for now, but open to future feedback.

apps/README.md
ejlayne: Thanks, fixed in new commit.

apps/README.md
ejlayne: Thanks, added your explanation in new commit.

apps/README.md
ejlayne: Fixed in new commit.

apps/README.md
ejlayne: Added changes in new commit. Let me know if you have any additional feedback, otherwise feel free to resolve comment.

apps/README.md
```diff
@@ -0,0 +1,37 @@
+`FluentUI Tester` is the test app that we use to test our FluentUI components during development. It uses the [react-native-test app](https://github.com/microsoft/react-native-test-app) under the covers, and loads the FluentUI bundle.
+
```
Saadnajmi, line 2: nit: loads the fluent-tester bundle, not FluentUI bundle

Saadnajmi:
So is the intention now to keep the documentation for all the apps at apps/README.md instead of having platform specific ones? Or is that just for now while we are only adding iOS?

Saadnajmi (approved):

apps/README.md
ejlayne: Fixed in recent commit.

ejlayne:
I'm doing separate PRs for each platform. For example, here's the macOS one: https://github.com/microsoft/fluentui-react-native/pull/465

HeyImChris (approved):

PR #467, opened 2020-09-21
Prototype Android Ripple effect on Button

tom-un:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

The Android Button in fluentui-react-native needs to have an option for a ripple effect.

This is a prototype wrapping the Android button root inside a [`TouchableNativeFeedback`](https://reactnative.dev/docs/touchablenativefeedback).   A better approach would be to use the [`Pressable`](https://reactnative.dev/docs/pressable), but that requires react-native 0.63 and fluentui-react-native is still on 0.62.2.

Ideally there would be a prop or token to turn on/off the ripple effect.   As this requires a wrapping TouchableNativeFeedback to be rendered I'm not sure how to pull it off.   With Pressable, it could just use the [`android_ripple`](https://reactnative.dev/docs/pressable#android_ripple) prop: it could be `undefined` when off or something when on.

### Verification

| furn-no-ripple | furn-ripple |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/Button.tsx
```diff
@@ -69,22 +73,26 @@ export const Button = compose<IButtonType>({
...
     // We shouldn't have to specify the source prop on Slots.icon, here, but we need another drop from @uifabricshared
     return (
-      <Slots.root>
-        <Slots.stack>
-          {info.icon && <Slots.icon source={renderData.slotProps!.icon.source} />}
-          {info.content && <Slots.content />}
-          {children}
-        </Slots.stack>
-      </Slots.root>
+      <Slots.touchable>
```
JasonVMo, line 76: We should be able to replace one of the view layers with the touchable, which is itself a view.

packages/components/Button/src/Button.types.ts
```diff
@@ -29,24 +29,27 @@ export interface IButtonInfo extends IPressableState {
...
 }
 
-export interface IButtonTokens extends FontTokens, IForegroundColorTokens, IBackgroundColorTokens, IBorderTokens {
+export interface IButtonTokens extends FontTokens, IForegroundColorTokens,
+                                       IBackgroundColorTokens, IBorderTokens {
   /**
    * The amount of padding between the border and the contents.
    */
-  contentPadding?: number | string;
+  contentPadding?: number|string;
```
JasonVMo, line 46: Where is this prettier setting coming from? Actually in general it looks like your prettier settings might be out of date in some way, are you using the ones from the repo?

packages/components/Button/src/Button.tsx
JasonVMo: So I would change the root to be the one that is sometimes a touchable. I also might try updating the experimental/Button which will replace this one shortly. In the new framework you don't have to have anything called root.

JasonVMo (changes requested):

packages/components/Button/src/Button.types.ts
tom-un: > Where is this prettier setting coming from? Actually in general it looks like your prettier settings might be out of date in some way, are you using the ones from the repo?

There are the settings from the repo.   Every time I hit Save in Visual Studio Code it rewrites the file based on the settings in `.vscode/settings.json`.   I'm surprised I'm the only one impacted by this.   In a previous PR I tried turning off the `"editor.formatOnSave": true` setting but was asked to revert that change: https://github.com/microsoft/fluentui-react-native/pull/418#discussion_r480435786

packages/components/Button/src/Button.tsx
tom-un: Actually, the [TouchableNativeFeedback](https://reactnative.dev/docs/0.5/touchablenativefeedback) must be a wrapper of a single View -- the TouchableNativeFeedback view itself doesn't have a style prop.   I guess I could make the top-most view be TouchableWithoutFeedback for non-Android plats, and TouchableNativeFeedback for Android only.

Saadnajmi:
I'm gonna close this since adding a ripple has since been addressed.

PR #478, opened 2020-09-23
Experimental theme package

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is both a refactoring change and a proposal of an alternative way of handling theming from a customer perspective.

#### `@fluentui-react-native/theme` module

This creates a new class called `ThemeReference` which is used to wrap theme objects. This class provides:

- the ability to layer a base theme, or another `ThemeReference`, together with a set of recipes to create a new theme. These recipes are either partial themes, or functions that produce partial themes
- an update mechanism, to listen for changes
- the ability to parent themes to other themes and have updates propogate as expected

This is designed to replace the code in theming-ramp, theme-registry, and theming-react-native (though that required factoring out the native module).

The typing on this is loose, the idea being that people could put their own variants of the theme and pass it through our context. Commenting for instance could extend our theme definition to add their own additional values, then look up information from their native module in a functional handler to merge the two approaches.

This also creates a new `ThemeProvider` which is a bit simpler as it only needs theme, not a registry.

- this shares the same ThemeContext as existing providers
- components will continue to work with either approach as they only reference the context for their lookups 

#### `@fluentui-react-native/win32-theme` module

This moves the code to produce an Office theme, including the code that works with the native module, into its own package. 

- This lets us take that code out of the bundle for platforms that don't need it.
- It opens the door for the native modules to vary slightly between platforms
- Customers that want their own native module can bypass any of our native module based solutions and just use our default theme

#### Usage

In practice the usage looks like:

```tsx
import { createOfficeTheme } from '@fluentui-react-native/win32-theme';
import { ThemeProvider, ThemeReference } from '@fluentui-react-native/theme';

// this will actually create a theme reference
const theme = createOfficeTheme('TaskPane');

// create a slight variant, based off the Office theme (this could be wrapped in a function as well)
const variantTheme = new ThemeReference(theme, parent => {
  // do some stuff
});

const render = () => {
  <ThemeProvider theme={theme}>
    //content
  </ThemeProvider>
}
```

### Verification

Did the following:

- added some UTs to the theme package
- ran through the web tester manually
- ran through the win32 tester side by side with an unchanged version

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/theme/README.md
```diff
@@ -0,0 +1,60 @@
+# @fluentui-react-native/use-slots
+
+A pattern and framework for building layering components together in an efficient, pluggable, and customizable manner.
```
tkamireh, line 3: building and layering?

packages/experimental/theme/README.md
```diff
@@ -0,0 +1,60 @@
...
+  const { children, ...rest } = props;
+  const propsForBase = doSomethingWithProps(rest);
+  // call the first part of the function
+  const continuation = BaseComponent._staged(propsForBase);
+  // now render that component just by calling the function
+  return continuation({}, children)
+}
+```
+
+Note that in this example the `WrapperComponent` is leveraging the code of the `BaseComponent` but will not create a dedicated tree entry. Also the `WrapperComponent` itself could be implemented as a staged component.
```
tkamireh, line 56: If I'm understanding this correctly, what `stagedComponent` give us is a way to compress from n (where n > 2) layered components in the render tree to 2. So if `BaseComponent` renders a `View` in it's continuation `WrapperComponent` looks like
```
<WrapperComponent>
      <View>
            ....
      </View>
</WrapperComponent>
```

So is this pulling out some of the ideas from the original `compose` and `composable` concepts into their own helpers?

packages/experimental/theme/src/mergeTheme.ts
```diff
@@ -0,0 +1,6 @@
+import { immutableMerge } from '@fluentui-react-native/immutable-merge';
+import { Theme, PartialTheme } from '@fluentui-react-native/theme-types';
+
+export function mergeTheme<TTheme extends Theme, TPartial extends PartialTheme>(base: TTheme, partial: TPartial): TTheme {
+  return immutableMerge(base, (partial as unknown) as TTheme);
```
tkamireh, line 5: is the casts to `unknown` necesary - and does immutableMerge not accept partials (or maybe TS isn't smart enough to realize it's a partial)?

packages/theming/win32-theme/src/createPartialOfficeTheme.ts
```diff
@@ -0,0 +1,33 @@
+import { ColorRamp, OfficePalette, PartialTheme } from '@fluentui-react-native/theme-types';
+import { OfficeThemingModule } from './NativeModule/officeThemingModule';
+import { paletteFromOfficeColors } from './paletteFromOfficeColors';
+
+const createColorRamp = ({ values, index = -1 }: Partial<ColorRamp>) => ({
+  values,
+  index,
```
tkamireh, line 7: I've seen `index` a few times, what does it get used for? And why do we default to -1?

packages/experimental/theme/README.md
JasonVMo: Doh, just copied this file (obviously) and need to update this.

packages/experimental/theme/README.md
JasonVMo: Yes, that's the way the experimental framework is built. The pieces can be used on their own, or together. Compose brings them all together, there is no more vanilla composable (which was confusing to folks).

That being said, this is the existing documentation file in use-slots, which accidentally got copied when I created the new package.

packages/experimental/theme/src/mergeTheme.ts
JasonVMo: TS isn't smart enough because a Partial<MyType> is just a mapped type where it adds ? to every key within. The theme partial type is built up from explicitly declared partial sub-types. You can build a deep partial but then it does weird things to objects like ColorRamps where you really should be specifying multiple elements.

packages/theming/win32-theme/src/createPartialOfficeTheme.ts
JasonVMo: This was actually moved from theming-react-native, @FalseLobster might have better context for that.

packages/experimental/theme/README.md
tkamireh: Makes sense :)

packages/experimental/theme/src/mergeTheme.ts
tkamireh: Ah right the deep partials are what you want

tkamireh (approved):
Besides documentation changes, lgtm

packages/theming/win32-theme/src/createPartialOfficeTheme.ts
FalseLobster: That was based on trying to align with a now out-of-date version of the Web's theming ideas.  I don't think it's relevant any more and should probably get removed since it's just complicating the theme dictionary.

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
```diff
@@ -1,139 +0,0 @@
-{
```
FalseLobster, line 1: Did this get deleted and not just moved?

packages/theming/win32-theme/src/createPartialOfficeTheme.ts
JasonVMo: I agree, though I want the design we move towards to be a bit more stable before we transition.

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
JasonVMo: Yes. Looking at it I don't think we need it since this was only being used in the case where the native module was not being used, but we already have a complete palette definition with the white palette so this seemed redundant.

FalseLobster: I think you're overlooking a few things here, but I'm still trying to run through these changes to make sure I understand the repercussions, so give me a moment to collect my thoughts

FalseLobster: I think my main concern is that the fallback module doesn't actually have a full palette whereas the debugpalette.json did.  This means that certain keys in the office palette object will be undefined when consumers turn on web debugging.  This will likely break existing consumers even if the Fluent theme translation only uses a subset of the native colors.

It's also worth noting that just because getPalette (a synchronous native module) is unsafe in web debugger, that doesn't mean the module isn't there.  The old approach just replaced the actual NM.getPalette() call with a web-debug safe version but left the rest of the NM intact, whereas the new one actually always uses the fallback Module.

The recipe based approach does look much nicer over all and simplifies things, though.

JasonVMo: Makes sense. I definitely want your feedback on whether I am using the native module correctly. In particular I think I may be conflating several things. The idea of the base theme name and the palette that gets looked up from getPalette() as I think I'm using the base theme name as the default lookup value for the palette. But it seems like one of them should be things like 'Colorful' and the other should be things like 'TaskPane'.

FalseLobster: Oh you're right I didn't even notice that at first.

packages/theming/win32-theme/src/createOfficeTheme.ts
```diff
@@ -0,0 +1,40 @@
...
+ * fluent defaults but will attempt to use the theming native module to get information about
+ * the office palette.
+ *
+ * This theme will also listen for native changes and reload itself when things change on the native side of things
+ *
+ * @param paletteName - optional specifier for the currently active office palette
+ */
+export function createOfficeTheme(paletteName?: string): ThemeReference {
+  const [module, emitter] = getThemingModule();
+  const ref = { module, emitter, defaultPalette: module.initialHostThemeSetting || '' };
```
FalseLobster, line 23: As you already noted initialHostThemeSetting is akin to Colorful/Black/Dark/Fresh whereas paletteName is basically a scheme within the broader office theme.

FalseLobster: paletteName could just be left undefined when passed into handlePaletteCall

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -19,7 +20,10 @@ export interface Theme {
   components: { [key: string]: object };
   componentTokens?: object;
   spacing: Spacing;
-  host: object; // platform specific host settings
+  host: {
+    // Office palette, if running in Office with the native module connected in the theme
+    palette?: OfficePalette;
```
FalseLobster, line 25: You don't have any quibbles about leaving this in type in a semi-official capacity?

packages/theming/win32-theme/src/createOfficeTheme.ts
FalseLobster: Since the themeRef ThemeTranslation closure should capture the paletteName, and the paletteName can be left undefined if the user didn't given, you can just replace that key value pair in the ref object with the initialHostThemeSetting / args.hostThemeSetting and forward that into createPartialTheme, where you can set
{
name: hostThemeSetting || '';
}

packages/experimental/theme/src/themeReference.ts
```diff
@@ -0,0 +1,101 @@
...
+  private listeners: OnThemeChange[] = [];
+  private getParent: () => TTheme;
+  private parentRef: ThemeReference<TTheme, TPartial> | undefined = undefined;
+
+  /**
+   * Create the theme reference, either as a plain wrapper, or a wrapper with additional transforms and/or merging
+   * @param base - can be either another ThemeReference object, or a fully specified theme
+   * @param recipes - any number of recipies to be applied on top of the theme object
+   */
+  constructor(base: TTheme | ThemeReference<TTheme, TPartial>, ...recipes: ThemeRecipe<TTheme, TPartial>[]) {
```
FalseLobster, line 41: Do we want to export some sort of baselineTheme ThemeReference?

FalseLobster: Neveeeeermind, I was just trying to avoid the boilerplate of 

`new ThemeReference(getBaselinePlatformTheme(), createOfficeTheme(), myThemeRecipe);`

but then I realized/remembered the createOfficeTheme was a ThemeReference not a ThemeRecipe and  is already layered on the default-theme.

I do wonder if we should take an optional baseThemeReference in createOfficeTheme in case they want to change that though,

FalseLobster: I'm not sure how common that scenario is in practice though.

FalseLobster (approved):
Overall this does seem friendlier to use, I was never a fan of the coupling I had to do between the ThemeRegistry and ThemeModuleHelper to get the old stuff working correctly.

packages/theming/theme-types/src/Theme.types.ts
JasonVMo: I've been debating this but I don't think so, though perhaps we rename it officePalette. I kind of envision host maybe eventually looking something like:

```ts
// host is basically a grouping for settings that come from native modules
host: {
  dark?: boolean;
  highContrast?: boolean;
  officePalette?: OfficePalette;
  win32?: Win32Settings;
  windows?: WindowsSettings;
  macos?: MacSettings;
  // or something like this
}
```

Then the types in ThemeReference are general so that someone can do the following:

```ts
type MyTheme = Theme & {
  host: {
    myAdditionalNativeSettings?: SomeType;
  }
}
```

Then basically fill those settings the same way the win32 theme does. You can imagine comments creating a theme, based off the Win32 theme, where they augment it with their native module but use our context and provider mechanism. Then we get rid of the dual context problem we've had in the past.

tkamireh: +1 to allowing for flexible and general usage, but one thing I'd be interested in is keeping all these names generic (`officePalette` seems a little constraining maybe) - there could be a world in the future where the native modules are written in a way where they're pluggable. Other brownfield scenarios where large applications serve as "hosts", aren't too far-fetched a notion imo.

JasonVMo: Yeah, I tend to agree though the question is where that coalesces. There are several levels of definitions and naming.

- **Base** - need a good name but these are things like `theme.colors.neutralPrimary` or `theme.colors.neutral3` or whatever we come up with
- **Semantic** - these are role based.  Right now they are things like `theme.colors.bodyBackground` but may change. Design is working on a new set of these now, and while they feel like the base layer may be volatile, these names might be stable(ish)
- **Native/Host** - at some point there is a different concept of naming that maps to what exists in native. That can either be exposed in the native module, or on the theme somewhere, or be hidden on the native side.

I think there are a couple of principles I've been thinking about:

- components should be authored against the semantic level. It is kind of the goldilocks level here and allows for a role based mapping of arbitrary color schemes.
- this means that customizations get authored against the semantic level, which makes sense
- this also implies native needs to map to semantic, rather than mapping to the base layer, which is great because this maps to the way these things tend to be built on the native side.
- this native -> semantic mapping can be lossy, meaning that some native values may have no mapping in the semantic layer. 
- This implies that there will likely be value in having those available somewhere in the theme, because then a component can be customized easily to depend on those colors on a given platform.

This is all subject to debate and discussion but this is just the way I've been starting to think about things.

packages/experimental/theme/src/themeReference.ts
JasonVMo: It's an interesting idea but I'm not sure how common it is in practice as well. It's easy enough to do, and we can do it later without changing the interface by factoring out a helper and exporting the more generic helper.  So I'm going to hold off for now.

packages/theming/win32-theme/src/createOfficeTheme.ts
JasonVMo: I updated the code around this to match the expected behavior in the latest iteration.

packages/framework/theming-react-native/src/NativeModule/ThemingModuleHelpers.ts
```diff
@@ -46,15 +45,12 @@ export function createThemingModuleHelper(themingModule?: OfficeThemingModule, e
   let _hostTheme = themingModule.initialHostThemeSetting || '';
   emitter &&
     emitter.addListener('onPlatformDefaultsChanged', (args: PlatformDefaultsChangedArgs) => {
-      _hostTheme = (args && args.hostThemeSetting) || _hostTheme;
```
FalseLobster, line 70: You can't rename this without native theming module changes:

```
        dynamic params = dynamic::array("onPlatformDefaultsChanged", dynamic::object("hostThemeSetting", getThemeSetting()));
        instance->callJSFunction("RCTDeviceEventEmitter", "emit", std::move(params));
```
That creates a JSON object `{ hostThemeSetting: 'Foo' }`

Sorry, name is probably clear, I think the native code calls it the Theme Setting so I was probably borrowing that nomenclature

FalseLobster: Thankfully we can name it whatever we want in our pure JS interfaces, so this is really just an implementation detail

packages/theming/win32-theme/src/createOfficeTheme.ts
```diff
@@ -20,20 +20,20 @@ function handlePaletteCall(palette: OfficePalette | CxxException): OfficePalette
...
+    const name = paletteName || '';
     const palette = handlePaletteCall(ref.module.getPalette(name));
-    return createPartialOfficeTheme(module, palette);
+    return createPartialOfficeTheme(module, ref.themeName, palette);
   });
 
   // set up the callback for theme changes on the native side
   const onPlatformDefaultsChanged = (args: PlatformDefaultsChangedArgs) => {
-    ref.defaultPalette = (args && args.hostThemeSetting) || ref.defaultPalette;
+    ref.themeName = (args && args.hostThemeName) || ref.themeName;
```
FalseLobster, line 33: Same here, the args object's shape is defined by the native side.

packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
```diff
@@ -0,0 +1,31 @@
...
+  disabled && console.warn(console.warn('Web Debugging forces Theming Native Module to fallback to fake color values.'));
+  return disabled;
+}
+
+export function getThemingModule(): [OfficeThemingModule, NativeEventEmitter | undefined] {
+  const module = NativeModules && NativeModules.Theming;
+  // if the native module exists return the module + an emitter for it
+  if (module) {
+    // mock getPalette if it should be disabled, otherwise return the module directly
+    return [disableGetPalette() ? { ...module, getPalette: fallbackGetPalette } : module, new NativeEventEmitter(module)];
```
FalseLobster, line 26: So this does leave the NativeModule intact like the old version did, but you're still using a different fallbackGetPalette.

The old version did a require on the debugpalette.json you deleted.  That json was more fully specified than the impl in the fallbackThemingModule code, and it means that key-value pairs in the palette will disappear if we don't update the fallbackGetPalette to include the rest of the entries in the .json.

Now that I'm thinking about it, though, maybe we should get rid of the .json and export it from the NM as a constant.  I think I probably did it this way to avoid a second check in to the native code at the time.

packages/theming/theme-types/src/Theme.types.ts
tkamireh: I generally agree with the laid out principles. I would add, as a direct consequence of native -> semantic mapping be lossy + the probable usage of base when semantic is insufficient, that affordances for native -> base mappings are probably unavoidable. And I think the code already reflects this.

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
JasonVMo: Are we sure the debug palette is actually more complete? I believe the one defined in JS is fully specified since all the values in the palette are required.

packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
JasonVMo: I think both the json and the code in the fallback have all the keys specified, there's an open question of which one is more correct, but they both should work.

packages/theming/win32-theme/src/createOfficeTheme.ts
JasonVMo: Good catch

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
JasonVMo: Actually looking at just the counts it looks like debugpalette does in fact have about 30 more entries. If that is complete, then we need to add those entries to the type in OfficePalette and then ensure they are specified.

JasonVMo: I'll diff it later this afternoon

JasonVMo: I added the missing values to the type and to the definition. There is one discrepancy, in the JSON file there are a series of values named things like StrokeOverRest and in the OfficePalette type they are named StrokeOverlayRest. I'm working off the assumption the types with Overlay were correct, not the Over ones.

packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
JasonVMo: I did a diff of the files and added the types to the OfficePalette type, then patched the keys from the debugjson.json file. We probably do want to copy a full standard White or Colorful theme definition to ensure that when the fallbacks happen they are doing the right thing.

packages/theming/theme-types/src/Theme.types.ts
FalseLobster: Yea I think I'm _mostly_ on board with this strategy, although I think the one place I might object to the mapping of native -> semantic is with the brand/accent colors.

FalseLobster: Sent too soon --- I think you had concerns about this mapping being difficult to reason about because (currently) there are so many different accent colors in the ramp and we only use a subset--- I.e. if we prune out Accent 20-45, is it weird they we assign App2 to Accent 15 and App3 to Accent50?

I'm concerned that not mapping App -> Base layer here undercuts the flexibility of the Fluent theme and limits the ability to inject different brands.  Maybe this is something design should have a say in though.   Any thoughts about that?

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
FalseLobster: Actually the Over ones are correct--- the .JSON was taken straight off a devmain checkpoint the mock module was hand authored.

Other than that, the latest revision looks good to me.

packages/theming/theme-types/src/Theme.types.ts
JasonVMo: I definitely want design to iterate a bit more. From talking to Travis it sounds like their notion of base colors are more fluid than their notion of semantic ones. (They call them Sets) In general design might retarget the set value from Neutral.89 to Neutral.73 and we shouldn't rely on meaning for Neutral.73.

Because the set/semantic colors have explicit branded values that would be where we would map brand colors.

packages/framework/theming-react-native/src/NativeModule/office/debugpalette.json
JasonVMo: Perfect, I updated the types, then changed the values and removed the commented out ones. I'm going to merge this PR once the CI loop passes. Thanks for all the feedback.

PR #482, opened 2020-09-24
Experimental iOS Component "Avatar"

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On iOS, macOS, and Android, we have the Avatar control. On Web and React Native, we have the PersonaCoin. They have slightly different API's and it'll be a while before we can converge on a shared API. However, we have partner teams who want to use an avatar component on iOS _now_. Furthermore, they require that the React Native Avatar look and behave _exactly_ like the native FluentUI Avatar, because they will have both in close proximity in their experience. Waiting for these two API's to converge would be a serious blocker.

Thus, I decided to introduce a new experimental package: "@fluentui-react-native/experimental-avatar". This will exist side by side with PersonaCoin, and start out as a wrapper of the native FluentUI Apple (iOS) control Avatar. Overtime, my intention is that as we converge the Avatar and PersonaCoin API's, we do so on this package, and eventually replace PersonaCoin (we've already decided that "Avatar" is the name to go with, after all).

There will be lots of duplication initially between experimental-avatar and PersonaCoin. That is to be expected. There are also some Todo's (make sure we have a flat list of properties, map overrideAccessibilityLabel), but this is more or less a complete map of the native control. 

### Verification

Added a new test on iOS that renders a few avatar views

Screen Shot 2020-09-24 at 1 05 47 AM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/RNTester/TestComponents/Avatar/AvatarTest.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Test, TestSection, PlatformStatus } from '../Test';
+
+const avatar: React.FunctionComponent<{}> = () => {
+  return (
+    <Stack style={stackStyle}>
+      <Avatar avatarData={{ primaryText: 'John Smith' }} />
+      <Avatar avatarData={{ primaryText: 'John Smith' }} avatarStyle="square" />
+      <Avatar avatarData={{ primaryText: 'John Smith', presence: 'available' }} />
+      <Avatar avatarData={{ primaryText: 'John Smith', secondaryText: 'johnsmith@microsoft.com', presence: 'outOfOffice' }} />
+      <Avatar avatarData={{ secondaryText: 'johnsmith@microsoft.com', presence: 'outOfOffice' }} />
```
Saadnajmi, line 15: As @JasonVMo has already pointed out, having a sub-object like so is not the best/efficient way to do this. The biggest TODO of this PR that stops it from being a full PR is having a flat list of props on the JSX "Avatar" component, which then maps to this object when passed to the inner native component

https://github.com/microsoft/fluentui-react-native/pull/372#discussion_r474326015

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * The secondary text to create initials with if primary text is not provided (e.g. an email address)
+   */
+  secondaryText?: string;
+
+  /**
+   * The image to be displayed
+   */
+  image?: Image;
```
Saadnajmi, line 37: We need to make sure we can set a base64 encoded image here, as per requirements.

apps/fluent-tester/src/RNTester/TestComponents/Avatar/AvatarTest.tsx
Saadnajmi: Perhaps we can use the memo cache as mentioned here (https://github.com/microsoft/fluentui-react-native/pull/462#discussion_r492195973) to map the props and make sure we don't rerender every time

JasonVMo: I think this usage shows some of the issues with the interface. From the perspective of the user, most of the data they want to input is actually in the sub-object, which is really a useless level of indirection for 1, but also has the risk of this style of usage with inline objects being specified. There are linting rules for RN that will detect inline styles but I don't think they work for other objects.

Similar to shimmer, it's fine for the native component to have an interface like this, the JS side can translate from a good JS interface to the native one.

JasonVMo: Also internally you can memo on the input object reference, but that breaks if people use it like you specified above. Alternatively you could split apart the avatarData into the separate props, then memo on the key combinations to form the internal object to pass to native, which is fine. But at that point what value did the avatarData prop have at the JS level if it has to be broken apart.

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Set to true to enable the pointer interaction on the avatar view, false by default.
+   * TODO Figure out how to make iOS only
+   */
+  hasPointerInteraction?: boolean;
+
+  // TODO Figure out how to make this a flat list at the AvatarProps level,
+  // and map to the inner custom property in the native view
+  // https://github.com/microsoft/fluentui-react-native/pull/372#discussion_r474326015
+  avatarData?: AvatarData;
```
JasonVMo, line 123: Why not just add the contents here to the props directly? Also why not add them directly on the native side? It's easy to collect props and construct an object on either side. If they are all just props it makes it easy to work with them correctly and ensure shallow compares work as expected.

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,162 @@
...
+  tokens: [
+    {
+      size: 'small',
+    },
+    avatarName,
+  ],
+  tokensThatAreAlsoProps,
+  slots: { root: NativeAvatarView },
+  slotProps: {
+    root: buildProps<AvatarProps, AvatarTokens>(
```
JasonVMo, line 147: I think this should be buildProps<NativeAvatarViewProps, AvatarTokens> as you are converting the tokens into native props.

JasonVMo (changes requested):
Mainly some feedback on the interface shape. TLDR: why not just promote all the things in AvatarData into props on both the native and JS side, then do any object construction you need on the native side?

packages/experimental/Avatar/src/Avatar.tsx
amgleitman: If we make this an `ImageURISource`, then we can easily do that by specifying a [data:... URI](https://en.wikipedia.org/wiki/Data_URI_scheme).

packages/experimental/Avatar/src/Avatar.tsx
Saadnajmi: @JasonVMo , my new plan (similar to what I did for Shimmer) is to expose the props on the native side instead of having the intermediate `AvatarData` object. So the next steps for this PR will involve a change to FluentUI Apple, then an update here once we consume it. I expect that to take a few days, and I'll return to this PR once I can integrate those changes =)

Saadnajmi: And I lied. Changing the native API is perhaps not the right approach. For context, in Swift, creating an AvatarView would look like this:

```swift
let avatarView = AvatarView(avatarSize: .large, withBorder: false, style: .circle)

avatarView.setup(primaryText: person.displayName, secondaryText: person.emailAddress, image: nil)
// OR
let avatarData = AvatarData(primaryText: person.displayName, secondaryText: person.emailAddress)
avatarView.setup(avatar: avatarData)
```
`primaryText` and `secondaryText` are not public properties on the native avatarview, but are set via the `setup` method. Calling the `setup` method again also clears out the old values (I guess it wanted to be an "immutable" view based on the properties you set it up with).

Changing the native API to make primary/secondaryText/etc publicwould mean introducing an alternative to setup method, and perhaps confuse clients. @harrieshin and @rdeassis expressed concern with doing so.

Instead, I went for the flat JS API that maps to a more complex native API route. I *believe* I did it correctly with the memoCache to ensure we don't have unnecessary re-renders, but I'd love to verify that with someone from CXE.

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -0,0 +1,80 @@
+#import <React/RCTComponent.h>
+#import <React/RCTViewManager.h>
+
+@import FluentUI;
+
+ @implementation RCTConvert (MSFAvatarView)
```
Saadnajmi, line 6: I probably need one of these for each enum.  I don't know why this works as is :D 
Also, nit: that one space

packages/experimental/Avatar/ios/AvatarViewManager.swift
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	override class func requiresMainQueueSetup() -> Bool {
+		return true
+	}
+	
+	override func constantsToExport() -> [AnyHashable : Any]! {
+			return [
+				"sizes" : [
+					"xSmall" :	AvatarSize.extraSmall.size.width,
```
HeyImChris, line 18: Extra spacing here

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -0,0 +1,80 @@
...
+	@"blocked": @(MSFPresenceBlocked),
+}), MSFPresenceNone, integerValue);
+
++ (MSFAvatarData *)MSFAvatarData:(id)json
+{
+	return [[MSFAvatarData alloc]initWithPrimaryText:[RCTConvert NSString:json[@"primaryText"]]
+									   secondaryText:[RCTConvert NSString:json[@"secondaryText"]]
+												image:[RCTConvert UIImage:json[@"image"]]
+											presence:[RCTConvert MSFPresence:json[@"presence"]]
+											   color:[RCTConvert UIColor:json[@"color"]]];
```
HeyImChris, line 47: nit; this whole return doesn't seem to be spaced properly

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -0,0 +1,80 @@
...
+RCT_EXPORT_VIEW_PROPERTY(avatarBackgroundColor, UIColor);
+RCT_EXPORT_VIEW_PROPERTY(customBorderImage, UIImage);
+
+RCT_REMAP_VIEW_PROPERTY(avatarStyle, style, MSFAvatarStyle)
+
+RCT_EXPORT_VIEW_PROPERTY(borderColor, UIColor);
+RCT_EXPORT_VIEW_PROPERTY(presence, MSFPresence);
+RCT_EXPORT_VIEW_PROPERTY(useOpaquePresenceBorder, bool);
+
+//TODO Remap this
```
HeyImChris, line 68: What's going on with this prop? TODOs should be paired with bug numbers

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,176 @@
...
+   * The color of the border of the avatar view
+   */
+  borderColor?: string;
+
+  /**
+   * When true, the presence status border is opaque. Otherwise, it is transparent.
+   */
+  usesOpaquePresenceBorder?: boolean;
+
+  // TODO Map the viewProp AccessibilityLabel to this
```
HeyImChris, line 98: Todo- bug number

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,176 @@
...
+  // overrideAccessibilityLabel? : string;
+
+  /**
+   * Used when avatarView doesn't have image or can't generate initials string
+   */
+  preferredFallBackImageStyle?: string;
+
+  /**
+   * Set to true to enable the pointer interaction on the avatar view, false by default.
+   * TODO Figure out how to make iOS only
```
HeyImChris, line 108: For now commenting the platforms it supports is a good start. Maybe @JasonVMo knows how we've been doing platform specific props?

HeyImChris (changes requested):

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
Saadnajmi: The property existed on the native view, and I wasn't really sure what to do with it. I'll either remove it or add a bug number

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
```diff
@@ -0,0 +1,61 @@
...
+    size: 'large',
+  });
+
+  return (
+    <Stack style={stackStyle}>
+      <CustomizedAvatar primaryText="John Smith" />
+      <CustomizedAvatar primaryText="John Smith" avatarStyle="square" />
+      <CustomizedAvatar primaryText="John Smith" presence="available" />
+      <CustomizedAvatar primaryText="John Smith" secondaryText="johnsmith@microsoft.com" presence="outOfOffice" />
+      <CustomizedAvatar secondaryText="johnsmith@microsoft.com" presence="outOfOffice" />
```
harrieshin, line 31: we should add phone number case or non English case? just a thought

Saadnajmi: We support phone numbers for secondaryText? I ca take a look at the cases we have in the native test app

harrieshin: phone number as primary text for phone sign in

Saadnajmi: OK, we don't have those examples in the iOS test app :D

Saadnajmi: I think i'll leave the "better test cases to a separate PR, because I think I can steal a bunch of stuff from the PersonaCoin test cases

packages/experimental/Avatar/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "onPublish": {
+    "main": "lib-commonjs/index.js",
+    "module": "lib/index.js"
+  },
+  "scripts": {
+    "build": "fluentui-scripts build",
+    "just": "fluentui-scripts",
+    "clean": "fluentui-scripts clean",
+    "lint": "fluentui-scripts eslint",
+    "test": "fluentui-scripts jest",
```
JasonVMo, line 17: should this also have a depcheck task?

JasonVMo: Just for more context, any package that has a depcheck task (generally `"depcheck": "fluentui-scripts depcheck"`) will have that task run as part of the CI loop to check for missing or unused dependency declarations.

packages/experimental/Avatar/src/Avatar.tsx
JasonVMo: I think there are a few ways to think about props. Beyond the props that just apply to all platforms, the next preference would be something that has:

- A name and semantic meaning that isn't platform specific
- Implementations for platforms where it applies, no-op for ones where it doesn't
- Comments denoting the expected behaviors and where it applies

If on the other hand it is truly platform specific, then the pattern seems to be to either put the platform in the prop name or fork the interface. Given the fact that TypeScript doesn't work well with interfaces that change shape per platform I'd prefer either semantic names or platform specific names to forked interfaces.

JasonVMo: This one seems to fall in the first bucket, enabling pointer interaction seems like a switch that could apply to any platform.

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,176 @@
+/** @jsx withSlots */
+import { compose, UseSlots, buildProps, mergeProps, withSlots } from '@fluentui-react-native/framework';
+import { Image, NativeModules, ViewProps } from 'react-native';
+import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
+import { getMemoCache } from '@fluentui-react-native/framework';
+
+const cache = getMemoCache();
+
+const avatarName = 'MSFAvatarView';
```
JasonVMo, line 9: I would disambiguate the avatarName from the native component name. If this is Avatar, it makes sense for the name to be 'Avatar' from the perspective of the displayName and theme lookup.

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const cachedAvatarDataProp = cache(
+      () => ({
+        primaryText: props.primaryText,
+        secondaryText: props.secondaryText,
+        image: props.image,
+        color: props.color,
+        customBorderImage: props.customBorderImage,
+        presence: props.presence,
+      }),
+      [props.primaryText, props.secondaryText, props.image, props.color, props.customBorderImage, props.presence],
```
JasonVMo, line 171: One thing to consider when caching is the likelihood of re-use vs memory overhead. Given that this data includes content values, and as such every element of a list might have a different avatar data, your primary goal will be to insure render to render consistency, rather than cross-component sharing. From this I would switch this from memo-cache to React.useMemo.  Here's a quick breakdown of when to use each:

## memo-cache
**How it works:** Persistent, hierarchical (per-key) cache, scope of storage is up to you as the user controls the cache reference.
**Best uses**: When data can often be shared across component instances. E.g. if you are rendering a hundred instances on a screen and the expectation is that much of the data can be shared across component instances.

## React.useMemo
**How it works**: Actually the baseline for almost all hook implementations (useRef, useCallback, etc.) This stores the value in a hook slot (just like useState would), and compares the keys via a mixed array. This will not share across components, and only stores the last used. But it is lighter weight and likely a touch faster.
**When to use**: When you just want to ensure that if you render a single component twice in a row, the sub-objects don't mutate. When the values to memoize involve content (like a label) that will be different for every component.

packages/libraries/core/src/index.ts
```diff
@@ -2,6 +2,7 @@ export * from '@fluentui-react-native/button';
 export * from '@fluentui-react-native/callout';
 export * from '@fluentui-react-native/checkbox';
 export * from '@fluentui-react-native/contextual-menu';
+export * from '@fluentui-react-native/experimental-avatar';
```
JasonVMo, line 5: Do we want to put it in the library when it is still experimental? Users can just pull it in directly. I would expect that we would want to figure out the intersection between Avatar and Persona before it goes in the library.

JasonVMo (changes requested):
Generally looks good, just a few comments.

packages/libraries/core/src/index.ts
Saadnajmi: Good point. If users can pull it in separately, then that may actually be better for getting this used by our first client, as they won't need to update their version of FURN just yet. Thanks for the recommendation =)

packages/experimental/Avatar/package.json
Saadnajmi: Must have missed it, thanks!

Saadnajmi:
https://github.com/microsoft/fluentui-react-native/pull/482#discussion_r503491466

@JasonVMo in this case, "pointerInteractions" refers specifically to the new iPadOS pointer support and the pointer interactions that were just added. https://www.youtube.com/watch?v=9iO52-MIBP0
So, this is very iOS specific at the moment, and I"ll look into naming it through one of your suggested ways

packages/experimental/Avatar/src/Avatar.tsx
Saadnajmi: Thanks for the explanation! I've switched to useMemo

packages/experimental/Avatar/ios/AvatarViewManager.swift
```diff
@@ -0,0 +1,27 @@
...
+		let avatarView = AvatarView(avatarSize: .small)
+		return avatarView
+	}
+
+	override class func requiresMainQueueSetup() -> Bool {
+		return true
+	}
+
+	override func constantsToExport() -> [AnyHashable : Any]! {
+			return [
```
HeyImChris, line 16: nit- the spacing in this method isn't right

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -0,0 +1,76 @@
...
+	@"outOfOffice": @(MSFPresenceOutOfOffice),
+	@"offline": @(MSFPresenceOffline),
+	@"unknown": @(MSFPresenceUnknown),
+	@"blocked": @(MSFPresenceBlocked),
+}), MSFPresenceNone, integerValue);
+
++ (MSFAvatarData *)MSFAvatarData:(id)json
+{
+	return [[MSFAvatarData alloc]initWithPrimaryText:[RCTConvert NSString:json[@"primaryText"]]
+                                     secondaryText:[RCTConvert NSString:json[@"secondaryText"]]
```
HeyImChris, line 44: nit- misaligned with first line's colon

HeyImChris (approved):

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
Saadnajmi: First line uses a tab, rest of the lines uses spaces assuming tab width is 2 spaces. I think github assumes a different tabwidth. I looked into what we could do to address that, and maybe we could add a .editorconfig https://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github/23522945#23522945

packages/experimental/Avatar/ios/AvatarViewManager.swift
Saadnajmi: Yep, there is an extra tab. I'll fix that

JasonVMo (approved):
Looks good!

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
Saadnajmi: #523 for this one

PR #489, opened 2020-09-25
ContextualMenu: manage focus state on hover

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Checkbox.settings
- update focus indicator to meet most recent design (2px solid border around text)
ContextualMenuItem.settings
- remove background color styling for hover. focus styling will manage hover as well
- make sure hovering on disabled items will focus without a focus indicator border
ContextualMenuItem.tsx
- call focus() and blur() imperatively on componentRef of menu item on hover events
- call onFocus() and onBlur() as part of onMouseEnter and onMouseLeave events
- onKeyup, add a checker for focused state. If item is not focused, do not trigger onItemClick event

### Verification

ContextualMenu was manually tested by confirming that hovering matched focus 

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -43,15 +44,19 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
...
+    const onItemHoverIn = React.useCallback(
+      () => {
+        componentRef.current.focus();
+      }, [componentRef]);
 
-    // attach the pressable state handlers
-    const pressable = useAsPressable({ ...rest, onPress: onItemClick });
+    const onItemHoverOut = React.useCallback(
+      () => {
+        componentRef.current.blur();
```
FalseLobster, line 56: I'm not sure I'd bother with the blur call.  The next item to take focus will remove focus from the previous item.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -62,7 +67,29 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
       icon: !!icon
     };
 
-    const cmRef = useViewCommandFocus(userProps.componentRef);
+    const onKeyUp = React.useCallback(
```
FalseLobster, line 70: We might have a useKeyCallback hook you could leverage already?  Worth a look.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -62,7 +67,29 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
...
+      e => {
+        if ((e.nativeEvent.key === 'Enter' || e.nativeEvent.key === ' ') && pressable.state.focused) {
+          onItemClick(e);
+        }
+      }, [onItemClick, pressable.state]);
+
+    const onMouseEnter = React.useCallback(
+      e => {
+        pressable.props.onMouseEnter && pressable.props.onMouseEnter(e);
+        pressable.props.onFocus && pressable.props.onFocus(e);
```
FalseLobster, line 80: onFocus should get called automatically by the call to focus() above.

Honestly I think the 

...pressable.props 

below is already providing an onMouseEnter/Leave that should work, you shouldn't need to override the implementation from the pressable hook, so I think you can just delete your onMouseEnter/Leave callbacks

FalseLobster (changes requested):

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
lenahong: I initially tried implementing without these but without calling pressable.props.onFocus()/onBlur(), mouse focus stays on the last hovered item when I hover off the menu.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
lenahong: Got it!

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
lenahong: useKeyCallback seems useful for keyboard events with only one key type. I have to support both ' ' and 'Enter' so it might be just convenient for me to define this as is

packages/components/Checkbox/src/Checkbox.settings.ts
```diff
@@ -45,8 +45,8 @@ export const settings: IComposeSettings<ICheckboxType> = [
     content: {
       variant: 'bodyStandard',
       style: {
-        borderStyle: 'dotted',
-        borderWidth: 1,
+        borderStyle: 'solid',
```
tkamireh, line 48: are these values coming from design?

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
FalseLobster: I refactored it in my PR to make it more general, we can touch it up later though.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
FalseLobster: You mean visually?

Can you just drive the visuals off hovered state instead of the focus state then? (By moving the tokens you're currently setting in settings for focused to the hovered override) ?

FalseLobster (approved):

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
lenahong: It doesn't seem to be a visual only issue. The main issue is when the mouse hovers off the menu, actual focus remains on the last hovered/focused menu item. When I press space bar or enter, the menu item gets invoked.

PR #495, opened 2020-09-28
Create test-tools project with mock theme and some component tests

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds a new package under utils called test-tools. This is a non-published, devDependency only, package for component testing. Right now it only contains a few things:

#### mock theme

This is just a fully specified theme that is exported and can be pulled into component tests if needed. There were a couple of locations where the theme was specified in tests so they were changed to point at the mock theme. This also allows some portions of the tests to be changed in one location.

#### enzyme component tests

This change also creates two related component tests that components we build can use to validate their correctness. They are `checkRenderConsistency` and `checkReRender`. The differences and similarities are as follows:

Scenario | `checkRenderConsistency` | `checkReRender`
--- | --- | ---
Scenario | Ensure that two identical instances of a component share as much as possible | Ensure that re-render doesn't cause unnecessary tree mutations 
Description | Calls the passed in render function twice and compares the results | Calls the render function, then triggers a tree update and re-render, and compares the results
Props | Props at each level are shallow compared using === | Props at each level are shallow compared using ===
Children | Ignored in the comparison | Ignored in the comparison
Functions | Ignored. Typically these might be built via useEffect hooks and will be unique per instance | Compared. Re-render should not rebuild functions passed to child components
Depth | 1 by default, set to 2 or more for HOCs. | 1 by default, set to 2 or more for HOCs

### Verification

Just tests, since this is a change to add tests

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/utils/test-tools/src/mockTheme.ts
```diff
@@ -0,0 +1,189 @@
+import { Theme, ColorRamp } from '@fluentui-react-native/theme-types';
```
tkamireh, line 1: Presumably, this only copies what was being used in the tests? No additions or subtractions?

packages/utils/test-tools/src/enzymeTests.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function compareTrees(a: PropTreeSnapshot, b: PropTreeSnapshot, paths: string[]): void {
+  const newPaths = paths.concat(a.name);
+  if (a.name !== b.name) {
+    throw new Error(`Shallow compare found two nodes with different names at ${paths.join(': ')}`);
+  }
+  if (a.children.length !== b.children.length) {
+    throw new Error(`Shallow compare found two nodes at ${paths.join(': ')} with different property counts`);
+  }
+  Object.keys(a.props).forEach((key) => {
```
tkamireh, line 46: for full correctness, don't you also need to check that all of `b.props` are covered by `a.props`?

packages/utils/test-tools/src/enzymeTests.ts
```diff
@@ -0,0 +1,71 @@
...
+  const t1 = snapshotPropTree(mount(render()), depth, filter);
+  const t2 = snapshotPropTree(mount(render()), depth, filter);
+  compareTrees(t1, t2, []);
+}
+
+export function checkReRender(render: JSXProducer, depth: number = 1) {
+  const filter = { children: true };
+  const w1 = mount(render());
+  const t1 = snapshotPropTree(w1, depth, filter);
+  w1.setProps({});
```
tkamireh, line 67: What does the call to `setProps` do? Is it passing in an empty set of props? Or is it just syntax for recreating your original set?

packages/utils/test-tools/src/mockTheme.ts
JasonVMo: Yep

packages/utils/test-tools/src/enzymeTests.ts
JasonVMo: Yep, good catch. I had the check in a previous iteration but lost it when I added the tree snapshotting. I will add a test for Object.keys(a.props).length === Object.keys(b.props).length to ensure things are covered

packages/utils/test-tools/src/enzymeTests.ts
JasonVMo: Neither, it just makes sure that update will do something. In essence it just recreates a hidden top-level props module. There are discussions around this in the enzyme community and this is the accepted way of forcing this to happen as far as I can tell.

packages/utils/test-tools/src/enzymeTests.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+describe('enzyme component test validation', () => {
+  it('renders the simple control twice', () => {
+    checkRenderConsistency(() => <Simple>Hello</Simple>);
+  });
+
+  it('recurses into multi-level control correctly', () => {
+    checkRenderConsistency(() => <MultiLevel>World</MultiLevel>, 2);
+  });
+
+  it('handles memoed functions', () => {
```
FalseLobster, line 66: I really like the idea behind these test helpers!  Looking forward to seeing them get more use on our actual components.

FalseLobster (approved):

PR #499, opened 2020-09-29
Move Separator component to new framework

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This moves the separator component to use the new framework. Functionality should be unchanged though the usage of customize will have a slightly different signature.

Given how concise the component is I just moved it all to be a single file, types excepted.

### Verification

I ran through automated tests, then ran the win32 tester side by side with a tester with no changes.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Separator/src/Separator.tsx
```diff
@@ -1,23 +1,31 @@
...
+export const Separator = compose<SeparatorType>({
   displayName: separatorName,
-  settings: settings,
-  slots: {
-    root: {
-      slotType: View,
-      filter: filterViewProps
-    }
+  tokens: [{ separatorWidth: 1 }, separatorName],
+  tokensThatAreAlsoProps: ['vertical'],
```
FalseLobster, line 12: what does tokensThatAreAlsoProps do under the hood?

packages/components/Separator/src/Separator.types.ts
```diff
@@ -1,26 +1,27 @@
...
   /**
    * Specifies the width of the separator. This will be interpreted as DIPs on Windows and Android, but as points on Mac and iOS.
    * @defaultValue 1
    */
   separatorWidth?: number;
+
+  /**
+   * Specifies the color of the separator
+   */
+  color?: string;
```
FalseLobster, line 21: Did you see the tokenset stuff the web folks were doing, out of curiosity?

apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -128,10 +128,8 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
   const TestComponent = selectedTestIndex == -1 ? EmptyComponent : sortedTestComponents[selectedTestIndex].component;
 
   const TestListSeparator = Separator.customize({
-    tokens: {
-      color: useTheme().colors.inputBorder,
-      separatorWidth: 2,
-    },
+    color: useTheme().colors.inputBorder,
+    separatorWidth: 2,
```
FalseLobster, line 132: I know you didn't write this, but its a little weird to be doing this in a render function :)

FalseLobster (approved):

packages/components/Separator/src/Separator.tsx
JasonVMo: The main function is that it controls which props are copied into tokens before styles are prepared. So in this case the tokens will come from the theme + defaults, including a value for vertical. But if vertical is specified in props it will be copied into tokens before being handed to the slotProps functions.

The more subtle function is that the buildProps helper will actually filter the keys based on the intersection of this and its own list. So if you have `buildProps(tokens => { /* build props here */ }, ['vertical', 'horizontal'])` and only vertical is listed, the keys will be reduced to only be vertical. This is because the tokens before props are applied are already keyed on the theme, so only the ones that may be changed via props need to be used as memo keys.

packages/components/Separator/src/Separator.types.ts
JasonVMo: I just looked at it now, it looks like there are two things if I'm reading it right:

- a naming convention for the equivalent of our things like `FontTokens` where they should be named `FontTokenSets` which sounds reasonable and something we should align on.
- the ability to override tokens via props. I'm less aligned on this one. My read is that this is happening because of the move away from having compose and customize as options. 

For the second one we could handle this, with the memo stuff we do it could be blended as an object reference fairly easily. But it seems like something that can be added later if necessary. For now I would hold off.

For the first one with the naming we should snap to that. I'm also temped to move the types into theme-types as they are sort of global theme/customization fragments.

JasonVMo: Note that in this case we had IForegroundColorTokens which always just contained color?: string and I put it in specifically to make it more readable, but maybe it is better to keep these composed of parts.

apps/fluent-tester/src/FluentTester/FluentTester.tsx
JasonVMo: Ugh, didn't notice this, I will fix it as this is wrong and I'm surprised it isn't crashing

JasonVMo: I guess it won't crash because useContext isn't a memo based hook, so it doesn't create storage. This gets called out of tree though, which means it will always use the initialized value, which is undefined, so I would expect that this always ends up hitting the fallback theme.

PR #500, opened 2020-09-29
FocusZone/RadioGroup Functionality

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- Added FocusZone as slot to RadioGroup to give circular navigation functionality
- Added FocusZone E2E Test
- Cleaned up FocusZone/RadioGroup test pages
- Removing RadioGroup E2E UWP test because FocusZone is not compatible with UWP

### Verification

Manual/Automated

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/FocusZoneTest.tsx
```diff
@@ -11,90 +11,68 @@ const directionalFocusZone: React.FunctionComponent<{}> = () => {
   return (
     <View>
       <Button content="Outside FocusZone" />
+      <Text variant="headerSemibold" accessible={false}>FocusZone with Bidirectional arrow key navigation</Text>
```
ksiler, line 14: What's the reason for adding `accessible={false}` to these text headers? Does removing it affect the test in any way? Having text be accessible shouldn't give focus to it when assistive technology (like narrator) isn't being used. If it is being used, I think we'd want narrator to focus on and read these headings.

samuelfreiberg: I was thinking it would take focus for some reason, brain fart :) Removing now

ksiler: No worries! It's not an uncommon thought. I've seen others use it as such before. Thanks for fixing it!

JasonVMo (approved):

samuelfreiberg:
@JasonVMo 
I've just discovered a potential problem with adding FocusZone as a slot to RadioGroup in this way. Currently, nested FocusZones are not properly handled. Having a FocusZone as a slot for RadioGroup would mean the consumer can't use another FocusZone to wrap the control to add additional functionality (such as disallowing certain navigation keys - a FocusZone possibility).

Do we envision this could be an issue? Should I expose the FocusZone props on the RadioGroupProps interface? This would allow the consumer to feed in FocusZone props as well that I could use on the FocusZone slot.

ie. `IRadioGroupProps extends FocusZoneProps {...}`

FalseLobster:
> @JasonVMo
> I've just discovered a potential problem with adding FocusZone as a slot to RadioGroup in this way. Currently, nested FocusZones are not properly handled. Having a FocusZone as a slot for RadioGroup would mean the consumer can't use another FocusZone to wrap the control to add additional functionality (such as disallowing certain navigation keys - a FocusZone possibility).
> 
> Do we envision this could be an issue? Should I expose the FocusZone props on the RadioGroupProps interface? This would allow the consumer to feed in FocusZone props as well that I could use on the FocusZone slot.
> 
> ie. `IRadioGroupProps extends FocusZoneProps {...}`

RadioGroup's keyboarding behavior is a subset of FocusZones.  I think we should leave FocusZone an implementation detail with a specific FocusZone configuration (and not expose the FocusZone API surface) until a partner justifies "why" to us.  Most people will want RadioGroup to "just work" --- wrap, arrow nav, with tab out --- and not need to customize this behavior.

The only thing that might be worth considering now is the behavior of the disabled state, IMO, since the RadioGroup should be disable-able and there might be interactions there.

FalseLobster:
I guess that works

PR #501, opened 2020-09-30
Add more button variants

FalseLobster:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Moves our Button closer to the converged v8 FluentUI button

- Stealth/Primary Button are no longer independent types and just a different variant with special tokens you can flip with a bool.  I.e. `<Button primary />` instead of `<PrimaryButton />`
- Renamed Stealth -> Ghost
- Added a token variant for `Fluid` which sets width to 100%
- I removed the inner Stack slot because there wasn't one in the converged component and it would interfere with someone trying to set `style={{ padding: 3 }}` if they wanted to
- Added a CompoundButton that has a sublabel
- Added a ToggleButton
  - Refactored useAsToggle as a shared hook
  - Generalized useKeyCallback and applied it in more places.
  - **Note**that the Win32 theming module doesn't have enough color information to create Compound/ToggleButtons of each color variant, so some of this will need to be driven by Theme
- The old theme object didn't have enough semantic colors to describe some of the variants and was doing things like mapping Ghost button to MenuItem colors as a result, this explictly adds new semantic colors for those variants and tries to adopt a similar naming schema I was seeing in the FluentUI v8.  Ideally these would be tool driven by the token pipeline but I don't think we should block more accurate colors on that work.
- I published the experimental Button separately from the @fluentui/react-native library to use it in the tester which is now being published.  I'm not sure we're ready to drop it in the core library yet, so maybe we need to find a way to inject new pages into the tester that aren't published if we don't want to publish the experimental button separately.
- I wrote a shared useButton hook for CompoundButton and regular Button to share that bundles some of the other hooks together.

**NOTE** some of this stuff is a little rough still but I wanted to start collecting feedback.

### Verification

newButtons

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/Button.styling.ts
```diff
@@ -1,65 +1,31 @@
...
   slotProps: {
     root: buildProps(
       (tokens: ButtonTokens, theme: Theme) => ({
         style: {
           display: 'flex',
-          alignItems: 'flex-start',
+          alignItems: 'center',
           flexDirection: 'row',
           alignSelf: 'flex-start',
+          justifyContent: 'center',
```
JasonVMo, line 17: I do wonder if some of these minHeight/Width or padding values should be tokens as well. Certainly the gap between the icon and content likely should be configurable.

packages/experimental/Button/src/CompoundButton/CompoundButton.styling.ts
```diff
@@ -0,0 +1,82 @@
...
+          },
+        },
+      } as ButtonTokens,
+    }),
+    defaultButtonTokens,
+    compoundButtonName,
+  ],
+  states: buttonStates,
+  slotProps: {
+    root: buildProps(
```
JasonVMo, line 31: I wonder if some of these could be shared with button. It seems like much of this is somewhat duplicated. Maybe something like the way borderStyles or fontStyles are defined.

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    secondaryContent: Text,
+    contentContainer: View,
+  },
+  filters: {
+    icon: filterImageProps,
+  },
+  render: (userProps: CompoundButtonProps, useSlots: UseSlots<CompoundButtonType>) => {
+    const { icon, content, secondaryContent, ...rest } = userProps;
+
+    const button = useButton(rest);
```
JasonVMo, line 27: This is a nice simplification.

packages/experimental/Button/src/CompoundButton/CompoundButton.types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface CompoundButtonTokens extends ButtonTokens {
+  secondaryContentTokens?: FontTokens & { color?: string };
+}
+
+export interface CompoundButtonProps extends ButtonProps {
+  /**
+   * Second line of text that describes the action this button takes.
+   */
+  secondaryContent?: string;
```
JasonVMo, line 43: Should this maybe be called caption?

JasonVMo: I also wonder if this could be implemented as recomposing Button, with replacing the content slot and the render function. Right now that might be a bit tricky but worth considering.

JasonVMo (commented):
Still looking through things but publishing some comments until I get through everything.

packages/experimental/Button/src/useButton.ts
```diff
@@ -0,0 +1,23 @@
+import { useAsPressable, useKeyCallback, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import { ButtonProps, ButtonState } from './Button.types';
+
+export const useButton = (props: ButtonProps): ButtonState => {
```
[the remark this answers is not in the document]
JasonVMo: I would also decouple this from the Button package, if we:

- make the input props and returned props ViewProps. This will help with setting to the internal nodes as well.
- don't use the ButtonState object, just return a tuple of [ViewProps, PressableState]
- then this could also be put in interactive hooks

JasonVMo (commented):
More comments for review part 2/3. Publishing before heading to a block of meetings.

packages/experimental/Button/src/CompoundButton/CompoundButton.test.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import { CompoundButton } from './CompoundButton';
+import * as renderer from 'react-test-renderer';
+
+it('Button default', () => {
```
JasonVMo, line 5: Probably should be CompoundButton default

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    icon: Image,
+    content: Text,
+    secondaryContent: Text,
+    contentContainer: View,
+  },
+  filters: {
+    icon: filterImageProps,
+  },
+  render: (userProps: CompoundButtonProps, useSlots: UseSlots<CompoundButtonType>) => {
+    const { icon, content, secondaryContent, ...rest } = userProps;
```
JasonVMo, line 25: I wonder if these should be split out in the returned closure. By that I mean something like:

```ts
const { icon, content, secondaryContent, ...merged } = mergeProps(button.props, final);
```

packages/experimental/Button/src/ToggleButton/ToggleButton.test.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import { ToggleButton } from './ToggleButton';
+import * as renderer from 'react-test-renderer';
+
+it('Button default', () => {
```
JasonVMo, line 5: TottleButton default

JasonVMo (approved):
Some things to look at but generally these changes look great.

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
JasonVMo: This is to ensure it can be set by a component in JSX if this component is being driven as part of a larger component.

packages/experimental/Button/src/ToggleButton/ToggleButton.test.tsx
FalseLobster: TottleButton has to be one of my favorite typos I've seen recently

packages/experimental/Button/src/Button.styling.ts
FalseLobster: I went ahead and added minHeight/Width since they were straightforward.  I'm punting on Padding for now because it's a little less obvious how we should expose it in the token interface.

packages/experimental/Button/src/CompoundButton/CompoundButton.types.ts
FalseLobster: I went with secondaryContent because that's what the [converged web button](https://github.com/microsoft/fluentui/blob/master/packages/react-button/src/components/CompoundButton/CompoundButton.types.ts) calls it.

FalseLobster: I see.  Are you suggesting I might take the contentContainer (View around content and secondaryContent) and put that in the  content slot of the base Button?

packages/experimental/Button/src/useButton.ts
```diff
@@ -0,0 +1,23 @@
...
+  // attach the pressable state handlers
+  const { onClick, ...rest } = props;
+  const pressable = useAsPressable({ ...rest, onPress: onClick });
+  const onKeyUp = useKeyCallback(onClick, ' ', 'Enter');
+
+  return {
+    props: {
+      ...pressable.props,
+      accessible: true,
```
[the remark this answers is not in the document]
FalseLobster: Agreed

packages/experimental/Button/src/useButton.ts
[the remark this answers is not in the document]
FalseLobster: I like this idea but I'm not sure it's quite as simple as you're suggesting if we want to get the typings right, so I'm going to punt on this for now.

FalseLobster:
azp /run

packages/experimental/Button/src/CompoundButton/CompoundButton.types.ts
JasonVMo: Effectively yes, though I don't know whether it is a net gain.

JasonVMo (approved):

PR #502, opened 2020-10-02
Update checkbox to use new framework

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This moves the checkbox component onto the new framework. The tokens are tweaked a small amount to add more customization options, the style portions of the old settings go into the style functions, as do the values from the old token arrays.

The main user change is for people who have customized checkbox. In this case the format has changed from:

```ts
const MyCheckbox = Checkbox.customize({
  tokens: { /* token values */ },
  _overrides: {
    hovered: {
      tokens: { }
    }
  }
});
```

to:

```ts
const MyCheckbox = Checkbox.customize({
  // token values
  hovered: {
    // hovered token values
  }
}
```

#### Token Retrieval

Previously the tokens used to create the styles were never exposed to callers. This can be limiting as some values may need to be handled by the functions. The use-styling hook now takes an optional object which can both set a custom lookup value, and retrieve tokens.

#### Token Sets

The various token combinations have been moved from tokens into theme types and renamed to be TokenSets, this hopefully aligns better with what the web folks are doing.

### Verification

Automated tests, then running through the win32 tester, then looking at the web tester. It looks like the web tester was broken by a recent change so I fixed that. On the web the checkboxes look correct but aren't clickable, which I believe was an issue before. I'll look into that more.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Checkbox/src/Checkbox.types.ts
```diff
@@ -19,14 +16,24 @@ export interface ICheckboxState extends IPressableState {
...
+/**
+ * Values that can be set via props or via tokens
+ */
+export interface CheckboxTokenProps {
   /**
-   * Determines position of Checkbox. True if boxSide='end'
+   * Allows you to set the checkbox to be at the before (start) or after (end) the label
    */
-  boxAtEnd?: boolean;
+  boxSide?: 'start' | 'end';
```
samuelfreiberg, line 28: Out of curiosity, why make this a token instead of a prop?

JasonVMo: I actually made it both a token and a prop. This allows someone to change the default to boxSide: 'end' if they want.

JasonVMo: Think of it as a token, that happens to be overridable via props. The value below, `tokensThatAreAlsoProps` determines which values get overridden from props.

samuelfreiberg (approved):

tkamireh:
@JasonVMo one question I have is whether we worry at all about inconsistency in the exposed APIs in controls that are shipped from core that use the experimental frameworks and the ones that don't? The API shapes exposed on the control are slightly different if I understand correctly.

Also this has been lingering for a bit 😎, any reason not to merge (depending on above concern) or just close it?

samuelfreiberg:
Is there an update on this? Should this PR be closed or merged? @JasonVMo

JasonVMo:
This is ancient so I'm closing it

PR #506, opened 2020-10-04
Tester themes

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This removes some dependencies on theming-react-native that didn't need to be present. At this point I believe it has been pretty much pruned from everywhere. To do this I moved useTheme up to be next to the ThemeContext definition and then rerouted references appropriately.

This also moves the tester from using the theme-registry + theming-react-native + theming-ramp over to ThemeReferences.

It also reworks the theme picker bar to have a more predictable set of selections. Part of this was removal of platform (which didn't work), it can be re-added later. Now though you can set it to use the base Fluent theme, the Office theme overlay, or a custom Caterpillar theme. Set light/dark mode, and apply a brand color to whatever theme you are in. Each of these overlays can be extended in the future.

This also publishes the experimental-text and button components so they can be pulled into the tester.

### Verification

Lots of work in the web tester to make sure things look right. I need to do a final pass on win32 but my VPN is not working at the moment.

Update 10/4: I looked over everything on win32 and it looks good.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Makes the tester on iPad test app cleaner, thanks! Also love the addition of the fluent theme with light/dark support, I'll have to update my components

Saadnajmi:
I got the following warning on my packager, just an fyi
```
[Wed Oct 07 2020 12:09:13.878]  WARN     Require cycle: ../fluent-tester/src/FluentTester/theme/CustomThemes.ts -> ../fluent-tester/src/FluentTester/theme/applyBrand.ts -> ../fluent-tester/src/FluentTester/theme/CustomThemes.ts

Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
[Wed Oct 07 2020 12:09:13.880]  WARN     Require cycle: ../fluent-tester/src/FluentTester/theme/CustomThemes.ts -> ../fluent-tester/src/FluentTester/theme/applyTheme.ts -> ../fluent-tester/src/FluentTester/theme/CustomThemes.ts
```

Saadnajmi (approved):

apps/fluent-tester/package.json
```diff
@@ -21,13 +21,15 @@
...
+    "@fluentui-react-native/experimental-text": ">=0.1.0 <1.0.0",
     "@fluentui-react-native/framework": ">=0.3.7 <1.0.0",
     "@fluentui-react-native/stack": ">=0.2.16 <1.0.0",
     "@fluentui-react-native/interactive-hooks": ">=0.5.2 <1.0.0",
-    "@uifabricshared/theming-react-native": ">=0.8.2 <1.0.0",
-    "@uifabricshared/theme-registry": ">=0.3.71 <1.0.0",
+    "@fluentui-react-native/default-theme": ">=0.2.1 <1.0.0",
+    "@fluentui-react-native/theme": ">=0.2.1 <1.0.0",
+    "@fluentui-react-native/theme-types": ">=0.2.1 <1.0.0",
+    "@fluentui-react-native/win32-theme": ">=0.3.1 <1.0.0",
```
HeyImChris, line 31: Why isn't it a problem that this shared package.json has win specific dependencies?

apps/fluent-tester/src/FluentTester/FluentTesterApp.tsx
```diff
@@ -1,18 +1,16 @@
...
 import { FluentTester, FluentTesterProps } from './FluentTester';
-import { customRegistry, registerThemes } from './CustomThemes';
 import { tests } from './testPages';
-
-registerThemes();
+import { testerTheme } from './theme/index';
 
 export const FluentTesterApp: React.FunctionComponent<FluentTesterProps> = (props) => {
-  const [theme, setTheme] = React.useState('');
+  // const [theme, setTheme] = React.useState('');
```
HeyImChris, line 10: delete?

apps/fluent-tester/src/FluentTester/TestComponents/Theme/ThemeTest.tsx
```diff
@@ -92,7 +90,7 @@ const SwatchList: React.FunctionComponent<{ hostKey: string }> = ({ hostKey }: {
   }, []);
   return (
     <View style={[commonTestStyles.view]}>
-      <Text variant="bodySemibold">getHostSettingsWin32(theme: ITheme).{hostKey}</Text>
+      <Text>getHostSettingsWin32(theme: ITheme).palette</Text>
```
HeyImChris, line 93: We haven't done the theming work on Apple platforms yet, but will this not eventually be a problem having a win32 specific call in a shared file?

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -0,0 +1,102 @@
+import { testerTheme } from './CustomThemes';
+
+export type OfficeBrand = 'Default' | 'Office' | 'Word' | 'Excel' | 'Powerpoint' | 'Outlook';
+type BrandRampKey = 'App1' | 'App2' | 'App3' | 'App4' | 'App5' | 'App6' | 'App7' | 'App8';
+type BrandRamps = { [K in OfficeBrand]: { [J in BrandRampKey]: string } };
+
+const brandColors: BrandRamps = {
```
HeyImChris, line 7: Why should a FluentUI repo know anything about the consumer Office apps? That feels like we're almost reversing the dependency chain by this repo knowing about consts the consumers should define

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * default brand just passes through what is in the theme
+   * */
+  Default: undefined,
+  /**
+   * Office brands, applied via a small App color ramp
```
HeyImChris, line 13: Why are we prioritizing Office brands when this has plenty of non-Office clients? Are all clients expected to add their branding constants here...?

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export type ThemeNames = 'Fluent' | 'Office' | 'Caterpillar';
+
+export const themeChoices = [
+  { label: 'Fluent', value: 'Fluent' },
+  { label: 'Office', value: 'Office' },
+  { label: 'Caterpillar', value: 'Caterpillar' },
+];
+
+export function applyCaterpillarTheme(parent: Theme): PartialTheme {
```
HeyImChris, line 13: I saw you mention this in the description. What is a Caterpillar Theme?

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
```diff
@@ -0,0 +1,78 @@
...
+            },
+          },
+        },
+      };
+}
+
+let currentTheme: ThemeNames = 'Fluent';
+
+const themingModule = getThemingModule()[0];
+const officeBase = themingModule
```
HeyImChris, line 56: Same question as before that this seems odd to have Office specific code in an open source framework...

apps/fluent-tester/package.json
JasonVMo: win32-theme isn't 100% win32 specific. It is effectively a theme that maps the Office Palette structure to the Fluent theme structure. Now the office theming native module is only implemented on win32, so it ends up falling back to the JS only implementation for all other platforms.

I was tempted to call this office-theme, and long-term that may make sense, but I'm unclear on which platforms use the office palette and which don't, so for right now using the more scoped naming.

This change ultimately creates a hybrid theme in the tester where it can dynamically bring in layers. It starts with default-theme, optionally specifying light and dark, optionally applying the office theme as a layer, then optionally applying a brand. Long-term we can extend this with more options, this is a start.

apps/fluent-tester/src/FluentTester/TestComponents/Theme/ThemeTest.tsx
JasonVMo: See the comment above, it should work. Previously the mac control versions were always routing through the win32 native module fallback code. With the changes to foundation-compose and the theming system this is no longer the case.

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
JasonVMo: Keep in mind the tester is just test code. This is using the understood Office brands to demonstrate what applying a brand looks like to the components. You can image also adding additional themes / brands in the tester like Material, Spotify, Teams, etc.

Because win32 is exposed and is an Office only tech, some of the fact that we have Office concepts are going to leak into FLURN. As an example we have the shape of the host interface for the win32 palette in @fluentui-react-native/theme-types. Because typedefs have no actual code attached to them though this doesn't incur cost and makes the usage for many office clients easier, even if we don't expect many others to make use of it.

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
JasonVMo: This is just example code and is non-shipping. The idea (taken from Fluent) is for a company like Caterpillar (I believe a real Fluent/Fabric customer) to be able to inject their brand into the theming.

Keep in mind all of these things were already present in the tester, this just reorganizes them.

PR #507, opened 2020-10-05
Fix macOS test app, update react-native-test-app dependency

Saadnajmi:
The macOS test app would not automatically launch the packager when run. Furthermore, when I did run it, I got the logbox error:
`Unhandled JS Exception: ReferenceError: Can't find variable: FluentTester`

Comparing the macOS test app source files with the iOS test app source files helped figure out the problems.

### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

 - For some reason, the `launchPackager.command` and `packager.sh`files got deleted from the macOS test app directory. Add them back. They are copies of the iOS one with one string change: `react-native -> react-native-macos`
- Change the name key inside `apps/macos/app.json` from `"FluentTester-macos` to `"FluentTester`. This matches how we do it on the other platforms that use `react-native-test-app`.
- Update `apps/macos/src/index.tsx` to export `FluentTesterApp`, not `FluentTester`
- Update the `react-native-test-app` dependency in our projects to use the latest patch version. Development has been consistent in this repo and it would be easier if we just got the latest non-breaking changes more easily.

### Verification

Ran the macOS, iOS, and Android test apps to make sure they ran properly.

Screen Shot 2020-10-05 at 12 20 42 PM
Screen Shot 2020-10-05 at 12 38 49 PM
Screen Shot 2020-10-05 at 1 38 25 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/macos/src/index.tsx
+++ b/apps/macos/src/index.tsx
@@ -1,17 +1,8 @@
 'use strict';
 
 import {FluentTesterApp} from '@fluentui-react-native/tester';
-import * as React from 'react';
-import {AppRegistry, SafeAreaView} from 'react-native-macos';
+import {AppRegistry} from 'react-native-macos';
 
-const FluentTester: React.FunctionComponent = () => {
-  return (
-    <SafeAreaView>
-      <FluentTesterApp />
-    </SafeAreaView>
-  );
-};
+AppRegistry.registerComponent('FluentTester', () => FluentTesterApp);
 
-AppRegistry.registerComponent('FluentTester', () => FluentTester);
-
-export default FluentTester;
+export default FluentTesterApp;
--- a/apps/macos/src/index.tsx
+++ b/apps/macos/src/index.tsx
@@ -1,8 +1,17 @@
 'use strict';
 
 import {FluentTesterApp} from '@fluentui-react-native/tester';
-import {AppRegistry} from 'react-native-macos';
+import * as React from 'react';
+import {AppRegistry, SafeAreaView} from 'react-native-macos';
 
-AppRegistry.registerComponent('FluentTester', () => FluentTesterApp);
+const FluentTester: React.FunctionComponent = () => {
+  return (
+    <SafeAreaView>
+      <FluentTesterApp />
+    </SafeAreaView>
+  );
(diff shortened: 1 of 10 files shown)


apps/macos/src/index.tsx
```diff
@@ -1,8 +1,17 @@
...
 
 import {FluentTesterApp} from '@fluentui-react-native/tester';
-import {AppRegistry} from 'react-native-macos';
+import * as React from 'react';
+import {AppRegistry, SafeAreaView} from 'react-native-macos';
 
-AppRegistry.registerComponent('FluentTester', () => FluentTesterApp);
+const FluentTester: React.FunctionComponent = () => {
+  return (
+    <SafeAreaView>
```
tido64, line 9: I don't think we need `SafeAreaView` for macOS.

tido64 (approved):
Looks fine to me.

HeyImChris (approved):
Thanks for fixing this!

apps/macos/src/index.tsx
```diff
@@ -1,17 +1,8 @@
 'use strict';
 
 import {FluentTesterApp} from '@fluentui-react-native/tester';
-import * as React from 'react';
-import {AppRegistry, SafeAreaView} from 'react-native-macos';
```
anandrajeswaran: Fun note: Big Sur just added safe areas to macOS as a better means of layout relative to a title bar than the current solution of the layout is either full screen or cannot underlap the title bar. Still the right change here, but we may need to re-evaluate as we look at this with the 11.0 SDK

Saadnajmi: I remember that! I vaugely remember Safe Area's being a new thing. I basically had two thoughts:

1) Leave it in as the extra JSX doesn't do anything right now and I'm sure SafeAreaView isn't implemented for react-native-macOS yet. This is the futureproof approach

2) Remove it and add it when it becomes a necessity.

As you can see, I've gone between both :D

PR #510, opened 2020-10-07
Fixing redbox on macOS from theming module

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

macOS doesn't currently provide a theme (the theming support on macOS is limited).

I think with commit 59d0857d4244fb6b35fb2c9ca7d01cf2a073f461 we can now end up using the spread operator on a string to construct an object, which seems unintentional, since the following code

`Object.assign(translateOfficeTheme(themingModule, paletteCache, themeId), { name: _hostTheme })`

changed to

`{ /*other stuff*/, ...(_hostTheme && { name: _hostTheme }) }`

which is problematic because the empty string '' is falsy -- so instead of skipping the name field, we just end up trying to spread the empty string and trying to construct an object out of it, which the Object.es6.js polyfill is very unhappy about. This isn't a problem on Win32 because the theme is not falsy (i.e. not empty) on Win32.

### Verification

I've checked that we no longer redbox on macOS. Changes do not affect Win32 because the theme is never falsy on Win32 (which is why we never see it on Win32)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/win32-theme/src/createPartialOfficeTheme.ts
+++ b/packages/theming/win32-theme/src/createPartialOfficeTheme.ts
@@ -33,6 +33,6 @@ export function createPartialOfficeTheme(module: OfficeThemingModule, themeName?
       // Office Branding Colors
       colors: getRamps(module),
     },
-    ...(themeName && { name: themeName }),
+    ...(!!themeName ? { name: themeName } : undefined),
   };
 }
--- a/packages/theming/win32-theme/src/createPartialOfficeTheme.ts
+++ b/packages/theming/win32-theme/src/createPartialOfficeTheme.ts
@@ -33,6 +33,6 @@ export function createPartialOfficeTheme(module: OfficeThemingModule, themeName?
       // Office Branding Colors
       colors: getRamps(module),
     },
-    ...(!!themeName ? { name: themeName } : undefined),
+    ...(themeName ? { name: themeName } : undefined),
   };
 }
--- a/change/@fluentui-react-native-win32-theme-2020-10-06-15-01-34-nakambo-mac-redbox-theming-fix.json
+++ b/change/@fluentui-react-native-win32-theme-2020-10-06-15-01-34-nakambo-mac-redbox-theming-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fixing redbox on macOS from theming module",
+  "packageName": "@fluentui-react-native/win32-theme",
+  "email": "nakambo@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-10-06T22:01:33.996Z"
+}


packages/theming/win32-theme/src/createPartialOfficeTheme.ts
```diff
@@ -33,6 +33,6 @@ export function createPartialOfficeTheme(module: OfficeThemingModule, themeName?
       // Office Branding Colors
       colors: getRamps(module),
     },
-    ...(themeName && { name: themeName }),
+    ...(!!themeName ? { name: themeName } : undefined),
```
HeyImChris, line 36: Doesn't seem right that macOS is using the `win32-theme` file

nakambo: I agree, but it looks tied in with simply using the Fluent Button:

```
createPartialOfficeTheme (index.js:formatted:523)
translateOfficeTheme (index.js:formatted:607)
getPlatformDefaults (index.js:formatted:621)
createPlatformThemeRegistry (index.js:formatted:828)
getThemeRegistry (index.js:formatted:846)
getTheme (index.js:formatted:891)
useStylingCore (index.js:formatted:1188)
(anonymous) (index.js:formatted:1213)
usePrepareProps (Button.tsx?4e81:49)
useUpdateRenderData (Composable.slots.ts?63de:87)
useCompoundPrepare (Composable.slots.ts?63de:120)
Button (Composable.ts?a5ad:99)
renderWithHooks (ReactNativeRenderer-dev.js?2a1c:10611)
```

At the moment, I'm just trying to unblock newer versions of Fluent on macOS -- from what I can tell, it's always been this way (even before the commit I indicated in my notes earlier). Folks can refactor\rework things as they see fit separately to remove this dependency.

Saadnajmi: Yeah, I think he only "theme" provided right now is the win32 / office theme, and we just haven't built out platform specific / other themes yet. Definitely should be on our list of things to do, would help with making the test apps look/feel more native.

nakambo: Should we not go ahead with the current change that restores previous behavior (prior to commit 59d0857d4244fb6b35fb2c9ca7d01cf2a073f461)? If not, what should I be doing?

As I mentioned, macOS is currently blocked on adopting newer versions of Fluent. I need the same SDX on this newer version of Fluent because I need the capabilities exposed therein on Win32. Fixing the block (which was introduced two weeks ago) feels more pressing to me than fixing a problem that has probably existed since a while (months, I'll guess).

Saadnajmi: It seems fine to me, but I'm less familiar with this part of the code and wanted someone more familiar to review. The reviewers you initially added should be good.

JasonVMo (approved):

packages/theming/win32-theme/src/createPartialOfficeTheme.ts
JasonVMo: The change looks good and the win32-theme dependency is something I/we are in the process of removing. I think one of the PRs I have out will actually remove this dependency. It was from theming-react-native containing the native module and being referenced from foundation-compose I believe. I think I changed the framework dependencies to only be theme-types and more generic packages, then made the win32/office theming parts optional in the tester.

PR #511, opened 2020-10-08
Use @react-native-community/picker to silence Logbox warning

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

We've had a persistent logbox error telling us to use the ejected module "@react-native-community/picker" instead of the one in react-native core. This PR is to address that.

Note: I had to cast a lot of objects to string to use the new picker
Note 2: There is a bug in the macOS implementation of RNCPicker. I created [this PR](https://github.com/react-native-community/react-native-picker/pull/144) to fix that, and we will have to wait till that goes in before we can use this.

### Verification

Ran the test app on iOS and macOS to make sure the test app still works.
It is probably worth me running it on all platforms as a sanity check. Will do so

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/src/Podfile
```diff
@@ -29,6 +29,7 @@ use_test_app! do |target|
   target.app do
 
     pod 'FluentUI-React-Native-Shimmer', :path => '../../../packages/components/Shimmer/FluentUIReactNativeShimmer.podspec'
+    pod 'RNCPicker', :path => '../../../node_modules/@react-native-community/picker/RNCPicker.podspec'
```
Saadnajmi, line 32: note: We really shouldn't need to do this and I need to figure out why it isn't just auto adding native modules.

apps/macos/src/Podfile
```diff
@@ -27,6 +27,9 @@ fi
 use_flipper!(false)
 use_test_app! do |target|
   target.app do
+
+    pod 'RNCPicker', :path => '../../../node_modules/@react-native-community/picker/RNCPicker.podspec'
```
Saadnajmi, line 31: Same as comment in the iOS Podfile. We shouldn't need to do this and we should figure out why autolinking isn't working.

NickGerleman:
I think this might break Win32 as is if it was using this before. Community modules won't have a view manager for NetUI, and generally use a different class name than the builtin (sometimes a different interface too).

Saadnajmi:
Closing for now. I have an idea for how to continue this in a way that works for all our platforms (create a wrapper Picker control that uses the win32 version for win32, community version for other platforms, and something less annoying than the full picker for iOS). Will probably revisit this for my team's hackathon next week =)

PR #514, opened 2020-10-15
Shimmer Static Library + Autolinking in test app (sorta)

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There are two main problems being addressed here:

1) We need Shimmer to be built into a static library. That built library will be included in our nuget package that we consume internally in the office codebase. This is achieved by adding an xcodeproj to the Shimmer component. This will let us build a static library "libFRNShimmer.a" that is included / built via cocoapods.

2) Autolinking was not working in our iOS test app to get the native Shimmer files. After lots of debugging, I determine the cause to be this:

Autolinking will include the native dependencies of packages listed in the app's package.json. It will NOT include the native dependencies of the dependencies of packages in the package.json.

Let me explain:
"fluent-tester-ios" (Our iOS test app) depends on "fluent-tester" (which contains the JS and some native modules), which depends on "fluentui-react-native/shimmer" (A Native Module with swift/objc code).
This does not autolink shimmer into fluent-tester-ios

If we add "fluentui-react-native/shimmer" into "fluent-tester-ios"''s package.json, autolinking then works.

Note: I feel like this is a bug, and we shouldn't need to do this. A package should be able to import it's dependencies' native dependencies. I'm not sure why we are seeing this behavior.

### Verification

Ran `pod install` and made sure I saw FRNShimmer get included 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Shimmer/ios/FRNShimmer.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,309 @@
...
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
+				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
+				MTL_FAST_MATH = YES;
+				ONLY_ACTIVE_ARCH = YES;
+				SDKROOT = iphoneos;
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
anandrajeswaran, line 185: Can we avoid putting build settings in pbxprojs where they are duplicated per config/flavor? We should extract out to an xcconfig file(s)

anandrajeswaran (changes requested):

packages/components/Shimmer/ios/FRNShimmer.xcodeproj/project.pbxproj
Saadnajmi: I knew that would be first :D, and yeah sure.

Saadnajmi:
Ideally I would like it to just generate the `libFRNShimmer.a` from just the `.podspec` and not need an xcodeproj. Till now I have not seen a way to do that =(

packages/components/Shimmer/ios/FRNShimmer.xcodeproj/project.pbxproj
Saadnajmi: Turns out I don't need the xcodeproj at all. Deleting

Saadnajmi: @anandrajeswaran I can remove the xcodeproj from here and grab the static library from the test app directory, where we already build it for CI. I talked to @tom-un and found out we do this for other projects as well, so it's an accepted pattern

Saadnajmi:
Making autolinking sorta work means depcheck fails.
That + the fact that we don't need an xcodeproj means I'll close this review and do the smaller changes in a separate review

PR #517, opened 2020-10-15
RadioButton Focus Rect Bug

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

- Before, the focus rect on a RadioButton was visible on render and whenever the RadioButton was selected (regardless if it was focused or not). Now, the focus rect is only visible when the RadioButton is focused

- Added componentRef prop to the RadioButton 

### Verification

Manually + Automated tests

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/RadioGroup/src/RadioButton.settings.ts
```diff
@@ -61,23 +62,22 @@ export const settings: IComposeSettings<IRadioButtonType> = [
...
         marginTop: 3,
         borderStyle: 'dashed',
-        borderColor: 'transparent',
         borderWidth: 1
       }
     },
-    // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out'
-    _precedence: ['disabled', 'hovered', 'selected'],
+    // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out
+    _precedence: ['disabled', 'hovered', 'isFocused', 'selected'],
```
lenahong, line 69: Rename this to 'focused' to keep the naming convention the same across our components?

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -9,21 +9,27 @@ import { filterViewProps } from '@fluentui-react-native/adapters';
...
 export const RadioButton = compose<IRadioButtonType>({
   displayName: radioButtonName,
 
   usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
-    const { content, buttonKey, disabled, ariaLabel, ...rest } = userProps;
+    const { content, buttonKey, disabled, ariaLabel, componentRef, ...rest } = userProps;
+
+    // Since we're overriding pressable's onFocus below, pressable.state.focused does not get updated.
+    // We need to track whether the RadioButton is focused or not manually.
+    const [isButtonFocused, setIsFocused] = React.useState(false);
```
lenahong, line 24: Did you consider using our existing useFocusState hook? I was wondering if you had a particular reason for doing it this way

packages/components/RadioGroup/src/RadioButton.settings.ts
```diff
@@ -61,23 +62,22 @@ export const settings: IComposeSettings<IRadioButtonType> = [
       style: {
         marginTop: 3,
         borderStyle: 'dashed',
-        borderColor: 'transparent',
         borderWidth: 1
       }
     },
-    // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out'
-    _precedence: ['disabled', 'hovered', 'selected'],
+    // This still uses the innerCircle's 'hover' style instead of the 'focused' style when need be. I'm commenting out
```
ksiler, line 68: I don't understand what this comment means, could you clarify?

packages/components/RadioGroup/src/RadioButton.settings.ts
samuelfreiberg: Changing this now. I initially thought there might be some negative interaction between pressable's 'focused' and this, but it looks to be fine.

packages/components/RadioGroup/src/RadioButton.settings.ts
samuelfreiberg: That looks like an old comment that isn't relevant anymore. Good catch, I'm removing it now

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: I was thinking since I already had to create a custom function for the root's onFocus(), I might as well just do this. And to be honest, I wasn't aware of the useFocusState hook. I just tried to implement it, but it broke it (every radiobutton had the focus rect around it). I think overriding the onFocus() is causing problems.

Would it be sufficient to leave as is?

samuelfreiberg: So the reason this won't work is because I'm already setting an _onFocus_ method on the root. The useFocusState's onFocus will be overridden by this one, and thus the focus state won't get updated. That's why I had to modify my already-created onFocus method.

PPatBoyd: You wouldn't be able to leverage onStateChange with useFocusState?

samuelfreiberg: The pressable's focus state never changes because I'm overriding the onFocus method that would have been used for the pressable state change. But maybe I'm misunderstanding how onStateChange works? I'm assuming it's just a callback when there's a state change.

samuelfreiberg: So the problem I'm having is this:

I feed my onFocus function as a prop to useFocusState. However, my onFocus function references the state of the RadioButton, so I can't feed it as a prop to useFocusState because that gets defined before the state gets defined (because state relies on that).

Maybe there's an easy solution that I'm not seeing at the moment?

lenahong: I think you can implement what you want to do with this new state definition.

Instead of defining this new isButtonFocused state, you can just call pressable.props.onFocus() inside your onFocusChange callback. Pressable.props' onFocus and onBlur should be managing the focus state already so you wouldn't need to reassign basically same state callbacks.

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -54,13 +61,16 @@ export const RadioButton = compose<IRadioButtonType>({
       if (!state.selected) {
         info.onChange && info.onChange(buttonKey);
       }
+      setIsFocused(true);
```
lenahong, line 64: So basically what I'm saying is: 

`const onFocusChange = React.useCallback((ev: NativeSyntheticEvent<{}> ) => {
      if (!state.selected) {
        info.onChange && info.onChange(buttonKey);
      }
      pressable.props.onFocus && pressable.props.onFocus(ev);
    }, [state, pressable.props, info, buttonKey]);`

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -54,13 +61,16 @@ export const RadioButton = compose<IRadioButtonType>({
...
     }, [state, pressable.props, info, buttonKey]);
 
     const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
       root: {
         rest,
+        ref: buttonRef,
         ...pressable.props,
         onFocus: onFocusChange,
+        onBlur: () => setIsFocused(false),
```
lenahong, line 73: You wouldn't need this either since pressable.props.onblur should be changing the focused state anyway

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -40,6 +46,7 @@ export const RadioButton = compose<IRadioButtonType>({
       ...pressable.state,
       selected: info.selectedKey === userProps.buttonKey,
       disabled: disabled || false,
+      focused: isButtonFocused
```
lenahong, line 49: You can remove this as well

lenahong (changes requested):

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: I believe we tried this when I first developed RadioGroup and it didn't work (I can't remember why exactly). I'm trying it now though, but I'm getting an error here: `pressable.props.onFocus(ev)` stating mismatching parameters. Going to look into it more now. Is the NativeSyntheticEvent the correct event to be passing in?

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Thanks you, updated in newest commit

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Thank you for the advice, changed in newest commit

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Changed in newest commit

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Changed in newest commit

lenahong (approved):

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -96,7 +100,7 @@ export const RadioButton = compose<IRadioButtonType>({
     root: [],
     button: [borderTokens],
     innerCircle: [backgroundColorTokens],
-    content: [foregroundColorTokens, textTokens],
+    content: [foregroundColorTokens, textTokens, [{ source: 'textBorderColor', lookup: getPaletteFromTheme, target: 'borderColor' }]],
```
tkamireh, line 103: Do you need a width on the border? My guess would be that it defaults to 0, or is there something I'm missing?

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -49,16 +51,18 @@ export const RadioButton = compose<IRadioButtonType>({
     // 1) Calls pressable's onFocus in order to keep track of our state's focus variable. It is dependent on pressable's
     //    focus variable. Without this, it wouldn't stay updated because we're overriding it's onFocus below for the rootProps.
     // 2) Selects the currently focused button by calling the RadioGroup's callback function.
-    const onFocusChange = React.useCallback((/* ev: NativeSyntheticEvent<{}> */) => {
-      // This check is necessary because this func gets called even when a button loses focus (not sure why?) which then calls the client's onChange multiple times
+    const onFocusChange = React.useCallback((ev: NativeSyntheticEvent<{}>) => {
```
tkamireh, line 54: is `NativeSyntheticEvent<{}>` really the right type here?  The cast to `FocusEvent` signals to me that you should just use `FocusEvent` throughout. This is one of those weird things where we can't really make guarantees on the type with absolute certainty (maybe turbomodules with type generation in the future and the corresponding viewmanager changes make this safer).

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: That's in the settings file

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: The onFocus callback on ViewWin32.Props is: `onFocus?: (ev: RN.NativeSyntheticEvent<{}>) => void;`

PR #519, opened 2020-10-16
Adding selectedKey prop to RadioGroup

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

- Added the _selectedKey_ prop to RadioGroup. This allows the developer to set the selected key. When using this, the developer must maintain selection state by observing onChange events and passing a new value when change. This was a direct ask from a partner.

### Verification

Manual + Automation

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
```diff
@@ -10,29 +10,43 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
...
+        <RadioButton content="Option A" buttonKey="A" ariaLabel="Test Aria Label" />
         <RadioButton content="Option B" buttonKey="B" />
-        <RadioButton content="Option C" buttonKey="C" />
+        <RadioButton content="Option C (disabled)" buttonKey="C" disabled={true} />
         <RadioButton content="Option D" buttonKey="D" />
       </RadioGroup>
       <Separator />
-      <RadioGroup label="RadioGroup with B as defaultSelectedKey" defaultSelectedKey="B" onChange={onChange}>
-        <RadioButton content="Option A" buttonKey="A" />
+      <RadioGroup label="Controlled RadioGroup" selectedKey={selectedKey} defaultSelectedKey="D" onChange={onChange2}>
```
ksiler, line 42: nit: If `selectedKey` overrides `defaultSelectedKey`, is there a need for `defaultSelectedKey` here?

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
```diff
@@ -10,29 +10,43 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
     console.log(key);
   };
 
+  const [selectedKey, setSelectedKey] = React.useState('C');
+
+  const onChange2 = React.useCallback(
```
ksiler, line 15: nit: can fix or leave it as is but I think this example makes it hard to see the difference between a controlled and uncontrolled RadioGroup. Perhaps you can recreate the partner's scenario on a simple level to reflect its use? I'm not sure what the partner's scenario is here so that may or may not be an easy thing to do.

ksiler (approved):

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
samuelfreiberg: This was for testing purposes to ensure selectedKey did in fact override defaultSelectedKey like it should. I'll remove it in my newest commit.

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
samuelfreiberg: The partner's scenario is the following: Multiple people working on the same document in Excel. The value of the RadioGroup might change asynchronously. If someone changes the value of the RadioGroup, and another person has the RadioGroup visible at the same time, they need to be able to manually change the RadioGroup value. I'm not sure there's a good way to recreate this exact scenario.

There's no functional difference between controlled/uncontrolled from the user's POV, so I think it would be a little difficult to showcase the difference here. The difference comes in the code and how the developer plans to use the RadioGroup. Since there are many scenarios, I'm not sure choosing just one will add any value. I also just took a look at [Fabric's RadioGroup page](https://developer.microsoft.com/en-us/fluentui#/controls/web/choicegroup) and they have the same example for both controlled/uncontrolled. I'm going to leave as is for now :)

ksiler: Interesting. And yeah that's a good point, no worries then :) thanks for implementing this!

PR #528, opened 2020-10-20
Update Podspecs of iOS components, add more metadata to some package.jsons

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

In preparation for an upcoming change to our .yml files (where I will build and package the static libraries for the iOS components into our publish NuGet Package), I wanted to get this review out sooner to separate / compartmentalize changes.

### Description of changes

I wanted to update our podspec names from "FluentUI-React-Native-<Component>" to simply "FRN<Component>". This is because:
A) It's shorter
B) Three letter prefixes seem to be commonly used (though I don't like introducing new ones... we now have MSF, RNF, and FRN. Open to feedback)
C) The names of the built static libraries would then be "libFRNShimmer.a" instead of "libFluentUI-React-Native-Shimmer.a"... which just seems to be a better name.

Additionally, I updated the package.json to have some extra metadata so that the .podspec files could pull it from there. This simplifies the podspec and is better  follows "One Source of Truth". I followed the [React Native Webview podspec](https://github.com/react-native-webview/react-native-webview/blob/master/react-native-webview.podspec).

While I was at it, I ended up alhpabetizing/ updating metadata of the fluent-tester podspec, which I decided to keep. 

### Verification

Locally built the iOS and macOS test apps.
Ran `yarn buildci` locally.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
@chiuam as an FYI, since I'm renaming some of the names in the podspecs and she is working on the Native Button implementation

Saadnajmi:
Also, totally open to feedback on naming. I chose "FRN" over "FluentUI-React-Native" because it's shorter and seemed descriptive enough but I'm not sure if there's a precedence for naming of cocoapods / Fluent / if we all just wanna use one naming scheme.

Saadnajmi:
Gonna tag @anandrajeswaran in case he has comments on naming 😛

anandrajeswaran:
@Saadnajmi thanks for looping me in, you know me well :). I actually prefer the longer name since Apple doesn't really have any precedence in their apps of exposing three-letter-prefixes as part of package names. Not sure if cocoapods has conflicting conventions. As an example analogous one you brought up, is MSF actually part of a package name? I wasn't aware that it was but definitely could be wrong.

I think preferring the explicit name at the expense of brevity might just be the old Objective C curmudgeon in me though, so if you and others think this is better (or this matches Cocoapods conventions) I definitely don't feel too strongly about it.

apps/ios/src/Podfile.lock
```diff
@@ -258,9 +258,9 @@ DEPENDENCIES:
   - DoubleConversion (from `../../../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`)
   - FBLazyVector (from `../../../node_modules/react-native/Libraries/FBLazyVector`)
   - FBReactNativeSpec (from `../../../node_modules/react-native/Libraries/FBReactNativeSpec`)
-  - FluentUI-React-Native-Avatar (from `../../../packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec`)
-  - FluentUI-React-Native-Shimmer (from `../../../packages/components/Shimmer/FluentUIReactNativeShimmer.podspec`)
   - Folly (from `../../../node_modules/react-native/third-party-podspecs/Folly.podspec`)
+  - FRNAvatar (from `../../../packages/experimental/Avatar/FRNAvatar.podspec`)
```
HeyImChris, line 262: Reminder these are potentially renamed as per Anand's comments on your other pr

packages/components/Shimmer/FRNShimmer.podspec
```diff
@@ -1,14 +1,16 @@
...
-version = package['version']
 
 Pod::Spec.new do |s|
-  s.name             = 'FluentUI-React-Native-Shimmer'
-  s.version          = "#{version}"
-  s.summary          = 'Fluent UI React Native Shimmer Control'
-  s.homepage         = "https://www.microsoft.com/design/fluent/#/"
-  s.author           = { "Microsoft" => "fluentuinativeowners@microsoft.com"}
+  s.name         = 'FRNShimmer'
+  s.version      = package['version']
```
HeyImChris, line 7: What's the difference between this and what we used to have for version?

packages/experimental/Avatar/FRNAvatar.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.summary      = package['description']
+  s.license      = package['license']
+
+  s.authors      = package['author']
+  s.homepage     = package['homepage']
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
+
+  s.ios.deployment_target = "11.0"
```
HeyImChris, line 17: Gross it's so old. I just checked RNTester and apparently that's still on iOS 10 though so I guess we can leave for now. Longer term I'd like to adopt a similar approach to Office where we have firm guidelines on what OS versions we support instead of just arbitrarily dropping versions when we find they don't work anymore

apps/fluent-tester/package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "@fluentui-react-native/tester",
   "version": "0.11.1",
+  "description": "A test app to test FluentUI React Native Components during development",
+  "license": "MIT",
+  "author": "Microsoft <fluentuinativeowners@microsoft.com>",
+  "homepage": "https://github.com/microsoft/fluentui-react-native",
```
HeyImChris, line 7: Not for this review but maybe over a doc day or hackathon we can update all the other components to have this metadata as well

HeyImChris (approved):

packages/components/Shimmer/FRNShimmer.podspec
Saadnajmi: We used to have a separate line do this:
`version = package['version']`
and then reference it. Not sure why the indirection when we can do it directly do it here. I copied this heavily off of react-native-webview btw, since that's my go-to react-native-community-module to look at

packages/experimental/Avatar/FRNAvatar.podspec
Saadnajmi: Yeah =( I kept it as 11 since that's what we have for FluentUI-Apple. That one should be at 13, but I think we have 1 holdover client. Good to bring up with @rdeassis and @harrieshin though! https://github.com/microsoft/fluentui-apple/blob/master/MicrosoftFluentUI.podspec

PR #529, opened 2020-10-21
Package Apple Static Libraries + JS Bundle in Nuget Package

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update our yml files to build the apple test app and package the built static libraries into our NuGet package.

There's a few things happening here:

1. Renamed apple-xcode-build.yml to apple-yarn-build.yml.
1. Removed some unnecessary parameters from apple-yarn-build.yml as they weren't relevant anymore
1. Create a new apple-xcode-build.yml that uses xcodebuild instead of yarn, giving us more control over compiling the test apps.
1. Create `apple-xcode-build-static-libs.yml` (and some xcconfigs) that uses `apple-xcode-build.yml` and builds the apple test apps in various configurations. Much of this is copied from FluentUI-Apple.
1. Add steps to `publish-nuget-job.yml` to bundle the javascript and build the static libraries.
1. Update the `package.nuspec` to pack the bundles and static libraries
### Verification

Made my own pipeline in the same ADO instance our other pipeline lives. Made sure we didn't break the existing NPM Publish by separating it to a different job

Screen Shot 2020-10-26 at 10 36 07 AM

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/templates/apple-xcode-build-static-libs.yml
```diff
@@ -0,0 +1,63 @@
+#
+# Build the FluentUI React Native Test App in different configurations
+#
+steps:
+  # macOS Framework Debug
```
HeyImChris, line 5: Having all these param changes into the same template seems like a good candidate for a matrix

.ado/templates/apple-xcode-build-static-libs.yml
```diff
@@ -0,0 +1,63 @@
...
+      xcode_configuration: 'Debug'
+      xcode_extraArgs: '-xcconfig $(Build.Repository.LocalPath)/.ado/xcconfig/publish_overrides.xcconfig'
+
+  # macOS Framework Release
+  - template: apple-xcode-build.yml
+    parameters:
+      xcode_sdk: 'macosx'
+      xcode_workspacePath: 'apps/macos/src/FluentUITester.xcworkspace'
+      xcode_actions: 'build'
+      xcode_scheme: 'ReactTestApp'
```
HeyImChris, line 21: `ReactTestApp`? What about FURNTestApp or something that has to do with FluentUI

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: Yeah.. I think I'll look into that for the next iteration and just turn this all info a separate job with a matrix. Wanted to get this PR out to get feedback first, I'll try and use a matrix here.

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: ReactTestApp is the scheme from ''react-native-test-app', I don't think we can control that name. It would be best if we had a scene that was just a status library, but I don't think 'react-native-test-app' had support for that

HeyImChris: Ah, we're using `react-native-test-app`- forgot about that. I think it's fine as is then :)

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: Side note, since this is copied from FluentUI Apple. If I refactor this, there's gains we can have in that repo too. @rdeassis and @markavitale as an FYI :D

package.nuspec
```diff
@@ -9,8 +9,23 @@
...
+      <!-- iOS device ship -->
+      <file src="DerivedData\polyester\Build\Products\Release-iphoneos\libFRN*.a" target="Ship-iphoneos"/>
+      <!-- iOS simulator debug -->
+      <file src="DerivedData\polyester\Build\Products\Debug-iphonesimulator\libFRN*.a" target="Debug-iphonesimulator"/>
+      <!-- iOS simulator ship -->
+      <file src="DerivedData\polyester\Build\Products\Release-iphonesimulator\libFRN*.a" target="Ship-iphonesimulator"/>
+      <!-- macOS debug -->
+      <file src="\DerivedData\polyester\Build\Products\Debug\libFRN*.a" target="Debug-macosx"/>
+      <!-- macOS ship -->
+      <file src="DerivedData\polyester\Build\Products\Release\libFRN*.a" target="Ship-macosx"/>
```
anandrajeswaran, line 23: Can we have an explicit list instead of using wildcards. Being explicit with a manifest has in practice been no inconvenience in practice and great at making changes intentional

.ado/xcconfig/publish_overrides.xcconfig
```diff
@@ -0,0 +1,13 @@
+// For publish builds, only provide line tables for symbolizing crashes
+CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only
+OTHER_SWIFT_FLAGS=-gline-tables-only
+
+// Optimize for size in publish builds
+SWIFT_OPTIMIZATION_LEVEL=-Osize
+SWIFT_REFLECTION_METADATA_LEVEL=none
```
anandrajeswaran, line 7: Don't we also publish debug builds?

.ado/templates/apple-xcode-build-static-libs.yml
anandrajeswaran: Sounds like a bug in react-native-test-app? We should at least create an issue?

Saadnajmi: Not so much a bug but a feature ask? I'll create an issue in their repo though

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: Oh, yes. I assume them that lines 6 and 7 have to be release only then?

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: Filed https://github.com/microsoft/react-native-test-app/issues/239

.ado/templates/apple-xcode-select.yml
```diff
@@ -3,6 +3,6 @@
 #
 steps:
   - bash: |
-      sudo xcode-select --switch '/Applications/Xcode_11.4.1.app'
-    displayName: 'Switch to XCode 11.4.1'
+      sudo xcode-select --switch '/Applications/Xcode_12.app'
+    displayName: 'Switch to XCode 12'
```
Saadnajmi, line 7: I should probably revert this, but also worth noting Azure Pipelines is dropping support for Xcode 11.4 in November... so might as well keep it in?

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: I had to not use these as the xcodebuild invocation would then fail. I'm not sure why, but I'll probably need to remove them from the review and add back later, say, for ARM64 support on macOS

apps/ios/src/Podfile
```diff
@@ -33,8 +33,8 @@ use_test_app! do |target|
     script_phase name: 'Start Packager',
                  script: start_packager_script,
                  execution_position: :before_compile
-    script_phase name: 'Bundle React Native code and images',
-                 script: bundle_script
+    # script_phase name: 'Bundle React Native code and images',open
+    #              script: bundle_script
```
Saadnajmi, line 37: This build phase also caused the xcodebuild invocation to fail, and I'm fairly sure we don't need it anyway. Will just remove instead of comment out if that looks fine to people.

package.nuspec
```diff
@@ -9,8 +9,27 @@
...
+      <!-- iOS simulator debug -->
+      <file src="DerivedData\Build\Products\Debug-iphonesimulator\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Debug-iphonesimulator"/>
+      <file src="DerivedData\Build\Products\Debug-iphonesimulator\FluentUI-React-Native-Shimmer\libFluentUI-React-Native-Shimmer.a" target="Debug-iphonesimulator"/>
+      <!-- iOS simulator ship -->
+      <file src="DerivedData\Build\Products\Release-iphonesimulator\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Ship-iphonesimulator"/>
+      <file src="DerivedData\Build\Products\Release-iphonesimulator\FluentUI-React-Native-Shimmer\libFluentUI-React-Native-Shimmer.a" target="Ship-iphonesimulator"/>
+      <!-- macOS debug -->
+
+      <!-- macOS ship -->
+
```
Saadnajmi, line 27: I left the comments in as we'll almost certainly need these soon, so might as well leave them in?

package.nuspec
Saadnajmi: Done

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: https://github.com/microsoft/react-native-test-app/issues/239

package.nuspec
```diff
@@ -9,8 +9,27 @@
         <repository type="git" url="$repoUri$" commit="$commitId$"/>
     </metadata>
     <files>
-        <file src="packages/**/*.*" />
-        <file src="scripts/**/*.*" />
-        <file src="*.*" />
+      <!-- iOS device debug -->
```
tkamireh, line 12: Why can't you capture all these files via glob patterns? You can use excludes if some artifacts are undesired.

package.nuspec
```diff
@@ -9,8 +9,27 @@
...
+      <!-- iOS simulator ship -->
+      <file src="DerivedData\Build\Products\Release-iphonesimulator\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Ship-iphonesimulator"/>
+      <file src="DerivedData\Build\Products\Release-iphonesimulator\FluentUI-React-Native-Shimmer\libFluentUI-React-Native-Shimmer.a" target="Ship-iphonesimulator"/>
+      <!-- macOS debug -->
+
+      <!-- macOS ship -->
+
+
+      <!-- iOS jsbundle -->
+      <file src="apps\ios\dist\index.bundle" target="ios"/>
```
tkamireh, line 30: Out of curiosity are these ship or dev bundles? Do you care?

package.nuspec
Saadnajmi: See @anandrajeswaran 's comment [here](https://github.com/microsoft/fluentui-react-native/pull/529#discussion_r509911726). We preferred being explicit

package.nuspec
Saadnajmi: Ah, we should probably have separate ones. Good catch =)

.ado/templates/apple-yarn-build.yml
```diff
@@ -0,0 +1,18 @@
...
+    workingDirectory: apps/${{ parameters.relative_directory }}
+    displayName: 'pod install'
+
+  - bash: |
+      echo "yarn $platform_directory"
+      yarn $platform_directory
+    workingDirectory: apps/${{ parameters.relative_directory }}
+    displayName: 'yarn ${{ parameters.relative_directory }}'
+    env:
+      platform_directory: ${{ parameters.relative_directory }}
```
tkamireh, line 18: For consistency, you should use the env variable either everywhere or nowhere in this step, probably.

.ado/templates/apple-xcode-select.yml
tkamireh: Anywhere else you need to update this?

package.nuspec
Saadnajmi: Would that just be a `yarn bundle --dev`?I'm looking at `scripts/src/tasks/metroTasks.ts` to infer that

.ado/azure-pipelines.publish.yml
```diff
@@ -12,27 +11,69 @@ pr: none
...
+      - script: |
+          yarn test
+        displayName: 'yarn test'
+
+      - script: |
+          yarn publish:beachball -- -b origin/master -n $(npmAuth) --access public -y
+        displayName: 'Publish NPM Packages'
+
+  - job: NuGetPublish
+    displayName: NuGet Publish
```
tkamireh, line 42: I might consider throwing the entire job into a template. Could be useful if we ever need to run test publish or decide to split things out into multiple pipelines

Saadnajmi: Would you like me to move both NPM Publish and NuGet Publish to two separate templates? I can definitely see the usefulness in that

.ado/templates/apple-xcode-build-static-libs.yml
Saadnajmi: Update @HeyImChris . Using a matrix sends the build to different agents, and I need the built libraries to all be on the same host machine afterwards so I can pack it. In my preiliminary testing, it didn't seem like I could split it out and I kept it all together

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: Ended up not using any of the xcconfig overrides. Will revisit in a future change, but will leave it off for now

anandrajeswaran: But we still need to ensure those optimizations are on for release builds (shouldn't be in xcconfig overrides but in static xcconfigs that live with the project itself and not just as part of the publish pipelines

.ado/templates/apple-xcode-select.yml
Saadnajmi: Nope! We already moved to XCode 12 in FluentUI Apple and in Office. Just here should be fine

.ado/azure-pipelines.publish.yml
```diff
@@ -12,27 +11,69 @@ pr: none
...
+          yarn bundle ios
+        workingDirectory: $(Build.Repository.LocalPath)/apps/ios
+        displayName: 'yarn bundle iOS'
+
+      - script: |
+          yarn bundle macos
+        workingDirectory: $(Build.Repository.LocalPath)/apps/macos
+        displayName: 'yarn bundle macOS'
+
+      # Select proper XCode version
```
HeyImChris, line 74: I've been meaning to fix this elsewhere but it's a lowercase 'c'- "Xcode"

package.nuspec
HeyImChris: Yea I think leaving them in makes sense

.ado/templates/apple-xcode-select.yml
```diff
@@ -1,8 +1,12 @@
...
+# Task Group: Xcode select proper version
 #
+parameters:
+  xcode_path_override: '' # Leaving this blank results in the repo default being applied
+
 steps:
-  - bash: |
-      sudo xcode-select --switch '/Applications/Xcode_11.4.1.app'
-    displayName: 'Switch to XCode 11.4.1'
+  - bash: scripts/xcode_select_current_version.sh
```
HeyImChris, line 8: Where is this script?

Saadnajmi: ...derp. Forgot to bring that over. I was copying what we did in FluentUI Apple when I noticed they don't explicitly set an xcode version and use the default. I'll bring it over

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: As it turns out, we already have some of these settings in xcconfigs we get for free from `react-native-test-app`
https://github.com/microsoft/react-native-test-app/pull/151/files
https://github.com/microsoft/react-native-test-app/blob/master/macos/ReactTestApp/ReactTestApp.common.xcconfig
https://github.com/microsoft/react-native-test-app/blob/master/macos/ReactTestApp/ReactTestApp.debug.xcconfig
https://github.com/microsoft/react-native-test-app/blob/master/macos/ReactTestApp/ReactTestApp.release.xcconfig
https://github.com/microsoft/react-native-test-app/blob/master/ios/ReactTestApp/ReactTestApp.common.xcconfig
https://github.com/microsoft/react-native-test-app/blob/master/ios/ReactTestApp/ReactTestApp.debug.xcconfig
https://github.com/microsoft/react-native-test-app/blob/master/ios/ReactTestApp/ReactTestApp.release.xcconfig

.ado/azure-pipelines.publish.yml
Saadnajmi: I'll leave this for a future change that needs it, as this is already a lot of refactoring and I don't want to unnecessarily add more on top of that with this change

HeyImChris (approved):

PR #536, opened 2020-10-22
Add "Segoe UI Semibold" font to defaults for semibold variants

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adds "Segoe UI Semibold" font to defaults for semibold variants. 

Previously, using "Segoe UI" for semibold variants rendered as bold when rich labels were being used. I have a native change that fixes this so this just updates the default fallback values.

### Verification

| 
image
 | 
image
 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
@@ -17,7 +17,6 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
   const BrushScriptMT = Text.customize({ tokens: { variant: 'heroStandard', fontFamily: 'Brush Script MT' } });
   const CourierNew = Text.customize({ tokens: { variant: 'headerStandard', fontFamily: 'Courier New' } });
   const Georgia = Text.customize({ tokens: { variant: 'subheaderStandard', fontFamily: 'Georgia' } });
-  const SegoeScript = Text.customize({ tokens: { variant: 'bodyStandard', fontFamily: 'Segoe Script' } });
   const TimesNewRoman = Text.customize({ tokens: { variant: 'secondaryStandard', fontFamily: 'Times New Roman' } });
   const Wingdings = Text.customize({ tokens: { variant: 'captionStandard', fontFamily: 'Wingdings' } });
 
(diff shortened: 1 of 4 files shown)


Saadnajmi:
This might cause problems with the apple platforms. I remember the very first big we ran into for MS Build was along the lines of segoe ui not existing on iOS/macOS

ksiler:
> This might cause problems with the apple platforms. I remember the very first big we ran into for MS Build was along the lines of segoe ui not existing on iOS/macOS

Thanks for bringing this up! Perhaps I should make "Segoe UI Semibold" the secondary font face and it apply it that way so it doesn't effect Apple. I'm not sure if this is the right move though. @tkamireh @FalseLobster do you have any recommendations?   

This is a bit of a special case where the primary font face ("Segoe UI") does not work for rendering semibold text on win32. We currently have "System" set for the other parts of font family (secondary, cursive, monospace, etc.) so we could stick "Segoe UI Semibold" in one of those spots but I'm not sure that's what we want to do. Should this be a question for design?

Saadnajmi:
Also side note, I'm pretty sure this change affects all platforms and not just win32 (looking at change notes). I'm happy to test things on iOS/macOS by pulling in the change! I'm sure not everyone has access to a Mac =)

tkamireh:
@ksiler I'm understanding correctly in that the native fix addresses this and this change is not strictly necessary? If we can specify what we want text to look like without bringing in an additional font face that seems preferable to me.

ksiler:
> @ksiler I'm understanding correctly in that the native fix addresses this and this change is not strictly necessary? If we can specify what we want text to look like without bringing in an additional font face that seems preferable to me.

@tkamireh Yes that's correct. I have a native change that covers this for win32 but thought I would update the defaults (or fallback values) in the event the theming module doesn't load correctly. I'm not sure how often or likely that will happen but if it does, semibold font variants would render as bold on win32. Is that something we're willing to live with to avoid adding additional font faces here? Or should we create a case for win32 like macos and ios?

tkamireh (approved):

Saadnajmi (approved):

PR #553, opened 2020-11-03
RadioButton: Adding selection on onPress functionality

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes

This PR adds functionality to change RadioButton selection when onPress is called (before, it was only during onFocus, now it's both).  We're changing selection on both onPress and onFocus to integrate better with NetUI. For example, for TaskPanes, if focus isn't already landed on the TaskPane and you click inside, keyboard focus does not get moved there, and thus no onFocus is called. This is where we need onPress selection functionality.

Some background: When keyboarding through a RadioGroup, you want selection to change as you move keyboard focus. This is why initially, only onFocus was needed. I was not aware of the complications that might have with certain NetUI panes.

### Verification

Manual + Automated Testing

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -22,11 +22,21 @@ export const RadioButton = compose<IRadioButtonType>({
     // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
     const info = React.useContext(RadioGroupContext);
 
+    /* We don't want to call the user's onChange multiple times on the same selection. */
+    const changeSelection = () => {
+      if(buttonKey != info.selectedKey)
+        info.onChange && info.onChange(buttonKey);
+    };
+
+    /* We're changing selection on both onPress and onFocus to integrate better with NetUI. For example, for TaskPanes, if focus isn't already landed
```
PPatBoyd, line 31: I would avoid references to underlying platform details (NetUI), I would just describe it as radio button changes selection when focus is moved between radio button items as well as on click

PPatBoyd (approved):
approved w/ suggestion of updating comment; this is xplat code, no?  If we need win32 specific we need to split that out

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Makes sense, changing in newest commit

samuelfreiberg:
> approved w/ suggestion of updating comment; this is xplat code, no? If we need win32 specific we need to split that out

Yes xplat, updated description

PR #556, opened 2020-11-09
Create a Native Button component

chiuam:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
This change wraps the native implementation of the mac & iOS button into the newly created component named Native Button. Both platforms share the same props and tokens, however some functionalities are exclusive to macOS due to the restrictiveness of the iOS native implementation. 

Customizations exclusive to macOS :
1. Additional button styles: acrylic (the workaround is to map acrylic to primary style for iOS) 
2. Background color tinting
3. Image support for Borderless button
4. Optional tinting for image
(2-4 will be no-op in iOS)

### Verification

image
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/FluentUIReactNativeButton.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.homepage     = "https://github.com/microsoft/fluentui-react-native"
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
+
+  s.osx.deployment_target = "10.13"
+  s.osx.source_files      = "macos/*.{swift,h,m}"
+
+  s.dependency 'React'
+  s.dependency 'MicrosoftFluentUI', '~> 0.1.16'
```
harrieshin, line 21: should we use subspec just to grab button related code?

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonVariantTestSection.tsx
```diff
@@ -1,20 +1,56 @@
...
+  const [buttonTitle2, setTitle2] = React.useState('Mac Native Button');
+  const onPress2 = () => setTitle2('Native Button Clicked');
+
+  const [buttonTitle3, setTitle3] = React.useState('Mac Native Button');
+  const onPress3 = () => setTitle3('Native Button Clicked');
+
+  if (Platform.OS == 'macos') {
+    return (
+      <View style={[stackStyle, commonTestStyles.view]}>
+        <TouchableOpacity onPress={onPress1}>
```
JasonVMo, line 19: Why do we need to wrap these in TouchableOpacity entries? Doesn't the button support onPress?

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonVariantTestSection.tsx
```diff
@@ -1,20 +1,56 @@
...
-      <Button ghost content="Ghost" style={commonTestStyles.vmargin} />
-      <Button fluid content="Fluid" style={commonTestStyles.vmargin} />
-      <Button primary fluid content="Fluid Primary" style={commonTestStyles.vmargin} />
-      <Button ghost fluid content="Fluid Ghost" style={commonTestStyles.vmargin} />
-      {/* <CompoundButton content="Default" secondaryContent="Compound" style={commonTestStyles.vmargin} />
-      <CompoundButton primary content="Primary" secondaryContent="Compound" style={commonTestStyles.vmargin} /> */}
-      <CompoundButton ghost content="Ghost" secondaryContent="Compound" style={commonTestStyles.vmargin} />
-    </View>
-  );
+  const [buttonTitle1, setTitle1] = React.useState('Mac Native Button');
```
JasonVMo, line 7: Test code so this is a nit, but this code is being defined for all platforms but only used for macos. It should be inside the platform.OS block.

packages/experimental/Button/src/Button.tsx
```diff
@@ -1,41 +1,55 @@
...
-          {content && <Slots.content key="content">{content}</Slots.content>}
-          {children}
-        </Slots.root>
-      );
-    };
-  },
-});
+export const Button =
+  Platform.OS === 'macos'
+    ? compose<ButtonTypeMac>({
```
JasonVMo, line 17: Right now I don't think this approach is a great experience for customers. (See the forking required in the test code) Fundamentally we are making the macos button inexplicably have a different and less functional interface. There are issues with each:

- Different - the fact that the interfaces are different is an incredible pain for people authoring as they have to fork/re-author everything. Typescript also doesn't have good affordances for multi-platform building. As a result it is certain that TS builds are only validating for one platform, type errors will slip through the other ones.
- Less Functional - taking recomposition scenarios out of it for a second (as those are hard), this effectively only allows a few pre-canned button variants, ignores tokens entirely (though it will still have a customize function on it). This also means that if you author your code in win32 first (which many Office feature authors are doing), then try to port it over to mac, you'll have to come up with new implementations for any custom button usages.

I think there are really two viable approaches here. Having Button be a unified front-end that will transparently use native buttons internally when they are sufficient for what the customer wants, or make a well defined native button control that is explicit in what you are using and supporting.

I'd probably start by creating an explicit native button control. If you are using native button anyways you suddenly have a tie between native/JS code that didn't necessarily exist before. (If you service deliver a new bundle you have to validate the versioning as the native code won't update)

The explicit control allows thinking through what an appropriate interface looks like, thinking through questions like:

- is this a native button for all platforms? All apple platforms? Mobile? Mac only?
- This leads into what is the best interface for a native button
- What is our versioning story for matching our native bits to our bundle? (I've been diving into SDK stuff for the past few weeks so this ends up being a big deal)

We can then look at routing button to native when it is available but that might be a step 2.

JasonVMo (changes requested):

apps/macos/src/Podfile
```diff
@@ -22,6 +22,9 @@ fi
 use_flipper!(false)
 use_test_app! do |target|
   target.app do
+
+    pod 'FluentUI-React-Native-Button', :path => '../../../packages/experimental/Button/FluentUIReactNativeButton.podspec'
```
HeyImChris, line 26: @JasonVMo Do we want new controls in experimental or is components still where they should live?

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonVariantTestSection.tsx
chiuam: Yeah, this is sort of a hack to get the onPress event working. I was thinking the proper way of doing it is to export the native onPress event to JS using RCTEventEmitter?

chiuam: I removed the TouchableOpacity wrappers and used RCTBubblingEventBlock to fire events from native.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonVariantTestSection.tsx
```diff
@@ -1,20 +1,58 @@
...
-      <Button ghost fluid content="Fluid Ghost" style={commonTestStyles.vmargin} />
-      {/* <CompoundButton content="Default" secondaryContent="Compound" style={commonTestStyles.vmargin} />
-      <CompoundButton primary content="Primary" secondaryContent="Compound" style={commonTestStyles.vmargin} /> */}
-      <CompoundButton ghost content="Ghost" secondaryContent="Compound" style={commonTestStyles.vmargin} />
-    </View>
-  );
+  if (Platform.OS == 'macos') {
+    const CustomizedNativeButton = Button.customize({
+      contentTintColor: processColor('white'),
+      restBackgroundColor: processColor('orange'),
```
Saadnajmi, line 10: You shouldn't need to call `processColor`. That is an internal React-Native API and when I used it in Shimmer, it was a bug. If you set `contentTintColor` and `restBackgroundColor` as `string` types instead of `Number`, RCTConvert should convert them to proper NSColors. I can follow up offline if that causes problems

Saadnajmi, line 10: I feel these names are very specific to the macOS implementation and should not be in the React Native Button. I feel there should be a more general name (one that probably comes from design like `accentColor`) that maps to `contentTintColor` internally when the JS component passes props to the native module.

packages/experimental/Button/FluentUIReactNativeButton.podspec
```diff
@@ -0,0 +1,22 @@
...
+  s.homepage     = "https://github.com/microsoft/fluentui-react-native"
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
+
+  s.osx.deployment_target = "10.13"
+  s.osx.source_files      = "macos/*.{swift,h,m}"
+
+  s.dependency 'React'
+  s.dependency 'MicrosoftFluentUI/Button_mac'
```
Saadnajmi, line 21: `s.osx.dependency`?

packages/experimental/Button/src/Button.types.ts
```diff
@@ -111,16 +111,62 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
   fluid?: boolean;
 }
 
+export interface ButtonMacProps {
+  /*
+   * Name of the icon to show on the Button.
+   */
+  image?: Image;
```
Saadnajmi, line 118: I learned from working on `Avatar` that passing an `image` is actually not quite simple, and it may be easier to pass an `ImageURISource` that the component internally calls `Image.resolveAssetSource(uri: ...)` on so that RCTConvert can properly convert it to a UIImage / NSImage

packages/experimental/Button/src/Button.types.ts
```diff
@@ -111,16 +111,62 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
...
+  /*
+   * Text to show on the Button.
+   */
+  title?: string;
+  /*
+   * ButtonStyle wrapped enum
+   */
+  buttonStyle?: ButtonMacStyle;
```
Saadnajmi, line 126: I'm not sure if @chripear 's change in FluentUI apple means we would have the same styles across native and React Native, but I would rather not define a separate type purely for Mac Styles. At the very least, I would prefer that this is just `ButtonStyle` and the platform check happens inside `ButtonStyle`. 
But again, I'd rather we just have the same name for the styles across platforms, or map the RN styles to the native styles we have

packages/experimental/Button/src/Button.types.ts
```diff
@@ -111,16 +111,62 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
...
+  /*
+   * A button can be enabled or disabled.
+   */
+  isEnabled?: boolean;
```
Saadnajmi, line 130: I also feel like this is somewhat platform specific? Although I don't really see anything in RN that would map to this

packages/experimental/Button/src/Button.types.ts
```diff
@@ -111,16 +111,62 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
...
+export interface ButtonMacTokens {
+  /*
+   * Button content tint color
+   */
+  contentTintColor?: number;
+  /*
+   * Button background color for rest state
+   */
+  restBackgroundColor?: number;
+}
```
Saadnajmi, line 146: I wonder if instead of defining these here, we could make a `ButtonTokens.macos.ts` and just import `ButtonTokens` and keep the platform specific tokens separate that way

Saadnajmi (changes requested):

packages/experimental/NativeButton/FluentUIReactNativeButtonMac.podspec
```diff
@@ -0,0 +1,22 @@
+require 'json'
+
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
+Pod::Spec.new do |s|
+  s.name         = 'FluentUI-React-Native-Button-Mac'
```
Saadnajmi, line 6: Could we just have one podspec. and specify different source files for osx or ios?

packages/experimental/NativeButton/ios/UIButtonViewManager.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import FluentUI
+
+@objc(MSFUIButtonViewManager)
+class UIButtonViewManager: RCTViewManager {
+
+	override func view() -> UIView! {
+		let button = UIButtonWrapper()
+	return button
```
Saadnajmi, line 9: nit: indentation looks off?

packages/experimental/NativeButton/ios/UIButtonViewManager.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import FluentUI
+
+@objc(MSFUIButtonViewManager)
+class UIButtonViewManager: RCTViewManager {
```
Saadnajmi, line 5: Also, why is this called UIButtonViewManager? UIButton comes from UIKit, but we are using  a Fluent Button?

packages/experimental/NativeButton/macos/MSFButtonViewManager.m
```diff
@@ -0,0 +1,30 @@
...
+RCT_ENUM_CONVERTER(MSFButtonStyle, (@{
+	@"primary": @(MSFButtonStylePrimary),
+	@"secondary": @(MSFButtonStyleSecondary),
+	@"acrylic": @(MSFButtonStyleAcrylic),
+	@"borderless": @(MSFButtonStyleBorderless),
+}), MSFButtonStylePrimary, integerValue);
+
+@end
+
+@interface RCT_EXTERN_MODULE(MSFButtonViewManager, RCTViewManager)
```
Saadnajmi, line 17: I personally like one view manager per platform better, but could we combine the two viewManagers into one and use ifdefs like what what @amgleitman has been doing in react-native-macos and community modules?

packages/experimental/NativeButton/FluentUIReactNativeButtonMac.podspec
chiuam: Good suggestion, I like that we can update the versions in one file.

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
```diff
@@ -0,0 +1,34 @@
...
+#import "RCTBridgeModule.h"
+
+@import FluentUI;
+
+@implementation RCTConvert (MSFButtonAdditions)
+RCT_ENUM_CONVERTER(MSFButtonStyle, (@{
+	@"primary": @(MSFButtonStylePrimaryFilled),
+	@"secondary": @(MSFButtonStyleSecondaryOutline),
+	@"borderless": @(MSFButtonStyleBorderless),
+  @"acrylic": @(MSFButtonStylePrimaryFilled), // mac specific button style
```
Saadnajmi, line 12: It feels like this is just here to get it to compile, but we aren't actually setting the iOS button to an "acrylic" style, rather, just primary filled? I would prefer we either update the comment to show this isn't * really * an acrylic button, or add that actual style to the iOS button (which probably won't happen bc it's not part of design).

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
```diff
@@ -0,0 +1,34 @@
...
+@end
+
+@interface RCT_EXTERN_MODULE(MSFUIButtonViewManager, RCTViewManager)
+
+//RCT_EXPORT_VIEW_PROPERTY(image, UIImage);
+RCT_EXPORT_VIEW_PROPERTY(onPress, RCTBubblingEventBlock);
+RCT_REMAP_VIEW_PROPERTY(buttonStyle, style, MSFButtonStyle);
+RCT_CUSTOM_VIEW_PROPERTY(title, String, MSFButton) {
+	// for simplicity only normal state is supported
+	[view setTitle:[RCTConvert NSString:json] forState:normal];
```
Saadnajmi, line 24: Setting it for normal sets it for all states, is that fine? https://stackoverflow.com/questions/1375145/is-there-a-way-to-settitle-on-a-uibutton-that-updates-all-uicontrolstates-at-onc

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
```diff
@@ -0,0 +1,34 @@
...
+}
+RCT_CUSTOM_VIEW_PROPERTY(isEnabled, BOOL, MSFButton) {
+	[view setEnabled:[RCTConvert BOOL:json]];
```
Saadnajmi, line 27: Does this need to be custom because isEnabled != enabled?

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
```diff
@@ -0,0 +1,34 @@
...
+}
+RCT_CUSTOM_VIEW_PROPERTY(image, UIImage, MSFButton) {
+	[view setImage:[RCTConvert UIImage:json]];
+}
```
Saadnajmi, line 31: Why does this need to be custom? Why can't it be a simple RCT_EXPORT_VIEW_PROPERTY?

packages/experimental/NativeButton/macos/MSFButtonViewManager.m
```diff
@@ -0,0 +1,31 @@
...
+
+RCT_EXPORT_VIEW_PROPERTY(title, NSString);
+RCT_EXPORT_VIEW_PROPERTY(accentColor, NSColor);
+RCT_EXPORT_VIEW_PROPERTY(onPress, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(isImageTinted, BOOL);
+RCT_REMAP_VIEW_PROPERTY(buttonStyle, style, MSFButtonStyle);
+RCT_CUSTOM_VIEW_PROPERTY(isEnabled, BOOL, MSFButton) {
+	[view setEnabled:[RCTConvert BOOL:json]];
+}
```
Saadnajmi, line 26: We don't want/need this on iOS too then?

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
chiuam: It compiles without adding the acrylic enum, it just suppresses the no enum found error in the console(iOS only). I'll update the comment.

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
chiuam: I think it's because RCTConvert doesn't support NSImage. https://github.com/facebook/react-native/blob/master/React/Base/RCTConvert.h

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
chiuam: I've been looking into extending RCTConvert to add custom types. Anyway, I think it's fine for v1.

packages/experimental/NativeButton/macos/MSFButtonViewManager.m
chiuam: I find it easier to read  and less prone to bugs with two separate viewManagers.

packages/experimental/NativeButton/ios/UIButtonViewManager.swift
chiuam: Sorry, I'll rename them...

packages/experimental/Button/src/Button.types.ts
chiuam: Now that we have a component for native button, all platforms should conform to the same interface.

packages/experimental/Button/src/Button.types.ts
chiuam: I don't think this is platform specific, RN also uses isEnabled flag to indicate the button status.

packages/experimental/Button/src/Button.types.ts
chiuam: Agreed, we now have one buttonStyle for all platforms.

packages/experimental/Button/src/Button.types.ts
chiuam: Updated image type to ImageURISource.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonVariantTestSection.tsx
chiuam: That didn't work for mac. Could it be RCTConvert doesn't support NSColor?

packages/experimental/NativeButton/src/NativeButton.types.ts
```diff
@@ -0,0 +1,48 @@
...
+   * A callback to call on button click event
+   */
+  onPress?: () => void;
+}
+
+export interface NativeButtonTokens {
+  /*
+   * Button background color
+   */
+  accentColor?: string;
```
Saadnajmi, line 36: Can we add a comment saying this token is a no-op on iOS? Or split this to separate iOS and macOS specific token files?

packages/experimental/NativeButton/src/NativeButton.types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export type NativeButtonStyle = 'primary' | 'secondary' | 'borderless' | 'acrylic';
```
Saadnajmi, line 39: Similar comment about iOS and macOS specific button styles

packages/experimental/NativeButton/ios/MSFButtonViewManager.m
```diff
@@ -0,0 +1,31 @@
...
+
+@import FluentUI;
+
+@implementation RCTConvert (MSFButtonAdditions)
+RCT_ENUM_CONVERTER(MSFButtonStyle, (@{
+	@"primary": @(MSFButtonStylePrimaryFilled),
+	@"secondary": @(MSFButtonStyleSecondaryOutline),
+	@"borderless": @(MSFButtonStyleBorderless),
+  @"acrylic": @(MSFButtonStylePrimaryFilled), // mac specific button style
+}), MSFButtonStylePrimaryFilled, integerValue);
```
Saadnajmi, line 13: nit: weird spacing

Saadnajmi (approved):

packages/experimental/NativeButton/src/NativeButton.types.ts
chiuam: Given there's only one token defined, I'll add a comment to it for now.

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
chiuam: Good call. I was using custom view to set the enabled property. I think isEnabled didn't work because it's a swift only property. We should just export the enabled property so no need to make it custom...

packages/experimental/NativeButton/macos/MSFButtonViewManager.m
chiuam: Will remove

packages/experimental/Button/src/Button.types.ts
chiuam: Renamed to enabled

packages/experimental/NativeButton/ios/MSFUIButtonViewManager.m
chiuam: Using custom view was redundant so removed.

packages/experimental/Button/src/Button.tsx
chiuam: I created a separate component for native button that consists of the native implementations. Right now, it only supports apple platforms, but I think it has the potential to support other platforms as well. Since the control takes a dependency on FluentUI native, we need to update the native version in FURN whenever we make changes to the native button.

JasonVMo (approved):
Looks good, thanks for addressing the earlier feedback.

PR #565, opened 2020-12-02
Update react-native-win32 version and add Text pressability test

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update react-native-win32 version which introduces key events to TextWin32 (needed for pressability).

Add pressability Test to the Text page.

Replace acceptsKeyboardFocus with focusable.

### Verification

Manually tested with the test win32 app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/win32/package.json
```diff
@@ -33,7 +33,7 @@
     "react-native-svg": "^11.0.0"
   },
   "devDependencies": {
-    "@office-iss/react-native-win32": "0.62.3",
+    "@office-iss/react-native-win32": "0.0.0-canary.63",
```
tkamireh, line 36: This should probably be a dependency and not a devDepenendency, if you could move it that would be great!

Also are we sure we want to use the canary versions? I don't quite remember the version schema and implications.

ksiler: Sure, I can move it.

And I'm not sure. There isn't any non-canary versions past 0.62.3 besides the 0.64.0-preview. Who would be the best person to ask?

tkamireh: I do see [this](https://www.npmjs.com/package/@office-iss/react-native-win32/v/0.63.2).@NickGerleman can you remind us of the versioning schemas and who you expect to consume canary versions.

NickGerleman: Oh yeah, you probably don't want the canary version. Those versions are published nightly and can use unofficial RN builds and be full of bugs.

We should have versions of react-native-win32 0.63.x.

NickGerleman: Looks like latest is 0.63.2

ksiler: Thanks @NickGerleman. If I want to bump the version to something later than 0.63.2 (to include changes after 0.63.2), should I wait until a new stable version is released? Do you have an idea of when that would be?

NickGerleman: For each major version (e.g. 0.62, 0.63, etc) we have a dedicated stable branch. These are updated when you cherry pick a change into the branch. So, if you wanted a change in 0.63 after the branch was created, you'd cherry pick a change into "0.63-stable" and we would automatically publish a new version 0.63.x (we do that weekly).

We recently cut a new branch for 0.64, which should have almost everything in our master branch, and targets RN 0.64. These new releases are pretty infrequent.

ksiler: Thanks for the explanation @NickGerleman, I just added a PR for that here: https://github.com/microsoft/react-native-windows/pull/6762.

PR #552, opened 2020-11-02
Move windows test app to react-native-test-app

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

We use `react-native-test-app` to handle the native code of our android, iOS, and macOS test apps, rather than us creating native projects for each and checking in the native code. As of `0.2.0`, `react-native-test-app` added support for windows. Let's move move the windows test app to use `react-native-test-app` as well.

With this change, all platforms (except win32, which doesn't seem to have any native code anyway) now use `react-native-test-app`. Perhaps we could then combine the test apps back into `apps/fluent-tester` now?

Steps:
- Add the `react-native-test-app` dependency to the windows test app `package.json`. Make sure it matches the version of the other test apps so that the node module can be hoisted. This also happened to alphabetize the `package.json`.
- Remove everything under `apps/windows/windows/` except for the `.gitignore`
- Update `app.json` to have a appKey.
- Run `yarn install-windows-test-app` to generate `FluentTester.sln`

Now, `yarn windows` should work as expected, meaning our CI should also work as expected.

This should close #447 

### Verification

Built the app on my windows PC.

Screen Shot 2020-11-02 at 2 32 20 PM

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #566, opened 2020-12-04
Update react-native-test-app dependency, which allows RN 0.63 in macOS test app, and run-android CLI support

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Let's update our `react-native-test-app` dependency to `0.3.2`. This allows us to do a number of changes, and sets us up for future changes:

- Update our macOS test app to React Native macOS 0.63
- Update iOS test app so we don't need to specify the scheme in `yarn ios`
- Update our Azure Pipeline to build the "FluentTester" xcode scheme instead of "ReactTestApp" (they are copies of each other anyway)
- Update our android test app to add `yarn android` support. We no longer need to open Android Studio and can just build and run the test app from the CLI.

Future changes that could be implemented with 0.3.2 update (not in this CR):

- iOS and android test apps to RN 0.63
    - I could not update the iOS or android test apps to RN 0.63 yet. That causes RN version mismatches between the app and bundling server.
- Windows test app on RN 0.63 and/or `react-native-test-app`
    - Updating this dependency will also pick up some bug fixes that will help move `windows` to use `react-native-test-app`. However, as per [this comment](https://github.com/microsoft/fluentui-react-native/pull/552#discussion_r520019278) on my other PR, that will likely need to wait till React Native Windows 0.64.
- Add a CI check that builds the android test app on every PR.
    - Tried seeing if we could add this cheaply, bbut ultimately decided I'll leave it for a separate CR and not bundle it here. Testing YML is no fun.

### Verification

Built the iOS, android, and macOS test apps locally

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/src/Podfile
```diff
@@ -1,6 +1,6 @@
 require_relative '../../../node_modules/react-native-test-app/test_app.rb'
 
-workspace 'FluentUITester.xcworkspace'
+workspace 'FluentTester.xcworkspace'
```
Saadnajmi, line 3: Need to change the name here to simplify `yarn ios` in our package.json. Can't simplify `yarn macos` the same way unfortunately

apps/android/README.md
```diff
@@ -5,19 +5,31 @@
 ## Launch `FluentUI Tester` app on Android
 
 Prereqs:
+
```
HeyImChris, line 8: What do the pluses do?

apps/ios/package.json
```diff
@@ -16,13 +16,13 @@
...
     "bundle-dev": "fluentui-scripts metro --dev --cli",
-    "ios": "react-native run-ios --scheme ReactTestApp --project-path src"
+    "ios": "react-native run-ios --project-path src"
   },
   "dependencies": {
     "@fluentui-react-native/tester": "^0.13.7",
     "react": "16.11.0",
-    "react-native": "0.62.2",
-    "react-native-test-app": "^0.2.1"
+    "react-native": "^0.62.2",
```
HeyImChris, line 24: I don't think we should use the caret here. Although I used to think these version carets were handy, there's something to be said about having an easy way to tell the exact dependency version we're on. It's a bit more overhead to update it when we need to, but that consistency in reproducibility is really nice if ever there's a problem.

apps/ios/src/Podfile
HeyImChris: Why not?

HeyImChris (changes requested):

Saadnajmi:
@tido64  It looks like the `run-ios` fails if I don't specifty the scheme and let it default to "FluentTester" instead of "ReactTestApp". Looking at the log, it builds `ReactTestApp.app` but tries to launch `FluentTester.app`

```
Error: Command failed: /usr/libexec/PlistBuddy -c Print:CFBundleIdentifier /Users/sanajmi/Library/Developer/Xcode/DerivedData/FluentTester-ftadnpdiazxfkkgpjihfkwdxhume/Build/Products/Debug-iphonesimulator/FluentTester.app/Info.plist
Print: Entry, ":CFBundleIdentifier", Does Not Exist
```

apps/ios/src/Podfile
tido64: rn-cli defaults to the workspace name on iOS, but on macOS it defaults to workspace name + `-macOS`. ¯\\_(ツ)_/¯

tido64:
Did you try cleaning out `node_modules` first? If you only bumped and re-ran `yarn`, the old version is still installed.

Saadnajmi:
@tido64 I was referring to our CI pipeline, which would have no previous installs. And yes

apps/android/README.md
Saadnajmi: I think that's just a new line. I didn't add a plus

tido64:
This is a bug in `@react-native-community/cli` and was fixed in https://github.com/react-native-community/cli/pull/1236. You should upgrade to ^4.13.0.

Saadnajmi:
> This is a bug in `@react-native-community/cli` and was fixed in [react-native-community/cli#1236](https://github.com/react-native-community/cli/pull/1236). You should upgrade to ^4.13.0.

RN 0.62.2 (which iOS and android are still using) bring in `@react-native-community/cli 4.8.0`, and I can't seem to override that. I'll leave the change to not specify the scheme for the future, and keep this focused on android and macOS

tido64:
> RN 0.62.2 (which iOS and android are still using) bring in `@react-native-community/cli 4.8.0`, and I can't seem to override that. I'll leave the change to not specify the scheme for the future, and keep this focused on android and macOS

It looks like we've got dupes of `@react-native-community/cli` from looking at `yarn.lock`. I can look into addressing that once this PR goes in.

apps/android/package.json
```diff
@@ -21,8 +21,8 @@
   "dependencies": {
     "@fluentui-react-native/tester": "^0.13.7",
     "react": "16.11.0",
-    "react-native": "0.62.2",
-    "react-native-test-app": "^0.2.1"
+    "react-native": "^0.62.2",
```
HeyImChris, line 24: Probably want to be explicit with which version we pull in? Reproducing and debugging dependency issues is way easier when we have the exact version specified

HeyImChris (approved):

apps/macos/package.json
```diff
@@ -15,12 +15,12 @@
...
     "start": "react-native start",
     "test": "fluentui-scripts jest",
-    "macos": "react-native run-macos --scheme ReactTestApp --project-path src"
+    "macos": "react-native run-macos --scheme FluentTester --project-path src"
   },
   "dependencies": {
-    "@fluentui-react-native/tester": "^0.13.8",
-    "react-native-macos": "^0.62.0-0",
-    "react-native-test-app": "^0.2.1"
+    "@fluentui-react-native/tester": "^0.13.9",
```
Saadnajmi, line 21:
```suggestion
    "@fluentui-react-native/tester": "^0.13.8",
```

apps/macos/package.json
```diff
@@ -32,11 +32,11 @@
     "jest": "^24.9.0",
     "metro-config": "^0.58.0",
     "metro-react-native-babel-preset": "^0.58.0",
-    "react-native": "^0.62.2",
+    "react-native": "0.63.3",
```
tido64, line 35: Should we use latest 0.63?
```suggestion
    "react-native": "0.63.4",
```

apps/macos/package.json
```diff
@@ -15,12 +15,12 @@
...
     "test": "fluentui-scripts jest",
-    "macos": "react-native run-macos --scheme ReactTestApp --project-path src"
+    "macos": "react-native run-macos --scheme FluentTester --project-path src"
   },
   "dependencies": {
-    "@fluentui-react-native/tester": "^0.13.8",
-    "react-native-macos": "^0.62.0-0",
-    "react-native-test-app": "^0.2.1"
+    "@fluentui-react-native/tester": "^0.13.9",
+    "react-native-macos": "0.63.1",
```
tido64, line 22: Should we use latest 0.63?
```suggestion
    "react-native-macos": "0.63.2",
```

apps/macos/package.json
```diff
@@ -32,11 +32,11 @@
     "jest": "^24.9.0",
     "metro-config": "^0.58.0",
     "metro-react-native-babel-preset": "^0.58.0",
-    "react-native": "^0.62.2",
+    "react-native": "0.63.3",
     "react-test-renderer": "~16.11.0"
```
tido64, line 36: Does this also need to be bumped to 16.13?

apps/macos/package.json
```diff
@@ -32,11 +32,11 @@
     "jest": "^24.9.0",
     "metro-config": "^0.58.0",
     "metro-react-native-babel-preset": "^0.58.0",
```
tido64, line 34: These should probably be bumped to 0.59 to stay in sync with RN 0.63

Saadnajmi:
@tido64  juuuust missed all of your messages 😅 I suppose I can make a followup

PR #567, opened 2020-12-08
Change Avatar props from "Image" to "ImageSource"

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Previously the experimental avatar component took an `Image` prop. This had some complications for passing the prop to a Native Module. Specifically, we need to call `Image.ResolveAssetSource(uri: ...)` on the image to actually resolve it from an ID into an image object that RCTConvert can create a UIImage from. 

Instead, let's change the prop to `imageSource: ImageURISource` and do the resolving call ourselves internally so that clients don't have to. Also, I feel `ImageURISource` is a more common API surface that people expect, and matches what we do in `PersonaCoin`.

### Verification

Updated the test app and converted an image we use in FluentUI Apple to base64 to test.

Screen Shot 2020-12-08 at 4 03 33 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
```diff
@@ -1,18 +1,30 @@
...
+};
```
Saadnajmi, line 13: Because this is so large, we could probably move this to it's own TS file. We want base64 instead of say, a PNG because I was trying to explicitly test base64 support

HeyImChris:
Does this not impact win or other platforms? The file path looks platform agnostic but I'm guessing win just isn't ready to stand up a specific Avatar control since they're still using the JS persona?

HeyImChris (approved):

Saadnajmi:
> Does this not impact win or other platforms? The file path looks platform agnostic but I'm guessing win just isn't ready to stand up a specific Avatar control since they're still using the JS persona?

Experimental Avatar is only implemented on iOS. I would like to combine the PersonaCoin JS implementation into it in the future, but right now it won't work on any other platform.

harrieshin (approved):

PR #572, opened 2020-12-11
Pass the avatar prop "size" properly to "avatarSize" swift property

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Avatar was not properly passing the "size" prop to the native "AvatarSize" property because we used the wrong macro. 
Changed the macro to fix this.

### Verification

Updated the Avatar test page to use "xxLarge" size to show the size difference better.

Screen Shot 2020-12-11 at 9 55 08 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
@@ -16,7 +16,7 @@ const avatar: React.FunctionComponent<{}> = () => {
   return (
     <Stack style={stackStyle}>
       <Text>Name</Text>
-      <Avatar primaryText="Kat Larson" />
+      <Avatar primaryText="Kat Larrson" />
       <Text>Name and Email</Text>
       <Avatar primaryText="Kat Larrson" secondaryText="Kat.Larrson@example.com" />
       <Text>Name, Email, and Image</Text>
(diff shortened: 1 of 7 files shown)


packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -51,7 +51,8 @@ + (MSFAvatarData *)MSFAvatarData:(id)json
 
 @interface RCT_EXTERN_MODULE(MSFAvatarViewManager, RCTViewManager)
 
-RCT_EXPORT_VIEW_PROPERTY(avatarSize, MSFAvatarSize);
```
Saadnajmi, line 54: The macro in question that needed to be updated.
This maps the JS "size" prop to the swift "avatarSize" property.

chiuam: So was "size" no-op in JS?

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
```diff
@@ -31,7 +31,7 @@ const avatar: React.FunctionComponent<{}> = () => {
 
 const stylizedAvatar: React.FunctionComponent<{}> = () => {
   const CustomizedAvatar = Avatar.customize({
-    size: 'large',
+    size: 'xxLarge',
```
chiuam, line 34: nit: it would be nice to show all the sizes in test app.

chiuam (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/AvatarTest.tsx
Saadnajmi: Yes, I think I'll update the test to do that.

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
Saadnajmi: It changed the frame of the view, but didn't actually set AvatarSize. This resulted in having the Avatar of a small size getting stretched out to a bigger size, like so:

Screen Shot 2020-12-11 at 9 48 09 PM

PR #574, opened 2020-12-19
Add support for Icons on Buttons

warren-ms:
This change allows assignment of IconProps to the icon property of Button and give it support for drawing Icon and all of the formats that it supports (otf, ttf, png, svg, etc). 

Also updating the test for Linear and Radial Gradients in SVGs

There are still a couple of issues to be taken as bugs afterwards. Some font icons slightly clip on the edge which I think is an issue on the native side, recoloring via token doesn't seem to work atm.

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/package.json
```diff
@@ -37,7 +37,8 @@
     "@fluentui-react-native/themed-stylesheet": "^1.1.0",
     "@fluentui-react-native/win32-theme": ">=0.5.1 <1.0.0",
     "@fluentui/react-native": ">=0.21.6 <1.0.0",
-    "react-native-svg": "^11.0.0"
+    "react-native-svg": "^11.0.0",
```
HeyImChris, line 40: I was under the impression the react-native-svg community module only supported iOS/Android. Has that changed to somehow include windows support?

warren-ms: I've added native support for it in office win32 netui.

packages/components/Button/src/Button.types.ts
```diff
@@ -81,7 +82,7 @@ export interface IButtonTokens extends FontTokens, IForegroundColorTokens, IBack
   /**
    * Source URL or name of the icon to show on the Button.
    */
-  icon?: string;
+  icon?: number | string | IconProps;
```
JasonVMo, line 85: Does this really need to be in tokens? I guess the use case is setting a default icon value but it does seem a bit strange. (I realize I may have put it there initially)

packages/components/Button/src/Button.types.ts
```diff
@@ -93,7 +94,7 @@ export interface IButtonProps extends Omit<IPressableProps, 'onPress'> {
   /*
    * Source URL or name of the icon to show on the Button.
    */
-  icon?: string;
+  icon?: number | string |IconProps;
```
JasonVMo, line 97: Given this appears twice it might be worth creating an IconPropType and using that in each location.

packages/components/Button/src/Button.types.ts
```diff
@@ -110,7 +111,7 @@ export interface IButtonProps extends Omit<IPressableProps, 'onPress'> {
 export interface IButtonSlotProps {
   root: React.PropsWithRef<IViewWin32Props>;
   stack: ViewProps;
-  icon: ImageProps;
+  icon: string | number | IconProps;
```
JasonVMo, line 114: This should be the prop type of the component hosted in this slot. So if this is an Icon component it should be IconProps. It isn't related to the icon prop on the public interface.

JasonVMo: This is the only comment that I would consider blocking, as it could lead to unexpected bugs like a number getting passed as props into the icon component, which would...well I have no idea, but it likely wouldn't work as expected.

JasonVMo (changes requested):

packages/components/Button/src/Button.types.ts
warren-ms: The issue with that is when you do something like require('.test.png'), what is returned is actually a number that is an index into the AssetRegistry.

JasonVMo: But this is the prop type of the internal component. This is the equivalent of passing the a number instead of a props interface to the Icon component. It is not the type for the icon prop on the button props interface.

warren-ms: Oh I see this is about the slot props. Ok its removed now.

packages/components/Button/src/Button.types.ts
warren-ms: Added now

JasonVMo (approved):

Saadnajmi:
@warren-ms

FYI, I think this broke stuff on the macOS test app because I get a Redbox error when I load the button test.
Screen Shot 2021-03-01 at 5 30 58 PM

warren-ms:
Thanks for the report. I think I'll have to scope the test down by platform. I assumed it was ok since it passed the macos check, but it doesn't seem to run the e2etest on macs. It should just be an issue with the tests and not affect any existing usage. I'll put out a change for the tests shortly.
________________________________

Saadnajmi:
Yeah, the only test we run ATM is checking that `run-ios` and `run-macos` work. It's good followup that we should at least have a test that at least opens every test pane (which will most assuredly fail :D )

PR #576, opened 2020-12-22
Bump react-native-win32 and add pressable text test

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Bump react-native-win32 version which introduces key events, selectable, and tooltip to Text.

Add pressable text test (win32 only)

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/package.json
```diff
@@ -37,6 +37,7 @@
     "@fluentui-react-native/themed-stylesheet": "^1.1.0",
     "@fluentui-react-native/win32-theme": ">=0.5.1 <1.0.0",
     "@fluentui/react-native": ">=0.21.7 <1.0.0",
+    "@office-iss/react-native-win32": "0.62.6",
```
tkamireh, line 40: probably want to hat this to "^0.62.6", idk why it wasn't like that before

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import { Alert, Linking, Platform, View } from 'react-native';
+import { Stack } from '@fluentui-react-native/stack';
+import { Text } from '@fluentui/react-native';
+import { stackStyle } from '../Common/styles';
+import { IKeyboardEvent, IHandledKeyboardEvent } from '@office-iss/react-native-win32';
+
+export const PressableUsage: React.FunctionComponent<{}> = () => {
+  if (Platform.OS == ('win32' as any)) {
```
tkamireh, line 9: Not strictly necessary to platform guard is it? keyDown just won't do anything on the other platform, worth adding a comment perhaps, maybe even a platform check in dev mode inside the Text component itself that logs a warning or error.

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (ev.nativeEvent.key === 'Enter' || ev.nativeEvent.key == ' ') Alert.alert('Alert', 'Success!');
+    };
+
+    const handledNativeKeyboardEvents: IHandledKeyboardEvent[] = [{ key: 'Enter' }, { key: 'Space' }];
+
+    return (
+      <View>
+        <Stack style={stackStyle} gap={5}>
+          <Text variant={'bodyStandard'}>
+            To learn more about microsoft, visit this{' '}
```
tkamireh, line 32: What does the rendered output look like? Is there a formatting thing I'm missing?

apps/win32/package.json
```diff
@@ -60,7 +60,7 @@
     "@wdio/sync": "5.20.1"
   },
   "peerDependencies": {
-    "@office-iss/react-native-win32": "0.62.3",
+    "@office-iss/react-native-win32": "0.62.6",
```
tkamireh, line 63: You may be able to get rid of this, it's not really a peer dep. Not sure it's a dev dep either, though it seems like some things may have broken when removing it from dev deps and placing it as a dep in other places? Any idea what was going on there?

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
ksiler: Are you referencing the `{' '}`? It's just a trailing space on the first part of the sentence. It looks like a regular sentence with an inline link.

packages/components/Button/package.json
```diff
@@ -31,7 +31,7 @@
     "@uifabricshared/foundation-settings": ">=0.8.0 <1.0.0"
   },
   "devDependencies": {
-    "@office-iss/react-native-win32": "0.62.3",
+    "@office-iss/react-native-win32": "0.62.6",
```
tkamireh, line 34: Like the other comment, these probably should be deps (unless we only use the typings and want to keep them opaque)

tkamireh (approved):
Generally looks good, left some questions

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
tkamireh: yes, nice. Text components don't render trailing whitespace? For a second, I thought there was some crazy template literal type stuff going on

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
Saadnajmi: I know that @HeyImChris recently added keyboarding to react-native-macos so maybe it does? He would be better to comment though

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
ksiler: I think they do, it's just added in by the editor

apps/win32/package.json
ksiler: No idea. I just tried it locally and it seems to be fine. I'll move it to dependencies.

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.tsx
ksiler: I see that, looks like it works slightly differently than it does for win32 though. I can check to see if it causes any issues.

tkamireh:
Looks like in fluent-tester-win32, react-native-win32 isn't used as a dep (we should just remove  it from that package).

apps/win32/package.json
ksiler: Actually just started seeing problems now. When removing it, the win32 package will no longer bundle. Adding it back to devDependencies fixes it.

PR #577, opened 2021-01-09
Add beta macOS implementation to experimental-avatar to test arm64 pipeline

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

I need a native module on macOS to test that the static libraries are indeed universal and compiled for x86 and arm64.
Since @chiuam is already working on button, I decided to add a macOS implementation to Avatar...

Note: I don't consider this macOS avatar implementation production ready, and have marked it "Beta" as such. There's a few API differences / hardcoded constants and such that we would need to recitfy. If you'll believe me, I REALLY did add it so I can test the e2e macOS pipeline for arm64. 

This change adds a macOS (native) implementation for `experimental-avatar`, and packages the compiled static library into our nuget package. We also fix a bug on the iOS implementation where a custom backgroundColor was not getting set properly.

### Verification

Ran both iOS and macOS test apps to make sure Avatar displayed properly.

Screen Shot 2021-01-11 at 4 27 12 PM

Screen Shot 2021-01-11 at 4 25 39 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Avatar/macos/AvatarViewManager.swift
```diff
@@ -0,0 +1,27 @@
...
+			"sizes" : [
+				"xSmall" : 16,
+				"small" : 24,
+				"medium" : 32,
+				"large" : 40,
+				"xLarge" : 52,
+				"xxLarge" : 72
+			]
+		]
+	}
```
Saadnajmi, line 26: API difference: 
iOS has these explicit sizes you can set the Avatar too. macOS supports any size. To keep the same API, I just hardcoded sizes for macOS avatar to use and call the constructor with the appropriate CGFloat,

packages/experimental/Avatar/macos/MSFAvatarViewManager.m
```diff
@@ -0,0 +1,34 @@
...
+
+@interface RCT_EXTERN_MODULE(MSFAvatarViewManager, RCTViewManager)
+
+RCT_REMAP_VIEW_PROPERTY(primaryText, contactName, NSString)
+
+RCT_REMAP_VIEW_PROPERTY(secondaryText, contactEmail, NSString)
+
+RCT_REMAP_VIEW_PROPERTY(imageSource, contactImage, UIImage)
+
+RCT_REMAP_VIEW_PROPERTY(backgroundColor, avatarBackgroundColor, UIColor)
```
Saadnajmi, line 14: Fun fact: even though this is macOS, you need to set these as `UIImage` and `UIColor` for it to work properly

harrieshin: can we add comments on this? i cringe a little

packages/experimental/Avatar/macos/AvatarViewManager.swift
harrieshin: yeah, can we bring this up with the design team? Avatar for mac is still bit undefine but there are lots of churn i don't know Taili's team might have more guidance :)

harrieshin:
its interesting that mac avatar background color doesn't match with iOS. i thought we updated the mac avatar background palettes to be the same?

harrieshin (approved):

packages/experimental/Avatar/macos/MSFAvatarViewManager.m
Saadnajmi: Yes.. though lots of controls will have things that maps to colors and images (like in Amanda's change). Is it still worth adding a comment if I have to add it everywhere?

packages/experimental/Avatar/macos/AvatarViewManager.swift
Saadnajmi: There's a figma somewhere with values I haven't looked up.. but that would be a FluentUI Apple change ideally. The purpose of this was really to unblock arm64 testing, so would it be cool to keep this enum here (perhaps with better defined constants) and integrate it back into FUA with the better design once we have it?

packages/experimental/Avatar/macos/MSFAvatarViewManager.m
HeyImChris: Does this just work because UIImage and UIColor are macros defined to the correct platform types?

Saadnajmi:
> its interesting that mac avatar background color doesn't match with iOS. i thought we updated the mac avatar background palettes to be the same?

The palettes are the same, but the hashing algorithm is different. When we did the work to integrate Avatar for it's first client (ToDo) it was requested but ultimately cut. So nothing should block us from "fixing" the hash algorithm to match.

packages/experimental/Avatar/macos/MSFAvatarViewManager.m
HeyImChris: If so, react-native-macos defines various RCT* wrappers around these native types that we could try using instead

.ado/templates/apple-xcode-build-static-libs.yml
```diff
@@ -62,7 +62,7 @@ steps:
       xcode_actions: 'build'
       xcode_scheme: 'FluentTester'
       xcode_configuration: 'Debug'
-      xcode_extraArgs: '-xcconfig $(Build.Repository.LocalPath)/.ado/xcconfig/publish_overrides.xcconfig'
+      xcode_extraArgs: '-xcconfig $(Build.Repository.LocalPath)/.ado/xcconfig/publish_overrides.xcconfig' ONLY_ACTIVE_ARCH=NO
```
Saadnajmi, line 65: So... this totally works, but It would be cleaner to not need to manually add the xcconfig here. I'll follow up with the react-native-test-app team on getting the change integrated upstream

packages/experimental/Avatar/macos/MSFAvatarViewManager.m
Saadnajmi: As in, specify something like RCTUIColor instead? I'll try that.
It seems that UIColor and UIImage map to their corresponding RCTConvert macros, which contain the correct macOS implementation. Both and and @chiuam had problems setting it to just NSColor.

HeyImChris: Yea we need to change that at some point, but probably fine for now as that's out of scope for this review

HeyImChris (approved):

packages/experimental/Avatar/macos/AvatarViewManager.swift
Saadnajmi: I asked Kaan to start, and he said the point values will most likely stay the same. Nevertheless, I filed https://github.com/microsoft/fluentui-apple/issues/391 to track adding this to FluentUI Apple

Saadnajmi:
FYI @harrieshin, I filed https://github.com/microsoft/fluentui-apple/issues/392 for the color mismatch issue

PR #578, opened 2021-01-13
Fix publish pipeline and properly build universal binaries for macOS

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

In my last change (#577) i had a typo where some arguments were outside of colons, causing yml parsing to fail. This caused the publish pipeline to fail. Let's fix that. 

Let's move the build setting to an xcconfig file, and pass that in as an arg to the yaml step. Let's also fix the template for the apple-build step to actually use the extra arg passed in. 

To speed up testing, I moved the macOS debug and release steps to build first, before iOS in the yaml. 

### Verification

Up until the second last commit, I manually added the 'templates/apple-xcode-build-static-libs.yml` as a step to the PR pipeline. You can see as such in the diff of files up to the 2nd last commit here: https://github.com/microsoft/fluentui-react-native/pull/578/files/e0829659213cd48380f664cecfa8a8afe6afa8a9
The Pipeline run is [here](https://github.com/microsoft/fluentui-react-native/pull/578/checks?sha=e0829659213cd48380f664cecfa8a8afe6afa8a9). The pipeline overall failed as one job ran longer than the 60 minutes, but the "build-static-libs" steps did pass otherwise.

I then reversed it in [the following commit](https://github.com/microsoft/fluentui-react-native/pull/578/commits/39a5d1f1f9ae5eeaac3782e3e7a03ee4e4666f3c)

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/templates/apple-xcode-build-static-libs.yml
```diff
@@ -14,62 +14,62 @@ steps:
...
     parameters:
-      xcode_sdk: 'macosx'
-      xcode_workspacePath: 'apps/macos/src/FluentTester.xcworkspace'
+      xcode_sdk: 'iphoneos'
+      xcode_workspacePath: 'apps/ios/src/FluentTester.xcworkspace'
       xcode_actions: 'build'
       xcode_scheme: 'FluentTester'
       xcode_configuration: 'Release'
-      xcode_extraArgs: '-xcconfig $(Build.Repository.LocalPath)/.ado/xcconfig/publish_overrides.xcconfig' ONLY_ACTIVE_ARCH=NO
+      xcode_extraArgs: ''
```
Saadnajmi, line 75: I only moved two steps from the bottom to the top (to build macOS first). I guess because the steps are all so similar, the diff makes it look a lot worse than it is.

chiuam: Is it necessary to revert the order back to building iOS first?

Saadnajmi: It is not necessary, the order does not matter.

chiuam (approved):

PR #580, opened 2021-01-13
Build arm64e for iOS device builds in publish pipeline

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Apple will eventually require `arm64e` (arm64 with pointer authentication) for iOS device apps.
https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication

Let's explicitly build that arch in our pipeline so that the resulting static libs in our NuGet package contain a universal binary with both `arm64` and `arm64e`. We can copy what we do in [FluentUI Apple](https://github.com/microsoft/fluentui-apple/pull/162) and add an extra xcconfig override to the yml steps.

### Verification

I added an extra job in the PR pipeline with the initial commit that will build the static libs. This will be used to verify that CI can do it successfully. I will remove it before checking in this change.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/azure-pipelines.yml
```diff
@@ -111,3 +111,41 @@ jobs:
...
+
+      - script: |
+          yarn bundle macos
+        workingDirectory: $(Build.Repository.LocalPath)/apps/macos
+        displayName: 'yarn bundle macOS'
+
+      # Select proper Xcode version
+      - template: templates/apple-xcode-select.yml
+
+      - template: templates/apple-xcode-build-static-libs.yml
```
Saadnajmi, line 151: This is copied from our publish CI pipeline temporarily to build the static libs in our PR CI pipeline and make sure the steps pass.

Testing CI is annoying and the easiest way I've found to do it so far is to do temporary stuff like this that I revert before check in.

.ado/azure-pipelines.publish.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     pool:
       vmImage: 'macos-10.15'
       demands: ['xcode', 'sh', 'npm']
-    timeoutInMinutes: 60 # how long to run the job before automatically cancelling
+    timeoutInMinutes: 90 # how long to run the job before automatically cancelling
```
HeyImChris, line 46: Having some data on why we need to increase this would be valuable. How long are the arm64e runs taking? How do they compare to the other architectures?

.ado/azure-pipelines.yml
```diff
@@ -111,3 +111,41 @@ jobs:
       - template: templates/apple-yarn-build.yml
         parameters:
           relative_directory: $(relative_directory)
+
+  - job: StaticLibBuild
```
HeyImChris, line 115: Just to be crystal clear- the goal here is to mimic what the publish pipeline does but this NuGet Publish step will be removed after you validate it succeeds right?

HeyImChris (changes requested):

.ado/azure-pipelines.publish.yml
Saadnajmi: It's a little iffy with differences per run (runs can be 2x longer just based on which agent it arbritraily got). I'll try and compile some simple numbers to compare.

.ado/azure-pipelines.yml
Saadnajmi: Yep. I don't want to add this to the PR Pipeline because it takes a while, and we already run `yarn iOS / macOS` which is mostly the same thing. But for PR's where I change the pipeline, I'm trying to be extra careful.

.ado/azure-pipelines.publish.yml
Saadnajmi: A good publish run is about 40 minutes. A bad publish run is about 60 minutes.
Adding the new architectures made the last run 46 minutes (6 minutes longer).
I estimate that a bad run would then be 75 minutes at most (60 + 15 minutes).
With that, I feel 90 is a good upper bound.

HeyImChris (approved):

PR #582, opened 2021-01-15
Update mobile test apps to React Native 0.63.4

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

In preparation for some planned work I have involving React Native 0.63 features (*cough* PlatformColor in theming *cough*), I want to update to update our test apps to RN 0.63.4.

I used the React Native Upgrade helper to know what packages and files to update (https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.4), and made the changes in iOS, android and macOS.
I missed some package updates when I updated the macOS app previously, so I added those back.
I also added "Eslint" as a dependency (not sure why that wasn't already there) for each of the test apps
I also deleted some redundany android source files from when I initially made the test app.

### Verification

macOS and iOS test apps run successfully. Android test app is being weird buuut it compiles so I'll take it.

Screen Shot 2021-01-15 at 11 33 01 AM
Screen Shot 2021-01-15 at 11 46 06 AM

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/package.json
```diff
@@ -20,20 +20,21 @@
   },
   "dependencies": {
     "@fluentui-react-native/tester": "^0.14.0",
-    "react": "16.11.0",
-    "react-native": "0.62.2",
-    "react-native-test-app": "^0.3.2"
+    "react": "16.13.1",
+    "react-native": "0.63.4",
```
HeyImChris, line 24: Just noticed this now but wouldn't we want iOS to still use react-native-macos so it has all our extra fixes we made that haven't made it upstream yet

Saadnajmi: I thought about this earlier too, from when we first moved the test apps to per platform. At the time I thought it made more sense for a public facing repo to prefer the facebook official React Native if possible. I can't really think of a benefit now besides seeing how FURN operates without our extra fixes in + it's easier to update the iOS app to the latest RN without waiting for react-native-macos to catch up. 

I could go either way, but would also do it in a separate change if we decided to go back to react-native-macos

HeyImChris: I think it's more important that this repro presents a consistent baseline to our clients. If iOS is shipping with a different version and therefore different fixes than macOS that's probably not a good message to send. Agree it can be fixed outside this PR, but should probably get addressed soon

HeyImChris (approved):

PR #583, opened 2021-01-18
react-native-fluentui setup for android intial commit

mnkshukl:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [yes ] android

### Pull request checklist

This PR has considered (when applicable): NA
- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Why do we need to name the bundle and app name? Was that causing issues with running the app / solve your issues?

mnkshukl:
> Why do we need to name the bundle and app name? Was that causing issues with running the app / solve your issues?

Mainly app.json and package.json bundle name should be same for making the bundle clubbed into assets.
Names changes are just part of that

apps/android/package.json
```diff
@@ -44,7 +44,7 @@
     ],
     "entry": "index.js",
     "outputPath": "./dist",
-    "bundleName": "index.bundle",
+    "bundleName": "index.android.bundle",
     "noPlatformSuffix": true
```
Saadnajmi, line 48: I think our tooling is set up so that you can set `noPlatformSuffix` to false if you want to include the platform in the bundle.

Saadnajmi (changes requested):

apps/android/package.json
mnkshukl: it adds the suffix at last like index.bundle.android, so falling back to original naming convention

apps/android/package.json
```diff
@@ -12,7 +12,8 @@
     "code-style": "fluentui-scripts code-style",
     "depcheck": "fluentui-scripts depcheck",
     "lint": "fluentui-scripts eslint",
-    "start": "react-native start",
+    "start": "react-native start --reset-cache",
+    "start --reset-cache": "react-native start --reset-cache",
```
Saadnajmi, line 16: You don't need to include the arguments to the key. With just the key "start" you can run "yarn start --reset-cache", and it will pass the argument to the script (`react-native start`)

Saadnajmi: Also, I think pakage.json doesn't like it when the key for a script has a space? Every other script I've seen has had no spaces

Saadnajmi (approved):

Saadnajmi:
Not sure if you have merging privileges, but I can merge this if you don't @mnkshukl

mnkshukl:
> Not sure if you have merging privileges, but I can merge this if you don't @mnkshukl

I don't have right now , Who can provide the merge access  ?

Saadnajmi:
Probably @FalseLobster is the one to talk to, also @ejlayne . I'll merge this one

PR #584, opened 2021-01-24
Add an Apple Theme

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We would like there to be an Apple Theme that uses the iOS / macOS semantic colors (using the PlatformColor API) and also the FluentUI Apple palette. We also want to use the typography weights / font sizes specified in the Apple HIG.

This change adds a new package, "@fluentui-react-native/apple-theme" that accomplishes that. It is heavily pattern matched off the default theme package. It also requires RN 0.63 as a minimum because it makes heavy use of the PlatformColor API. 

This change also adds that theme as selectable in FluentTester. That bit is subject to change as the theme breaks some assupmtions FluentTester had (for example: the light/dark picker does nothing since PlatformColor does not listen to the react native appearance API).

We also make a change to the NativeButton component to map the token "accentBackground" to the accentColor of the button. This isn't really part of the apple theme change, but a nice addition nonetheless.

### Verification

Ran the macOS test app with the Apple theme and a custom brand.
Screen Shot 2021-01-21 at 4 42 51 PM

Also checked the JS Button
Screen Shot 2021-01-20 at 2 50 57 PM

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/theming/apple-theme/src/appleSemanticColors.macos.ts
```diff
@@ -0,0 +1,178 @@
...
+  gridColor: ColorValue;
+
+  /** The color to use for text in header cells in table views and outline views */
+  headerTextColor: ColorValue;
+
+  /** The colors to use for alternating (odd Rows) content, typically found in table views and collection views */
+  alternatingOddContentBackgroundColor: ColorValue;
+
+  /** The colors to use for alternating (even Rows)content, typically found in table views and collection views */
+  alternatingEvenContentBackgroundColor: ColorValue;
```
chiuam, line 77: Like we discussed offline, I couldn't find these colors in the Apple doc. Should we remove them or just map them to alternatingContentBackgroundColor?

packages/theming/apple-theme/src/appleTheme.colors.ios.ts
```diff
@@ -0,0 +1,223 @@
...
+    disabledBodyText: appleSemantic.quaternaryLabel,
+    disabledSubtext: appleSemantic.tertiaryLabel,
+    disabledBodySubtext: appleSemantic.quaternaryLabel,
+
+    focusBorder: 'transparent',
+    variantBorder: appleSemantic.separator,
+    variantBorderHovered: appleSemantic.separator,
+    defaultStateBackground: appleSemantic.systemFill,
+
+    // TODO
```
chiuam, line 31: Add documentation to the TODOs in this file?

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
```diff
@@ -0,0 +1,223 @@
...
+    buttonContent: appleSemantic.controlTextColor,
+    buttonIcon: appleSemantic.controlTextColor,
+
+    // macOS Button has no hover effect
+    buttonHoveredBackground: fluentApple.communicationBlue,
+    buttonHoveredBorder: 'transparent',
+    buttonHoveredContent: appleSemantic.controlTextColor,
+    buttonHoveredIcon: appleSemantic.controlTextColor,
+
+    // TODO what does focus even mean on macOS?
```
chiuam, line 120: Like via keyboard focus?

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
```diff
@@ -0,0 +1,223 @@
...
+    brandPressedContent: appleSemantic.controlTextColor,
+    brandPressedIcon: appleSemantic.controlTextColor,
+
+    // TODO System Effect
+    brandDisabledBackground: fluentApple.gray100,
+    brandDisabledBorder: 'transparent',
+    brandDisabledContent: appleSemantic.controlTextColor,
+    brandDisabledIcon: appleSemantic.controlTextColor,
+
+    // TODO What is checked?
```
chiuam, line 191: Checkbox / toggle button?

packages/theming/apple-theme/src/fluentAppleColors.macos.ts
```diff
@@ -0,0 +1,490 @@
...
+    neutralBackground2: '#FFFFFF',
+    neutralBackground3: '#000000',
+    neutralForeground2: '#000000',
+    neutralForeground3: '#272727',
+    neutralInverted: '#FFFFFF',
+    neutralStroke2: '#000000',
+  };
+}
+
+export function getStockAppleDarkPalette(): ApplePalette {
```
chiuam, line 405: Why do we need separate functions for light and dark palettes? (unlike iOS)

packages/theming/apple-theme/src/fluentAppleColors.macos.ts
```diff
@@ -0,0 +1,490 @@
...
+    successTint30: '#043615',
+    successTint40: '#021D0B',
+    warningPrimary: '#FFC328',
+    warningShade10: '#FFC83E',
+    warningShade20: '#FFDD15',
+    warningShade30: '#FFDD87',
+    warningTint10: '#E0AB24',
+    warningTint20: '#997518',
+    warningTint30: '#4D3A0C',
+    warningTint40: '#291F07',
```
chiuam, line 479: Can we fetch the colors from FUA instead of hardcoding them?

chiuam:
Can the JS button title color be themed as well?

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
```diff
@@ -13,7 +13,7 @@ export const lightnessOptions = [
 ];
 
 export class TesterThemeReference extends ThemeReference {
-  private _themeName: ThemeNames = 'Fluent';
+  private _themeName: ThemeNames = 'Apple';
```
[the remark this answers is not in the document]
Saadnajmi: Yes.. this is where I would like @FalseLobster 's input. This is intended to be the default theme on iOS/macOS, but it differs from the already-defined default theme that uses the Fabric Web Palette. I could pass the theme when we initialize FluentTester in each test app at index.ts, but I'm not sure if I need to change the picker options in the drop down then?

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
```diff
@@ -0,0 +1,223 @@
...
+    buttonBackgroundHovered: fluentApple.communicationBlue,
+    buttonBackgroundCheckedHovered: fluentApple.communicationBlueTint10,
+    buttonBackgroundPressed: fluentApple.communicationBlueTint10,
+    buttonBackgroundDisabled: fluentApple.gray100,
+    buttonText: appleSemantic.controlTextColor,
+    buttonTextHovered: appleSemantic.controlTextColor,
+    buttonTextChecked: appleSemantic.controlTextColor,
+    buttonTextCheckedHovered: appleSemantic.controlTextColor,
+    buttonTextPressed: appleSemantic.controlTextColor,
+    buttonTextDisabled: appleSemantic.controlTextColor,
```
Saadnajmi, line 62: @chiuam The JS button text is themed with tokens from here, along with similar tokens in the ControlColorTokens section

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
Saadnajmi: Yes. I made the assumption that would match the "pressed" state, but would love to hear other opinions from you or @harrieshin or designers or whoever else

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
Saadnajmi: I think so. I think Focus does something different on windows than macOS. On macOS (and I assume iOS..) we add the blue rectangle around the control, but otherwise the control looks at rest. I think on windows it takes on a new state like hovered or pressed.

I made the assumption that all the "Focused" state tokens should match rest. Again would love to hear opinions from others like @HeyImChris or @harrieshin or designers or such

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
```diff
@@ -0,0 +1,223 @@
...
+    link: appleSemantic.linkColor,
+    linkHovered: appleSemantic.linkColor,
+    linkPressed: appleSemantic.selectedControlColor,
+
+    /* ControlColorTokens */
+
+    // Default values without any style
+    buttonBackground: fluentApple.communicationBlue,
+    buttonBorder: 'transparent',
+    buttonContent: appleSemantic.controlTextColor,
```
Saadnajmi, line 111: @chiuam buttonContent refers to the text for the experimental JS button

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
```diff
@@ -0,0 +1,223 @@
...
+    ghostPressedSecondaryContent: appleSemantic.secondaryLabelColor,
+
+    brandSecondaryContent: appleSemantic.secondaryLabelColor,
+    brandFocusedSecondaryContent: appleSemantic.secondaryLabelColor,
+    brandHoveredSecondaryContent: appleSemantic.secondaryLabelColor,
+    brandPressedSecondaryContent: appleSemantic.secondaryLabelColor,
+
+    buttonDisabledSecondaryContent: appleSemantic.tertiaryLabelColor,
+    buttonHoveredSecondaryContent: appleSemantic.tertiaryLabelColor,
+    buttonPressedSecondaryContent: appleSemantic.tertiaryLabelColor,
```
Saadnajmi, line 221: These mappings are somewhat iffy, since we don't really have an equivalent of "secondary" for any of the FluentUI Apple controls

packages/theming/apple-theme/src/fluentAppleColors.macos.ts
Saadnajmi: That was my original intention, but I think it ends up being more indirection than it's worth. If we fetch them from FUA, we introduce a cocoapod dependency. We would still need to update the pod to increment the colors or add mappings for new colorsm so it's not really that we get the newest colors for free. Furthermore, this whole palette interface is subject to change as we move to the design token pipeline, and it's easier to rewrite and verify JS code in one repo, than JS and native code in two repos.

packages/theming/apple-theme/src/appleSemanticColors.macos.ts
Saadnajmi: The choice to separate the one macOS semantic color into two react native PlatformColors was done by whoever designed the PlatformColor API for react-native-macos. The intention of the `appleSemanticColors.<platform>.ts` files were to simply expose all the available PlatformColors, regardless of if we mapped it to any Token or Palette or not. The mappings of AppleColor->Palette happen in `appleTheme.colors.<platform>ts`.

Rather than remove the two colors from here, I would rather we choose not to use them in `appleTheme.colors.macos.ts` and use something else for list backgrounds, since it's confusing.

packages/theming/apple-theme/src/appleTheme.colors.ios.ts
Saadnajmi: I should just remove all of these. They were there while I was still mapping all the colors but now that I have a complete set, I can probably just remove them, and let people comment on the mappings

packages/theming/apple-theme/src/fluentAppleColors.macos.ts
Saadnajmi: Ah... we don't need the light and dark palettes anymore. That was an intermediary step while I was debugging. They can be fully removed.

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
Saadnajmi: Yep. I assuemd that the tokens would match their state at rest, and the system would just add the keyboard focus rectangle.

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
harrieshin: for the tokens that doesn't make sense to the platform, I am curious to hear how to no-op it or flag it somehow like "this token doesn't make sense for the platform. don't use"

Saadnajmi: Even if none of our FluentUI Apple controls have "secondary" content on them, there are JS components that do, so we would need to set something on them.

To not use specific tokens on a platform would need us to either:
- Don't set the tokens at all. This means they inherit from whatever the parent theme is, which could lead to unintended styling, and seems bad.
- Adding platform specific implementations of those JS controls that don't use the offending tokens. This seems like a bad idea because differing the API of a compoment that much may get confusing
- Mapping the offending tokens to some other token (like buttonBackgroundHover maps to buttonBackgroundRest). This is what I settled on to ensure that we control the styling to do what we want.

packages/theming/apple-theme/src/fluentAppleColors.macos.ts
Saadnajmi: Edit: Will need to make a native theming module to get access to the [macOS System effects API](https://developer.apple.com/documentation/appkit/nscolor/2998826-withsystemeffect). Will leave as a follow up

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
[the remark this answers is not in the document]
Saadnajmi: An update here: 
Win32/windows CI fail because `FluentTester` takes a dependency on `apple-theme`, which takes a dependency on `react-native-macos`, which doesn't import cleanly into windows/win32. 

Ideally, FluentTester should not need to know about the apple theme, and we should just pass it in via the macOS test app. I have not yet found a way to do this, and will need to modify FluentTester to take that in as a prop or something like that...

apps/fluent-tester/src/FluentTester/testPages.macos.ts
```diff
@@ -4,6 +4,7 @@ import { ButtonTest, HOMEPAGE_BUTTON_BUTTON } from './TestComponents/Button';
 import { CalloutTest, HOMEPAGE_CALLOUT_BUTTON } from './TestComponents/Callout';
 import { CheckboxTest, HOMEPAGE_CHECKBOX_BUTTON } from './TestComponents/Checkbox';
 import { ContextualMenuTest, HOMEPAGE_CONTEXTUALMENU_BUTTON } from './TestComponents/ContextualMenu';
+import { ExperimentalButtonTest, HOMEPAGE_BUTTON_BUTTONEXPERIMENTAL } from './TestComponents/ButtonExperimental';
```
HeyImChris, line 7: Why the `BUTTON` in the middle? It looks like most have that at the end after the module name

packages/theming/apple-theme/README.md
```diff
@@ -0,0 +1,10 @@
+# Apple-theme
+
+Code and definitions for creating an Apple Theme for FluentUI React Native. Currently only implemented for macOS.
```
HeyImChris, line 3: Probably worth mentioning the bug tracking iOS progress

packages/theming/apple-theme/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+Some heuristics followed in this theme:
+
+- For most tokens, we default to prefer the apple system colors, using the FluentUI Apple Palette where it makes sense. These mappings are subject to change as we increment on the design.
+- On macOS, there tends to not be a "hover" state for components such as Button, so those tokens are mapped to be identical to the rest state tokens.
```
HeyImChris, line 8: Maybe to make this a little cleaner we can have a macOS and iOS (currently empty) section of platform specific information

packages/theming/apple-theme/src/appleTheme.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { appleTypography } from './appleTheme.typography';
+
+export function appleSpacing(): Spacing {
+  return { s2: '4px', s1: '8px', m: '16px', l1: '20px', l2: '32px' };
+}
+
+export const appleComponents = {
+  Button: {
+    tokens: {
```
HeyImChris, line 15: Where do these tokens come from? Wouldn't it vary between iOS/macOS?

packages/theming/apple-theme/src/appleTheme.ts
```diff
@@ -0,0 +1,30 @@
+import { Spacing, Theme } from '@fluentui-react-native/theme-types';
+import { getFluentUIApplePalette } from './fluentAppleColors';
+import { getAppleSemanticPalette } from './applePlatformColors';
+
+import { paletteFromAppleColors } from './appleTheme.colors';
+
+import { appleTypography } from './appleTheme.typography';
+
+export function appleSpacing(): Spacing {
+  return { s2: '4px', s1: '8px', m: '16px', l1: '20px', l2: '32px' };
```
HeyImChris, line 10: Where do these values come from/how are they used?

packages/theming/apple-theme/src/appleTheme.typography.ts
```diff
@@ -0,0 +1,55 @@
+import { FontSize, FontSizes, FontWeightValue, Spacing, Typography, Variants } from '@fluentui-react-native/theme-types';
+
+export function appleTypography(): Typography {
+  const appleDict = {
+    sizes: {
```
HeyImChris, line 5: These values should really be pulled from Appkit/UIKit directly. The way it is now, when Apple inevitably tweaks these values, we're going to have to come in here and redo all this work.  I think the best design would be to have a native module that reads them directly from the OS APIs. I implemented [something similar](https://office.visualstudio.com/ISS/_git/sdx-platform?path=%2Fpackages%2Fpolyester%2Fapple%2Fpolyester%2Fmac%2FPLYColorPalette_mac.m&_a=contents&version=GBmaster) in Polyester that seems pretty analogous to what we want here

packages/theming/apple-theme/src/fluentAppleColors.ts
```diff
@@ -0,0 +1,316 @@
...
+  neutralBackground3: ColorValue;
+  neutralForeground2: ColorValue;
+  neutralForeground3: ColorValue;
+  neutralInverted: ColorValue;
+  neutralStroke2: ColorValue;
+}
+
+export type FluentUIApplePalette = FluentUIAppleBasePalette & FluentAppleButtonTokens;
+
+export function getFluentUIApplePalette(): FluentUIApplePalette {
```
HeyImChris, line 95: Do these need to be hardcoded or can we pull them from OS APIs as well?

HeyImChris (changes requested):

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: This is organized alphabetically. Because this starts with E, it's in the middle. 

Also, there are now 3 button test pages: Button (V1 of compose framework), ExperimentalButton (the new experimental compose framework), and NativeButton (also using the experimental compose framework)

packages/theming/apple-theme/src/appleTheme.ts
Saadnajmi: They would. A few of these pages were initially split out between macOS and iOS before I reverted the iOS changes and just had the macOS changes. As such, all the code you see now is intended for macOS and will fail on iOS

```
export function createAppleTheme(): ThemeReference {
  const appleTheme = Platform.select({
    macos: () => new ThemeReference(defaultAppleThemeMacOS),
    ios: () => {
      console.log('iOS Theme not currently supported, using default theme');
      return new ThemeReference(createDefaultTheme());
    },
  })();
  return appleTheme;
 ```

File: packages/theming/apple-theme/src/appleTheme.ts (hunk 15 above)
Comment (Saadnajmi) [reply packages/theming/apple-theme/src/appleTheme.ts:10]: These values are copied from the default theme. AFAIK, they aren't really used anywhere yet, though @falselobster can correct me if I'm wrong. 

We align to a a 4px grid anyway, so these felt fine...

File: apps/fluent-tester/src/FluentTester/testPages.macos.ts (hunk 11 above)
Comment (HeyImChris) [reply apps/fluent-tester/src/FluentTester/testPages.macos.ts:7]: Hah I know it's alphabetized :) I meant the word BUTTON in the middle of the page name. I was asking why not make it `HOMEPAGE_BUTTONEXPERIMENTAL`

File: packages/theming/apple-theme/src/appleTheme.typography.ts (hunk 16 above)
Comment (Saadnajmi) [reply packages/theming/apple-theme/src/appleTheme.typography.ts:5]: These sizes are not actually coming from Appkit, but from the [Fluent vNext macOS Figma](https://www.figma.com/file/RKqH3UfoIZpO1ZZIlMSp0j/Fluent-vNext-macOS). They are designed to be close to the macOS HIG, but can't quite match it as we haven't moved to a unified type ramp between the different FLURN platforms yet.

Agree that we should pull values like Colors from a native module. for V1 I didn't want to add the extra complexity of that just yet, and I think typography will end up not coming from a native module since they will actually come from design tokens.

File: apps/fluent-tester/src/FluentTester/testPages.macos.ts (hunk 11 above)
Comment (HeyImChris) [reply apps/fluent-tester/src/FluentTester/testPages.macos.ts:7]: Or `HOMEPAGE_BUTTONEXPERIMENTAL_BUTTON` since all the other ones have button at the end

File: packages/theming/apple-theme/src/fluentAppleColors.ts (hunk 17 above)
Comment (Saadnajmi) [reply packages/theming/apple-theme/src/fluentAppleColors.ts:95]: These don't come from the OS, they're copied from the FluentUI Apple Asset Catalog. This could be a native module, but again, I avoided the extra complexity of a native module for V1. Furthermore, we don't * need * most of these colors, and they'll either end up removed, or shared as we move to Fluent vNext.

File: apps/fluent-tester/src/FluentTester/testPages.macos.ts (hunk 11 above)
Comment (Saadnajmi) [reply apps/fluent-tester/src/FluentTester/testPages.macos.ts:7]: Ah, my bad! I didn't write these so I dunno haha. Probably weirdness when they went from V1 -> Experimental and had a bunch of duplicate implementations of controls

File: packages/theming/apple-theme/src/appleTheme.typography.ts (hunk 16 above)
Comment (Saadnajmi) [reply packages/theming/apple-theme/src/appleTheme.typography.ts:5]: I do realize I explicitly wrote that they come from the HIG in the description though, ha. I should probably clarify that somewhere

File: packages/theming/apple-theme/src/appleTheme.ts (hunk 15 above)
Comment (harrieshin) [reply packages/theming/apple-theme/src/appleTheme.ts:10]: follow up with @rdeassis if spacing has already been tokenized in Apple. I think font and colors are but not sure about spacing...

Comment (Saadnajmi) [reply packages/theming/apple-theme/src/appleTheme.ts:10]: Looking [here](https://github.com/microsoft/fluentui-apple/blob/88073f40a14fb15af56688ab6b54dad91fb5d47d/tools/sgen/input/FluentUIStyle.yml#L147), spacing is defined as such:

```
Spacing:
  none: 0pt
  xxxSmall: 2pt
  xxSmall: 4pt
  xSmall: 8pt
  small: 12pt
  medium: 16pt
  large: 20pt
  xlarge: 24pt
  xxlarge: 36pt
  xxxlarge: 48pt
  xxxxlarge: 72pt
  ````
  
  It's not... exactly clear how these two line up but really the shape of the Theme interface should just evolve to match the design token pipeline that we have so I don't have to guess these equivalences in the future..

apps/fluent-tester/src/FluentTester/testPages.macos.ts
FalseLobster: Random aside--- I'm not crazy about these names, to be honest.  The coding style looks out of place on JS.  I feel like HomepageButtonExperimentalButton would look more normal, not that you have to fix it.  And, if I remember correctly, it's not really a button its more of an ID...

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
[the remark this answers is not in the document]
FalseLobster: we can use metro's per-platform module resolution to change the behavior of our fluent theme.  you could imagine a fluentTheme.macos.ts that calls createAppleTheme() and a fluentTheme.win32.ts that calls createOfficeTheme() or whatever.

packages/theming/apple-theme/src/appleTheme.ts
FalseLobster: Yea I think that's an accurate assessment.  This interface may be sort of a prototype/placeholder in absence of a fully spec'd out token pipeline.

packages/theming/apple-theme/src/appleTheme.typography.ts
FalseLobster: Generally speaking I've found it helpful for our consumers to have access to native ideas in addition to what we're (eventually going to be) consuming through the token pipeline, fwiw.  No objection to tackling the problem later though

packages/theming/apple-theme/src/fluentAppleColors.ts
```diff
@@ -0,0 +1,316 @@
...
+  return {
+    blue10: '#4F6BED',
+    blueMagenta20: '#8764B8',
+    blueMagenta30: '#5C2E91',
+    communicationBlue: DynamicColorMacOS({
```
FalseLobster, line 100: How does DynamicColorMacOS work, out of curiosity?  Is this like PlatformColor where it's handled natively or is this using the appearance module (I think it's called) to do this dynamically from the JS-side?

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -25,7 +25,7 @@ export interface Theme {
   spacing: Spacing;
   host: {
     // appearance of the theme, this corresponds to the react-native Appearance library values, though can be overwritten
-    appearance: 'light' | 'dark';
+    appearance: 'light' | 'dark' | 'auto';
```
FalseLobster, line 28: Should this go into name?: ?

packages/theming/apple-theme/src/fluentAppleColors.ts
FalseLobster: I'm all for doing what's most ergonomic, but it might also be good to lean into shared patterns where appropriate to make it easier to interpret.

packages/theming/apple-theme/src/appleTheme.ts
FalseLobster: You could define these in separate files (i.e. theme.macos.ts and theme.ios.ts) if it makes more sense.  I don't know how much cross over there will be

packages/theming/apple-theme/src/fluentAppleColors.ts
Saadnajmi: I think DynamicColorMacOS is simply analogous to [DynamicColorIOS](https://reactnative.dev/blog/2020/07/06/version-0.63#native-colors-platformcolor-dynamiccolorios) which was introduced with 0.63. I think it handles everything natively and supersedes the appearance API. I'm not opposed to just having a light/dark palette that the theme switches as that may be easier. However, I did notice PlatformColor() supersedes the react-native appearance API and will just follow whatever the system has

packages/theming/theme-types/src/Theme.types.ts
Saadnajmi: While using PlatformColor(), the theme seemed to not use the react-native appearance API at all, so the Light/Dark picker we had did nothing and it just followed the system. 

I think I wanted to add this so that I could then have the picker be disabled if it saw the theme handled it's own changes, but I never added the code to disable the picker/check

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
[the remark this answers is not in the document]
Saadnajmi: I thought that the Fluent theme in this context was just a fancy name for the default theme though? To which it would then become having the default theme return the new apple theme if on ios or macos, which didn't seem like a good idea

packages/theming/apple-theme/src/appleTheme.colors.macos.ts
FalseLobster: I think the way you solved this makes sense.

packages/theming/apple-theme/src/appleTheme.ts
Saadnajmi: I'll leave that as a task for when we add the iOS theme.
The tokens here came from looking at the design of the [FluentUI Apple macOS button](https://github.com/microsoft/fluentui-apple/blob/4b9992b3bbeb46776de86d0c8c14ac3c08a207f0/macos/FluentUI/Button/Button.swift#L584)

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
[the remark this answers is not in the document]
Saadnajmi: Platform specific files fixed this! Thanks again!

packages/theming/apple-theme/README.md
Saadnajmi: Most of the iOS information would be the same  (no focus state, prefer system colors), so I'll leave that as a task for when we add the iOS theme

packages/theming/apple-theme/src/appleTheme.ts
Saadnajmi: I changed this to match the type ramp from FluentUI Apple vNext (assuming s2 corresponds to xSmall, etc). This still isn't used anywhere, but at least it matches :D

packages/theming/apple-theme/README.md
Saadnajmi: Created Github #596 to track publicly

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: I'm going to leave this as not part of this review, more changes than I would like :D

packages/theming/apple-theme/src/appleTheme.ts
Saadnajmi: Decided to split the rest of the fies by platform, even if we just have a macOS theme.

packages/theming/apple-theme/README.md
```diff
@@ -0,0 +1,10 @@
+# Apple-theme
+
+Code and definitions for creating an Apple Theme for FluentUI React Native. Currently only implemented for macOS. iOS Apple theme progress is tracked at Github #596
+
+Some heuristics followed in this theme:
+
+- For most tokens, we default to prefer the apple system colors, using the FluentUI Apple Palette where it makes sense. These mappings are subject to change as we increment on the design.
+- On macOS, there tends to not be a "hover" state for components such as Button, so those tokens are mapped to be identical to the rest state tokens.
```
HeyImChris, line 8: "rest _of the_ state tokens"? Or is a `rest state token` some token I don't know of

HeyImChris: If it's the opposite of a pressed state, usually Apple refers to that as the `normal` state

packages/theming/apple-theme/src/appleTypography.macos.ts
```diff
@@ -0,0 +1,57 @@
+import { FontSize, FontSizes, FontWeightValue, Typography, Variants } from '@fluentui-react-native/theme-types';
+
+/**
+ * The Typography is designed to match the styles defined in the Apple HIG:
+ * https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/
+ * These mappings and variants are subject to change as we moved to a unified cross platform Fluent typography ramp
+ */
+
+export function appleTypography(): Typography {
+  const appleDict = {
```
HeyImChris, line 10: Is there a bug tracking these issues coming from native components? Probably worth follow up to make it robust so we never have issues with it again instead of tweaking these values for every release :)

HeyImChris (approved):

packages/theming/apple-theme/README.md
Saadnajmi: rest state as in normal state. I'll add clarification that this is the "normal" state on apple Platforms

packages/theming/apple-theme/src/appleTypography.macos.ts
Saadnajmi: #598 filed, adding as a comment as well

PR #586, opened 2021-01-28
setup Basic Android azure pipeline

mnkshukl:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [yes] android

### setup Android azure basic pipeline which setups basics android bundling and assemble aar and run UI tests


.ado/azure-pipelines.yml
```diff
@@ -111,3 +111,44 @@ jobs:
...
+          relative_directory: 'android'
+
+    steps:
+      - checkout: self
+        persistCredentials: true
+
+      # Clean Derived Data
+      - script: |
+          rm -rf $(Build.Repository.LocalPath)/DerivedData
+        displayName: 'Clean DerivedData'
```
Saadnajmi, line 134: This is Apple specific, you don't need this

.ado/azure-pipelines.yml
```diff
@@ -111,3 +111,44 @@ jobs:
...
+    displayName: Android PR
+    pool:
+      vmImage: 'ubuntu-latest'
+    timeoutInMinutes: 60 # how long to run the job before automatically cancelling
+    cancelTimeoutInMinutes: 5 # how much time to give 'run always even if cancelled tasks' before killing them
+    strategy:
+      maxParallel: 4
+      matrix:
+        android:
+          relative_directory: 'android'
```
Saadnajmi, line 125: why is this a matrix?

apps/android/app.json
```diff
@@ -2,5 +2,5 @@
   "name": "FluentTester",
   "displayName": "FluentTester",
   "components": [{"appKey": "FluentTester"}],
-  "resources": ["dist/res", "dist/main.jsbundle"]
+  "resources": ["dist/res", "dist/index.bundle"]
```
Saadnajmi, line 5: Good catch...

.ado/templates/android-dep-setup.yml
```diff
@@ -0,0 +1,5 @@
+steps:
+  - task: Bash@3
+    inputs:
+      targetType: 'inline'
+      script: echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'ndk;21.0.6113669'
```
Saadnajmi, line 5: Is there not an equivalent azure pipeline task to do this?

Saadnajmi (commented):
For the iOS and macOS apps, we set up a basic pipeline which more or less just makes sure "react-native run-ios/macos" works. There's certainly more we can do, but that was our start. What was the goal with this pipeline?

.ado/templates/android-dep-setup.yml
mnkshukl: default Agents come with specific NDK version 22.x.x, which does not fit for latest android gradle as well.
There are no azure tasks I could find to download them as well.

.ado/templates/android-dep-setup.yml
```diff
@@ -0,0 +1,40 @@
...
+        
+        # Install AVD files
+        echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'system-images;android-27;google_apis;x86'
+        
+        # Create emulator
+        echo "no" | $ANDROID_HOME/tools/bin/avdmanager create avd -n android_emulator -k 'system-images;android-27;google_apis;x86' --force
+        
+         # list emulator
+        echo "Available emulator"
+        $ANDROID_HOME/emulator/emulator -list-avds
```
Saadnajmi, line 29: Why do we need to list the emulator? To make it easier to debug when you're looking at logs of previous runs?

.ado/templates/android-dep-setup.yml
```diff
@@ -0,0 +1,40 @@
...
+    inputs:
+      targetType: 'inline'
+      script: |
+        #!/usr/bin/env bash
+        
+        # Install AVD files
+        echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'system-images;android-27;google_apis;x86'
+        
+        # Create emulator
+        echo "no" | $ANDROID_HOME/tools/bin/avdmanager create avd -n android_emulator -k 'system-images;android-27;google_apis;x86' --force
```
Saadnajmi, line 25: Mostly out of curiosity, whats the minimum API we (Microsoft / Office) plans to support?
For macOS it's N-2 from the latest (so macOS 10.14, 10.15, and 11)
for iOS, it's N-1 (iOS 13 and 14)

.ado/azure-pipelines.yml
```diff
@@ -110,4 +110,39 @@ jobs:
...
+    
+      # build android apk and run available Intrumented test
+      - task: Gradle@2
+        inputs:
+          gradleWrapperFile: 'apps/$(relative_directory)/src/gradlew'
+          workingDirectory: 'apps/$(relative_directory)/src'
+          tasks: 'assembleDebug cAT'
+          publishJUnitResults: false
+          javaHomeOption: 'JDKVersion'
+          sonarQubeRunAnalysis: false
```
Saadnajmi, line 148: Could we add a step that also runs `yarn android`, similar to what we do in `templates/apple-yarn-build.yml` ? That is what we list on the repo's README as the way to launch the android app and I assume is the way newcomers would try and run the android test app. Running that command (which just invokes `react-native run-android`) should also invoke gradle in the process

.ado/templates/android-dep-setup.yml
mnkshukl: right now we are aiming to support min(office/outlook) which is 23
it is a bit complicated in android , and there is no thumb rule across.

.ado/templates/android-dep-setup.yml
mnkshukl: just incase pipeline fails it is can be tracked , if emulator was installed well.
As in next step we wait for emulator to start and up running before building

.ado/azure-pipelines.yml
mnkshukl: assemble command loads the bundle created by bundle command and also is more reliable and faster than  as we don't need for NPM server to load bundle (which takes time before starting any UI tests.)  Your Thoughts ?

Saadnajmi: I see, that makes sense. I guess we used the run-ios/macos as an initial test case before we wrote any other UI tests. 
If the assemble command is enough to guarantee that `run-android` also works (since it's in our readme) that would be great =).

Saadnajmi (approved):

PR #609, opened 2021-02-23
Android tester app

tamasane:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

- The original FluentTester app was not friendly to mobile screens and it made difficult to test controls on small screen. Moreover, the test pages contained additional Platform status data which made added too much information on the screen.
- The idea was to create separate files for mobile platforms which further developed into creating a shared file with rendering based on platform checks.
- Instead of using the two panes layout, the newer layout is a two pager layout where the first screen shows a list of available controls, which clicked upon load their corresponding Test page. The idea is to share the logic for both Android/iOS mobile platforms but this PR considers the change only for Android.

### Verification

The app is tested on a phone and tablet layout Android emulator.

|  | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Ooh.. This PR has been a long time coming. I'll do my best to review it soon =)

ejlayne:
@tamasane looks like you may need to run "yarn change" to fix the CI errors you're getting.

apps/fluent-tester/src/FluentTester/TestComponents/Common/styles.ts
```diff
@@ -131,3 +131,16 @@ export const separatorStackStyle: IStackProps['style'] = {
...
+
+export const mobileStyles = StyleSheet.create({
+  testListItems: {
+    width: '100%',
+    height: 'auto',
+    fontSize: 18,
+    paddingVertical: 8,
+  },
+  testListSeparator: {
+    borderColor: '#e1e1e1',
```
Saadnajmi, line 143: nit: I would rather not place this value (gray100) here and rather have it specified via tokens.

apps/fluent-tester/src/FluentTester/TestComponents/Test.tsx
```diff
@@ -54,6 +54,8 @@ const styles = StyleSheet.create({
   },
 });
 
+const isMobile = Platform.OS == 'android';
```
Saadnajmi, line 57: Can you add iOS to this as well? Or just make an issue tracking adding iOS here as well and add the comment so that I or someone else can do that?

apps/fluent-tester/src/FluentTester/TestComponents/Test.tsx
```diff
@@ -64,55 +66,57 @@ export const Test = (props: TestProps) => {
...
-              iOS: <Text style={styles.status}>{props.status.iosStatus}</Text>
-            </Text>
-            <Text style={[styles.statusLabel]} variant="bodySemibold">
-              macOS: <Text style={styles.status}>{props.status.macosStatus}</Text>
-            </Text>
-            <Text style={[styles.statusLabel]} variant="bodySemibold">
-              Android: <Text style={styles.status}>{props.status.androidStatus}</Text>
-            </Text>
-          </Stack>
+      {isMobile ? null : (
```
Saadnajmi, line 69: Rather than define it null here, could we just do the platform check wherever we render this component and not render it for mobile paltforms?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -18,7 +18,7 @@ export const themePickerStyles = StyleSheet.create({
 
   dropdown: {
     height: 30,
-    width: 90,
+    width: 120,
```
Saadnajmi, line 21: This would affect all platforms, so you might want to make this android specific. 
On iOS I personally want to get rid of all the dropdowns and hide them behind a modal, because they look really bad on that platform.

apps/fluent-tester/src/FluentTester/FluentTester.android.tsx
```diff
@@ -0,0 +1,115 @@
...
+                      style={mobileStyles.testListItems}
+                      testID={description.testPage}
+                    >
+                      {description.name}
+                    </Text>
+                    <Separator style={mobileStyles.testListSeparator} />
+                  </View>
+                );
+              })}
+            </ScrollView>
```
Saadnajmi, line 98: I feels like this could be separated out into a separate component so that we don't have a block here. That also might help it be re-usable if it's split out

apps/fluent-tester/src/FluentTester/FluentTester.android.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const onBackPress = () => {
+    setOnMainScreen(true);
+    BackHandler.removeEventListener('hardwareBackPress', onBackPress);
+    return true;
+  };
+
+  return (
+    <DisplayIfVisible isVisible={true}>
+      <DisplayIfVisible isVisible={onMainScreen}>
```
Saadnajmi, line 73: Why do we need these?

apps/fluent-tester/src/FluentTester/FluentTester.android.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    </Text>
+                    <Separator style={mobileStyles.testListSeparator} />
+                  </View>
+                );
+              })}
+            </ScrollView>
+          </View>
+        </View>
+      </DisplayIfVisible>
+      <DisplayIfVisible isVisible={!onMainScreen}>
```
Saadnajmi, line 102: Oh I see.. you're making "pages" by having different parts render based on the boolean. 
"Main Screen" sounds like screen 1 if we had a multi display configuration (like an external monitor). Could we name this something like "onTestListView" or something more specific?

Or maybe just split it out to a separate component with comments to describe the pattern?

Saadnajmi: Maybe too complicated, but rather than have a completely android specific page, maybe we could add the "DisplayIfVisible" wrapper to FluentTester.tsx, and use a "useTwoPanes" flag that is set appropriately for each platform, then we can share the logic? And with this review, have it default to only use the new one pane solution for android, which is what you've tested.

That way, we could have useTwoPanes set to `true` on Desktop and Tablet (Or Surface Duo.. however that works) and we get the original tester, and `false` on iOS and android, and we get your new tester logic.

Saadnajmi (changes requested):

apps/fluent-tester/src/FluentTester/TestComponents/Common/styles.ts
tamasane: added to themedStyles and set the color to refer menuDivider

apps/fluent-tester/src/FluentTester/TestComponents/Test.tsx
tamasane: added check for iOS (excluded iPad)

apps/fluent-tester/src/FluentTester/TestComponents/Test.tsx
tamasane: updated to render on non-mobile platforms

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
tamasane: will pick up this work item after this  [PR#633](https://github.com/microsoft/fluentui-react-native/pull/633) is merged

apps/fluent-tester/src/FluentTester/FluentTester.android.tsx
Saadnajmi: Thanks! I like this approach =)

apps/fluent-tester/src/FluentTester/FluentTesterApp.tsx
```diff
@@ -2,14 +2,17 @@
 
 import { ThemeProvider } from '@fluentui-react-native/theme';
 import * as React from 'react';
+import { Platform } from 'react-native';
 import { FluentTester, FluentTesterProps } from './FluentTester';
 import { tests } from './testPages';
 import { testerTheme } from './theme/index';
 
+const isMobile = Platform.OS == 'android' || (Platform.OS == 'ios' && !Platform.isPad);
```
Saadnajmi, line 10: Remove iOS for now. I realized I don't have a way to handle "going back" because BackHandler is android specific. I'll figure that out in a followup and re-add the platforms to this check.

Saadnajmi (changes requested):
Just remove iOS/iPad from isMobile for now and this should be good to go!

Could you also add notes on your verification plan so that section doesn't remain empty when we commit this?

apps/fluent-tester/src/FluentTester/FluentTesterApp.tsx
tamasane: Yep removing iOS for now.

Saadnajmi (approved):

PR #613, opened 2021-02-24
Android theme

tamasane:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

- Added android-theme to support android platform. The token system follows semantic tokens as in Figma FluentUI  Android which are referred in ThemeReference API.

### Verification

The new theme looks as follows:

https://user-images.githubusercontent.com/23552252/110751711-40308d00-826a-11eb-81f7-bf50833869b8.mp4

( The appearance change is emulated in the demo by making local changes to react-native-test-app and currently dark/light mode does not switch automatically as the react-native-test-app does not provide subscription to uiMode changes )

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/android-theme/src/androidTypography.ts
```diff
@@ -0,0 +1,51 @@
...
+      bodyStandard: { face: 'primary', size: 'body', weight: '400' },
+      bodySemibold: { face: 'primary', size: 'body', weight: '600' },
+      subheaderStandard: { face: 'primary', size: 'subheader', weight: '400' },
+      subheaderSemibold: { face: 'primary', size: 'subheader', weight: '600' },
+      headerStandard: { face: 'primary', size: 'header', weight: '700' },
+      headerSemibold: { face: 'primary', size: 'header', weight: '800' },
+      heroStandard: { face: 'primary', size: 'hero', weight: '400' },
+      heroSemibold: { face: 'primary', size: 'hero', weight: '700' },
+      heroLargeStandard: { face: 'primary', size: 'heroLarge', weight: '400' },
+      heroLargeSemibold: { face: 'primary', size: 'heroLarge', weight: '700' },
```
Saadnajmi, line 46: These look apple specific. Specifically "SF Pro", and all the font weights. Should probably remove

tamasane: Changed Typography as per [Figma Fluent Android](https://www.figma.com/file/MkkE13z6zALstwLlbczIMXrA/Fluent-Android?node-id=7455%3A94).

packages/theming/android-theme/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@fluentui-react-native/android-theme",
+  "version": "0.0.1",
```
Saadnajmi, line 3: Nit: I think we start at 0.1.0 for most new packages

Saadnajmi:
Could you add screenshots of before and after with the theme? Would love to see how it looks with the new theme

packages/theming/android-theme/src/androidTheme.ts
```diff
@@ -0,0 +1,42 @@
...
+      borderWidth: 1,
+      minHeight: 48,
+      minWidth: 92,
+    },
+  },
+  RNFText: {
+    tokens: {
+      variant: 'bodyStandard',
+    },
+  },
```
Saadnajmi, line 31: This was also apple specific, but if it works, then great!

packages/theming/android-theme/package.json
tamasane: Noted. Thanks for letting me know 😄 👍🏻

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
```diff
@@ -4,10 +4,15 @@ import { ThemeReference } from '@fluentui-react-native/theme';
 import { Platform } from 'react-native';
 import { applyBrand, OfficeBrand } from './applyBrand';
 import { applyTheme, ThemeNames } from './applyTheme';
+import { createAndroidTheme } from '@fluentui-react-native/android-theme';
 
-const themeOptions: ThemeOptions = { paletteName: 'TaskPane', appearance: 'dynamic' };
+const themeOptions: ThemeOptions = Platform.select({
+  android: { appearance: 'dynamic' },
+  default: { paletteName: 'TaskPane', appearance: 'dynamic' },
+});
```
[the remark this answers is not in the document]
tamasane: Agree about this. Though this brings me to another question - why are the platform specific theming packages separated instead of adding platform specific files and entry point within the default-theme? Is there any specific reason? 👀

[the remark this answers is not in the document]
Saadnajmi: It came down to the ThemeReference API. The ThemeReference API that the wonderful people like @FalseLobster and @JasonVMo and their teams wrote takes in a "Base Theme", and list of "Theme Recipes" (Either partial themes or functions that return partial themes) to create one final merged theme. The ThemeReference API also lets you specify listeners and ways to invalidate the theme.

When I wrote the Apple Theme, I found that because of the way Native Modules work (this could change for TurboModules), I couldn't just make a ThemeRecipe to layer on the default-theme. I needed a full ThemeReference, complete with it's own baseTheme, partial theme (from the native module) and listeners (either macOS system events or React-Native events on iOS). The ThemeReference API does not let me merge two ThemeReferences together, only a ThemeReference with ThemeRecipes.

Because of that extra level of complexity, I decided to let the AppleTheme be it's own complete ThemeReference that is meant to be used _instead_ of the default-theme, rather than just a ThemeRecipe layered on top. I think this works much better, since the two themes don't have much in common anyway. The only downside is I need to find a way to tell clients "Please use the apple-theme" 😁.

An argument could be made that the default-theme should just return the apple-theme/android-theme on those platforms. I am open to making that change.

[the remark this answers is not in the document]
Saadnajmi: Oh yeah. I should also mention the win32-theme was implemented as a separate package, so I was copying that approach too.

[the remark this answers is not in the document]
tamasane: > An argument could be made that the default-theme should just return the apple-theme/android-theme on those platforms. I am open to making that change.

The availability of other platform themes doesn't make sense. Making this change for the android-theme.

packages/theming/android-theme/README.md
```diff
@@ -0,0 +1,5 @@
+# Android-theme
+
+Code and definitions for creating an Android Theme for FluentUI React Native.
+
+The theme follows definitions from [Figma - Fluent Android](https://www.figma.com/file/MkkE13z6zALstwLlbczIMXrA/Fluent-Android). The theme is work in progress and changes are expected.
```
Saadnajmi, line 5: I don't think this Figma is quite public facing, so it may be better leave it out of the README, and maybe link something like FluentUI Android instead?

packages/theming/android-theme/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@fluentui-react-native/theme-types": ">=0.5.0 <1.0.0",
+    "@fluentui-react-native/theme": ">=0.4.1 <1.0.0"
+  },
+  "devDependencies": {
+    "@uifabricshared/build-native": "^0.1.1",
+    "@uifabricshared/eslint-config-rules": "^0.1.1"
+  },
+  "peerDependencies": {
+    "react": "^16.13.1",
+    "react-native": "^0.62.2"
```
Saadnajmi, line 44: We recently updated the repo to use React Native 0.63 as the minimum, would you be able to update this here too?
https://github.com/microsoft/fluentui-react-native/pull/631

packages/theming/android-theme/src/androidTheme.colors.ts
```diff
@@ -0,0 +1,198 @@
...
+    disabledBackground: p.surfacesTertiary,
+    disabledText: p.textDisabledHighContrast,
+    disabledBodyText: p.textDisabled,
+    disabledSubtext: p.gray200, //TODO
+    disabledBodySubtext: p.gray300, //TODO
+
+    focusBorder: 'transparent', //TODO
+    variantBorder: 'transparent', //TODO
+    variantBorderHovered: 'transparent', //TODO
+    defaultStateBackground: 'transparent', //TODO
```
Saadnajmi, line 27: Are these TODO's still relevant? If not, please remove them before check in

packages/theming/android-theme/src/createAndroidTheme.ts
```diff
@@ -0,0 +1,29 @@
+import { ThemeReference } from '@fluentui-react-native/theme';
+import { getAndroidTheme } from './androidTheme';
+import { Appearance } from 'react-native';
+import { Theme } from '@fluentui-react-native/theme-types';
+
+export type AppearanceOptions = 'light' | 'dark';
```
Saadnajmi, line 6: Does this need to be exported? Or did you copy this from the default theme?

Saadnajmi (changes requested):

packages/theming/android-theme/.npmignore
```diff
@@ -0,0 +1,14 @@
+src
```
Saadnajmi, line 1: As pointed out in #https://github.com/microsoft/fluentui-react-native/pull/643 You'll want to remove the first line here too. =)

packages/theming/android-theme/src/createAndroidTheme.ts
tamasane: yep copied the approach of default theme, though these are not needed and removed.

packages/theming/android-theme/.npmignore
tamasane: done :)

packages/theming/android-theme/README.md
tamasane: added reference to FluentUI Android resource files on Github

packages/theming/android-theme/package.json
tamasane: updated to 0.63.4 =)

packages/theming/android-theme/src/androidTheme.colors.ts
```diff
@@ -0,0 +1,198 @@
...
+    listItemBackgroundCheckedHovered: 'transparent',
+
+    listHeaderBackgroundHovered: 'transparent',
+    listHeaderBackgroundPressed: 'transparent',
+
+    actionLink: 'todo',
+    actionLinkHovered: 'todo',
+    link: 'todo',
+    linkHovered: 'todo',
+    linkPressed: 'todo',
```
Saadnajmi, line 99: If you don't have every color defined just yet, you can specify this as a partial palette and leave this out. But keeping a string 'todo' likely will just cause the app to crash/error if it ever shows a link while using the android theme.

Saadnajmi: https://github.com/microsoft/fluentui-react-native/blob/03896e322970c7f2c052ddf1ea8be8dfac6e4e7e/packages/theming/theme-types/src/Color.types.ts#L454

packages/theming/android-theme/README.md
```diff
@@ -0,0 +1,5 @@
+# Android-theme
+
+Code and definitions for creating an Android Theme for FluentUI React Native.
+
+The theme follows color, typography, spacing and other values to closely match [FluentUI Android](https://github.com/microsoft/fluentui-android/tree/master/FluentUI/src/main/res/values). The theme is work in progress and changes are expected.
```
Saadnajmi, line 5: It's probably fine to just leave this link as the base GitHub repo (https://github.com/microsoft/fluentui-android) but that's a nitpick and I wouldn't consider it necessary for check in.

Saadnajmi (commented):
The remaining todo's are my only real comment left (Either track them with an issue and place that # in the code as a comment, or remove the todos altogether). Otherwise, looks great!

Saadnajmi:
Re: "( The appearance change is emulated in the demo by making local changes to react-native-test-app and currently dark/light mode does not switch automatically as the react-native-test-app does not provide subscription to uiMode changes )" Is that a bug or feature request we need to make to [react-native-test-app](https://github.com/microsoft/react-native-test-app) ?

Saadnajmi (approved):

PR #615, opened 2021-02-26
Add a Native Theming Module for macOS

### LINKED ISSUE #212: macOS Button incorrectly styled
According to [Fluent design website](https://developer.microsoft.com/en-us/fluentui#/controls/crossplatform/button), the macOS Button should look like this:

image
image

But currently, it looks like this:

image

Creating this issue to track this inconsistency. Please let me know if you have any questions.

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Let's add a native module for the apple theme on macOS.

1. Convert the current macOS theme (defined entirely in JS) to be the base / fallback theme.
2.  Add a Native Module that loads some values from the OS (namely typography).
3.  Create a new ThemeReference that takes in the original JS Apple Theme as the base theme, and has a callback that loads the native theme as a partial theme on top of the base theme.

In order to do this, I had to change the base theme of FluentTester on macOS from "default-theme" to "apple-theme". This is because as it was implemented before, the apple-theme was pulled out of the createAppleTheme() ThemeReference, and applied on top of the default-theme. Any fancy listeners or such we set up in createAppleTheme() wouldn't be transferred over, only the final theme.  As such, we ran into weird cases where the native module partial theme would not be applied, or would be applied the next time the theme was invalidated. A possible followup may be to allow ThemeReference recipes to include other ThemeReferences and have the listeners transfer over. The current solution I have seems fine though.

Some other quality of life improvements:
1. Update the Base Apple theme color mappings to be more accurate. Use communicationBlue for all the primaryBackgrounds, and communincationBlueShade20 for the pressed state (that seems to match Apple's pressed system effect the most).
2. Override some Button styling in the apple them to make the (JS) button component look closer to the NativeButton.  Apart from width, they are about as close as I could get them to be. 

This should close #212 

### Verification

Ran FluentTester and manually changed light / dark mode switches. 

https://user-images.githubusercontent.com/6722175/110047166-ffd58a00-7d12-11eb-997e-46a7437b044b.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/apple-theme/src/createAppleTheme.macos.ts
```diff
@@ -1,6 +1,42 @@
...
+          console.error(`Error retrieving apple theming module! ${error}`);
+        }
+        // Layer the native apple theming module values on top of the base Apple theme
+
+        // TODO this doesn't work. Why?
+        // appleThemeReference.update(applePartialTheme);
+        // Set the individual properties instead
+        appleThemeReference.theme.colors = applePartialTheme.colors;
+        appleThemeReference.theme.typography = applePartialTheme.typography;
+        appleThemeReference.invalidate();
```
Saadnajmi, line 25: Unless I'm obviously doing something wrong here, This feels like a bug. I'll update this comment with a bug number unless I or someone else finds out what's wrong beforehand

packages/theming/apple-theme/macos/AppleThemeModule.swift
```diff
@@ -0,0 +1,427 @@
...
+			/* PaletteBackgroundColors & PaletteTextColors */
+
+			"background" : RCTColorToHexString(NSColor.windowBackgroundColor.cgColor),
+			"bodyStandoutBackground" : RCTColorToHexString(NSColor.underPageBackgroundColor.cgColor),
+			"bodyFrameBackground" : RCTColorToHexString(NSColor.windowBackgroundColor.cgColor),
+			"bodyFrameDivider" : RCTColorToHexString(NSColor.separatorColor.cgColor),
+			"bodyText" : RCTColorToHexString(NSColor.textColor.cgColor),
+			"bodyTextChecked" : RCTColorToHexString(NSColor.selectedTextColor.cgColor),
+			"subText" : RCTColorToHexString(NSColor.placeholderTextColor.cgColor),
+			"bodyDivider" : RCTColorToHexString(NSColor.separatorColor.cgColor),
```
Saadnajmi, line 67: Re: Bug that fetching colors doesn't work properly.

Even though I am calling this again every-time the system appearance changes (say from light to dark), reevaluating this code will still give the original appearances. 

Example:
Boot in Light Mode
Background: #ececec
Switch to Dark Mode 

EXPECTED:
Background: #323232

ACTUAL:
Background: #ececec

I dunno what I'm doing wrong here, tagging other apple people @HeyImChris @harrieshin in case they catch something obvious

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -15,13 +13,33 @@ export function appleSpacing(): Spacing {
...
+    },
+    stack: {
+      style: {
+        paddingVertical: 4.5,
+        minHeight: 28,
+      },
+    },
+    icon: {
+      style: {
+        marginEnd: 10,
```
Saadnajmi, line 37: @harrieshin icon padding

packages/theming/apple-theme/macos/AppleThemeModule.swift
harrieshin: you are actually recreating the dictionary?

Saadnajmi: Yep. I can set breakpoints inside here and verify it's getting hit everyone I switch light / dark

Saadnajmi: Removed the code for now. I'll follow up to see if this is a valid apple bug

packages/experimental/NativeButton/src/NativeButton.tsx
```diff
@@ -14,9 +14,9 @@ export const NativeButton = compose<NativeButtonType>({
     root: buildProps(
       (tokens, theme) => ({
         style: {
-          height: 35,
+          height: 36,
```
HeyImChris, line 17: For things like this, what prompted this slight tweak?

packages/theming/apple-theme/FluentUIReactNativeAppleTheme.podspec
```diff
@@ -0,0 +1,25 @@
...
+  s.summary      = package['description']
+  s.license      = package['license']
+
+  s.authors      = package['author']
+  s.homepage     = package['homepage']
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
+
+  # s.ios.deployment_target = "11.0"
```
HeyImChris, line 17: Is the plan here that when we get iOS support we uncomment this?

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -15,13 +13,33 @@ export function appleSpacing(): Spacing {
 }
 
 export const appleComponents = {
+  // These values correspond to the "Large"  Button Size values of the FluentUI Apple Button
   Button: {
     tokens: {
-      borderRadius: 5,
+      borderRadius: 6,
```
HeyImChris, line 19: Did you just pull this from an actual NSButton you tested yourself? Wondering why it was wrong in the first place

packages/theming/apple-theme/src/createAppleTheme.macos.ts
```diff
@@ -1,6 +1,27 @@
...
+
+  const module = NativeModules && NativeModules.MSFAppleThemeModule;
+  if (module) {
+    module.getApplePartialThemeWithCallback((error, applePartialTheme) => {
+      if (error) {
+        console.error(`Error retrieving apple theming module! ${error}`);
+      }
+      // Layer the native apple theming module values on top of the base Apple theme
+
+      // TODO this doesn't work. Why?
```
HeyImChris, line 16: Every TODO should be paired with a bug number, also any relevant info you have here would help explain it better to the next engineer who might look at it would be nice. e.g. why do we need this commented out line in the first place?

HeyImChris (changes requested):

packages/theming/apple-theme/macos/AppleThemeModule.swift
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	/// Map the current FluentUI React Native font sizes approximately to their corresponding apple text style,
+	/// For older vesrions of macOS, fallback to the values described in the apple HIG:
+	/// https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/
+	/// These mappings and variants are subject to change as we moved to a unified cross platform Fluent typography ramp
+	static func fontSizes() -> [AnyHashable : Any] {
+		if #available(OSX 11.0, *) {
+			return [
+				"caption" : NSFont.preferredFont(forTextStyle: .caption1, options: [:]).pointSize,
```
harrieshin, line 45: are these weights all regular?

Saadnajmi: No idea what the weights returns since it's the system API. However, in the method "fontVariants()" you can see me basically define a bunch of text styles with weights
```
			"captionStandard": [
				"face" : families["primary"],
				"size" : sizes["caption"],
				"weight" : weights[.medium],
			],
			"secondaryStandard": [
				"face" : families["primary"],
				"size" : sizes["secondary"],
				"weight" : weights[.regular],
			],
			///
```

packages/theming/apple-theme/src/createAppleTheme.macos.ts
Saadnajmi: The intention here was to add it before checkin in case I either found the solution, or confirmed with someone on CXE that this was an actual bug.

However, I think I figured out why this isn't working, and can probably remove this. There's a bunch of FluentTester code I'll have to change to fix it though :/

packages/theming/apple-theme/src/appleTheme.macos.ts
Saadnajmi: I pulled these from the FluentUI Apple macOS Button, by looking at what we set it as in Swift. Those come from a Figma (I assume).
It was wrong because I wrote it wrong the first time.

packages/theming/apple-theme/FluentUIReactNativeAppleTheme.podspec
Saadnajmi: Yep. I'm also happy to just remove this, didn't notice this was here

packages/experimental/NativeButton/src/NativeButton.tsx
Saadnajmi: I wanted to align to a 4 px grid, but this is minor and can be removed and done separately. 
We probably shouldn't hardcode the height and width in the style but I'm not a flexbox expert enough to know what to actually do so that this view just uses the intrinsicContentSize

packages/theming/apple-theme/macos/AppleThemeModule.swift
Saadnajmi: Update: using KVO rather than listening to NSNotifications worked. I still listen to the ColorPreferencesChangedNotification to see when the controlAccentColor changes.

```
func startObserving() {
	kvoToken = NSApplication.shared.observe(\.effectiveAppearance) { (application, change) in
		guard self.bridge != nil else {
			return
		}
		// Do work
	}
}

func stopObserving() {
	kvoToken?.invalidate()

}
```

packages/theming/apple-theme/src/createAppleTheme.macos.ts
Saadnajmi: I figured out what the problem was. It wasn't enough to update the themeReference with a partial theme. The details are weird but tl;dr I had to change the base/default theme for FluentTester on macOS to be the apple-theme. That's probably better anyway.

Saadnajmi: Details for future reference:

A ThemeReference does not take other ThemeReferences as a way to layer themes on top of it, only partial themes or functions that return partial themes. This meant that even though I set up a callback and listener to load the partial theme from the native module on top of the base Apple theme in the createAppleTheme() ThemeReference, that didn't propagate up to the actual TesterThemeReference (what the Apple Theme was being applied on top of) getting invalidated and recalculated. That led to weird cases where the native theme wouldn't show up at all, or would only show up when the tester theme was invalidated for some other reason. 

The fix is to make the base theme of TesterThemeReference to the apple theme, and let the partial theme from the native module be a partial theme applied onto it.

Saadnajmi:
@heyimchris @harrieshin It looks like the macOS CI is failing because it can't find macOS 11 API's, even though I wrapped them in if #available checks. That seems weird? The CI machine was running 10.15 last I checked

Saadnajmi:
> @HeyImChris @harrieshin It looks like the macOS CI is failing because it can't find macOS 11 API's, even though I wrapped them in if #available checks. That seems weird? The CI machine was running 10.15 last I checked

Looks like updating xcode did the trick :D

harrieshin (approved):

packages/theming/apple-theme/FluentUIReactNativeAppleTheme.podspec
HeyImChris: Probably worth removing for now or adding a TODO with a bug number tracking adding iOS support

HeyImChris (approved):

Issue #626, opened 2021-03-03
ThemePickers on iOS aren't dark mode compliant and also look horrible


Saadnajmi:
[Microsoft Internal ADO Bug](https://office.visualstudio.com/APEX/_backlogs/backlog/FX/Features/?workitem=4920042)

### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [x] iOS
    - **Xcode version**:
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [ ] windows
  - [ ] android

### Please provide a reproduction of the bug:

I'll leave this bug focused on the dark mode but, and not the "Don't use Pickers at all / use the community module for picker instead" bit"

Unlike every other platform, the Picker component is not a dropdown menu but an always expanded carousel. Furthermore they aren't dark mode compliant, as seen in this screenshot. Notice you can't read the text. 

![Screen Shot 2021-03-01 at 10 54 47 PM](https://user-images.githubusercontent.com/6722175/109862546-2ec20200-7c26-11eb-91ca-dde58db41393.png)

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: Normal
Products/applications affected: (if applicable)

Saadnajmi:
Closed

Issue #625, opened 2021-03-03
Header Separator in FluentTester is always black (bad for Dark Themes)


Saadnajmi:
[Microsoft Internal ADO Bug](https://office.visualstudio.com/APEX/_backlogs/backlog/FX/Features/?workitem=4919892)

### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [x] iOS
    - **Xcode version**:
  - [x] macOS
    - **Xcode version**:
  - [x] win32
  - [x] windows
  - [x] android

### Please provide a reproduction of the bug:

https://user-images.githubusercontent.com/6722175/109864059-fcb19f80-7c27-11eb-9685-8f121ae48a1d.mov

See the attached video, and notice the Separator between the Header (under the FluentUI Tests title) is always black, even in dark mode. On iOS where we want a pure black background, this makes it invisible. Ideally there should be a different color for dark mode.

My proposed fix is to instead of using a standard Separator, use a themed Separator like what we do with the TestListSeparator. Something like:
```
  const HeaderSeparator = Separator.customize((t) => ({
    color: t.colors.bodyFrameDivider,
    separatorWidth: 2,
  }));
  ```
  
  the correct token to use can be up for debate. 
  
  I can make the fix iOS only, but I think this would be appreciated for all platforms

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes
Requested priority: Normal

Products/applications affected: (if applicable)

Issue #624, opened 2021-03-03
TestList cuts off test on iPad (possibly other platforms)


Saadnajmi:
[Microsoft Internal ADO Bug](https://office.visualstudio.com/APEX/_backlogs/backlog/FX/Features/?workitem=4919898)

### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [x] iOS
    - **Xcode version**:
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [ ] windows
  - [ ] android

### Please provide a reproduction of the bug:

Ignoring the fact that FluentTester is very much not mobile friendly (PR #613 should help fix that), the test list is cut off on iPad when in portrait mode. This is partially because the TestList style is a hardcoded width, rather than using flexbox + a minimum width.

I can make a fix on iOS only, but I think this is one that could be applied to all platforms. I'll need to set up a windows VM to test the UWP and win32 platforms though.

![Screen Shot 2021-02-26 at 4 28 37 PM](https://user-images.githubusercontent.com/6722175/109863727-9af13580-7c27-11eb-828e-7970e10f9767.png)

#### Actual behavior:

Test List cuts off text

#### Expected behavior:

Text should either expand, wrap around to a new line, or have trailing ellipses

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: Normal

Products/applications affected: (if applicable)

Issue #623, opened 2021-03-03
FluentTester doesn't expand to fill screen on mobile


Saadnajmi:
[Microsoft Internal ADO Bug](https://office.visualstudio.com/APEX/_backlogs/backlog/FX/Features/?workitem=4919889)

### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using) 
- **OS version**: (fill this out if relevant)
- **Platform**:
  - [x] iOS
    - **Xcode version**: 12.4
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [ ] windows
  - [x] android

### Please provide a reproduction of the bug:

![Screen Shot 2021-03-01 at 10 54 47 PM](https://user-images.githubusercontent.com/6722175/109862546-2ec20200-7c26-11eb-91ca-dde58db41393.png)

FluentTester on iOS (and probably Android) doesn't grow to fill the whole screen. This is because <FluentTesterApp> is wrapped in a SafeAreaView on both platforms that doesn't specify `flex: 1` in their style. Adding that let's the SafeAreaView grow, and thus the child FluentTester view grow. 

We can either add the style specifically on iOS and Android, or if SafeAreaView is a no-op on desktop, just always wrap <FluentTester> in a <SafeAreaView>.

### Priorities and help requested:

Are you willing to submit a PR to fix? **Yes**

Requested priority: **Normal**

Products/applications affected: (if applicable)

Saadnajmi:
@tamasane is this an issue on Android too?

tamasane:
> @tamasane is this an issue on Android too?

For Android, the following styles in FluentTester.tsx resolved this issue:

```
const getThemedStyles = themedStyleSheet((t: Theme) => {
  return {
    root: {
      backgroundColor: t.colors.background,
      height: '100%',
      flexDirection: 'column',
      justifyContent: 'flex-start',
      alignItems: 'stretch',
      padding: 4,
    },
  };
});
```

Saadnajmi:
@tamasane Thanks! I've got a fix with similar styling I'll send out soon. I just want to test on windows first. Works on every other platform though :)

PR #627, opened 2021-03-03
Add a base iOS Theme (JS Only)

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Like what we did with macOS, let's add an iOS theme that pulls some constants from FluentUI Apple.
Before we can add a Native Module to pull stuff from the OS/FluentUI Apple, we need a base JS theme to fallback to while the native module theme is fetcehd. That is what this PR addresses. We aren't on RN 0.63 yet, so i can't use PlatformColor. Luckily, we have enough semantic colors in FluentUI Apple that I got away with not needing it.

Note: FluentTester looks bad on iOS. I opened #623, #624, #625, and #626 to address specific bugs and will send PR's for them separately.

### Verification

Ran the iOS test app, and switched to the apple theme. I locally fixed some stuff like the view not growing also just to see how it WOULD look after all the other issues are fixed.

Screen Shot 2021-03-03 at 12 33 57 AM |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/ios/src/index.tsx
```diff
@@ -2,11 +2,18 @@
...
 import {FluentTesterApp} from '@fluentui-react-native/tester';
 import * as React from 'react';
-import {AppRegistry, SafeAreaView} from 'react-native';
+import {AppRegistry, SafeAreaView, Text, useColorScheme} from 'react-native';
 
 const FluentTester: React.FunctionComponent = () => {
+  const isDarkMode = useColorScheme && useColorScheme() === 'dark';
+
+  const backgroundStyle = {
+    backgroundColor: isDarkMode ? 'black' : 'white',
```
Saadnajmi, line 11: Note: https://github.com/microsoft/react-native-test-app/pull/286 will remove the need for this once it's checked in and published

apps/ios/src/index.tsx
```diff
@@ -2,11 +2,18 @@
...
+    flexGrow: 1,
```
Saadnajmi, line 12: This addresses #623 , I need to see if we should include this fix in Android as well. Tagging @tamasane

apps/ios/src/Podfile.lock
```diff
@@ -9,15 +9,15 @@ PODS:
...
     - ReactCommon/turbomodule/core (= 0.62.2)
-  - FluentUI-React-Native-Avatar (0.4.1):
+  - FluentUI-React-Native-Avatar (0.4.3):
     - MicrosoftFluentUI/AvatarView_mac (~> 0.1.28)
     - MicrosoftFluentUI/Controls_ios (~> 0.1.28)
     - React
-  - FluentUI-React-Native-Button (0.2.1):
+  - FluentUI-React-Native-Button (0.3.1):
     - MicrosoftFluentUI/Button_mac (~> 0.1.28)
     - MicrosoftFluentUI/Controls_ios (~> 0.1.28)
```
harrieshin, line 18: fyi next fluent publish, this need to change to button_ios and avatar_ios!

apps/ios/src/index.tsx
harrieshin: is it published yet?

apps/ios/src/Podfile.lock
Saadnajmi: Yeah, I should probably log a bug to update all the FluentUI Apple dependencies

packages/theming/apple-theme/src/fluentAppleColors.ios.ts
```diff
@@ -0,0 +1,321 @@
...
+  iconDisabled: ColorValue; //= UIColor(light: gray300, lightHighContrast: gray500, dark: gray600, darkHighContrast: gray400)
+  /** icon appears on top of surfaces that uses `Colors.primary` as background color */
+  iconOnAccent: ColorValue; //= UIColor(light: .white, dark: .black)
+
+  /**
+   * In Dark mode, our system use two sets of background colors -- called base and elevated -- to enhance the perception of depth when one dark interface is layered above another.
+   * The dark base colors are darker, making background interface appear to recede, and the elevate colors are lighter, making foreground interfaces appear to advance
+   */
+
+  surfacePrimary: ColorValue; //= UIColor(light: .white, dark: .black, darkElevated: gray950)
```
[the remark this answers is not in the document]
Saadnajmi: Reference from the original code, which also had the darkElevated. I thought it would be useful to keep them because you can see where I got the hex values from

apps/ios/src/index.tsx
Saadnajmi: Yes, I'll just include the new package with this PR

Saadnajmi: Eh... Updating RNTA from 0.3.2 to 0.4.3 works on iOS and macOS but breaks android. With my other changes with FluentTester layout, we don't need to update yet so I'll leave it off this review

apps/ios/src/index.tsx
Saadnajmi: Nevermind. FIxed it separately in #630

packages/theming/apple-theme/src/appleTypography.ios.ts
```diff
@@ -0,0 +1,56 @@
...
+      heroSemibold: { face: 'primary', size: 'hero', weight: '600' },
+      heroLargeStandard: { face: 'primary', size: 'heroLarge', weight: '400' },
+      heroLargeSemibold: { face: 'primary', size: 'heroLarge', weight: '600' },
+    } as Variants,
+  };
+
+  return appleDict;
+}
+
+export function appleSpacing(): Spacing {
```
harrieshin, line 54: why is spacing define in applyTypography.ios.ts?

Saadnajmi: Good catch. I'll move it

harrieshin (approved):

PR #630, opened 2021-03-04
Fix up some FluentTester layout bugs

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android
- [x] web

### Description of changes

There's a bunch of issues with FluentTester that this PR aims to address, namely #623, #624, and #625

This work came up as I was working on the macOS and iOS theme, because the layout issues made some stuff just unreadable/unusable. This was finally a good opportunity to dive in and fix some stuff up, validating for all platforms.

To summarize the changes:

- Changed the Base View of FluentTester on iOS to SafeAreaView (keeping it view on all other platforms. Its a no-op on other platforms anyway, but better safe than sorry), and removed the wrapping SafeAreaView from the iOS and Android apps. All platforms have more or less the same `index.tsx` now. This also fixes some flexbox issues we had where we forgot to set "flex:1" on android on iOS causing weird layout problems.
- Removed the minimum height and width from the rootView style. Those minimums may have been too wide for portrait or too tall for landscape on mobile. Moreover, we shouldn't need to specify minimums and should let flexbox do the right thing
- Stylized the Separators so that they adapt to Dark Mode better. I'm open to choosing a better token / color though, I kinda just guessed here.
- Changed the testList style to not have a hardcoded width but instead have flexgrow and flexshrink of both 0. This basically means the test list always stays the same size, what I would expect of a sidebar to do. For whatever reason, win32 also needed a minWidth set or else it wouldn't show any of the test items, so I added that as well.

See this video of the test list size never changing:

https://user-images.githubusercontent.com/6722175/109934239-e8a28800-7c91-11eb-8b96-1ebc38a86f43.mov

### Verification

I ran the tester on a few platforms, namely iOS, macOS, Android, and Web
I still need to test Windows, and Win32. Those require me to setup a Windows VM on my Macbook so that might take some time. 😕

Screenshots on various platforms. The iPad screenshot is dark mode to show that the FluentTester view now extends to fit the whole screen

#### iOS (iPad)

Screen Shot 2021-03-04 at 2 21 37 AM

#### macOS

Screen Shot 2021-03-04 at 2 23 22 AM

#### Android

Screen Shot 2021-03-04 at 2 25 13 AM

#### Win32

win32

#### Windows

windows

#### Web

Screen Shot 2021-03-04 at 2 19 34 AM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester/FluentTester.tsx
@@ -3,7 +3,7 @@ import { StealthButton, Separator } from '@fluentui/react-native';
 import { Text } from '@fluentui-react-native/experimental-text';
 import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
 import * as React from 'react';
-import { ScrollView, View, Text as RNText } from 'react-native';
+import { ScrollView, View, Text as RNText, Platform, SafeAreaView } from 'react-native';
 import { TestDescription } from './TestComponents';
 import { BASE_TESTPAGE } from './TestComponents/Common/consts';
 import { fluentTesterStyles } from './TestComponents/Common/styles';
@@ -51,8 +51,6 @@ const getThemedStyles = themedStyleSheet((t: Theme) => {
       flex: 1,
       flexGrow: 1,
       flexDirection: 'column',
-      minHeight: 550,
-      minWidth: 300,
       justifyContent: 'flex-start',
       alignItems: 'stretch',
       padding: 4,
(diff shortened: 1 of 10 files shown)


apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -73,18 +71,28 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
 
   const TestComponent = selectedTestIndex == -1 ? EmptyComponent : sortedTestComponents[selectedTestIndex].component;
 
+  const HeaderSeparator = Separator.customize((t) => ({
+    color: t.colors.bodyFrameDivider,
+    separatorWidth: 2,
```
harrieshin, line 76: width 2? thats suprising.

harrieshin (approved):

apps/fluent-tester/src/FluentTester/FluentTester.tsx
harrieshin: because natively, we do "half pixel"

Saadnajmi: Native on which Platform? This was originally 1, but I can change it back

HeyImChris (approved):

PR #631, opened 2021-03-04
Update to react-native 0.63

acoates-ms:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Move react-native dependencies to react-native 0.63

### Verification

Ran through the test pages in win32 testapp.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


acoates-ms:
I've marked the version update as a minor version update.  I wonder if it should be a major update though?  -- Thoughts?

packages/components/Button/package.json
```diff
@@ -28,19 +28,19 @@
...
     "@uifabricshared/build-native": "^0.1.1",
     "@uifabricshared/eslint-config-rules": "^0.1.1",
-    "react-native": "0.62.2"
+    "react-native": "^0.63.4"
   },
   "peerDependencies": {
-    "react": "^16.8.0",
-    "react-native": ">=0.60.0"
+    "react": ">=16.13.1",
+    "react-native": ">=0.63.4"
```
Saadnajmi, line 43: Do we need to specify it as both a devDependency and peerDependency? My worry is that having it a devDependency increases the likelihood the monorepo can pull in two versions of react native that don't play nice.

My last thought on this was to have every package (Excpt the test apps) specify them as peer dependencies, then use [yarn resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) to specify one version of react-native/react-native-macos/windows/win32 for the whole monorepo.

packages/components/Button/src/Button.types.ts
```diff
@@ -5,7 +5,7 @@ import { ITextProps } from '@fluentui-react-native/text';
 import { IPressableProps } from '@fluentui-react-native/pressable';
 import { FontTokens, IForegroundColorTokens, IBackgroundColorTokens, IBorderTokens } from '@fluentui-react-native/tokens';
 import { IFocusable, IPressableState } from '@fluentui-react-native/interactive-hooks';
-import { IViewWin32Props } from '@office-iss/react-native-win32';
+import type { IViewProps } from '@fluentui-react-native/adapters';
```
Saadnajmi, line 8: Thank you for this :D No reason the JS components should pull react-native-win32 on all platforms

packages/components/Callout/package.json
```diff
@@ -29,14 +29,14 @@
...
     "@uifabricshared/build-native": "^0.1.1",
     "@uifabricshared/eslint-config-rules": "^0.1.1",
-    "react-native": "0.62.2"
+    "react-native": "^0.63.4"
   },
   "peerDependencies": {
-    "react": "^16.8.0",
-    "react-native": ">=0.60.0"
+    "react": ">=16.13.1",
+    "react-native": ">=0.63.4"
```
Saadnajmi, line 39: General question for Microsoft packages/components, do we prefer to keep the minimum 0.62 since that technically would still work, or to keep the minimum the highest version of RN as possible?

packages/components/Checkbox/src/Checkbox.types.ts
```diff
@@ -5,7 +5,7 @@ import { ViewProps } from 'react-native';
 import { IRenderData } from '@uifabricshared/foundation-composable';
 import { ITextProps } from '@fluentui-react-native/text';
 import { FontTokens, IForegroundColorTokens, IBackgroundColorTokens, IBorderTokens } from '@fluentui-react-native/tokens';
-import { IViewWin32Props } from '@office-iss/react-native-win32';
+import type { IViewWin32Props } from '@office-iss/react-native-win32';
```
Saadnajmi, line 8: What does adding type do here? But also, could we just use adapters here?

yarn.lock
```diff
@@ -14093,47 +14013,10 @@ react-native-windows@^0.62.1:
...
-    whatwg-fetch "^3.0.0"
-
-react-native@0.63.3:
-  version "0.63.3"
-  resolved "https://registry.yarnpkg.com/react-native/-/react-native-0.63.3.tgz#4a7f6540e049ff41810887bbd1125abc4672f3bf"
-  integrity sha512-71wq13uNo5W8QVQnFlnzZ3AD+XgUBYGhpsxysQFW/hJ8GAt/J5o+Bvhy81FXichp6IBDJDh/JgfHH2gNji8dFA==
+react-native@^0.63.4:
+  version "0.63.4"
+  resolved "https://registry.yarnpkg.com/react-native/-/react-native-0.63.4.tgz#2210fdd404c94a5fa6b423c6de86f8e48810ec36"
+  integrity sha512-I4kM8kYO2mWEYUFITMcpRulcy4/jd+j9T6PbIzR0FuMcz/xwd+JwHoLPa1HmCesvR1RDOw9o4D+OFLwuXXfmGw==
```
Saadnajmi, line 14019: Yay it's only one version in the lock file :D

packages/components/Checkbox/src/Checkbox.types.ts
acoates-ms: It specifies that you are only importing type information, no actual JS.  In this case it shouldn't make any difference since IViewWin32Props is only a type.  But it makes it more explicit that its just a type information import.

packages/components/Callout/package.json
acoates-ms: 0.62 would probably at least mostly work.  But for instance we are now setting the Focusable property (0.63 supported on all platforms), rather than acceptsKeyboardFocus (0.62 only on windows/macos and our office RN fork).   We could try to support 0.62 also, but then we'd need some weird casts to allow us to set both properties, and ensure that we do set both when someone specifies one of them.  And generally be quick a headache, that would be undertested.

As far as I know all our clients are moving to 0.63, so we should just reduce our testing / support matrix while we can.

Saadnajmi: The point about the focusable property makes sense! That does set our minimum to 0.63 for just about everything, thanks!

apps/fluent-tester/package.json
```diff
@@ -39,15 +39,15 @@
...
     "@fluentui/react-native": ">=0.21.13 <1.0.0",
-    "@office-iss/react-native-win32": "^0.62.6",
     "react-native-svg": "^11.0.0"
   },
   "devDependencies": {
+    "@office-iss/react-native-win32": "^0.63.7",
     "@react-native-community/eslint-config": "^1.1.0",
     "@rnx-kit/cli": "^0.0.3",
     "@types/jasmine": "3.5.10",
     "@types/react": "^16.9.34",
```
NickGerleman, line 49: Do we have new react typings?

apps/web/package.json
```diff
@@ -24,7 +24,7 @@
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
```
NickGerleman, line 26: Need new react-dom typings?

NickGerleman (approved):

NickGerleman:
> I've marked the version update as a minor version update. I wonder if it should be a major update though? -- Thoughts?

Bumping a peer dependency is technically a breaking change.

acoates-ms:
> > I've marked the version update as a minor version update. I wonder if it should be a major update though? -- Thoughts?
> 
> Bumping a peer dependency is technically a breaking change.

yeah, talked to Alicia.  I forgot FURN is still on 0.X versions, so a minor bump is sufficient.

apps/web/package.json
acoates-ms: react-dom typings went from 16.9.X to 17.X.  So I dont think so.

apps/fluent-tester/package.json
acoates-ms: react typings went from 16.9.X to 17.X.  So I dont think so.

PR #633, opened 2021-03-05
Revamp ThemePicker in FluentTester Header

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

There were two main problems with the ThemePicker in our test app:

1- They looked horrible on mobile, namely iOS
2- We were using the react-native core picker, rather than the community module. As such, we constantly got a warning to use the community module picker instead.

For the first point, on iOS, I replaced the use of Pickers with the community module [react-native-menu](https://github.com/react-native-menu/menu). I also adopted the mobile Picker that we added for Android, refactoring it a bunch to make it cleaner. Unlike Android, I had to add a back button in the header when we're on a test component view to go back to the test list (android uses the hardware back button).

Fir the second point, we can't simply use the community module because it doesn't support win32 (nor should it). So we need platform specific files...
- `ThemePicker.macos.tsx` is now uses the community module, but otherwise is the same.
- `ThemePicker.ios.tsx` uses the community module, and replaces the Pickers with Contextual Menus (ActionSheets on iOS 13).

I would have done the other platforms (android, UWP), but I don't know how to  do that, sooo #639 now exists.

This also addresses #626 as the iOS Picker inherits a theme so that the text colors correctly in Dark Mode

### Verification

Tested the Picker on iOS (phone and iPad), macOS, android, windows, and UWP.

https://user-images.githubusercontent.com/6722175/110668401-e314d880-8190-11eb-8f2c-1b6b99da8f15.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Dropping these two links here because using a context menu on iOS seems better..
https://developer.apple.com/videos/play/wwdc2020/10205/
https://github.com/react-native-menu/menu

Saadnajmi:
@tamasane just to clarify, would you prefer I remove android from this change? I couldn't add you as a reviewer for some reason so tagging you here

tamasane:
> @tamasane just to clarify, would you prefer I remove android from this change? I couldn't add you as a reviewer for some reason so tagging you here

Yes removing it here would be good. I will follow up on it as we progress on android-theme.

apps/ios/src/Podfile
```diff
@@ -28,6 +28,8 @@ use_test_app! do |target|
     pod 'FluentUI-React-Native-Avatar', :path => '../../../packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec'
     pod 'FluentUI-React-Native-Shimmer', :path => '../../../packages/components/Shimmer/FluentUIReactNativeShimmer.podspec'
     pod 'FluentUI-React-Native-Button', :path => '../../../packages/experimental/NativeButton/FluentUIReactNativeButton.podspec'
+    pod 'react-native-menu', :path => '../../../node_modules/@react-native-menu/menu'
+    pod 'RNCPicker', :path => '../../../node_modules/@react-native-picker/picker'
```
harrieshin, line 32: lets either be generous and fix the personcoin or we don't feel lucky and remove this dependency

harrieshin (approved):

PR #637, opened 2021-03-08
Add iOS theme native module

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Like what we did for macOS, let's add a native module to the iOS theme that pulls the Theme/Constants directly from FluentUI Apple, rather than JS. There are a few differences from the macOS theme:
- Using KVO to observe appearance changes in our native module didn't work like it did in macOS. Instead, I add listeners on the JS side using the React Native Appearance API (similar to what the default theme does). We may want to add more listeners for accessibility events like BoldText / dynamic fonts / etc (https://reactnative.dev/docs/accessibilityinfo). 
- The macOS theme has a base apple theme, and ap partial theme loaded  from a native module as a recipe layered on top. The iOS theme has an empty object as a base,  a recipe for the base theme, and a recipe for the native module theme. I had to do it this way, because the iOS theme didn't play nicely when I used `DynamicColorIOS` & `PlatformColor` for it's palette. This is because those types resolve to `string | symbol` rather than `string`, which is incompatible with some of our theming infrastructure. To solve this, I do what the default theme does and define two (light and dark) themes that we switch between. 
    - This brings up the question: Why could I use `PlatformColor / DynamicColorMacOS` for the macOS theme? Probably some weirdness with the fact that iOS uses react-native, and macOS uses react-native-macos. I think there's definitely a followup to make sure theming plays nicely, or to just add support for `PlatformColor/DynamicColorIOS` to our theming infrastructure explicitly. 

### Verification

Verified that the native theme was in fact overlaid on top of the base JS theme by changing one value (heroLargeSemibold font size) to be oviously wrong in the JS, and seeing if it would get overwritten by the native theme. I also switched between light and dark and made sure the right callbacks were firing to load the appropriate themes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/apple-theme/ios/AppleThemeModule.swift
+++ b/packages/theming/apple-theme/ios/AppleThemeModule.swift
@@ -2,7 +2,7 @@ import Foundation
 import FluentUI
 
 @objc(MSFAppleThemeModule)
-class AppleThemeModule: RCTEventEmitter {
+class AppleThemeModule: NSObject {
 
 	public enum Events: CaseIterable {
 		case traitCollectionDidChange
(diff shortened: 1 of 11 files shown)


packages/theming/apple-theme/src/createAppleTheme.ios.ts
```diff
@@ -1,17 +1,38 @@
...
     return current === 'light' ? BaseAppleLightThemeIOS : BaseAppleDarkThemeIOS;
-  });
+  };
+
+  const loadNativeTheme = (appleThemeReference: ThemeReference) => {
+    const module = NativeModules && NativeModules.MSFAppleThemeModule;
+    if (module) {
+      module.getApplePartialThemeWithCallback((error, applePartialTheme) => {
+        if (error) {
+          console.error(`Error retrieving apple theming module! ${error}`);
```
HeyImChris, line 20: Is there a reason you're using backquote instead of a regular single quote?

packages/theming/apple-theme/ios/AppleThemeModule.swift
```diff
@@ -0,0 +1,379 @@
...
+			"serif": systemFontFamilyName
+		]
+
+	}
+
+	/// Map the current FluentUI React Native font sizes approximately to their corresponding FluentUI Apple  text style,
+	/// These mappings and variants are subject to change as we moved to a unified cross platform Fluent typography ramp
+	static func fontSizes() -> [AnyHashable : Any] {
+			return [
+				"caption" : FluentUI.Fonts.caption2.pointSize,
```
HeyImChris, line 281: Caption uses `caption2` and secondary uses `caption1`? Sounds backwards to me but I also am not looking at either at the moment

HeyImChris (approved):

packages/theming/apple-theme/src/createAppleTheme.ios.ts
Saadnajmi: that might just have been prettier when I decided to add the `${error}`. I don't like backticks and I'll see if I can just switch to single quotes and string concatenation. Good catch!

packages/theming/apple-theme/ios/AppleThemeModule.swift
Saadnajmi: Yeah, my mappings are.. questionable. I think neither FluentUI Apple nor FluentUI React Native use a unified Fluent Typography Type Ramp (I can't find the Figma right now). We don't have a "secondary" text style in FluentUI Apple, so I just chose caption 1 for that, and caption 2 for the actual "caption" text style.

PR #640, opened 2021-03-11
Fix publish

afoxman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update nuget spec with corrected names for ios and macos bundle files.

### Verification

Verified by adding nuget packing to PR loop, watching it succeed, then removing the PR loop changes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.ado/azure-pipelines.yml
+++ b/.ado/azure-pipelines.yml
@@ -75,8 +75,15 @@ jobs:
     pool:
       vmImage: 'macos-10.15'
       demands: ['xcode', 'sh', 'npm']
-    timeoutInMinutes: 90 # how long to run the job before automatically cancelling
+    timeoutInMinutes: 60 # how long to run the job before automatically cancelling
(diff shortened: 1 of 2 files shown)


afoxman:
Verified that this successfully produces a nuget package.

The built bundle file names are now different, which is what caused this break. I was careful to rename the bundles -- in the nuget package -- back to original names, so that any consuming code won't break. We should verify that the consuming app continues to work.

Saadnajmi (approved):

Saadnajmi:
> Verified that this successfully produces a nuget package.
> 
> The built bundle file names are now different, which is what caused this break. I was careful to rename the bundles -- in the nuget package -- back to original names, so that any consuming code won't break. We should verify that the consuming app continues to work.

At the moment nothing consumes the bundles (just the static libraries) but keeping the same name should help make sure the tooling to import it downstream won't break anyway :). Thank you!

PR #645, opened 2021-03-13
Add Icon support for macOS

amgleitman:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This adds Icon support for macOS using the recently published @microsoft/react-native-svg-desktop fork.

### Verification

Validated via the test page. Things work on macOS, and there are no changes to iOS.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/package.json
```diff
@@ -39,9 +39,9 @@
     "@fluentui-react-native/themed-stylesheet": "^1.2.0",
     "@fluentui-react-native/win32-theme": ">=0.6.0 <1.0.0",
     "@fluentui/react-native": ">=0.23.2 <1.0.0",
+    "@microsoft/react-native-svg-desktop": "^12.1.1",
     "@react-native-menu/menu": "^0.1.2",
-    "@react-native-picker/picker": "^1.9.11",
-    "react-native-svg": "^11.0.0"
```
Saadnajmi, line 44: Does this technically effect win32 too? Just curious

Saadnajmi (approved):

apps/fluent-tester/package.json
amgleitman: It probably would, so I think it'd probably be a good idea to do some extra validation on Windows. That being said, the fact that the CI fails as is worries me a bit, so probably a good thing to double check.

Saadnajmi: I personally cloned the repo on my HP Z440 and followed the instructions on `apps/win32` to run the test app. Your windows machine also doesn't need to be enlisted, it can be a personal one and the test app will still run. If you don't have a windows machine on hand easily, I can try helping with that. Done that for others before

apps/macos/src/Podfile
```diff
@@ -28,6 +28,7 @@ use_test_app! do |target|
     pod 'FluentUI-React-Native-Button', :path => '../../../packages/experimental/NativeButton/FluentUIReactNativeButton.podspec'
     pod 'FluentUI-React-Native-Avatar', :path => '../../../packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec'
     pod 'RNCPicker', :path => '../../../node_modules/@react-native-picker/picker'
+    pod 'RNSVG', :path => '../../../node_modules/@microsoft/react-native-svg-desktop'
```
chiuam, line 31: Do we need to update the iOS podfile as well? I think iOS should use the same library since FluentTester pulls in 'react-native-svg-desktop' not 'react-native-svg'.

samuelfreiberg:
I just checked out your branch and tried to run the Win32 test app locally. It didn't boot correctly, a React LogBox Error came up (empty, not sure why). But this is why the CI Win32 E2E tests are failing - the Win32 app is not booting at all.

Issue #657, opened 2021-04-08
FluentTester can't render Font Family Arial Black on iOS


Saadnajmi:
### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [x] iOS
    - **Xcode version**: 12.4
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [ ] windows
  - [ ] android

### Please provide a reproduction of the bug:

Open FluentTester on iOS
Navigate to "Text Test"

#### Actual behavior:

You get an error saying "Unrecognized font family Arial Black"
You can dismiss the error and see the rest of the page

![Screen Shot 2021-04-08 at 1 35 07 PM](https://user-images.githubusercontent.com/6722175/114079414-ca6f0f80-986f-11eb-8aad-1a04e8d92a64.png)

#### Expected behavior:

No error, see the test page

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: Low

Products/applications affected: (if applicable)

PR #663, opened 2021-04-20
Android Text Theme

tamasane:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

- The android-theme did not contain correct typography. The customisation of text on android is limited as very few font families are supported and there is bug in setting the font-weight values. This change addresses these limitations and applies workarounds to make typography compliant to Fluent Design guidelines.
- This change also updates the CustomUsage in Fluent-tester-app which shows example use cases for other supported font-families on android.

### Verification

| before-android-text-part1 | after-android-text-part01 |
| before-android-text-part2 | after-android-text-part02 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { View } from 'react-native';
+import { Platform, View } from 'react-native';
 import { Stack } from '@fluentui-react-native/stack';
 import { Text } from '@fluentui/react-native';
 import { stackStyle } from '../Common/styles';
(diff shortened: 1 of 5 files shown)


Saadnajmi:
I added @ksiler to this review because she's the last CXE person I remember working on Text. Also tagged @chiuam because I think she'll be working on a similar bug for iOS (#657)

packages/theming/android-theme/src/androidTypography.ts
```diff
@@ -17,33 +17,35 @@ export function androidTypography(): Typography {
...
-      primary: 'Roboto',
+      primary: 'sans-serif', // sans-serif, weight=400
+      primaryBold: 'sans-serif-medium', // sans-serif, weight=500
+      primaryLight: 'sans-serif-light', // sans-serif, weight=300
       secondary: 'System',
       cursive: 'System',
       monospace: 'monospace',
       sansSerif: 'sans-serif',
       serif: 'serif',
     },
```
Saadnajmi, line 34: Is it OK for themes to define their own font families? I would assume so but I would worry that would get confusing for clients writing x-plat code unless we explicitly documented it somewhere, like the in the README.md of the Android Theme package.

ejlayne: This is probably a good question to discuss with Travis when we start hooking up to the pipeline, and agree on an answer for this.

apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
```diff
@@ -13,11 +13,31 @@ export const CustomizeUsage: React.FunctionComponent<{}> = () => {
...
+        ArialBlack: 'Arial Black',
+        CourierNew: 'Courier New',
+        Georgia: 'Georgia',
+        TimesNewRoman: 'Times New Roman',
+      },
+    }),
+  };
+
+  const ArialBlack = Text.customize({ tokens: { variant: 'heroLargeStandard', fontFamily: CustomFontNames.ArialBlack } });
   const BrushScriptMT = Text.customize({ tokens: { variant: 'heroStandard', fontFamily: 'Brush Script MT' } });
```
ksiler, line 37: If BrushScriptMT and Wingdings aren't available on Android, we should probably remove or replace them with fonts that are available on all platforms so feel free to change them!

ksiler: Along the same lines, I'd probably change the ArialBlack text to just Arial since android doesn't support ArialBlack.

packages/theming/android-theme/src/androidTypography.ts
```diff
@@ -17,33 +17,35 @@ export function androidTypography(): Typography {
...
+      // The font weights do not work for Android since RN 0.60 (https://github.com/facebook/react-native/issues/25696)
+      // The workaround is to use font aliases from android source (https://android.googlesource.com/platform/frameworks/base/+/master/data/fonts/fonts.xml)
       regular: '400' as FontWeightValue,
-      medium: '500' as FontWeightValue,
       semiBold: '500' as FontWeightValue,
     },
     families: {
-      primary: 'Roboto',
+      primary: 'sans-serif', // sans-serif, weight=400
+      primaryBold: 'sans-serif-medium', // sans-serif, weight=500
```
ksiler: Should this be named primaryMedium instead? primaryBold makes me think the font weight would be '700' instead of '500'.

packages/theming/android-theme/src/androidTypography.ts
ksiler: @E.J.Layne was there anymore discussion here? I'm not really sure what else could be done. We had to do something similar for win32 where we needed to define a font face for semibold variants (since the font weight didn't work) and I stuck it under secondary instead of defining a new font family but perhaps I should have named it primarySemibold like it's done here.

apps/fluent-tester/src/FluentTester/TestComponents/Text/CustomizeUsage.tsx
tamasane: Added usage of supported fonts ComingSoon and Arial for Android.

packages/theming/android-theme/src/androidTypography.ts
tamasane: Yep! changed to primarySemibold in the updated change

ksiler (approved):

Issue #639, opened 2021-03-10
Use the Community Module Picker in FluentTester


Saadnajmi:
FluentTester uses the React Native Core Picker. This is deprecated, and will eventually be removed. We should use the community module picker. That has the additional complexity of linking the native module (auto linking won't work bc it's a transitive dependency), so we either need to combine the platform specific test apps into one big test app and use auto linking, or link it manually per platform. We've already linked it for macOS and iOS.

Win32 can't use the community module, so we'll keep using core there.

- **Platform's done**:
  - [x] iOS
  - [x] macOS
  - [ ] windows
  - [ ] android

### Priorities and help requested:

Are you willing to submit a PR to fix? (Yes, No)

Requested priority: (Blocking, High, Normal, Low)

Products/applications affected: (if applicable)

rurikoaraki:
Fixed with #1762

PR #664, opened 2021-04-21
Revamp Android ThemePicker

tamasane:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

This change improves android ThemePicker in the FluentTester header. 
- uses [community picker](https://www.npmjs.com/package/@react-native-picker/picker/v/1.9.11https://www.npmjs.com/package/@react-native-picker/picker/v/1.9.11), thus addresses #639 
- updated styles to be mobile-friendly
- split out platform specific code as `Themepickers.android.tsx`.

### Verification

Ran on android emulator(mobile and tablet)

https://user-images.githubusercontent.com/23552252/115585442-3bd7a500-a2e9-11eb-894a-6a89d04ac5d6.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
+++ b/apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
@@ -1,16 +1,9 @@
 import * as React from 'react';
-import { View, StyleSheet, Platform } from 'react-native';
+import { View, Picker, StyleSheet } from 'react-native';
 import { Text } from '@fluentui-react-native/experimental-text';
 import { lightnessOptions, testerTheme } from './CustomThemes';
 import { themeChoices, ThemeNames } from './applyTheme';
 import { brandOptions, OfficeBrand } from './applyBrand';
-import { Theme, useTheme } from '@fluentui-react-native/framework';
-import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
-
-const isAndroid = Platform.OS == 'android';
-
-// To be replaced with single import when Windows support comes in.
-const Picker = isAndroid ? require('@react-native-picker/picker').Picker : require('react-native').Picker;
 
 export const themePickerStyles = StyleSheet.create({
   pickerRoot: {
@@ -18,21 +11,16 @@ export const themePickerStyles = StyleSheet.create({
   },
 
   picker: {
-    flexDirection: isAndroid ? 'column' : 'row',
+    flexDirection: 'row',
     alignItems: 'center',
     padding: 4,
   },
-});
 
-export const getThemedDropdownStyles = themedStyleSheet((t: Theme) => {
-  return {
-    dropdown: {
-      height: 30,
-      width: 120,
-      fontSize: 12,
-      color: t.colors.bodyText,
-    },
-  };
+  dropdown: {
+    height: 30,
+    width: 120,
+    fontSize: 12,
+  },
 });
 
 type PartPickerEntry = { label: string; value: string };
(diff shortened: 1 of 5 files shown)


apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -41,8 +53,24 @@ export const PartPicker: React.FunctionComponent<PartPickerProps> = (props: Part
     },
     [setValue, onChange],
   );
-  return (
-    <Picker selectedValue={value} style={themePickerStyles.dropdown} onValueChange={onValueChange}>
+
+  const themedDropdownStyles = getThemedDropdownStyles(useTheme());
+  const dropdownIconColor = useTheme().colors.buttonIcon;
```
Saadnajmi, line 58:
```javascript
const theme = useTheme();
const themedDropdownStyles = getThemedDropdownStyles(theme);
const dropdownIconColor = theme.colors.buttonIcon;
```
Especially with hooks, I think the number of times you call the hook matters, as it uses order of calls to determine what to return.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -1,26 +1,38 @@
...
-  },
+export const getThemedDropdownStyles = themedStyleSheet((t: Theme) => {
+  return {
+    dropdown: {
+      height: 30,
+      width: 120,
+      fontSize: 12,
+      color: t.colors.bodyText,
+    },
+  };
```
Saadnajmi, line 35: The picker menu looked like it was light mode even in dark mode. Is there a `backgroundColor` type prop we can set here to help with that?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -1,26 +1,38 @@
...
 import { lightnessOptions, testerTheme } from './CustomThemes';
 import { themeChoices, ThemeNames } from './applyTheme';
 import { brandOptions, OfficeBrand } from './applyBrand';
+import { Theme, useTheme } from '@fluentui-react-native/framework';
+import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
+
+const isAndroid = Platform.OS == 'android';
+
+// To be replaced with single import when Windows support comes in.
+const Picker = isAndroid ? require('@react-native-picker/picker').Picker : require('react-native').Picker;
```
Saadnajmi, line 13: react-native-picker already supports windows (but not win32). I'm guessing you mean once someone updates the windows test app to support it? 
Also I feel like using `require()` inside typescript is bad practice although I'm not actually sure..

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -41,8 +53,24 @@ export const PartPicker: React.FunctionComponent<PartPickerProps> = (props: Part
...
+      onValueChange={onValueChange}
+      dropdownIconColor={dropdownIconColor}
+      mode="dropdown"
+    >
+      {contents.map((entry: PartPickerEntry, index: number) => (
+        <Picker.Item label={entry.label} value={entry.value} key={`entry${index}`} />
+      ))}
+    </Picker>
+  ) : (
+    <Picker selectedValue={value} style={themedDropdownStyles.dropdown} onValueChange={onValueChange}>
```
Saadnajmi, line 73: I know this is a minor difference from the original file, but since macOS and iOS are already split out to platform specific files, could we also do so with the android one?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
tamasane: I did this but did not push it because I got confused 😅

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
tamasane: Yep, windows test app needs to be supported.
Regarding `require()` I found that it was the best way to use shared logic around a component coming from different packages.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
tamasane: There is no prop/style of backgroundColor in picker. The background color is set based on system (Android UI mode) theme. This has something to do with reactInstanceManager updating configuration on receiving notification from system. (Even auto switching of theme doesn't work when system theme is  switched). This issue is on react-native-test-app's end (I think or I am unaware of the right way to get this configuration working on the app). Talking with @arazabishov to get this working. 

Offline I have got it working by modifying the RNTA in `node_modules`.

https://user-images.githubusercontent.com/23552252/115561699-746c8400-a2d3-11eb-80ab-026e784a8e2f.mp4

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
tamasane: done 💯

Saadnajmi (approved):

PR #667, opened 2021-04-27
Support submenu for ContextualMenu component

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Submenu.tsx
- menu mounted off a SubmenuItem
- handles dismiss events in sequence when menu item inside submenu is clicked
- parent of ContextualMenuItems

SubmenuItem.tsx
- menu item that a new Submenu will mount to
- similar to ContextualMenuItem but has a component ref and will open a submenu on hover

Focus and Ally work remaining**

### Verification

Manual test with FluentUI Tester example 
submenu
### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
Please add a page to the E2E tests so this is included in the CI pipeline. You can find more info on authoring [E2E tests here](https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/E2E). There's a ton of other pages you can follow for the pattern (it's really not difficult at all, you can just copy/paste from another test and change the name/variables to the new component name).

Saadnajmi:
Would you be able to post a screenshot of what the nested menus look like? I personally don't have easy access to my (work) windows machine since I moved for COVID and having a screenshot in the PR description would help be an easier reference!

packages/components/ContextualMenu/src/Submenu.settings.ts
```diff
@@ -0,0 +1,25 @@
...
+  {
+    tokens: {
+      backgroundColor: 'menuBackground',
+      beakWidth: 20,
+      borderColor: 'buttonBorder',
+      borderWidth: 1,
+      gapSpace: 0,
+      minPadding: 0
+    },
+    root: {
```
samuelfreiberg, line 14: Should this have **accessible: true**?

packages/components/ContextualMenu/src/Submenu.tsx
```diff
@@ -0,0 +1,91 @@
...
+  },
+  styles: {
+    root: [backgroundColorTokens, borderTokens],
+    container: []
+  },
+  render: (Slots: ISlots<SubmenuSlotProps>, renderData: SubmenuRenderData, ...children: React.ReactNode[]) => {
+    if (renderData.state == undefined) {
+      return null;
+    }
+    return <CMContext.Provider value={renderData.state.context}><Slots.root><Slots.container>{children}</Slots.container></Slots.root></CMContext.Provider>;
```
samuelfreiberg, line 87: Nit: Easier to read with each slot/component on a different line.

`return (
  <CMContext.Provider value={renderData.state.context}>
    <Slots.root>
      <Slots.container>
        {children}
      </Slots.container>
    </Slots.root>
  </CMContext.Provider>;`

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+      ...rest
+    } = userProps;
+
+    // Grabs the context information from Submenu (currently selected menuItem and client's onItemClick callback)
+    const context = React.useContext(CMContext);
+
+    const onItemClick = React.useCallback(
+      (e) => {
+        if (!disabled && (onClick || context.onItemClick)) {
+          context ?.onDismissMenu();
```
samuelfreiberg, line 43: Just out of curiosity, what is this operator do?

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+          if (onClick) {
```
samuelfreiberg, line 44: Use ternary operator?
`onClick ? OnClick() : (context.onItemClick && context.onItemClick(itemKey));`

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+            context.onItemClick && context.onItemClick(itemKey);
+          }
+          e.stopPropagation();
+        }
+      },
+      [context, disabled, itemKey, onClick],
+    );
+
+    const cmRef = useViewCommandFocus(componentRef);
+    //const cmRef = React.useRef(null);
```
samuelfreiberg, line 56: Nit: Remove if not needed?

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const onItemHoverIn = React.useCallback(
+      (e) => {
+        componentRef.current.focus();
```
samuelfreiberg, line 59: If the user doesn't give componentRef as a prop, will this cause issues with the default value of `React.useRef(null)`?

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+        onHoverIn(e);
```
samuelfreiberg, line 60: Is this an optional prop? Does onHoverIn() always exist?

lenahong:
> Looking at the screenshot -- the main menu item w/ submenuitems should probably have an indicator that there's a submenu; do we need a new slot for that or just to fill one?

I did add an image slot for the indicator. I just need to fill one

packages/components/ContextualMenu/src/SubmenuItem.tsx
lenahong: onHoverIn should always exist since SubmenuItem extends a pressable

packages/components/ContextualMenu/src/SubmenuItem.tsx
lenahong: It's not a required callback in context so it will call it if it has one

packages/components/ContextualMenu/src/SubmenuItem.tsx
lenahong: no, it won't cause an issue because it will focus on current of the newly created ref instead

PPatBoyd (approved):

ejlayne:
> > Looking at the screenshot -- the main menu item w/ submenuitems should probably have an indicator that there's a submenu; do we need a new slot for that or just to fill one?
> 
> I did add an image slot for the indicator. I just need to fill one

@lenahong Double checking on this, did you add an image/icon to indicate there's a submenu? If so, please update the image in the description so we can see what it looks like.

PPatBoyd (approved):

PR #669, opened 2021-04-29
Experimental Shimmer

chiuam:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [X] android

### Description of changes

This PR removes the iOS native shimmer module and implements a new experimental component. The component uses the Animated library from react-native for animation effects, and react-native-svg for gradient and Svgs elements. It's not supported on win32 due to incompatibility with NetUI (app freezes) and absence of a platform native driver - performance takes a big hit if we don't leverage it. 

### Usage

1. `<Shimmer elements={circle} width={500} />`
The prop 'elements' takes an array of ShimmerElement objects to build a shimmer view with one or more ShimmerElementType (rect / line, circle). Note: image is not supported as a sub-element.
3. `<Shimmer height={10} />`
If no element is passed in, the component will return a rect shimmer by default.

### Verification

https://user-images.githubusercontent.com/67026167/116898414-57558080-ac04-11eb-8081-1aa1971fbbff.mov

https://user-images.githubusercontent.com/67026167/116898434-5d4b6180-ac04-11eb-929d-226a422d133c.mov

https://user-images.githubusercontent.com/67026167/116898474-69cfba00-ac04-11eb-9130-dcef2fd915d6.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [X] Voiceover


packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+    Animated.timing(startValue, {
+      toValue: memoData.toValue,
+      duration: memoData.duration,
+      delay: memoData.delay,
+      useNativeDriver:false
+    })]),
+  ).start()
+  })
+  return startValue;
+}
```
Saadnajmi, line 24: I think you've in effect made a custom hook here? If so, you might want to rename this to something like "useShimmerAnimation" to reflect that. https://reactjs.org/docs/hooks-custom.html

packages/experimental/Shimmer/tsconfig.json
```diff
@@ -4,4 +4,4 @@
     "outDir": "lib"
   },
   "include": ["src"]
-}
+}
```
Saadnajmi, line 7: nit: newline? If only so that symbol goes away 😅

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+      [props.gradientTintColor, props.shimmerTintColor, props.containerWidth, props.containerHeight, props.toValue, props.delay, props.duration],
+    );
+
+    let startValue = renderAnimations(memoizedShimmerData);
+
+
+    if(props.uri != null) {
+      return (rest: ShimmerProps) => <Root {...mergeProps(props, rest)}>
+      <Defs>
+        <AnimatedLinearGradient id="gradient" x1={startValue} y1="0" x2="-1" y2="-1" >
```
Saadnajmi, line 60: I think the more standard syntax I've seen for conditional JSX returns looks something like this

```jsx
return (
 {condition ? (
 <JSX>...</JSX>
 ) : (
 <OtherJSX>...</OtherJSX>
)}
```
It might help with lining up indentation and such.
https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

Saadnajmi: You might also be able to do that inline and conditionally render just the <Image> or <Rect> based on props.uri. Not sure if the JSX rules works with SVG elements though 🤔

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface ShimmerTokens extends ShimmerPropTokens {
+  viewBox?: string;
+}
+
+export interface ShimmerSlotProps extends  ShimmerPropTokens {
+  root: SvgProps;
+  rect: RectProps;
+  circle: CircleProps;
+}
```
Saadnajmi, line 25: Maybe a big ask, but could we document what each of these parameters / tokens are? Either with comments above the field in the type, or in a markdown file at the root of this repo?

packages/experimental/Shimmer/package.json
```diff
@@ -16,16 +13,17 @@
...
-    "test": "fluentui-scripts jest",
     "update-snapshots": "fluentui-scripts jest -u",
     "verify-api": "fluentui-scripts verify-api-extractor",
     "update-api": "fluentui-scripts update-api-extractor"
   },
   "dependencies": {
     "@fluentui-react-native/framework": "0.5.1",
-    "@fluentui-react-native/adapters": ">=0.6.0 <1.0.0",
-    "@fluentui-react-native/tokens": ">=0.8.0 <1.0.0"
+    "@uifabricshared/foundation-compose": "^1.8.1",
```
Saadnajmi, line 22: Do we need this dependency still? It wouldn't just be "@fluentui-react-native/framework"?

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
+/** @jsx withSlots */
+import { useState, useEffect, useMemo } from "react";
+import { Circle, ClipPath, Defs, LinearGradient, Rect, Stop, Svg, Image } from 'react-native-svg';
+import { shimmerName, ShimmerProps, ShimmerType } from './Shimmer.types';
+import { mergeProps } from '../../../framework/merge-props/lib/mergeProps';
+import { buildUseStyling, compose, UseSlots, withSlots } from '../../../experimental/framework/lib';
```
Saadnajmi, line 6: nit: importing from relative path instead of "@fluentui-react-native/framework" or whatever the package name is

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,99 @@
...
+  let startValue = useState(new Animated.Value(0))[0]
+  useEffect(() => {
+  Animated.loop(
+    Animated.sequence([
+    Animated.timing(startValue, {
+      toValue: memoData.toValue,
+      duration: memoData.duration,
+      delay: memoData.delay,
+      useNativeDriver:false
+    })]),
```
Saadnajmi, line 20: nit: indentation

package-lock.json
```diff
@@ -0,0 +1,100 @@
+{
```
Saadnajmi, line 1: Delete this file

apps/fluent-tester/package.json
```diff
@@ -30,6 +30,7 @@
     "@fluentui-react-native/experimental-avatar": ">=0.7.1 <1.0.0",
     "@fluentui-react-native/experimental-button": "0.5.1",
     "@fluentui-react-native/experimental-native-button": ">=0.5.3 <1.0.0",
+    "@fluentui-react-native/experimental-shimmer": "0.1.0",
```
Saadnajmi, line 33: So I know I told you to move this to experimental.. but I wonder if this point we can just stop calling stuff made with the experimental framework "experimental" and your shimmer could just be the newest update to the existing "@fluentui-react-native/shimmer" package...

packages/experimental/Shimmer/src/Shimmer.tsx
chiuam: I removed the conditional check and added some more slots

packages/experimental/Shimmer/package.json
chiuam: Yep, no longer needed.

apps/fluent-tester/package.json
chiuam: I prefer keeping the prefix for consistency and clarity... maybe folks from CXE can weigh in

Saadnajmi: Ah, that's fair. If that's the case, can we not delete the original shimmer, keep the experimental shimmer, and delete the original shimmer in a followup?

I don't feel too strongly about this, so not blocking.

Saadnajmi: Reason partially being, I don't know how deleting a package from a monorepo behaves with NPM. Especially if we are probably going to bring it back up later when `experimental-shimmer` just becomes `shimmer`

packages/experimental/Shimmer/src/Shimmer.tsx
chiuam: yep, renamed

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: Not at all, added documentation.

apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.tsx
```diff
@@ -1,50 +1,73 @@
...
+  }
   const CustomizedShimmer = Shimmer.customize({
-    shimmerAlpha: 0,
-    shimmerWidth: 180,
-    shimmerAngle: 0,
-    shimmerSpeed: 400,
-    shimmerDelay: 0.1,
-    viewTintColor: 'rgb(100, 100, 100)',
-    cornerRadius: 10,
+    gradientTintColor: "pink",
```
chiuam, line 58: Token customization isn't working for some reason...

packages/experimental/Shimmer/src/Shimmer.styling.ts
```diff
@@ -0,0 +1,52 @@
...
+        toValue: tokens.toValue,
+        duration: tokens.duration,
+        delay: tokens.delay,
+        shimmerTintColor: tokens.shimmerTintColor,
+        width: tokens.width,
+        height: tokens.height,
+        angle: tokens.angle,
+        gradientOpacity: tokens.gradientOpacity,
+        accessibilityRole: 'progressbar',
+        accessibilityLabel: "Loading content",
```
Saadnajmi, line 31: This isn't localized, and I have a feeling that could cause problems down the line. What happens if we just set the role and not the label? And if this is required, I think downstream in sdx_platform there is localization code set up so we could defer responsibility to clients to set the label until we have a better story here.

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+export enum ShimmerElementType {
+  /**
+   * Rect element type
+   */
+  rect = 1,
+  /**
+   * Circle element type
+   */
+  circle = 2,
+}
```
Saadnajmi, line 18: nit: I've seen people do something more like this for enums
```
export ShimmerElementType = 'rect'  | 'circle' 
```

I'm not sure, but maybe that's more the more standard way to do enums now? I'm not sure of the pros/cons but I'd love to learn more about it if others know..

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Represents the shimmer element
+ */
+export interface ShimmerElement {
+  /**
+   * Type of the shimmer element
+   */
+  type?: ShimmerElementType;
+  /**
+   * Specifies the element width, non-op for circle element
```
Saadnajmi, line 28: width in points? pixels?

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  yPos?: number;
+}
+
+export interface ShimmerTokens {
+  /**
+   * The final value for the animation to reach the end
+   * @defaultValue '30'
+   */
+  toValue?: number;
```
Saadnajmi, line 59: I don't think I understand. Is this a duration? or the # of steps in the animation?

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Height of the shimmer view
+   * @defaultValue '100'
+   */
+  height?: number;
+  /**
+   * Specifies the gradient angle, value should be less than 1
+   * @defaultValue '0'
+   */
+  angle?: number;
```
Saadnajmi, line 94: if less than 1, what units are these in? radians? does 0->1 map from 0 degrees to 90 degrees?

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+  circle: CircleProps;
+  image: ImageProps;
+}
+
+export interface ShimmerProps extends ShimmerTokens {
+  /**
+   * Image to be used as a shimmer element
+   * @defaultValue 'null'
+   */
+  uri?: ImageURISource;
```
Saadnajmi, line 114: Is the image something like a template image that gets tinted?

PPatBoyd (changes requested):

apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.tsx
```diff
@@ -1,50 +1,73 @@
...
...1eB7V4LVi5kGC2MACuere8I/eufov9az9e/wCQvcf7w/kKo0UUUV//1PeaKKKKK1/C02y/eIniROPqOf8AGtCWHZ4mglHSWNs/UDH+FU/Fv/H1b/8AXM/zrIt4nnnSGMZZzgV2ltBHb2yQIPlUY+tZGraIGzNZLg9TEOh+n+FYCo5kEYU7ydu3vn0rRvNFu7eLzARKAPmCDkf41m0lFFb3hH71z9F/rWfr3/IXuP8AeH8hVGiiiiv/1feaKKKKKnsZfIvIZs42OCfp3rsJog88Ew5MZP5Fcf4Vh+LP+Pq3/wCuZ/nUvhezwrXjjk/LH9O5rdoqFrW3a5FyYUMw6PjmpqzNT0iG6zJFiKb1xw31/wAa5u6tpraXy5kKnt6H6VDRW94R+9c/Rf61n69/yF7j/eH8hVGiiiiv/9b3miiiiiiuz0mbz9Ogkzk7cN9RxWP4rGbq3Hqh/nW/DGsUSRIMKigCn0UUUVFdW8NzEY5kDL/Kub1TR5rbMsOZYf8Ax5frWXW94R+9c/Rf61n69/yF7j/eH8hVGiiiiv/X95ooooooro/Ck263mgJ+424fQ/8A6qh8Vq/nwOFJAQjOO+aVdflCgNZhjjk7yP6Uv/CQSf8APj/5EP8AhR/wkEn/AD4/+Pn/AAo/4SCT/nx/8fP+FH/CQSf8+P8A4+f8KP8AhIJP+fH/AMiH/Cj/AISCT/nx/wDIh/wo/wCEgk/58v8Ax8/4Vl6hNHcyCSO08g/xbTkH8MVq+EkYC4YqQp2gEj61m69/yF7j/eH8hVGiiiiv/9D3miiiiiir2i3i2V55kgbYylWx1rbOu2B6+b/37o/tzT/+mv8A37o/tzT/APpr/wB+6P7c0/8A6a/9+6P7c0//AKa/9+6P7c0//pr/AN+6P7c0/wD6a/8Afuj+3NP/AOmv/fuj+3NP/wCmv/fuj+3NP/6a/wDfuj+3bD1l/wC/dc9qU63N9LOgIVzkA9ar0UUUV//R95oooooooooooooooooooooooor/0veaKKKKKKKKKKKKKKKKKKKKKKKK/9P3miiiiiiiiiiiiiiiiiiiiiiiiv/U95oooooooooooooooooooooooor/1feaKKKKKKKKKKKKKKKKKKKKKKKK/9b3miiiiiiiiiiiiiiiiiiiiiiiiv/X95oooooooooooooooooooooooor/2Q==',
```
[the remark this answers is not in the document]
Saadnajmi: It's a base64 encoded image, basically take a pull PNG and turn it into a string. It's used in a few places like PersonaCoin for the status indicators, and experimental-avatar for the image of the profile.
I put one of these in the top of the AvatarTest page, but there's probably some better place to put assets like this than directly in code 😅 I just didn't spend the time to look into it and now ppl follow my example :D

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+export function useShimmerAnimation(memoData: any) {
+  let startValue = useState(new Animated.Value(0))[0]
+  useEffect(() => {
+    Animated.loop(
+      Animated.sequence([
+        Animated.timing(startValue, {
+          toValue: memoData.toValue,
+          duration: memoData.duration,
+          delay: memoData.delay,
+          useNativeDriver: false,
```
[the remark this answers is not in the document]
Saadnajmi: There's a bug related to looping with the native driver https://github.com/microsoft/react-native-macos/pull/763
There's an open PR in Facebook upstream that fixes it, so I think Amanda was working on bringing it to react-native-macos before turning it on.

apps/fluent-tester/package.json
Saadnajmi: I think it was mostly stuff like the CHANGELOG.JSON carrying over, or the API changing a bunch between 2 versions of shimmer.

If we deleted old, then told clients to use experimental-shimmer, then eventually moved `experimental-shimmer` to just be `shimmer` I wanted to be sure that didn't confuse people or the NPM registry. It sounds like it won't confuse the latter, and I'm pretty sure nobody consumed the old shimmer so we should be good on both fronts.

That being said... how do we feel about dropping the `experimental` prefix altogether? It seems like the experimental compose framework is just... the framework now, and is even named as such ( `@fluentui-react-native/framework`). There isn't a precedent for new controls using the new framework to be in `packages/components` instead of `packages/experimental` (besides the old shimmer, which I wrote with the new framework) but... we can set one?

Saadnajmi: I also just remembered: I made the original Shimmer component with the new compose framework but put it in `packages/components` after asking @FalseLobster where to put it. and I only put `experimental-avatar` in `packages/experimental` to keep it distinguished from PersonaCoin (I still want to rectify those both together someday..). 
Others can chime in but I think we're at the point that we don't need to put new stuff in `packages/experimental` just because it's using the compose v2 framework?

packages/experimental/Shimmer/src/Shimmer.styling.ts
chiuam: Good catch. If I don't specify the label, VO would just read out the accessibility role label. The prop isn't required here so I will remove it.

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: In dp, if clients want they can use PixelRatio to get the device pixel density and calculate the dimensions based on that?

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: Read up on enums in typescript. I learned that numeric enums aren't type-safe and it's unnecessary in our case so it's probably better to use union types.

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: Yeah it's more like the distance to travel from startValue. For the default case, the animation will animate from 0 to 30 over the course of 7000 milliseconds(duration). Since speed = distance / time, speed ≈ 0.004 (higher the speed, faster it moves). I should probably reword it..

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: No, it doesn't get tinted but Svg supports image tinting if there's need for it

apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.tsx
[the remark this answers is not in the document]
chiuam: Yeah, looks like it's more common to use uri data images than static resources in the project. I'd like to know the standard way of using images for test as well

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <ClipPath id="shimmerView">
+            {rows}
+            {rect && <Slots.rect width={props.rect.width} x={props.rect.x} y={props.rect.y} />}
+            {circle && <Slots.circle cx={props.circle.cx} cy={props.circle.cy} r={props.circle.r} />}
+          </ClipPath>
+          <Rect
+            x="0"
+            y="0"
+            width="100%"
+            height="100%"
```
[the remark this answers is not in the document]
chiuam: These are the default values and should be static. x y are the coordinates of the Svg object in the parent view, width height are the dimensions which should be shown in full-scale for all cases.

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: I think I was worried that a shimmer's tint and the image's tint might be different shades of grey / not match in light/dark mode without tinting. It might not be a problem that all though, and could be a followup if it ever does become a problem

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: Oh wow! Today I learned that the unions are more type safe. Thanks for looking into it :D

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: I was more being nitpicky about having units in documentation, but also because I truly did not know

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: That's helpful. I think I'd even appreciate just the default example you specified in documentation.
Like if you put "For the default case, the animation will animate from 0 to 30 over the course of 7000 milliseconds(duration). Since speed = distance / time, speed ≈ 0.004 (higher the speed, faster it moves)" as a comment

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+          if (element.type == ShimmerElementType.rect) {
+            rows.push(<Slots.rect key={i} width={element.width} height={element.height} x={element.xPos} y={element.yPos} rx={element.borderRadius} ry={element.borderRadius} />);
+          } else if (element.type == ShimmerElementType.circle) {
+            rows.push(<Slots.circle key={i} r={element.height / 2} cx={element.xPos} cy={element.yPos} />);
+          }
+        }
+      }
+      return (
+        <Slots.root {...mergedProps}>
+          <Defs>
```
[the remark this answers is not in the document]
chiuam: Very good point. Since startValue uses the state hook which causes re-render on every state change, this does cause performance issues. I learned that useRef does the same thing as useState - preserves the data after a re-render, except it doesn't cause re-render. Since we are not updating the component on state change, useRef is probably the more suitable approach here.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <AnimatedLinearGradient id="gradient" x1={startValue} y1={memoizedShimmerData.angle} x2="-1" y2="-1" >
+              <Stop offset="10%" stopColor={uri ? memoizedShimmerData.gradientTintColor : memoizedShimmerData.shimmerTintColor} stopOpacity={uri ? "0" : "1"} />
+              <Stop offset="20%" stopColor={memoizedShimmerData.gradientTintColor} stopOpacity={memoizedShimmerData.gradientOpacity} />
+              <Stop offset="30%" stopColor={uri ? memoizedShimmerData.gradientTintColor : memoizedShimmerData.shimmerTintColor} stopOpacity={uri ? "0" : "1"} />
+            </AnimatedLinearGradient>
+          </Defs>
+          {uri && <Slots.image href={props.uri} />}
+          <ClipPath id="shimmerView">
+            {rows}
+            {rect && <Slots.rect width={props.rect.width} x={props.rect.x} y={props.rect.y} />}
```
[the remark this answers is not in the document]
chiuam: Yeah, I thought it'd be cleaner for the client to construct a group view with an array, also I believe that's how Fluent web does it too.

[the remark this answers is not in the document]
Saadnajmi: Ah, I see that [FluentUI React Shimmer](https://developer.microsoft.com/en-us/fluentui#/controls/web/shimmer) does it with the props, and [FluentUI React Northstar Skeleton](https://fluentsite.z22.web.core.windows.net/0.54.0/components/skeleton/definition) does it with children. That's probably why I got slightly confused. Thanks!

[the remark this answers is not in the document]
chiuam: Thanks for sharing the northstar version, I didn't know about it! One thing I like about using props is that the sub-elements are reusable on the client side, kinda like the example I have in the test page, (not sure if that's possible with children). I have a feeling that the northstar version should be the gold standard we should follow though...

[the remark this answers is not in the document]
Saadnajmi: Reusable in that you can use the same element in multiple shimmers? Or that You can render an individual element? I think there's ways to do both. I also agree that Northstar is probably the thing to base off of, but I get that's a fairly big API change too and probably out of scope if we already like our props design

packages/experimental/Shimmer/src/Shimmer.tsx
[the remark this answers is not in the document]
chiuam: Talked offline. I'm still not sure if rendering the animation redraws the SVG, opened https://github.com/microsoft/fluentui-react-native/issues/675  to track this.

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -0,0 +1,135 @@
...
+export const shimmerName = 'Shimmer';
+/**
+ * Specifies the possible type of the shimmer elements.
+ * Required when building more than one element at a time.
+ */
+export enum ShimmerElementType {
+  /**
+   * Rect element type
+   */
+  rect = 1,
```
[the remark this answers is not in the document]
chiuam: used union types instead, I think the convention is to use camel case

packages/experimental/Shimmer/src/Shimmer.tsx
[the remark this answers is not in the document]
chiuam: The former. I'm not sure what the advantages of using children are aside from readability, I'm just leaning towards props for now until we have stronger reasons to do otherwise.

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: I'm not sure about the units tbh. For the mapping, y1 should be anywhere from -1 to 1, -1 being 90 degrees. I think this is something we can iterate on though.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            rows.push(<Slots.rect key={i} width={element.width} height={element.height} x={element.xPos} y={element.yPos} rx={element.borderRadius} ry={element.borderRadius} />);
+          } else if (element.type == ShimmerElementType.circle) {
+            rows.push(<Slots.circle key={i} r={element.height / 2} cx={element.xPos} cy={element.yPos} />);
+          }
+        }
+      }
+      return (
+        <Slots.root {...mergedProps}>
+          <Defs>
+            <AnimatedLinearGradient id="gradient" x1={startValue} y1={memoizedShimmerData.angle} x2="-1" y2="-1" >
```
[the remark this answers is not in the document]
chiuam: The four coordinates represent the starting and ending positions of the gradient. 

1. Gradient is vertical when y1 == y2 && x1 != x2 
2. Gradient is horizontal when x1 == x2 && y1 != y2 
3. Gradient is angled when x1 != x2 && y1 != y2 

Since we are animating the x1 value, horizontal gradient support might not be available in this iteration. Moving y1 from 0 to -1 creates the angled effect.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            {rows}
+            {rect && <Slots.rect width={props.rect.width} x={props.rect.x} y={props.rect.y} />}
+            {circle && <Slots.circle cx={props.circle.cx} cy={props.circle.cy} r={props.circle.r} />}
+          </ClipPath>
+          <Rect
+            x="0"
+            y="0"
+            width="100%"
+            height="100%"
+            fill="url(#gradient)"
```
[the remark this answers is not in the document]
chiuam: I think it looks up the tag that references the fill color - https://github.com/react-native-svg/react-native-svg#lineargradient

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: I think dp is the default 'unit' in react native, not sure if it's redundant to specify that in doc

packages/experimental/Shimmer/src/Shimmer.tsx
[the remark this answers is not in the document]
Saadnajmi: I think the main benefit is clients have stronger control over the children, where they can style each child individually / layout circles and lines and rects how they see fit. 
It sounds like we weren't going for that with this Shimmer, though I still think it's a useful followup in the future, if it comes up as a client need.

PPatBoyd (approved):

packages/experimental/Shimmer/package.json
```diff
@@ -1,10 +1,7 @@
 {
-  "name": "@fluentui-react-native/shimmer",
-  "version": "0.6.3",
-  "description": "A cross-platform Shimmer component using the Fluent Design System",
-  "license": "MIT",
-  "author": "Microsoft <fluentuinativeowners@microsoft.com>",
-  "homepage": "https://github.com/microsoft/fluentui-react-native",
+  "name": "@fluentui-react-native/experimental-shimmer",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
HeyImChris: True, with the rename it's probably fine then

HeyImChris (approved):

Saadnajmi:
I just remembered. This will cause the NuGet pipeline to fail because it will try to pack the static library for the original Shimmer but not find it because we deleted it :( 

NuGet pipeline was broken anyway so I guess we should go fix it : ' )

PR #682, opened 2021-05-11
Use SwiftUI Avatar implementation from FluentUI Apple

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

WIP to test pipeline / NuGet consumption.

### Description of changes

This change updates our Avatar component (which was a native module wrapping the FluentUI Apple Avatar) to use the new FluentUI Apple Vnext SwiftUI implementation of Avatar. I also cleaned up the test page for Avatar and made it look a lot nicer.

The change is non-trivial: It required me to define a new obbj-c class and macros to pass properties from Javascript to the SwiftUI View. The class will need to be implemented one per every new component. The macros can be reused by future SwiftUI components. The macros and implementation were inspired by https://betterprogramming.pub/react-native-meets-swiftui-d1606a8e1681

Explanation:
SwiftUI talks to normal Swift / obj-c through a proxy / viewWrapper class. The proxy class gives you access to a state object, and a UIView that contains the SwiftUI view. The important bit: You can access the view from the proxy class, but you cannot access the proxy class from the view.
```
Proxy Class (MSFAvatar)
|-> View (UIView  wrapper around SwiftUI View)
|-> State (Used to set properties on the SwiftUI View)
```

This does not work well with React Native. React Native View Managers expect to be able to set properties directly on the view via the view class, and know not about this indirection through the proxy object. To solve this, I created an extension on the proxy class that I named `MSFAvatarStorage` that allows me to store a key value pair of view->proxy classes. This means that given a view, I can now look up it's proxy object, then access it's state object to set properties.

```
Proxy Class (MSFAvatar)
|-> View (UIView  wrapper around the SwiftUI view)
|-> State (used to set properties on the SwiftUI view)
|-> Storage (view->proxy object mappings)
```

The last bit I needed was extra macros to actually do this indirection:
```
#define RCT_EXPORT_SWIFTUI_PROPERTY(name, type, proxyClass) \
 RCT_REMAP_VIEW_PROPERTY(name, __custom__, type)  \
 - (void)set_##name:(id)json forView:(UIView *)view withDefaultView:(UIView *)defaultView RCT_DYNAMIC { \
     NSMutableDictionary *storage = [proxyClass storage]; \
     proxyClass *proxy = storage[[NSValue valueWithNonretainedObject:view]];  \
     proxy.state.name = [RCTConvert type:json]; \
 }
 
 ...
 
 // In MSFAvatarViewManager
 RCT_EXPORT_SWIFTUI_PROPERTY(primaryText, NSString, MSFAvatar)
 
 ```

And now we have a working SwiftUI based native component in React Native!

### Verification

Rewrote the test app to test the SwiftUI based props. They work!

https://user-images.githubusercontent.com/6722175/120049216-e7f26700-bfde-11eb-855f-c0d9f5ab4166.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### CODE REVIEW

**Comment by HeyImChris** [comment]
"It needed a bunch of extra code which I'll not describe for now bc this is a WIP PR"

Now the PR is Ready for Review, could be worth going into more details?

File: .ado/azure-pipelines.yml [hunk 1]
```diff
@@ -146,3 +146,35 @@ jobs:
           publishJUnitResults: false
           javaHomeOption: 'JDKVersion'
           sonarQubeRunAnalysis: false
+  # Uncomment this job to test changes to changes to the NuGet package
```
HeyImChris, line 149: extraneous "to changes"?

apps/macos/src/Podfile
```diff
@@ -24,8 +24,8 @@ use_test_app! do |target|
   target.app do
     platform :osx, '10.14'
 
-    pod 'FluentUI-React-Native-Button', :path => '../../../packages/experimental/NativeButton/FluentUIReactNativeButton.podspec'
     pod 'FluentUI-React-Native-Avatar', :path => '../../../packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec'
+    pod 'FluentUI-React-Native-Button', :path => '../../../packages/experimental/NativeButton/FluentUIReactNativeButton.podspec'
     pod 'FluentUI-React-Native-FocusZone', :path => '../../../packages/components/FocusZone/FluentUIReactNativeFocusZone.podspec'
     pod 'FluentUI-React-Native-Radio-Button', :path => '../../../packages/components/NativeRadioButton/FluentUIReactNativeRadioButton.podspec'
     pod 'FluentUI-React-Native-MenuButton', :path => '../../../packages/components/MenuButton/FluentUIReactNativeMenuButton.podspec'
```
HeyImChris, line 31: Looks like you alphabetized the Button which is awesome. Might as well fix up the rest of them while we're here (e.g. Menu is before Radio)

packages/experimental/Avatar/ios/MSFAvatarStorage.h
```diff
@@ -0,0 +1,23 @@
+//
+//  MSFAvatarStorage.h
+//  Pods
+//
+//  Created by Saad Najmi on 5/11/21.
```
HeyImChris, line 5: Delete "Created by" line for new files

packages/experimental/Avatar/ios/AvatarViewManager.swift
```diff
@@ -15,12 +23,12 @@ class AvatarViewManager: RCTViewManager {
 	override func constantsToExport() -> [AnyHashable : Any]! {
 		return [
 			"sizes" : [
-				"xSmall" : AvatarSize.extraSmall.size.width,
-				"small" : AvatarSize.small.size.width,
-				"medium" : AvatarSize.medium.size.width,
-				"large" : AvatarSize.large.size.width,
-				"xLarge" : AvatarSize.extraLarge.size.width,
-				"xxLarge" : AvatarSize.extraExtraLarge.size.width
+				"xSmall" : 16,
```
HeyImChris, line 26: The way it is now, FURN And FUA sizes can get out of sync which might be something worth avoiding if we can pull these directly from FUA?

packages/experimental/Avatar/ios/MSFAvatarStorage.m
```diff
@@ -0,0 +1,30 @@
+//
+//  MSFAvatarStorage.m
+//  FluentUI-React-Native-Avatar
+//
+//  Created by Saad Najmi on 5/11/21.
```
HeyImChris, line 5: Delete created by

packages/experimental/Avatar/ios/MSFAvatarViewManager.m
```diff
@@ -1,82 +1,105 @@
...
+    NSMutableDictionary *storage = [proxyClass storage]; \
+    proxyClass *proxy = storage[[NSValue valueWithNonretainedObject:view]];  \
+    proxy.state.keyPath = [RCTConvert type:json]; \
+}
+
+#define RCT_EXPORT_CUSTOM_SWIFTUI_PROPERTY(name, type, proxyClass) \
+RCT_REMAP_VIEW_PROPERTY(name, __custom__, type)  \
+- (void)set_##name:(id)json forView:(UIView *)view withDefaultView:(UIView *)defaultView RCT_DYNAMIC
+
+
```
HeyImChris, line 29: nit- Probably don't need this many empty lines between the API and the next line?

HeyImChris (changes requested):

Saadnajmi:
> "It needed a bunch of extra code which I'll not describe for now bc this is a WIP PR"
> 
> Now the PR is Ready for Review, could be worth going into more details?

Good catch. I updated the description with details on the new MSFAvatarStorage class and the new macros I wrote.

packages/experimental/Avatar/ios/AvatarViewManager.swift
Saadnajmi: The SwiftUI Avatar sizes gets its' size values from the token pipeline inside FUA. We've had the question before of how we reconcile the setting of tokens between FURN and FUA in the case of wrapped FUA views. The direction I've been going towards is to have the source of truth be the token pipeline inside FURN, as ideally they should match. On that note, something I could investigate next is to get those tokens into FURN so that I don't have to hardcode this size ramp inside the swift file and can do it in JS instead.

packages/experimental/Avatar/ios/AvatarViewManager.swift
```diff
@@ -3,9 +3,17 @@ import FluentUI
 
 @objc(MSFAvatarViewManager)
 class AvatarViewManager: RCTViewManager {
+	
 	override func view()->UIView! {
-		let avatarView = AvatarView(avatarSize: .small)
-		return avatarView
+		let viewWrapper = MSFAvatar()
```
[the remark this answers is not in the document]
Saadnajmi: Github interprets tabs at 8 spaces for some reason.  I could put a .editorconfig file in the FURN repo root to fix that but I just haven't. We also don't really run a linter or anything on any FURN apple code to fix these sort of things.

packages/experimental/Avatar/ios/AvatarViewManager.swift
Saadnajmi: I don't think sizeThatFits is applicable here. The problem is two layout systems: UIKit and Yoga/flexbox

I need the sizes because I need to hardcode the `width / length` of the react native view using flexbox (how the web / React Native does layout). My solution previously was to pass the size from native so that I could set the width and height in React Native. There's probably a better, more complex solution to layout involving subclassing RCTShadowView (a class that React Native internally uses to layout views). I haven't investigated that.

rdeassis: I see. Just to be clearer on the approach I meant with sizeThatFits:
After you call sizeThatFits, shouldn't be able to pass the dimensions to the React Native layer since you'll have the view in the right size?

Saadnajmi: Are the numbers I have for each enum not the correct height/width (including rings)? Should I instead be instantiating an Avatar at each size, calling sizeThatFits, and passing the height/width from there?

HeyImChris, line 26: If these numbers are supposed to be in sync with something then it's probably worth a comment as to where we keep them in sync (e.g. FUA Avatar)

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -1,132 +1,102 @@
...
    */
-  borderColor?: ColorValue;
+  isRingVisible?: boolean;
 
   /**
-   * When true, the presence status border is opaque. Otherwise, it is transparent.
+   * A boolean indicating whether the Inner Ring Gap is transparent
    */
-  usesOpaquePresenceBorder?: boolean;
+  isTransparent?: boolean;
```
HeyImChris, line 78: Pretty misleading name here if it only applies to the inner ring gap and not the Avatar control as a whole being transparent

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -1,132 +1,102 @@
...
   /**
-   * Used when avatarView doesn't have image or can't generate initials string
+   * A boolean indicating out of office status
    */
-  preferredFallBackImageStyle?: string;
+  isOutOfOffice?: boolean;
 
   /**
-   * Set to true to enable the iPadOS pointer interactions on the avatar view, false by default.
+   * A boolean indicating whether the gap between the avatar and ring is visible.
```
HeyImChris, line 86: `isInnerRingGapHidden` would be the more Appley way of phrasing this API. Is this API derived from somewhere else or can we change the names?

HeyImChris (approved):

packages/experimental/Avatar/ios/AvatarViewManager.swift
Saadnajmi: That sounds good. I can at the very least leave a comment here why these are exported and where they come from.

packages/experimental/Avatar/src/Avatar.tsx
Saadnajmi: The prop name comes from the FluentUI Apple control and maps 1:1. I can't really change that.

packages/experimental/Avatar/src/Avatar.tsx
Saadnajmi: The prop name comes from the FluentUI Apple control and maps 1:1. I can't really change that.

rdeassis (approved):

packages/experimental/Avatar/ios/AvatarViewManager.swift
Saadnajmi: I added a comment. Thanks for the pushing this to be more clear.

packages/experimental/Avatar/src/Avatar.tsx
rdeassis: This one applies on only to the ring inner gap, but also to the outer gap of the ring along with the presence icon (if available) and its outline.

packages/experimental/Avatar/ios/AvatarViewManager.swift
rdeassis: These are the sizes of the avatar (without considering the ring/gaps/presence if available).
`size: [ xSmall: 16pt, small: 24pt, medium: 32pt, large: 40pt, xlarge: 52pt, xxlarge: 72pt ]`

We do intent to change the frame on the avatar to be fixed (to the largest possible for the size enum specified).

PR #686, opened 2021-05-12
Add bundle typechecking

afoxman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Add a new `rnx-kit` Metro plugin which runs TypeScript when building a bundle.

### Verification

Validated against all `apps` by running bundling. Simulated type failures and observed the expected FAIL result.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
The CI failure where it can't find "@fluentui-react-native/tester" *might* be because some package versions locally are out of sync with what's in NPM. Our NPM publish pipeline stopped working recently, and in my attempt to fix it I think a few packages (including @fluentui-react-native/tester) are now locally a higher version than what is published in NPM (For example, @fluentui-react-native/tester is 0.23.3 locally but 0.23.1 in npm).  Perhaps the bundle is failing because it can't find the correct package number.

If that is the case, the fix would be for me to go introduce a bump to the packages that got out of sync.

Saadnajmi:
> The CI failure where it can't find "@fluentui-react-native/tester" _might_ be because some package versions locally are out of sync with what's in NPM. Our NPM publish pipeline stopped working recently, and in my attempt to fix it I think a few packages (including @fluentui-react-native/tester) are now locally a higher version than what is published in NPM (For example, @fluentui-react-native/tester is 0.23.3 locally but 0.23.1 in npm). Perhaps the bundle is failing because it can't find the correct package number.
> 
> If that is the case, the fix would be for me to go introduce a bump to the packages that got out of sync.

#688 Fixes that!

Saadnajmi:
OK.. so the CI is still failing with the new package versions published too. I guess that wasn't the error :'(

afoxman:
Hi Saad, thank you for taking a look at the errors for me. The problem is in the TypeScript plugin I wrote (coming from rnx-kit repo). It assumes that all other packages in the repo are built and that .d.ts files are available, but that's not the case in FURN. I'm working on a fix for it and will update the version number in this PR once the fix is done/tested.

afoxman:
The way FURN builds and bundles is going to require more changes to the plugin. For now, I'm going to close this PR.

Saadnajmi:
> Hi Saad, thank you for taking a look at the errors for me. The problem is in the TypeScript plugin I wrote (coming from rnx-kit repo). It assumes that all other packages in the repo are built and that .d.ts files are available, but that's not the case in FURN. I'm working on a fix for it and will update the version number in this PR once the fix is done/tested.

@afoxman does running `yarn build` at the repo root before bundling not generate the .d.ts files you're looking for?

afoxman:
> @afoxman does running `yarn build` at the repo root before bundling not generate the .d.ts files you're looking for?
Yes, that produces the missing files. It also runs a full typescript pass, duplicating all of the typescript work being done in Metro, so that didn't seem like a good idea.

PR #683, opened 2021-05-11
WIP: Fix Pipelines

Saadnajmi:
Work In Progress, tryna get NPM to publish again

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

(a summary of the changes made, often organized by file)

### Verification

(how the change was tested, including both manual and automated tests)

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Screenshot or description before this change | Screenshot or description with this change |

### Pull request checklist

This PR has considered (when applicable):
- [ ] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #688, opened 2021-05-13
Manually bump packages that are out of sync with NPM

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

We had an issue with our NPM Publish pipeline that I tried to fix with #683 . Due to some weirdness with beachball, our publish bot, and me running an ADO Publish pipeline manually, that PR got merged prematurely. I then made #685 to fix the worst fallout of that premature PR. This PR is a second one to fix more fallout.

In #683, a bunch of packages were bumped locally, but not published to NPM. Examples are "@fluentui-react-native/tester", which was at 0.23.1 in NPM but 0.23.3 locally. 

This PR goes to the affected packages, and bumps them one more time with a no-op change so that our NPM Publish pipeline will publish the latest packages and we can be in sync again.

Note: I created these no-op changes by going to the `index.tsx` file of each package, adding a `//no-op` comment to the file , creating a commit with those no-op comments, running `beachball change` to generate the change files, and then removing the no-op comments and committing. The end result should be just change files and maybe a few minor new lines or extraneous comment removals.

### Verification

Locally ran `yarn publish:beachball` at the repo root, which can't publish because I don't have the necessary NPM token to do so. What it will do is show the state of the repo after a publish locally, and things looked good to me. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Shimmer/src/index.ts
+++ b/packages/experimental/Shimmer/src/index.ts
@@ -1,2 +1,3 @@
 export * from './Shimmer';
 export * from './Shimmer.types';
+//no-op
--- a/packages/experimental/Shimmer/src/index.ts
+++ b/packages/experimental/Shimmer/src/index.ts
@@ -1,3 +1,2 @@
 export * from './Shimmer';
 export * from './Shimmer.types';
-//no-op
--- a/apps/fluent-tester/src/index.ts
+++ b/apps/fluent-tester/src/index.ts
@@ -1,1 +1,2 @@
 export * from './FluentTester';
+//no-op
--- a/apps/fluent-tester/src/index.ts
+++ b/apps/fluent-tester/src/index.ts
@@ -1,2 +1,1 @@
 export * from './FluentTester';
-//no-op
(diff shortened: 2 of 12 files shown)


packages/experimental/Shimmer/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from './Shimmer';
-export * from './Shimmer.types';
\ No newline at end of file
+export * from './Shimmer.types';
```
HeyImChris, line 2: Is this file's change worth pushing up? Better to avoid needless edits to files if we don't need them to not muddy the file's history

HeyImChris (approved):

packages/experimental/Shimmer/src/index.ts
Saadnajmi: My local copy of prettier or eslint actually won't let me save this file without adding a new line... so yes? 😅

HeyImChris: Sounds good! If the linter thinks we need it then I think it's fair to have it

PR #689, opened 2021-05-13
Android button

tamasane:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

This change makes Button android-complaint by matching the redlines and effects. Some important highlights of the change are: 
- Separated files from default:  This change introduces _Button.android.tsx_ and _Button.settings.android.ts_ for platform specific code. It also separates `createIconProps` into _Button.icon.ts_ for code sharing. The separate files are needed to support structural changes as android button requires an additional layer of `Pressable` for ripple effect. Platform specific files are considered over `Platform.select` in this scenario to support more customisation (with v-next token pipeline coming) without impacting the bundle size.
- Pressable needs to be wrapped in a View: The ripple effect of Pressable overflows outside of the specified borderRadius. The workaround for this is to wrap it in a View with `overflow: hidden`.
- The button is accessible via Keyboard Tab/Swipe gesture navigation and Talkback makes the correct announcements.
- This PR has not considered v-next tokens/theme.
- Icon is updated to apply styles to Svg icons. The margin and padding styles could not be applied directly to `<SvgUri />`, so wrapped into `<View>` as a workaround.

### Verification

| Before                                       | After (Light)                                      | After (Dark)                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
|  |  |  |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/Button.android.tsx
```diff
@@ -0,0 +1,82 @@
...
+            {info.content && <Slots.content />}
+            {children}
+          </Slots.stack>
+        </Slots.ripple>
+      </Slots.root>
+    );
+  },
+  slots: {
+    root: View,
+    ripple: Pressable,
```
tamasane, line 68: When the button has Pressable as root, the ripple effect overflows. 

Pressable-ripple-leaks

packages/components/Button/src/Button.settings.android.ts
```diff
@@ -0,0 +1,88 @@
...
+      accessible: true,
+      focusable: true,
+      accessibilityRole: 'button',
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      android_ripple: {
+        color: 'buttonBackgroundPressed',
+      },
+    },
+    content: {
+      // accessible: false,
```
samuelfreiberg, line 35: Nit - remove comments if not needed anymore

packages/components/Button/src/Button.android.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Text } from '@fluentui-react-native/text';
+import { settings } from './Button.settings';
+import { backgroundColorTokens, borderTokens, textTokens, foregroundColorTokens, getPaletteFromTheme } from '@fluentui-react-native/tokens';
+import { filterViewProps } from '@fluentui-react-native/adapters';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { useAsPressable, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import { Icon } from '@fluentui-react-native/icon';
+import { createIconProps } from './Button.icon';
+
+export const Button = compose<IButtonType>({
```
Saadnajmi, line 16: Is the difference between the base button and the android one so big that we actually need  a platform specific file? It will be hard to track differences between the two as they evolve.

Saadnajmi, line 16: I also wonder if we should be making these modfications to `experimental-button` since that uses the new compose framework, and seems to be the direction we want to go when I asked. 
I personally couldn't get styling to work properly on the experimental button with the apple theme, so the apple them mostly works well with this button.

packages/components/Button/src/Button.android.tsx
Saadnajmi: I'm not sure I see the overflow in that video?

Saadnajmi: Also, does this comment thread give any insight in general? https://github.com/microsoft/fluentui-react-native/pull/467/files#r499365330

packages/components/Button/src/Button.icon.ts
```diff
@@ -0,0 +1,22 @@
+import { Image } from 'react-native';
+import { RasterImageIconProps, IconProps } from '@fluentui-react-native/icon';
```
Saadnajmi, line 2: Did we not have icon support on the original FURN button before?

packages/components/Button/src/Button.settings.android.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const settings: IComposeSettings<IButtonType> = [
+  {
+    tokens: {
+      backgroundColor: 'buttonBackground',
+      color: 'buttonText',
+      borderColor: 'buttonBorder',
+      borderWidth: 1,
+      borderRadius: 12,
+    },
```
Saadnajmi, line 13: I still hold that adding overrides here in this platform specific file is mixing info that should be coming from a theme into the base, theme-less component. When I spoke with Tushar offline about this, I think a medium compromise we talked about was that we need more control token mappings in the theme so that we can just set the tokens in the theme rather than override them the `components` section of a theme's definition (which increases bundle size). 

For this specific PR, I'd rather we add more control tokens for 'buttonBorderRadius` or `buttonBorderWidth` in `@fluentui-react-native/theme-types` than introduce them here

packages/components/Button/src/Button.tsx
```diff
@@ -10,29 +10,8 @@ import { backgroundColorTokens, borderTokens, textTokens, foregroundColorTokens,
 import { filterViewProps } from '@fluentui-react-native/adapters';
 import { mergeSettings } from '@uifabricshared/foundation-settings';
 import { useAsPressable, useKeyCallback, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
-import { Icon, RasterImageIconProps, IconProps } from '@fluentui-react-native/icon';
-
-function createIconProps(src: number | string | IconProps) {
-  if (src === undefined) return null;
-
```
Saadnajmi, line 17: Oh I see, this was just moved. Probably good for readability!

packages/components/Button/src/Button.types.ts
```diff
@@ -112,6 +112,7 @@ export interface IButtonProps extends Omit<IPressableProps, 'onPress'> {
 
 export interface IButtonSlotProps {
   root: React.PropsWithRef<IViewProps>;
+  ripple?: PressableProps;
```
Saadnajmi, line 115: Ooh, an optional slot! I wasn't sure we could do that. I've been wondering how to handle platform specific slots...

I'm gonna link this thread again in case it gives insight because we might not need to add a new slot? https://github.com/microsoft/fluentui-react-native/pull/467/files#r496065158

packages/components/Button/src/Button.android.tsx
tamasane: >I'm not sure I see the overflow in that video?

If you look at the corners, you will see some translucent layer flowing out.

>Also, does this comment thread give any insight in general? https://github.com/microsoft/fluentui-react-native/pull/467/files#r499365330

Yep, started this thread in response to the argument that a single wrapper View could be used or not. So I began with just using Pressable as the root view, but noticed that ripple flows outside of the border radius, and the workaround was to wrap Pressable with a View and apply styles to hide it.

packages/components/Button/src/Button.android.tsx
tamasane: Yep, I was in a doubt regarding this, but since android button has an extra slot and thus, changed (and extra) slot settings; it was necessary for readability of code.

tamasane: Will try if I could get `experimental-button` to work in the same PR.

packages/components/Button/src/Button.icon.ts
tamasane: Yep icons are already supported, these are just the imports in the separated common code.

packages/components/Button/src/Button.types.ts
tamasane: Had a look at that PR already! As per the [docs](https://reactnative.dev/docs/touchablenativefeedback), Pressable API should be preferred. Though, this makes android-button compatible with RN 0.63 and higher.

packages/components/Button/src/Button.settings.android.ts
tamasane: >I still hold that adding overrides here in this platform specific file is mixing info that should be coming from a theme into the base, theme-less component.

I am planning to do this in a change to move the current theme to v-next (probably identifying this as the first step for token pipeline integration). The end goal is to fetch the information from the theme as mentioned.

packages/components/Button/src/Button.types.ts
samuelfreiberg, line 115: Is Ripple an android-specific slot? At the very least, I'd suggest adding a comment that tells people this is an android-specific slot 

It feels a bit strange to me that we have two android-specific files (Button.android.tsx and Button.settings.android.ts), but  for this one we don't. I understand it's a small addition, but should it be consistent?

tamasane: `<Pressable />` is used for ripple which is android-specific. Though adding a comment wont hurt 😅, so doing that. 

_Button.android.tsx_ and _Button.settings.android.tsx_ exist because `Platform.select` would affect the file size and readability, whereas having `ripple?` in types file is a very insignificant change to have a separate file.

Saadnajmi: When bundling, metro should be smart enough to strip the `Platform.Select` from the bundles so it would not add to file size unless I'm mistaken

ejlayne: @tamasane I think Sammy's question is about ripple? is fair, and should be a principle that we decide on for the repo. Should we put platform specific props in their own files, or should be put them in a common file and just leave comments. It may be ok with just one platform specific property, but what about two? Or three? Or more? I'm ok with the comment for now, but I could see pulling these out into platform specific files if we start to see more platform-specific types/props.

samuelfreiberg: Yes, that's more where my question was headed (sorry for being vague). This instance should be fine, but once additional platform specific props are needed within the repo, I'd like us to all agree on a organization principle on where to add these. As EJ said stated above

samuelfreiberg (approved):

ejlayne:
@tom-un with this Android Button PR merge, does it make sense to close your other Android button PR? https://github.com/microsoft/fluentui-react-native/pull/467

Issue #642, opened 2021-03-11
Support PlatformColor in our Theme definition


Saadnajmi:
#### Platform
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android
 
#### Describe the feature that you would like added

As it is implemented right now, the `ColorValue` in '@fluentui-react-native/theme-types' resolves to `string`, while the `ColorValue` type in React Native 0.63 resolves to `string | symbol` (because it can also be a PlatformColor, or DynamicColorIOS/MacOS). This causes an error if we try to create a theme using the PlatformColor API. 

I would like to update our theme definition to use the new React Native `ColorValue` type instead of our own, now that we are on 0.63. It's been mentioned before we would like to add support for PlatformColor in our theme and this seems the simplest way.

The only "complicated" bit is we need to update react-native-svg to support PlatformColor, as it's "SvgProps" expect a string instead of a ColorValue. Now that we have a fork of that, it may be easier to do that. 

#### What component or utility would this be added to

@fluentui-react-native/theme-types, and everything that depends on it. 
The Apple theme already has some PlatformColor colors in it's macOS palette (but I'm pretty sure it's a bug that I could do that), but not the iOS Palette. Switching to use DynamicColorIOS for the iOS apple theme would be the one of my goals after we update the theme type.

#### Have you discussed this feature with our team, and if so, who

Yes, @harrieshin @HeyImChris and @amgleitman

Saadnajmi:
Filed issue https://github.com/react-native-svg/react-native-svg/issues/1567 to track changes needed in RN SVG to allow us to use ColorValue

PR #694, opened 2021-05-13
Add support for PlatformColors - change colors to use ColorValue as type

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

There have been several requests for the color tokens to accept `ColorValue`, which is the default color type for RN > 0.63 (ex. #642). This PR changes the colors in the repo to use the `ColorValue` type instead of `string` so our color tokens can accept `PlatformColors`.

The main hangups on this change are:
1. `react-native-svg` and `react-native-picker` are still on RN 0.61, and thus don't support the use of `ColorValue` as the type for their colors. This means we have to pass `string`s for those colors to components that use those packages. The main concern for that within the repo is the use of svg-based icons in `Icon.tsx`. For now I've casted the type to force it to be compatible, but if people have other suggestions, I would certainly welcome them. There is an open PR for supporting `PlatformColor` on the svg repo (react-native-svg/react-native-svg#1561), and I logged an isue for the picker repo (react-native-picker/picker#266). In the meantime SvgIcons will take their color from `svgIconProps` which is fixed to `string` type, with some fallback logic.
2. ~~I have to add a decent number of casts to the code due to the fact that ColorValue is of type `string | unique symbol` and assigning a `ColorValue` to an object property makes the TS compiler complain, as `string | symbol` is not the same as `string | unique symbol`. Casting is the suggested workaround. This issue is tracked at DefinitelyTyped/DefinitelyTyped#46864. I'm concerned that this may introduce a build break for people who depend on us, as I've had to add casting to several places to get things to build.~~ Was able to finally get this working, but we have to use `ColorValue` directly from `react-native` for things to work. Because of this, I'm removing the `ColorValue` export, as I think the user experience of fixing that will be better than fixing the break that would happen if we push this change out without removing `ColorValue`.
3. Made a new `theming-utils` package to house some theme-agnostic logic, had to move  `AppearanceOptions` and `ThemeOptions` under the `theme-types` package to avoid circular dependency.

### Verification

I've run the change through the Fluent UI tester on win32. It seems that the win32 app freezes if you change the Theme or the Brand of the app (this repros without changes, and I'll look into the issue), but toggling between light/dark modes works and colors update accordingly.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Checkbox/CheckboxTest.tsx
```diff
@@ -103,7 +103,7 @@ const tokenCheckbox: React.FunctionComponent<{}> = () => {
 
   const theme = useTheme();
   const textBoxBorderStyle = {
-    borderColor: theme.colors.inputBorder,
+    borderColor: theme.colors.inputBorder as ColorValue,
```
alloy, line 106: What actually is the TS type and the runtime type of `inputBorder`?

rurikoaraki: It's ColorValue (which with this PR, will be the same type as ColorValue in React Native). You can find it defined here: https://github.com/microsoft/fluentui-react-native/blob/3c0f5496f0daeb748c6b2daac8d0465a1d58a672/packages/theming/theme-types/src/Color.types.ts#L249

At runtime, at the moment on all platforms except MacOS this is a string. On MacOS this will be a PlatformColor.

FalseLobster (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Checkbox/CheckboxTest.tsx
rurikoaraki: I removed this cast and cast the entire object into a TextStyle isntead.

packages/theming/theme-types/package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {},
   "devDependencies": {
     "@types/react": "^16.9.34",
+    "@types/react-native": "^0.63.4",
     "react": "16.13.1",
     "react-native": "^0.63.4",
```
Saadnajmi, line 39: nit: I feel like this is just a dependency now that we use ColorValue specifically..

packages/theming/theme-types/src/Color.types.ts
```diff
@@ -11,7 +11,10 @@
  *
  * `{semantic: "windowBackgroundColor"}`
  */
-export type ColorValue = string;
+
+import { ColorValue } from 'react-native';
+
+export { ColorValue } from 'react-native';
```
Saadnajmi, line 17: So that's how you do that :D 
Repeating again that we may just want to get rid of the type since it's redundant.

packages/utils/tokens/src/border-tokens.ts
```diff
@@ -1,11 +1,11 @@
 import { ViewStyle } from 'react-native';
 import { IOperationSet } from '@uifabricshared/foundation-tokens';
-import { ITheme } from '@uifabricshared/theming-ramp';
+import { ColorValue, ITheme } from '@uifabricshared/theming-ramp';
```
Saadnajmi, line 3: nit: Import from `@fluentui-react-native/theme-types` or just `react-native`

packages/utils/tokens/src/color-tokens.ts
```diff
@@ -1,6 +1,5 @@
-import { ITheme, IThemeColorDefinition } from '@uifabricshared/theming-ramp';
+import { ColorValue, ITheme, IThemeColorDefinition } from '@uifabricshared/theming-ramp';
```
Saadnajmi, line 1: same comment about importing here

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -4,100 +4,101 @@ import { ButtonTokens } from '.';
...
+    },
     disabled: {
-      backgroundColor: t.colors.brandDisabledBackground,
-      color: t.colors.brandDisabledContent,
-      borderColor: t.colors.brandDisabledBorder,
-      iconColor: t.colors.brandDisabledIcon,
+      backgroundColor: t.colors.buttonDisabledBackground,
+      color: t.colors.buttonDisabledContent,
+      borderColor: t.colors.buttonDisabledBorder,
+      iconColor: t.colors.buttonDisabledIcon,
```
Saadnajmi, line 24: These are a change of tokens across a bunch of styling. Was that the intention?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.ios.tsx
```diff
@@ -5,29 +5,14 @@ import { MenuAction, MenuView } from '@react-native-menu/menu';
...
       flexDirection: 'row',
     },
-    picker: {
-      flexDirection: 'column',
-      alignItems: 'center',
-      padding: 4,
-      flexShrink: 0,
-    },
-    pickerItem: {
-      color: t.colors.bodyText,
```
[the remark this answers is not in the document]
rurikoaraki: Oh okay, I'll put it back. I didn't see it being referenced directly, so I was confused why this was here.

packages/components/Button/package.json
```diff
@@ -27,12 +27,12 @@
     "@fluentui-react-native/interactive-hooks": ">=0.9.0 <1.0.0",
     "@fluentui-react-native/pressable": ">=0.6.0 <1.0.0",
     "@fluentui-react-native/text": ">=0.9.1 <1.0.0",
+    "@fluentui-react-native/theme-types": "^0.6.0",
     "@fluentui-react-native/tokens": ">=0.8.0 <1.0.0",
     "@uifabricshared/foundation-composable": ">=0.8.0 <1.0.0",
     "@uifabricshared/foundation-settings": ">=0.8.0 <1.0.0"
   },
   "devDependencies": {
-    "@office-iss/react-native-win32": "^0.63.7",
```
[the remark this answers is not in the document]
rurikoaraki: I need to look at this again because I might have been overzealous in removing unused dependencies.

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -54,10 +54,11 @@ function renderSvg(iconProps: IconProps) {
   const { width, height } = iconProps;
   const viewBox = iconProps.svgSource.viewBox;
 
+  // react-native-svg is still on 0.61, and their color prop doesn't handle ColorValue
```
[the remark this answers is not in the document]
rurikoaraki: This basically casts to string right now, so it might still be a blocker. Another option is to only allow strings for icon color tokens, but it might be better to just wait for this to come along for now, since I have other blockers.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.ios.tsx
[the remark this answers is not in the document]
Saadnajmi: I should perhaps add a comment in code why it exists.

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -2,44 +2,50 @@ import { buttonName, ButtonTokens, ButtonSlotProps, ButtonProps } from './Button
 import { Theme, UseStylingOptions, buildProps } from '@fluentui-react-native/framework';
 import { borderStyles, layoutStyles, fontStyles } from '@fluentui-react-native/tokens';
 import { buttonStates, defaultButtonTokens } from './ButtonTokens';
+import { TextProps } from '@fluentui-react-native/experimental-text';
+import { IViewWin32Props } from '@office-iss/react-native-win32';
```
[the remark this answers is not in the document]
rurikoaraki: Using IViewWin32Props is driven by the root style for this button being defined as using IViewWin32Props in Button.types.ts: https://github.com/microsoft/fluentui-react-native/blob/3c0f5496f0daeb748c6b2daac8d0465a1d58a672/packages/experimental/Button/src/Button.types.ts#L118

I think using IViewProps makes sense, but I think it's out of scope for this change since it's a change to Button and not for supporting ColorValue..

packages/experimental/Button/src/ButtonTokens.ts
rurikoaraki: I uh... am not sure why git thinks there's a diff, since the actual text looks the same? The change is just that I had to cast this object as ButtonTokens to avoid a compiler error. I'll double check that nothing has changed.

packages/experimental/NativeButton/src/NativeButton.types.ts
```diff
@@ -1,3 +1,4 @@
+import { ColorValue } from '@fluentui-react-native/theme-types';
```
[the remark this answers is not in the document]
rurikoaraki: My only concern with doing that is we do export ColorValue so removing it might break consumers of the package. If we were to remove the ColorValue type from the package it'd likely be thought of as a breaking change...

packages/framework/theme-registry/src/Registry.test.ts
```diff
@@ -37,105 +38,105 @@ interface IFakeTheme {
...
+  },
 };
 
 function processor(parent: IFakeTheme): Partial<IFakeTheme> {
   return {
     settings: {
       base: {
         root: {
-          textColor: 'light' + (parent.settings.base.root as IFakeStyle).textColor,
+          textColor: 'light' + (parent.settings.base.root as IFakeStyle).textColor.toString(),
```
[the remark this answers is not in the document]
rurikoaraki: Fair point. This was test code, so I felt a little better about doing this, but I'll look into whether I can do this in a better way.

packages/utils/tokens/src/color-tokens.ts
rurikoaraki: Latest commit addresses this.

packages/utils/tokens/src/border-tokens.ts
rurikoaraki: Latest commit addresses this.

packages/theming/theme-types/package.json
rurikoaraki: Latest commit addresses this.

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -1,5 +1,6 @@
 import { SvgProps, ImageProps } from 'react-native-svg';
 import { ImageURISource } from 'react-native';
+import { ColorValue } from '@uifabricshared/theming-ramp';
```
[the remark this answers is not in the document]
rurikoaraki: Latest commit addresses this.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.ios.tsx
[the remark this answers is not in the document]
rurikoaraki: Latest commit addresses this.

packages/theming/theme-types/src/Color.types.ts
rurikoaraki: I responded to the other comment on why I'm hesitant to just remove this. Feel free to reactivate if you feel differently.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.android.tsx
```diff
@@ -49,13 +51,18 @@ export const ThemePickers: React.FunctionComponent<{}> = () => {
 
   const theme = useTheme();
   const themedPickerStyles = getThemedDropdownStyles(theme);
+
+  type DropdownEntry = { label: string; value: string };
+
+  // react-native-picker is still on 0.61, and their color prop doesn't handle ColorValue
+  const dropdownIconColor = theme.colors.buttonIcon as PickerProps<DropdownEntry>['dropdownIconColor'];
+
```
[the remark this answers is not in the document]
rurikoaraki: Opened react-native-picker/picker#266

packages/experimental/Button/src/ButtonTokens.ts
Saadnajmi: Oh.. they are the same. I think it got confused with whitespace lol. Probably fine!

Saadnajmi: I think the whitespace changes that show up in every review is inconsistent based on whether the dev in question had prettier or eslint installed.. It might be worth doing a full pass of every file somehow in 1 separate PR so that individual PRs are cleaner

packages/components/Button/package.json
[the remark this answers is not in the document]
rurikoaraki: Reverted this change, since I'm not sure what this was used for and it's not really required for this change.

apps/fluent-tester/src/FluentTester/TestComponents/Checkbox/CheckboxTest.tsx
rurikoaraki: Turns out it helps to do full builds to test changes. It seems that with full builds and being better about declaring types for objects that we create. we're able to avoid casting, so this is not an issue anymore.

rurikoaraki: Well I spoke too soon. I can remove casts, but I still need to have the change done in DT in order to get this to compile fully. I have a PR open here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53368

rurikoaraki: Okay it builds! I just had to use ColorValue from react-native directly and update all of the dependencies accordingly, instead of using ColorValue exported from theme-types (even though it's just the react-native one underneath)

packages/experimental/Icon/src/Icon.tsx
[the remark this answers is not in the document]
Saadnajmi: Revisiting this: I think a cast will crash and we should separate out the icon tokens and allow them to only be a number or string. As it is implemented rn, I think the icon will inherit the text color token if no icon color is supplied

packages/experimental/NativeButton/src/NativeButton.types.ts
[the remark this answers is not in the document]
Saadnajmi: Looks like we removed it :D

packages/experimental/Avatar/src/Avatar.tsx
```diff
@@ -67,7 +67,7 @@ export type otherAvatarProps = {
   /**
    * Background Color of initials
    */
-  backgroundColor?: string;
+  backgroundColor?: ColorValue;
```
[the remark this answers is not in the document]
Saadnajmi: I think this should be fine, since MeControl on iOS will only ever pass it a string

Saadnajmi (approved):

packages/experimental/Icon/src/Icon.tsx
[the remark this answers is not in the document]
rurikoaraki: Would it be an issue if I added a color prop to SvgIconProps which was string only but left the color prop alone so that font icons could take advantage of the ColorValue change? I guess it could break things downstream so I guess that's not the best solution.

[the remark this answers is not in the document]
rurikoaraki: Addressed with latest commit

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -56,16 +56,20 @@ function renderSvg(iconProps: IconProps) {
   const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width: width, height: height }, [width, height])[0]);
 
   // react-native-svg is still on 0.61, and their color prop doesn't handle ColorValue
+  // If a color for the icon is not supplied, fall back to white or black depending on appearance
+  const theme = useTheme();
+  const iconColor = svgIconProps.color ? svgIconProps.color : theme.host.appearance === 'dark' ? '#FFFFFF' : '#000000';
```
Saadnajmi, line 61: A nit: We could just have the default value of the color prop be set in Icon itself, and hide the need to set it ourself here.

Saadnajmi (approved):

packages/experimental/Icon/src/Icon.tsx
Saadnajmi: I'm dumb and didn't realize this is Icon.tsx. ignore me!

Saadnajmi, line 61: If the theme doesn't set it's own appearance or host, do we need to default to use the react-native appearance API?

rurikoaraki: Thanks for the suggestion, I added that functionality. There's only so much we can do to try and support a fallback - worst case you'll get black, which is what you'd get if you didn't supply a color anyway.

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -6,6 +6,7 @@ import { SvgUri } from 'react-native-svg';
 import { mergeStyles } from '@fluentui-react-native/framework';
 import { stagedComponent, mergeProps, getMemoCache } from '@fluentui-react-native/framework';
 import { useTheme } from '@fluentui-react-native/theme-types';
+import { getCurrentAppearance } from '@fluentui-react-native/default-theme';
```
Saadnajmi, line 10: This adds a direct dependency to default-theme for icon. I feel components shouldn't know about any specific theme?

packages/experimental/Icon/src/Icon.tsx
Saadnajmi: That's fair. Going through one hoop of checking the theme's appearance is probably enough. Thanks for looking into that!

packages/experimental/Icon/src/Icon.tsx
Saadnajmi: or this could be a function that is moved to the theme definition so every theme can use it? That maybe out of scope though

rurikoaraki: Yeah I was being lazy. I'll move it to a theming folder under utils

packages/theming/default-theme/src/createDefaultTheme.ts
```diff
@@ -1,34 +1,8 @@
 import { ThemeReference } from '@fluentui-react-native/theme';
 import { defaultFluentDarkTheme, defaultFluentTheme } from './defaultTheme';
 import { Appearance } from 'react-native';
-import { Theme } from '@fluentui-react-native/theme-types';
-
-export type AppearanceOptions = 'light' | 'dark';
```
rurikoaraki, line 6: Not sure if moving these is a good idea. I can keep them here, but then moving getCurrentAppearance out of the default-theme package feels weird since it feels like a circular dependency...

packages/experimental/Icon/src/Icon.tsx
Saadnajmi: Ah, I didn't realize it would require creating a new package 😅, that does feel heavy for such a simple change.
IIRC, the purpose of the default theme having getCurrentAppearance() was because the default theme could override the appearance instead of what the system defined. 

I feel moving that functionality is still beneficial and better than having a component dependent on a specific theme. Could it not go in the `@fluentui-react-native/theme` package under `packages/experimental/theme`? But this is beginning to feel out of scope so I think I'll be fine with whatever you feel works best!

rurikoaraki: Thanks for the suggestion! Taking a look, the `experimental/theme` feels "downstream" of `default-theme` (if that makes sense), and `default-theme` still uses `getCurrentAppearance()`, so I think keeping it close to that makes more sense compared to moving it under `experimental/theme`. I'm okay with having a utils folder under theming, might come in handy in the future.

PR #695, opened 2021-05-18
Add react-native-svg-transformer to iOS and macOS

### LINKED ISSUE #644: Icon example page should include direct SVG imports on macOS and iOS
#### Platform
- [x] iOS
- [x] macOS
- [ ] win32
- [ ] windows
- [ ] android
 
#### Describe the feature that you would like added

In Fluent UI React Native, Icons can be defined using SVGs directly imported from the source tree. However, the automated checks seem to have a problem with the tool that allows us to import SVGs directly. While this doesn't affect what Icon is capable of, we really should fill this feature gap.

amgleitman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This removes platform-divergent behaviors on some of the pages in the FURN test app. Specifically, we do the following:
* Add react-native-svg-transformer to the iOS and macOS test apps, thereby allowing us to resolve #644.
* Remove the font-based icon from the Button test page. The font icon difference will stay on the Icon test page as we determine the best way to link Font Awesome's .otf into the app.

### Verification

Validated that the Button and Icon test pages still work fine on iOS and macOS.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonIconTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonIconTest.tsx
@@ -2,9 +2,9 @@
 import { Button } from '@fluentui/react-native';
 import { Stack } from '@fluentui-react-native/stack';
 import * as React from 'react';
-import { Platform, View } from 'react-native';
+import { View } from 'react-native';
 import { stackStyle } from '../Common/styles';
-import { SvgIconProps, FontIconProps } from '@fluentui-react-native/icon';
+import { SvgIconProps } from '@fluentui-react-native/icon';
 import TestSvg from './test.svg';
 
 const CustomizedIconButton = Button.customize({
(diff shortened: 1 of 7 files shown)


Saadnajmi:
My guess for why this fails windows is because is because we have an SVG icon on the button test page, and RN-SVG doesn't work with react-native-windows (but it does for win32)

Saadnajmi:
It also looks like this removes the font button from being tested on win32? That seems like removing a thing we would want?

amgleitman:
> It also looks like this removes the font button from being tested on win32? That seems like removing a thing we would want?

We're mainly trying to showcase the ability to add an icon to a button. With the Icon test page being a thing, I think the font button might be sort of redundant.

Also, there's still the issue of figuring out the best way to link custom fonts in iOS and macOS, so this is just an easy way of staying unblocked.

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonIconTest.tsx
```diff
@@ -20,20 +20,27 @@ export const ButtonIconTest: React.FunctionComponent<{}> = () => {
     viewBox: '0 0 500 500',
   };
 
+  // SVG-based icons are not available on all platforms yet
+  const svgIconsEnabled = ['ios', 'macos', 'win32'].includes(Platform.OS as string);
```
Saadnajmi, line 24: Maybe a place for @tamasane or @mnkshukl to follow up if / when Android works

Saadnajmi: Also, I think SVG's work with react-native-web as well? Probably not blocking to include that since nothing really depends on the web platform...

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonIconTest.tsx
```diff
@@ -13,48 +13,35 @@ const CustomizedIconButton = Button.customize({
...
-      viewBox: '0 0 500 500',
-    };
+  // SVG-based icons are not available on all platforms yet
+  const svgIconsEnabled = ['ios', 'macos', 'win32'].includes(Platform.OS as string);
 
-    return (
-      <View>
-        <Stack style={stackStyle}>
-          <Button icon={testImage} content="Button with png Icon" tooltip="button tooltip" />
-          <Button icon={{ fontSource: fontProps, color: 'blue' }} content="Button with font Icon" tooltip="button tooltip" />
```
Saadnajmi, line 36: Since it's still a change to the win32 test app to remove this, I'd want someone from CXE to also look at this, so I'm just gonna tag @warren-ms . Things look good to me for the Apple side

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonIconTest.tsx
amgleitman: Yep, and once new platforms get verified we can add them here.

warren-ms (approved):

PR #699, opened 2021-05-21
Better-Aligning RadioGroup Text with RadioButton

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes

The text of the RadioButton was a bit lower than expected. This is because, even when invisible, the border width (2px) is being taken into account when aligning. This small change better centers the text with the RadioButton. Tested at all DPI's.

### Verification

Visually, tested at all DPIs.

| 
image
 | 
image
 |
image

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

ejlayne:
@samuelfreiberg can you please also add a picture of the focused state to this PR (even just the after), since that is what started this conversation?

samuelfreiberg:
Yup, just updated the PR to add a picture of the focused state!

packages/components/RadioGroup/src/RadioButton.settings.ts
```diff
@@ -60,7 +60,7 @@ export const settings: IComposeSettings<IRadioButtonType> = [
     content: {
       variant: 'subheaderStandard',
       style: {
-        marginTop: 3,
+        marginTop: 2,
```
Saadnajmi, line 63: why isn't this just 0? If we're centering it, we wouldn't need an offset?

ejlayne: I think when I played with this in NetUI spy, the "O" of "Option" looked slightly north of center when the top margin is 0, so it looked off-center the other way. But Sammy can post a picture here and we can take a look.

samuelfreiberg: I've tested it with all values/DPIs, and this seemed to work best. I believe '0' doesn't work bet because the variant of the font + the size of it's border. Although the border is invisible, it still exists and the size is taken into account when centering

PR #700, opened 2021-05-21
Add prop hideInsideGapForBorder to experimental-avatar

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds a prop to hide the gap between the Avatar and the border. I'm only doing this change for iOS for now as the macOS native avatar border API is a bit different, and would require more work than this simple addition to get working.

### Verification

Added a test case to the iOS test app.

Screen Shot 2021-05-21 at 5 47 18 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec
+++ b/packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = "13.0"
   s.ios.source_files      = "ios/*.{swift,h,m}"
-  s.ios.dependency 'MicrosoftFluentUI', '~> 0.2.2'
+  s.ios.dependency 'MicrosoftFluentUI', '~> 0.2.6'
 
   s.osx.deployment_target = "10.14"
   s.osx.source_files      = "macos/*.{swift,h,m}"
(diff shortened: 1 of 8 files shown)


packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec
```diff
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = "13.0"
   s.ios.source_files      = "ios/*.{swift,h,m}"
-  s.ios.dependency 'MicrosoftFluentUI', '~> 0.2.2'
+  s.ios.dependency 'MicrosoftFluentUI', '~> 0.2.6'
```
harrieshin, line 19: why is this 0.2.6 while others are 0.2.7 in pod.lock file?

harrieshin (approved):

packages/experimental/Avatar/FluentUIReactNativeAvatar.podspec
Saadnajmi: `~> 0.2.6` resolves to the latest version of `0.2.x`, which at the time of running `pod install` was 0.2.7

PR #703, opened 2021-05-25
Bump the anrdroid theme version manually

HeyImChris:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [X] android

### Description of changes

Publish CI is failing with the following error:
`ERROR: Attempting to bump to a version that already exists in the registry: @fluentui-react-native/android-theme@0.2.6`

This package version is supposed to be updated automatically, but something broke and it's out of sync with NPM. After checking on npmjs.com, the latest version is 0.2.7, so let's update to 0.2.8.

### Verification

This can be verified in the publish CI which is currently failing for everyone.

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/android-theme/package.json
+++ b/packages/theming/android-theme/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui-react-native/android-theme",
-  "version": "0.2.7",
+  "version": "0.2.8",
   "description": "A FluentUI React Native theme that pulls constants from FluentUI Android",
   "repository": {
     "type": "git",


harrieshin (approved):

PPatBoyd (approved):

packages/theming/android-theme/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui-react-native/android-theme",
-  "version": "0.2.7",
+  "version": "0.2.8",
```
Saadnajmi, line 3: What you actually want is a changefile, not a manual bump. You can do this by making an edit, running `yarn change` and then reverting your edit. 

But also we're unblocked for now, so let's save this

Saadnajmi (changes requested):

packages/theming/android-theme/package.json
Saadnajmi: Although maybe just manually bumping the # here is the root cause and we don't want a change file. 🤔.
Maybe this # was always one below what it needed to be?

HeyImChris:
Looks like another review is handling this so will close for now.

PR #705, opened 2021-05-26
Fix for FURN Controls not getting keyboard focus on Click 

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Enables Button, Checkbox, Experimental Button, Link and Radio Button  to receive keyboard focus after being clicked. 

### Verification

Tested locally in the FluentUI Tester

| GIF coming soon. |  |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/Button.tsx
```diff
@@ -20,12 +26,16 @@ export const Button = compose<IButtonType>({
       content,
       onAccessibilityTap = userProps.onClick,
       accessibilityLabel = userProps.content,
+      componentRef = userProps.componentRef != undefined ? userProps.componentRef : React.useRef(null),
```
lenahong, line 29: I think just `componentRef = React.useRef(null)` should do what you want!

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -41,10 +41,13 @@ export const RadioButton = compose<IRadioButtonType>({
...
     }, []);
 
+    // Ensure focus is placed on button after click
+    const changeSelectionWithFocus = useOnPressWithFocus(componentRef, changeSelection);
+
     /* RadioButton changes selection when focus is moved between each RadioButton and on a click */
     const pressable = useAsPressable({
       ...rest,
-      onPress: changeSelection,
+      onPress: changeSelectionWithFocus,
```
samuelfreiberg, line 50: I would double-check that the RadioButton gets focus when navigating with keyboard as well. When arrowing up/down in a RadioGroup, selection+focus should change

packages/components/Checkbox/src/Checkbox.tsx
```diff
@@ -9,14 +9,30 @@ import { filterViewProps } from '@fluentui-react-native/adapters';
...
-    const { ariaLabel, checked, defaultChecked, boxSide, disabled, label, onChange, ...rest } = userProps;
+    const {
+      ariaLabel,
+      checked,
+      defaultChecked,
+      boxSide,
+      disabled,
+      label,
+      onChange,
+      componentRef = userProps.componentRef != undefined ? userProps.componentRef : React.useRef(null),
```
lenahong, line 33: `componentRef = React.useRef(null),`

packages/components/Button/src/Button.tsx
samuelfreiberg, line 29: I believe this can be written as: 

`componentRef = userProps.componentRef ?? React.useRef(null)`

See [Nullish Operator](https://www.javascripttutorial.net/es-next/javascript-nullish-coalescing-operator/#:~:text=ES2020%20introduced%20the%20nullish%20coalescing,double%20question%20marks%20(%20%3F%3F%20).&text=It%20returns%20the%20right%20operand,is%20either%20null%20or%20undefined%20.)

samuelfreiberg: This is same for all the usages below

lenahong (approved):

packages/components/Button/src/Button.tsx
ksiler: What if the user provides their own ref?

packages/components/RadioGroup/src/RadioButton.tsx
ksiler: It does

packages/components/Button/src/Button.tsx
ksiler: Oh nice! Will definitely use that.

CampbellOwen: I'm pretty sure the assignment operator here is for providing default values during the pattern matching - so if `userProps.componentRef` is defined it won't assign it to `React.useRef(null)`

ksiler: I see, that seems to be the case. Thanks for the explanation! I'll update it in my next iteration.

rurikoaraki (approved):

NikoAri (approved):
:shipit:

Saadnajmi:
@chiuam as an FYI since she's looking into focus stuff on the button for macOS soon

PR #706, opened 2021-05-26
Comment out tests that crash the app for iOS and macOS

lindaweng:
### Platforms Impacted
- [ x] iOS
- [ x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Commented out the tests that crash the tester app for iOS and macOS. Tests commented out are Callout, ContextualMenu, and Focus Trap Zone.

### Verification

Tested by running both tester apps and confirming that the tests no longer showed up

|Simulator Screen Shot - iPhone 11 - 2021-05-26 at 16 56 52|Simulator Screen Shot - iPhone 11 - 2021-05-26 at 16 56 32 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/testPages.ios.ts
+++ b/apps/fluent-tester/src/FluentTester/testPages.ios.ts
@@ -1,11 +1,8 @@
 import { TestDescription } from './TestComponents';
 import { AvatarTest, HOMEPAGE_AVATAR_BUTTON } from './TestComponents/Avatar';
 import { ButtonTest, HOMEPAGE_BUTTON_BUTTON } from './TestComponents/Button';
-import { CalloutTest, HOMEPAGE_CALLOUT_BUTTON } from './TestComponents/Callout';
 import { CheckboxTest, HOMEPAGE_CHECKBOX_BUTTON } from './TestComponents/Checkbox';
-import { ContextualMenuTest, HOMEPAGE_CONTEXTUALMENU_BUTTON } from './TestComponents/ContextualMenu';
 import { ExperimentalButtonTest, HOMEPAGE_BUTTON_BUTTONEXPERIMENTAL } from './TestComponents/ButtonExperimental';
-import { FocusTrapTest, HOMEPAGE_FOCUSTRAPZONE_BUTTON } from './TestComponents/FocusTrapZone';
 import { HOMEPAGE_ICON_BUTTON, IconTest } from './TestComponents/Icon';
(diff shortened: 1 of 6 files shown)


apps/fluent-tester/src/FluentTester/testPages.ios.ts
```diff
@@ -35,26 +35,26 @@ export const tests: TestDescription[] = [
...
+  // {
+  //   name: 'Callout Test',
+  //   component: CalloutTest,
+  //   testPage: HOMEPAGE_CALLOUT_BUTTON,
+  // },
+  // {
+  //   name: 'ContextualMenu Test',
+  //   component: ContextualMenuTest,
+  //   testPage: HOMEPAGE_CONTEXTUALMENU_BUTTON,
+  // },
```
Saadnajmi, line 47: We have a preference to not have commented out code in public repos, so it would be preferred to just delete the code altogether. There should be a list of all the tests in just `testPages.ts` that is easy enough to reference to remember what tests exists.

Saadnajmi (changes requested):

apps/fluent-tester/src/FluentTester/testPages.ios.ts
harrieshin: similar to macOS test app should we do disable tableviewcell look? or should we really delete?

Saadnajmi: A disabled button for tests that don't work on a platform would be a good idea, but a more complex change. I think that could be a good followup.
I think right now between the components / experimental components, there isn't really one platform that actually has every test so it's fine to just delete them for now.

lindaweng: Should I also delete the import statements as well since they aren't being used? I'm not sure what the tableviewcell look is.

Saadnajmi: If you don't delete import statements, I think you'll get an `unused import` error in the PR checks (probably what's failing under PR (Build, test, and bundle Javascript) right now)., so definitely remove them.

Sure, I think it's no harm to add those two tests to `testPages.ts`. At some point I do think we should consolidate to just have one `testPages` file, but I still feel that's outside the scope of this simple change.

Saadnajmi (approved):

Saadnajmi:
It looks like you need to run `yarn change` at the repo root and update this PR before it can be merged.

PR #709, opened 2021-05-28
Initial theme tokens package and global token file

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Adding new theme-tokens package which contains the initial global tokens files. Integrated tokens files into default theme to have an example of how to use the tokens in the project.

### Verification

Booted tester app, made sure that the theme table updated based on which global token was passed in. Screenshots will be posted shortly.

| theming_before | theming_after <br> (redDark is set to teal primary token in this example) |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -1,21 +1,22 @@
...
-      subheader: 16 as FontSize,
-      header: 20 as FontSize,
-      hero: 28 as FontSize,
+      caption: globalTokens.font.size['100'] as FontSize,
+      secondary: globalTokens.font.size['200'] as FontSize,
+      body: globalTokens.font.size['300'] as FontSize,
+      subheader: globalTokens.font.size['400'] as FontSize,
+      header: globalTokens.font.size['500'] as FontSize,
+      hero: globalTokens.font.size['700'] as FontSize,
       heroLarge: 42 as FontSize,
```
Saadnajmi, line 15: heroLarge isn't in the tokens?

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -1,21 +1,22 @@
...
     } as FontSizes,
     weights: {
-      regular: '400' as FontWeightValue,
-      semiBold: '600' as FontWeightValue,
+      regular: globalTokens.font.weight.regular.toString() as FontWeightValue,
+      semiBold: globalTokens.font.weight.semibold.toString() as FontWeightValue,
```
Saadnajmi, line 19: why do we need toString?

Saadnajmi (approved):

packages/theming/default-theme/src/defaultTheme.ts
rurikoaraki: FontWeightValue is string type, and the font weights in the tokens file are numbers.

@TravisSpomer maybe we should change the output of the font weight to strings to have consistency, to get rid of the cast here.

packages/theming/default-theme/src/defaultTheme.ts
rurikoaraki: Yeah, 42 is not in the current token file. The `defaultTheme` has the older theming tokens, so not everything maps. I'm thinking in a future commit I'll add a vNext version which has the proper mappings that are listed in the newer designs?

Saadnajmi: Yeah I remember the typography ramp was off, and I just kinda winged the mappings for the apple theme. Makes sense!

packages/theming/default-theme/src/defaultTheme.ts
TravisSpomer: Sure; we can change it to whatever's convenient for the platform. The outputs of the "json" folder are just things I threw together as an example; I could make a separate set of React Native outputs that use strings for font weights, and any other customizations you need.

(Sorry for the delay; I don't seem to get email notifications for @mentions to things under the `microsoft` scope. 😒)

rurikoaraki: Thanks! That would be helpful. I'm working on the change to integrate the pipeline into this repo, so if I can just change the input file to output the right thing, I can do that on my end, too.

PR #710, opened 2021-05-29
Update Native Apple Code for better downstream integration

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change refactors some of our native iOS and macOS code for better downstream integration into the office codebase.

Namely:
- Every native module's `<Component>-Bridging-Header.h` has been renamed to `<Component>ViewManager.h`. 
  - These files were never bridging headers, and cocoa pods / swift never interpreted them as such. There is potential to improve the RN docs on how to write a native swift module here as well.
- Adds the `BUILD_LIBRARY_FOR_DISTRIBUTION` flag to each native module's pod spec
  - This flag allows for ABI stability in our swift static libraries. Note that the main side effect of this flag is it produces a `.swiftinterface` file in the static libraries' swift module. I do not include the swift module in the NuGet to be included downstream in office. In my testing I found it was not needed to get iOS NativeButton to link, so I opted not to include the swif module. This may change in the future as I try to figure out why the macOS NativeButton static library won't link.
- Adds the `OTHER_SWIFT_FLAGS=-gline-tables-only` flag to each native module's pod spec
  - This is a potential security risk that is fixed, as per the comment here https://github.com/microsoft/fluentui-react-native/pull/710#discussion_r688016727 . 
- removes the `CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only` build flag
  - This caused issues with the previous `BUILD_LIBRARY_FOR_DISTRIBUTION` flag and build failures. I'm not sure why. 
- Sets the Swift version to 5.0
  - I added this because I noticed that Azure pipelines would actually use the Swift 5.3.2 compiler, so I wanted to be explicit which version of swift we are compiling for.

Prior to this change, we could not link the NativeButton component downstream into office due to some linker issues. This change updates a few parts of our native iOS and macOS code such that we can now link the iOS NativeButton. The macOS NativeButton static library still does not properly link, I'll leave that as a followup.

### Verification

I had a local change downstream in office to consume an intermediate NuGet produced by this PR and test that linking worked. That's where I saw that iOS and not macOS works. Link for those who can see it: https://office.visualstudio.com/OE/_build/results?buildId=9621115&view=logs&j=4257c0a7-7659-5375-d6e0-cde9f28a3d36

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/xcconfig/publish_overrides.xcconfig
```diff
@@ -1,2 +1,13 @@
+// For publish builds, only provide line tables for symbolizing crashes
+CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only
+// OTHER_SWIFT_FLAGS=-gline-tables-only
```
Saadnajmi, line 3: @markavitale do you know what the impact of not including this build flag may be? I copied these flags from FluentUI Apple, and I suppose we were never building with it to begin with so it should be alright. I was confused why this caused build failures earlier though.

Saadnajmi: Does the xcconfig on line 2 just do a separate but very similar thing?
`CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only`

It also sounds like it's a limiter so we are just getting more info than we need?

markavitale: Yep, one is for clang compiler (Objective-C, C, C++, etc), and the other is for the Swift compiler (Swift code). We probably can't ship something with full debug information (which includes things like filepaths of the machines that did the compilation), so the alternative would be dropping the ability to symbolize crashes involving FURN.

.ado/xcconfig/publish_overrides.xcconfig
```diff
@@ -1,2 +1,13 @@
+// For publish builds, only provide line tables for symbolizing crashes
+CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only
+// The following build setting caused build errors, so it is commented out
```
HeyImChris, line 3: Probably worth explaining the actual build error and/or just removing the commented out line altogether? Do we expect to need to add it back at some point?

package.nuspec
```diff
@@ -9,18 +9,20 @@
...
     <!-- iOS device ship -->
     <file src="DerivedData\Build\Products\Release-iphoneos\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Ship-iphoneos"/>
     <file src="DerivedData\Build\Products\Release-iphoneos\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Ship-iphoneos"/>
     <file src="DerivedData\Build\Products\Release-iphoneos\FluentUI-React-Native-Date-Picker\libFluentUI-React-Native-Date-Picker.a" target="Ship-iphoneos"/>
+
     <!-- iOS simulator debug -->
-    <file src="DerivedData\Build\Products\Debug-iphonesimulator\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Debug-iphonesimulator"/>
-    <file src="DerivedData\Build\Products\Debug-iphonesimulator\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Debug-iphonesimulator"/>
-    <file src="DerivedData\Build\Products\Debug-iphonesimulator\FluentUI-React-Native-Date-Picker\libFluentUI-React-Native-Date-Picker.a" target="Debug-iphonesimulator"/>
+    <file src="DerivedData\Build\Products\Debug-iphoneos\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Debug-iphoneos"/>
```
HeyImChris, line 22: Why do we switch to packaging the iphoneos libs instead of the simulator ones?

package.nuspec
```diff
@@ -32,18 +34,13 @@
...
     <!-- macOS ship -->
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-FocusZone.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-MenuButton\libFluentUI-React-Native-MenuButton.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Radio-Button\libFluentUI-React-Native-Radio-Button.a" target="Ship-macosx"/>
 
-
-    <!-- iOS Release jsbundle -->
-    <file src="apps\ios\dist\index.ios.jsbundle" />
```
HeyImChris, line 44: We don't need these for downstream consumption? Is it because clients SDXs are defining their own index files?

package.nuspec
Saadnajmi: That is a typo :')

package.nuspec
Saadnajmi: Nope. these were unused so I just opted to remove them

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: Re Marks' comment here https://github.com/microsoft/fluentui-react-native/pull/710#discussion_r688016727

I wanted to ask what the fallout is by not using it. Looks like it could be a risk so I'm doing some additional testing to make sure we are including the right amount of symbols.

.ado/xcconfig/publish_overrides.xcconfig
```diff
@@ -1,2 +1,13 @@
...
+// OTHER_SWIFT_FLAGS=-gline-tables-only
+
+// Optimize for size in publish builds
+SWIFT_OPTIMIZATION_LEVEL=-Osize
+
 // Build for all architectures, not just the active one
 ONLY_ACTIVE_ARCH=NO
+
+// Specify the exact Swift version used for reproducibility
+SWIFT_VERSION = 5.0
```
chripear, line 13: nit: add EOL (newline) at EOF

chripear (approved):

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: I'll continue the thread in the other comment thread

.ado/xcconfig/publish_overrides.xcconfig
Saadnajmi: Just gonna drop this here for reference. https://gist.github.com/CodaFi/b1e0d0f37cd2890c07c5

Saadnajmi: To be overcautious, I added `-gline-tables-only` to every pod spec to be over-cautious. Looking at a local build log, it also seemed that the greater project `ReactTestApp` was also inheriting that arg from elsewhere. Between those two, I feel this is safer now. I could probably revisit how to inherit xcconfig args more cleanly here, but 🤷🏾

PR #714, opened 2021-06-02
Fix comment in android-theme to trigger package update

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The NPM publish pipeline still appears to be stuck on updating the android-theme package. I'm not sure what's going on there but would like to unblock other packages that are waiting to be updated. I've fixed some comments in android-theme to initiate a change in the package. Hopefully, this will unblock the other package updates.

### Verification

Ran `yarn bump-versions` to verify package bump locally

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

Saadnajmi:
Perhaps I'm naive in hoping that if someone other than me does the same thing of updating the android-theme with a change file, it may work, but I'm down to try again! 

I think this will unblock the pipeline temporarily, until some other change goes in. If it gets stuck again, I suggest we try to do #703 again, last time we closed it without merging it.

PR #716, opened 2021-06-03
Checkbox

tamasane:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android
- [x] web

### Description of changes

1. Checkbox redlines are matched with vNext in individual platforms.
    - Added platform-wise SVGs for checkmark.
    - Made use of Icon as a slot for checkmark. Created a new optional slot `checkmarkIcon`. Made previous `checkmark` slot optional. Enabled checkmarkIcon on supported platforms (android,ios,macos,web)
    - Added platform-wise margin/padding settings in the theme (Default settings are for web).
    - Added `checkmarkColor`, `checkboxBackground`, `checkboxBackgroundDisabled`, `checkboxBorderColor` tokens to the theme
    - Created `Checkbox.settings.old` to support win32, windows as SVG checkmark is not enabled on them.

2. Icon
    - Allowed usage of color from style prop in `renderSvg()`. This allows to target `checkmarkColor` on Icon slot.
    - Used `processColor` to convert string to color for macos.

3. Theme-wise changes (iOS, macOS, Android)
    - Added component settings.
    - Added theme token values.

### Verification
| Android                                       | iOS                                      |
| android-checkbox | ios-checkbox |

macOS
macos-checkbox

Web
web-checkbox

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Common/styles.ts
```diff
@@ -45,6 +45,7 @@ export const commonTestStyles = StyleSheet.create({
     fontSize: 12,
     width: 150,
     marginBottom: 8,
+    paddingVertical: 0,
```
Saadnajmi, line 48: is 0 not the default value?

packages/components/Checkbox/src/@types/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '*.svg' {
+  import { SvgProps } from 'react-native-svg';
+  const content: React.FC<SvgProps>;
+  export default content;
+}
```
Saadnajmi, line 5: Did you add this yourself? I would be surprised if you had to manually add an index.d.ts file

packages/components/Checkbox/package.json
```diff
@@ -27,7 +27,8 @@
     "@fluentui-react-native/text": ">=0.9.1 <1.0.0",
     "@fluentui-react-native/tokens": ">=0.8.0 <1.0.0",
     "@uifabricshared/foundation-composable": ">=0.8.0 <1.0.0",
-    "@uifabricshared/foundation-settings": ">=0.8.0 <1.0.0"
+    "@uifabricshared/foundation-settings": ">=0.8.0 <1.0.0",
+    "@fluentui-react-native/icon": ">=0.5.1 <1.0.0"
```
Saadnajmi, line 31: Could we keep this in alphabetical order? Doing `yarn add <package>` at the root of this specific package tends to auto-do that and I'd prefer to just adopt that for future adds to make diffs minimal and just have an accepted order.

packages/components/Checkbox/src/Checkbox.settings.old.ts
```diff
@@ -0,0 +1,95 @@
...
+import { checkboxName, ICheckboxType } from './Checkbox.types';
+
+export const checkboxSelectActionLabel = 'Toggle the Checkbox';
+
+export const settings: IComposeSettings<ICheckboxType> = [
+  {
+    tokens: {
+      borderColor: 'menuItemText',
+      color: 'menuItemText',
+      backgroundColor: 'menuBackground',
```
Saadnajmi, line 11: Since this is win32 specific, would it be appropriate to make this settings.win32.ts?

packages/components/Checkbox/src/Checkbox.settings.ts
```diff
@@ -6,10 +6,11 @@ export const checkboxSelectActionLabel = 'Toggle the Checkbox';
...
   {
     tokens: {
-      borderColor: 'menuItemText',
-      color: 'menuItemText',
-      backgroundColor: 'menuBackground',
+      backgroundColor: 'transparent',
       textBorderColor: 'transparent',
+      borderColor: 'checkboxBorderColor',
+      checkmarkColor: 'checkmarkColor',
+      checkmarkVisibility: 0,
```
Saadnajmi, line 13: Are these new tokens?

packages/components/Checkbox/src/Checkbox.settings.ts
```diff
@@ -19,31 +20,38 @@ export const settings: IComposeSettings<ICheckboxType> = [
...
       },
     },
+    checkmarkSvg: {
+      width: 8,
+      height: 7,
+      style: {
+        marginVertical: 4.5,
+        marginHorizontal: 4,
+      },
+    },
```
Saadnajmi, line 54: Would these not have the same / similar styles?

packages/components/Checkbox/src/Checkbox.settings.win32.ts
```diff
@@ -0,0 +1 @@
+export * from './Checkbox.settings.old';
```
Saadnajmi, line 1: Ah, you do have a settings.win32... I feel like this is an unnecessary level of redirection and we could just leave a comment on top of the original file saying "These settings are win32 only because ...."

packages/components/Checkbox/src/Checkbox.settings.windows.ts
```diff
@@ -0,0 +1 @@
+export * from './Checkbox.settings.old';
```
Saadnajmi, line 1: And then over here, we could export checkbox.settings.win32.ts

packages/components/Checkbox/src/Checkbox.tsx
```diff
@@ -54,7 +56,7 @@ export const Checkbox = compose<ICheckboxType>({
 
     const state: ICheckboxState = {
       ...pressable.state,
-      disabled,
+      disabled: !!disabled,
```
Saadnajmi, line 59: What is the double !! doing?

packages/components/Checkbox/src/Checkbox.types.ts
```diff
@@ -87,7 +88,8 @@ export interface ICheckboxTokens extends FontTokens, IForegroundColorTokens, IBa
 export interface ICheckboxSlotProps {
   root: React.PropsWithRef<IViewProps>;
   checkbox: IViewProps;
-  checkmark: ITextProps;
+  checkmark?: ITextProps;
+  checkmarkSvg?: IconProps;
```
Saadnajmi, line 92: Would it be appropriate to make this one slot that is `ITextProps | IconProps`

Saadnajmi: Or just call it "CheckboxIcon" instead of CheckboxSVG specifically? The former feels more descriptive to me

packages/components/Checkbox/src/assets/checkmark_ios.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="9" height="8" viewBox="0 0 9 8" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M0.853553 4.14646C0.658291 3.9512 0.341709 3.9512 0.146447 4.14646C-0.0488155 4.34172 -0.0488155 4.6583 0.146447 4.85357L2.64645 7.35357C2.85435 7.56147 3.19589 7.54597 3.38411 7.3201L8.38411 1.3201C8.56089 1.10797 8.53223 0.792684 8.32009 0.615902C8.10795 0.43912 7.79267 0.467782 7.61589 0.67992L2.9664 6.25931L0.853553 4.14646Z" fill="currentColor"/>
+</svg>
```
Saadnajmi, line 3: Where did you get these svgs from? Figma?

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -55,16 +55,20 @@ function renderSvg(iconProps: IconProps) {
   const viewBox = iconProps.svgSource.viewBox;
   const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width: width, height: height }, [width, height])[0]);
 
+  // The svg color can be set using either style.color or iconProps.color, where style.color is preferred in case both are set.
+  const getColor = (color1: Color, color2: Color) => (color1 === undefined ? color2 : color1);
+  const color = getColor((style as any).color, iconProps.color);
```
Saadnajmi, line 60: I feel like there's an immutable style merge function that we have that does the same ish thing. I'll defer to @falseLobster

packages/theming/apple-theme/src/appleColors.ios.ts
```diff
@@ -421,5 +421,10 @@ export function paletteFromAppleColors(isDark: boolean): ThemeColorDefinition {
     buttonDisabledSecondaryContent: fluentApple.buttonTitleDisabled,
     buttonHoveredSecondaryContent: fluentApple.communicationBlue,
     buttonPressedSecondaryContent: fluentApple.communicationBlueTint20,
+
+    checkboxBackground: fluentApple.communicationBlue,
+    checkboxBackgroundDisabled: fluentApple.surfacePrimary,
+    checkboxBorderColor: fluentApple.gray600,
+    checkmarkColor: fluentApple.iconOnAccent,
```
Saadnajmi, line 428: Thank you for updating the apple platform as well =) Did you also get these colors from Figma?

packages/theming/theme-types/src/Color.types.ts
```diff
@@ -439,6 +439,10 @@ export interface ControlColorTokens {
   buttonDisabledSecondaryContent: ColorValue;
   buttonHoveredSecondaryContent: ColorValue;
   buttonPressedSecondaryContent: ColorValue;
+
+  checkboxBackground: ColorValue;
+  checkboxBackgroundDisabled: ColorValue;
+  checkmarkColor: ColorValue;
```
Saadnajmi, line 445: Do these names match the fluentui-token-pipeline names? Also FYI @rurikoaraki since she's working on that

Saadnajmi (changes requested):

Saadnajmi:
Normally I'd say we should make a new `experimental-checkbox` package that uses the new framework but it seems @JasonVMo already has a PR that just updates the normal component to the new framework but it hasn't moved in months https://github.com/microsoft/fluentui-react-native/pull/502/files

Should we... just do that now too?

apps/fluent-tester/src/FluentTester/TestComponents/Common/styles.ts
tamasane: No. TextInput has some default padding on Android.

As you can see here the placeholder text appears cut.
Screenshot 2021-06-03 at 9 45 58 PM

packages/components/Checkbox/src/@types/index.d.ts
tamasane: Yes. I added it manually, as per [FURN SVG Setup README](https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/FluentTester/TestComponents/Svg).

packages/components/Checkbox/package.json
tamasane: Sure. Thanks for mentioning this.

packages/components/Checkbox/src/Checkbox.settings.ts
tamasane: Yes these are the new tokens. Added them to respective theme as well.

packages/components/Checkbox/src/Checkbox.types.ts
tamasane: Have to have separate slot names so that styling can be separated. (Combining styles may cause Invalid props supplied to one type of slot). 
`checkmarkIcon` seems reasonable, so making that change.

packages/components/Checkbox/src/assets/checkmark_ios.svg
tamasane: Yep vNext Figmas

packages/experimental/Icon/src/Icon.tsx
tamasane: This change is made to supply the color to `<SvgUri />` which is used as fill value in the SVG. I didn't know how we could target a theme-token to a prop. For all the usages in the repo, the theme-tokens are merged into styles. 
So made this to have an override for the checkmark color using style. (and any future requirements of overriding SVG icon colors using style)

packages/theming/apple-theme/src/appleColors.ios.ts
tamasane: Yep. Used the matching tokens with the same values as in Figma.

packages/theming/theme-types/src/Color.types.ts
tamasane: Ohh didn't check that :/

packages/components/Checkbox/src/Checkbox.settings.old.ts
tamasane: windows too !

packages/components/Checkbox/src/Checkbox.tsx
tamasane: converts `undefined` to `false`

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
```diff
@@ -21,8 +21,10 @@ export const lightnessOptions = [
   { label: 'Dark', value: 'dark' },
 ];
 
+const defaultAsThemeName = ['android', 'ios', 'macos', 'web'].includes(Platform.OS as string);
```
Saadnajmi, line 24: Might be easier to inverse this and just say "if win32, then Office?" :D

packages/components/Checkbox/src/@types/index.d.ts
Saadnajmi: Interesting. Thanks!

packages/components/Checkbox/src/Checkbox.settings.old.ts
Saadnajmi: Ah. I don't really like the name "old" here.. but I get why you're doing it. I think I'd rather it just be ".windows" and for .win32 you just import the windows settings. Or "settingsWithoutSVG" to make it more descriptive.

packages/theming/theme-types/src/Color.types.ts
Saadnajmi: It would be nice if they match up so it's easier to integrate but if they aren't in the pipeline yet, then that's fine!

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -46,6 +46,28 @@ export const appleComponents = {
...
+    checkmarkIcon: {
+      width: 10,
+      height: 10,
+      style: {
+        marginVertical: 2,
+        marginHorizontal: 2,
+      },
+    },
+    // This disables other available states like: hovered, focused, pressed.
+    _precedence: ['disabled', 'boxAtEnd', 'checked'],
```
Saadnajmi, line 69: Good catch! What's "boxAtEnd?"

packages/components/Checkbox/src/Checkbox.types.ts
```diff
@@ -87,7 +88,8 @@ export interface ICheckboxTokens extends FontTokens, IForegroundColorTokens, IBa
 export interface ICheckboxSlotProps {
   root: React.PropsWithRef<IViewProps>;
   checkbox: IViewProps;
-  checkmark: ITextProps;
+  checkmark?: ITextProps;
+  checkmarkIcon?: IconProps;
```
Saadnajmi, line 92: Could we still combine these slots into one? or is that not feasible?

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
tamasane: Yep! Making this change.

packages/theming/apple-theme/src/appleTheme.macos.ts
tamasane: `boxAtEnd` adds `marginStart:0` instead of `marginEnd:0` for the checkbox square

packages/components/Checkbox/src/Checkbox.types.ts
tamasane: I think they can be merged, by introducing Platform.select() for the slot. But that would remove the strict check for the supplied props to the slot. Should I do that?

Saadnajmi: I think we intentionally don't want the API to differ greatly between platforms. which is why I like the optional slots.  I was more thinking about how the image in the button could be a font, PNG, or SVG, yet there's only one slot it goes to and I was wondering if we could mimick that.

packages/theming/apple-theme/src/appleTheme.macos.ts
Saadnajmi: Interesting.. and I guess that wasn't something we couldn't have handled with just the style prop and needed this _precendence thing?

packages/components/Checkbox/src/Checkbox.types.ts
tamasane: The font icons in button are based on codepoint in a ttf file. Making that change here would unnecessarily complicate things. Also it would become tedious to keep track for different platforms. Keeping these separate with the hope that once SVG support comes to windows, we can have a single checkmark.

packages/theming/apple-theme/src/appleTheme.macos.ts
tamasane: Yep, could be handled with style prop. (@JasonVMo's version with new framework addressed this).

apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
```diff
@@ -22,7 +22,7 @@ export const lightnessOptions = [
 ];
 
 export class TesterThemeReference extends ThemeReference {
-  private _themeName: ThemeNames = 'Office';
+  private _themeName: ThemeNames = Platform.OS == ('win32' as any) ? 'Office' : 'Default';
```
rurikoaraki, line 25: Uh this was a mistake commit on my part - you can change it back to default for everyone

Saadnajmi: Thanks! I was going to ask about that :D

packages/theming/apple-theme/src/appleTheme.macos.ts
Saadnajmi: OK. I guess I'll worry about that for a future review with the new compose checkbox then!

packages/components/Checkbox/src/Checkbox.types.ts
Saadnajmi: OK, if it's much more complex to have them together, keeping separate makes sense! Thanks for the explanation =)

packages/components/Checkbox/src/Checkbox.settings.windows.ts
```diff
@@ -1 +1,95 @@
-export * from './Checkbox.settings.old';
+import { IComposeSettings } from '@uifabricshared/foundation-compose';
+import { checkboxName, ICheckboxType } from './Checkbox.types';
+
+export const checkboxSelectActionLabel = 'Toggle the Checkbox';
+
+export const settings: IComposeSettings<ICheckboxType> = [
```
Saadnajmi: Maybe add a comment in the file that the settings differ because windows doesn't support SVG yet

Saadnajmi (approved):
I think my only concern is if we can merge the colors using some merge styles method in FURN. (This comment https://github.com/microsoft/fluentui-react-native/pull/716#discussion_r644957960).

I'll defer to @rurikoaraki or @samuelfreiberg if they know how to do that though

PPatBoyd:
Are the current screenshots accurate?  I may be off base but a couple look a little off with regards to relative size or alignment.  Does android, ios, and/or macos do pixel dithering?
Alignment of inscribed elements raises some interesting issues with scaling -- if layout rounding occurs, it can throw off alignment.  There's a couple ways to deal with it, can you share the goal visual?

tamasane:
@PPatBoyd, I am not sure about pixel dithering issue happening or not. But here are the goal visuals.

| Android | iOS | macOS |
| --- | --- | --- |
| android-checkbox-goal | ios-checkbox-goal | macos-checkbox-goal |

PPatBoyd:
> @PPatBoyd, I am not sure about pixel dithering issue happening or not. But here are the goal visuals.
> 
> Android	iOS	macOS
> android-checkbox-goal	ios-checkbox-goal	macos-checkbox-goal

thanks for sharing these!  Specifically per-platform the iOS checks appear to be smaller than expected (primarily dimensions but also thickness), macOS appears misaligned (top of check is close to the edges), and Android appears to also have misalignment (a pixel or so low), there's a faint white visual that may just be artifacting but I don't expect so, and maybe the check is a little off I can't quite tell for sure?  There isn't the full compare set but it's roughly similar between circular and rounded square (circular being a little more apparent for some).  Is the web visual a separate React fluentui impl or a react-native-web projection?

These are really really nitpicky comments, but it doesn't take much for these specific controls to have a noticeable deviation from ideal.  There are also implications from DPI scaling (Android caring more than iOS/macOS with more DPI scaling values) and text scaling (if either do relational sizing to text) that can exacerbate these issues or shift them into variations -- if you can take some time to compare with each scaling (in isolation and combination) that would be great so we understand the scope of each platform's issues / requirements.  There are potential avenues around / tradeoff decisions / mitigations, but those also aren't necessarily cross-platform compatible... which is rough 😬

packages/experimental/Icon/src/Icon.tsx
rurikoaraki: Here's info on `mergeStyles`, which might be what you're thinking of: https://github.com/microsoft/fluentui-react-native/tree/3c0f5496f0daeb748c6b2daac8d0465a1d58a672/packages/framework/merge-props. There's also a README on `immutableMerge`: https://github.com/microsoft/fluentui-react-native/tree/3c0f5496f0daeb748c6b2daac8d0465a1d58a672/packages/framework/immutable-merge. Both work on objects, while the color prop is a string/`PlatformColor`, so it's a bit overkill here.

As a suggestion, the code could be simplified to `const color = style.color ? style.color : iconProps.color` without defining `getColor`.

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -65,16 +65,20 @@ function renderSvg(iconProps: IconProps) {
     ? '#FFFFFF'
     : '#000000';
 
+  // The svg color can be set using either style.color or iconProps.color, where style.color is preferred in case both are set.
+  const colorString = (style as any).color !== undefined ? (style as any).color : iconColor;
+  const color = (Platform.OS == 'macos' ? processColor(colorString as ColorValue) : colorString) as Color;
```
Saadnajmi, line 70: I totally missed this earlier, but I have found that calling processColor directly is a bug. Depending on the order of calls, it can swap RGBA bits into ARGB, which renders the wrong color. also, of the reason we're doing this only for macOS is because macOS uses PlatformColor, converting to a string won't do anything since platformColors cannot be represented as strings. We should make a follow-up here to remove this logic.

Saadnajmi: I think we made sure that the iconProps color is always a string until we support ColorValue objects on rn-svg. I can find the PR / code where me made that change if we're curious. 
But basically you should never pass an SVG a ColorValue for a color because ColorValue could be an object like a PlatformColor. Instead, pass it the icon color bc we made sure that should be a string I think.

Saadnajmi (commented):
Comment about icon color is a change we should follow up on.

packages/experimental/Icon/src/Icon.tsx
Saadnajmi: No, RN-SVG doesn't support ColorValue at all because it predates RN 0.63 and hasn't been updated to support it yet. 
@rurikoaraki and I talked about it some when she added support for PlatformColor in our theme definition, and we made sure to keep tokens that apply to SVGs (basically Icons) to only be strings and not ColorValues.

A followup (for me) is to update RN-SVG to support ColorValue. There's an [open PR](https://github.com/react-native-svg/react-native-svg/pull/1561) that adds support for PlatformColor, but it's not very extensible to out of tree platforms

samuelfreiberg:
It seems this PR broke Checkbox focus on Win32/Windows. It also changed some of the default Win32 styling. We need to change a lot within this PR, including the structure of the Checkbox.settings files. Let's chat when you can @tamasane

harrieshin:
technically most of the time on iOS client should NOT use checkbox for iOS because it doesn't follow Apple Human Interface guideline. only time checkbox is kinda allowed is when user is selected a cell from a tableview(list). How can we actually stop clients from abusing this control?

PR #720, opened 2021-06-08
Callout string anchor

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Callout anchoring typically anchors to a component.  In some specialized circumstances a component may want to offer another anchor reference for different behavior.  A textbox offering anchoring to the caret or a substring is the focus of this change.

It's worth noting that this behavior is not processed on the JS side for Win32, to leverage extended functionality with dynamic anchors.  Since the Callout component is a wrapped native component and will stay that way for Win32/macOS, this poses no issue.  For other platforms that do not require using a native component (iOS, Android, Windows) JS side processing should be added to avoid making a native component a requirement.  Such a change would have interesting nuance for different kinds of potential sub-component anchors and is left as an exercise for the developer extending to the next platform.  Early projections for nuance include round-tripping layout information from one component to another for determining an anchor rect or potentially an 'anchor registry' component or module (where a module could push logic back to the native side and avoid unnecessary round-tripping, etc.).

### Verification

### Pull request checklist

- [ N/A] Automated Tests
- [ ] Documentation and examples -- the primary example for Win32 is not available in the Win32 test app, and is foregone for now.
- [ N/A] Keyboard Accessibility
- [ N/A] Voiceover
- [ N/A] Internationalization and Right-to-left Layouts


Saadnajmi:
> For other platforms that do not require using a native component (iOS, Android, Windows) JS side processing should be added to avoid making a native component a requirement.

How would you implement callout except as a native component? The only idea I have is positioning a view with an absolute position instead of a relative one. But then it's not on a separate window

PPatBoyd:
> > For other platforms that do not require using a native component (iOS, Android, Windows) JS side processing should be added to avoid making a native component a requirement.
> 
> How would you implement callout except as a native component? The only idea I have is positioning a view with an absolute position instead of a relative one. But then it's not on a separate window

right, that's why I view it as macOS and Win32 only -- Windows iOS and Android don't let you have discrete separate app surfaces.  You can't escape your container.  End of the day all callouts are also absolutely positioned because they aren't positioned by layout; it's just win32 and macOS should be in their own discrete container, and iOS/Android/UWP keep everything on one surface.  

... possible exception considering split screen iOS apps, I think you can bring up two Word windows up side by side?  Not just different apps?  That still wouldn't power callouts but may present some interesting issues more akin to win32.

Having native-side logic is more reasonable through a native module I think, but it doesn't obviate the absolutely positioned component.

sagokhal:
>certain components may proffer a string as an anchor target, such as  anchoring to a point inside the component.

maybe clarify that this'll need some sort of native registry to map that string to actual rects?

packages/components/Callout/src/Callout.tsx
```diff
@@ -15,31 +15,34 @@ export const Callout = compose<ICalloutType>({
...
-    const [targetNativeTag, setTargetNativeTag] = React.useState<number>(undefined);
+    const [nativeTarget, setNativeTarget] = React.useState<number | string | null>(null);
 
     React.useLayoutEffect(() => {
-      if (target?.current) {
-        setTargetNativeTag(findNodeHandle(target.current));
+      if (typeof target === 'string') {
+        setNativeTarget(target);
+      }
+      if (typeof target !== 'string' && target?.current) {
```
sagokhal, line 24: `else if`?

sagokhal (approved):

Saadnajmi:
> > certain components may proffer a string as an anchor target, such as  anchoring to a point inside the component.
> 
> maybe clarify that this'll need some sort of native registry to map that string to actual rects?

Is that how that works? Is this a win32 specific thing?

Issue #674, opened 2021-05-06
Fluent Tester on Win32 does not get keyboard focus initially


ejlayne:
### Environment Information

- **Package version(s)**: (fill this out, include which package manager you're using)
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [ ] iOS
    - **Xcode version**:
  - [ ] macOS
    - **Xcode version**:
  - [x ] win32
  - [ ] windows
  - [ ] android

### Please provide a reproduction of the bug:

1. git pull, yarn, yarn build, cd apps/win32, yarn bundle, yarn run-win32
2. Win32 Tester launches
3. Hit Tab on keyboard, and continue hitting Tab

#### Actual behavior:

No-op, do not see keyboard focus moving between components when I hit Tab. Have to alt tab and then alt tab back to get focus into app for keyboarding to work. 

Possibly a bug with Fluent Button component. However, it looks like focus wants to start on the "Theme: Default" dropdown component, but when I alt-tab/alt-tab-back into the app, and then hit tab, it looks like focus flickers on Theme: Default for a split second before moving to Light/Dark: Auto's dropdown.

#### Expected behavior:

Keyboard focus visibly moves between buttons and interactable elements in Fluent Tester. Expect this to work when I originally launch the Fluent Tester, as focus should go somewhere.

### Priorities and help requested:

Are you willing to submit a PR to fix? No

Requested priority: Normal

Products/applications affected: Just Fluent Tester, but want to make sure it isn't a bug with Fluent Button.

ksiler:
Linking this bug from ADO: AB#5099897

It's possible they're the same bug

ejlayne:
@ksiler looks like the linked bug is closed. Can we close this issue too?

ksiler:
@ejlayne Sorry, I should have followed up here. The bug I linked above is unrelated to this issue. This bug seems to be the same: AB#5127814

ksiler:
Closing this issue. Bug has been resolved.

PR #722, opened 2021-06-09
Fix Fluent Tester on Win32 not receiving keyboard focus on initial render

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This fixes issue #674 for Win32 where the fluent tester does not receive keyboard focus on initial render. Wrapping the tester content in a FocusTrapZone places initial focus on the first focusable item (in this case, the "Theme: Default" dialog).

### Verification

Manually tested.

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
Do we know how this regression took place? I don't recall this being an issue in the past. Does this have to do with the previous issue of our FURN controls not taking keyboard focus?

ksiler:
> Do we know how this regression took place? I don't recall this being an issue in the past. Does this have to do with the previous issue of our FURN controls not taking keyboard focus?

I don't recall it being an issue in the past either but I haven't been able to pinpoint when it occurred. I'll continue looking and comment again here. It does seem like this solution won't work since it's preventing me from clicking on anything but the dropdowns in the header. I'll also look into this and update it with a better solution.

apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -204,16 +204,37 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
...
+          {enableSinglePaneView ? <MobileTestComponentView /> : <TestComponentView />}
+        </View>
+      </RootView>
+    );
+  };
+
+  const Win32TesterContent: React.FunctionComponent<{}> = () => {
+    return (
+      <RootView style={themedStyles.root}>
+        <FocusTrapZone>
```
Saadnajmi, line 225: Could RootView be a FocusTrapZone? 
Related: Was the commits in which I refactored FluentTester for mobile (introducing RootView and the like) be when the regression happened?

Saadnajmi: Dumb question: Why not just a FocusZone?

ksiler: I just created a branch off of 9a6fb7a97c6aa0f8cea946f1d9963bab027eebc6 which is the last commit before your change and I still get the same issue. I skipped running `yarn` and just did `yarn build` under apps, then `yarn bundle` under apps\win32, and `yarn run-win32` to pull the tester up. That should be the right steps to take to test a previous state of the app, right? Just double checking since I have gone back even further and can't seem to find a point where keyboard focus was set on initial render. If not a change in the repo, I'm not sure what else could be effecting this change in behavior.

ksiler: I've also tried running the above with `yarn` and got the same result.

Saadnajmi: That sounds about right to me. And March 4 is the earliest I remember making changes to FluentTester directly so I guess it wasn't that :/

apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -204,16 +204,35 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
...
+          {enableSinglePaneView ? <MobileTestComponentView /> : <TestComponentView />}
+        </View>
+      </RootView>
+    );
+  };
+
+  const Win32TesterContent: React.FunctionComponent<{}> = () => {
+    return (
+      <FocusTrapZone style={themedStyles.root}>
+        {enableSinglePaneView ? <MobileHeader /> : <Header />}
```
samuelfreiberg, line 225: **Code Duplication**

We can avoid the code duplication here for everything within the <FocusTrapZone /> and <RootView /> above. Maybe make a functional component just for the content within those two sections?

So something like:
  const TestContent: React.FunctionComponent<{}> = () => {
    return (
      <View>
        {enableSinglePaneView ? <MobileHeader /> : <Header />}

        <HeaderSeparator />

        <View style={fluentTesterStyles.testRoot}>
          {enableSinglePaneView ? <MobileTestList /> : <TestList />}
          {enableSinglePaneView ? <MobileTestComponentView /> : <TestComponentView />}
        </View>
      </View>
    );
  };

  const TesterWrapper: React.FunctionComponent<{}> = () => {
    return (
      <View>
        {Platform.OS === ('win32' as any) ? (
          <FocusTrapZone style={themedStyles.root}>
            <TestContent />
          </FocusTrapZone>
        ) : (
          <RootView style={themedStyles.root}>
            <TestContent />
          </RootView>
        )}
      </View>
    );
  };

Just a thought to clean up the code a bit

Saadnajmi: RootView itself was just a custom wrapper to let there be a `SafeAreaView` at the top level instead of a `View` for iOS. I think it may be completely fine for us to add an extra case to the platform select of `RootView` to be FocusTrapZone on win32, unless FocusTrapZone itself must be wrapped by a View

ksiler: Yeah, definitely. I wanted to clean this up but got busy and wanted to push out a working change. @Saadnajmi I tried adding another case to platform select but I couldn't get the syntax right for the win32 case... do you know how I might do that?

Saadnajmi: Does this not work?
`
  const RootView = Platform.select({
    ios: SafeAreaView,
    win32: FocusTrapZone,
    default: View,
  });
  `

ksiler: No, because win32 is not part of Platform (only android, ios, windows, macos, and default). I think usually we do "win32 as any" but I'm not sure how to make it work in this case.

Saadnajmi: :') That's frustrating. I guess then Sammy's suggestion is the next best thing.

rurikoaraki (approved):
Code duplication can be mitigated but as it is I think this change looks good. Thanks for fixing this!

apps/fluent-tester/src/FluentTester/FluentTester.tsx
ksiler: I did a modified version of Sammy's suggestion (just had to add some styling to the two additional View's). Thanks for the feedback!

apps/fluent-tester/src/FluentTester/FluentTester.tsx
ksiler: @Saadnajmi sorry for getting back to this late: I chose FocusTrapZone because it was the easiest out of the two to get working in this scenario. The difference is that FTZ traps focus and FZ does not. FTZ was easier to work with because it placed focus automatically (w/o setting any props) whereas FZ would require some more props to be set.

apps/fluent-tester/src/FluentTester/FluentTester.tsx
Saadnajmi: Awesome! I'll probably revisit the whole RootView thing later since it's not as extensible as I hoped.

Saadnajmi (approved):

PR #726, opened 2021-06-15
Add ActivityIndicator Prototype

lindaweng:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added 2 tests with ActivityIndicator prototype. One made from a 3/4 circle arc combined with 2 end circles and another made from a Path SVG. We will most likely proceed with the circle arc and end circles due to the difficulty to scale a Path SVG. 

Props for animating have been created, mimicking the React Native ActivityIndicator props. Tokens for line thickness, color and size will be added later.

### Verification

Ran in the iOS tester app and confirmed that the ActivityIndicators worked as they should.

Video Demos
https://user-images.githubusercontent.com/30802320/122101890-d4b70800-cde2-11eb-94b8-e44f820cf04d.mov
https://user-images.githubusercontent.com/30802320/122101900-d84a8f00-cde2-11eb-9b0e-f95da9b918d7.mov

| Screen Shot 2021-06-15 at 2 11 25 PM | Simulator Screen Shot - iPhone 11 - 2021-06-15 at 13 59 47 Simulator Screen Shot - iPhone 11 - 2021-06-15 at 13 59 57 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Before people comment "Why is that SVG messed up", I wanted to point out that we prototyped two different methods of creating the ActivityIndicator: One with RN core components (Views basically), and one using SVG's. From the prototyping, it seems SVG's aren't as scalable as we hoped, and I asked Linda to keep the tests in to elucidate that =)

.idea/fluentui-react-native.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Saadnajmi, line 1: We should probably delete all the .idea/ files. Nothing against jetbarains, but these are generated files we don't necessarily want as source and I think most use VS / VS Code for this repo anyway.

If you do want to keep using Jetbrains stufff, then I would add `.idea/*` to our gitignore file so git ignores adding them.

apps/android/package.json
```diff
@@ -19,7 +19,7 @@
     "android": "react-native run-android "
   },
   "dependencies": {
-    "@fluentui-react-native/tester": "^0.24.2",
+    "@fluentui-react-native/tester": "^0.25.1",
```
Saadnajmi, line 22: I would encourage you to do the following to lower the amount of diffs that we get in this PR:

`git checkout activityindicator_prototype`
`git pull upstream master`
`git push`

Saadnajmi (changes requested):

apps/android/package.json
Saadnajmi: Actually I just did it. The purpose was that `activityindicator_prototype` was out of sync with master but your branch wasn't, so your PR looked like it had a lot more changes than it did. After syncing it, the diff went from 184 files to 14 :')

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,134 @@
...
+          })
+      ).start();
+    });
+  }
+  const interpolateSpin = spinAnimation.interpolate({
+    inputRange: [0, 359],
+    outputRange: ['0deg', '359deg'],
+  });
+
+  // or display: none, but since React Native ActivityIndicator still takes up space when hidden, using opacity
```
rurikoaraki, line 79: This may not be in scope, but one thing about using opacity to control hidden-ness is that it doesn't trick screen readers, so they will still announce the presence of things hidden in this way. If I remember correctly display:none works better for screen readers, but I might also just be making things up.

rurikoaraki: If the opacity is only for animating and the component is removed when hidden, then this is fine as it is

rurikoaraki: Also just realized this is tester code, so all of this is probably a moot point.

Saadnajmi: This is all tester code... but it should actually be one of the props :P 

I'm actually in favor of ignoring this prop altogether. It feels like a legacy prop and clients can hide views themselves by modifying the style.. and also less code for us to maintain. It would introduce a difference between the Fluent ActivityIndicator and the core one, but I think that's fine...

Saadnajmi: Thanks for the points @rurikoaraki , bringing up that just changing the opacity would not work well for accessibility

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,134 @@
...
+// Top end circle:
+  // top: R - D - borderRadius
+    // top edge is already 1 borderRadius below the big circle
+  // left: R - D - borderRadius
+    // left edge is already 1 borderRadius to the right
+// Bottom end circle:
+  // top: R + D - 2 * borderRadius
+    // 1 borderRadius explained same as top end circle. Extra borderRadius since bottom end circle is 1 borderRadius below top end circle already.
+  // left: R - D - borderRadius
+    // same explanation as above
```
Saadnajmi, line 26: For multiline comments, we like doing the following 

/**
  * comment
  * comment
  */

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,134 @@
...
+  },
+  endBottom: {
+    width: border,
+    height: border,
+    borderRadius: border/2,
+    backgroundColor: '#919191',
+    top: radius + xydist - 2*border,
+    left: radius - xydist - border,
+  }
+});
```
Saadnajmi, line 53: Since this takes a border and radius, perhaps this should be a function that returns a Stylesheet given a border or radius as inputs? We do a similar thing with "getThemedStyles()" somewhere in FluentTester.tsx

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+
+export type ActivityIndicatorProps = ViewProps & {
+    animating?: boolean;
+    hidesWhenStopped?: boolean;
+}
+const ActivityIndicator = (props: ActivityIndicatorProps) => {
+  const spinAnimation = useRef(new Animated.Value(0)).current;
+  if (props.animating) {
```
Saadnajmi, line 62: This should be in it's own package `@fluentui-react-native/activityindicator` rather than just a function in the test code. That would ensure people can actually import it and keep a separation of responsibility between the component and the test.
You can copy one of the existing packages in `packages/experimental/*` as a template and then change the names / files to be your component. I have some info on how to do that in our contribution guide https://github.com/microsoft/fluentui-react-native/blob/master/CONTRIBUTING.md

apps/android/package.json
lindaweng: Great, thank you!

.idea/fluentui-react-native.iml
lindaweng: Just deleted.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
lindaweng: Yeah, that makes sense. A function would allow it to be called which will probably also make it easier when tokenizing the size and line thickness.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
lindaweng: Just added as its own package. Along the way, there were some extra steps that aren't on the contributing guide and it might be helpful to add those to the guide. (Deleting cache and running `yarn install` to load the package as a dependency for the app and some import/export tips.)

packages/experimental/ActivityIndicator/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@fluentui-react-native/experimental-activity-indicator",
+  "version": "0.1.0",
+  "description": "A cross-platform Fluent Activity Indicator component",
+  "main": "src/index.ts",
+  "module": "src/index.ts",
+  "typings": "lib/index.d.ts",
+  "onPublish": {
```
Saadnajmi, line 8: For my own knowledge, which package did you use to base what info / scripts to copy? I'm curious so I know what to recommend for the future in case some package is missing a script

packages/experimental/ActivityIndicator/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "main": "lib-commonjs/index.js",
+    "module": "lib/index.js"
+  },
+  "scripts": {
+    "build": "fluentui-scripts build",
+    "just": "fluentui-scripts",
+    "clean": "fluentui-scripts clean",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "verify-api": "fluentui-scripts verify-api-extractor",
+    "update-api": "fluentui-scripts update-api-extractor"
```
Saadnajmi, line 18: I think we also want depcheck here?

packages/experimental/ActivityIndicator/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "@fluentui-react-native/tokens": ">=0.9.0 <1.0.0"
+  },
+  "devDependencies": {
+    "@types/react-native": "^0.63.0",
+    "@uifabricshared/build-native": "^0.1.1",
+    "@uifabricshared/eslint-config-rules": "^0.1.1",
+    "react-native": "^0.63.4"
+  },
+  "peerDependencies": {
+    "react-native": ">=0.63.4"
```
Saadnajmi, line 31: Nit: One is "^", the other ">=". I think they both effectively resolve to the same thing (cheatsheet here https://devhints.io/semver) for versions less than 1.x.x.

But we should use the same symbols. It's especially concerning if you copied this from another package because it means we should go fix that other package and any other packages that differ.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+// radius is from center to halfway border, border is line thickness
+export const getActivityIndicatorStyle = (radius: number, border: number) => {
+  const xydist = Math.sqrt(2)/2 * radius;
+
+  return {
+    semicircle: {
+      width: radius*2 + border, // from outside edge of circle on one side to the opposite side
+      height: radius*2 + border,
+      borderWidth: border,
+      borderColor: '#919191',
```
Saadnajmi, line 31: Since this is used in more than one place, it's preferable to separate it to it's own const variable. Also maybe leave a comment about where you got that exact value (I'm assuming from Figma)

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+}
+
+
+export type ActivityIndicatorProps = ViewProps & {
+    animating?: boolean;
+    hidesWhenStopped?: boolean;
+}
+export const ActivityIndicator = (props: ActivityIndicatorProps) => {
```
Saadnajmi, line 59: nit: space in between these two things. Though, I would assume `prettier` would do that for you, else you fail PR checks.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+      Animated.loop(
+          Animated.timing(spinAnimation, {
+            toValue: 359,
+            duration: 750,
+            useNativeDriver: true,
+            easing: Easing.linear
+          })
+      ).start();
+    });
+  }
```
Saadnajmi, line 72: Something about defining a hook only on a boolean feels wrong to me... but I'm not sure if it's actually a problem.
For future reference, could you link the tutorial /documentaion where you learned how to do this in case we need to go back and revisit this later?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const interpolateSpin = spinAnimation.interpolate({
+    inputRange: [0, 359],
+    outputRange: ['0deg', '359deg'],
+  });
+
+  // or display: none, but since React Native ActivityIndicator still takes up space when hidden, using opacity
+  let hideOpacity = 1;
+  if (props.animating == false && props.hidesWhenStopped == true) {
+    hideOpacity = 0;
+  }
```
Saadnajmi, line 82: As Ruriko discussed, we probably want to use `display:none` because the view would still be seen by screen readers. Another thing we can check is use the Xcode debugger + VoiceOver to see if a native ActivityIndicator announces itself to screenreader when the "hidesWhenStopped" prop is set. I can help with this too if that sounds unclear.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <Animated.View style={[getActivityIndicatorStyle(40, 10).semicircle, { transform: [{rotateZ: interpolateSpin}, {perspective: 10}] }, {opacity: hideOpacity}]}>
```
Saadnajmi, line 84: I would separate out the style into it's own variable for readability. It'll also help when we use the compose framework, because that has some API's for building style props for you

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <View style={getActivityIndicatorStyle(40, 10).endTop}></View>
+      <View style={getActivityIndicatorStyle(40, 10).endBottom}></View>
```
Saadnajmi, line 86: We definitely should separate it out so we don't call the function 3 times where one is sufficient. Then we could just do `style={themedStyles.endTop}` which we do in FluentTester.tsx

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+    </Animated.View>
+  )
+};
+
+// Default props mimic the React Native ActivityIndicator
+ActivityIndicator.defaultProps = {
+  animating: true,
+  hidesWhenStopped: true
+}
```
Saadnajmi, line 95: I didn't know defaultProps existed. Good to know =)

packages/experimental/ActivityIndicator/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ActivityIndicator';
```
Saadnajmi, line 1: nit: Should probably add newline for end here and in ActivityIndicator.tsx + tsconfig.json .

I'm not sure if/why it's better but I think it's the preferred style.

Saadnajmi (changes requested):

packages/experimental/ActivityIndicator/package.json
lindaweng: I copied from Shimmer, recommended by the contributing guide.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: I just tested voice over with the Accessibility Inspector and it skips over all 3 ActivityIndicators (spinning, stopped and not hidden, stopped and hidden). I'm not sure why, but it may be because it's a custom component?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: This is where I got the useEffect and loop part: https://reactnativeforyou.com/react-native-animation-tutorials-how-to-create-looping-animations-in-react-native/

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
Saadnajmi: Thanks =) If you have a list of those, we can definitely improve the the contribution guide :D 
Thanks for teasing the guide out through practice, haha

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
Saadnajmi: Oh yes... we probably need to properly set it's accessibilityRole / accessibilityLabel prop. I think what Callstack does in react-native-paper might be a good one to base off of https://github.com/callstack/react-native-paper/blob/1638b94a0fd8a481dfa90bf5833d887023eaec5d/src/components/ActivityIndicator.tsx#L157

packages/experimental/ActivityIndicator/package.json
lindaweng: I just looked through the other packages and they all have some of each style so I think it's okay, but I agree that using a consistent notation would be nice.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
Saadnajmi: Looking at the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#explanation), it seems we are violating the "Don't put hooks inside conditionals" part. Their suggestion is to put the condition inside the hook itself. 
Hooks rely on their order of calls matching between different renders, which is why conditions stopping them from being called messes with them. 

Another alternative could be to do what Callstack does and have separate start/stop animation methods https://github.com/callstack/react-native-paper/blob/1638b94a0fd8a481dfa90bf5833d887023eaec5d/src/components/ActivityIndicator.tsx#L83
I think they need that because their animation is much more complicated though.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: I'm not quite sure what you're referring to but I put the `radius*2 + border` in its own variable. The comment describes how I got to that formula based on the other variables.
As for the `borderColor`, yes I got that from Figma and I also separated it out into its own variable.

Saadnajmi: oops. Meant backgroundColor 😅

packages/experimental/ActivityIndicator/package.json
Saadnajmi: That's fair. For the react-native Dependency sepcifically, I think we try to be more strict. we can leave this as a followup since I think it doesn't have any real effect.

packages/experimental/ActivityIndicator/package.json
lindaweng: I'm not sure what that means or does but from looking at other packages, it seems that some have it (Avatar and Button) and some don't (Shimmer and Icon).

Saadnajmi: depcheck is a script that makes sure that the dependencies we specify are accurate. It will error if we have an unused dependency in that package, or use a dependency that isn't listed. We should probably add it, then make a followup bug to add it to the other packages missing it, as it's a pretty important check.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -75,15 +79,19 @@ export const ActivityIndicator = (props: ActivityIndicatorProps) => {
     outputRange: ['0deg', '359deg'],
   });
 
-  // or display: none, but since React Native ActivityIndicator still takes up space when hidden, using opacity
-  let hideOpacity = 1;
+  // React Native ActivityIndicator still takes up space when hidden, so to perfectly match would use opacity
+  // using display: 'none' because screen reader might still read it when hidden with opacity
+  let displayValue = 'flex';
```
Saadnajmi, line 84: ```const displayValue = (props.animating == false && props.hidesWhenStopped == true) ? 'flex': 'none'```

We prefer consts and immutability when possible to eliminate the class of bugs where a variable can be reassigned to a new value after it has been created. It's also perfectly valid to have conditional initialization like I did.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: Yeah, after reading that, it makes sense and I put the conditional inside the `.useEffect()`. Even with Callstack, although they have their animations defined outside, the conditionals for which one eventually gets called is within `.useEffect()` as well.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -75,15 +79,19 @@ export const ActivityIndicator = (props: ActivityIndicatorProps) => {
...
+    displayValue = 'none';
   }
+
+  const activityIndicatorStyles = getActivityIndicatorStyle(40, 10);
+
   return (
-    <Animated.View style={[getActivityIndicatorStyle(40, 10).semicircle, { transform: [{rotateZ: interpolateSpin}, {perspective: 10}] }, {opacity: hideOpacity}]}>
-      <View style={getActivityIndicatorStyle(40, 10).endTop}></View>
-      <View style={getActivityIndicatorStyle(40, 10).endBottom}></View>
+    <Animated.View style={[activityIndicatorStyles.semicircle, { transform: [{rotateZ: interpolateSpin}, {perspective: 10}] }, {display: displayValue as 'flex'|'none'}]}>
```
Saadnajmi, line 92: Is there a reason that this style is separated out into multiple objects, instead of all just being contained in `activityIndicatorStyles.semicircle`?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: I just tested changing the opacity to hide the indicator with the accessibility role set and the voice over skips the hidden activity indicator so it seems to be okay.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: They are all styling different views so I don't think they could be combined into one object. The `.semicircle` is styling the circle arc, (maybe I should change this name), the `.endTop` and `.endBottom` are styling the respective rounded ends. If they were all moved into one object, at least one of the three wouldn't be the right size or in the right spot. Is this what you meant by "contained in"?

Saadnajmi: I meant why isn't `{ transform: [{rotateZ: interpolateSpin}, {perspective: 10}] }, {display: displayValue as 'flex'|'none'}` just part of `activityIndicatorStyles.semicircle` ?

Saadnajmi (approved):

PR #729, opened 2021-06-15
Token persona coin integration

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This change redirects the colors used in the persona coin to be token from the global tokens instead of hardcoding them directly. The colors used are the same hex value as the hardcoded ones, even though the names of the colors may be different.
In the process I noticed that charcoal has been mis-defined, so that was also fixed in the input file, and the outputs have been regenerated.

### Verification

Will be comparing colors before and after change in tester.

| image | image |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
I can follow up offline, but does the algorithm for name -> color also match between here and what we have in Fluent android / iOS / macOS? They all use the java hash algorithm which I can try to point out.

packages/components/PersonaCoin/src/PersonaCoin.helpers.ts
```diff
@@ -52,34 +53,34 @@ const sizeTable: {[P in PersonaSize]: PersonaSizeConfig} = {
...
+  lightPink: globalTokens.color.berry.primary,
+  pink: globalTokens.color.hotPink.primary,
+  magenta: globalTokens.color.grape.primary,
+  purple: globalTokens.color.purple.primary,
+  orange: globalTokens.color.pumpkin.primary,
+  lightRed: globalTokens.color.red.primary,
+  darkRed: globalTokens.color.burgundy.primary,
+  violet: globalTokens.color.orchid.primary,
+  gold: globalTokens.color.brass.primary,
+  burgundy: globalTokens.color.darkRed.primary,
```
Saadnajmi, line 79: could we also just.. update the key names to match?

rurikoaraki:
> I can follow up offline, but does the algorithm for name -> color also match between here and what we have in Fluent android / iOS / macOS? They all use the java hash algorithm which I can try to point out.

I'm not entirely sure what you're asking about. The way I matched the colors together was to look at the hex color value and found the matching token from the global tokens file, so there isn't an automated thing for this.

packages/components/PersonaCoin/src/PersonaCoin.helpers.ts
rurikoaraki: Still trying to get a handle on what we're okay with breaking vs. not. Long run it makes sense update these so that the names all align, talking with @FalseLobster on whether this is an okay change now.

Saadnajmi:
> > I can follow up offline, but does the algorithm for name -> color also match between here and what we have in Fluent android / iOS / macOS? They all use the java hash algorithm which I can try to point out.
> 
> @Saadnajmi I'm not entirely sure what you're asking about. The way I matched the colors together was to look at the hex color value and found the matching token from the global tokens file, so there isn't an automated thing for this.

Yeah, sorry I guess I can explain that better.

Given a primary+secondary string (say, a name + email), the Avatar/PersonaCoin control must decide what color to make the background (and also the foreground, aka the color of the text, to ensure enough contrast). On iOS, macOS, and Android, we elected to concatenate the strings, hash it using the Java Hash method (re-implementing it on apple), and use that hash as an index after modding into the array of colors.

Here is where we do it on macOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/macos/FluentUI/AvatarView/AvatarView.swift#L496
Here's where we do it on iOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/ios/FluentUI/Avatar/InitialsView.swift#L145

We very deliberately invested effort to have macOS match iOS (which itself matched android) once we started having cross-platform scenarios like MeControl that needed the colors of their Avatar match between native and react native, and (less so) between platforms to ensure a consistent experience. This was also one of the major reasons I couldn't re-use `PersonaCoin` for MeControl on iOS and wrote my own native module `experimental-avatar`, which just wrapped the FluentUI Apple control.

This can definitely be a followup change though, since this PR is just making sure we consume the same colors, and what I described adds complexity.

rurikoaraki:
> > > I can follow up offline, but does the algorithm for name -> color also match between here and what we have in Fluent android / iOS / macOS? They all use the java hash algorithm which I can try to point out.
> > 
> > 
> > @Saadnajmi I'm not entirely sure what you're asking about. The way I matched the colors together was to look at the hex color value and found the matching token from the global tokens file, so there isn't an automated thing for this.
> 
> Yeah, sorry I guess I can explain that better.
> 
> Given a primary+secondary string (say, a name + email), the Avatar/PersonaCoin control must decide what color to make the background (and also the foreground, aka the color of the text, to ensure enough contrast). On iOS, macOS, and Android, we elected to concatenate the strings, hash it using the Java Hash method (re-implementing it on apple), and use that hash as an index after modding into the array of colors.
> 
> Here is where we do it on macOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/macos/FluentUI/AvatarView/AvatarView.swift#L496
> Here's where we do it on iOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/ios/FluentUI/Avatar/InitialsView.swift#L145
> 
> We very deliberately invested effort to have macOS match iOS (which itself matched android) once we started having cross-platform scenarios like MeControl that needed the colors of their Avatar match between native and react native, and (less so) between platforms to ensure a consistent experience. This was also one of the major reasons I couldn't re-use `PersonaCoin` for MeControl on iOS and wrote my own native module `experimental-avatar`, which just wrapped the FluentUI Apple control.
> 
> This can definitely be a followup change though, since this PR is just making sure we consume the same colors, and what I described adds complexity.

Oh I see. I'm not immediately finding anything like that for win32, but it's possible that logic exists somewhere else. I can see if we have the logic and if not, implement it. I agree it can be a separate PR.

Saadnajmi:
> > > > I can follow up offline, but does the algorithm for name -> color also match between here and what we have in Fluent android / iOS / macOS? They all use the java hash algorithm which I can try to point out.
> > > 
> > > 
> > > @Saadnajmi I'm not entirely sure what you're asking about. The way I matched the colors together was to look at the hex color value and found the matching token from the global tokens file, so there isn't an automated thing for this.
> > 
> > 
> > Yeah, sorry I guess I can explain that better.
> > Given a primary+secondary string (say, a name + email), the Avatar/PersonaCoin control must decide what color to make the background (and also the foreground, aka the color of the text, to ensure enough contrast). On iOS, macOS, and Android, we elected to concatenate the strings, hash it using the Java Hash method (re-implementing it on apple), and use that hash as an index after modding into the array of colors.
> > Here is where we do it on macOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/macos/FluentUI/AvatarView/AvatarView.swift#L496
> > Here's where we do it on iOS: https://github.com/microsoft/fluentui-apple/blob/4cea132624f1b463c98039b6f480c618a71af88f/ios/FluentUI/Avatar/InitialsView.swift#L145
> > We very deliberately invested effort to have macOS match iOS (which itself matched android) once we started having cross-platform scenarios like MeControl that needed the colors of their Avatar match between native and react native, and (less so) between platforms to ensure a consistent experience. This was also one of the major reasons I couldn't re-use `PersonaCoin` for MeControl on iOS and wrote my own native module `experimental-avatar`, which just wrapped the FluentUI Apple control.
> > This can definitely be a followup change though, since this PR is just making sure we consume the same colors, and what I described adds complexity.
> 
> Oh I see. I'm not immediately finding anything like that for win32, but it's possible that logic exists somewhere else. I can see if we have the logic and if not, implement it. I agree it can be a separate PR.

Yeah.. looking at PersonaCoin.tokens.initialsBackground.ts, it almost looks like the same backgroundColor is always used unless overridden by the client. I guess I misunderstood how the PersonaCoin chose backgroundColor.

packages/components/PersonaCoin/src/PersonaCoin.helpers.ts
rurikoaraki: Introduced a new prop, `coinColorFluent`, and marked the old prop as deprecated. Updated the test page to use the new prop.

packages/components/PersonaCoin/package.json
```diff
@@ -23,14 +23,14 @@
...
-    "@uifabricshared/foundation-tokens": ">=0.9.0 <1.0.0",
-    "@uifabricshared/theming-ramp": ">=0.14.0 <1.0.0"
+    "@uifabricshared/foundation-compose": "^1.9.2",
+    "@fluentui-react-native/adapters": ">=0.6.1 <1.0.0",
+    "@fluentui-react-native/theme-tokens": ">=0.2.1 <1.0.0",
+    "@fluentui-react-native/tokens": ">=0.9.2 <1.0.0",
+    "@uifabricshared/foundation-composable": ">=0.8.2 <1.0.0",
+    "@uifabricshared/foundation-settings": ">=0.9.1 <1.0.0",
+    "@uifabricshared/foundation-tokens": ">=0.9.1 <1.0.0",
+    "@uifabricshared/theming-ramp": ">=0.14.2 <1.0.0"
```
Saadnajmi: Are these updates a part of the PR? Probably fine but I thought we mostly let beachball handle that.

Saadnajmi (approved):
The global tokens map is great :')

packages/components/PersonaCoin/package.json
rurikoaraki: I'm pretty sure these changes came in from a merge from master... The only new line from the PR should be adding theme-tokens.

PR #730, opened 2021-06-17
Fix prettier for FURN

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

I noticed that running prettier via `yarn prettier` wasn't working in the repo. This attempts to fix that.
If you want to run the prettier check, you can run `yarn prettier` from root. If you want to fix the issues, run `yarn prettier-fix`.

The reason I added a new command was because I couldn't figure out how to get lage to pass the --fix argument into `just prettier`. When I would try to just append `--fix true` to `yarn prettier`, I would get an error message and it would look like it was trying to pass `--` in twice so it'd look like `prettier -- --fix true` which would fail, but it would ignore the arg if I just passed `fix true` without the `--`.

Also this doesn't fix prettier running in the scripts folder, ironically. Still looking into that.

### Verification

Re-saved files using VSCode configured with Prettier, made sure there were no new changes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

ksiler:
This is great! Thanks for looking into this :)

ksiler (approved):

lage.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     bundle: ['build-tools'],
     clean: [],
     depcheck: ['build-tools'],
+    just: ['build-tools'],
```
JasonVMo, line 10: Given that you are sending calls to execute the prettier and prettier-fix commands this should probably have an entry for prettier and prettier-fix instead of just

JasonVMo (approved):
It would probably have been easier to split this into the build system changes and the run of the command itself but combined is fine. One small comment for the lage.config.js file to match the commands you are running since we need to build the TS in our scripts package before executing the prettier commands (which use the output of that build)

As a followup do we have plans for adding this to the CI?

rurikoaraki:
> As a followup do we have plans for adding this to the CI?

@JasonVMo Yeah I was thinking about doing this. There's a few things I want to look into (sometimes prettier complains about files that don't need fixing), but we should have it hooked up to CI.

PR #731, opened 2021-06-17
Add Ring to PersonaCoin

tamasane:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This change adds activity rings to PersonaCoin without affecting its current structure. The anatomy of activity ring consists of: inner-ring, glow and outer-ring. The inner-ring and outer-ring form a background for the glow. Thus, the activity ring is achieved using 2 slots: `activityGlow`, i.e. the glow part and `activityRing`, i.e. the background rings. 

The rings can be displayed by passing a config prop `activityRing` with one of the following config defined:
1. ringColor - color of the glow (default value from theme token - personaActivityGlow)
2. ringBackgroundColor - color of the inner, outer rings (default values from theme token - personaActivityRing)
3. ringThickness - thickness for the glow (as well as the inner and outer rings) (default is set to 4) 
4. innerGap - thickness/width of inner-ring
5. transparent - if set true, makes ring transparent
6. accent - if set true, glow uses brand accent color

Added mobile friendly test for android, tried making it available to ios as well but didnt get around linking '@react-native-community/slider'. Tagging @Saadnajmi for helping with it.

### Verification
| Android                                       | iOS                                    | 
| android | ios |

macOS
macos

win32
win32

windows
windows

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/PersonaCoin/src/PersonaCoin.types.ts
```diff
@@ -2,6 +2,7 @@ import type { IViewProps } from '@fluentui-react-native/adapters';
 import { ImageProps, ViewProps, ImageURISource, TextProps } from 'react-native';
 import { IBackgroundColorTokens, IForegroundColorTokens } from '@fluentui-react-native/tokens';
 import { IRenderData } from '@uifabricshared/foundation-composable';
+import { Color } from '../../../theming/theme-types/lib';
```
rurikoaraki, line 5: You should import this from the package directly instead of referencing it using the relative path, so that if the package changes its location, the import statement still works. This also aligns with other imports in the repo.

The theme-types package is `@fluentui-react-native/theme-types`.

rurikoaraki (commented):
Curious - is this meant to be a mobile-specific feature? If so, is there anything preventing the activity ring from being used on Mac/Windows?

packages/components/PersonaCoin/src/PersonaCoin.types.ts
```diff
@@ -2,7 +2,7 @@ import type { IViewProps } from '@fluentui-react-native/adapters';
 import { ImageProps, ViewProps, ImageURISource, TextProps } from 'react-native';
 import { IBackgroundColorTokens, IForegroundColorTokens } from '@fluentui-react-native/tokens';
 import { IRenderData } from '@uifabricshared/foundation-composable';
-import { Color } from '../../../theming/theme-types/lib';
+import { Color } from '@fluentui-react-native/theme-types';
```
Saadnajmi, line 5: Huh.. this exists....
``` export type Color = keyof Palette | ColorValue;```

I would jus use ColorValue here.. since every token we have should be a ColorValue so it still works with tokens

**Comment by Saadnajmi** [comment]
Filed #734 to track the slider bug

**Comment by Saadnajmi** [comment]
A question: Should this just be named "ring" and not ActivityRing? Aligning on names is not the super-most-important thing to do, but I think the Fluent Vnext Avatar just calls it a ring.

File: apps/ios/src/Podfile [hunk 3]
```diff
@@ -29,6 +29,7 @@ use_test_app! do |target|
     pod 'FluentUI-React-Native-Date-Picker', :path => '../../../packages/experimental/NativeDatePicker/FluentUIReactNativeDatePicker.podspec'
     pod 'react-native-menu', :path => '../../../node_modules/@react-native-menu/menu'
     pod 'RNSVG', :path => '../../../node_modules/react-native-svg'
+    pod 'react-native-slider', :path => '../../../node_modules/@react-native-community/slider'
```
Saadnajmi, line 32: Thanks! I never actually told you how to do this 😅 sorry for not following up!

Saadnajmi:
> Curious - is this meant to be a mobile-specific feature? If so, is there anything preventing the activity ring from being used on Mac/Windows?

This is probably because "Slider" only works on android, iOS, and windows, and not macOS. My team should probably look into adding that support, since it's probably just not ejected yet.

packages/components/PersonaCoin/src/PersonaCoin.types.ts
```diff
@@ -60,6 +61,17 @@ export type PersonaCoinFluentColor =
...
+export type RingThickness = number | 'xSmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge';
+
+export interface IRingConfig {
+  accent?: boolean;
+  transparent?: boolean;
+  ringColor?: Color; // glow
+  ringBackgroundColor?: Color; // inner/outer rings
+  ringThickness?: RingThickness;
+  innerGap?: RingThickness;
+}
```
Saadnajmi, line 73: Can this just a type called RingProps?

packages/theming/android-theme/src/colorsTokens.ts
```diff
@@ -199,5 +199,8 @@ export function paletteFromAndroidColors(p: FluentUIAndroidPalette): ThemeColorD
     checkboxBackground: p.buttonBackground,
     checkboxBackgroundDisabled: p.checkboxDisabled,
     checkboxBorderColor: p.checkboxBorder,
+
+    personaActivityRing: p.surfacesPrimary,
+    personaActivityGlow: p.buttonBackground,
```
Saadnajmi, line 204: We (the apple / FX team) probably need to add these tokens to the apple theme as well?

apps/fluent-tester/src/FluentTester/TestComponents/PersonaCoin/StandardUsage.tsx
```diff
@@ -45,6 +46,19 @@ const allColors: WithUndefined<PersonaCoinFluentColor>[] = [
...
+  const pickerStyles = { color: theme.colors.inputText as ColorValue, ...commonStyles.header };
+  return (
+    <Picker prompt={prompt} style={pickerStyles} selectedValue={selected} onValueChange={onChange}>
+      {collection.map((value, index) => (
+        <Picker.Item label={value} key={index} value={value} />
+      ))}
+    </Picker>
+  );
+};
+
```
Saadnajmi, line 61: Not for this review: But I like this pattern of injecting style into an existing component. I would love to integrate this elsewhere especially because Picker is in core for win32 and not in core for all the other platforms.

packages/components/PersonaCoin/src/PersonaCoin.tokens.ring.ts
```diff
@@ -0,0 +1,64 @@
...
+      width: effectiveSize,
+      height: effectiveSize,
+      position: 'absolute',
+      top: -effectiveRingThickness,
+      left: -effectiveRingThickness,
+    },
+  };
+}
+
+function _buildGlowStyles(tokenProps: IPersonaCoinTokens, theme: ITheme): ViewProps {
```
Saadnajmi, line 36: why the leading underscore?

tamasane: this was something I saw in other tokens.xx.ts files. so maintained uniformity here too !

packages/theming/android-theme/src/colorsTokens.ts
tamasane: Yep! Or maybe I could just do it here!

packages/components/PersonaCoin/src/PersonaCoin.types.ts
tamasane: ring is a prop to the Avatar / PersonaCoin, and these values are used to create the styles for the ring slots. So calling it a RingConfig seems more appropriate =)

apps/ios/src/Podfile
tamasane: I just had a quick look at the Podfile, and I said to myself lets try it! And it worked, so happy =)

packages/components/PersonaCoin/src/PersonaCoin.tokens.ring.ts
Saadnajmi: looks like it's a convention for functions that get transformed by the StyleFunction at the end. Interesting...

apps/android/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@fluentui-react-native/tester": "^0.27.1",
+    "@react-native-community/slider": "^3.0.3",
```
Saadnajmi, line 25: Can you add this to the iOS package.json too?

packages/theming/android-theme/src/colorsTokens.ts
Saadnajmi: Also, where did you get these tokens from?

I ask because I have two feelings:

`personaActivityRing` should just match the PersonaCoin initials background color (which I think is called coinColor), and not really be a token. The same token (something like AvatarBackgroundColor/PersonaBackgroundColor) should change both, with a prop to override the color if you need it something else.

For `personaActivityGlow`, I can't imagine you'd ever want it something other than transparent, or white for light mode / black for dark mode?

apps/android/package.json
tamasane: Oh is it required? Like the slider is functional on iOS already. Heres a quick demo!

https://user-images.githubusercontent.com/23552252/123345697-437b2a80-d574-11eb-8a10-7a8c90ca394e.mov

Saadnajmi: Not required, but good for pedantic-ness =)
Our test apps are structured weirdly, and missing dips in package.json have caused weird errors in the past. Also I want to combine the test apps one day so this would be good to know the actual set per app

packages/theming/android-theme/src/colorsTokens.ts
tamasane: Okay so `personaActivityRing` refers to the inner and outer rings - the background rings to the main presence ring
And the `personaActivityGlow` refers to the presence ring.
Both these are just default values that have to should be taken if `ringColor` (glow/presence/personaActivityGlow) & `ringBackgroundColor` (personaActivityRing) are not supplied in `RingConfig`.

Saadnajmi: OK I'm looking at what we do in FluentUI Apple Vnext and it looks like we have similar tokens..
https://github.com/microsoft/fluentui-apple/blob/vnext-prototype/ios/FluentUI/Vnext/Avatar/AvatarTokens.swift
https://github.com/microsoft/fluentui-apple/blob/11b9a86f8ef8f2b161604d0ccf8b4908850c3b85/tools/sgen/input/FluentUIStyle.yml#L135

rurikoaraki (commented):
I have a few questions and I think some of Saad's questions about the tokens might be worth answering before this gets merged, but on the whole the changes look good to me.

packages/components/PersonaCoin/src/PersonaCoin.tokens.ring.ts
```diff
@@ -0,0 +1,64 @@
...
+  const { physicalSize } = calculateEffectiveSizes(tokenProps);
+  const { ring } = tokenProps;
+
+  if (!ring) return {};
+
+  const innerGap = ring.innerGap != undefined ? getRingThickness(ring.innerGap) : getRingThickness(ring.ringThickness || 'xxlarge');
+
+  const effectiveRingThickness = 2 * getRingThickness(ring.ringThickness || 'xxlarge') + innerGap;
+  const effectiveSize = physicalSize + 2 * effectiveRingThickness;
+  const ringColor = ring.transparent ? 'transparent' : ring.ringBackgroundColor || theme.colors.personaring;
```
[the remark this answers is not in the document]
tamasane: Yep, this definitely makes sense!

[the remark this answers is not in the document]
Saadnajmi: Wait.. why would we ever set the ring to transparent instead of just `showRing=false` ? To get the glow?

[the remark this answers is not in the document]
tamasane: yep! please see the transparent rings demo (on changed background) in the description

packages/theming/android-theme/src/colorsTokens.ts
Saadnajmi: I'll ignore this and leave this for a future aligning talk. FYI @EJLayne

apps/android/package.json
Saadnajmi: Can you add that video to the PR description? Thanks!

packages/components/PersonaCoin/src/PersonaCoin.types.ts
```diff
@@ -1,9 +1,11 @@
 import type { IViewProps } from '@fluentui-react-native/adapters';
-import { ImageProps, ViewProps, ImageURISource, TextProps } from 'react-native';
+import { ImageProps, ViewProps, ImageURISource, TextProps, ColorValue } from 'react-native';
 import { IBackgroundColorTokens, IForegroundColorTokens } from '@fluentui-react-native/tokens';
 import { IRenderData } from '@uifabricshared/foundation-composable';
+import { Palette } from '@fluentui-react-native/theme-types';
 
 export const personaCoinName = 'RNFPersonaCoin';
+export type Color = keyof Palette | ColorValue;
```
Saadnajmi, line 8: I think you misunderstood what I meant. Can we not use this type at all, and on line70 and 71, just have those be of type `ColorValue` ?

packages/components/PersonaCoin/src/PersonaCoin.types.ts
Saadnajmi: Alright. 
A note: CXE is moving away from using "I" as a prefix for interfaces. so "RingConfig" is fine.

packages/components/PersonaCoin/src/PersonaCoin.types.ts
tamasane: Oops.. on it! Thanks for clarifying that.

Saadnajmi (approved):
Overall looks good, LGTM!

apps/android/package.json
Saadnajmi: Also the desktop screenshots. Maybe just have one screenshot per platform instead of showing off each individual prop, for ease of looking back at this PR in the future =)

PR #733, opened 2021-06-18
Wrap FluentUI DateTimePicker (iOS) for React Native

chripear:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
DatePickerManager.swift: The native side of the NativeDatePicker module, exposing the `present` method of FluentUI's DateTimePicker for iOS with flattened parameters.  This manager stores a single DatePickerDelegate instance to ensure that the callback block lives long enough to run.

DatePickerDelegate.swift: This delegate keeps track of the callback supplied by the TypeScript side, and invokes it with appropriate parameters when called by DateTimePicker.

MSFDatePickerManager.m: This file invokes the necessary React Native macros to export the `present` method to the TypeScript side, along with enums used for some of its parameters.

NativeDatePicker.tsx: This file puts the picker method and associated enums into more convenient TypeScript terms, using object unpacking on the `present` method to eliminate the need for a long list of positional parameters.  It also defines a convenience function to facilitate conversion of ISO 8601 strings supplied to the callback method into Date objects.

### Verification

NativeDatePickerTest.tsx: This includes test cases and usage examples for several configurations of NativeDatePicker, accessible through buttons.  It shows the resulting picked values in text fields.

NativeDatePickerTest

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/package.json
```diff
@@ -28,10 +28,12 @@
   "dependencies": {
     "@fluentui-react-native/android-theme": ">=0.3.1 <1.0.0",
     "@fluentui-react-native/apple-theme": ">=0.6.1 <1.0.0",
+    "@fluentui-react-native/button": ">=0.17.2 <1.0.0",
```
Saadnajmi, line 31: This should be included with line 45 here, "@fluentui/react-native"

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
```diff
@@ -0,0 +1,22 @@
...
+
+    public init(manager: DatePickerManager, picker: DateTimePicker, didPickBlock: @escaping RCTResponseSenderBlock) {
+        self.manager = manager
+        self.picker = picker
+        self.didPickBlock = didPickBlock
+    }
+
+    public func dateTimePicker(_ dateTimePicker: DateTimePicker, didPickStartDate startDate: Date, endDate: Date) {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions.insert(.withFractionalSeconds)
```
Saadnajmi, line 16: why do we need this?

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
```diff
@@ -0,0 +1,22 @@
...
+        self.didPickBlock([formatter.string(from: startDate), formatter.string(from: endDate)])
+        if let manager = manager {
+            manager.release(delegate: self)
```
Saadnajmi, line 19: Nit: Maybe give this a name that makes it not sound like it's built into swift? That's what I thought till I scrolled more down.

packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
```diff
@@ -0,0 +1,66 @@
...
+        dateTitle: String?,
+        dateSubtitle: String?,
+        timeTitle: String?,
+        timeSubtitle: String?,
+        callback: @escaping RCTResponseSenderBlock
+    ) {
+        DispatchQueue.main.async {
+            guard let viewController = UIApplication.shared.windows.filter({$0.isKeyWindow}).first?.rootViewController else {
+                fatalError("Unable to get a UIViewController from current shared application context.")
+            }
```
Saadnajmi, line 32: Maybe add a comment to the stack overflow you got this from / why you need to do this?

packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
```diff
@@ -0,0 +1,66 @@
...
+                startDate: startDate ?? Date(),
+                endDate: endDate,
+                datePickerType: datePickerType,
+                dateRangePresentation: dateRangePresentation,
+                titles: titles)
+        }
+    }
+
+    func release(delegate: DatePickerDelegate) {
+        precondition(delegate === lastDelegate, "DatePickerDelegate requested to be released should match the last one initialized.")
```
Saadnajmi, line 63: I'm curious, did you run into errors where this was not the case?

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
```diff
@@ -0,0 +1,83 @@
+import { NativeModules } from 'react-native';
+export const NativeDatePicker = NativeModules.MSFDatePickerManager;
+export const { MSFDateTimePickerMode, MSFDateTimePickerDatePickerType, MSFDateTimePickerDateRangePresentation } = NativeDatePicker.getConstants();
+
+type NativeDatePickerMode = keyof typeof MSFDateTimePickerMode;
+type NativeDatePickerType = keyof typeof MSFDateTimePickerDatePickerType;
+type NativeDatePickerDateRangePresentation = keyof typeof MSFDateTimePickerDateRangePresentation;
+
+interface DatePickerParameterObject {
```
Saadnajmi, line 9: Nit: Maybe we can we add a comment saying this is coming from the iOS DatePicker? I talked to one person who looked at NativeButton, try to port it to windows, and get confused where some of the very apple-specific props and declarations were coming from.

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+// We get date values back from the native side as strings in ISO 8601 format and UTC.
+// We want to immediately put them back into `Date` objects in local time.
+NativeDatePicker.parseISOString = (dateISOString: string): Date => {
+    if (dateISOString == null) {
+        return null
+    }
+    var dateParts: number[] = dateISOString.split(/\D+/).map(x => parseInt(x, 10));
+    dateParts[1]--; // Date.UTC's `month` arg is zero-based
+    var dateUTC = Date.UTC.apply(null, dateParts);
+    return new Date(dateUTC);
```
Saadnajmi, line 75: We can't do something like Date.parse()? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
chripear: Why do we need the function, or why do we need the ISO8601DateFormatter?  The former is the callback from the FluentUI-Apple class, which is how we communicate the picked values.  The latter is to convert a date into a string, since we can't pass a Date object directly back to RN.

chripear: If the question is why we need `.withFractionalSeconds`, it's because JS doesn't understand how to parse an ISO 8601 date without that part.  I did write a helper function to manually parse dates, but with this the built-in Date constructor can also do it (though I read it was ['strongly discouraged'](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) to use that)

packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
chripear: No, otherwise I would have used a different implementation.  I originally had the manager keep track of an array of delegates in case something caused us to need to keep track of more than one at a time. (Multi-window maybe?  The test app doesn't support it so IDK.)  Harrie convinced me to stick with a single value, but I put this line here so that we will know if some case breaks our assumption that there can be only one.

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
chripear: I decided to write an explicit parser because of this note in [Date() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date):

> Note: Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies.

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
chripear: Added in `30c439e`

packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
chripear: Added in `30c439e`.  I didn't see any other FURN code cite stackoverflow, so I just added some explanatory commentary.

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
chripear: Updated in `30c439e`

apps/fluent-tester/package.json
chripear: Updated in `30c439e`

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
Saadnajmi: Yeah sorry I meant the fractional seconds bit. 
The fact that you are discouraged from using the constructor is... unfortunate. I'm not a fan of doing manual parsing of dates, but it seems we might have to if we don't want to add a new dependency / use the base one? Wow, that's annoying

apps/fluent-tester/package.json
chripear: Just kidding

ERR! [11:48:57 AM] x Missing dependency on @fluentui-react-native/button
ERR! [11:48:57 AM] x -- /Users/chripear/git/fluentui-react-native/apps/fluent-tester/src/FluentTester/TestComponents/NativeDatePicker/NativeDatePickerTest.tsx

Saadnajmi: Huh, well that's weird. I guess it doesn't like transitive dependencies. I guess it's also better to be explicit 🤷🏾
Sorry about that...

packages/experimental/NativeDatePicker/ios/DatePickerDelegate.swift
chripear: I figure if somebody consuming this module wants to just use the constructor, I am ensuring it probably works fine.  But given the warning I wanted to have an example of doing it manually too.

Saadnajmi: Ah, this is example code. That is a good distinction to make, thank you!

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
Saadnajmi: I wonder if it's more standardized across React Native, but even then there's multiple javascript engines that could be used. I tend to be against manual date parsing (as I mentioned in a different comment) but I suppose this is the way to ensure it's always correct? Fingers crossed :D

Saadnajmi (changes requested):
Forgot:
Please add it to the nuspec so that we can consume it downstream =) 

Otherwise, looks good!

Saadnajmi (approved):

PR #735, opened 2021-06-22
Add support for RadioGroup on macOS

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, RadioGroup doesn't work on macOS because it has a dependency on FocusZone - a control that allows focus to navigate by groups. This change adds a native module to FocusZone to support RadioGroup on macOS. Since we already have RCTFocusZone implemented for Editor Pane in Devmain (CL#21476810 by @nakambo), let's copy and paste it to FURN where it's more sharable. 

### Verification
Tested RadioGroup on macOS and win32.

https://user-images.githubusercontent.com/67026167/122969732-fb34f000-d35a-11eb-8457-93aa73437f5c.mov

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Don't forget to add FocusZone to the NuGet package nuspec file

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -104,7 +104,7 @@ export const RadioButton = compose<IRadioButtonType>({
   settings,
   slots: {
     root: View,
-    button: { slotType: View, filter: filterViewProps },
+    button: { slotType: View },
```
Saadnajmi, line 107: Does filterViewProps break stuff on macOS? I think that's one of those things we never stood up on macOS that we probably should. I'm sure removing that breaks things for other platforms...

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -46,6 +46,76 @@ export const appleComponents = {
       variant: 'bodyStandard',
     },
   },
+  RadioButton: {
+    tokens: {
+      borderColor: 'rgba(0, 0, 0, 0.15)',
```
Saadnajmi, line 51: nit: separate to it's own variable since we re-use it or make it it's own token? There were some new tokens introduced in the Checkbox PR for an example of new tokens

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -46,6 +46,76 @@ export const appleComponents = {
...
+    root: {
+      accessible: false,
+      focusable: false,
+      style: {
+        alignItems: 'flex-start',
+      },
+    },
+    button: {
+      style: {
+        backgroundColor: 'rgba(255, 255, 255, 1)',
```
Saadnajmi, line 63: This isn't just 'white' ?

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -46,6 +46,76 @@ export const appleComponents = {
...
+        },
+        hovered: {
+          innerCircle: {
+            style: {
+              opacity: 0,
+            },
+          },
+        },
+        button: {
+          backgroundColor: 'rgba(75, 145, 247, 1)',
```
Saadnajmi, line 101: What color is this? This isn't just controlAccentColor?

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -46,6 +46,76 @@ export const appleComponents = {
...
+      backgroundColor: 'white',
+    },
+    root: {
+      accessible: false,
+      focusable: false,
+      style: {
+        alignItems: 'flex-start',
+      },
+    },
+    button: {
```
Saadnajmi, line 61: We don't need to add shadow-tops like shadowRadius to match the macOS system control?

Saadnajmi (changes requested):

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
...
+			BOOL stop = NO;
+			block(view, &stop);
+			if (stop)
+			{
+				break;
+			}
+		}
+		else
+		{
+			[queue addObjectsFromArray:[view subviews]];
```
nakambo, line 81: We should also grab Daniel's CL# 22682075 that deals with nested focusable views in a focusable view?

packages/components/FocusZone/macos/RCTFocusZoneManager.m
```diff
@@ -0,0 +1,55 @@
...
+
+RCT_CUSTOM_VIEW_PROPERTY(navigateAtEnd, NSString, RCTFocusZone)
+{
+	if ([json isEqualToString:@"NavigateStopAtEnds"])
+	{
+		[view setNavigateAtEnd:NavigateStopAtEnds];
+	}
+	else if ([json isEqualToString:@"NavigateWrap"] || [json isEqualToString:@"NavigateContinue"])
+	{
+		[view setNavigateAtEnd:NavigateStopAtEnds];
```
nakambo, line 42: we should retain that comment about this not being implemented yet

packages/components/RadioGroup/src/RadioButton.tsx
chiuam: No, filterViewProps doesn't break stuff on macOS. I had to remove it from button to allow the focus ring to land on the radio button. I tested my change on win32 and it was working fine, not sure about the other platforms though...

packages/theming/apple-theme/src/appleTheme.macos.ts
chiuam: Yes... I should probably create a token for that too

packages/components/RadioGroup/src/RadioButton.tsx
Saadnajmi: I think only teasing win32 is fine for this control, since it requires FocusZone which is only on these two platforms. 
Still, maybe we should reach out to someone in CXE to make sure it's OK to remove this?

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -26,7 +26,7 @@ export const RadioButton = compose<IRadioButtonType>({
 
     /* We don't want to call the user's onChange multiple times on the same selection. */
     const changeSelection = () => {
-      if (buttonKey != info.selectedKey) {
+      if (buttonKey != info.selectedKey && !disabled) {
```
samuelfreiberg, line 29: Just curious, in what scenario ill changeSelection() be called when the button is disabled? Doesn't disabling the button disallow focus/press on that button?

chiuam: I think this has to do with removing the filterViewProps from button. The button is selectable even when it's disabled...

packages/components/RadioGroup/src/RadioButton.tsx
samuelfreiberg: Is removing filterViewProps only necessary to add a feature on macOS? If so, it doesn't seem like a good pattern to remove stuff from one platform to add something to a different platform. 

Filtering out view props is a common pattern in all our controls for slots that are of type View (excluding the root). It feels strange to remove this here. Perhaps dividing these into .platform files would be better?

RadioButton.win32.tsx and RadioButton.macOS.tsx?

@ksiler @PPatBoyd @lenahong What are your thoughts on this?

samuelfreiberg:
Do we know the difference in functionality between RCTFocusZone and FURN's FocusZone? Are there any functionality gaps? Are they able to behave the exact same?

Saadnajmi:
> Do we know the difference in functionality between RCTFocusZone and FURN's FocusZone? Are there any functionality gaps? Are they able to behave the exact same?

@Nakambo implemented it with CL#21476810 in SD. His description:

> This is props(-and-name)-compatible with it's NetUI counterpart (which is currently in CR) -- therefore, it will function as a drop-in replacement to what Fluent expects from the JS side (at least for the capabilities needed for our scenario).
...
Testing:
	* Works as expected -- in terms of arrow keys and tab.
		-- we can't remember the "last" focused view in the zone. I've omitted this capability

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
...
+			{
+				action = FocusZoneActionShiftTab;
+				break;
+			}
+	}
+
+	return action;
+}
+
+- (NSView *)getNextView:(BOOL)next toFocusInRowOrColumn:(BOOL)horizontal
```
harrieshin, line 128: unless there is a reason why we need to have "get" in the prefix of this function name it doesn't seem very objc style function naming. applies for other functions too

harrieshin: "tofocusinroworcolumn" doesn't make much sense. or Hoirzontal doesn't really mean boolean

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
+#import "RCTFocusZone.h"
+
+// from <HIToolbox/Events.h>
+enum {
```
harrieshin, line 4: should the key code really be defined here?

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
...
+	FocusZoneActionNone,
+	FocusZoneActionTab,
+	FocusZoneActionShiftTab,
+	FocusZoneActionRightArrow,
+	FocusZoneActionLeftArrow,
+	FocusZoneActionDownArrow,
+	FocusZoneActionUpArrow,
+} FocusZoneAction;
+
+static const CGFloat FocusZoneBuffer = 3;
```
harrieshin, line 22: what is this for?

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
...
+	return viewToFocus;
+}
+
+- (NSView *)getViewToFocusInNextRow:(BOOL)next
+{
+	NSView *firstResponder = GetFirstResponder([self window]);
+	NSRect firstResponderRect = [firstResponder convertRect:[firstResponder bounds] toView:self];
+
+	__block NSView *viewToFocus = nil;
+	__block CGFloat closestDistance = 1.0e+30;
```
harrieshin, line 178: can you add comment why this is the default value?

packages/components/FocusZone/macos/RCTFocusZone.mm
```diff
@@ -0,0 +1,300 @@
...
+
+static NSView *GetFirstResponder(NSWindow *window)
+{
+	NSResponder *responder = [window firstResponder];
+	while (responder != nil && ![responder isKindOfClass:[NSView class]])
+		responder = [responder nextResponder];
+	return [responder isKindOfClass:[NSView class]] ? (NSView *)responder : nil;
+}
+
+static void EnumerateFocusableViews(NSView *root, void (^block)(NSView *, BOOL *))
```
harrieshin, line 60: maybe typedef the block so reader know what this block is actually doing?

packages/components/FocusZone/macos/RCTFocusZone.mm
harrieshin: does addobjectsfromarray accept nil? or does it throw exception if [view subviews] is nil?

packages/components/RadioGroup/src/RadioButton.tsx
ksiler: @chiuam how did `filterViewProps` prevented the focus ring from landing on the radio button? I'd like to understand that more. 

I agree that we shouldn't be removing stuff just to add something to one platform. Even though it doesn't break anything now, we might want to use platform specific props in the future. 

@Saadnajmi / @chiuam  Is it possible to implement a filterViewProps for macOS? How hard would that be? Seems like react-native-macos has View, Text, and Image components here: https://github.com/microsoft/react-native-macos/tree/master/Libraries Is assume this is where you'd pull the platform specific props from?

Saadnajmi: Would that just be making an `adapters.macos.ts` file here? https://github.com/microsoft/fluentui-react-native/tree/907fea51098ba59a8235128e17fb82a327b7adb2/packages/utils/adapters/src

It might help if we had a simple overview of what filterviewprops/adapters was doing or why it's necessary. I assume it's because the props passed to the core components (View, Text, Image) can be different on each platform (especially out of tree platforms), and so this... does something to help that? Does it exclude props? How do we stop it from excluding a prop that may be necessary, like `enableFocusRing`? That's one of the things I've been confused about :)

ksiler: Yeah I believe so.

Your assumption is my understanding as well. The adapter file allows you to pull in platform specific props and define them through the masks (for View, Text, and Image) and it can exclude props if you desire. I know for win32, we exclude some android and ios specific props and include win32 specific props. Does that answer your question?

It sounds like `enableFocusRing` is a macOS specific prop? Where does it come from/where is it defined?

chiuam:
Closing this as we went with the native route for radio button. https://github.com/microsoft/fluentui-react-native/pull/838

PR #736, opened 2021-06-22
Add a NuGet publish dry run job to PR pipeline

Saadnajmi:
### Platforms Impacted
- [] iOS
- [] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

A common point of failure in the Publish pipeline is that the NuGet package fails to publish. This change adds a check that does a dry run of publishing the NuGet package, AKA, it does everything but the final `nuget push`. 

Pros:
- We can catch NuGet failures earlier

Cons:
- This extra check takes like an hour where the others take like 15 minutes :( 

I originally had decided against adding the check because it takes so long, but I don't think it'll affect the developer inner loop because people rarely check in code that quickly it seems.

### Verification

Check the checks of this PR :D 

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/azure-pipelines.yml
```diff
@@ -146,3 +146,43 @@ jobs:
           publishJUnitResults: false
           javaHomeOption: 'JDKVersion'
           sonarQubeRunAnalysis: false
+
+  - job: NuGetPublishDryRun
```
HeyImChris, line 150: How often is the publish pipeline breaking due to someone's changes? I'm wondering if a failed publish merits blocking checkins if the publishes failing are commonly due to external factors

Saadnajmi: Usually one of us forgets to update the .nuspec when we make a new native module, or we might put the paths in the nuspec wrong. At least one time the token to publish expired, which you know of because you helped fix it.
Once it was because we tried to integrate rex-kit and changed the bundles from .jsbundle to .bundle.
Most of the time it's Apple Engineers changing something on the apple side, not external.


Also, I would be happy to have this just *publish* the NuGet too because having a NuGet from a WIP PR is helping for testing downstream. I think that would block others working on JS only stuff if the NuGet pipeline broke again because of expired certs or the like though.

HeyImChris (approved):
How was this tested that the dry run option still catches problems like we've hit in the past?

Saadnajmi:
> How was this tested that the dry run option still catches problems like we've hit in the past?

Tested.... :D 

I can try making a commit that changes the nuspec to obviously fail, and see if that breaks CI =) Like, changing .jsbundle to .bundle or adding a typo to a static lib path

Saadnajmi:
I decided to keep the change of having ".jsbundle" instead of ".bundle" in our NuGet package because I think that's what will be better downstream in devmain. It's not currently used so no effect.

PR #738, opened 2021-06-24
Fixing Checkbox Focus + Checkbox Folder Structure

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes

1. A Recent PR caused a regression for Focus in the Checkbox. Specifically, **focusable: true** was removed from the .settings file
2. I'm reverting _Checkbox.settings.ts_ back to the default values for Win32/Windows rather than Web
3. Removing platform-specific Checkbox.settings files for Win32/Windows, and instead adding platform-specific files for Android/Mac

### Verification

Manual testing

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Checkbox/src/Checkbox.settings.ios.ts
```diff
@@ -0,0 +1 @@
+export * from './Checkbox.settings.android';
```
Saadnajmi, line 1: I don't think this is correct.
Checkbox.settings.windows.ts was created because windows/win32 doesn't support SVG's. It looks like you renamed settings.windows to settings.android and are using it for iOS/macOS. That means the settings for non-SVG supporting platforms are being used for SVG supporting platforms. 

Also, I'd rather not have one platforms settings be imported into another platforms. I'd rather have some common name like "settings / settingsWithoutSVG" that each imports if that's the actual distinction.

packages/components/Checkbox/src/Checkbox.settings.ts
```diff
@@ -3,14 +3,14 @@ import { checkboxName, ICheckboxType } from './Checkbox.types';
 
 export const checkboxSelectActionLabel = 'Toggle the Checkbox';
 
+/* Default values for Win32/Windows */
```
Saadnajmi, line 6: If these are the windows/win32 default, would it be more proper to name this "settings.windows.ts" and have "settings.win32.ts" import from there?

Saadnajmi (changes requested):

packages/components/Checkbox/src/Checkbox.settings.ios.ts
samuelfreiberg: I didn't rename the .windows one to .android. I took the base Checkbox.settings.ts from before, and made that into Checkbox.settings.android.ts. I then made the base Checkbox.setting.ts into the _normal_ Win32/Windows .settings file from before [this PR](https://github.com/microsoft/fluentui-react-native/pull/716/)

It's all a bit confusing, but let me try to simplify it some more.

Checkbox.settings.ts should have never been changed to be the default settings files for android/iOS/macOS/Web (which it was in the previous PR). Checkbox.setting.ts should be the default file for Win32/Windows, and the other platforms (if needed) should have separate platform-specific files. 

Also, I understand your sentiment of not having one platform import from another platform. However, if we do combine the Android/MacOS/iOS into one file named something like _Checkbox.settingsWithSVG_, what happens if there comes a time where we actually need to make a change to only iOS and not Android?

packages/components/Checkbox/src/Checkbox.settings.ts
samuelfreiberg: I understand this sentiment, however, I am trying to follow patterns done everywhere else in the repo. All the base files in this repo (whether it be each controls' .tsx file, .settings file, or .types file) are all Win32 specific. We haven't added .win32 specific files for any of these controls (to my knowledge).

Maybe, however, because we're introducing other platform-specific files in this folder, it would make sense to add .win32 specific files as well.

I think this should be a discussion we hold at an in-person meeting, as it will pave the repo structure moving forward as more platform-specific files come into play

packages/components/Checkbox/src/Checkbox.settings.ios.ts
Saadnajmi: I see, sorry I got confused looking at what git _thought_ you did in the headers for the file changes.
Come that time, I assume we would either do a `if Platform === 'ios'` or `Platform.Select({...})` check, or separate out the types even further. 
This is probably valid for talking about in person since this and the other thread are related =)

samuelfreiberg: No problem, I just noticed that and it's definitely confusing :)

packages/components/Checkbox/src/Checkbox.settings.ts
samuelfreiberg: @Saadnajmi  Some extra info: I didn't actually change anything here. All I did was revert a couple of the things Tushar did in the last PR. The only thing I added was this comment to clarify this is Win32.

samuelfreiberg: @Saadnajmi I'm writing up a post to discuss the folder structure and what not, I just want to make sure we get this change in before tomorrow (the fork is on sunday and GDPR is in production and uses our Checkbox).

We need to have the Checkbox focus fix in asap for GDPR.

samuelfreiberg: And I want to make sure the broken version of Checkbox doesn't get pulled into Devmain for the fork

Saadnajmi: I see. How about we check this in, with a followup conversation on naming so we don't block fixing this downstream?
If this is truly a revert, then I suppose the risk is minimal.

Saadnajmi (approved):

packages/components/Checkbox/src/Checkbox.settings.ts
samuelfreiberg: Sounds good, I will make a task to discuss it! Thanks

Issue #739, opened 2021-06-24
Win32 specific settings files should be moved to settings.win32.ts


Saadnajmi:
#716 has issues on win32 breaking focus and issues on macOS converting colors. 
We should either back it out or make a followup PR ASAP, given that the former (win32) component is used in production.

Saadnajmi:
Looks like #738 is addressing the windows side.

Saadnajmi:
Windows side fixed with #738 

I'll keep this issue open for the open issue about platform specific settings

Saadnajmi:
Renaming bug to reflect change.

PR #742, opened 2021-06-24
Remove Icon change that breaks Icon on macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This should resolve the macOS side of #739 

I am undoing code added in #716 that would otherwise break macOS. A longer discussion of why this is bad is [here](https://github.com/microsoft/fluentui-react-native/pull/716/files#r655534695).

Tl;Dr: 
`processColor` is a private API that should not be called directly. Furthermore, we are converting an object (ColorValue) to a string and then calling processColor() on it, which is incorrect (ColorValue might be a semantic color, not a string).
Let's just remove the code.

I'm also adding a really simple README to Checkbox to address a comment by @harrieshin that clients should probably not use checkbox on iOS and use a switch instead.

### Verification

This is reverting code that was added, so it should be safe. I'll still click through the macOS app to make sure things work as expected.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Checkbox/README.md
+++ b/packages/components/Checkbox/README.md
@@ -1,0 +1,9 @@
+# Checkbox
+
+A cross-platform Checkbox component using the Fluent Design System.
+
+Supported Platforms: Android, iOS, macOS, web, windows, win32
+
+Note that on iOS, the more common control to use is a Switch
+https://developer.apple.com/design/human-interface-guidelines/ios/controls/switches/
+https://reactnative.dev/docs/switch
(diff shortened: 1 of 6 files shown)


harrieshin (approved):

packages/components/Checkbox/README.md
```diff
@@ -0,0 +1,9 @@
+# Checkbox
+
+A cross-platform Checkbox component using the Fluent Design System.
```
rurikoaraki, line 3: Yay!

rurikoaraki (approved):

PR #744, opened 2021-06-25
chore: bump @rnx-kit/cli and set up @rnx-kit/dep-check

tido64:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Bumps `@rnx-kit/cli` to 0.5.27, which includes `@rnx-kit/dep-check`. Configures test apps so their react/react-native dependencies are aligned.

Additional changes were also made in order to make the Android test app boot, like adding `hermes-engine` (which is default by now), and bumping Gradle to a more recent version.

Note that `hermes-engine` had to be added to depcheck's `ignoreMatches` because it is only referenced in Gradle build scripts.

### Verification

CI should pass.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


tido64:
Noticed this weirdness here: https://github.com/microsoft/fluentui-react-native/blob/6843bd3785e2b2bfe47b8d645a3afde73af92bb8/apps/windows/package.json#L19-L20

If this is not intentional, I can try and re-align it later.

yarn.lock
```diff
@@ -14178,16 +14218,14 @@ react-native-svg@^12.1.1:
     css-select "^2.1.0"
     css-tree "^1.0.0-alpha.39"
 
-react-native-test-app@^0.3.2:
-  version "0.3.13"
-  resolved "https://registry.yarnpkg.com/react-native-test-app/-/react-native-test-app-0.3.13.tgz#ec4cd2d35110b4dbb079b4df91fb8127af359bb6"
-  integrity sha512-b+QoP3Wt+2XKvwQx1l6OqF17KhBQOvFoj+WVGNzFs+gVDumNUInFdzjKXo4G3ZTXJqTmGoBKYRpIuWEKoNJK9Q==
+react-native-test-app@^0.5.9:
+  version "0.5.14"
```
tido64, line 14222: Not using the latest 0.5.x due to a crash on startup. The fix for that is here: https://github.com/microsoft/react-native-test-app/pull/370

Saadnajmi:
> Noticed this weirdness here:
> 
> https://github.com/microsoft/fluentui-react-native/blob/6843bd3785e2b2bfe47b8d645a3afde73af92bb8/apps/windows/package.json#L19-L20
> 
> If this is not intentional, I can try and re-align it later.

I think when we updated to 0.63, we didn't update the windows test app because either:
A) I didn't know where to get the newest template test app from without using RNTA, and moving to RNTA broke ci checks
B) Updating rnw broke ci checks

Pretty sure it was just A

apps/android/package.json
```diff
@@ -24,34 +24,43 @@
...
     "@react-native-community/eslint-config": "^1.1.0",
-    "@rnx-kit/cli": "^0.0.3",
+    "@rnx-kit/cli": "^0.5.26",
     "@uifabricshared/build-native": "^0.1.1",
     "@uifabricshared/eslint-config-rules": "^0.1.1",
     "babel-jest": "^24.9.0",
     "jest": "^24.9.0",
     "metro-react-native-babel-preset": "^0.59.0",
     "react-native-svg-transformer": "^0.14.3",
+    "react-native-test-app": "^0.5.9",
```
Saadnajmi, line 43: Not ^0.6.3?

tido64:
@Saadnajmi Sounds like we should try and bump it in #552.

apps/android/package.json
```diff
@@ -24,40 +24,50 @@
...
   },
   "depcheck": {
     "ignoreMatches": [
-      "react-native-test-app",
+      "@react-native-community/slider",
       "@react-native-picker/picker",
-      "@react-native-community/slider"
+      "hermes-engine",
+      "react-native-test-app"
     ]
```
Saadnajmi, line 71: Could you explain the extra additions in the PR description?

apps/android/src/build.gradle
```diff
@@ -1,13 +1,15 @@
 buildscript {
```
Saadnajmi, line 1: For my own knowledge, all of the `build.gradle`, `gradle/*`, `Bradley`, etc.. files  at `apps/android/src` are not something that RNTA provides and something we have to bring ourselves, correct?

apps/macos/package.json
```diff
@@ -21,38 +21,43 @@
...
         "macos"
       ],
       "entryPath": "index.js"
-    }
+    },
+    "capabilities": [
+      "core-macos",
+      "react",
+      "test-app"
+    ]
```
Saadnajmi, line 60: purposefully not bringing hermes to macOS and iOS?

package.json
```diff
@@ -24,7 +24,7 @@
     "change": "beachball change",
     "check-for-changed-files": "cd scripts && yarn fluentui-scripts check-for-modified-files",
     "checkchange": "beachball check --changehint \"Run 'yarn change' to generate a change file\"",
-    "depcheck": "lage depcheck",
+    "depcheck": "rnx-dep-check && lage depcheck",
```
Saadnajmi, line 27: Why do we need. both? Are we not quite wanting to change all the subrepo depchecks to rnx-depcheck?

apps/android/package.json
tido64: `@rnx-kit/dep-check` currently recommends `^0.5.9`. I can bump `react-native-test-app` in rnx-kit, then do a second pass here.

apps/android/package.json
tido64: Done.

apps/android/src/build.gradle
tido64: That is correct. @ArazAbishov probably knows the why, but my understanding is that it's just how Gradle works. I don't know if there's much we can do about that.

apps/macos/package.json
tido64: If we want to bring in Hermes, I would do it in a separate PR as it's more controversial.

Saadnajmi: Just making sure =) 
We purposefully chose not to use Hermes on the apple platforms just yet. Also, Hermes on arm64 Macs won't work till RN 0.65...

package.json
tido64: `@rnx-kit/dep-check` is more about managing dependencies and aligning them within and across repos. You can read more about it here: https://github.com/microsoft/rnx-kit/blob/main/packages/dep-check/DESIGN.md

Whereas depcheck's purpose is to find unused or missing dependencies. They don't overlap in functionality.

tido64: The name clash is a bit unfortunate. If you have better suggestions, I'm all ears.

Saadnajmi: Thanks for that doc. That was very helpful in explaining what the PR is trying to achieve. Definitely a problem I've run into before with this repo =)

`dep-align`? I'm sure you've already thought of that though

afoxman (approved):

Saadnajmi (approved):
We should perhaps update the README's of each test app to say that they are managed by rex-kit and not just RNTA, with a link to this (https://github.com/microsoft/rnx-kit/blob/main/packages/dep-check/DESIGN.md) to explain what the `capabilities` and `rnx-kit` fields are for. We're having some new people (specifically in the UWP space) onboard into FURN so the more info the better!


Otherwise, looks good =)

tido64:
> We should perhaps update the README's of each test app to say that they are managed by rex-kit and not just RNTA, with a link to this (https://github.com/microsoft/rnx-kit/blob/main/packages/dep-check/DESIGN.md) to explain what the `capabilities` and `rnx-kit` fields are for. We're having some new people (specifically in the UWP space) onboard into FURN so the more info the better!

Good callout. I'll prep a separate PR since this one is getting a bit big.

PR #753, opened 2021-06-30
[A11y]: Fix Radio Button not announcing position in Radio Group on win32

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change addresses bug [AB#5129857](https://office.visualstudio.com/e853b87d-318c-4879-bedc-5855f3483b54/_workitems/edit/5129857) which states:

> When navigating through a RadioGroup, the screen reader should read off the position of the currently focused RadioButton (along with all the other content information). For example, if you have a RadioGroup consisting of:
> 
> RadioButton A
> RadioButton B
> RadioButton C
> RadioButton D
> 
> When focused on RadioButton B, the narrator's last statement should be "RadioButton 2 of 4" (something along those lines). 

To fix this, this change adds a `buttonKeys` array to RadioGroupContext which stores all the buttonKeys in radio group. Each radio button retrieves the size of array to pass to ViewWin32's accessibilitySetSize and the position their buttonKey appears in the array to pass to ViewWin32's accessibilityPositionInSet.

### Verification

Tested manually and inspected  elements with A11yInsights.

Before:

Screen readers like Narrator did not announce the position of a radio button in a radio group when a user navigates to it.

After:

Screen readers like Narrator announce the position of a radio button in a radio group when as user navigates to it.

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
@@ -35,9 +35,8 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
       <Separator />
       <RadioGroup label="Controlled RadioGroup" selectedKey={selectedKey} onChange={onChange2}>
         <RadioButton content="Option A" buttonKey="A" ariaLabel="Test Aria Label" />
-        {/* The values for ariaPosInSet and ariaSetSize show someone can override the auto-generated values */}
-        <RadioButton content="Option B" buttonKey="B" ariaPosInSet={57} />
-        <RadioButton content="Option C" buttonKey="C" ariaSetSize={18} />
+        <RadioButton content="Option B" buttonKey="B" />
+        <RadioButton content="Option C" buttonKey="C" />
         <RadioButton content="Option D" buttonKey="D" />
       </RadioGroup>
     </View>
--- a/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
@@ -35,6 +35,7 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
       <Separator />
       <RadioGroup label="Controlled RadioGroup" selectedKey={selectedKey} onChange={onChange2}>
         <RadioButton content="Option A" buttonKey="A" ariaLabel="Test Aria Label" />
+        {/* The values for ariaPosInSet and ariaSetSize show someone can override the auto-generated values */}
         <RadioButton content="Option B" buttonKey="B" ariaPosInSet={57} />
         <RadioButton content="Option C" buttonKey="C" ariaSetSize={18} />
         <RadioButton content="Option D" buttonKey="D" />
--- a/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
@@ -35,8 +35,8 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
       <Separator />
       <RadioGroup label="Controlled RadioGroup" selectedKey={selectedKey} onChange={onChange2}>
         <RadioButton content="Option A" buttonKey="A" ariaLabel="Test Aria Label" />
-        <RadioButton content="Option B" buttonKey="B" />
-        <RadioButton content="Option C" buttonKey="C" />
+        <RadioButton content="Option B" buttonKey="B" ariaPosInSet={57} />
+        <RadioButton content="Option C" buttonKey="C" ariaSetSize={18} />
         <RadioButton content="Option D" buttonKey="D" />
       </RadioGroup>
     </View>
--- a/packages/components/RadioGroup/src/RadioButton.tsx
+++ b/packages/components/RadioGroup/src/RadioButton.tsx
@@ -17,7 +17,7 @@ export const RadioButton = compose<IRadioButtonType>({
   displayName: radioButtonName,
 
   usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
-    const { content, buttonKey, disabled, ariaLabel, componentRef = React.useRef(null), ...rest } = userProps;
+    const { content, buttonKey, disabled, ariaLabel, componentRef = React.useRef(null), ariaSetSize, ariaPosInSet, ...rest } = userProps;
 
     // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
     const info = React.useContext(RadioGroupContext);
@@ -81,6 +81,8 @@ export const RadioButton = compose<IRadioButtonType>({
         accessibilityLabel: ariaLabel ? ariaLabel : content,
         accessibilityState: { disabled: state.disabled, selected: state.selected },
         accessibilityActions: [{ name: 'Select', label: radioButtonSelectActionLabel }],
+        accessibilityPositionInSet: ariaPosInSet,
+        accessibilitySetSize: ariaSetSize,
         onAccessibilityAction: onAccessibilityAction,
       },
       content: { children: content },
--- a/packages/components/RadioGroup/src/RadioButton.tsx
+++ b/packages/components/RadioGroup/src/RadioButton.tsx
@@ -17,7 +17,7 @@ export const RadioButton = compose<IRadioButtonType>({
   displayName: radioButtonName,
 
   usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
-    const { content, buttonKey, disabled, ariaLabel, componentRef = React.useRef(null), ariaPosInSet, ariaSetSize, ...rest } = userProps;
+    const { content, buttonKey, disabled, ariaLabel, componentRef = React.useRef(null), ...rest } = userProps;
 
     // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
     const info = React.useContext(RadioGroupContext);
@@ -81,8 +81,8 @@ export const RadioButton = compose<IRadioButtonType>({
         accessibilityLabel: ariaLabel ? ariaLabel : content,
         accessibilityState: { disabled: state.disabled, selected: state.selected },
         accessibilityActions: [{ name: 'Select', label: radioButtonSelectActionLabel }],
-        accessibilityPositionInSet: ariaPosInSet,
-        accessibilitySetSize: ariaSetSize,
+        accessibilityPositionInSet: info.buttonKeys.findIndex((x) => x == buttonKey) + 1,
+        accessibilitySetSize: info.buttonKeys.length,
         onAccessibilityAction: onAccessibilityAction,
       },
       content: { children: content },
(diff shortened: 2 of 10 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
```diff
@@ -35,8 +35,8 @@ const basicRadioGroup: React.FunctionComponent<{}> = () => {
       <Separator />
       <RadioGroup label="Controlled RadioGroup" selectedKey={selectedKey} onChange={onChange2}>
         <RadioButton content="Option A" buttonKey="A" ariaLabel="Test Aria Label" />
-        <RadioButton content="Option B" buttonKey="B" />
-        <RadioButton content="Option C" buttonKey="C" />
+        <RadioButton content="Option B" buttonKey="B" ariaPosInSet={57} />
+        <RadioButton content="Option C" buttonKey="C" ariaSetSize={18} />
```
ksiler, line 39: While these values are inaccurate, this test aims to show someone can override the auto-generated values. A screen reader reading this would say "57 out of 4" or "3 out of 18".

packages/components/RadioGroup/src/RadioButton.types.ts
```diff
@@ -31,6 +31,16 @@ export interface IRadioButtonProps extends IPressableProps {
...
+  /*
+   ** Defines the current radio button's position in the radio group. If not provided, this will be auto-generated.
+   */
+  ariaPosInSet?: number;
+
+  /*
+   ** Defines the number of radio buttons in the group. If not provided, this will be auto-generated.
+   */
+  ariaSetSize?: number;
+
```
ksiler, line 43: I've wondered whether or not we should make these properties available on radio button since I can't think of a scenario where someone would want to override the generated value... but I added them in part because I wasn't sure how else to get the properties passed down from radio group to radio button. If someone knows how I could pass them down without adding them as properties on radio button, let me know! I think that'd be ideal but maybe there is a case for someone to override the values...

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
rurikoaraki: Could you add a comment about that in the code so that we understand what it's there for in the future?

packages/components/RadioGroup/src/RadioButton.types.ts
ksiler: I've considered adding these properties to RadioGroupContext but that doesn't work for ariaPosInSet since each radio button's position is unique and isn't shared data like ariaSetSize would be.

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
ksiler: Added one

packages/components/RadioGroup/src/RadioGroup.tsx
```diff
@@ -76,6 +76,26 @@ export const RadioGroup = compose<IRadioGroupType>({
...
+      /* eslint-disable @typescript-eslint/ban-ts-ignore */
+      // @ts-ignore - TODO, fix typing error
+      children = React.Children.map(children, (child: React.ReactChild, index: number) => {
+        if (React.isValidElement(child)) {
+          const childProps = child.props;
+          const extraProps = {
+            ariaSetSize: childProps.ariaSetSize ?? React.Children.count(children),
+            ariaPosInSet: childProps.ariaPosInSet ?? index + 1,
+          };
+          return React.cloneElement(child, {
```
rurikoaraki, line 90: Hmm... A little weary seeing "clone" and building a new prop object for every render. It seems like this could result in excessive rendering?

Looks like you could just pass in the "extraProps" part according to [their documentation](https://reactjs.org/docs/react-api.html#cloneelement): "The resulting element will have the original element’s props with the new props merged in shallowly." That might also help with render checks as posInSet and setSize are both numbers, so they should pass render checks, but the rest of the props are preserved.

rurikoaraki (changes requested):
I'd like to know a bit more about how this change effects rendering of the RadioButton, but other parts of the change look good.

packages/components/RadioGroup/src/RadioGroup.tsx
samuelfreiberg: I think another strategy could be this:

Add `ariaSetSize: int` to RadioGroup's context, and set the value around line 81. Then, the children can just grab the value from RadioGroup's context and set their prop on the value.

I would also add something like: `ariaPosInSet: [ {radioButtonKey: string, ariaPosInSet: int} ]` to RadioGroup's context. Then, when iterating through the children using the map function, you can create a local array of these values. After .map is complete, you can set the context to the local array value. Using this, RadioButton's can grab this array, check for matching buttonKey, and set their posInSet to the corresponding value.

packages/components/RadioGroup/src/RadioButton.types.ts
samuelfreiberg: For the RadioGroup context, could this work:

Add ariaSetSize: int to RadioGroup's context, and set the value around line 81. Then, the children can just grab the value from RadioGroup's context and set their prop on the value.

I would also add something like: ariaPosInSet: [ {radioButtonKey: string, ariaPosInSet: int} ] to RadioGroup's context. Then, when iterating through the children using the map function, you can create a local array of these values. After .map is complete, you can set the context to the local array value. Using this, RadioButton's can grab this array, check for matching buttonKey, and set their posInSet to the corresponding value.

packages/components/RadioGroup/src/RadioGroup.tsx
ksiler: That's a good point. I've been a little stumped on this because I don't know how to retrieve the children of RadioGroup outside of the render portion of this file. Ideally, I'd like to create a function that's defined in RadioGroup (maybe using [React.useEffect](https://reactjs.org/docs/hooks-reference.html#useeffect)) that will only be called when the number of children change and either passes down the data to the children as props or share the data through RadioGroupContext. But it appears we only know how many children there are in the render portion. And whatever I place there will have to run every render 😕

ksiler: I'll try to work with Sammy's suggestion and maybe add an array of the children in RadioGroupContext that will update with each render. The radio buttons can then pull the info they need from the array.

rurikoaraki: I think if you remove the "childProps" part in cloneElement the children won't rerender excessively (at least, according to the documentation) but you'll still add the props you need to the children.

But yeah this is rough since the RadioGroup gets RadioButtons as children directly. I kind of wonder if/how the Picker handles this scenario since I think it has a similar structure.

packages/components/RadioGroup/src/RadioButton.types.ts
ksiler: Thanks for the suggestion! I ended up introducing a `buttonKeys` array to RadioGroupContext which stores all the buttonKeys in radio group. From each child, I can retrieve the size of the set by inspecting the size of the array and figure out the position in the set by finding the index of the matching buttonKey. Now I no longer need these props.

packages/components/RadioGroup/src/RadioGroup.tsx
ksiler: I ended up adding a `buttonKeys` array to RadioGroupContext and altered this if statement to retrieve the buttonKey for each child and place it into that array. I can then grab the info I need from the array from each child, eliminating the need to clone child elements and pass down props.

packages/components/RadioGroup/src/RadioButton.tsx
```diff
@@ -81,6 +81,8 @@ export const RadioButton = compose<IRadioButtonType>({
         accessibilityLabel: ariaLabel ? ariaLabel : content,
         accessibilityState: { disabled: state.disabled, selected: state.selected },
         accessibilityActions: [{ name: 'Select', label: radioButtonSelectActionLabel }],
+        accessibilityPositionInSet: info.buttonKeys.findIndex((x) => x == buttonKey) + 1,
```
samuelfreiberg, line 84: Playing devils advocate - what happens if findIndex can't find the buttonKey for some reason? What does this return?

packages/components/RadioGroup/src/RadioButton.types.ts
samuelfreiberg: Much better idea - great job!

packages/components/RadioGroup/src/RadioGroup.tsx
```diff
@@ -76,6 +77,17 @@ export const RadioGroup = compose<IRadioGroupType>({
       return null;
     }
 
+    // Populate the buttonKeys array
+    if (children) {
+      /* eslint-disable @typescript-eslint/ban-ts-ignore */
+      // @ts-ignore - TODO, fix typing error
+      renderData.state.context.buttonKeys = React.Children.map(children, (child: React.ReactChild) => {
+        if (React.isValidElement(child)) {
+          return child.props.buttonKey;
```
samuelfreiberg, line 86: What if a user has something other than a RadioButton as a child to RadioGroup? What would the behavior be? Just thinking about edge cases

packages/components/RadioGroup/src/RadioButton.tsx
ksiler: If findIndex can't find an element that satisfies the testing function, it returns -1. If we add one to that we get zero which is the default value for the native PositionInSet property. So if for some reason it can't find the button key for a radio button, it will assign it zero and the screen reader won't announce anything about the position or size of set for that radio button. But I don't foresee that happening since we pull the button keys from each child directly during each render.

packages/components/RadioGroup/src/RadioGroup.tsx
ksiler: Good call. I tested that scenario by adding a Button as a child of Radio Group in the test app (in between option "B" and "C"). It doesn't seem to effect desired result. When we map through each child and return `child.props.buttonKey`, it won't include any null or undefined values:

`map<T, C>(children: C | C[], fn: (child: C, index: number) => T):
            C extends null | undefined ? C : Array<Exclude<T, boolean | null | undefined>>;`
            
In this case, button doesn't have a prop called buttonKey so it doesn't include the value in the array buttonKeys. I assume it will do the same with any other child that's not a Radio Button (since it wouldn't have a prop called buttonKey either).

ksiler: And even if it did have a prop called buttonKey... I think that scenario would be really rare and kind of conflicting. Why would we have something other than RadioButton as a child of RadioGroup that also uses button keys 🤔 if we wanted to be super safe here, we could check the child to make sure it's a RadioButton before returning it's button key.

rurikoaraki: Imo if it's only going to affect the accessibility props I don't think it's worth the effort to be really robust if it's unexpected that this case would happen. Until it's a legitimate scenario (or somehow it ends up being a common issue), I think we can let it be.

rurikoaraki (approved):
Looks good!

packages/components/RadioGroup/src/RadioGroup.tsx
samuelfreiberg: Sounds good with me! Thanks for testing

samuelfreiberg (approved):

PR #756, opened 2021-07-01
Add Expander native component (WinUI 2.6)

amdingler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Adding a native windows Expander control (WinUI 2.6) to FURN. 
- packages/experimental/Expander/windows: Contains native module code for the Expander control
- packages/experimental/Expander/src: Contains code to compose the native view into a component

Notable design choices:
- packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.idl: ExpanderView inherits from a ContentPresenter, and the Expander control is displayed from within the ContentPresenter control. This is because we are unable to inherit from the Expander control directly.

Future work:
- Resolve layout issues between Yoga and XAML layout systems within the RNW framework. Conflicting layout systems caused the following issues and work arounds in the Expander implementation:
   - Expander requires a fixed collapsed and expanded height
   - height must be exposed as props to allow the proper collapsed/expanded height changes. If the user tries to use the height prop, it is overridden. Can potentially fix this using token states to update the Expander height.
   - Expander does not account for height changes caused by window resizing
   - User cannot interact with non-native components within the header. In this scenario, the Expander will expand instead.
- Remove implicit ordering of Expander header and content. The user should be able to specify which child element is the header and which is the content. 

### Verification

Tested on a [React Native for Windows v0.64.x tester app](https://github.com/amdingler/expander-tester). Expander was tested for light mode, dark mode, and keyboard accessibility.

**Basic Expander**

https://user-images.githubusercontent.com/47195059/127239274-c1378b17-2429-410e-9afd-d769772e7c35.mp4

**Full Expander Functionality**

https://user-images.githubusercontent.com/47195059/127241306-3870077d-9a22-4f2c-85e7-3cdc2d329d5c.mp4

**Dark Mode**

Dark_Mode_Expander

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Expander/FluentUIReactNativeButton.podspec
```diff
@@ -0,0 +1,27 @@
...
+  s.ios.deployment_target = "13.0"
+  s.ios.source_files      = "ios/*.{swift,h,m}"
+  s.ios.dependency 'MicrosoftFluentUI', '~> 0.2.2'
+
+  s.osx.deployment_target = "10.14"
+  s.osx.source_files      = "macos/*.{swift,h,m}"
+  s.osx.dependency 'MicrosoftFluentUI', '~> 0.2.2'
+
+
+end
```
Saadnajmi, line 27: This is Apple specific and you can delete this.

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,32 @@
...
+  tokens: [{}, expanderName],
+  slotProps: {
+    root: buildProps(
+      (tokens, theme) => ({
+        style: {
+          height: 500,
+          width: 500,
+          margin: 8,
+        },
+        accentColor: theme.colors.accentButtonBackground,
```
Saadnajmi, line 21: Is this an actual thing you can style?

amdingler: I believe we're leaning towards exposing most of WinUI's custom styling options, so I will probably remove the accentColor token. Otherwise, this value would be used to theme a few default styling options similar to the NativeButton.

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,43 @@
...
+      }),
+      [],
+    ),
+  },
+  slots: { root: ExpanderComponent },
+  render: (userProps: ExpanderProps, useSlots: UseSlots<ExpanderType>) => {
+    const Root = useSlots(userProps).root;
+    const [expandedState, setExpandedState] = React.useState(userProps.expanded);
+    return (rest: ExpanderViewProps, ...children: React.ReactNode[]) => <Root
+      style={{
```
amdingler, line 29: @Saadnajmi Adding style and onChange here causes the pipeline to fail because they aren't in tokens/props. I don't want to expose these to users, so is there somewhere else they should be added to resolve these errors? If I remember correctly, utilizing buildUseStyling states could be an alternative to this?

Saadnajmi: After looking through your code, I think I understand your problem. You have props you need to pass to your Root Slot. These props don't depend on tokens, so you can't really make them in slotProps, which is where I would have suggested doing it to not expose them to the user. I haven't personally used the states part of compose yet, so I'll have to do some research of my own to investigate what a proper solution is.

amdingler: Would making collapsedHeight and expandedHeight tokensThatAreAlsoProps be good design and solve the issue? I made them props since each instance of the Expander could have unique content height and therefore unique expanded and collapsed heights.

Saadnajmi: That's definitely something we did in ActivityIndicator and something that could work. I think it would be a valid design decision. It's a part of styling that we also want to be a customizable property of each control.

Saadnajmi:
Something I forgot to mention: If you could add screenshots / videos to your PR, it would be really helpful to understand how the component looks / should behave when reviewing. And is good historical data when looking through git history :D

packages/experimental/Expander/windows/ReactNativeExpander/ReactPackageProvider.h
```diff
@@ -0,0 +1,21 @@
...
+    {
+        ReactPackageProvider() = default;
+
+        void CreatePackage(IReactPackageBuilder const &packageBuilder) noexcept;
+    };
+} // namespace winrt::ReactNativeExpander::implementation
+
+namespace winrt::ReactNativeExpander::factory_implementation
+{
+
```
rectified95, line 18: nit: make formatting consistent with line 8.
It's a style choice you can make, with my personal preference being to avoid extra spaces, such as on line 11 too

packages/experimental/Expander/windows/ReactNativeExpander/ReactPackageProvider.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include "ExpanderViewManager.h"
+
+using namespace winrt::Microsoft::ReactNative;
+
+namespace winrt::ReactNativeExpander::implementation
+{
+
+void ReactPackageProvider::CreatePackage(IReactPackageBuilder const &packageBuilder) noexcept
+{
+    packageBuilder.AddViewManager(L"ExpanderViewManager", []() { return winrt::make<ExpanderViewManager>(); });
```
rectified95, line 16: Consider adding changing the VM name to "RNExpander" or something similar in order to align the string name with other native modules (see: DatePicker, OrientationLocker, etc.).

yarn.lock
```diff
@@ -14043,7 +14043,7 @@ react-native-web@0.12.3, react-native-web@^0.12.3:
     prop-types "^15.6.0"
     react-timer-mixin "^0.13.4"
 
-react-native-windows@^0.63.0:
+react-native-windows@^0.63.0, react-native-windows@^0.63.0-0:
```
rectified95, line 14046: Was this added by yarn? This doesn't seem right, and AFAIK the `-0` suffix breaks the lookup of higher versions.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "winrt/Microsoft.ReactNative.h"
+#include "NativeModules.h"
+
+namespace winrt::ReactNativeExpander::implementation {
+
+    class ExpanderViewManager : public winrt::implements<
```
rectified95, line 8: nit: Typically, nested namespaces are used also in order to avoid extra indentation - you could shift everything to the left here

packages/experimental/Expander/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@fluentui-react-native/component-cache": "^1.2.0",
+    "@fluentui-react-native/framework": "0.5.15"
+  },
+  "devDependencies": {
+    "@types/react-native": "^0.63.0",
+    "@uifabricshared/build-native": "^0.1.1",
+    "@uifabricshared/eslint-config-rules": "^0.1.1",
+    "react": "16.13.1",
+    "react-native": "^0.63.4",
+    "react-native-windows": "^0.63.0-0"
```
rectified95, line 37: We typically don't want to specify pre-release (`-x`) versions as dependencies

packages/experimental/Expander/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "peerDependencies": {
+    "react": "16.13.1",
+    "react-native": ">=0.63.4",
+    "react-native-windows": ">=0.62"
```
rectified95, line 42: Since RN is min 63.4, this is too low for RNW

packages/experimental/Expander/just.config.js
```diff
@@ -0,0 +1,3 @@
+const { preset } = require('@uifabricshared/build-native');
+
```
rectified95, line 2: Is this file something all native FURN components need?

packages/experimental/Expander/package.json
Saadnajmi: I think I/we may have told her to use this syntax so that she can get both 0.63 and 0.64, since `^0.63.0` resolves to `>=0.63.0 <0.64.0` (https://devhints.io/semver). This syntax should resolve to `>=0.63.0 < 1`

That being said, having `react-native-windows` as a peer dependency should solve the issue and we can maybe remove the "-0", since the app consuming this component will bring it's own version of `react-native/react-native-windows`, regardless of what the package.json specifies.

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,38 @@
...
+    root: buildProps(
+      (tokens) => ({...tokens})
+    ),
+  },
+  slots: { root: ExpanderComponent },
+  render: (userProps: ExpanderProps, useSlots: UseSlots<ExpanderType>) => {
+    const Root = useSlots(userProps).root;
+    const [expandedState, setExpandedState] = React.useState(userProps.expanded);
+
+    return (rest: ExpanderViewProps, ...children: React.ReactNode[]) => <Root
```
Saadnajmi, line 25: Indentation here seems off?

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,38 @@
...
+      height= {expandedState? userProps.expandedHeight : userProps.collapsedHeight}
+      onChange={async (event: ExpanderChangeEvent) => {
+        if (event.nativeEvent.expanded != null) {
+          event.persist();
+          if (!event.nativeEvent.expanded) {
+            await delay(175);
+          }
+          setExpandedState(event.nativeEvent.expanded);
+        }
+      }}
```
Saadnajmi, line 35: I would define these 2 (height, onChange) as separate variables and then set them for better readability. The indentation levels threw me off here for a bit.

packages/experimental/Expander/src/Expander.types.ts
```diff
@@ -0,0 +1,174 @@
...
+export type HorizontalAlignment = 'center' | 'left' | 'right' | 'stretch';
+export type ExpanderViewProps = ExpanderProps & ExpanderTokens;
+
+export type ExpanderChangeEvent = SyntheticEvent<
+  Readonly <{
+    nativeEvent: {
+      expanded: boolean
+    },
+  }>
+>;
```
Saadnajmi, line 166: This looks like a Flow type, not a typescript type. Am I wrong?

Saadnajmi:
There is not a test page for Expander. Is that because it will not work on RNW 0.63? If not, could we add that?

Saadnajmi:
One more comment: Could we add a super simple barebones README.md to the package that maybe just links the WinUI documentation of the native component and says "This component wraps the native windows component".

packages/experimental/Expander/just.config.js
Saadnajmi: It's something that the initial components had, and I wrote my contribution guide to say "Copy all the files in an existing package and edit as appropriate". It's likely we need it, though I'm not sure why.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    // IViewManagerWithExportedEventTypeConstants
+    ConstantProviderDelegate ExpanderViewManager::ExportedCustomBubblingEventTypeConstants() noexcept {
+        return nullptr;
+    }
+
+    ConstantProviderDelegate ExpanderViewManager::ExportedCustomDirectEventTypeConstants() noexcept {
+        return [](winrt::IJSValueWriter const& constantWriter) {
+            WriteCustomDirectEventTypeConstant(constantWriter, "onChange");
```
rectified95, line 88: Check if removing this line beaks anything - I think `onChange` is registered by default in RNW code

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    void ExpanderViewManager::ReactContext(IReactContext reactContext) noexcept {
+        m_reactContext = reactContext;
+    }
+
+    // IViewManagerWithNativeProperties
+    IMapView<hstring, ViewManagerPropertyType> ExpanderViewManager::NativeProps() noexcept {
+        auto nativeProps = winrt::single_threaded_map<hstring, ViewManagerPropertyType>();
+
+        // TODO: finish updating props when Expander.types are further defined
```
rectified95, line 37: is this TODO done now?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+        else {
+            m_reactContext.CallJSFunction(
+                L"RCTLog",
+                L"logToConsole",
+                MakeJSValueArgWriter("warn", "React Native for Windows does not yet support nesting non-FrameworkElement components under <Expander>"));
+        }
+    }
+
+    void ExpanderViewManager::RemoveAllChildren(winrt::Windows::UI::Xaml::FrameworkElement const& parent) noexcept {
+        return;
```
rectified95, line 127: I think you can fill these implementations fairly easily, see existing RNW VMs for examples.
Applies to ReplcaeChild and RemoveChildAt too

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // IViewManagerWithChildren
+    int32_t ExpanderViewManager::ReplaceChild(winrt::Windows::UI::Xaml::FrameworkElement const& parent, winrt::Windows::UI::Xaml::UIElement const& oldChild, winrt::Windows::UI::Xaml::UIElement const& newChild) noexcept {
+        return 0;
+    }
+
+    void ExpanderViewManager::AddView(winrt::Windows::UI::Xaml::FrameworkElement const& parent, winrt::Windows::UI::Xaml::UIElement const& child, int64_t index) noexcept {
+        auto contentPresenter = parent.try_as<winrt::Windows::UI::Xaml::Controls::ContentPresenter>();
+        auto expanderWrapper = contentPresenter.Content();
+        auto expander = expanderWrapper.try_as<Microsoft::UI::Xaml::Controls::Expander>();
```
rectified95, line 102: Can merge these 2 lines, since you're not doing anything else with `expanderWrapper`
Also, wen using `try_as` you should be checking for null since it may occur. However, if that happened here, then we can't handle it in any smart way, so might as well crash - use `.as<>()` which will throw an exception.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+                    L"RCTLog",
+                    L"logToConsole",
+                    MakeJSValueArgWriter("warn", "React Native for Windows does not yet support nesting more than two components under <Expander>.\nThe first component will be the Expander header, and the second will be the Expander content."));
+            }
+        }
+        else {
+            m_reactContext.CallJSFunction(
+                L"RCTLog",
+                L"logToConsole",
+                MakeJSValueArgWriter("warn", "React Native for Windows does not yet support nesting non-FrameworkElement components under <Expander>"));
```
rectified95, line 122: Is this something we have a chance of supporting in the future, or rather a WinUI assertion?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.h
```diff
@@ -0,0 +1,33 @@
...
+    public:
+        ExpanderView(Microsoft::ReactNative::IReactContext const& reactContext);
+        void UpdateProperties(Microsoft::ReactNative::IJSValueReader const& reader);
+
+    private:
+        Microsoft::ReactNative::IReactContext m_reactContext{ nullptr };
+        bool m_updating{ false };
+        bool m_expanded{ false };
+
+        Microsoft::UI::Xaml::Controls::Expander::Collapsed_revoker m_expanderCollapsingRevoker{};
```
rectified95, line 19: align variable name with type name

rectified95: or add comment in cpp explaining the error in naming

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "ExpanderView.g.h"
+#include "winrt/Microsoft.ReactNative.h"
+#include "NativeModules.h"
+
+namespace winrt::ReactNativeExpander::implementation {
+
+    class ExpanderView : public ExpanderViewT<ExpanderView> {
```
rectified95, line 9: Just checked we do collapse this first level of indentation in RNW code

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.h
```diff
@@ -0,0 +1,33 @@
...
+
+    private:
+        Microsoft::ReactNative::IReactContext m_reactContext{ nullptr };
+        bool m_updating{ false };
+        bool m_expanded{ false };
+
+        Microsoft::UI::Xaml::Controls::Expander::Collapsed_revoker m_expanderCollapsingRevoker{};
+        Microsoft::UI::Xaml::Controls::Expander::Expanding_revoker m_expanderExpandingRevoker{};
+
+        void SetResourceColor(Microsoft::ReactNative::JSValue const& value, hstring key);
```
rectified95, line 22: nit: I think the more common order of parameters would be the reverse

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+    using namespace Microsoft::ReactNative;
+    using namespace Windows::Foundation;
+    using namespace Windows::UI::Xaml::Media;
+}
+
+namespace winrt::ReactNativeExpander::implementation {
+
+    ExpanderView::ExpanderView(winrt::IReactContext const& reactContext) : m_reactContext(reactContext) {
+        auto expander = Microsoft::UI::Xaml::Controls::Expander();
+        this->VerticalContentAlignment(Windows::UI::Xaml::VerticalAlignment::Top);
```
rectified95, line 19: @asklar Should we use the aliases `xaml::` namespace here? Or is that something that is only necessary in RNW core code, no longer necessary with our update of the default winui version?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+        this->Content(expander);
+
+        RegisterEvents();
+    }
+
+    void ExpanderView::RegisterEvents() {
+        auto expander = (this->Content()).try_as<Microsoft::UI::Xaml::Controls::Expander>();
```
rectified95, line 26: `->` and `.` operators have the same precedence and are evaluated LTR in C++, so you can remove the parentheses around Content

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+            resDict.Insert(winrt::box_value(key), color);
+        }
+    }
+
+    void ExpanderView::OnCollapsing(winrt::Windows::Foundation::IInspectable const& sender,
+                          Microsoft::UI::Xaml::Controls::ExpanderCollapsedEventArgs const& args) {
+        if (!m_updating) {
+            m_reactContext.DispatchEvent(
+                *this,
+                L"topCollapsing",
```
rectified95, line 221: Since you're firing `onChange` and  `onCollapsing` at the same time`, I think you can remove onCollapsing, as it doesn't seem to convey any extra information; same for onExpanding

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+                }
+            }
+            else if (propertyName == "height") {
+                if (!propertyValue.IsNull()) {
+                    expander.Height(propertyValue.AsDouble());
+                    this->Height(propertyValue.AsDouble());
+                }
+            }
+            else if (propertyName == "contentHorizontalAlignment") {
+                if (!propertyValue.IsNull()) {
```
rectified95, line 96: for some of the properties you clear the value if the incoming argument is null - should that be done for all props, or not really?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+            else if (propertyName == "headerBackground") {
+                this->SetResourceColor(propertyValue, L"ExpanderHeaderBackground");
+            }
+            else if (propertyName == "headerForeground") {
+                this->SetResourceColor(propertyValue, L"ExpanderHeaderForeground");
+            }
+            else if (propertyName == "headerBorderThickness") {
+                if (!propertyValue.IsNull()) {
+                    auto resDict = expander.Resources();
+                    Windows::UI::Xaml::Thickness thickness = Windows::UI::Xaml::ThicknessHelper::FromUniformLength(propertyValue.AsDouble());
```
rectified95, line 138: nit can be auto

rectified95: also i think you're alising this namespace so winrt:: might work if i'm looking correctly

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
asklar: yes, this should use the appropriate alias so that it will "just work" with WinUI 3.

I'm still a little confused as to the nature of this change overall... I thought the idea was to use react-native-xaml to expose the control to JS, it seems like this change is just hand-rolling duplicated logic we already have in place?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,255 @@
...
+            else if (propertyName == "chevronBorderPressedBrush") {
+                this->SetResourceColor(propertyValue, L"ExpanderChevronBorderPressedBrush");
+            }
+        }
+        m_updating = false;
+    }
+
+    void ExpanderView::SetResourceColor(Microsoft::ReactNative::JSValue const& value, hstring key) {
+        if (!value.IsNull()) {
+            auto expander = (this->Content()).try_as<Microsoft::UI::Xaml::Controls::Expander>();
```
asklar, line 209: missing `if` (otherwise line 211 will crash)

packages/experimental/Expander/README.md
```diff
@@ -0,0 +1,452 @@
...
+## Usage
+
+Import Expander from @fluentui-react-native/experimental-expander
+
+`import { Expander } from @fluentui-react-native/experimental-expander`
+
+Add `Expander` as follows:
+
+```javascript
+<Expander collapsedHeight={50} expandedHeight={100}>
```
rectified95, line 28: might want to call out the need to specify these height values

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
+#include "pch.h"
```
asklar, line 1: I'm a little confused as to the nature of this change overall... I thought the idea was to use react-native-xaml to expose the control to JS, but it seems like this change is just hand-rolling duplicated logic that we already have in place?

packages/experimental/Expander/README.md
```diff
@@ -0,0 +1,452 @@
...
+- [`chevronBorderThickness`](#chevronborderthickness)
+- [`chevronBorderBrush`](#chevronborderbrush)
+- [`chevronBorderPointerOverBrush`](#chevronborderpointeroverbrush)
+- [`chevronBorderPressedBrush`](#chevronborderpressedbrush)
+
+### Important notes
+
+- `collapsedHeight` and `expandedHeight` must be set for Expander to display correctly
+- Do not set the `height` or `onChange` props. These props are used to update the height of Expander, and the Expander height will not update if these props are overridden. If you need to override either of these props, you must implement the height change functionality as well.
+- Non-native components within the header will not be interactable (i.e. if there is a `Button` in the header, and you press it, the Expander will expand, and the `Button` functionality will not be performed). We recommend using [react-native-xaml](https://github.com/asklar/react-native-xaml) controls in the header if you would like to interact with the control and there is not a native implementation.
```
rectified95, line 93: Isn't `interactive` the word? :D

packages/experimental/Expander/README.md
```diff
@@ -0,0 +1,452 @@
...
+- [`chevronPressedForeground`](#chevronpressedforeground)
+- [`chevronBorderThickness`](#chevronborderthickness)
+- [`chevronBorderBrush`](#chevronborderbrush)
+- [`chevronBorderPointerOverBrush`](#chevronborderpointeroverbrush)
+- [`chevronBorderPressedBrush`](#chevronborderpressedbrush)
+
+### Important notes
+
+- `collapsedHeight` and `expandedHeight` must be set for Expander to display correctly
+- Do not set the `height` or `onChange` props. These props are used to update the height of Expander, and the Expander height will not update if these props are overridden. If you need to override either of these props, you must implement the height change functionality as well.
```
rectified95, line 92: What would the user need to do to override the onChange?

packages/experimental/Expander/README.md
```diff
@@ -0,0 +1,452 @@
...
+
+| Type   | Required |
+| ------ | -------- |
+| string | No       |
+
+---
+
+### `contentBorderBrush`
+
+Content border color. Input value should be hexadecimal string.
```
rectified95, line 348: I thought that you could also pass string names of colors - can you check if that works?

rectified95 (commented):
Left a bunch of minor comments, good job! :)

packages/experimental/Expander/just.config.js
NickGerleman: This defines tasks like linting and building.

packages/experimental/Expander/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "devDependencies": {
+    "@types/react-native": "^0.63.0",
+    "@uifabricshared/build-native": "^0.1.1",
+    "@uifabricshared/eslint-config-rules": "^0.1.1",
+    "react": "16.13.1",
+    "react-native": "^0.63.4",
+    "react-native-windows": "^0.63.0-0"
+  },
+  "peerDependencies": {
+    "react": "16.13.1",
```
NickGerleman, line 40: The version of React can change with newer versions of RN, so this should allow newer versions as well.

packages/experimental/Expander/src/Expander.types.ts
```diff
@@ -0,0 +1,174 @@
+import { ColorValue } from 'react-native';
+import type {SyntheticEvent} from 'react-native/Libraries/Types/CoreEventTypes';
```
NickGerleman, line 2: Typescript will see this as "any" instead of the real type you want, since the internals of RN are in Flow.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
NickGerleman: I am not sure the aliased types are publicly exported outside of RNW.

yarn.lock
NickGerleman: -0 means to allow any prerelease build as well. It can still resolve to non-prerelease versions though.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
asklar: we expose cppwinrtincludes.h and the ui.*.h headers

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
asklar: we will move to a better api soon - see IViewManagerCreateWithProperties and microsoft/react-native-windows#5539

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
amdingler: I'll update it to clear the values for properties with associated dependency properties (height, width, alignment). I'm not sure if there's a clearValue equivalent for properties set through the resource dictionary

packages/experimental/Expander/README.md
amdingler: I meant if the user implements an onChange callback, our callback that adjusts the height will be gone. If the user wants to use onChange, they should use the onExpanding and onCollapsing events instead.

rectified95: Ok, I'd then rephrase this part: "you must implement the height change functionality as well"

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,272 @@
...
+            if (auto self = ref.get()) {
+                self->OnExpanding(sender, args);
+            }
+        });
+    }
+
+    void ExpanderView::UpdateProperties(winrt::IJSValueReader const& reader) {
+        m_updating = true;
+        auto const& propertyMap = JSValueObject::ReadFrom(reader);
+        auto expander = this->Content().try_as<Microsoft::UI::Xaml::Controls::Expander>();
```
asklar, line 45: if expander is null we'll crash below - check for null / assert / do something

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,272 @@
...
+                    }
+                    else {
+                        expander.ExpandDirection(Microsoft::UI::Xaml::Controls::ExpandDirection::Down);
+                    }
+                }
+            }
+            else if (propertyName == "expanded") {
+                if (propertyValue.IsNull()) {
+                    expander.ClearValue(Microsoft::UI::Xaml::Controls::Expander::IsExpandedProperty());
+                    m_expanded = false;
```
asklar, line 68: `m_expanded` is only ever written to and therefore unnecessary, remove the member variable

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,272 @@
...
+                }
+                else if (expander.IsExpanded() != propertyValue.AsBoolean()) {
+                    m_expanded = propertyValue.AsBoolean();
+                    expander.IsExpanded(m_expanded);
+                }
```
asklar, line 73: can simplify to `else { expander.IsExpanded(propertyValue.AsBoolean()); }`

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const Root = useSlots(userProps).root;
+    const [expandedState, setExpandedState] = React.useState(userProps.expanded);
+    const expanderHeight = expandedState? userProps.expandedHeight : userProps.collapsedHeight;
+    async function onExpanderChange(event: ExpanderChangeEvent) {
+      if (event.nativeEvent.expanded != null) {
+        event.persist();
+        if (!event.nativeEvent.expanded) {
+          await delay(175);
+        }
+        setExpandedState(event.nativeEvent.expanded);
```
Saadnajmi, line 31: Why do we need the `persist()` call and `delay(175)` call? Maybe add inline comments for future readers as well.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,272 @@
...
+            else if (propertyName == "headerBackground") {
+                this->SetResourceColor(propertyValue, L"ExpanderHeaderBackground");
+            }
+            else if (propertyName == "headerForeground") {
+                this->SetResourceColor(propertyValue, L"ExpanderHeaderForeground");
+            }
+            else if (propertyName == "headerBorderThickness") {
+                auto resDict = expander.Resources();
+                if (!propertyValue.IsNull()) {
+                    auto thickness = xaml::ThicknessHelper::FromUniformLength(propertyValue.AsDouble());
```
asklar, line 148: fyi (not blocking for this PR), we now (0.66) have JSValue readers for Thickness which let you specify a uniform thickness, or individual thicknesses `[1,2,3,4]` or `{left:1, top:2, right:3, bottom:4}`

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,41 @@
...
+      (tokens) => ({...tokens})
+    ),
+  },
+  slots: { root: ExpanderComponent },
+  render: (userProps: ExpanderProps, useSlots: UseSlots<ExpanderType>) => {
+    const Root = useSlots(userProps).root;
+    const [expandedState, setExpandedState] = React.useState(userProps.expanded);
+    const expanderHeight = expandedState? userProps.expandedHeight : userProps.collapsedHeight;
+    async function onExpanderChange(event: ExpanderChangeEvent) {
+      if (event.nativeEvent.expanded != null) {
```
Saadnajmi, line 26: In my own native component I am implementing (https://github.com/microsoft/fluentui-react-native/pull/827/files#diff-a20f47efb37c1fffa3d336f27135a753a289722fdd477b6e5c70291be7f28784R36), I ended up just doing `(event: Any)`, which according to @NickGerleman is how typescript sees this flow type anyway. You might want to consider just doing that since the type isn't actually getting read properly.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -0,0 +1,272 @@
...
+            else if (propertyName == "chevronPressedForeground") {
+                this->SetResourceColor(propertyValue, L"ExpanderChevronPressedForeground");
+            }
+            else if (propertyName == "chevronBorderThickness") {
+                auto resDict = expander.Resources();
+                if (!propertyValue.IsNull()) {
+                    auto thickness = xaml::ThicknessHelper::FromUniformLength(propertyValue.AsDouble());
+                    resDict.Insert(winrt::box_value(L"ExpanderChevronBorderThickness"), winrt::box_value(thickness));
+                } else {
+                    resDict.Remove(winrt::box_value(L"ExpanderChevronBorderThickness"));
```
asklar, line 205: can you clarify if the resource dictionary already has a value for this key upon creation, and we are Inserting into it to override the value, or if it's empty and we're inserting to set a value?

Does this work with fast refresh? IIRC ResourceDictionary is quirky in that it doesn't re-evaluate at runtime once an element is fully created and rendered

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+        return [](winrt::IJSValueWriter const& constantWriter) {
+            WriteCustomDirectEventTypeConstant(constantWriter, "onChange");
+            WriteCustomDirectEventTypeConstant(constantWriter, L"Collapsing");
+            WriteCustomDirectEventTypeConstant(constantWriter, L"Expanding");
+        };
+    }
+
+    // IViewManagerWithChildren
+    int32_t ExpanderViewManager::ReplaceChild(winrt::Windows::UI::Xaml::FrameworkElement const& parent, winrt::Windows::UI::Xaml::UIElement const& oldChild, winrt::Windows::UI::Xaml::UIElement const& newChild) noexcept {
+        return 0;
```
asklar, line 96: missing impl?

Saadnajmi (commented):
I think the JS side looks good! We talked about this offline, but the only open question I have left is whether we can add the Expander test to FluentTester, so future people have a reference to code that uses the component. Great work!

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    void ExpanderViewManager::AddView(winrt::Windows::UI::Xaml::FrameworkElement const& parent, winrt::Windows::UI::Xaml::UIElement const& child, int64_t index) noexcept {
+        auto contentPresenter = parent.try_as<winrt::Windows::UI::Xaml::Controls::ContentPresenter>();
+        auto expanderWrapper = contentPresenter.Content();
+        auto expander = expanderWrapper.try_as<Microsoft::UI::Xaml::Controls::Expander>();
+
+        if (auto content = child.try_as<winrt::Windows::UI::Xaml::FrameworkElement>()) {
```
asklar, line 104: I'm not sure this check makes sense - both Header and Content take an IInspectable, is there a reason we're downcasting here?

packages/experimental/Expander/windows/ReactNativeExpander/ReactNativeExpander.vcxproj
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="utf-8"?>
```
asklar, line 1: Looks like this project might have been created with an older version of RNW which doesn't have ExperimentalFeatures.props - this is important in order to enable apps using this module to customize their own version of the SDK, WinUI, etc.

asklar (changes requested):

packages/experimental/Expander/src/Expander.tsx
amdingler: When the expander collapses and the height manually changes, the collapsing animation gets cut short, so I added a delay for the height change. I'll add a comment.

packages/experimental/Expander/windows/ReactNativeExpander/ReactNativeExpander.vcxproj
amdingler: I created the project with 0.63.0 initially since the FURN repo required it. Is there a good way to allow for customization without generating a new project? Or would generating a 0.64.0 project and copying the important files over be the easiest?

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
amdingler: The resource dictionary doesn't have a value for ExpanderChevronBorderThickness, and we are inserting to set a value. It is working with fast refresh.

If the resource dictionary doesn't have a key/value, is it pulled from [here](https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/Expander/Expander_themeresources.xaml)? Just curious how the relationship between the two works

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
amdingler: I'm working on this implementation, and was wondering what scenarios cause me to hit this method so that I can test it. I've tried changing a child from Text to Button, and `RemoveChildAt` and `AddView` were hit instead of `ReplaceChild`

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -21,18 +21,30 @@ export const Expander = compose<ExpanderType>({
...
+        userProps.onExpanding();
+      }
+    }
+
+    return (rest: ExpanderViewProps, ...children: React.ReactNode[]) =>
+      <Root
+        {...mergeProps(userProps, rest)}
+        height={expanderHeight}
+        onCollapsing={_onCollapsing}
+        onExpanding={_onExpanding}>{children}
```
rectified95, line 47: nit move `children` to next line

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -21,18 +21,30 @@ export const Expander = compose<ExpanderType>({
...
+        userProps.onCollapsing();
+      }
+      // Need to delay the height change so that the animation runs
+      await delay(175);
+      setExpandedState(false);
+    }
+
+    const _onExpanding = () => {
+      setExpandedState(true);
+      if (userProps.onExpanding !== undefined) {
```
rectified95, line 37: nit, I think you can make this more compact by writing:
`userProps.onExpanding && userProps.onExpanding()`
double-check though if it correctly detects the empty callback case

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
```diff
@@ -42,210 +42,206 @@ namespace winrt::ReactNativeExpander::implementation {
...
-            auto color = value.To<winrt::Brush>();
+    void ExpanderView::SetResourceColor(hstring key, Microsoft::ReactNative::JSValue const& value) {
+        if (auto expander = this->Content().try_as<Microsoft::UI::Xaml::Controls::Expander>()) {
             auto resDict = expander.Resources();
-            resDict.Insert(winrt::box_value(key), color);
+            if (!value.IsNull()) {
+                auto color = value.To<winrt::Brush>();
+                resDict.Insert(winrt::box_value(key), color);
+            } else {
+                resDict.Remove(winrt::box_value(key));
```
rectified95, line 227: making sure the resource dictionary structure won't throw an exception while retireving an element that's not there

packages/experimental/Expander/windows/ReactNativeExpander/ReactPackageProvider.cpp
```diff
@@ -13,7 +13,7 @@ namespace winrt::ReactNativeExpander::implementation
 
 void ReactPackageProvider::CreatePackage(IReactPackageBuilder const &packageBuilder) noexcept
 {
-    packageBuilder.AddViewManager(L"ExpanderViewManager", []() { return winrt::make<ExpanderViewManager>(); });
+    packageBuilder.AddViewManager(L"RNExpander", []() { return winrt::make<ExpanderViewManager>(); });
```
rectified95, line 16: making sure you updated the string in tsx as well, didn't see it in a quick scan over the files

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp
```diff
@@ -85,49 +84,65 @@ namespace winrt::ReactNativeExpander::implementation {
...
-
-        if (auto content = child.try_as<winrt::Windows::UI::Xaml::FrameworkElement>()) {
-            if (index == 0) {
-                expander.Header(content);
-            }
-            else if (index == 1) {
-                expander.Content(content);
+        auto expander = parent.as<xaml::Controls::ContentPresenter>().Content().as<Microsoft::UI::Xaml::Controls::Expander>();
+        if (index == 0) {
+            if (auto newContent = expander.GetValue(Microsoft::UI::Xaml::Controls::Expander::HeaderProperty())) {
```
rectified95, line 110: got confused by the naming here: the logic is correct, but I'd change `newContent` to `header/currentHeader`

rectified95 (approved):
Great stuff! Left a few nit comments. Also, make sure you finish off `ReplaceChild()` and resolve the conflict in yarn.lock

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -22,20 +22,29 @@ export const Expander = compose<ExpanderType>({
...
-          await delay(175);
-        }
-        setExpandedState(event.nativeEvent.expanded);
+
+    const _onCollapsing = async () => {
+      if (userProps.onCollapsing !== undefined) {
+        userProps.onCollapsing();
       }
+      // Need to delay the height change so that the animation runs
+      await delay(175);
```
asklar, line 31: nit - I wonder if there is a better way to be notified of "the animation has completed", e.g. [Completed](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.animation.timeline.completed?view=winrt-20348) event.

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.idl
```diff
@@ -0,0 +1,9 @@
+namespace ReactNativeExpander{
+
+    [default_interface]
+    runtimeclass ExpanderView : Windows.UI.Xaml.Controls.ContentPresenter {
```
asklar, line 4: We should add a comment to the IDL to explain why we need a ContentPresenter in the first place

asklar (approved):

packages/experimental/Expander/api-extractor.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "@uifabricshared/build-native/api-extractor.json"
```
PPatBoyd, line 2: nit shouldn't be needed w/ api-extractor removal

Saadnajmi: Good catch! We can (and should) delete this file now that we don't use it.

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -0,0 +1,50 @@
...
+  render: (userProps: ExpanderProps, useSlots: UseSlots<ExpanderType>) => {
+    const Root = useSlots(userProps).root;
+    const [expandedState, setExpandedState] = React.useState(userProps.expanded);
+    const expanderHeight = expandedState? userProps.expandedHeight : userProps.collapsedHeight;
+
+    const _onCollapsing = async () => {
+      if (userProps.onCollapsing !== undefined) {
+        userProps.onCollapsing();
+      }
+      // Need to delay the height change so that the animation runs
```
PPatBoyd, line 30: is this a pattern rn-windows folks have been taking to generate layout animations w/o using LayoutAnimation [due to limitations e.g. its global nature] ? This is a 175ms wait on the JS thread, no?

packages/experimental/Expander/src/Expander.tsx
PPatBoyd: What's the feature we're enabling w/ an animation completed callback?

packages/experimental/Expander/src/Expander.tsx
PPatBoyd: userProps?.onExpanding()

packages/experimental/Expander/src/Expander.tsx
PPatBoyd: little surprised the linter didn't pick it up 🤔

packages/experimental/Expander/src/Expander.tsx
amdingler: When the Expander is collapsed or expanded, we manually adjust the height of the Expander. For the collapsing event, the height is changed too early and clips the native control's collapsing animation, so we add a slight delay to the thread before making the height adjustment.

packages/experimental/Expander/src/Expander.types.ts
```diff
@@ -0,0 +1,161 @@
...
+  /*
+   * Chevron background color at rest
+   */
+  chevronBackground?: ColorValue;
+  /*
+   * Chevron foreground color at rest
+   */
+  chevronForeground?: ColorValue;
+  /*
+   * Chevron background color on pointer over
```
PPatBoyd, line 117: the chevron gets different background and foreground colors from the rest of the control's `background`, e.g. we can't reuse BackgroundTokens ?

the variations should be handled in styling and only need chevronBackground chevronForeground

packages/experimental/Expander/src/Expander.tsx
amdingler: Yes, we added a wait on the JS thread to allow enough time for the native animation to complete before changing the height of the Expander.

asklar: @ppatboyd RNW does not support RN layout animations. All animations here are XAML animations.

packages/experimental/Expander/src/Expander.types.ts
amdingler: I pulled these tokens from values that can be put in the [ResourceDictionary](https://docs.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.expander?view=winui-2.6#remarks), so they can all be different colors.

packages/experimental/Expander/src/Expander.tsx
```diff
@@ -24,27 +24,24 @@ export const Expander = compose<ExpanderType>({
     const expanderHeight = expandedState? userProps.expandedHeight : userProps.collapsedHeight;
 
     const _onCollapsing = async () => {
-      if (userProps.onCollapsing !== undefined) {
-        userProps.onCollapsing();
-      }
+      userProps.onCollapsing && userProps.onCollapsing();
```
asklar, line 27: I think you can write this as `userProps?.onCollapsing()`

packages/experimental/Expander/windows/ReactNativeExpander/ExpanderView.cpp
asklar: this might require "tapping" the app to re-calculate colors/brushes, see https://github.com/microsoft/microsoft-ui-xaml/issues/5457 and https://github.com/microsoft/microsoft-ui-xaml/issues/4443

packages/experimental/Expander/src/Expander.tsx
amdingler: This still calls the function even if the user didn't provide a callback, and that throws an error

asklar: I put the optional chaining on the wrong thing :D :
`userProps.onCollapsing?.()` this will call only if onCollapsing is not falsy

Saadnajmi: TIL I have learned the interesting syntax of `?.()` ...

apps/fluent-tester/src/FluentTester/TestComponents/Expander/ExpanderTest.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { Stack } from '@fluentui-react-native/stack';
+import { stackStyle, commonTestStyles as commonStyles } from '../Common/styles';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { EXPANDER_TESTPAGE } from './consts';
+import { View, Switch } from 'react-native';
+
+const expanderTest: React.FunctionComponent = () => {
+  /** This test page has not yet been tested and does not currently build because
+   * the react-native-test-app does not yet support WinUI 2.6
+   */
```
Saadnajmi, line 13: I would leave a link to the RNTA issue that you opened here https://github.com/microsoft/react-native-test-app/issues/444

Thanks for adding this!

Saadnajmi (approved):

Saadnajmi (approved):

PR #757, opened 2021-07-02
Create use-tokens package, make use-styling package be based on use-tokens

JasonVMo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This creates a new package called @fluentui-react-native/use-tokens which encapsulates the token extraction functionality used by the theming system. The existing implementation of this system found in use-styling will now use the new use-tokens package.

The purpose of this change is to allow a simpler pattern for creating themed controls. Particularly those that simply wrap a native text control. 

The idea here is to extract the key pieces of value from the framework into more granular methods that people can learn to use as needed. The current framework brings everything together behind the scenes which makes it hard to understand what is going on under the covers.

Some next steps:
- **Customize pattern** - the current implementation of the hook exposes a `.customize()` function but doesn't have a helper to make it into a control building pattern.
- **use-slot package** - the basics of having a pluggable slot, both for production and consumption can be broken into a more granular package without breaking the framework.
- **Update old framework** - the old framework can likely be made to depend on this as well. I decided to be a bit more granular though.

From the readme file:

At the core of this system is the idea that: **`Tokens + Theme ==> Styles`**. This can be thought of as:

- Your styles are produced via a recipe, ideally memoized via the returned cache or by the use of a `useMemo` hook or its equivalent
- The tokens are the ingredients for your recipe
- Some of those tokens come from constants
- Some of those tokens come from the theme

The overall signature of the method is as follows:

```ts
export type StyledTextTokens = Pick<TextProps[style], 'color', 'fontFamily', 'fontWeight', 'fontSize'>;

const useTokens = buildUseTokens<StyledTextTokens, Theme>(
  /* function for looking up a component in a theme, allows this to be used independently of the shape of the theme */
  getComponentInfoFromTheme,

  /* Start of tokens, these are applied in order */
  /* object, these are effectively constants to be used as defaults */
  { token1: 'foo', token2: 400 },
  /* theme function */
  (theme: Theme) => ({ token3: theme.colors.backgroundColor }),
  /* name to lookup */
  'NameToLookup',
);
```

This `useTokens` hook can then be used to build styles for the component as follows:

```tsx
const StyledText: React.FunctionComponent<TextProps> = (props) => {
  // split the props
  const { style, children, ...rest } = props;

  // left out of the hook to allow theme to be obtained in whatever way is appropriate. Generally from a context.
  const theme = useTheme();

  // get a set of tokens from the theme
  const [tokens, cache] = useTokens(theme);

  // create a style object from the tokens using the cache. Note that if all the tokens are actually directly styles the tokens could be used
  // directly. This is just to illustrate the pattern for a more complex component
  const [tokenStyle] = cache(() => ({ ...tokens }), []);

  // now merge any styles coming from the props in on top as they should override
  const mergedStyles = mergeStyles(tokenStyle, style);

  // now just render the text element
  return (
    <Text {...rest} style={mergedStyles}>
      {children}
    </Text>
  );
};
```

### Verification

Verification is still in progress beyond automated tests.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/framework/use-tokens/src/__snapshots__/useTokens.samples.test.tsx.snap
```diff
@@ -0,0 +1,163 @@
...
+  <Text
+    style={
+      Object {
+        "color": "black",
+        "fontFamily": "Arial",
+        "fontSize": 12,
+      }
+    }
+  >
+    <Text
```
rurikoaraki, line 14: I'm not understanding why there's identical Text elements and one is nested within the other. Can you explain where the inner element is coming from?

rurikoaraki (commented):
One question about the snapshot file, otherwise looks good.

packages/framework/use-tokens/src/__snapshots__/useTokens.samples.test.tsx.snap
JasonVMo: Sorry was on vacation, I believe the inner text element is the equivalent of the native component when doing a full mounting. i.e. an implementation detail of the testing framework.

rurikoaraki: Thanks for the explanation, makes sense.

rurikoaraki (approved):

PR #768, opened 2021-07-07
Change Adapters to Export a Union of all Platform Defaults

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes

In regards to [this bug](https://office.visualstudio.com/OC/_workitems/edit/5110816).

There are some scenarios where partners' typechecking is resolving the correct platform resolution (out of our scope). This change makes it so adapters.ts exports a union of all the platforms to ensure that even if platform resolution doesn't correctly happen, they will still have all necessary props.

In this specific scenario, adapters.ts wasn't resolving to adapters.win32.ts for MCWW. Thus, they weren't receiving some callbacks they need, such as onFocus. 

I had to make some additional changes to other files that align with the changes to IViewProps, in Link and Persona/PersonaCoin. However, they aren't significant changes.

### Verification

Manual 

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Checkbox/package.json
```diff
@@ -31,7 +31,8 @@
     "@fluentui-react-native/tokens": ">=0.9.3 <1.0.0",
     "@uifabricshared/foundation-composable": ">=0.8.2 <1.0.0",
     "@uifabricshared/foundation-settings": ">=0.9.1 <1.0.0",
-    "react-native-svg": "^12.1.1"
+    "react-native-svg": "^12.1.1",
+    "@fluentui-react-native/pressable": ">=0.6.9 <1.0.0"
```
Saadnajmi, line 35: nit: can we keep this alphabetical?
Running `yarn add -D @fluentui-react-native/pressable` should do that automatically

packages/components/Persona/src/Persona.tokens.root.ts
```diff
@@ -1,11 +1,12 @@
...
 import { styleFunction } from '@uifabricshared/foundation-tokens';
 import { IPersonaTokens } from './Persona.types';
 import { ITheme } from '@uifabricshared/theming-ramp';
+import type { IViewProps } from '@fluentui-react-native/adapters';
 
 const _rootKeyProps: (keyof IPersonaTokens)[] = ['coinSize', 'size'];
 
-function _buildRootStyle(tokenProps: IPersonaTokens): ViewProps {
+function _buildRootStyle(tokenProps: IPersonaTokens): IViewProps {
   const rootStyle: ViewStyle = {
```
Saadnajmi, line 10: Is the expectation that all controls use IViewProps instead of ViewProps from React Native?

packages/utils/adapters/src/adapters.ts
```diff
@@ -1,10 +1,11 @@
 import { TextProps, ViewProps, ImageProps } from 'react-native';
 import { IFilterMask } from './filter.types';
-import type { ITextWin32Props } from '@office-iss/react-native-win32';
+import type { ITextWin32Props, IViewWin32Props } from '@office-iss/react-native-win32';
+import { IViewWindowsProps } from 'react-native-windows';
 
 // export core interface types
 export type ITextProps = TextProps & Partial<ITextWin32Props>;
-export type IViewProps = ViewProps;
+export type IViewProps = ViewProps & Partial<IViewWin32Props> & Partial<IViewWindowsProps>;
```
Saadnajmi, line 8: There are some macOS specific props that are not included in here as well, like `nextKeyView`

samuelfreiberg: I don't see a macOS adapter file, so that might be why. Do we need to create one?

packages/components/Checkbox/package.json
samuelfreiberg: Yes, I have a bad habit of adding it programmatically :) I'll fix this now

packages/utils/adapters/src/adapters.ts
Saadnajmi: Oh yeah, there isn't one. Probably, since this package seems pretty important and we're starting to use a lot of controls on macOS

packages/components/Persona/src/Persona.tokens.root.ts
samuelfreiberg: I think so - the adapters package filters out the necessary props for each separate platform, so I don't see the negative to it. Although I'm open to hear others' thoughts.

packages/components/Link/src/Link.tsx
```diff
@@ -10,11 +10,11 @@ import { foregroundColorTokens, textTokens, borderTokens } from '@fluentui-react
...
 import { ISlots, withSlots } from '@uifabricshared/foundation-composable';
-import { IViewProps } from '@fluentui-react-native/adapters';
+import { ITextProps } from '@fluentui-react-native/text';
 
-export type ILinkHooks = [IWithLinkOptions<IViewProps>, ILinkState];
+export type ILinkHooks = [IWithLinkOptions<ITextProps>, ILinkState];
 
-export function useAsLink(userProps: IWithLinkOptions<IViewProps>, ref: React.RefObject<any>): ILinkHooks {
+export function useAsLink(userProps: IWithLinkOptions<ITextProps>, ref: React.RefObject<any>): ILinkHooks {
   const { url, onPress, ...rest } = userProps;
```
ksiler, line 18: Why is this changed to `ITextProps`?

ksiler: Oh, I see that `ILinkProps` is defined the same way (`IWithLinkOptions<ITextProps>`), so I'm guessing you're matching that? I'm a bit confused because the props generated from this function are being passed to the root View (see lines 59 and 71). So I'm wondering how View makes use of `ITextProps`.

packages/components/Persona/src/Persona.tokens.root.ts
ksiler: I don't see a negative to it either... I suppose we were using ViewProps in some cases because there wasn't a need for platform specific props but I don't see an issue changing it over.

Saadnajmi: I mostly asked so I could update contributing guidance / understand the ramifications of that. I think we've avoided doing that on the apple side because IViewProps doesn't include macOS and at the time none of us guide understood what it meant. That, and Jason told me we were moving away from the "I" prefix so I just kinda filtered those out of my head 😅

ksiler: Yeah, the "I" prefix may have been have been a habit carried over from other programming languages? Perhaps we should create a issue about it for discussion.

ksiler (changes requested):
This looks good to me but I'd like to get more clarification on the ITextProps vs IViewProps change on Link.

packages/components/Link/src/Link.tsx
samuelfreiberg: There was a mismatch in properties with the new changes to IViewProps that broke the bundle, so I tried something new. 

My logic was: When we call `useAsLink(rest, componentRef)` on line 59, the parameter `rest` is of type `ILinkProps extends IWithLinkOptions<ITextProps>`. So, even before this change, the only props that could have been passed to the root were within `IWithLinkOptions<ITextProps>`. There doesn't seem to be anything special in useAsLink that uses ViewProps. I was actually a bit confused why there were ViewProps there in the first place. Maybe I'm missing something?

ksiler: I think maybe useAsLink was using ViewProps for the `onKeyUp` property. And the reason why it works with ITextProps is because we added `onKeyUp` to ITextWin32Props?

samuelfreiberg: I think you're right. To be safe though I changes the adapters package a little so Link doesn't have to change. There was one prop that was Windows-specific that was causing an issue that I simply omitted from adding. Everything should still work correctly,

packages/utils/adapters/src/adapters.ts
```diff
@@ -1,10 +1,10 @@
 import { TextProps, ViewProps, ImageProps } from 'react-native';
 import { IFilterMask } from './filter.types';
-import type { ITextWin32Props } from '@office-iss/react-native-win32';
+import type { ITextWin32Props, IViewWin32Props } from '@office-iss/react-native-win32';
 
 // export core interface types
 export type ITextProps = TextProps & Partial<ITextWin32Props>;
-export type IViewProps = ViewProps;
+export type IViewProps = ViewProps & IViewWin32Props;
```
ksiler, line 7: Make this `Partial<IViewWin32Props>` like we did for Text? I don't think there's any required props in IViewWin32Props but it might be worth to safeguard this in case that changes.

ksiler (approved):

PR #771, opened 2021-07-08
Speculative fix for iOS Date Picker

chripear:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Excel noted a problem when integrating the FURN date picker into card view:

2021-07-01 11:26:40.950891+0530 Excel[21929:11984296] [Presentation] Attempt to present <MSFPageCardPresenterController: 0x6190004eac80> on <MSONavigationController: 0x61a000094280> (from <MSONavigationController: 0x61a000094280>) whose view is not in the window hierarchy.

This suggests that our guess of a UIViewController from which to present the calendar view is incorrect in their case.  I found a React API that will give us the UIViewController.  It seems likely this will fix the Excel team's issue but it is also a much less clunky way to obtain the view controller anyway, so it should be our default guess regardless.
 
### Verification
Used test app to verify that behavior of date picker is unchanged.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
+++ b/packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
@@ -29,7 +29,7 @@ public class DatePickerManager: NSObject {
         DispatchQueue.main.async {
             // We need a UIViewController to present the actual DateTimePicker view.
 			guard let viewController = RCTPresentedViewController() else {
-                fatalError("Unable to get the current UIViewController from React Native.")
+                preconditionFailure("Unable to get the current UIViewController from React Native.")
             }
 
             let picker = DateTimePicker()
--- a/packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
+++ b/packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
@@ -27,13 +27,9 @@ public class DatePickerManager: NSObject {
         callback: @escaping RCTResponseSenderBlock
     ) {
         DispatchQueue.main.async {
-            // We need a UIViewController to present the actual DateTimePicker
-            // view.  With the advent of multi-window in iOS 13, we shouldn't
-            // use UIApplication.shared.keyWindow?.rootViewController, because
-            // keyWindow is deprecated.  We need to pick one of what could be
-            // multiple key windows.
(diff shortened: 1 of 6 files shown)


packages/experimental/NativeDatePicker/ios/DatePickerManager.swift
```diff
@@ -27,13 +27,9 @@ public class DatePickerManager: NSObject {
...
-            // We need a UIViewController to present the actual DateTimePicker
-            // view.  With the advent of multi-window in iOS 13, we shouldn't
-            // use UIApplication.shared.keyWindow?.rootViewController, because
-            // keyWindow is deprecated.  We need to pick one of what could be
-            // multiple key windows.
-            guard let viewController = UIApplication.shared.windows.filter({$0.isKeyWindow}).first?.rootViewController else {
-                fatalError("Unable to get a UIViewController from current shared application context.")
+            // We need a UIViewController to present the actual DateTimePicker view.
+			guard let viewController = RCTPresentedViewController() else {
+                fatalError("Unable to get the current UIViewController from React Native.")
```
anandrajeswaran, line 32: unrelated to the current change but somewhat important: we need to avoid fatalError. It encodes the path of the source file in the binary which is not ok from a security perspective. Should use preconditionFailure instead. A pretty good writeup - https://lapcatsoftware.com/articles/fatalError.html

chripear: Wow.  That's not good.

packages/experimental/NativeDatePicker/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui-react-native/experimental-native-date-picker",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
Saadnajmi, line 3: These types of updates should be handled by beachball btw. You should never have to manually do it.

chripear: I don't want any trouble with the bot labor union... reverted.

Saadnajmi (approved):

PR #777, opened 2021-07-12
Activity Indicator

lindaweng:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Activity Indicator
- Implemented with a View wrapping an Animated SVG. Composed as a new experimental component.
- 3 tokens: size, line thickness, and color
  - Size defaults to "medium"
  - Line thickness defaults to same size as size
- Props animating and hidesWhen Stopped mirror the [React native ActivityIndicator](https://reactnative.dev/docs/activityindicator)

Accessibility
- Works in light and dark modes and colors match the vNext designs
  - Open issue: No support for high contrast yet. React native doesn't have support for high contrast and there are no designs for high contrast either.
- Voiceover works on iOS and Android
  - Open issue on iOS: `accessibilityRole` doesn't work for SVGs so may have to use `accessibilityLabel` even though it doesn't have support for localization in FURN yet
  - Update on this issue (Videos are also updated): `accessibilityRole` gets put into AccessibilityValue on iOS which is not localized. We are instead using and exposing `accessibilityLabel` so clients can add their own label depending on context and may localize this themselves.

### Usage
`<ActivityIndicator size="large" />`
Open issue: `customize` function does not work. It also does not work for Shimmer, so this will be a follow up.

### Verification

Tested with:
- Light/Dark mode
  - iOS by running from XCode
  - Android with dropdown menu in tester app
- Accessibility (Voice over)
  - iOS with Accessibility Inspector
  - Android with TalkBack
    - Minor detail being that since the entire SVG is animating, the square around the ActivityIndicator “bounces” (see below video)
- Different combinations of sizes, line thicknesses, and colors
  - When size is specified but line thickness is not and when both are specified with different sizes

#### iOS
https://user-images.githubusercontent.com/30802320/125316472-241b2600-e306-11eb-8264-2a225206c4ad.mov

https://user-images.githubusercontent.com/30802320/125983188-8b81286b-d634-4f40-8f08-a6bb7a5461ad.mov

#### Android
https://user-images.githubusercontent.com/30802320/125317176-cd621c00-e306-11eb-92ef-fdd957a1bf06.mov

https://user-images.githubusercontent.com/30802320/125983169-962f0355-819d-48f9-abd6-5323dfca78a6.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


Saadnajmi:
2 things: 
1) It looks like either the `activityindicator_prototype` branch or your branch are out of sync with master (causing the PR to show more edits than you actually did), so I would try to update those. I updated the former branch, but nothing seems to have changed on the GitHub PR.
2) Despite open issues, I think this PR would be OK to target master. At this point, it's no longer a prototype but a full composed component =)

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Stack } from '@fluentui-react-native/stack';
+import { stackStyle } from '../Common/styles';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { ACTIVITYINDICATOR_TESTPAGE } from './consts';
+
+const activityIndicatorTest: React.FunctionComponent<{}> = () => {
+  /** Bug: Customize doesn't do anything
+   * Tried having tokens not be props, but didn't work
+   * Not sure how to test/check where the tokens are passed through other than just looking at the final render
+   */
```
Saadnajmi, line 13: Rather than just "Bug", I would put "GH: ###" with the issue number so that we know where it is tracked. You can also just say the first line "Customize doesn't work" and put more info in the bug.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const CustomizedActivityIndicator = ActivityIndicator.customize({
+    activityIndicatorColor: '#458792',
```
Saadnajmi, line 15: Is there a more descriptive name for this hex value? We (Apex generally prefer to put names for hardcoded values for clarity.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Text>Small</Text>
+      <ActivityIndicator size="small" />
+      <Text>Medium</Text>
+      <ActivityIndicator size="medium" />
+      <Text>Large</Text>
+      <ActivityIndicator size="large" />
+      <Text>Extra Large</Text>
+      <ActivityIndicator size="xLarge" />
+
+      <Text>Size and Line Thickness props of different sizes</Text>
```
Saadnajmi, line 30: I would have a more descriptive label, something like "size=XLarge, lineThickness=large" so it's more obvious at a glance what changed.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <ActivityIndicator size="xLarge" lineThickness="large" />
+      <Text>Color props</Text>
+      <ActivityIndicator activityIndicatorColor="#921571" />
```
Saadnajmi, line 33: I wonder if we could make the prop named "color", and have the token "activityIndicatorColor" map to that prop. This is a nit/opinion and not a blocking issue imo, feel free to disagree.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Text>Animating: False, hidesWhenStopped: False</Text>
+      <ActivityIndicator animating={false} hidesWhenStopped={false} />
```
Saadnajmi, line 35: Could we have a section where we hook up the Boolean props to a switch? Similar to what I do [here](https://github.com/microsoft/fluentui-react-native/pull/682) (which I just copied elsewhere from FURN

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+  },
+];
+
+export const ActivityIndicatorTest: React.FunctionComponent<{}> = () => {
+  const status: PlatformStatus = {
+    win32Status: 'Backlog',
+    uwpStatus: 'Backlog',
+    iosStatus: 'Backlog',
+    macosStatus: 'Backlog',
+    androidStatus: 'Backlog',
```
Saadnajmi, line 58: Not that this is really updated anywhere else, but we can probably move iOS/Android out of backlog to "Beta" or something.

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  const description = 'ActivityIndicator is a visual representation that data is being loaded. It is made with an Animated SVG.';
```
Saadnajmi, line 61: nit: "It is implemented"

packages/experimental/ActivityIndicator/src/ActivityIndicator.styling.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ActivityIndicatorTokens,
+} from './ActivityIndicator.types';
+
+export const diameterSizeMap: { [key: string]: number } = {
+  xSmall: 12,
+  small: 16,
+  medium: 24,
+  large: 32,
+  xLarge: 36,
+};
```
Saadnajmi, line 15: @rurikoaraki Not an ask, but this may be a good candidate for a design token that is not a color to integrate from the pipeline come time. same for the line thickness below.

packages/experimental/ActivityIndicator/src/ActivityIndicator.styling.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    activityIndicatorName,
+  ],
+  slotProps: {
+    root: buildProps(
+      (tokens: ActivityIndicatorTokens) => ({
+        activityIndicatorColor: tokens.activityIndicatorColor,
+        lineThickness: tokens.lineThickness,
+        size: tokens.size,
+        // Bug: accessibilityRole does not work on iOS
```
Saadnajmi, line 39: Nit: If this won't be fixed before we check it in, File a GitHub issue and put the issue number here

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
+/** @jsx withSlots */
+import { useRef, useEffect, useMemo } from 'react';
+import { Animated, Easing } from 'react-native';
+import { Svg, Path } from 'react-native-svg';
+import { compose, mergeProps, withSlots, UseSlots, buildUseStyling } from '@fluentui-react-native/framework';
+import { activityIndicatorName, ActivityIndicatorProps, ActivityIndicatorType } from './ActivityIndicator.types';
+import { diameterSizeMap, lineThicknessSizeMap, stylingSettings } from './ActivityIndicator.styling';
+
+// Can also just put this within the compose, would that be better?
```
Saadnajmi, line 9: I think this is fine here. I have a feeling if you put it inside the `render` of compose, every time we call render, it would recreate this function.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+        size: props.size ? diameterSizeMap[props.size] : diameterSizeMap[tokens['size']],
+        lineThickness: props.lineThickness
+          ? lineThicknessSizeMap[props.lineThickness]
+          : props.size
+          ? lineThicknessSizeMap[props.size]
+          : lineThicknessSizeMap[tokens['lineThickness']],
+        animating: props.animating != undefined ? props.animating : true,
+        hidesWhenStopped: props.hidesWhenStopped != undefined ? props.hidesWhenStopped : true,
+      }),
+      [],
```
Saadnajmi, line 45: I think you need to add the keys here?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+      width: memoizedActivityIndicatorData.size,
+      height: memoizedActivityIndicatorData.size,
+      style: {
+        transform: [{ rotateZ: interpolateSpin }, { perspective: 10 }],
+        opacity: hideOpacity,
+      },
+      accessibilityState: { busy: memoizedActivityIndicatorData.animating },
+    };
+    return (rest: ActivityIndicatorProps) => {
+      const { ...mergedProps } = mergeProps(props, rest);
```
Saadnajmi, line 87: I think you can add `otherProps` here to the argument list to just have one set of mergedProps?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    // React Native ActivityIndicator still takes up space when hidden, so to perfectly match would use opacity
+    // hiding opacity makes the screen reader on iOS and Android skip over it
+    const hideOpacity = memoizedActivityIndicatorData.animating == false && memoizedActivityIndicatorData.hidesWhenStopped == true ? 0 : 1;
+
+    const path = getActivityIndicatorPath(
+      memoizedActivityIndicatorData.size,
+      memoizedActivityIndicatorData.lineThickness,
+      memoizedActivityIndicatorData.activityIndicatorColor as string,
+    );
```
Saadnajmi, line 74: I think this is actually something that would be better benefit from useMemo, since we'll probably generate the same path multiple times.

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
```diff
@@ -0,0 +1,50 @@
+import { ColorValue, Animated } from 'react-native';
+import { SvgProps } from 'react-native-svg';
+
+export const activityIndicatorName = 'ActivityIndicator';
+/**
+ * Specifies the possible sizes of the ActivityIndicator
+ */
+// Somehow need to convert these to numbers
```
Saadnajmi, line 8: Is this comment still relevant?

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
```diff
@@ -0,0 +1,50 @@
...
+  lineThickness?: ActivityIndicatorSize;
+  /**
+   * Size of the ActivityIndicator view
+   * @defaultValue 'medium'
+   */
+  size?: ActivityIndicatorSize;
+}
+
+export interface ActivityIndicatorSlotProps extends ActivityIndicatorProps {
+  root: Animated.AnimatedProps<SvgProps>;
```
Saadnajmi, line 30: :D

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const spinAnimation = useRef(new Animated.Value(0)).current;
+    useEffect(() => {
+      if (memoizedActivityIndicatorData.animating) {
+        Animated.loop(
+          Animated.timing(spinAnimation, {
+            toValue: 359,
+            duration: 750,
+            useNativeDriver: false,
```
Saadnajmi, line 55: Should this not be true? Or rather, we need to do a platform check here right?

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
lindaweng: I just put a random color that would stand out, but I can change it to a named color instead.

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
lindaweng: Nope.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: Whoops, thought that was true. I don't think we need to platform check because it only doesn't work for web which we aren't supporting anyways.

Saadnajmi: web, win32, and probably windows. I think an extra platform check would be a good way to avoid a confusing bug for someone maybe months down the line) who is trying to port this component to another platform and wondering why their code isn't working :D

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export interface ActivityIndicatorProps extends ActivityIndicatorTokens {
```
Saadnajmi, line 33: this should extend SVGProps

apps/fluent-tester/src/FluentTester/TestComponents/ActivityIndicator/ActivityIndicatorTest.tsx
lindaweng: I tried adding switches but the animation does not behave as expected. I think it's because of the combination of the `useState` and `useEffect` with the animation it's somehow not rerendering correctly. Even though it should since my condition is inside of the hook.

https://user-images.githubusercontent.com/30802320/125356254-f4831280-e333-11eb-89d4-602ca1a01736.mov

lindaweng: The weird jerking is because of native driver was set to true. If I set native driver to false, it stops jerking, but it never stops even when Animating is false. This is because there isn't a "stopAnimating" function, so when it re-renders, the Animated.loop isn't run but it's still animating from when Animating was true.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: I'm not sure I can put this within memo because it depends on size, line thickness, and color resolving themselves in memo first. If it's within memo, they will also need to resolve themselves as arguments to the getPath function.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,102 @@
...
+        opacity: hideOpacity,
+      },
+    };
+
+    const otherRootProps = {
+      style: {
+        opacity: hideOpacity,
+      },
+      accessibilityState: { busy: memoizedActivityIndicatorData.animating },
+    };
```
Saadnajmi, line 92: Let's say we removed the call to useMemo altogether, is the logic (minus interpolateSpin) being done here something that could be moved to SlotProps?

Saadnajmi: I ask because 
A) I still think the call to useMemo is mostly unnecessary and we can remove it
B) It feels like what we're doing here is taking the input from `ActivityIndicatorProps` and transforming it into the actual props we want to pass down to each slot, which is what `slotProps` is suppose to be for.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
Saadnajmi: Yeah that makes sense. IMO, it feels like the call to useMemo is getting in the way of writing a cleaner / more performant implementation, because we're caching the easy to compute stuff but not the hard-to-compute stuff (path). I would personally look to see if we can just remove it. When I used it in Avatar, there were different requirements around that control that encouraged us adding that call. I don't think it makes as much sense for ActivityIndicator anymore.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: Are you talking about the slotProps within the .styling.tsx? I don't think that file takes in any props that the user passes in. These props here all depend on what the user passes or doesn't pass in so I don't think it could be moved to slotProps

Saadnajmi: Ah, that's a good point. Thanks for the clarification!

Saadnajmi: Which reminds me, I think we need to set `tokensThatAreAlsoProps` in the .styling function

lindaweng: Do we still have to set that with props extending all of the tokens?

Saadnajmi: Yeah, looking at the readme, it looks like it copies the props that are in the `tokensThatAreAlsoProps` into tokens https://github.com/microsoft/fluentui-react-native/tree/9ae107336b8e6a20dffbfb802facf0f624729fe0/packages/experimental/use-styling

lindaweng: Wait, from the readme it seems like it should be `propsThatAreAlsoTokens` because it describes that as "Copy some/none/all props to Tokens" and `tokens = { ...tokens, ...props }`. This confuses me a bit because then, shouldn't I just put all of the "tokens" within props and not have anything within tokens? But I feel like that wouldn't work because then they aren't defined as tokens. I just feel like it's all going in a circle.

Saadnajmi: If I understand correctly, this is because there are multiple places that a token can come from:
- A Theme
- Passed in through `.customize()`
- Passed in as a prop

So you don't necessarily want all your tokens to just be props. You lose the ability for a theme to customize your control that way. That being said, `propsThatAreAlsoTokens` is valid. Fwiw, I think this used to be called `tokenProps` before I complained not knowing what that meant lol.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -69,30 +57,26 @@ export const ActivityIndicator = compose<ActivityIndicatorType>({
...
+    const hideOpacity = animating == false && hidesWhenStopped == true ? 0 : 1;
 
-    // Depends on size, line thickness and color after memoization so can't be in the memo itself
     const path = getActivityIndicatorPath(
-      memoizedActivityIndicatorData.size,
-      memoizedActivityIndicatorData.lineThickness,
-      memoizedActivityIndicatorData.activityIndicatorColor as string,
+      slotProps.root['size'],
+      slotProps.root['lineThickness'],
+      slotProps.root['activityIndicatorColor'],
```
Saadnajmi, line 65: slotProps.root.size doesn't work here?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -28,26 +28,14 @@ export const ActivityIndicator = compose<ActivityIndicatorType>({
...
-    );
+    const animating = props.animating != undefined ? props.animating : true;
+    const hidesWhenStopped = props.hidesWhenStopped != undefined ? props.hidesWhenStopped : true;
 
     const spinAnimation = useRef(new Animated.Value(0)).current;
     useEffect(() => {
-      if (memoizedActivityIndicatorData.animating) {
+      if (animating) {
         Animated.loop(
           Animated.sequence([
```
Saadnajmi, line 40: On my machine, I had to remove the sequence and set useNativeDriver=false in order for the animation to work properly. Is that not the case on yours?

packages/experimental/ActivityIndicator/src/ActivityIndicator.styling.tsx
```diff
@@ -0,0 +1,55 @@
...
+  tokensThatAreAlsoProps: 'all',
+  slotProps: {
+    root: buildProps(
+      (tokens: ActivityIndicatorTokens) => ({
+        activityIndicatorColor: tokens.activityIndicatorColor,
+        size: diameterSizeMap[tokens.size],
+        lineThickness: tokens.lineThickness != 'medium' ? lineThicknessSizeMap[tokens.lineThickness] : lineThicknessSizeMap[tokens.size],
+        accessibilityRole: 'progressbar',
+        accessible: true,
+        style: { width: diameterSizeMap[tokens.size] },
```
Saadnajmi, line 43: You don't need height here too?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: No, since size is not a normal prop in ViewProps and root is of type ViewProps.

packages/experimental/ActivityIndicator/src/ActivityIndicator.styling.tsx
lindaweng: The View automatically adjusts to the height of the ActivityIndicator, but the width is automatically the width of the stack. Setting the height would be fine, as a just in case.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+      }
+    });
+    const interpolateSpin = spinAnimation.interpolate({
+      inputRange: [0, 359],
+      outputRange: ['0deg', '359deg'],
+    });
+
+    // React Native ActivityIndicator still takes up space when hidden, so to perfectly match would use opacity
+    // hiding opacity makes the screen reader on iOS and Android skip over it
+    const hideOpacity = animating == false && hidesWhenStopped == true ? 0 : 1;
```
Saadnajmi, line 60: Nit, maybe move this after line 34 so all our props / local variable creation is in the same block

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+      slotProps.root['lineThickness'],
+      slotProps.root['activityIndicatorColor'],
+    );
+
+    // props for the AnimatedSvg, which is Slots.svg
+    const otherSvgProps = {
+      style: {
+        transform: [{ rotateZ: interpolateSpin }, { perspective: 10 }],
+      },
+    };
```
Saadnajmi, line 73: Nit: We can maybe remove the comment and give a more specific name like `animatedSVGProps`
Another nit: Since we mostly added a random value for perspective, could we put a comment explaining this was needed for Android?

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * Specifies the possible sizes of the ActivityIndicator
+ */
+export type ActivityIndicatorSize = 'xSmall' | 'small' | 'medium' | 'large' | 'xLarge';
+
+export interface ActivityIndicatorTokens {
+  /**
+   * ActivityIndicator element color
+   * @defaultValue 'BDBDBD' for light mode, '666666' for dark mode
+   */
+  activityIndicatorColor?: ColorValue;
```
Saadnajmi, line 15: Forgot to mention this earlier. Because SVG's don't support ColorValue yet, I think the type here is better to be 'string'.

Saadnajmi (approved):
Some minor comments aside, looks good! =)

packages/experimental/ActivityIndicator/src/ActivityIndicator.styling.tsx
Saadnajmi: The width/height difference there might be a side effect of flexbox and the <Stack> element that ActivityIndicator is hosted in. I think we should set both. Thanks for the clarification though =)

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: Starting from just this morning (which is weird since it worked yesterday), if useNativeDriver is true with the sequence, the indicator only spins once and then stops. All I have to do to get it completely working is setting useNativeDriver to false. I don't have to remove the sequence

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
lindaweng: Just curious, how do you know that svgs don't support ColorValue? And what's the difference?

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: Could the comment be taken out for cleaner code, or do you still think it's necessary to mention that screenreaders do skip over 0-opacity elements?

packages/experimental/ActivityIndicator/src/ActivityIndicator.types.tsx
lindaweng: Nevermind, I think I got it.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
Saadnajmi: I think the comment is cleaner code =) 
At the very least, if I came across this code without having seen this comment, I would have been tempted to rewrite it.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -82,7 +81,7 @@ export const ActivityIndicator = compose<ActivityIndicatorType>({
       const { ...mergedProps } = mergeProps(props, rest, otherRootProps);
       return (
         <Slots.root {...mergedProps}>
-          <Slots.svg {...otherSvgProps}>{path}</Slots.svg>
+          <Slots.svg {...animatedSvgProps}>{path}</Slots.svg>
```
Saadnajmi, line 84: Don't we also need to merge this with `slotProps.svg` ? Which I suppose is just width/height

Saadnajmi:
It looks like the PR checks are failing. To reproduce the error locally, you can run `yarn buildci` from root. It should be as simple as adding `react` and `react-native-svg` as dependencies to ActivityIndicator's package.json =)

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
Saadnajmi: Ah ok.. we're passing in values that don't exist on the type of `slotProps.root`. I think the fix there would be to change the type from `ViewProps` to `ViewProps & ActivityIndicatorTokens` 
``` 
export interface ActivityIndicatorSlotProps {
    root: ViewProps & ActivityIndicatorTokens;
    svg: Animated.AnimatedProps<SvgProps>;
 }
 ```

lindaweng: Yeah, that would work. I would then change the sizes from small, medium, etc. to the actual numbers inside of ActivityIndicator.tsx instead of within styling.

packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
lindaweng: No, because buildProps already adds those props within styling.tsx.

Saadnajmi: yes, but to get those in here, we need to grad it from UseStyling(props).svg, aka `slotProps.svg`, do we not?

lindaweng: No, the changes directly apply from .styling.tsx. I tried changing it to different heights and the changes show up. Also, you can see that we don't merge slotProps.root to get the accessibility and style to work on root either.

Saadnajmi: Oh, interesting. Thanks for the clarification!

PR #783, opened 2021-07-13
Extract macOS FocusZone implementation from devmain

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- Added NuGet and CocoaPods machinery to publish FocusZone library.
- Copied `RCTFocusZone` and `RCTFocusZoneManager` files from `devmain/word/endpoints/apple`.
- Stripped out [devmain-specific] feature gate and `AssertTag` calls.
- Fixed a couple of ARC issues (`autorelease` calls that needed to be removed)
- Addressed comments made in #735.
    - Extracted key codes `enum` to its own header.
    - Added a comment explaining `FocusZoneBuffer` usage.
    - Added a `typedef` for the block used by `EnumerateFocusableViews`.
    - Removed 'get' prefix from methods.
    - Passed `FocusZoneAction` enum values instead of `BOOL`s as arguments to several methods; added helper functions to compute the BOOL values from the enum value.
    - Changed an explicit large float value to CGFLOAT_MAX.
    - Added back a comment indicating a not-yet-implemented feature.

### Verification
- Checked baseline behavior of devmain Word Editor pane keyboard focus.
- Removed `RCTFocusZone` and `RCTFocusZoneManager` implementations from devmain and verified that Word Editor pane throws a React Native runtime error as a result.
- Verified Word Editor pane behavior is the same as the baseline behavior when using this version of RCTFocusZone and RCTFocusZoneManager linked from FURN instead of the version that is currently in devmain.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/FocusZone/macos/RCTFocusZone.h
```diff
@@ -0,0 +1,23 @@
...
+	FocusZoneDirectionNone
+} FocusZoneDirection;
+
+typedef enum {
+	NavigateStopAtEnds,
+} NavigateAtEnd;
+
+@interface RCTFocusZone : RCTView
+
+@property(nonatomic) BOOL disabled;
```
harrieshin, line 19: nit : isDisabled?

packages/components/FocusZone/macos/RCTFocusZoneManager.m
```diff
@@ -0,0 +1,56 @@
...
+	}
+	else
+	{
+		[view setNavigateAtEnd:[defaultView navigateAtEnd]];
+	}
+}
+
+- (RCTView *)view
+{
+  return [RCTFocusZone new];
```
harrieshin, line 53: its by design each time view is called we create a new object?

packages/components/FocusZone/macos/RCTFocusZoneManager.m
```diff
@@ -0,0 +1,56 @@
...
+
+RCT_CUSTOM_VIEW_PROPERTY(navigateAtEnd, NSString, RCTFocusZone)
+{
+	if ([json isEqualToString:@"NavigateStopAtEnds"])
+	{
+		[view setNavigateAtEnd:NavigateStopAtEnds];
+	}
+	else if ([json isEqualToString:@"NavigateWrap"] || [json isEqualToString:@"NavigateContinue"])
+	{
+		// Not yet implemented
```
harrieshin, line 42: GitHub ticket? or some other tracking mechanism? should be fail somehow obviously?

chripear: In devmain, this line was `AssertTag(false, 0x228146d7 /* tag_86u1x */);    // NYI`

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -0,0 +1,334 @@
+#import "KeyCodes.h"
+#import "RCTFocusZone.h"
+
+typedef enum {
```
harrieshin, line 4: nit : is there a limitation here where we can't use the recommended NS_ENUM(NSInteger, FocusZoneAction)  ?

chripear: Not that I know of.  I avoided changes unless they were specifically called out in the previous PR.

packages/components/FocusZone/macos/RCTFocusZone.h
```diff
@@ -0,0 +1,23 @@
...
+
+typedef enum {
+	FocusZoneDirectionBidirectional,
+	FocusZoneDirectionHorizontal,
+	FocusZoneDirectionVertical,
+	FocusZoneDirectionNone
+} FocusZoneDirection;
+
+typedef enum {
+	NavigateStopAtEnds,
```
harrieshin, line 14: why is this an enum?

packages/components/FocusZone/macos/RCTFocusZoneManager.m
chripear: In devmain, this line was `return [[RCTFocusZone new] autorelease];`.  Unknown what the design intent is here.  @nakambo might be able to comment.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -0,0 +1,334 @@
...
+
+static NSView *GetFirstResponder(NSWindow *window)
+{
+	NSResponder *responder = [window firstResponder];
+	while (responder != nil && ![responder isKindOfClass:[NSView class]])
+		responder = [responder nextResponder];
+	return [responder isKindOfClass:[NSView class]] ? (NSView *)responder : nil;
+}
+
+static void EnumerateFocusableViews(NSView *root, IsViewAtEndOfFocusableRange block)
```
harrieshin, line 56: IsViewAtEndOfFocusableRange is this meant to be boolean?

packages/components/FocusZone/macos/RCTFocusZone.h
chripear: This is the only one of three values that is implemented.  See the 'not yet implemented' part of RCTFocusZoneManager.mm.

packages/components/FocusZone/macos/RCTFocusZone.m
chripear: See the typedef.  It takes a pointer to BOOL as an output parameter, but returns void.

packages/components/FocusZone/macos/RCTFocusZoneManager.m
Saadnajmi: Yes, the purpose is to create a new object each time. That's mandated by native modules to not cache it.

Saadnajmi (approved):

packages/components/FocusZone/macos/RCTFocusZone.h
harrieshin: either this shouldn't be an enum because we have no plans supporting them. or it should be an enum and implement the rest (or plan on implementing it) also, the enum class names seems weird to the enum options. is it like navigation behavior? there should be also comments on these too.

packages/components/FocusZone/macos/RCTFocusZone.m
harrieshin: but the callers are directly doing bool * and only this function declaration uses the typedef which seems weird

packages/components/FocusZone/macos/RCTFocusZone.m
harrieshin: seems like a simple fix to try it out?

packages/components/FocusZone/macos/RCTFocusZone.m
chripear: Do you have a recommendation in mind?

For example, should callers define the block before using it so that this type name will appear there?
e.g.
```
IsViewAtEndOfFocusableRange block = ^(NSView *candidateView, BOOL *stop) {...};
EnumerateFocusableViews(self, block);
```

or should the block return a BOOL instead of having an output parameter?

packages/components/FocusZone/macos/RCTFocusZone.h
chripear: This property is exported and the name needs to match the JS API.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -0,0 +1,334 @@
...
+	FocusZoneActionNone,
+	FocusZoneActionTab,
+	FocusZoneActionShiftTab,
+	FocusZoneActionRightArrow,
+	FocusZoneActionLeftArrow,
+	FocusZoneActionDownArrow,
+	FocusZoneActionUpArrow,
+} FocusZoneAction;
+
+typedef void (^IsViewAtEndOfFocusableRange)(NSView *, BOOL *);
```
harrieshin, line 14: we should have the param names here so we know what it is used for.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -0,0 +1,334 @@
...
+	BOOL isAdvance = IsAdvanceWithinZoneAction(action);
+	BOOL isHorizontal = IsHorizontalNavigationWithinZoneAction(action);
+
+	NSView *firstResponder = GetFirstResponder([self window]);
+	NSRect firstResponderRect = [firstResponder convertRect:[firstResponder bounds] toView:self];
+
+	__block NSView *viewToFocus = nil;
+	__block CGFloat closestDistance = CGFLOAT_MAX;
+
+	EnumerateFocusableViews(self, ^(NSView *candidateView, BOOL *stop)
```
harrieshin, line 142: right so here you should be using the tyedef?

packages/components/FocusZone/macos/RCTFocusZone.m
harrieshin: right we should acutlaly use the typedef like you just said in the caller

packages/components/FocusZone/macos/RCTFocusZoneManager.m
harrieshin: what's is the convention in furn? we can't assert or nslog. preconditionfailure might be agressive but possible.

packages/components/FocusZone/macos/RCTFocusZone.h
chripear: Removed the enum

harrieshin (approved):

PR #755, opened 2021-07-01
chore: use react-native-test-app for Windows

tido64:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

This PR resumes the work of #552. Resolves #447.

- Migrates Windows test app to `react-native-test-app`
- Re-aligns `react-native-windows` with current `react-native` version
- Dedupes `metro` and friends
- Changed e2e tests to use the embedded bundles instead of dev server to avoid timeouts

### Verification

image

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [x] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

### Blocked by

- [x] https://github.com/microsoft/fluentui-react-native/issues/765
- [x] https://github.com/microsoft/fluentui-react-native/issues/767
- [x] Call stacks are pointing to `code` instead of the actual file on disk

PR #785, opened 2021-07-13
Disable "gap" prop on Stack when running in Chakra

### LINKED ISSUE #767: Assertion hit in `deepFreezeAndThrowOnMutationInDev` when navigating away from Separator page
While investigating a crash in the [PR migrating the Windows test app](https://github.com/microsoft/fluentui-react-native/pull/755) to `react-native-test-app`, I found that navigating away from the Separator page soft-crashes the test app on an assertion in `deepFreezeAndThrowOnMutationInDev`.

### Environment Information

- **Package version(s)**: `master`
- **OS version**: Windows 10 Enterprise 21H1 19043.1083

- **Platform**:
  - [ ] iOS
    - **Xcode version**:
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [x] windows
  - [ ] android

1. Check out this branch: https://github.com/microsoft/fluentui-react-native/pull/755
2. Build and run windows test app
    ```sh
    yarn
    yarn build
    cd apps/windows
    yarn install-windows-test-app
    start windows/FluentTester.sln
    yarn start
    ```
3. Once the app starts, click on the Separator page, then navigate to any other page, e.g. Button

#### Actual behavior:

The test app soft-crashes in `deepFreezeAndThrowOnMutationInDev` and turns blank.

A workaround is to disable `gap`:

```diff
diff --git a/apps/fluent-tester/src/FluentTester/TestComponents/Separator/SeparatorTest.tsx b/apps/fluent-tester/src/FluentTester/TestComponents/Separator/SeparatorTest.tsx
index e0cf9e1b..18df01ba 100644
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Separator/SeparatorTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Separator/SeparatorTest.tsx
@@ -11,8 +11,8 @@ const RedSeparator = Separator.customize({ color: 'red' });

 const separator: React.FunctionComponent<{}> = () => {
   return (
-    <Stack style={stackStyle} gap={5}>
-      <Stack gap={4} style={separatorStackStyle}>
+    <Stack style={stackStyle} gap={0}>
+      <Stack gap={0} style={separatorStackStyle}>
         <Button content="Button4" />
         <BlueSeparator vertical />
         <Button content="Button5" />
```

#### Expected behavior:

It doesn't soft-crash.

Requested priority: High

rurikoaraki:
Fixes #767 

### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

A workaround fix for an issue with Charka, RN, and Stack where we were seeing errors while running the RN Test App for Windows.

This issue seems to be a timing issue with using Chakra. While the code in the Stack component isn't inherently doing anything that isn't allowed by RN, it trips up a function which runs in debug bundles. We can't switch away from Chakra in UWP since the base test app, which we'll be switching to with #755, runs with Chakra and probably won't be using Hermes until 0.65. For now, we can disable the `gap` prop causing the issue in Chakra. Since Edge has switched to Chromium and RNW is trying to switch to Hermes, Chakra is less popular in usage so we think it's okay to not have the prop working when running Chakra.

I've tried other workarounds such as using Object.defineProperty directly, or using View components instead of marginTop/marginLeft to implement the spacing, to no avail. Short of removing the prop entirely, this is the best solution we have for this issue.

### Verification

Tested fix in UWP test app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Stack/src/Stack.tsx
+++ b/packages/components/Stack/src/Stack.tsx
@@ -12,6 +12,7 @@ import { backgroundColorTokens, borderTokens } from '@fluentui-react-native/toke
 import { buildStackRootStyles, buildStackInnerStyles } from './Stack.tokens';
 import { StyleProp, ViewStyle } from 'react-native';
 
+// Needed for TS to understand that __jsiExecutorDescription exists.
 declare global {
   /* eslint-disable-next-line @typescript-eslint/no-namespace*/
   namespace NodeJS {
@@ -30,12 +31,7 @@ const _styleKey = 'style';
 const render = (Slots: ISlots<IStackSlotProps>, renderData: IStackRenderData, ...children: React.ReactNode[]): JSX.Element => {
   const { gap, horizontal, wrap } = renderData.state!;
 
-  if (
-    gap &&
-    gap > 0 &&
-    children &&
-    (!global.hasOwnProperty('__jsiExecutorDescription') || global.__jsiExecutorDescription !== 'ChakraRuntime')
-  ) {
+  if (gap && gap > 0 && children && global.__jsiExecutorDescription !== 'ChakraRuntime') {
     const extraStyle: ViewStyle = horizontal ? { marginLeft: gap } : { marginTop: gap };
     /* eslint-disable @typescript-eslint/ban-ts-ignore */
     // @ts-ignore - TODO, fix typing error
--- a/packages/components/Stack/src/Stack.tsx
+++ b/packages/components/Stack/src/Stack.tsx
@@ -28,10 +28,11 @@ const render = (Slots: ISlots<IStackSlotProps>, renderData: IStackRenderData, ..
     children = React.Children.map(children, (child: React.ReactChild, index: number) => {
       if (React.isValidElement(child) && index > 0) {
         const childProps = child.props;
-        const mixedStyle = _mixinStyle(childProps[_styleKey], extraStyle);
-        const styleProp = Object.create({});
-        Object.defineProperty(styleProp, 'style', { value: mixedStyle, configurable: true });
-        return React.cloneElement(child, styleProp);
+        const extraProps = { style: _mixinStyle(childProps[_styleKey], extraStyle) };
+        return React.cloneElement(child, {
+          ...childProps,
+          ...extraProps,
+        });
       }
       return child;
     });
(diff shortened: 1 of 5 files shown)


packages/components/Stack/src/Stack.tsx
```diff
@@ -28,11 +28,10 @@ const render = (Slots: ISlots<IStackSlotProps>, renderData: IStackRenderData, ..
...
       if (React.isValidElement(child) && index > 0) {
         const childProps = child.props;
-        const extraProps = { style: _mixinStyle(childProps[_styleKey], extraStyle) };
-        return React.cloneElement(child, {
-          ...childProps,
-          ...extraProps,
-        });
+        const mixedStyle = _mixinStyle(childProps[_styleKey], extraStyle);
+        const styleProp = Object.create({});
+        Object.defineProperty(styleProp, 'style', { value: mixedStyle, configurable: true });
```
Saadnajmi, line 33: Nit: Maybe add a comment about why these two lines are needed with an issue # ?

Saadnajmi (approved):

packages/components/Stack/src/Stack.tsx
rurikoaraki: Solution has changed, since this actually didn't fix the bug.

packages/components/Stack/src/Stack.tsx
```diff
@@ -21,7 +30,12 @@ const _styleKey = 'style';
 const render = (Slots: ISlots<IStackSlotProps>, renderData: IStackRenderData, ...children: React.ReactNode[]): JSX.Element => {
   const { gap, horizontal, wrap } = renderData.state!;
 
-  if (gap && gap > 0 && children) {
+  if (
+    gap &&
+    gap > 0 &&
+    children &&
+    (!global.hasOwnProperty('__jsiExecutorDescription') || global.__jsiExecutorDescription !== 'ChakraRuntime')
```
rurikoaraki, line 37: TODO: Need to add a comment for this

Also fix the experimental Stack.

tido64: `__jsiExecutorDescription` would just be `undefined` if it isn't set, so you can drop the first check:

```suggestion
    global.__jsiExecutorDescription !== 'ChakraRuntime'
```

rurikoaraki: Thanks, change has been made.

Saadnajmi (approved):

tido64 (approved):

PR #788, opened 2021-07-15
Callout DismissBehaviors

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Callouts may need certain behaviors that are controlled by the native platform projection; partially due to the nature of the platform host and partially due to the async nature of React-Native.  These properties help control the behavior of the Callout for scenarios that need it.

Note that these are strong suggestions and not hard rules; again due to the nature of the platform host.

Mind that `preventDismissOnKeyDown` does not prevent escape, and has limited use outside of keyboard navigation purposes (between the callout surface and other surfaces).

### Verification

Added tests to CalloutTest page.

### DIFF
--- a/apps/win32/package.json
+++ b/apps/win32/package.json
@@ -21,6 +21,7 @@
     "bundle-dev": "react-native rnx-bundle",
     "run-win32": "rex-win32 --bundle index.win32 --component FluentTester --windowTitle \"FluentUI Tester\" --basePath ./dist --pluginProps --debugBundlePath index --jsEngine v8",
     "run-win32-web": "rex-win32 --bundle index --component FluentTester --basePath ./dist --useWebDebugger --windowTitle \"FluentUI Tester\" --useFastRefresh --pluginProps --debugBundlePath index --jsEngine v8",
+    "run-win32-dev-web": "rex-win32 --bundle index --component FluentTester --basePath ./dist --useDevMain --useWebDebugger --windowTitle \"FluentUI Tester\" --useFastRefresh --pluginProps --debugBundlePath index --jsEngine v8",
     "run-win32-devmain": "rex-win32 --bundle index.win32 --component FluentTester --basePath ./dist --useDevMain --windowTitle \"FluentUI Tester\" --pluginProps --debugBundlePath index --jsEngine v8",
     "e2etest": "rimraf reports/* && wdio",
     "report": "allure generate allure-results --clean",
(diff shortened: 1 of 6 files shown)


packages/components/Callout/src/Callout.types.ts
```diff
@@ -26,6 +26,8 @@ export type DirectionalHint =
   | 'bottomCenter'
   | 'bottomRightEdge';
 
+export type DismissBehaviors = 'preventDismissOnKeyDown' | 'preventDismissOnClickOutside';
```
lenahong, line 29: Would defining a third behavior 'preventAllDismissal' cover all possible behaviors?? That way users don't have to pass in an array and just a string value.

lenahong (approved):

apps/win32/package.json
```diff
@@ -21,6 +21,7 @@
     "bundle-dev": "react-native rnx-bundle",
     "run-win32": "rex-win32 --bundle index.win32 --component FluentTester --windowTitle \"FluentUI Tester\" --basePath ./dist --pluginProps --debugBundlePath index --jsEngine v8",
     "run-win32-web": "rex-win32 --bundle index --component FluentTester --basePath ./dist --useWebDebugger --windowTitle \"FluentUI Tester\" --useFastRefresh --pluginProps --debugBundlePath index --jsEngine v8",
+    "run-win32-dev-web": "rex-win32 --bundle index --component FluentTester --basePath ./dist --useDevMain --useWebDebugger --windowTitle \"FluentUI Tester\" --useFastRefresh --pluginProps --debugBundlePath index --jsEngine v8",
```
samuelfreiberg, line 24: Maybe add a quick one-liner to the docs if this adds a new debugging capability?

packages/components/Callout/src/Callout.types.ts
PPatBoyd: Nah.  I don't want to suggest that a truly persistent callout is currently available.

PPatBoyd: it'd still need to be an array anyway because these behaviors can be individually relevant

PPatBoyd: and there'll be more later

apps/win32/package.json
PPatBoyd: eh it's just combining the -devmain and -web commands that should be pretty frequently used

samuelfreiberg (approved):

PR #793, opened 2021-07-20
Update Contributing guide and Tester app launch guides

lindaweng:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

When creating a new component, we discovered that the contributing guide wasn't complete in how to package the component and make it into an importable module. That section is now added for reference for future development.

Also, when running the iOS, macOS, and Android tester apps on an M1 Mac, some new steps are needed. Those are added to their respective launch guides to keep them up to date.
- They mainly stem from the fact that M1 uses an arm processor instead of x86

### Verification

Documentation only update

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


CONTRIBUTING.md
```diff
@@ -67,6 +67,23 @@ Most components should use the compose framework as it offers the comprehensive
    1. `<new-component>.<types | settings | platform | blah>.tsx` (Optional)
       - Optional extra files to subdivide your code however you see fit. You can also add platform specific files as you see fit.
 
+### Packaging your component (or make it into an importable module)
```
harrieshin, line 70: I think there should be brief explanation why we are packing before going into how

apps/android/README.md
```diff
@@ -9,6 +9,7 @@ Prereqs:
 - FluentUI Tester is built uses react-native-test app, so install its [prereqs](https://github.com/microsoft/react-native-test-app#react-native-test-app) and follow instructions depending on whether you're developing on Windows or macOS.
 - Install [Android Studio](https://developer.android.com/studio)
 - Make sure you have setup an [Android Virtual Device](https://developer.android.com/studio/run/managing-avds)
+  - M1 Macs: Current best emulator is to still download from Github, [version 3](https://github.com/google/android-emulator-m1-preview)
```
harrieshin, line 12: nit instead of m1 should we say apple silicon incase there will be m2 chips etc in the future?

harrieshin: applies for other documentation as well

Saadnajmi: I would also put a date. Something like "As of July 2021, if you are using an Apple Silicon Mac, ...

CONTRIBUTING.md
```diff
@@ -67,6 +67,23 @@ Most components should use the compose framework as it offers the comprehensive
...
+   1. `yarn build`
+   1. `cd apps/ios/src`
+   1. `pod install` since there are Apple components that need to be reconnected
+   1. Restart VSCode/Editor
+   1. Close packager
+   1. Close simulator
+   1. `cd apps/ios`
+   1. `yarn ios`
+1. Add your new module as a dependency in `apps/fluent-tester/package.json`
+1. Note: Make sure that all naming is consistent: the component you are exporting, the file name you are exporting, and what you are importing
```
Saadnajmi, line 85: I'm not sure we actually need _all_ of these steps. I think this can be condensed to something like: "Add your new package to `apps/fluent-tester/package.json` so that we can import and test it, and run `yarn && yarn build`. If you your component has native apple code, remember to run `pod install`."

apps/android/README.md
```diff
@@ -23,6 +24,16 @@ yarn build
 yarn android
 ```
 
+M1 Macs:
```
Saadnajmi, line 27: Rather than have have 3 different additions, I think it might make sense to just have one section like "If you are using an Apple Silicon Mac" and condense the 3 sections into one.

apps/macos/README.md
```diff
@@ -14,6 +14,7 @@ yarn build
 ```
 
 2. Then go into `apps/macos/src` folder and run pod install to pull in the project-level Cocoapod dependencies defined in the podfile, and to generate a valid xcworkspace:
+   - M1 Macs: If you get the error: "missing compatible arch", you may need to add `alias pod='arch -x86_64 pod'` to your .bash_profile or .zprofile
```
Saadnajmi, line 17: Wouldn't this apply to the iOS app too?
Also, ".zprofile or .bash_profile". Or better yet, link a stack overflow article/cocoapod issue # that explains why we have to do it, so if in the future this has changed, the reader can know by clicking on the link.

CONTRIBUTING.md
lindaweng: I just added all of the steps that I did to get it working, but it looks like deleting and restarting everything isn't needed. Is there somewhere else where the "deleting cache" steps could go? Sometimes bugs go away once things are reset.

Saadnajmi: I've seen the metro bundler spit out these steps whenever I run into an error running it, so I think that's something that is easiest enough to find yourself.

You could put it in a new troubleshooting section too if you like.

apps/macos/README.md
lindaweng: Yes, I also put it in the iOS readme.
Do you just mean switching the order of the profiles?
[This](https://github.com/CocoaPods/CocoaPods/issues/10518#issuecomment-798912624) was the best GitHub issue I found documenting the workaround with the `arch -x86_64` but I don't think the thread will be updated in the future since it's closed.

Saadnajmi: Oh I just noticed you listed both. I only read ".zprofile" the first time I looked at this so I wanted you to mention both.
I think this is the most "official" bug for it https://github.com/CocoaPods/CocoaPods/issues/9907
It also sounds like if you install CocoaPods with `brew` it's fixed now? I might try that.

lindaweng: Yep, I just tried deleting the cache and then `pod install` without the alias in .zprofile and it works! Also I'm pretty sure the `sudo gem install cocoa pods` from setting up React works as well. I don't think we need this anymore.

CONTRIBUTING.md
lindaweng: From shifting this around, I'm gonna cut this section and move it to the last step after "Adding a new test". It fits a bit better there and more logical that you new component needs to be packaged in order to be used in the new test you wrote.

CONTRIBUTING.md
```diff
@@ -70,9 +70,12 @@ Most components should use the compose framework as it offers the comprehensive
...
-1. If you're component contains a native module, you will also need to add the path to your component's `podspec` (that we created earlier) in your test app's Podfile. This extra step is due to the fact that FluentTester has separate platform test apps that each share a common JS package. The react native community CLI does not support [autolinking transitive dependencies](https://github.com/react-native-community/cli/issues/1347), so we need to add it manually here.
+1. If your component contains a native module, you will also need to add the path to your component's `podspec` (that we created earlier) in your test app's Podfile. This extra step is due to the fact that FluentTester has separate platform test apps that each share a common JS package. The react native community CLI does not support [autolinking transitive dependencies](https://github.com/react-native-community/cli/issues/1347), so we need to add it manually here.
+1. Add your new module as a dependency in `apps/fluent-tester/package.json` and run `yarn && yarn build` from the root folder.
+   1. If your component has native apple code, run `pod install`.
```
Saadnajmi, line 77: nit: Run pod install on the corresponding macOS or iOS test app.

Saadnajmi (approved):
Overall looks good to me. I'd like some other eyes on it before we merge it though. Tagging @chripear @chiuam

apps/android/README.md
```diff
@@ -23,6 +23,25 @@ yarn build
...
+```sh
+adb reverse tcp:8081 tcp:8081
+yarn start
+```
+
+- APKs: The emulator is running on arm64 so any APKs that need to be installed need to be for arm64. Here is a [link](https://www.apkmirror.com/apk/google-inc/android-accessibility-suite/android-accessibility-suite-9-1-0-358315219-release/android-accessibility-suite-9-1-0-358315219-android-apk-download/) to the accessibility APK which includes TalkBack (voiceover)
+  - The ADB location is not automatically detected on the emulator so you need to manually set it.
+    1.  Go to expanded settings (three dots on the side) -> settings
+    2.  Turn off "Use detected ADB location"
+    3.  Click the folder and find the location of adb (~/Library/Android/sdk/platform-tools/adb). You will need to show hidden folders (Shift + CMD + .) to see Library
```
harrieshin, line 42: it seems bit inconsistent some we talk about "cd foo" on the terminal sometime we talk about finder guy stuff. should we just unify and just do all terminal commands? also, "ls -l" also shows hidden files in terminal if you need the equivalent.

lindaweng: This specifically is within the Android emulator settings panel so terminal commands won't work. Also for this instance, the field won't accept a copy/pasted file location so you're required to go through finder to find the location.

harrieshin (approved):

CONTRIBUTING.md
chripear: This is probably the second document that a newbie will read in this repo.  At that point I do not expect it will be clear what "on the corresponding macOS or iOS test app" means.  I would like this to be more explicit, e.g. "from the macOS or iOS test app `src` folder (`apps/macos/src` or `apps/ios/src` respectively)."

lindaweng: That makes sense. I personally first ran the test apps first which requires you to run `pod install`, but it is nice to clarify to ensure everyone understands what to do.

chiuam (approved):

chripear (approved):

PR #798, opened 2021-07-21
Add alias tokens plumbing to FURN theming

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Adding alias tokens to FURN theming.
- Added additional alias tokens based on web values and names. These names mostly match mobile so we should be able to use them for Android/iOS. Windows/Mac aren't ready yet, so this is mostly for proof of concept use.
- Added plumbing to create alias tokens to be added to theme.color when default color themes are created.
- Added test page for tokens to show list of alias tokens and their values.

Selected colors are not included as there is some differences between how pipeline handles selected colors and how design thinks of selected colors.

### Verification

Verfied alias tokens change values when moving to dark mode
Alias values currently do not take Office theme into account when determining color values.

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-tokens/src/getAliasTokens.ts
```diff
@@ -0,0 +1,10 @@
+import * as lightAliasTokens from './light/tokens-alias';
+import * as darkAliasTokens from './dark/tokens-alias';
+import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+
+export function getAliasTokens(mode: AppearanceOptions) {
+  if (mode === 'light') return lightAliasTokens.default;
+  else if (mode === 'dark') return darkAliasTokens.default;
+
+  return lightAliasTokens.default;
```
Saadnajmi, line 9: I'm not sure if typescript will let us, but could we make this a switch statement without a default case so it _has_ to fill every case and therefore protect us against future AppearanceOption additions?

packages/theming/default-theme/src/defaultColors.ts
```diff
@@ -138,6 +139,7 @@ export function getStockWebPalette(): ThemeColorDefinition {
       accent: globalTokens.color.brand.primary,
       blackTranslucent40: 'rgba(0,0,0,.4)',
     }),
+    ...createAliasTokens('light'),
```
Saadnajmi, line 142: I like that we could add this on to the defaultColors =) I wonder if we could remove the `paletteFromFabricColors` call above in the future if nothing _actually_ depends on those colors.

apps/fluent-tester/src/FluentTester/testPages.web.ts
```diff
@@ -61,4 +62,9 @@ export const tests: TestDescription[] = [
     component: PersonaTest,
     testPage: HOMEPAGE_PERSONA_BUTTON,
   },
+  {
+    name: 'Tokens Test',
+    component: TokenTest,
+    testPage: HOMEPAGE_TOKEN_BUTTON,
+  },
```
Saadnajmi, line 69: Could we add this to all the platforms? Or will it not work / hasn't been tested yet? I can test mobile + macOS later if that helps

packages/theming/theme-tokens/src/getAliasTokens.ts
rurikoaraki: I'll have to look into whether TS natively has something like "assertNever" (which basically does what you're asking for), I'll get back to you

apps/fluent-tester/src/FluentTester/testPages.web.ts
rurikoaraki: I think it'll work? I just didn't verify outside of win32. I can add it to other platforms and test uwp.

Saadnajmi: I'll clone your branch and test it on macOS and iOS (and maybe Android, that one needs some additional setup from me) later tonight. It would be nice to see the token test show up on each platform since these are definitely cross platform.

packages/theming/theme-tokens/src/getAliasTokens.ts
rurikoaraki: https://github.com/microsoft/fluentui-react-native/pull/799

Saadnajmi:
When I run this locally, I get two warnings (I get the same ones with the Theme test page. I'm not sure if they show up on win32 but would we be able to fix these two if we're going to make a new test page?

> 'VirtualizedList: missing keys for items, make sure to specify a key or id property on each ' + 'item or provide a custom keyExtractor.'

> VirtualizedLists should never be nested inside plain ScrollViews with the same ' + 'orientation - use another VirtualizedList-backed container instead.

rurikoaraki:
@Saadnajmi Thanks for testing this out! I can definitely deal with the first error, less sure about the second, but I'll take a look.

packages/theming/default-theme/src/defaultColors.ts
rurikoaraki: Maybe in the distant future? But it's possible clients use those color entries on a theme so it wouldn't necessarily be safe to just randomly rip them out. I'd have a bit more confidence if fluentui stops using them/deprecates them, since they're based off the fluentui Semantic colors.

Saadnajmi: fluentui like the web one? That makes sense.

rurikoaraki: Yeah, the web one. You can see those colors here: https://github.com/microsoft/fluentui/blob/dab45a2afe609709735b9b6e604a1af40d50e809/packages/theme/src/types/ISemanticColors.ts

Saadnajmi: Thinking about this more, I think if we _did_ want to deprecate those colors (along with any others) in favor of our tokens, we would just have to rewrite the themes/components we export to not use those colors (default, win32, android, apple themes) and then mark the set as deprecated? Definitely out of scope though I'm just musing.

rurikoaraki:
@Saadnajmi So I got rid of one of the warnings but the other one (VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead. ) is because we're using a FlatList inside a ScrollView.

Unfortunately that's a result of how the test app is structured - a ScrollView hosts the Test Page content. We /could/ get rid of the FlatView that's used to render all the swatches but then all of the swatches would be rendered at once and that's pretty costly since the color object is pretty big. We could also force the outer ScrollView to be a FlatList or something which would be kinda wonky, but otherwise this is what we have 😕 Thoughts?

packages/theming/default-theme/src/defaultColors.ts
rurikoaraki: Yeah I think that's how it would go.

FalseLobster: I am wondering if there's a better place to introduce the alias tokens than in the paletteFromFabricColors function considering this function is sort of on a deprecation path.  Let me know if you want a more specific suggestion and I'll look into it

FalseLobster (approved):

Saadnajmi:
> @Saadnajmi So I got rid of one of the warnings but the other one (VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead. ) is because we're using a FlatList inside a ScrollView.
> 
> Unfortunately that's a result of how the test app is structured - a ScrollView hosts the Test Page content. We /could/ get rid of the FlatView that's used to render all the swatches but then all of the swatches would be rendered at once and that's pretty costly since the color object is pretty big. We could also force the outer ScrollView to be a FlatList or something which would be kinda wonky, but otherwise this is what we have 😕 Thoughts?

Yeah I noticed that too after looking into it, and it feels very out of scope to restructure the test page like we would need to. It doesn't look like there's a simple fix for this issue :/ But I am really happy to see one of the warnings fixed so thank you! :D

Saadnajmi (approved):

packages/theming/default-theme/src/defaultColors.ts
rurikoaraki: I can think of adding it to createOfficeTheme as a ThemeRecipe for the ThemeReference, but otherwise can't really think of anything better. If you have any better suggestions, let me know, but that's what I'll pivot to.

rurikoaraki: I think this works well actually to set up for the "layering" or "recipe" part of putting the tokens together. The ThemeRecipe approach allows us to have information about the Office host available to determine alias theme colors.

PR #799, opened 2021-07-21
Add assert never util

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add an assertNever util to FURN. The util aims to help with ensuring that we are handling all cases of a type union or an enum. If you add this to an if...else or switch case, then if a new case is added to those blocks, they will start throwing build-time errors to alert you to having to handle another case.

### Verification

Code builds. If I add a new type to Shimmer then the code has a build error.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/utils/assert-never/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Example
+
+```ts
+enum A {
+  foo,
+  bar,
+}
+
+...
```
Saadnajmi, line 13: Could we also add an example with a union type since that's more common in typescript / gives us the built in type checking? Something like?

```
type Size = 'small' | 'medium' | 'large' 

const ringSize: Size = ...

 switch(ringSize) {
   case 'small':      
    return 'small';
   case 'medium':
     return 'medium';
   case 'large':
     return 'large';
   default:
     assertNever(ringSize); // Will cause build error if another entry is added to Size
 }

```

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -84,6 +85,8 @@ export const Shimmer = compose<ShimmerType>({
             );
           } else if (element.type == 'circle') {
             rows.push(<Circle key={i} r={element.height / 2} cx={element.xPos} cy={element.yPos} />);
+          } else {
+            assertNever(element.type);
```
Saadnajmi, line 89: Nice to add a use case elsewhere in the repo so we have more code to reference =)

Saadnajmi (approved):

Saadnajmi:
I didn't realize this would take a whole new package, but I suppose it's a lightweight and useful addition

rurikoaraki:
Yeah the other option to use add a dependency to a different package, but it has functionality that we probably wouldn't use so I figured this would be lighter.

packages/experimental/Shimmer/src/Shimmer.tsx
rurikoaraki: Unfortunately this was the only place where it seemed to be applicable, because a lot of other cases are really flexible 😓

packages/utils/assert-never/README.md
rurikoaraki: Sure, can do.

packages/utils/assert-never/src/assertNever.ts
```diff
@@ -0,0 +1,3 @@
+export function assertNever(x: never): never {
```
JasonVMo, line 1: It looks like this does two things. the x: never clause will cause a build-time break, whereas the throw statement will cause a runtime exception. Maybe consider adding an additional routine that is expectNever which causes the build-time error but allows for execution to continue. The scenario here would be places where things end up being forwarded (like a middle component in a chain where the above and below components both have more options. Just a thought.

JasonVMo (commented):
These changes look good to me, one suggestion that you can take or leave.

packages/utils/assert-never/src/assertNever.ts
rurikoaraki: Yeah there's a package out there that does exactly what you're talking about (https://github.com/aikoven/assert-never). I've never run into a scenario where we've run into the throwing code, actually, so I was wondering what the use case was for having the option to bypass the throw. I can just pull that package in instead so we have that option when we need it

rurikoaraki:
I'll just pull in the assertNever package that already exists and use that. That way we'll be able to bypass throwing if we need to.

PR #800, opened 2021-07-22
@types/react dependency cleanup

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Edit package.json dependencies to specify a maximum @types/react version.

Since our react version is fixed to 16.13.x, our @types/react version should also be 16.13.x at maximum to make sure our versions match (in reality, @types/react has version 16.9.x and then jumps to 16.14, so we keep 16.9.x instead). This will also keep the bots from speeding ahead too far in versions.

Jumping @types/react versions caused some changes as there was a change in how React.ElementRef was defined in DT.

### Verification

Code compiles.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

Saadnajmi:
@tido64 Is this one of those things that rnx-depcheck would guard against in the future?

tido64:
> @tido64 Is this one of those things that rnx-depcheck would guard against in the future?

We've only integrated dep-check in the test apps so far. I'll see if I can do the rest.

tido64 (approved):
Thanks for cleaning this up ❤️

PR #802, opened 2021-07-23
Add submenu chevron as inline svg

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- Create an inline svg and add it to chevron slot in SubmenuItem
- Support chevronColor token for customization

### Verification
chevron

### Pull request checklist

- [X] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/package.json
```diff
@@ -30,7 +30,8 @@
     "@fluentui-react-native/text": ">=0.9.12 <1.0.0",
     "@fluentui-react-native/tokens": ">=0.9.8 <1.0.0",
     "@uifabricshared/foundation-composable": ">=0.8.4 <1.0.0",
-    "@uifabricshared/foundation-settings": ">=0.9.2 <1.0.0"
+    "@uifabricshared/foundation-settings": ">=0.9.2 <1.0.0",
+    "react-native-svg": "^12.1.1"
```
Saadnajmi, line 34: I think we prefer `react-native-svg` as a peer dependency, so that we don't accidentally bring in our own version of this package in case a client already packages it. It's similar to how we moved react/react-native/etc to peerDependencies

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -124,14 +131,15 @@ export const SubmenuItem = compose<SubmenuItemType>({
     icon: { slotType: Icon as React.ComponentType<object> },
     content: Text,
     rightstack: { slotType: View },
-    chevron: { slotType: Icon as React.ComponentType<object> }
+    chevron: SvgXml
```
Saadnajmi, line 134: What's the difference between `SvgXml` and `SVG`? I've seen the latter as a slot but not the former.

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -98,12 +99,18 @@ export const SubmenuItem = compose<SubmenuItemType>({
       },
       content: { children: text, testID },
       icon: createIconProps(icon),
+      chevron:{}
```
rurikoaraki, line 102: nit: there should be a space between the colon and open bracket here. If you run prettier on the file, that should automatically get fixed for you.

If you have VSCode you can configure it to run formatting on save. There's instructions in the [main README](https://github.com/microsoft/fluentui-react-native#prettier).

packages/components/ContextualMenu/src/SubmenuItem.tsx
lenahong: It extends the Svg interface but take an xml value as its source

rurikoaraki (approved):

ValentinaKozlova (approved):

PR #806, opened 2021-07-23
Create use-slot package

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In a similar manner to the use-tokens package, this continues making the framework more granular and usable without the full wrapping of compose. This is a backwards compatible change, factoring out functionality from the use-slots package, putting it in a new use-slot package, then reworking that package to depend on this one. Snapshots are identical between the two.

The new package creates a hook function called `useSlot` that allows slots to be used individually without the framework. Usage looks like:

```tsx
const Container = useSlot(View, viewProps);
return <Container>{children}</Container>;
```

If pluggability is desired (when used manually), you can do something like this:

```tsx

type OverrideableSlot = MyComponentProps & { override?: React.FunctionComponent<MyComponentProps> };

const SimpleControl = (props: OverrideableSlot) => {
  const { children, override, ...rest } = props;
  const Container = useSlot(override || View, rest);
  return <Container>{children}</Children>
}
```

The `useSlot` hook provides layer compression (if the slot type supports it) and pluggability. There is no need to use it if neither of these things are required. E.g. a `View` that is not pluggable doesn't need a slot, it can be rendered normally. A HOC which has other custom components as children may benefit, even if the slot isn't part of the API.

Note that if you are using `compose` all the previous scenarios still work.

### Verification

Tested via automated tests

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/composition/package.json
```diff
@@ -35,6 +35,7 @@
   "license": "MIT",
   "dependencies": {
     "@fluentui-react-native/immutable-merge": "^1.1.2",
+    "@fluentui-react-native/use-slot": ">=0.1.1 <1.0.0",
     "@fluentui-react-native/use-slots": ">=0.5.3 <1.0.0",
```
rurikoaraki, line 39: Oof, the names of these packages being really similar is unfortunate. What do you think about adding "framework-" in front of "use-slot" to differentiate it a little better?

JasonVMo: They definitely are. Long term though if we are moving away from compose as the main way to write components, then the use-slots package can go away. Maybe as a second step rename the old one to something a bit more ponderous?

Maybe compose-use-slots or something like that? I'd prefer to do that as a quick PR to not complicate this one but it is great feedback.

rurikoaraki: Sure, that seems fine to me. Out of curiosity are there any issues with renaming existing packages if someone external was using them?

packages/framework/use-slot/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "@fluentui-react-native/merge-props": ">=0.3.2 <1.0.0"
+  },
+  "devDependencies": {
+    "@uifabricshared/build-native": "^0.1.1",
+    "react": "16.13.1",
+    "react-native": "^0.63.4"
+  },
+  "peerDependencies": {
+    "react": ">=16.13.1",
+    "react-native": ">0.60.0"
```
rurikoaraki, line 46: Curious why this is much lower than the version used in devDependencies?

JasonVMo: This was copied from use-slots so I didn't really reassess this. What this is saying is effectively 'when you bundle me, you need to bundle a version of react above X and a version of react-native above Y.' This particular package would work with 0.60 and beyond, but we should figure out what the min-supported version for the library really is and make sure all the declared peer dependencies are consistent. Likely dep-check can help with that but it's a good follow-up.

Saadnajmi: I think we implicitly decided RN 0.63 was the minimum supported version when we updated to RN 0.63 and changed all of our dependencies to reflect that.

JasonVMo: Turns out I copied use-slots to start this package which was one of only three packages that didn't have the correct peer dependencies set. I updated this package and the other ones in the project that were incorrect as well.

packages/experimental/composition/package.json
JasonVMo: If we have users who are using use-slots directly then they would need to rename their references to pick up new versions. In most cases the internals of use-slots have been exposed via re-exporting via the framework package which insulates us from a lot of this.

Saadnajmi:
A request: Some of our interns set up our [wiki](https://github.com/microsoft/fluentui-react-native/wiki) to display all of our markdown files. Maybe as a followup if needed, could we update that to reflect this new package when it goes in? Specifically this page: https://github.com/nabil-k/fluentui-react-native/blob/master/packages/experimental/README.md

Saadnajmi:
From the example you listed in your PR, would it be possible to have a real world code example? Like replacing the image slot of a button with an Avatar or ActivityIndicator or the like? I'm not sure what two slots will ever have the same exact props. I suppose if it was just a View, you could replace it with anything.
```  
type OverrideableSlot = MyComponentProps & { override?: React.FunctionComponent<MyComponentProps> };

const SimpleControl = (props: OverrideableSlot) => {
  const { children, override, ...rest } = props;
  const Container = useSlot(override || View, rest);
  return <Container>{children}</Children>
}
```

Saadnajmi (approved):

rurikoaraki (approved):

PR #819, opened 2021-07-26
Add compressible utility to the framework package

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This adds a new utility `compressible` which can create components that can be tree compressed (using the `stagedRender` pattern) which also have `customize` functionality on them.

I also added a test which can serve as a bit of an example of how to author a component of this type. It has two controls defined and shows layer compression, customizing, and slot replacement.

This change is purely additive, it should not change existing implementations in any way.

### Verification

Automated tests and builds

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


JasonVMo:
I put automerge on this since I will be leaving a bit early today for a hiking trip and given that this is an additive change it should be safe to merge whenever you all feel comfortable with it. I'm open to discussion/feedback/changes as well (obviously) but will be OOF with limited internet connectivity until Wednesday around lunchtime.

packages/framework/framework/src/compressible.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+const useLabelTokens = buildUseTokens<LabelTokens>({
+  headerVariant: 'header',
+  captionVariant: 'caption',
+});
+
+const Label = compressible<LabelProps, LabelTokens>((props: LabelProps, useTokens: UseTokens<LabelTokens>) => {
+  const theme: Theme = {} as Theme;
+  const [tokens] = useTokens(theme);
+  const { headerText, captionText, headerSlot, captionSlot } = props;
+  const Header = useSlot<VariantTextProps>(headerSlot || VariantText, { variant: tokens.headerVariant });
```
JasonVMo, line 86: Note that while this is allowing plugging the slot via props it could easily be done via tokens as well. It can also be completely internal. It's up to the author to determine this.

packages/framework/framework/src/compressible.ts
```diff
@@ -0,0 +1,21 @@
+import { Theme } from '@fluentui-react-native/theme-types';
+import { stagedComponent, StagedRender } from '@fluentui-react-native/use-slot';
+import { CustomizableComponent } from '@fluentui-react-native/use-tokens';
+import { UseTokens } from './useTokens';
+import { TokenSettings } from './useStyling';
+
+export function compressible<TProps, TTokens>(
```
FalseLobster, line 7: @JasonVMo This needs to be exported from index.ts

FalseLobster (approved):

PR #821, opened 2021-07-27
Fix `lint` step failing due to picking up build artifacts

tido64:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

`lint` step fails the second time around due to trying to lint build artifacts.

- Cleans up duplicate ESLint installations
- Cleans up duplicate ESLint presets
- Bumps `eslint` to 7.3.1
- Cleans up all linter violations where possible; added suppressions elsewhere

### Verification

Not much to verify.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/Button.android.tsx
```diff
@@ -66,6 +66,7 @@ export const Button = compose<IButtonType>({
     root: View,
     ripple: Pressable,
     stack: { slotType: View, filter: filterViewProps },
+    // eslint-disable-next-line @typescript-eslint/ban-types
     icon: { slotType: Icon as React.ComponentType<object> },
```
Saadnajmi, line 70: Are these comments things that are probably worth a followup from our side?

packages/deprecated/foundation-composable/src/Composable.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
 import * as React from 'react';
```
Saadnajmi, line 3: Why did we have to ban these?

packages/framework/immutable-merge/src/Merge.test.ts
```diff
@@ -133,7 +133,7 @@ const singleWithChanges = {
 
 const _colorKey = 'color';
 
-const changeMeHandler = (...objs: (object | undefined)[]) => {
+const changeMeHandler = (...objs: (Record<string, unknown> | undefined)[]) => {
   // written always assuming only one entry
```
Saadnajmi, line 137: This seems like an API change?

packages/utils/interactive-hooks/src/useAsPressable.ts
```diff
@@ -12,6 +12,7 @@ import { usePressability } from './usePressability';
 /**
  * hover specific state and callback helper
  */
+// eslint-disable-next-line @typescript-eslint/ban-types
```
Saadnajmi, line 15: My impression is this line undoes the ban that lets us cast stuff as `object`. I wonder if we should just remove it from our config given we use it so much.

scripts/just.config.js
```diff
@@ -28,7 +28,7 @@ module.exports = function preset() {
   );
 
   task('depcheck', checkPublishing);
-  task('lint', eslintTask({ files: ['src/.'] }));
+  task('lint', eslintTask({ files: ['src/'] }));
```
Saadnajmi, line 31: Is this the one line change that stops us from listing artifacts?

packages/deprecated/foundation-composable/src/Composable.ts
Saadnajmi: Or is it just because all of these are in the `deprecated` path so we don't quite care to lint them?

packages/utils/interactive-hooks/src/useAsPressable.ts
NickGerleman: `object` is almost never what you want. `{}` generally isn't either, but ends up being safe when used as a prop type in react components (https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492)

NickGerleman (approved):
😍

NickGerleman:
Anywhere in PR pipeline that enforces lint step succeeds?

packages/components/Button/src/Button.android.tsx
tido64: So this is the error:
```
ERR! /~/packages/utils/test-tools/src/enzymeTests.ts
ERR!   12:10  error    Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
ERR! Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys  @typescript-eslint/ban-types
```
It's probably worth taking a look. I know too little of the code base to have an opinion.

packages/framework/immutable-merge/src/Merge.test.ts
tido64: This is a test, so I'm not too worried. If it compiles and the test passes, it should be fine.

Saadnajmi:
> Anywhere in PR pipeline that enforces lint step succeeds?

The "PR (Build, test, and bundle JavaScript)" should include linting.

packages/utils/interactive-hooks/src/useAsPressable.ts
tido64: +1 to what @NickGerleman said.

I posted this above, but here it is again:
```
ERR! /~/packages/utils/test-tools/src/enzymeTests.ts
ERR!   12:10  error    Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
ERR! Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys  @typescript-eslint/ban-types
```
It's probably worth taking a look. I know too little of the code base to have an opinion.

Saadnajmi: Should I take it then that for most things we cast to object, there's probably a more appropriate type that it could be cast to instead?

scripts/just.config.js
tido64: Honestly, there were multiple issues and I ended up having to fix them all simultaneously. I think the `.` is what caused the issue, but I can't say 100%.

packages/components/Button/src/Button.android.tsx
Saadnajmi: Oh, interesting. That sounds very typescript specific, and is good to know. Thanks!

Saadnajmi (approved):

tido64:
> > Anywhere in PR pipeline that enforces lint step succeeds?
> 
> The "PR (Build, test, and bundle JavaScript)" should include linting.

Yup, it works so well that the second time you run `yarn build` from repo root, it will start linting everything inside `lib`, `lib-commonjs`, etc. 😛

packages/deprecated/foundation-composable/src/Composable.ts
tido64: These use `object` extensively. Instead of adding `// eslint-disable-next-line @typescript-eslint/ban-types` everywhere, I opted to disable entirely since they are deprecated.

packages/utils/interactive-hooks/src/useAsPressable.ts
tido64: Almost always. The proposed `Record<string, unknown>` is a good alternative if you don't know what the value type is. If you do know the value type, you should use that instead. You can try converting one at a time and see what happens. I wouldn't recommend bulk-converting all at the same time. That's a recipe for disaster 😛

tido64:
@NickGerleman, @Saadnajmi: Thanks for reviewing. I'm going to land this now to avoid merge conflicts since I'm touching so many files.

PR #827, opened 2021-07-27
Menu Button on macOS (Initial check in)

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change introduces macOS support for the MenuButton component. As it is, it only supports a limited subset of the component's features, and lacks customization. I would like to add this initial version so that we can integrate it downstream and get feedback from clients. I tried to make no changes to the win32 implementation so as to not break compatibility.

I elected to implement this as a native module rather than composed of a `Button` and `ContextualMenu` component because I had a lot of trouble standing up the `ContextualMenu` component on macOS. `ContextualMenu` depends on `Callout`, which is only implemented for win32 and has a win32 specific API. On macOS, contextualMenus are generally not hosted in the equivalent of a `Callout` anyway; they are their own objects altogether. 

Future followups can/wil include:
- Extending the functionality of this component to support more of the props available on win32
- Standing up `ContextualMenu` for macOS (new API and all), and then rewriting this component to actually be composed of `Button` and `ContextualMenu`. 
- Porting this component to the new compose framework, along with ContextualMenu.

**Supported props:**
- menuItems,
  -  title, 
  - tooltip, 
  - submenus, 
  - disabled,
  - onItemClick (only 1 subMenu level deep)
- content, 
- Icon (png only)
- disabled, 
- onItemClick

**Unsupported but next planned to be added**
- Layout (Right now the widht/height are hard coded. I looked into it and I'll probably need to implement a shadow view to get the proper layout).

### Verification

https://user-images.githubusercontent.com/6722175/127942675-2b912247-4044-4219-a84d-28cdb3be5daa.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/testPages.macos.ts
```diff
@@ -15,6 +15,7 @@ import { HOMEPAGE_SEPARATOR_BUTTON, SeparatorTest } from './TestComponents/Separ
 import { HOMEPAGE_SVG_BUTTON, SvgTest } from './TestComponents/Svg';
 import { HOMEPAGE_TEXT_BUTTON, TextTest } from './TestComponents/Text';
 import { HOMEPAGE_THEME_BUTTON, ThemeTest } from './TestComponents/Theme';
+import { MenuButtonTest, HOMEPAGE_MENU_BUTTON } from './TestComponents/MenuButton';
```
HeyImChris, line 18: Nit- reorder the imports so they match everywhere else?

packages/components/MenuButton/FluentUIReactNativeMenuButton.podspec
```diff
@@ -0,0 +1,21 @@
...
+  s.name         = 'FluentUI-React-Native-MenuButton'
+  s.version      = package['version']
+  s.summary      = package['description']
+  s.license      = package['license']
+
+  s.authors      = package['author']
+  s.homepage     = package['homepage']
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
```
HeyImChris, line 15: Maybe just be explicit that we want "5.0" here so minor version updates are more obvious/purposeful?

packages/components/MenuButton/README.md
```diff
@@ -0,0 +1,5 @@
+# MenuButton
+
+MenuButton is a wrapper control that combines and simplifies the API of Button and ContextualMenu.
+
+On macOS, it is implemented as a native module, and thus lacks the same level of customization as it's win32 counterpart.
```
HeyImChris, line 5: The way this is phrased makes it sound sort of negative, but I don't think it's all bad. We should generally provide the level of customization that the OS allows and if that's different from Windows, it doesn't mean the control is lacking support, it's just that it has the expected functionality for the platform

packages/components/MenuButton/macos/MSFMenuButtonManager.m
```diff
@@ -0,0 +1,72 @@
+#import <React/RCTComponent.h>
```
HeyImChris, line 1: What's the `MSF` file prefix for?

packages/components/MenuButton/macos/MSFMenuButtonManager.m
```diff
@@ -0,0 +1,72 @@
...
+  NSMenu *menu = [[NSMenu alloc] init];
+  [menu setAutoenablesItems:NO];
+
+  NSArray *menuItems = [RCTConvert NSArray:json];
+  for (NSDictionary *menuItemJson in menuItems) {
+    NSMenuItem *menuItem = [RCTConvert menuItem:menuItemJson];
+
+    if ([menuItemJson containsKey:@"hasSubmenu"])
+    {
+      bool hasSubmenu = [RCTConvert BOOL:menuItemJson[@"hasSubmenu"]];
```
HeyImChris, line 38: Obj-c uses `BOOL` not `bool` so it's a good idea to match that

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,111 @@
+//
```
HeyImChris, line 1: I don't think this matters in terms of functionality but for my own curiosity, what's the reason to write the manager in obj-c and the menu button in swift?

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // MARK: - Private Methods
+
+  private func updateMenu() {
+    guard let menu = menu else {
+      return
+    }
+
+    // Insert an initial empty item into index 0, since index 0 is never displayed
```
HeyImChris, line 71: Is this documented somewhere? I'm probably missing context but it sounds fishy that an item in the menu wouldn't ever be displayed...

HeyImChris (changes requested):

packages/components/MenuButton/macos/MenuButton.swift
Saadnajmi: Not really. I believe it is because we are using a nonstandard (but accepted) to set the title/image of this MenuButton.

By default, NSPopupButton will either display the first menu item or the currently selected menu item (depending on if the boolean `pullsDown` is set) to display in the "button" part. We don't want to use a menu item from the menu, we want to have a custom title and image that is NOT in the menu. To achieve this, apple recommends setting the NSPopupButton's cell's `usesItemFromMenu` property to false, and then providing that cell a custom menu item to render instead. This is what I am doing from lines 87 to 93. My guess is a side effect is that first menu item that was supposed to be used to render the title/image of the button is now never shown. 

This info comes from this sample code, which is pretty old (pre ARC) but is the only apple official sample code I found https://developer.apple.com/library/archive/samplecode/MenuMadness/Introduction/Intro.html#//apple_ref/doc/uid/DTS40008870

packages/components/MenuButton/macos/MenuButton.swift
Saadnajmi: The Manager must be written in obj-c because it uses Macros to talk to React Native. The View itself however can be written in Swift with @objc() wrappers around appropriate properties and methods. React Native docs describe how to this here: https://reactnative.dev/docs/native-modules-ios#exporting-swift

I also personally try to use as much Swift as possible, since that's the newer / nicer / safer language to use.

packages/components/MenuButton/README.md
Saadnajmi: We're limiting OS level customization too though. For example, I could have allowed you to set whatever `bezel` you want on the NSPopupButton to get different styles of menu buttons. I also don't allow you to use custom view as NSMenuItems (which is supported by Apple) with this approach. Ultimately, I chose to go with limited customization and a native module because the other approach (standing up FURN `ContextualMenu` on macOS) was experiencing difficulties, and we had already made this "helper" component for our client team that I could take advantage of and provide a native experience. 
Once this specific feature request is done, I still need to go back and stand up either `Callout` or `ContextualMenu` for future work, which means this implementation might be revisited.

packages/components/MenuButton/macos/MSFMenuButtonManager.m
Saadnajmi: It's supposed to be `Microsoft Fluent` I think. It's what we use in FluentUI Apple. This is not a FluentUI Apple control but I still wanted a 3 letter prefix to follow our naming convention (https://microsoft.github.io/objc-guide/Classes/Naming.html)

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: What do you mean?

harrieshin: I think he is proposing HOmePage_menu_button, menubuttontest instead

packages/components/MenuButton/macos/MSFMenuButtonManager.m
harrieshin: that's interesting discussion. are we going to have MSF in all of our native implementation under furn?

Saadnajmi: There needs to be something. I didn't really want to introduce a new 3 letters. Some of the other JS only components use the prefix "RNF" in some parts as well. The RN community sometimes uses "RCT" (short for react I think), but that didn't feel right.

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: Ah. Good catch.

packages/components/MenuButton/macos/MenuButton.swift
harrieshin: for my own knowledge, does VO properly ignores the empty one ?

Saadnajmi: https://user-images.githubusercontent.com/6722175/128072198-0ec89a1e-7d89-4b4c-a790-d695711f2f61.mov

Looks like it.

packages/components/MenuButton/README.md
chripear: nit: its (no apostrophe)

packages/components/MenuButton/macos/MSFMenuButtonManager.m
```diff
@@ -0,0 +1,74 @@
...
+  [menu setAutoenablesItems:NO];
+
+  NSArray *menuItems = [RCTConvert NSArray:json];
+  for (NSDictionary *menuItemJson in menuItems) {
+    NSMenuItem *menuItem = [RCTConvert menuItem:menuItemJson];
+
+    if ([menuItemJson containsKey:@"hasSubmenu"])
+    {
+      BOOL hasSubmenu = [RCTConvert BOOL:menuItemJson[@"hasSubmenu"]];
+      if (hasSubmenu) {
```
chripear, line 39: How about using `objectForKey` instead of implementing an NSDictionary extension?

If `[RCTConvert BOOL:nil]` is OK and returns either `NO` or `nil`:
`if ([RCTConvert BOOL:[menuItemJson objectForKey:@"hasSubmenu"]]) {`

Otherwise add Elvis:
`if ([RCTConvert BOOL:[menuItemJson objectForKey:@"hasSubmenu"] ?: whateverThingRCTConvertIsExpectingThatReturnsNO]) {`

chripear: (Blame my finely honed double-lookup senses from Python)

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,131 @@
...
+
+  private func updateMenu() {
+    guard let menu = menu else {
+      return
+    }
+
+    for (index, menuItem) in menu.items.enumerated() {
+      menuItem.target = self
+      menuItem.action = #selector(sendOnItemClickEvent)
+      if menuItem.hasSubmenu, let submenu = menuItem.submenu {
```
chripear, line 76: The normal swift idiom would be

`if let submenu = menuItem.submenu {`

where `menuItem.submenu` would be nil if `menuItem` doesn't have a submenu.  Then we needn't bother with a `hasSubmenu` boolean.  Can we do that?

packages/components/MenuButton/README.md
Saadnajmi: fixed

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,131 @@
...
+    let initialEmptyItem = NSMenuItem()
+    menu.insertItem(initialEmptyItem, at: 0)
+  }
+
+  private func updateDropDownCell() {
+    guard let dropDownCell = cell as? NSPopUpButtonCell else {
+      preconditionFailure()
+    }
+
+    // MenuButton needs a MenuItem set on it's cell to display the title and image properly
```
chripear, line 97: nit: its

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,131 @@
...
+      updateMenu()
+    }
+  }
+
+  open override var image: NSImage? {
+    get {
+      return menuButtonImage
+    }
+    set {
+      // We must set the image on the dropdown cell rather than the button,
```
chripear, line 53: nit: period instead of comma (Word would call this a comma splice)

packages/components/MenuButton/macos/MenuButton.swift
```diff
@@ -0,0 +1,131 @@
...
+        for subMenuItem in submenu.items {
+          subMenuItem.tag = index //store the index of the "super" menuItem for lookup in the action
+          subMenuItem.target = self
+          subMenuItem.action = #selector(sendOnSubItemClickEvent)
+        }
+      }
+    }
+
+    // Insert an initial empty item into index 0, since index 0 is never displayed
+    // We must do this after we assign the tags to the submenuItems to preserve index order
```
chripear, line 87: nit: Add a period at the end of both lines.  Comments can either be a phrase, or complete sentence(s).  Multiline comments containing multiple sentences should be proper paragraphs.

packages/components/MenuButton/macos/MSFMenuButtonManager.m
Saadnajmi: `if ([menuItemJson objectForKey:@"hasSubmenu"])` works as expected. Thanks!

packages/components/MenuButton/macos/MenuButton.swift
Saadnajmi: Yeah I guess I didn't need to check `hasSubmenu`. Thanks!

chripear (approved):

packages/components/MenuButton/macos/MSFMenuButtonManager.m
HeyImChris: It shouldn't be RCT but feels equally off to make it the same as the Fluent repo. Ideally it should be something unique for this repo. Maybe RNF like you suggested?

HeyImChris (approved):

packages/components/MenuButton/macos/MSFMenuButtonManager.m
Saadnajmi: I'm gonna go with FRN because the usages of "RNF" seem like a legacy thing and newer components have dropped the prefix altogether

Saadnajmi: Actually, never mind. I talked to Amanda and we want to think about it more. I made an ADO bug as a followup assigned to me.

PR #830, opened 2021-07-29
🚧 [DRAFT] Add border & shadow effects to default theme, shadow component to apply them easily

FalseLobster:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR...
- Hardcodes borderRadius global tokens into the default theme for all platforms
- Hardcodes shadow alias tokens into the default theme for all platforms
- Creates a Shadow component using the new compressible API
- Serves as a discussion/rough proposal about how we might want to introduce these concepts into the Theme object and if a shared component might be valuable

#### Problem

The Fluent Design System uses two shadows for each shadow in its "Shadow Ramp", an ambient shadow and a key shadow.  In the web, css's box-shadow allows you to apply multiple shadow effects to a single HTML tag, but in react-native iOS, win32, and (presumably) macOS + windows only permit 1 shadow per View.  

Win32, in particular, has some idiosyncrasies you need to be aware of to make them work and look good-- backgroundColors are required on both Views if we want the shadow to match the rounded borders prevalent throughout the design.  This may be true for other platforms as well, but the alpha must come through the ViewStyle shadowOpacity prop and not be part of the rgba on shadowColor.

Android is an odd-one out here, in that react-native for android currently only supports a single 'elevation' prop instead of the usual x/y offset and blurRadius and opacity props.  It's unclear what Android Shadows should look like in our current Design specs, but it might also be useful to have a component that does not render the unnecessary view when the platform doesn't call for it (assuming design will just pick single elevation values, only one view will be required.

Ideally each control would apply the appropriate shadow in its own styling, but I thought it might be useful for us  and our partners to have something that makes this process a little easier.

#### To Do
- [ ] Reorganize the mess I made to the various platform themes
- [ ] Evaluate if I can pull these values from Ruriko's global/alias token work
- [ ] Drive consensus on the placement/naming conventions in the Theme object
- [ ] Confirm they work on all platforms (see Platforms Impacted)
- [ ] Drive consensus around the utility and desired API shape of the Shadow component
- [ ] Consider what sort of tests are appropriate
- [ ] Consider Theme "modes" (opacities are different in Dark mode)

#### Win32

image

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/utils/test-tools/src/mockTheme.ts
```diff
@@ -175,6 +175,18 @@ export const mockTheme: Theme = {
...
+    borderRadiusSmall: 2,
+    borderRadiusMedium: 4,
+    borderRadiusLarge: 6,
+    borderRadiusXLarge: 8,
+    shadow2: { ambient: undefined, key: undefined },
+    shadow4: { ambient: undefined, key: undefined },
+    shadow8: { ambient: undefined, key: undefined },
+    shadow16: { ambient: undefined, key: undefined },
+    shadow28: { ambient: undefined, key: undefined },
+    shadow64: { ambient: undefined, key: undefined },
```
Saadnajmi, line 188: Why are all these undefined?

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -88,6 +88,65 @@ function _defaultTypography(): Typography {
...
+          borderRadiusMedium: 4,
+          borderRadiusLarge: 8,
+          borderRadiusXLarge: 12,
+        }
+      : {
+          borderRadiusSmall: 2,
+          borderRadiusMedium: 4,
+          borderRadiusLarge: 6,
+          borderRadiusXLarge: 8,
+        };
```
Saadnajmi, line 105: Are these coming from Figma

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -88,6 +88,65 @@ function _defaultTypography(): Typography {
...
+      ((Platform.OS === 'android' || Platform.OS === 'macos') && {
+        shadow2: undefined,
+        shadow4: undefined,
+        shadow8: undefined,
+        shadow16: undefined,
+        shadow28: undefined,
+        shadow64: undefined,
+      }),
+  );
+  return effects as Effects;
```
Saadnajmi, line 147: A lot of these checks might be easier if we assume that `default` means web/win32, and have the default theme (if no theme is specified) be the platform themes (android/apple theme). That way, we could move the platform specifics to the platform themes and feel less bad lol. How would you feel about that?

packages/theming/apple-theme/src/appleTheme.ios.ts
```diff
@@ -126,6 +127,7 @@ export const BaseAppleLightThemeIOS: Theme = {
   typography: appleTypography(),
   spacing: appleSpacing(),
   components: appleComponents,
+  effects: defaultEffects(), // TODO: Saad, if this makes it into CR I'm counting on you to yell at me :D
```
Saadnajmi, line 130: But what if I want it on the apple theme? :'(

packages/experimental/Shadow/package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/microsoft/fluentui-react-native.git"
+  },
+  "keywords": [
+    "react-native",
+    "shadow",
+    "fluentui"
+  ],
```
Saadnajmi, line 50: What does the `keywords` field do?

packages/theming/apple-theme/src/appleTheme.ios.ts
JasonVMo: I haven't been paying attention to the theming stuff quite as much but why don't the apple and android themes base themselves off the default theme and then just override the apple/android portions as needed? You can still do applyTypography or appleSpacing calls to override it but it provides a fallback so that when we need a new section it doesn't require changes to all the platform specific themes.

This isn't a blocking comment for this PR, just curious why this approach was taken.

packages/experimental/Shadow/src/Shadow.types.ts
```diff
@@ -0,0 +1,22 @@
...
+  innerShadowSlot?: React.ComponentType<ViewProps>;
+  depth2?: ShadowTokens;
+  depth4?: ShadowTokens;
+  depth8?: ShadowTokens;
+  depth16?: ShadowTokens;
+  depth28?: ShadowTokens;
+  depth64?: ShadowTokens;
+}
+
+export interface ShadowProps extends ViewProps, ViewStyle, IBorderTokens {
```
JasonVMo, line 20: I'm curious why this needs to extend ViewStyle and BorderTokens? For changing tokens I would suggest encouraging people to use .customize(). For style changes (e.g. ViewStyle) the internal memoization in mergeStyles should handle the overriding by setting the style prop. In other words it feels like this component can start as a View that has an additional depth prop. It's always easy to add more later.

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const theme = useTheme();
+  let [tokens, cache] = useTokens(theme);
+  const { depth, padding, paddingHorizontal, style, ...rest } = props;
+  [tokens, cache] = applyTokenLayers(
+    tokens,
+    ['depth2', 'depth4', 'depth8', 'depth16', 'depth28', 'depth64'],
+    cache,
+    (layer) => layer === 'depth' + depth,
+  );
+  [tokens, cache] = applyPropsToTokens(props, tokens, cache, ['borderColor', 'borderRadius', 'borderStyle', 'borderWidth']);
```
JasonVMo, line 62: If you take away the extra props added to ShadowProps and just have depth you can take out this line as the only prop that would affect your tokens would be `depth` which you already handle with the apply token layers

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -0,0 +1,93 @@
...
+    ...rest,
+    style: mergeStyles(
+      {
+        backgroundColor: tokens.backgroundColor,
+        shadowColor: tokens.color,
+        ...borderStyles.from(tokens, theme),
+        ...tokens.key,
+      },
+      style,
+    ),
```
JasonVMo, line 73: The object identity of the style used in merge styles needs to be consistent, not just the values within. This can be done via either `useMemo` or the `cache` you have at that point.

`useMemo` is a per-component run to run cache. Meaning if component instance A renders a second time it will re-use the object if the passed in key values are the same. Usage would look like this:

```ts
const rootStyle = React.useMemo(() => ({
  backgroundColor: tokens.backgroundColor,
  shadowColor: tokens.color,
  ...tokens.key
}), [tokens, theme]);
```

The `cache` splits itself based on layer permutations and theme automatically so it is ready to use at this point. It stores styles across instances so multiple instances of `Shadow` with the same depth will end up using the same styles as long as the input parameters are the same. Usage of cache would look something like:

```ts
const [tokenStyles] = cache(() => ({
  rootStyle: {
    // same as above
  },
  innerStyle: {
    // second section below
  }),
  []); // no keys required since tokens + theme are already considered
```

JasonVMo (changes requested):
Very exciting to see this!

packages/theming/apple-theme/src/appleTheme.ios.ts
Saadnajmi: Two reasons:

Short:
1) I couldn't wrap a ThemeReference in another ThemeReference. I could only wrap a theme or ThemeRecipe (I think that was the name). 
1) I didn't want edge cases where the apple theme didn't override some visuals from the default theme and you had some mixed tokens that came together unevenly. I'm not sure that's an actual problem but it was a thought. 

Long answer:

The Apple theme was initially implemented with a native module, so I needed a full theme reference with extra listeners and such. Rhe issue was I found that I couldn't set the base theme of a ThemeReference to another themeReference, only a theme. Not all the listeners transferred over, and I was getting bugs where my new theme would stay in light mode when it should be dark, etc. The Android Theme used the apple theme as a guide, so it was it's own ThemeReference too. I think the fact that a ThemeReference cannot wrap another ThemeReference was probably a bug, but I couldn't figure it out at the time and making a new ThemeReference seemed like the way to go. In the time interim, I managed to update `react-native-macos` with a new API to remove the need for a native module, so both the apple and android theme are defined entirely in JS, and could maybe be refactored as layers on the default theme.

rurikoaraki:
FWIW Travis told me that he'd work on adding the ability to have shadow alias tokens in the pipeline next week.

packages/experimental/Shadow/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@fluentui-react-native/experimental-shadow",
+  "version": "0.0.1",
```
rurikoaraki, line 3: Oh are we supposed to start packages at 0.0.1? Whoops.

packages/experimental/Shadow/package.json
rurikoaraki: I guess it adds some key terms for searching on npmjs?

https://docs.npmjs.com/cli/v6/configuring-npm/package-json#keywords

packages/experimental/Shadow/package.json
Saadnajmi: We've historically done 0.1.0. Is there a package we have that is 0.0.1?

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -88,6 +88,65 @@ function _defaultTypography(): Typography {
   return defaultsDict;
 }
 
+export function defaultEffects(): Effects {
+  const effects =
+    Platform.OS === 'android' || Platform.OS === 'ios'
+      ? {
+          borderRadiusSmall: 2,
```
rurikoaraki, line 95: These should probably be pulled from the global tokens in the future. Right now they don't account for mobile platforms so we need to add those.

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -15,15 +15,37 @@ export interface Spacing {
...
 
+export type ShadowStyle = ShadowStyleIOS & Pick<ViewStyle, 'elevation'>;
+
+export interface ShadowValue {
+  ambient?: ShadowStyle;
+  key: ShadowStyle;
+}
+
+export interface Effects {
+  borderRadiusSmall: number;
```
rurikoaraki, line 26: Maybe the border Radius can be its own interface/grouping? I don't know if I would group it under "Effects."

Design has border radius as its own group of global tokens separate from shadows/elevation, so I would probably align with that.

packages/experimental/Shadow/src/Shadow.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type ShadowDepth = '2' | '4' | '8' | '16' | '28' | '64';
+
+export interface ShadowTokens extends IBorderTokens {
+  ambient?: ViewStyle;
+  key?: ViewStyle;
+  color?: ColorValue;
+  backgroundColor?: ColorValue;
+  innerShadowSlot?: React.ComponentType<ViewProps>;
+  depth2?: ShadowTokens;
```
rurikoaraki, line 12: I'm curious why the depths and the other tokens are part of the same interface. The current structure allows for recursion/nesting so you can do something like depth2.depth4.backgroundColor = whatever, which seems odd to me.

packages/utils/test-tools/src/mockTheme.ts
FalseLobster: I was being lazy for the hackathon and I'm not sure it matters terribly much for the test this mock is used in

packages/theming/default-theme/src/defaultTheme.ts
FalseLobster: yea I hand copied them.  Mobile devices get one set, Large screen gets another.  I did skirt over the fact that web has separate values for mobile than the native mobile values, but react-native-web isn't really a first class platform for us right now

packages/experimental/Shadow/package.json
FalseLobster: I don't think its a big deal either way, but beach ball with bump it to 0.1.0 after this gets published anyway since I described this a minor change not a patch.

FalseLobster: I think I did 0.0.1 simply because that was the default for npm init

packages/theming/theme-types/src/Theme.types.ts
FalseLobster: Yea, in the Fabric 7 theme object had both of these under theme.effects and that's what I was initially copying but it looks like the new Fluent theme object organizes things like theme.alias.color.neutral, theme.alias.shadow

We don't necessarily have to copy them, but it's worth considering prior art and I'm curious to hear what other folks think

packages/theming/default-theme/src/defaultTheme.ts
FalseLobster: totally agreed.  Any opinions on how/if I should handle this in the interim?

rurikoaraki: I think as it is is fine. The global tokens we have in FURN have the win32 tokens available if you'd like to use them, but otherwise I'll come back and fix it later.

packages/experimental/Shadow/src/Shadow.types.ts
FalseLobster: I didn't love this and its the part of the reason I wanted to start soliciting feedback because it worked for my hackathon needs but didn't feel right.

I guess my thought was if we don't expose ViewStyle props as props on the Shadow we get in a weird place for a lot of users.  A standalone Shadow component is a fairly basic concept kind of like a View++, and it's hard for me to imagine people not tweaking things like borderRadius and padding on it regularly.  From an outsider's perspective, I want to treat it like it's a single View.  Internally, I'm merging the user provided style prop onto the root slot, but borderRadius is needed on both slots, padding needs to go on inner slot, as well as I presume many of the flexbox style props.

Leaving ViewStyle out of the props interface means that anyone that needs something to go on the inner slot customize it.  Maybe that's ok?

packages/theming/apple-theme/src/appleTheme.ios.ts
FalseLobster: I want it in the apple theme, too Saad... :)  I just realized it would've probably been better organization to put the platform specific stuff in the platform specific places after I already shoved it all in the default theme.

I actually did in the default theme initially to be hackathon-quick because my mental model was similar to Jason's until I realized the platform themes don't fall back like thought, so I had to add it to the platform themes in a weird way.

JasonVMo: I'm pretty sure ThemeReferences can be layered, it's probably a lack of documentation on my part that made it seem they couldn't. For reference a ThemeReference has the following constructor:

```ts
constructor(base: TTheme | ThemeReference<TTheme, TPartial>, ...recipes: ThemeRecipe<TTheme, TPartial>[]) {
```

So if you want to do this in the future it looks like `const myTheme = new ThemeReference(baseTheme, ...myRecipes);`

They were designed to be layered as much as required. You could even go fluent -> apple -> ios / macos if you want. The second reason makes sense, though I wonder if some use of required types might help here. Ideally we want the compiler to catch those errors for us.

packages/experimental/Shadow/src/Shadow.types.ts
JasonVMo: That makes more sense now. I do think the default should be to start with .customize only and then add things to props as needed as it is much harder to go the other way. That being said if we think there are a set of commonly modified values then it might be worth exposing them as props.

Maybe a way to reason about it is if you are rendering a page and you want to change how all of your shadows look customize works great. If you want them to be much more dynamic and may have lots of permutations then props would work.

Right now we expect people to have multiple depth values on a page which is why its a prop. (I'm just trying to work through how to make these decisions)

packages/experimental/Shadow/src/Shadow.tsx
JasonVMo: Also if you do need props in here definitely look at the experimental-text PR, I think the patchTokens stuff is a bit more elegant than applyPropsToTokens.

packages/theming/apple-theme/src/appleTheme.ios.ts
Saadnajmi: I'll try having the apple theme be a layer/ThemeRecipe on the default theme again. I don't remember the exact issue I ran into, it might have also just been related to how we constructed the `testerTheme` in FluentTester

packages/experimental/Shadow/src/Shadow.types.ts
FalseLobster: yea we apply Tokens in layers which is kind of convenient but in this case it's also a little strange.  We'll flatten the nested Token objects based on the function passed to applyTokenLayers

`(layer) => layer === 'depth' + depth`

It that returns true, then the nested tokens will be merged on top of the underlying layer.  It's basically how the depth prop functions right now... but maybe that isn't the right approach for this component.  Let me think about it a little more.

packages/experimental/Shadow/package.json
FalseLobster: yea npm init asked me for some keywords so I just picked a few

FalseLobster: See also the bottom of this page: https://www.npmjs.com/package/@fluentui/react

packages/theming/apple-theme/src/appleTheme.ios.ts
FalseLobster: Yea, the layering does work as far as i know--- see the examples I crafted in my AccChecker branch:  

https://github.com/FalseLobster/fluentui-react-native/commit/16913d5d97548adad399e1886fc82fb35d603f33

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -0,0 +1,93 @@
...
+      key: theme.effects.shadow28.key,
+    },
+    depth64: {
+      ambient: theme.effects.shadow64.ambient,
+      key: theme.effects.shadow64.key,
+    },
+  };
+}, 'Shadow');
+
+export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps, useTokens: UseTokens<ShadowTokens>) => {
```
Saadnajmi, line 52: My "I'm dumb, plz explain", overarching question:

Why does this need to be a separate component instead of just an extra set of styles that are applied to existing components? In the end the Shadow tokens are just being converted into styles that are applied to a view, why can't we just apply it to the root view of existing components?

FalseLobster:
Superseded by #1779

PR #833, opened 2021-07-29
Convert experimental text to new approach

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This rewrites the @fluentui-react-native/experimental-text control to use the new `compressible` approach instead of compose. This also strives to provide a canonical example of how to write simple components in this manner though text is simple enough that I didn't use a slot for the inner text as that is a fundamental react-native type and that is no longer pluggable. We can always add it later if desired.

In rewriting it I added two additional routines to make doing these things easier.

**`useFluentTheme()` helper**

From the side discussion in teams, asking people to remember to write `useTheme() || defaultFluentTheme` in their code is just asking for errors. Beyond that it hard codes the fallback value in a million places. Instead I added a new helper in framework called `useFluentTheme` which falls back automatically, then used it in experimental text.

**`patchTokens`**

The apply props to tokens pattern felt unwieldy for a couple of reasons but the main one is that the typical react component building often starts with something like:

```ts
const { prop1, prop2, prop3, ...rest } = props;
```

This has two uses. First to extract values, second to remove those values from props without the need to filter them later. To make this easy I created the `patchTokens` api which just updates the tokens with override values from props and does the appropriate caching automatically. We may want to remove the old API at some point but until more things have been written I'm not 100% certain it is a good idea.

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/framework/framework/src/useFluentTheme.ts
```diff
@@ -0,0 +1,6 @@
+import { defaultFluentTheme } from '@fluentui-react-native/default-theme';
+import { Theme, useTheme } from '@fluentui-react-native/theme-types';
+
+export function useFluentTheme(): Theme {
+  return useTheme() || defaultFluentTheme;
+}
```
Saadnajmi, line 6: If we're introducing this, could we have this return the apple or android theme on each appropriate platform?

packages/experimental/Text/src/Text.tsx
```diff
@@ -39,56 +42,45 @@ export type TextProps<TBase = ITextProps> = TBase &
...
-  },
-});
+    [],
+  );
+
+  // return a continuation function that allows this text to be compressed
+  return (extra: TextProps, children: React.ReactNode) => {
+    const mergedProps = { ...rest, ...extra, style: mergeStyles(tokenStyle, props.style, extra.style) };
+    return <RNText {...mergedProps}>{children}</RNText>;
+  };
```
Saadnajmi, line 82: I hadn't thought of how to phrase this in your previous PR, but do we have a description (in documentation or elsewhere) of what it means to be `compressed`? As in, if someone knew to the repo (who knows not of using slots to optimize down the number of renders) comes and looks at this component, would they know what this means?

packages/experimental/Text/src/Text.tsx
```diff
@@ -39,56 +42,45 @@ export type TextProps<TBase = ITextProps> = TBase &
...
-  /** render function for the component */
-  render: (props: TextProps, useSlots: UseSlots<TextType>) => {
-    // stage one, execute any hooks, styling lookups to build the styled slot
-    const Root = useSlots(props).root;
-    // return a function used to complete the render
-    return (rest: TextProps, children: React.ReactNode) => <Root {...mergeProps(props, rest)}>{children}</Root>;
-  },
-});
+    [],
+  );
```
Saadnajmi, line 76: Is the type of thing that buildUseStyling / slotProps were doing behind the scenes for us within `compose` that we now must do manually with `compressible`?

packages/framework/framework/src/useFluentTheme.ts
JasonVMo: Are the apple themes JS only? I think that's a reasonable discussion to have though this line keeps the functionality consistent with what we have been using, so I would split that out from this change.

packages/experimental/Text/src/Text.tsx
JasonVMo: We do in the use-slot README, but likely need to agree on consistent terminology here. I think I've referred to it in the past as a staged component, but that is phrasing things in terms of the implementation details rather than the goal. Maybe I should change the word compressed to compressible here?

packages/framework/framework/src/useFluentTheme.ts
Saadnajmi: They are. The macOS one uses a `react-native-macos` specific API, but that's about it I think.

packages/experimental/Text/src/Text.tsx
JasonVMo: Yep.

- `states: ['disabled']` - this causes the `applyTokenLayers` call to happen
- `tokensThatAreAlsoProps` - causes the equivalent of `patchTokens`
- `slotProps: { root: buildProps< ...` - is the equivalent to the `const [tokenStyle] = cache(` call.
- `filters: { root: filterTextProps }` - is really only required because we aren't splitting the non-text props out of the input props before forwarding them on.

For many components the `applyTokenLayers` or `patchTokens` calls aren't required. In terms of caching styles that is an important concept to understand but the helpers were just as much code as the direct code so they end up being just extra things to learn.

JasonVMo: The rewrite is functionally the same and produces the same render tree, with the exception of not supporting `compose` and the things that can be done via compose.

packages/experimental/Text/src/Text.tsx
Saadnajmi: I think compressible makes sense. In the use-slot README, I saw references to stagedComponent but not compressible. Am I looking at the wrong README? I'm hoping that the right README is easy enough to find by someone ctrl+clicking on `compressible` in VS Code (since that's how I've been going about it). Or also a comment above the function definition when you do a mouse hover over

FalseLobster: I linked Saad our side conversation about the naming here, maybe we should've renamed stagedComponent to compressible like you originally suggested? Can you still use compressible without customize? (If so, does that basically make it the same as stagedComponent?) or did we make the useTokens param required?

Saadnajmi: The reason for my comment here is less about the name, and making the name's meaning discoverable to someone who is looking at this code for the first time. The name feels appropriate to how I understand it, but I imagine if this was my first time looking at the code I wouldn't have semantically understood what compressible meant, that it was related to stagedComponent, or that the README in use-slots is the right place to learn that info.

packages/experimental/Text/src/Text.tsx
Saadnajmi: Thanks! I like that this example has comments that broke into steps the levels of layering that are happening. I think that will definitely be helpful in the future for also understanding how `compose` layers tokens and styles onto a component.

packages/experimental/Text/src/Text.tsx
rurikoaraki: I would +1 this. I've been thinking that it would be good to do a pass and comment all of these functions with JSDoc styling so that when you're typing with something that has Intellisense, the function's purpose is easy to see (similar to what happens when you use things from RN). Personally wouldn't find this blocking, but I think it would make it easier to look at the framework and work with it if there's documentation accessible from the IDE itself.

JasonVMo: That makes sense. I would propose the overall pattern use the words compress or compressible in discussions, then I would likely rename stagedComponent to stagedRender as the implementation detail. I would likely create an alias as well (const stagedComponent = stagedRender) to reduce the risk of breaks. I would do that in a separate PR I think.

Saadnajmi: Thanks. I'm starting to feel that some of these things should be split from the PR like the `useFluentTheme` hook and documentation/intellisense documentation updates.

Saadnajmi (approved):

PR #836, opened 2021-07-30
Tabs Control for win32

dake3601:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Implemented Tabs Control for win32.

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
```diff
@@ -0,0 +1,251 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { Tabs, TabsItem, Text } from '@fluentui/react-native';
+import { stackStyle } from '../Common/styles';
+import { TABS_TESTPAGE } from './consts';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { SvgIconProps } from '@fluentui-react-native/icon';
+import { Button } from '@fluentui-react-native/button';
```
ksiler, line 8: Place Button up on line 3 with all the other FURN control imports?

packages/components/Tabs/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@uifabricshared/foundation-compose": "^1.9.5",
+    "@uifabricshared/foundation-composable": ">=0.8.2 <1.0.0",
+    "@uifabricshared/foundation-settings": ">=0.9.1 <1.0.0",
+    "@fluentui-react-native/adapters": ">=0.6.2 <1.0.0",
+    "@fluentui-react-native/focus-zone": ">=0.6.0 <1.0.0",
+    "@fluentui-react-native/interactive-hooks": ">=0.10.10 <1.0.0",
+    "@fluentui-react-native/pressable": ">=0.6.11 <1.0.0",
+    "@fluentui-react-native/text": ">=0.9.8 <1.0.0",
+    "@fluentui-react-native/tokens": ">=0.9.4 <1.0.0"
+  },
```
ksiler, line 37: Looks like the CI Loop is complaining about a couple unused dependencies here. I'd remove contextual menu and pressable if they're not being used and add icon since it's missing.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+   ** The currently selected TabsItem's key
+   */
+  selectedKey: string | null;
+
+  /*
+   ** Index of currently selected key
+   */
+
+  getTabId?: (key: string, index: number) => void;
+
```
ksiler, line 22: I thought this function was supposed to allow developers to customize how IDs are generated for each tab item (see [Fluent UI Pivot](https://developer.microsoft.com/en-us/fluentui#/controls/web/pivot)). If that's still the case, I'd probably add more details in the comment above it and make the return type 'string' instead of 'void'.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+export interface TabsInfo {
+  headersOnly?: boolean;
+  label?: boolean,
+}
+
+export interface TabsState {
+  context: ITabsContext;
+  info: TabsInfo;
+}
```
ksiler, line 52: Is there a reason to have both `context` and `info` here? Could the properties in `TabsInfo` be added to `ITabsContext`?

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+import { ITextProps } from '@fluentui-react-native/text';
+import type { IViewWin32Props } from '@office-iss/react-native-win32';
+import { IRenderData } from '@uifabricshared/foundation-composable';
+import { IForegroundColorTokens, FontTokens, IBackgroundColorTokens } from '@fluentui-react-native/tokens';
+import { FocusZoneProps } from '@fluentui-react-native/focus-zone';
+import { View, ViewProps } from 'react-native';
+
+export const tabsName = 'Tabs';
+
+export interface ITabsContext {
```
ksiler, line 11: Is it possible to name this `TabsContext` instead of `ITabsContext`?

packages/components/Tabs/src/Tabs.types.ts
ksiler: Oh sorry I see that there's a description for it below under `TabsProps`. In that case, could you make the function return type 'string' instead of 'void'?

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /*
+   ** Sets whether to only render the header
+   */
+  headersOnly?: boolean;
+
+  /*
+   ** Sets whether to only render the header
+   */
+  componentRef?: React.RefObject<View>;
```
ksiler, line 96: Change or remove the comment for this property since it's copied from above.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  testID?: string;
+}
+
+export interface TabsTokens extends IForegroundColorTokens, FontTokens, IBackgroundColorTokens {}
+
+export interface TabsSlotProps {
+  root: React.PropsWithRef<IViewWin32Props>;
+  label: ITextProps;
+  container: ViewProps & FocusZoneProps;
```
ksiler, line 106: Could you explain why ViewProps and FocusZoneProps are listed here for container?

packages/components/Tabs/src/TabsItem.settings.ts
```diff
@@ -0,0 +1,119 @@
...
+      variant: 'bodyStandard',
+      borderWidth: 2,
+      // borderColor: 'neutralForeground1',
+      borderColor: 'transparent',
+      borderRadius: 4,
+    },
+    root: {
+      accessible: true,
+      focusable: true,
+      accessibilityRole: 'button',
```
ksiler, line 20: accessibilityRole should be 'tab'

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+export const TabsItem = compose<TabsItemType>({
+  displayName: tabsItemName,
+
+  usePrepareProps: (userProps: TabsItemProps, useStyling: IUseComposeStyling<TabsItemType>) => {
+    const {
+      icon,
+      headerText = '',
+      onAccessibilityTap = userProps.onClick,
+      accessibilityLabel = userProps.headerText,
```
ksiler, line 30: Should this be `ariaLabel` instead?

ksiler: Actually, I see how `accessibilityLabel` works. Now I'm wondering what would happen if someone sets `ariaLabel` and `accessibilityLabel` on TabItem. Would the `ariaLabel` property work?

ksiler: Also, I think the count should be included in the accessibilityLabel if you're generating it from userProps

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,97 @@
...
+  componentRef?: React.RefObject<IFocusable>;
+}
+
+export interface TabsItemTokens extends IButtonTokens {
+  textBorderColor?: string;
+}
+
+export interface TabsItemSlotProps {
+  root: React.PropsWithRef<IViewProps>;
+  ripple?: PressableProps; // This slot exists to enable ripple-effect in android. It does not affect other platforms.
```
ksiler, line 83: I'd remove this slot since it looks specific to Android and this PR is for Win32

packages/components/Tabs/src/Tabs.types.ts
dake3601: Because we need to style the container that holds the TabsItems with flex. If we just have FocusZoneProps TypeScript complains about styling the FocusZone.

packages/components/Tabs/src/TabsItem.tsx
Saadnajmi: ^ I was about to ask the same questions. Note that if accessibilityLabel included count, I think we would then need some localization to get the right string format? i.e: ` "headerText + " " + count ` might not be the correct ordering for every language.

packages/components/Tabs/src/Tabs.types.ts
dake3601: On Tabs.tsx, line 15, we bumped into the need of having two variables with the same name, se we kept the "I" for that one:
`export const TabsContext = React.createContext<ITabsContext>({...});`

We are open to any suggestions for renaming them.

packages/components/Tabs/src/Tabs.types.ts
nabil-k: Yeah the reason we have both context and info is because context is for properties that we want the children of Tabs to have access for. While info just stores information about the Tabs state that we want only Tabs to know about.

packages/components/Tabs/src/TabsItem.tsx
nabil-k: When you say "might not be the correct ordering for every language" are you saying that in some languages people prefer to hear/read the number first then the item?

dake3601: > Actually, I see how `accessibilityLabel` works. Now I'm wondering what would happen if someone sets `ariaLabel` and `accessibilityLabel` on TabItem. Would the `ariaLabel` property work?

Currently it would not work. Should we keep accessibilityLabel or ariaLabel as the prop to pass to TabsItem?

Saadnajmi: Yes. In general when an accessibilityLabel is a combination of multiple elements in code (Like a First Name + Last Name, label + count, x of y items, etc), it needs to be localized. In native apple code, we have a system for doing that. I think downstream SDX'es that use FURN have a system to localize their code as well. FURN doesn't quite yet, but it hasn't been a problem AFAIK because most of our labels have just been static text or roles like `tab` (which presumably would get localized by the OS). I'm not sure how it works on win32 but I'm sure Krystal is the right person to ask there

ksiler: > Should we keep accessibilityLabel or ariaLabel as the prop to pass to TabsItem?

I think we should use accessibilityLabel for both Tabs and TabsItem. I believe ariaLabel was added to RadioGroup/Button and Checkbox by @samuelfreiberg but they appear to be the same property. Sammy, can you remind me why ariaLabel was added as a property on those controls? Could accessibilityLabel have been used instead? Having both is a bit confusing.

ksiler: As for localization, I agree with everything Saad said. I don't have a good answer for how we'd localize it on the FURN side so I'll have to think more on that. But allowing the default value for accessibilityLabel to be overwritten will allow developers to have some control if the default value doesn't work for them.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,152 @@
+/** @jsx withSlots */
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text } from '@fluentui-react-native/text';
+import { FocusZone } from '@fluentui-react-native/focus-zone';
+import { tabsName, TabsType, TabsProps, TabsState, TabsSlotProps, TabsRenderData, ITabsContext } from './Tabs.types';
+import { compose, IUseComposeStyling } from '@uifabricshared/foundation-compose';
```
Saadnajmi, line 7: Not blocking, but it's worth noting that this control is made with the legacy framework, rather than the newer compose,  which I will try to stop calling experimental since I don't think it is at this point.

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,96 @@
...
+  key: string;
+}
+
+export interface TabsItemState {
+  info: TabsItemInfo;
+}
+
+// Props for the tabs item
+export interface TabsItemProps extends IButtonProps {
+  /*
```
rurikoaraki, line 46: nit: I think the commenting syntax for JSDoc is supposed to be two asterisks following the opening slash (so similar to the comments in the Info interface)

packages/components/Tabs/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "lint": "fluentui-scripts eslint",
+    "test": "fluentui-scripts jest",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "verify-api": "fluentui-scripts verify-api-extractor",
+    "update-api": "fluentui-scripts update-api-extractor",
+    "prettier": "fluentui-scripts prettier",
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "dependencies": {
+    "@fluentui-react-native/button": "^0.17.9",
```
rurikoaraki, line 26: super nit: we tend to try and keep these in alphabetical order

packages/components/Tabs/src/TabsItem.settings.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export const tabsItemSelectActionLabel = 'Select a TabsItem';
+
+export const settings: IComposeSettings<TabsItemType> = [
+  {
+    tokens: {
+      color: 'neutralForeground3Brand',
+      variant: 'bodyStandard',
+      borderWidth: 2,
+      // borderColor: 'neutralForeground1',
```
rurikoaraki, line 13: nit: Please remove commented code

packages/components/Tabs/package.json
Saadnajmi: Running `yarn add <package>` at the `packages/components/Tab` directory will also auto sort it for you

samuelfreiberg:
Have you updated the spec document with all the final engineering decisions?

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  label?: string;
+
+  /*
+   ** The key of the TabsItem that will initially be selected
+   */
+  defaultSelectedKey?: string;
+
+  /*
+   **  OPTIONAL: An aria label for narrator. If not provided, it will be set to the label of the Tabs
```
samuelfreiberg, line 65: Nit: Aren't all these props optional? I would remove the "OPTIONAL:" text here.

packages/components/Tabs/src/Tabs.settings.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const settings: IComposeSettings<TabsType> = [
+  {
+    tokens: {
+      color: 'menuItemText',
+    },
+    root: {
+      accessible: true,
+      accessibilityRole: 'tablist',
+      style: {},
```
rurikoaraki, line 12: nit: Is this empty style prop needed?

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  ariaLabel?: string;
+
+  /*
+   ** The key of the selected option. If you provide this, you must maintain selection state by observing
```
samuelfreiberg, line 70: I would add text that says this is mutually exclusive to defaultSelectedKey. I know you stated it below, but I think mutually exclusive makes more sense

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { ISlots, withSlots } from '@uifabricshared/foundation-composable';
+import { settings } from './Tabs.settings';
+import { filterViewProps } from '@fluentui-react-native/adapters';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { foregroundColorTokens, textTokens, backgroundColorTokens } from '@fluentui-react-native/tokens';
+import { useSelectedKey } from '@fluentui-react-native/interactive-hooks';
+
+export const TabsContext = React.createContext<ITabsContext>({
+  selectedKey: null,
+
```
samuelfreiberg, line 17: Nit: Remove extra whitespace on line. 17 and 28

packages/components/Tabs/src/Tabs.types.ts
rurikoaraki: Maybe something like `TabsContextData`?

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,152 @@
...
+      </TabsContext.Provider>
+    );
+  },
+
+  settings,
+  slots: {
+    root: View,
+    label: Text,
+    tabPanel: { slotType: View, filter: filterViewProps },
+    container: FocusZone,
```
samuelfreiberg, line 143: Isn't this also a View?

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,96 @@
+import * as React from 'react';
+import { ViewProps } from 'react-native';
+import { IRenderData } from '@uifabricshared/foundation-composable';
+import type { IViewProps } from '@fluentui-react-native/adapters';
+import { IFocusable, IPressableState } from '@fluentui-react-native/interactive-hooks';
+import { ITextProps } from '@fluentui-react-native/text';
+import { IconProps } from '@fluentui-react-native/icon';
+import { IButtonProps, IButtonTokens } from '@fluentui-react-native/button';
+
+export const tabsItemName = 'TabsItem';
```
samuelfreiberg, line 10: Did we do research on what this should be called? IMO "tab header" sounds more intuitive, but I could go either way

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export interface TabsItemProps extends IButtonProps {
+  /*
+   ** The text string for the option
+   */
+  headerText?: string;
+
+  /*
+   ** The text string for the TabsItem count
```
samuelfreiberg, line 51: What do you mean by the "text string"? This is a number, not a string?..

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return (
+      <TabsContext.Provider
+        // Passes in the selected key and a hook function to update the newly selected tabsItem and call the client's onTabsClick callback
+        value={renderData.state.context}
+      >
+        <Slots.root>
+          {renderData.state.info.label && <Slots.label />}
+          <Slots.container>{children}</Slots.container>
+          <Slots.tabPanel>
+            <TabsContext.Consumer>
```
rurikoaraki, line 129: Not blocking: This doesn't seem like a great place to use React's Context - maybe I'm missing something, but to me it looks like all of the information that is provided by the context is immediately here or can be provided to the TabsItem by the Tabs, so I feel like it's a bit overkill to use Context here.

I recognize that you probably followed a pattern from a different part of the codebase, so not blocking since there's precedent for this already.

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,96 @@
...
+import { IButtonProps, IButtonTokens } from '@fluentui-react-native/button';
+
+export const tabsItemName = 'TabsItem';
+
+export interface TabsItemInfo extends IPressableState {
+  /**
+   * Disables the TabsItem.
+   * @default false
+   */
+  disabled?: boolean;
```
samuelfreiberg, line 17: Why does this need to be in the state? Why not just props?

I don't believe you need this in the state if you're not using it in the render() function.

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * TabsItem text.
+   */
+  headerText?: boolean;
+
+  /**
+   * Indicates if TabsItem is selected.
+   */
+  selected: boolean;
```
samuelfreiberg, line 32: Same comment as the one above

packages/components/Tabs/src/TabsItem.settings.ts
```diff
@@ -0,0 +1,95 @@
...
+        display: 'flex',
+        alignItems: 'center',
+        flexDirection: 'column',
+        alignSelf: 'flex-start',
+        justifyContent: 'center',
+      },
+    } as IViewProps,
+    indicator: {
+      style: {
+        minHeight: 2,
```
samuelfreiberg, line 31: Have you tested this at all DPIs?

packages/components/Tabs/src/Tabs.tsx
rurikoaraki: Okay I see that you're keeping children pretty flexible so you can't guarantee that a TabItem is the only child of Tabs. So I guess it makes sense to have a Context to keep things flexible. Though I kind of wonder what's expected to happen if you get a non-TabsItem child

samuelfreiberg:
E2E Testing is incomplete, please see: https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/E2E#authoring-e2e-test

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+      onClick,
+      itemKey,
+      itemCount,
+      ...rest
+    } = userProps;
+
+    // Grabs the context information from Tabs (currently selected TabsItem and client's onTabsClick callback)
+    const info = React.useContext(TabsContext);
+
+    /* We don't want to call the user's onTabsClick multiple times on the same selection. */
```
rurikoaraki, line 42: This is more of an opinion but I'd personally err on being "dumb," for lack of a better word, and send all events as we get them. We should let the user decide how to handle duplicate clicks and not filter events out preemptively, in case they want that information (i.e. multiple clicks may indicate a perf issue).

That said I'm not sure how "opinionated" FURN usually is so I'd be curious to hear other people's thoughts.

samuelfreiberg (changes requested):

apps/fluent-tester/src/FluentTester/testPages.win32.ts
```diff
@@ -110,6 +111,11 @@ export const tests: TestDescription[] = [
     component: MenuButtonTest,
     testPage: HOMEPAGE_MENU_BUTTON,
   },
+  {
```
samuelfreiberg, line 114: E2E testing is incomplete, there's no spec document.

Please see: https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/E2E#authoring-e2e-test

packages/components/Tabs/src/TabsItem.tsx
rurikoaraki: What is "count" here? Is that supposed to be `accessibilityPosInSet`?

I think if we set `accessibilityPosInSet`/`accessibilitySetSize` the screen readers will do the loc work for us

packages/components/Tabs/src/TabsItem.tsx
samuelfreiberg: It looks like this specific piece of code was taken from RadioButton. If I remember correctly (this was a long time ago), the user callback was being called multiple times on **one** click (don't quite remember why, I think it had to deal with changing selection on both onFocus and onPress)

It looks like something similar is going on here, where there's two slightly different things going on here between onPress and onFocus.

packages/components/Tabs/src/TabsItem.types.ts
dake3601: Selected seems to be needed to be kept in the state for styling. If we remove it from the state, when there is no focus on the control, the selected styling disappears.

packages/components/Tabs/src/TabsItem.tsx
rurikoaraki: I see. I think it might be worth updating the comment to be more specific about the issue that's being avoided. at least to me it currently reads more like a behavior decision than a fix for a bug.

packages/components/Tabs/src/TabsItem.types.ts
dake3601: We named it Tabs**Item** taking from XAML and Fluent UI using Pivot**Item**. Also, Fluent UI seems to be moving to [TabItem](https://github.com/microsoft/fluentui/blob/master/packages/react-tabs/src/components/Tabs/TabItem.tsx).

packages/components/Tabs/src/Tabs.tsx
ksiler: @samuelfreiberg are you referring to the container slot?

packages/components/Tabs/src/TabsItem.tsx
ksiler: @rurikoaraki "count" here refers to an item count you can add to the tab item text like:

image

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,90 @@
...
+  /*
+   ** Whether or not the tabs item is selectable
+   */
+  disabled?: boolean;
+
+  /*
+   ** An optional string for the Narrator to read for each TabsItem. If not provided, this will be set to the tabsItem's content
+   */
+  ariaLabel?: string;
+
```
ksiler, line 63: I think you should also add `ariaPosInSet` and `ariaSetSize` to TabItemProps so someone can define the n of m properties if they aren't generated (for instance, if TabsItems isn't a direct child of Tabs). You can find an example of the implementation in RadioButton: https://github.com/microsoft/fluentui-react-native/blob/master/packages/components/RadioGroup/src/RadioButton.tsx#L85

apps/fluent-tester/src/E2E/BootPage/specs/BootPages.spec.win.ts
```diff
@@ -118,6 +119,12 @@ describe('Click on each test page and check if it renders', function () {
     expect(TextTestPage.isPageLoaded()).toBeTruthy();
   });
 
+  it('Tabs Test Page', () => {
```
samuelfreiberg, line 122: Does this also work on UWP? Should it be added to the Windows E2E testing as well?

dake3601: It does not work because there is no FocusZone on UWP. The Tabs UWP Control is still a work in progress. We will probably submit another PR for it.

packages/components/Tabs/src/TabsItem.tsx
samuelfreiberg: @ksiler I don't quite remember why we decided on exposing `ariaLabel` instead of `accessibilityLabel` on RadioGroup/RadioButton/Checkbox, but I remember we had a conversation about it and that's what we agreed on during the spec review.

I'm ok with using accessibilityLabel here instead of ariaLabel, and then perhaps we can look into deprecating ariaLabel for accessibilityLabel in RadioGrou/Checkbox

packages/components/Tabs/src/TabsItem.settings.ts
nabil-k: No we haven't, when you say DPI do you mean dots per inch? Also how would we go about testing it.

packages/components/Tabs/src/Tabs.types.ts
ksiler: Maybe you could create another slot to handle the style/orientation of the container holding the children of Tabs? Named "stack" or "inner" or something else and have it be a View or Stack? Just throwing out an idea. Not sure if it's preferable to the solution you already have.

packages/components/Tabs/src/TabsItem.settings.ts
samuelfreiberg: Yes, dots per inch. You can easily change it in your computer settings. I would try each configuration and just visually make sure this component looks normal.

packages/components/Tabs/src/Tabs.tsx
nabil-k: If there's a non tabsitem thats a child it will simply just get rendered and it's expected not to affect the functionality of the tabs control. We also expect in general that a user won't try to pass non pivot items.

packages/components/Tabs/src/TabsItem.types.ts
nabil-k: Thanks for the reference, those props have been added.

packages/components/Tabs/src/TabsItem.settings.ts
dake3601: We tried it and it looks good!

packages/components/Tabs/src/TabsItem.tsx
dake3601: @samuelfreiberg do you have a better way of wording this?

packages/components/Tabs/src/Tabs.tsx
samuelfreiberg: Container. I was confused because in the .types file, the container is: `ViewProps & FocusZoneProps`

dake3601: The container is a FocusZone because we need its keyboarding properties. We assigned ViewProps in the .types files to be able to style the focusZone (without ts complaining) in order to have row flex around the tabsitems.

ksiler: @samuelfreiberg should FocusZone have ViewProps as a base? Similar to FocusTrapZone?

ksiler: another option might be to do:

```
 <Slots.container>
        <View style={{flexDirection:'row'}}>
              {children}
        </View>
 </Slots.container>
```

but I'm not sure if that's preferable

packages/components/Tabs/src/Tabs.types.ts
samuelfreiberg: A Krystal said above, I would make a separate Slot called stack or something. Make this a View and have that be a child to the FocusZone. 

I'm also thinking about the future of the control. If we end up allowing the user to change the orientation (vertical vs. horizontal), and other styling things, it makes more sense to separate this into separate slots.

packages/components/Tabs/src/Tabs.tsx
samuelfreiberg: I made a post below, but here's what I said:

I would make a separate slot called stack. Make this a View and have that be a child to the FocusZone. This should contain the styling info for the children of the FocusZone.

If we end up allowing the user to change the orientation (vertical vs. horizontal), and other styling props, it makes more sense to separate this into separate slots.

packages/components/Tabs/src/TabsItem.tsx
dake3601: I set accessibilityLabel, accessibilityPosInSet, and accessibilitySetSize as the accessibility props.

packages/components/Tabs/src/Tabs.settings.ts
```diff
@@ -0,0 +1,27 @@
...
+    tokens: {
+      color: 'menuItemText',
+    },
+    root: {
+      accessible: true,
+      accessibilityRole: 'tablist',
+    },
+    label: {
+      style: {
+        fontFamily: 'inherit',
```
samuelfreiberg, line 15: I would remove this - it recently caused an issue with RadioGroup on MacOS and I don't think it's necessary here

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,150 @@
...
+      context: {
+        selectedKey: selectedKey ?? data.selectedKey,
+        onTabsClick: data.onKeySelect,
+        getTabId: onChangeTabId,
+        updateSelectedTabsItemRef: onSelectTabsItemRef,
+        views: map,
+      },
+      info: {
+        headersOnly: headersOnly ?? false,
+        label: !!label && Platform.OS === 'windows',
```
samuelfreiberg, line 78: Why is `Platform.OS === 'windows'` needed here?

dake3601: Because only the windows Tabs Control has a label on native examples found. Should we still include it for all platforms?

samuelfreiberg: But the Label prop is available to all other platforms (via Tabs.types.ts), correct? It wouldn't be user-friendly if we surface the prop to all platforms, and say a MacOS user sets the label prop, and then we don't use it. The user will get confused and most likely file a bug.

Since it's an optional prop (and pretty harmless one, at that), I say we use it for all platforms.

dake3601: Correct, added it back to all platforms.

packages/components/Tabs/src/Tabs.settings.ts
```diff
@@ -0,0 +1,26 @@
...
+      color: 'menuItemText',
+    },
+    root: {
+      accessible: true,
+      accessibilityRole: 'tablist',
+    },
+    label: {
+      style: {
+        fontSize: 16,
+        fontWeight: '600',
```
ksiler, line 16: Are these styles being pulled from somewhere?

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  usePrepareProps: (userProps: TabsProps, useStyling: IUseComposeStyling<TabsType>) => {
+    const {
+      label,
+      accessibilityLabel = userProps.label,
+      selectedKey,
+      headersOnly,
+      defaultSelectedKey,
+      getTabId,
+      componentRef = React.useRef(null),
```
ksiler, line 41: There was a recent change that fixed the usage of conditional hooks: https://github.com/microsoft/fluentui-react-native/pull/841 

I'd apply the same change here as well.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -0,0 +1,159 @@
...
+export const TabsItem = compose<TabsItemType>({
+  displayName: tabsItemName,
+
+  usePrepareProps: (userProps: TabsItemProps, useStyling: IUseComposeStyling<TabsItemType>) => {
+    const {
+      icon,
+      headerText = '',
+      onAccessibilityTap = userProps.onClick,
+      accessibilityLabel = userProps.headerText,
+      componentRef = React.useRef(null),
```
ksiler, line 31: Same comment I made in Tabs.tsx. Refer to https://github.com/microsoft/fluentui-react-native/pull/841

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Key for tabItem
+   */
+  key: string;
+}
+
+export interface TabsItemState {
+  info: TabsItemInfo;
+}
+
+export interface TabsItemProps extends IButtonProps {
```
ksiler, line 38: Should we use IButtonProps here? Or instead list the props for the individual slots (like icon, text, etc.)?

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,97 @@
...
+  accessibilitySetSize?: number;
+
+  /**
+   * A RefObject to access the IFocusable interface. Use this to access the public methods and properties of the component.
+   */
+  componentRef?: React.RefObject<IFocusable>;
+}
+
+export interface TabsItemTokens extends IButtonTokens {
+  textBorderColor?: string;
```
ksiler, line 78: I think this might have been copied over from RadioButton. If it's not used, it should be removed.

packages/components/Tabs/src/TabsItem.types.ts
```diff
@@ -0,0 +1,97 @@
...
+    */
+  accessibilitySetSize?: number;
+
+  /**
+   * A RefObject to access the IFocusable interface. Use this to access the public methods and properties of the component.
+   */
+  componentRef?: React.RefObject<IFocusable>;
+}
+
+export interface TabsItemTokens extends IButtonTokens {
```
ksiler, line 77: Same comment as above. Should we be using IButtonTokens here?

packages/components/Tabs/src/Tabs.tsx
nabil-k: Added that change in.

packages/components/Tabs/src/Tabs.settings.ts
ksiler: Talked about this. The styles are pulled in from WinUI figma files. I think we should make these more platform specific. So if you're adding a label for win32 maybe use `variant: BodySemibold`.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+      },
+      [setSelectedTabsItemRef],
+    );
+
+    const onChangeTabId = React.useCallback((key: string, index: number) => {
+      if (getTabId) {
+        return getTabId(key, index);
+      }
+      return `${key}-Tab${index}`;
+    }, []);
```
samuelfreiberg, line 64: Should `getTabId` be in the dependencies array?

nabil-k: Yeah it should, I've added that.

packages/components/Tabs/src/TabsItem.types.ts
sriya112: Just updated so we're not using IButtonProps anymore

packages/components/Tabs/src/TabsItem.types.ts
sriya112: Updated so that we're not using IButtonTokens anymore

packages/components/Tabs/src/Tabs.settings.ts
sriya112: For win32, it's now updated to `SubheaderSemibold`

packages/components/Tabs/src/TabsItem.tsx
rurikoaraki: @dake3601 I would probably want something like:

"There's a bug where the user callback is being called multiple times on one click. We check that there is an actual change in selection before forwarding the message to the user callback so that we don't process more than one callback for each click"

samuelfreiberg:
There are two empty files being checked in (I can't leave a comment on them):
packages/components/Tabs/CHANGELOG.json
packages/components/Tabs/CHANGELOG.md

Are these needed? Why are the empty?

dake3601:
> There are two empty files being checked in (I can't leave a comment on them):
> packages/components/Tabs/CHANGELOG.json
> packages/components/Tabs/CHANGELOG.md
> 
> Are these needed? Why are the empty?

Those empty files where added as we followed the [contributing guide](https://github.com/microsoft/fluentui-react-native/blob/master/CONTRIBUTING.md#creating-a-new-component), specifically step 2 of creating a new component mentions that we should "Make sure to empty the changelog file".

PPatBoyd:
is tabs changelog.md being empty correct? 🤔

dake3601:
Yes, and changelog.md and changelog.json

samuelfreiberg (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+const tabsRenderSeparately: React.FunctionComponent = () => {
+  const [selectedKey, setSelectedKey] = React.useState('rectangleRed');
+
+  const onTabsClick = (key: string) => {
+    setSelectedKey(key);
+  };
+
+  const getTabId = (key: string) => {
+    return `ShapeColorPivot_${key}`;
```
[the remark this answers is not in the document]
nabil-k: > j.w. is there specific gain from naming and providing the key? Does the tab control default to integers that can be used by default, or is it a required prop? For generated tabs I'm not sure what having ShapeColorPivot_# vs just # would enable; or add extra steps to handling the tab change?

For this example the reason we have ShapeColorPivot_# is for the Narrator to read out to the user what the current color for the selected shape is. getTabId is more for accessibility use cases.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { tabsName, TabsType, TabsProps, TabsState, TabsSlotProps, TabsRenderData, TabsContextData } from './Tabs.types';
+import { compose, IUseComposeStyling } from '@uifabricshared/foundation-compose';
+import { ISlots, withSlots } from '@uifabricshared/foundation-composable';
+import { settings } from './Tabs.settings';
+import { mergeSettings } from '@uifabricshared/foundation-settings';
+import { filterViewProps } from '@fluentui-react-native/adapters';
+import { foregroundColorTokens, textTokens, backgroundColorTokens } from '@fluentui-react-native/tokens';
+import { useSelectedKey } from '@fluentui-react-native/interactive-hooks';
+
+export const TabsContext = React.createContext<TabsContextData>({
```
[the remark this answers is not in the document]
PPatBoyd: I didn't understand createContext and Provider/Consumer, ignore :)

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
```diff
@@ -0,0 +1,257 @@
...
+        </TabsItem>
+      </Tabs>
+    </View>
+  );
+};
+
+const tabsCountIcon: React.FunctionComponent = () => {
+  const svgProps: SvgIconProps = {
+    src: TestSvg,
+    viewBox: '0 0 500 500',
```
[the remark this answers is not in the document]
nabil-k: The viewbox is set in the svg however for some reason removing  viewBox: '0 0 500 500' prevents the svg from being visible,

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+export interface TabsProps extends Pick<FocusZoneProps, 'isCircularNavigation'> {
+  /*
+   ** Descriptive label for the Tabs. This will be displayed as the title of the Tabs to the user
+   */
+  label?: string;
+
+  /*
+   ** The key of the TabsItem that will initially be selected
+   */
+  defaultSelectedKey?: string;
```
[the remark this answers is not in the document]
nabil-k: The updateSelectedTabsItemRef is only triggered after the user clicks on something new, at which point defaultSelectedKey no longer matters.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+   ** Sets whether to only render the header
+   */
+  headersOnly?: boolean;
+
+  /*
+   ** A RefObject to access Tabs.
+   */
+  componentRef?: React.RefObject<View>;
+
+  testID?: string;
```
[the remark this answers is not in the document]
samuelfreiberg: Used for E2E testing

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+    } = userProps;
+
+    // Grabs the context information from Tabs (currently selected TabsItem and client's onTabsClick callback)
+    const info = React.useContext(TabsContext);
+
+    /* We don't want to call the user's onTabsClick multiple times on the same selection. */
+    const changeSelection = () => {
+      if (itemKey != info.selectedKey) {
+        info.onTabsClick && info.onTabsClick(itemKey);
+        info.getTabId && info.getTabId(itemKey, info.tabsItemKeys.findIndex(x => x == itemKey) + 1);
```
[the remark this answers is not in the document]
nabil-k: We called it getTabId because getTabId is a callback function that is available for the users to overwrite via the tab's getTabId prop. The same goes for onTabsClick. In the changeSelection function we are just setting up the callback functions (onTabsClick and getTabId) with the necessary updated values  for the user to use.

packages/components/Tabs/src/Tabs.types.ts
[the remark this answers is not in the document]
sriya112: It's been added in other components and we saw:

'If testID is specified in React Native app for Windows, the locator strategy should choose accessibility id. A unique accessiblity id/testID per Window is recommended for React Native Windows E2E testing when authoring the test app and test cases.

To use this, we must add a prop to our component or UI element in question called “testID”. In our test page, set the “testID” for the component, and we can then select it in our Page Object using the imported By method above from a base class.'

[here](https://github.com/microsoft/fluentui-react-native/blob/457ed8ab797732cd6128cb6847dd9053cd69660e/apps/fluent-tester/src/E2E/README.md)

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+        info.updateSelectedTabsItemRef && componentRef && info.updateSelectedTabsItemRef(componentRef);
```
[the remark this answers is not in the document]
nabil-k: We need that in order to set the defaultTabbableElement. It allows users to tab out in the middle of a pivot control and resume where they once where.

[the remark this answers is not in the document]
nabil-k: I wrote a reply to this yesterday but I guess I forgot to submit my reply haha.  We have updateSelectedTabsItemRef  as a callback because  we would like to set the defaultTabbableElement through the updateSelectedTabsItemRef  method.

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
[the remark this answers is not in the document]
nabil-k: > Maybe there could be issues with dynamic tabs? I'm mostly thinking about static tabs expecting that's the most common scenario, and that dynamic would raise the value in choosing to specify keys. I don't have any issues coming to mind that would make autogen keys specifically bad, but I have less insight into the tab control patterns

Yeah I can't seem to find anything particularly bad about autogen. Yesterday we took ur feedback and looked into it and today we have a rough solution for autogen keys but we're looking for ways to improve it.

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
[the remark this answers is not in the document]
ksiler: @warren-ms do you know why the viewBox setting in TestSvg doesn't seem to work and we have to set it as part of SvgIconProps?

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /*
+   ** Updates the selected tabsItem and calls the client’s onTabsClick callback
+   */
+  onTabsClick?: (key: string) => void;
+
+  /*
+   ** Updates the selected tabsItem's ref to set as the default tabbable element
+   */
+  updateSelectedTabsItemRef?: (ref: React.RefObject<any>) => void;
```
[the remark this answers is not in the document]
ksiler: If a key was used, we'd need a way to get the associated ref for that key since `defaultTabbableElement` is expecting a ref value for FocusZone.

apps/fluent-tester/src/FluentTester/testPages.windows.ts
```diff
@@ -62,6 +63,11 @@ export const tests = [
     component: CheckboxTest,
     testPage: HOMEPAGE_CHECKBOX_BUTTON,
   },
+  {
```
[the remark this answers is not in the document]
dake3601: We only added that control to win32 as we only spec'd it out for win32, windows and macos. Even then, there are style, keyboarding, and accessibility differences among these platforms. 

We added it to the windows tester app even though it is not currently working because of a dependency to FocusZone, like RadioGroup.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+  tabsItemKeys?: string[];
+
+  /*
+   ** A Map to for a TabItems corresponding view
+   */
+  views?: Map<string, React.ReactNode[]> | null;
+}
+
+export interface TabsInfo {
+  headersOnly?: boolean;
```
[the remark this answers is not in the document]
dake3601: We have Info on the state to decide if a slot will be rendered, which is a behavior we found on other components such as Button. I believe that this is because you can't directly access user props on the render.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /*
+   ** The key of the TabsItem that will initially be selected
+   */
+  defaultSelectedKey?: string;
+
+  /*
+   ** An accessibility label for narrator. If not provided, it will be set to the label of the Tabs
+   */
+  accessibilityLabel?: string;
```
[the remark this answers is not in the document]
dake3601: We thought of adding it here to be obvious on the props available for the control, but it is redundant. Should we remove it from here?

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+  selectedKey: null,
+  onTabsClick: (/* key: string */) => {
```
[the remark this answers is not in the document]
dake3601: We don't know if there is a way to pass optional props to the context on initialization. Other Controls available follow the same structure.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+      </TabsContext.Provider>
+    );
+  },
+
+  settings,
+  slots: {
+    root: View,
+    label: Text,
+    container: FocusZone,
+    stack: { slotType: View, filter: filterViewProps },
```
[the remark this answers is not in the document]
dake3601: The stack is what gives Tabs its styling and container provides with keyboarding. Before we had a FocusZone with View Stylings, but on previous comment reviews, this was the suggested way to handle the slots.

packages/components/Tabs/src/TabsItem.tsx
ksiler: Yes, I just started a thread about it. I'll have a PR out shortly.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /*
+   ** Array of tabsItem keys in the group
+   */
+  tabsItemKeys?: string[];
+
+  /*
+   ** A Map to for a TabItems corresponding view
+   */
+  views?: Map<string, React.ReactNode[]> | null;
```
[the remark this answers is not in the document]
nabil-k: Tabs is what knows what the currently selected tabs item is, so by doing something like views.get(selectedTabItem) in tabs we're able to easily get and render the selected view. If the views where held by tabs item it would be difficult to propagate that data from tabs item to tabs.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+      root: { rest, ref: componentRef, accessibilityLabel: accessibilityLabel, accessibilityRole: 'tablist'},
+      label: { children: label },
+      container: { isCircularNavigation: isCircularNavigation, defaultTabbableElement: selectedTabsItemRef },
+    });
+
+    return { slotProps, state };
+  },
+
+  render: (Slots: ISlots<TabsSlotProps>, renderData: TabsRenderData, ...children: React.ReactNode[]) => {
+    if (renderData.state == undefined) {
```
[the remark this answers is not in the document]
dake3601: We are not sure, we based it off from RadioGroup. We have updated ```renderData.state == undefined``` to ```!renderData.state```.

packages/components/Tabs/src/Tabs.types.ts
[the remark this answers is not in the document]
dake3601: The views map holds the children of the TabsItem by their key. This views is held on the context so it can be accessed and rendered under Tabs.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export interface TabsContextData {
+  /*
+   ** The currently selected TabsItem's key
+   */
+  selectedKey: string | null;
+
+  /*
+   ** Index of currently selected key
+   */
```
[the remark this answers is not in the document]
dake3601: We will be updating the comment. The expected use is as a callback to customize how IDs are generated for each tab header. Useful if you're rendering content outside and need to connect aria-labelledby. Here is an example of the callback:

```const tabsRenderSeparately: React.FunctionComponent = () => {
  const [selectedKey, setSelectedKey] = React.useState('rectangleRed');

  const onTabsClick = (key: string) => {
    setSelectedKey(key);
  };

  const getTabId = (key: string) => {
    return `ShapeColorPivot_${key}`;
  };

  return (
    <View style={stackStyle}>
      <View
        accessible={true}
        focusable={true}
        accessibilityLabel={getTabId(selectedKey)}
        style={{
          width: 100,
          height: selectedKey === 'squareRed' ? 100 : 200,
          backgroundColor: selectedKey === 'rectangleGreen' ? 'green' : 'red',
        }}
      />
      <Tabs label="Tabs" onTabsClick={onTabsClick} getTabId={getTabId} headersOnly={true} selectedKey={selectedKey}>
        <TabsItem headerText="Rectangle Red" itemKey="rectangleRed" />
        <TabsItem headerText="Square Red" itemKey="squareRed" />
        <TabsItem headerText="Rectangle Green" itemKey="rectangleGreen" />
      </Tabs>
    </View>
  );
};```

PR #837, opened 2021-07-30
Fix icon depending on theme being available

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Icon currently depends on theme provider being used, which is not a strict requirement. Add defaultThemeFluent to calls of useTheme() in case theme is not available.

This should be replaced with the useFluentTheme hook in the future, but we should unblock people first while that's being sorted out.

### Verification

Verified in tester that issue is fixed.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-icon-cbd8fdac-3f04-446e-8b8f-033d961d88c8.json
+++ b/change/@fluentui-react-native-icon-cbd8fdac-3f04-446e-8b8f-033d961d88c8.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Quickly fix icon bug",
+  "packageName": "@fluentui-react-native/icon",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):

Saadnajmi:
Wasn't aware of this bug. Thank you for the quick PR! 

Question: Should it be `createDefaultTheme` instead of `defaultFluentTheme`? The former will attach an appearance change listener. I'm not sure if the typing will work if you do that though since it's a `Theme` vs a `ThemeReference`

rurikoaraki:
@Saadnajmi Yeah no worries! This was reported by a partner team.

This is following the other calls to useTheme() in the repo, I think the idea is that there's a fallback if there's no theme provided, so it's okay that it doesn't react to appearance changes.

PR #838, opened 2021-08-02
Add native support for radio button on macOS

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR adds a native module for the radio button component. This achieves the native look and feel and functionality of FocusZone via the JS RadioGroup control. 

Open Issues:
1.  Tabbing into focus zone doesn't focus the currently selected button.
2.  When VO cursor is synchronized with keyboard focus, changing the VO cursor should only update keyboard focus without changing button states when entering FocusZone.

### Verification

https://user-images.githubusercontent.com/67026167/127902922-7bde9665-d217-427b-9d58-f79c5a652848.mov

https://user-images.githubusercontent.com/67026167/127902938-1063a983-a877-4156-b960-45ae2fe3f4f9.mov

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [X] Voiceover


packages/components/NativeRadioButton/macos/RadioButtonViewManager.m
```diff
@@ -0,0 +1,16 @@
+#import <React/RCTComponent.h>
+#import <React/RCTViewManager.h>
+
+@interface RCT_EXTERN_MODULE(RadioButtonViewManager, RCTViewManager)
+
+RCT_EXPORT_VIEW_PROPERTY(title, NSString);
+RCT_EXPORT_VIEW_PROPERTY(enabled, BOOL);
+RCT_REMAP_VIEW_PROPERTY(buttonKey, keyEquivalent, NSString)
+RCT_CUSTOM_VIEW_PROPERTY(state, BOOL, NSButton) {
+	if ([[RCTConvert NSNumber:json] isEqualToNumber:[NSNumber numberWithBool:YES]]) {
```
harrieshin, line 10: we don't need to  set the value off cases?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+@objc(RadioButtonViewManager)
+class RadioButtonViewManager: RCTViewManager {
+
+	override func view()->NSView! {
+		let radioButton = RadioButton.init(radioButtonWithTitle: "", target: nil, action: nil)
+		radioButton.action = #selector(radioButton.callOnPress)
```
harrieshin, line 8: we can't combine this directly in line 7?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.m
chiuam: Button state is off by default, so no need to validate that case, unless I'm missing something?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
chiuam: I think radioButton needs to be initialized before I can access it as an object.

harrieshin: am I misunderstanding something? it can't be target self action #selector(callonpress) ?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.m
harrieshin: if the state changes and caller asks for state shouldn't it be getting the most up to date value?

chiuam: I think NSButton does that under the hood? Not sure if I'm addressing your question but the state prop allows the client to set the button's initial state, it's not meant to be read on the JS side.

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
chiuam: Target must be a NSButton object and `self` here is RCTViewManger so I had to create an instance of RadioButton and set itself as the target. `callOnPress` belongs to the RadioButton class, so it needs to be accessed via a RadioButton object. 

I couldn't get init radioButtonWithTitle working in RadioButton so I came up with this workaround instead.

packages/components/NativeRadioButton/macos/RadioButton.swift
```diff
@@ -0,0 +1,21 @@
+class RadioButton: NSButton {
+
+	@objc public var onPress:RCTBubblingEventBlock?
+	
+	@objc public func callOnPress() {
+		self.window?.makeFirstResponder(self)
+		if (onPress != nil) {
+			onPress!(nil)
+		}
+	}
```
Saadnajmi, line 10: nit, can we prefix this with `send` instead of `call` to match NativeButton? It's what I was planning to do in MenuButton too.

packages/components/NativeRadioButton/macos/RadioButton.swift
```diff
@@ -0,0 +1,21 @@
...
+	
+	/// Set button state to 'on' when it becomes focused
+	override func becomeFirstResponder() -> Bool {
+		/// fire onChange event
```
Saadnajmi, line 14: nit, // not /// since it's inline comments rather than documentation style comments

packages/components/NativeRadioButton/macos/RadioButton.swift
```diff
@@ -0,0 +1,21 @@
...
+		if (state != NSButton.StateValue.on) {
+			callOnPress()
+		}
```
Saadnajmi, line 17: should we override the state property and call `callOnPress()` inside didSet ?

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  state?: boolean;
+  /**
+   * Unique identifier for each radio button.
+   */
+  buttonKey?: string;
+  /**
+   * On press event block
+   */
+  onPress?: () => void;
+};
```
Saadnajmi, line 34: Nit: Could we change these to be closer to `RadioButton's` properties? AKA, `content` instead of `title`, and `disabled` instead of `enabled`? The idea being to keep the API as similar as possible to the original RadioButton to make it easier for clients?

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+      minHeight: 20,
+    },
+    nativeRadioButtonName,
+  ],
+  slots: { root: NativeRadioButtonView },
+  slotProps: {
+    root: buildProps((tokens) => ({
+      style: {
+        marginLeft: 3,
+        marginTop: 3,
```
Saadnajmi, line 60: why 3?

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        width: tokens.minWidth,
+        height: tokens.minHeight,
```
Saadnajmi, line 62: style has a `minWidth` and `minHeight` key. I think those are better suited for these tokens.

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+@objc(RadioButtonViewManager)
+class RadioButtonViewManager: RCTViewManager {
```
Saadnajmi, line 4: I think we should use a 3 letter prefix here https://microsoft.github.io/objc-guide/Headers/Naming.html

I used MSF in my PR for MenuButton, but there's a comment asking If that's the right one.

packages/components/NativeRadioButton/macos/RadioButton.swift
chiuam: I agree, will rename it to sendCallBack() to match NativeButton.

packages/components/NativeRadioButton/macos/RadioButton.swift
chiuam: Probably not because the callback function shouldn't get triggered when the default state is On.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: I know we talked a bit about this offline, my rationale was since this API is tailored to native platforms, it's appropriate to follow the native naming conventions.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: A number I randomly picked to make it look aesthetically pleasing to my taste... not from any design spec whatsoever.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: But `width` and `height` must be defined to get the native views showing in the app?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
chiuam: I removed the prefix because it's not a Fluent control, though I agree with you that we should use a 3 letter prefix.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chripear: nit: Maybe then add a brief comment for the benefit of others with the same question, e.g. "These values are chosen for aesthetics" or "to match the look of native buttons"

packages/components/NativeRadioButton/FluentUIReactNativeRadioButton.podspec
```diff
@@ -0,0 +1,21 @@
...
+  s.homepage     = "https://github.com/microsoft/fluentui-react-native"
+
+  s.source           = { :git => "https://github.com/microsoft/fluentui-react-native.git", :tag => "#{s.version}" }
+  s.swift_version    = "5"
+
+  s.osx.deployment_target = "10.14"
+  s.osx.source_files      = "macos/*.{swift,h,m}"
+
+  s.dependency 'React'
+end
```
chripear, line 21: nit: line break

chripear (approved):

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
```diff
@@ -47,20 +48,47 @@ const basicRadioGroup: React.FunctionComponent = () => {
...
+        <NativeRadioButton title="Option C (disabled)" enabled={false} style={{ width: 150 }} buttonKey="C" />
+        <NativeRadioButton title="Option D" buttonKey="D" />
+      </RadioGroup>
+      <Separator />
+      <RadioGroup label="RadioGroup 2">
+        <NativeRadioButton title="Option A" buttonKey="A" />
+        <NativeRadioButton title="Option B" state={true} onPress={() => alert('Option B pressed')} buttonKey="B" />
+        <NativeRadioButton title="Option C (disabled)" enabled={false} style={{ width: 150 }} buttonKey="C" />
+        <NativeRadioButton title="Option D" buttonKey="D" />
+      </RadioGroup>
```
Saadnajmi, line 66: I think we should rename the `state` prop to somethingg like `isSelected`. state feels very misleading since that's something very special in React terminology.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
Saadnajmi: We talked offline more. We thought about merging the changes here into the existing `@fluentui-react-native/radiobutton` package so that it's a JS component on win32 and a native component on macOS. There are a few issues with that approach, namely that some non-optional props from the JS control are not implemented in the macOS native component yet, and other props would be no-ops. We weren't sure what's a better experience for clients: Using the same package but having some props not work on macOS vs win32, or having two different packages altogether. 

The conclusion we came to was to try both, and collect client feedback. My `MenuButton` component shares the package with win32 and has many props as non-optional. Amanda's `NativeRadioButton` component is a separate package with different props altogether. Both are about to be consumed by different clients so we can see the pros and cons of each approach.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
Saadnajmi: I think you can just set minHeight/minWidth and it'll display too. You might have an issue where the control grows too big and takes up more space then you wanted it to (I did). Id that's the case, then yeah I guess we can just use width/Height.

An aside, I was wondering why out native modules need a hardcoded size and I came across this StackOverflow. The solution to get our components to use something akin to `intrinsicContentSize` might be to implement a shadow view for each component https://stackoverflow.com/a/59746747 Definitely out of scope and something I wanna investigate myself later

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
harrieshin: technically fluent snaps to 4 pt grid

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
chiuam: Talked offline, seems like you'll follow up on this issue https://office.visualstudio.com/APEX/_workitems/edit/5319675

apps/fluent-tester/src/FluentTester/TestComponents/RadioGroup/RadioGroupTest.tsx
chiuam: fixed in commit fb736cb070687bb15f4028a3bafafaeccc538eec

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: Thanks for the write-up

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: minHeight & minWidth worked, thanks for the suggestion.

packages/components/NativeRadioButton/macos/RadioButtonViewManager.m
Saadnajmi: I still feel like we should expicitly set the OFF value, in case someone later tries to set this (outside of just as a default value) and is very confused why it won't.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const nativeRadioButtonName = 'NativeRadioButton';
+
+const NativeRadioButtonView = ensureNativeComponent('RadioButtonView');
+
+export type NativeRadioButtonProps = {
+  /**
+   * Radio button title, default title is set to "Button".
+   */
+  title?: string;
```
Saadnajmi, line 15: It's actually an empty string, no?

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Whether the radio button is selectable or not, true by default.
+   */
+  enabled?: boolean;
+  /**
+   * Whether the radio button is selected or not, false by default.
+   */
+  isSelected?: boolean;
```
Saadnajmi, line 23: maybe change `isSelected` to just `selected` to match the case of `enabled`?

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+      minHeight: 20,
+    },
+    nativeRadioButtonName,
+  ],
+  slots: { root: NativeRadioButtonView },
+  slotProps: {
+    root: buildProps((tokens) => ({
+      style: {
+        marginLeft: 4, // Fluent design on macOS
+        marginTop: 4, // Fluent design on macOS
```
Saadnajmi, line 58: I don't feel this comment is very descriptive to why this is "4" specifically. Maybe "Fluent controls are designed to  snap to a 4 px grid"?

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
```diff
@@ -0,0 +1,16 @@
...
+
+	override func view()->NSView! {
+		let radioButton = RadioButton.init(radioButtonWithTitle: "", target: nil, action: nil)
+		radioButton.action = #selector(radioButton.sendCallback)
+		radioButton.target = radioButton
+		return radioButton
+	}
```
Saadnajmi, line 11: If you override init in your subclass (I think you have to override the designated initializer), you could move all this logic to inside RadioButton.swift, and just do 
`return RadioButton()`. You can see me do a similar thing in MenuButton.swift.

package.nuspec
```diff
@@ -31,11 +31,14 @@
...
     <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-FocusZone.a" target="Debug-macosx"/>
     <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-MenuButton.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-Radio-Button\libFluentUI-React-Native-Radio-Button.a" target="Debug-macosx"/>
     <!-- macOS ship -->
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-FocusZone.a" target="Ship-macosx"/>
     <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-MenuButton.a" target="Ship-macosx"/>
+    <file src="DerivedData\Build\Products\Release\FluentUI-React-Native-Radio-Button\libFluentUI-React-Native-Radio-Button.a" target="Ship-macosx"/>
```
Saadnajmi, line 40: nit: Maybe `RadioButton` instead of `Radio-Button` to match MenuButton and FocusZone not having a hyphen?

Saadnajmi: Actually ignore this. MenuButton and FocusZone should have a hyphen to match their nom package names

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
```diff
@@ -0,0 +1,68 @@
+/** @jsx withSlots */
+import { ViewProps } from 'react-native';
+import { compose, buildProps, mergeProps, UseSlots, LayoutTokens } from '@fluentui-react-native/framework';
+import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
+import { withSlots } from '@uifabricshared/foundation-composable';
```
Saadnajmi, line 5: I think you can import this from the new framework and not take a dependency on the deprecated one

Saadnajmi: Yeah, like we do in Shimmer.tsx
```
import { ..., withSlots, ... } from '@fluentui-react-native/framework';
```

chiuam: I'm not sure why it can't find `withSlots` from @fluentui-react-native/framework, that only happens in non-experimental components.

packages/components/NativeRadioButton/macos/RadioButtonViewManager.swift
chiuam: Yeah I tried that and it didn't work. The problem is init.radioButtonWithTitle is a convenience initializer, I don't think swift allows a call to a convenience initializer inside a designated initializer.

packages/components/NativeRadioButton/src/NativeRadioButton.tsx
chiuam: That's right, good catch

package.nuspec
chiuam: Would having different names cause any trouble in publishing the NPM package?

Saadnajmi: I didn't mean to change the npm package name, just the static library name. But when I looked into it more, the nom package names all had hyphens and those are the "sources of truth" to me so I feel you did it right.

Saadnajmi (approved):

package.nuspec
```diff
@@ -31,11 +31,14 @@
     <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Debug-macosx"/>
     <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-FocusZone.a" target="Debug-macosx"/>
     <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-MenuButton.a" target="Debug-macosx"/>
```
Saadnajmi, line 33: There is a typo I made here in both debug and ship. If you can fix it that would be great lol

samuelfreiberg:
I didn't see this PR until now... But instead of making a separate component folder for this (NativeRadioButton), why don't we put it in the RadioGroup folder? It doesn't seem best from an organizational standpoint to make new folder for each platform control. 

IE. Components -> Win32RadioGroup
                           -> MacOSRadioGroup
                           -> AndroidRadioGroup

Should be:

Components -> RadioGroup -> .win32
                                               -> .macOS
                                               -> .android
                         

Thoughts? @Saadnajmi @chiuam

Saadnajmi:
> I didn't see this PR until now... But instead of making a separate component folder for this (NativeRadioButton), why don't we put it in the RadioGroup folder? It doesn't seem best from an organizational standpoint to make new folder for each platform control.
> 
> Foe example,
> 
> ```
>  \Components\     -> Win32RadioGroup
>                   -> MacOSRadioGroup
>                   -> AndroidRadioGroup
> ```
> 
> Should be:
> 
> Components -> RadioGroup -> .win32
> -> .macOS
> -> .android
> 
> Thoughts? @Saadnajmi @chiuam

We actually just talked about implementation and we're going to aim to remove this extra package altogether, and share the package/API with the existing RadioGroup =)

samuelfreiberg:
@Saadnajmi Perfect :)

PR #840, opened 2021-08-03
Focus zone empty tabbable

nakambo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There's code in Focus Zone written in a way such that once a default tabbable element is set, it can never be reset, no matter if the user of the focus zone passes a `undefined` defaultTabbableElement, or not pass it at all. Let's tweak that a little so that not-passing\resetting the defaultTabbableElement has the same effect as never having it set in the first place.

### Verification

Manual testing:

- Added action in the FocusZoneTest component to test the effect of setting any of the nine buttons as the default tabbable element plus another to clear it and checked that clearing the default tabbable makes the focus zone behave as expected (as if one was never set)
- Checked this sequence: set -> clear -> set -> clear

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/FocusZoneTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/FocusZoneTest.tsx
@@ -80,29 +80,37 @@ const commonUsageFocusZone: React.FunctionComponent = () => {
 };
 
 const navigation2DFocusZone: React.FunctionComponent = () => {
-  const buttonRef = React.useRef<View>(null);
+  const buttonRefs = [];
+  const onClicks = [];
+  const [defaultTabbableElementIndex, setDefaultTabbableElementIndex] = React.useState<number>(5);
+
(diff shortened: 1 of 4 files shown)


samuelfreiberg:
Hey Navneet, is this change in reference to a specific bug or task? What scenario are you trying to fix with this?

nakambo:
> Hey Navneet, is this change in reference to a specific bug or task? What scenario are you trying to fix with this?

yes -- I was trying to fix something in the Editor pane, and I noticed that if I set a defaultTabbableElement on a focus zone, and on a subsequent re-render of said FocusZone *don't* pass it, the FZ still thinks that the defaultTabbableElement is set. I confirmed this in NetUI spy that the defaultTabbableElement is indeed never cleared.

There's no bug\task filed for this.

samuelfreiberg (approved):

PR #841, opened 2021-08-03
Fix usage of conditional hooks

nakambo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There are various places in code where we do something like the following, which probably intends to read as grab the `componentRef` from `userProps` if present, otherwise make a new one:

`const { componentRef = React.useRef(null), ... rest } = userProps`

However, this is problematic if during one render pass a given component does not receive `componentRef` (so we make a ref) and in the next render pass it does (so we never try to use that ref we made earlier). The reason this is specifically problematic is that the `React.useRef(null)` **will not be called** if we were successful in extracting `componentRef` from `props`.

The implication here is that we invalidate a very important [rule of hooks](https://fb.me/rules-of-hooks/) -- that they must always be called in the same order. This will in turn lead to unexpected error messages in downstream\client code that uses Fluent components like the below:

> Warning: React has detected a change in the order of Hooks called by Button. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://fb.me/rules-of-hooks
> 
>    Previous render            Next render
>    ------------------------------------------------------
> 1. useMemo                    useMemo
> 2. useRef                     useCallback
>    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

To fix, let's just make sure we always call `React.useRef(null)` -- we can still toss it aside if not wanted by slightly tweaking how the code is structured.

### Verification

Testing:

* Sanity tested button, checkbox, and radio button components in the Win32 fluent tester
* I didn't attempt to repro the problem in the Fluent tester, but did validate the fix in the product where I originally discovered this problem.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-button-3af61ccd-ab0e-4c6b-89cc-4df652833ff7.json
+++ b/change/@fluentui-react-native-button-3af61ccd-ab0e-4c6b-89cc-4df652833ff7.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix usage of conditional hooks",
+  "packageName": "@fluentui-react-native/button",
+  "email": "nakambo@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@fluentui-react-native-checkbox-3e34480f-33f8-420e-a249-fa0a2088b16d.json
+++ b/change/@fluentui-react-native-checkbox-3e34480f-33f8-420e-a249-fa0a2088b16d.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix usage of conditional hooks",
+  "packageName": "@fluentui-react-native/checkbox",
+  "email": "nakambo@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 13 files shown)


Saadnajmi:
I wonder if we can add the linter rule mentioned in the rules of hooks page to our repo to catch stuff like this in the future
https://www.npmjs.com/package/eslint-plugin-react-hooks

Saadnajmi (approved):
Approved, but you should probably wait for one of the CXE people to also approve

samuelfreiberg:
This looks good - I just want to double-check, have you gone through the test app and played with some of these components a bit to see if everything works as intended? Like RadioGroup, for example

nakambo:
> This looks good - I just want to double-check, have you gone through the test app and played with some of these components a bit to see if everything works as intended? Like RadioGroup, for example

Yes -- I played with these controls some -- `Sanity tested button, checkbox, and radio button components in the Win32 fluent tester`

Let me know if there are specific steps you'd like for me to run through.

samuelfreiberg (approved):

PR #842, opened 2021-08-03
Create useFluentTheme hook

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is split out from the experimental-text PR and creates a useFluentTheme helper which will try to obtain the theme from the context and will fall back to the default fluent theme.

This is important as people start building components that query the theme directly to ensure we standardize. The Icon component is an example of this and so part of this PR switches the usage in there.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Icon/src/Icon.tsx
+++ b/packages/experimental/Icon/src/Icon.tsx
@@ -2,12 +2,10 @@ import * as React from 'react';
 import { IconProps, SvgIconProps, FontIconProps } from './Icon.types';
 import { Image, ImageStyle, Platform, View } from 'react-native';
 import { Text } from '@fluentui-react-native/text';
-import { mergeStyles } from '@fluentui-react-native/framework';
+import { mergeStyles, useFluentTheme } from '@fluentui-react-native/framework';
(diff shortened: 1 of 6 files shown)


PPatBoyd (approved):

Saadnajmi (approved):
Approved with the caveat that we should perhaps add a note/comment/readme saying this doesn't support the apple/android theme and that's a followup for us (me) to do.

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -84,7 +82,7 @@ function renderSvg(iconProps: IconProps) {
 }
 
 export const Icon = stagedComponent((props: IconProps) => {
-  const theme = useTheme() || defaultFluentTheme;
+  const theme = useFluentTheme();
```
rurikoaraki, line 85: Thanks for fixing these!

rurikoaraki (approved):

PR #845, opened 2021-08-04
Add ability to update brand tokens

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added theme recipe to win32 theme to use ramp information to populate brand-related alias tokens. Entries are based on web tokens (so there will be a follow up to edit/replace this with the mapping from palette to tokens for win32 that I got earlier this week). Long term branded information will come from pipeline, but in the short term we can draw this information from the native module.

Also updated the tester to regenerate the office theme when the theme is changed via picker and update the brand-related information to include the new colors and entries (some entries are missing, so will have the wrong color in the tester - will follow up with separate PR).

Long term we should probably come up with a way to draw that information from either the pipeline or the host somehow and not hardcode it into the tester since that makes assumptions on how the themes are created and which entries are brand-dependent, but for now this is fine.

### Verification

Test app reflects host colors accurately

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
+++ b/apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
@@ -1,5 +1,21 @@
 export type OfficeBrand = 'Default' | 'Office' | 'Word' | 'Excel' | 'Powerpoint' | 'Outlook';
-type BrandRampKey = 'App1' | 'App2' | 'App3' | 'App4' | 'App5' | 'App6' | 'App7' | 'App8';
+type BrandRampKey =
+  | 'App1'
+  | 'App2'
+  | 'App3'
+  | 'App4'
+  | 'App5'
+  | 'App6'
+  | 'App7'
+  | 'App8'
+  | 'AppTint40'
+  | 'AppTint30'
+  | 'AppTint20'
+  | 'AppTint10'
+  | 'AppPrimary'
+  | 'AppShade10'
+  | 'AppShade20'
+  | 'AppShade30';
 type BrandRamps = { [K in OfficeBrand]: { [J in BrandRampKey]: string } };
 
 const brandColors: BrandRamps = {
@@ -19,6 +35,14 @@ const brandColors: BrandRamps = {
     App6: '#d83b01',
     App7: '#a22c01',
     App8: '#6c1e01',
+    AppTint40: '#f9dcd1',
+    AppTint30: '#f4beaa',
+    AppTint20: '#e8825d',
+    AppTint10: '#dd4f1b',
+    AppPrimary: '#d83b01',
+    AppShade10: '#c33400',
+    AppShade20: '#a52c00',
+    AppShade30: '#792000',
   },
   Word: {
     App1: '#E3ECFA',
@@ -29,6 +53,14 @@ const brandColors: BrandRamps = {
     App6: '#2B579A',
     App7: '#124078',
     App8: '#002050',
+    AppTint40: '#D2E0F4',
+    AppTint30: '#AEC6EB',
+    AppTint20: '#6794D7',
+    AppTint10: '#2E6AC5',
+    AppPrimary: '#185ABD',
+    AppShade10: '#1651AA',
+    AppShade20: '#13458F',
+    AppShade30: '#0E336A',
   },
   Excel: {
     App1: '#E9F5EE',
@@ -39,6 +71,14 @@ const brandColors: BrandRamps = {
     App6: '#217346',
     App7: '#0E5C2F',
     App8: '#004B1C',
+    AppTint40: '#CAEAD8',
+    AppTint30: '#A0D8B9',
+    AppTint20: '#55B17E',
+    AppTint10: '#218D51',
+    AppPrimary: '#107C41',
+    AppShade10: '#0F703B',
+    AppShade20: '#0C5F32',
+    AppShade30: '#094624',
   },
   Powerpoint: {
     App1: '#FCF0ED',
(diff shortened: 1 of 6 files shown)


apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -1,5 +1,21 @@
...
-type BrandRampKey = 'App1' | 'App2' | 'App3' | 'App4' | 'App5' | 'App6' | 'App7' | 'App8';
+type BrandRampKey =
+  | 'App1'
+  | 'App2'
+  | 'App3'
+  | 'App4'
+  | 'App5'
+  | 'App6'
+  | 'App7'
+  | 'App8'
```
Saadnajmi, line 10: I had assumed these were the same as Primary/Tint/Shade that you had below (just with more confusing names). Are these values used somewhere?

Saadnajmi: OK I see that you have completely different hex values. What does App1-8 semantically mean?

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
```diff
@@ -50,15 +50,14 @@ function applyCaterpillarTheme(parent: Theme): PartialTheme {
...
-  : {};
 
 /** apply the currently active theme layering */
 export function applyTheme(parent: Theme, name: ThemeNames): PartialTheme {
   switch (name) {
     case 'Office':
-      return officeBase;
+      return themingModule
+      ? createOfficeTheme({ paletteName: 'WhiteColors' }).theme
+      : {};
```
Saadnajmi, line 60: Why is this no longer a partial theme?

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
rurikoaraki: So App1-8 are the old brand color names from Office. They're being replaced by the AppTint-AppShade colors.

Since the older entries are based on the old App colors, I elected to just add them instead of ripping it out

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
rurikoaraki: The partial was just running the base createPartialOfficeTheme part of making a win32 theme. Since I added to the theme recipes, the theme being spit out here wasn't reflecting the "true" result you would get from accessing the theme on win32, so I thought it'd be a good idea to replace this, since we use the tester to show what result you'd get if you just ran the defaults.

Saadnajmi: If the default theme is the new default (lol wording), is there any use of the Office theme? Or is it just an example theme that is only present in our test app?

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -19,6 +35,14 @@ const brandColors: BrandRamps = {
...
     App7: '#a22c01',
     App8: '#6c1e01',
+    AppTint40: '#f9dcd1',
+    AppTint30: '#f4beaa',
+    AppTint20: '#e8825d',
+    AppTint10: '#dd4f1b',
+    AppPrimary: '#d83b01',
+    AppShade10: '#c33400',
+    AppShade20: '#a52c00',
+    AppShade30: '#792000',
```
Saadnajmi, line 45: For my own knowledge, is this that new unified 13 level palette that designers were talking about between light and dark?

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -83,6 +139,29 @@ export const applyBrand = (currentBrand: string) => {
...
+          brandBackgroundHover: ramp.AppShade10,
+          brandBackgroundPressed: ramp.AppShade30,
+          compoundBrandBackground1: ramp.AppPrimary,
+          compoundBrandBackground1Hover: ramp.AppShade10,
+          compoundBrandBackground1Pressed: ramp.AppShade20,
+          brandStroke1: ramp.AppPrimary,
+          brandStroke2: ramp.AppTint40,
+          compoundBrandStroke1: ramp.AppPrimary,
+          compoundBrandStroke1Hover: ramp.AppShade10,
+          compoundBrandStroke1Pressed: ramp.AppShade20,
```
Saadnajmi, line 164: For my own knowledge, Both the left hand side and the right hadn't side are technically alias tokens, yes?

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
rurikoaraki: No, unfortunately that only exists for the web palettes. This is an 8 level palette that I believe is in line with what's currently used in win32. I do want to try and get better alignment between platforms here long term.

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
rurikoaraki: Yeah, I think technically that's true. The right side will change value based on which app you're using, so it would be considered an alias token.

apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
rurikoaraki: The Office theme is used for win32 still. The default theme is the base, but the Office theme allows for integration with palettes, which can be useful for people on the win32 platform.

PR #849, opened 2021-08-04
Clean up output from token pipeline

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Token pipeline was spitting out some duplicates so changed the script to delete unnecessary files. Global token files should be the same across modes, and we're not using the control token files. Also added a global token folder so that global token is taken from one source that's static.

### Verification

Code builds. The outputs should be the same as before.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


yarn.lock
```diff
@@ -1181,21 +1181,6 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@fluentui-react-native/framework@0.5.4":
```
rurikoaraki, line 1184: This is a lockfile change that is a result of a different change, I think. Pretty sure it's already in the repo

Saadnajmi: I've seen this hop back and forth in different reviews. It might be worth double checking it does exist. If it doesn't I'm sure the next person who runs `yarn` will add it back.

Saadnajmi: Wait.. should this even be here given that this is a local package?

Saadnajmi (approved):

yarn.lock
rurikoaraki: uh... great question? Anyway I just merged master and it didn't come back, so I assume this is just gone

PR #855, opened 2021-08-05
Revert "Bump lage from 0.31.0 to 0.32.0"

Saadnajmi:
Reverts microsoft/fluentui-react-native#839

This broke stuff. Let's undo while we investigate if it's our bug or a bug on lage.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2107,6 +2107,14 @@
   dependencies:
     "@microsoft/node-core-library" "3.19.3"
 
+"@microsoft/task-scheduler@^2.7.1":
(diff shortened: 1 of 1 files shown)


rurikoaraki:
Might want to tell the package manager that we cap at a certain version so it doesn't bother us about upgrading

rurikoaraki (approved):

Saadnajmi:
@rurikoaraki or keep it so that when it does bother us it reminds us to figure out what's broken? :D

rurikoaraki:
@Saadnajmi lol fair enough

PR #862, opened 2021-08-06
Win32 Shimmer and commonJS Shimmer unifying/cleaning changes

PPatBoyd:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Create a win32 Shimmer control to represent loading UI whose appearance reflects the UI to be loaded.

https://user-images.githubusercontent.com/41588892/128555856-6e0d750a-3f74-497d-b683-f64543a63e5a.mp4

### Verification

Manually verified between myself and @chiuam

### Pull request checklist

- [ ] Automated Tests
- [] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Ooooh looks very nice!

apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -12,10 +12,8 @@ import { useTheme } from '@fluentui-react-native/theme-types';
 import { ThemePickers } from './theme/ThemePickers';
 
 // uncomment the below lines to enable message spy
-/*
```
rurikoaraki, line 15: Is this intended to be committed?

apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/consts.ts
```diff
@@ -1,2 +1,5 @@
 export const HOMEPAGE_SHIMMER_BUTTON = 'Homepage_Shimmer_Button';
 export const SHIMMER_TESTPAGE = 'Shimmer_TestPage';
+
+// import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
```
rurikoaraki, line 4: nit: Please remove commented code

apps/fluent-tester/src/FluentTester/FluentTester.tsx
PPatBoyd: glossed over it reviewing my changes but tbh I'm somewhat tempted to leave it in -- I'm always uncommenting it for debug purposes and keeping it there as a local change is mildly frustrating.. but there are other ways to accommodate that I should look into -- reverting

packages/experimental/Shimmer/src/Shimmer.styling.win32.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Absolute positioning is used to overlay the clipping mask on top of the shimmer wave.
+     */
+    clippingMask: buildProps(
+      (tokens: ShimmerTokens) => ({
+        backgroundColor: tokens.backgroundColor,
+      }),
+      ['backgroundColor'],
+    ),
```
Saadnajmi, line 55: For my own knowledge, these extra slots were needed because win32 doesn't SVG masks, correct? Or are they something that could be applicable to the cross-plat shimmer should we choose to refactor it in the future?

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { ClipPath, Defs, LinearGradient, Path, Rect, Stop, Svg } from 'react-native-svg';
+import { styleSettings } from './Shimmer.styling.win32';
+import { shimmerName, ShimmerProps } from './Shimmer.types';
+import { ShimmerType } from './Shimmer.types.win32';
+import { RCTNativeAnimatedShimmer } from './consts.win32';
+import { convertRectToSvgPath, convertCircleToSvgPath } from './SvgShapeToPath';
+import { withSlots } from '@fluentui-react-native/framework';
+
+/** Absolute positioning is used to overlay the clipping mask on top of the shimmer wave. */
+const clippingMask: React.FunctionComponent<ShimmerType['slotProps']['clippingMask']> = (
```
Saadnajmi, line 13: Wait, why couldn't this just be ShimmerType.slotProps.clippingMask? Same for wave below

Saadnajmi: Or maybe define ShimmerType.slotProps.clippingMask as a separate type that both use?

packages/experimental/Shimmer/src/Shimmer.styling.ts
```diff
@@ -5,13 +5,10 @@ import { shimmerName, ShimmerProps, ShimmerSlotProps, ShimmerTokens } from './Sh
...
   tokens: [
     () => ({
-      toValue: 30,
       duration: 7000,
       delay: 0,
-      gradientTintColor: 'white', // Baseline component is the light mode color
-      shimmerTintColor: '#E1E1E1', // Baseline component is the light mode color
-      width: 200,
-      height: 100,
+      shimmerWaveColor: 'white', // Baseline component is the light mode color
```
rurikoaraki, line 10: If we have these colors in the global tokens it might be good to convert to pull from those so we have less hardcoded colors in our repo.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -39,79 +37,63 @@ export const Shimmer = compose<ShimmerType>({
 
     const memoizedShimmerData = useMemo(
       () => ({
-        gradientTintColor: props.gradientTintColor ? props.gradientTintColor : tokens['gradientTintColor'],
-        shimmerTintColor: props.shimmerTintColor ? props.shimmerTintColor : tokens['shimmerTintColor'],
-        containerWidth: props.width ? props.width : tokens['width'],
-        containerHeight: props.height ? props.height : tokens['height'],
-        toValue: props.toValue ? props.toValue : tokens['toValue'],
+        shimmerWaveColor: props.shimmerWaveColor ? props.shimmerWaveColor : tokens['shimmerWaveColor'],
+        shimmerColor: props.shimmerColor ? props.shimmerColor : tokens['shimmerColor'],
```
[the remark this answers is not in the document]
PPatBoyd: yeah, clarity

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -39,79 +37,63 @@ export const Shimmer = compose<ShimmerType>({
...
+        containerWidth: props?.style['width'] ? props?.style['width'] : '100%',
+        containerHeight: props?.style['height'] ? props?.style['height'] : '100%',
```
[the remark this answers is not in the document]
PPatBoyd: width and height were removed as tokens

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -39,79 +37,63 @@ export const Shimmer = compose<ShimmerType>({
...
-                  y={element.yPos}
-                  rx={element.borderRadius}
-                  ry={element.borderRadius}
-                />,
-              );
-            } else if (element.type == 'circle') {
-              rows.push(<Circle key={i} r={element.height / 2} cx={element.xPos} cy={element.yPos} />);
-            } else {
-              assertNever(element.type);
-            }
```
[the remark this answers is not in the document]
PPatBoyd: the prop was optional before, made it required

packages/experimental/Shimmer/src/Shimmer.win32.tsx
PPatBoyd: `Cannot access 'ShimmerType.slotProps' because 'ShimmerType' is a type, but not a namespace. ` tbh this confused me too.  I'll use the slot prop types directly, I didn't really like depending on it being "coordinated"

apps/fluent-tester/src/FluentTester/testPages.win32.ts
```diff
@@ -6,20 +7,26 @@ import { ContextualMenuTest, HOMEPAGE_CONTEXTUALMENU_BUTTON } from './TestCompon
...
 import { TestDescription } from './TestComponents';
-import { MenuButtonTest, HOMEPAGE_MENU_BUTTON } from './TestComponents/MenuButton';
-import { HOMEPAGE_TOKEN_BUTTON, TokenTest } from './TestComponents/Tokens';
 
 export const tests: TestDescription[] = [
+  {
+    name: 'ActivityIndicator Test',
+    component: ActivityIndicatorTest,
+    testPage: HOMEPAGE_ACTIVITYINDICATOR_BUTTON,
+  },
```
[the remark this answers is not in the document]
PPatBoyd: I had added it because I wanted to check on the commonJS animation behavior; I'll repeal -- though checking on it I also realized it's using transform and the type isn't always a string, wish I hadn't dropped those native changes... zz.

packages/experimental/Shimmer/src/Shimmer.tsx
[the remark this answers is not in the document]
rurikoaraki: For context, `assertNever` enforces that each entry of an enum of union type is handled by a conditional statement or a switch/case. It'll cause a build error (or throw at runtime) if you add an entry but neglect to handle it in the code. So for example in this case if we ever add another ShimmerElementType, then this code will cause a build error if it's not updated to handle the new type.

So I would argue that there's still value in having it here

packages/experimental/Shimmer/src/Shimmer.styling.ts
PPatBoyd: yeah I need to check in w/ lineage on their design figma; these may be existing tokens (e.g. I use bodyDivider for one) or need to be new

packages/experimental/Shimmer/src/Shimmer.styling.win32.ts
PPatBoyd: 2 SVGs are needed for the translation strategy, so it'll come xplat

apps/fluent-tester/src/FluentTester/TestComponents/Svg/SvgTest.tsx
```diff
@@ -13,10 +14,25 @@ const styles = StyleSheet.create({
...
+        />
+        <Text>Change Color</Text>
+      </View>
+      <Separator />
+      <Svg width="50" height="50">
+        <Rect x="10" y="10" width="20" height="20" fill={useColorA ? colorA : colorB} stroke="black" />
+      </Svg>
+    </React.Fragment>
   );
 };
```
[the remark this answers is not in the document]
PPatBoyd: yeah it actually wasn't updating render properly

packages/experimental/Shimmer/src/Shimmer.styling.ts
Saadnajmi: I think the original hex codes came from the older FluentUI Apple palette. #E1E1E1 maps to a few things from there. I agree that a followup should be to use the global token, especially if we can get that without a theme reference

packages/experimental/Shimmer/src/Shimmer.styling.win32.ts
```diff
@@ -0,0 +1,57 @@
...
+    })),
+    shimmerWaveContainer: buildProps(
+      (tokens: ShimmerTokens) => ({
+        delay: tokens.delay,
+        duration: tokens.duration,
+        shimmerColor: tokens.shimmerColor,
+      }),
+      ['delay', 'duration', 'shimmerColor'],
+    ),
+
```
Saadnajmi, line 35: nit: remove space?

Saadnajmi: Though I guess idk what to do with line 46 since it has the comment. It just feels weird that it's inconsistent.

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -64,7 +64,7 @@ const brandColors: BrandRamps = {
 
 export const brandOptions = Object.keys(brandColors).map((brand) => ({ label: brand, value: brand }));
 
-export const applyBrand = (currentBrand: string) => {
+export const applyBrand = (currentBrand: string): { [key: string]: any } => {
```
rurikoaraki, line 67: nit: this could be PartialTheme

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -1,114 +1,156 @@
...
+   * @defaultValue 12
    */
-  width?: number;
+  radius?: number;
+
   /**
-   * Specifies the shimmer view height. For circle element type, height is divided by 2 to get the radius for Svg
-   * @defaultValue '100'
+   * Note: cx and cy should be optional properties [or removed], with relative positioning being the default [or only] positioning mechanism.
+   * Note: The behavior of element intersections is undefined.
```
Saadnajmi, line 30: If we're going to assert this, I feel we should add open a GitHub issue and put the # in the comment so this note doesn't just dissappear.

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -1,114 +1,156 @@
...
+   *
+   * @defaultValue 2000
+   */
+  duration?: number;
+
   /**
-   * Specifies the gradient opacity, value should be less than 1
-   * @defaultValue '.7'
+   * Specifies the gradient opacity.
+   * @defaultValue 1
```
Saadnajmi, line 121: Is this meant to change from 0.7?

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -1,114 +1,156 @@
...
-   */
-  uri?: ImageURISource;
+export interface ShimmerProps extends ViewProps, ShimmerTokens {
   /**
-   * Array of ShimmerElement in a single view
+   * Shimmer shapes that define the masking effect of the Shimmer control.
    */
-  elements?: Array<ShimmerElement>;
+  elements?: Array<ShimmerCircleElement | ShimmerRectElement>;
 }
```
Saadnajmi, line 154: Why did we remove the support for overlaying over an image? That's one of the scenarios we support in Native apple code with the FluentUI Apple Shimmer

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,100 @@
...
+    </Svg>
+  );
+};
+
+const wave: React.FunctionComponent<ShimmerTokens> = (props: ShimmerTokens) => {
+  const shimmerColor: any = props.shimmerColor;
+  const shimmerWaveColor: any = props.shimmerWaveColor;
+
+  return (
+    <Svg {...props} width="100%" height="100%">
```
Saadnajmi, line 32: nit, filter out the props?
```
const {shimmerColor, shimmerWaveColor, ...rest} = props

return (
  <Svg {...rest} width="100%" height="100%">
```

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const waveContainer: React.FunctionComponent<ShimmerTokens> = (props: ShimmerTokens) => {
+  return <RCTNativeAnimatedShimmer {...{ ...props, style: { backgroundColor: props.shimmerColor, overflow: 'hidden' } }} />;
+};
+
+export const Shimmer = compose<ShimmerType>({
+  displayName: shimmerName,
+  ...styleSettings,
+
```
Saadnajmi, line 50: nit: I think we normally call this stylingSettings? not a big deal though, that name swaps out between the old and new framework anyway.

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,100 @@
...
+      let clipPathsAsMask: string[] = ['M -1 -1 h 100000 v 100000 h -100000 z '];
+
+      /**
+       * Now extend the path to include the provided element shapes, with their path parameter strings generated from their corresponding properties.
+       * Ideally this would be better possible with functions from the shape elements themselves and not to rely on our own calculations.
+       */
+
+      if (elements) {
+        for (let i = 0; i < elements.length; i++) {
+          const element = elements[i];
```
Saadnajmi, line 83: super nit: use a foreach loop?

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,100 @@
...
+          );
+        }
+      }
+
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.shimmerWaveContainer>
+            <Slots.shimmerWave />
+          </Slots.shimmerWaveContainer>
+          <Slots.clippingMask clipPath={clipPathsAsMask.join(' ')} />
```
Saadnajmi, line 95: could we pull out the join to it's own var outside of the return?

packages/experimental/Shimmer/src/Shimmer.tsx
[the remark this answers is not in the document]
PPatBoyd: I missed how to apply assertNever to a discriminated union (vs the union-of-types interface from before) -- updating :) also removing the ShimmerElementType that I neglected removing

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
PPatBoyd: ...I swear that caused a build break w/ another type. correcting

packages/experimental/Shimmer/src/Shimmer.types.ts
PPatBoyd: I did mean to change that, I haven't seen what the value is in the default shimmer being translucent -- or icon being a top line slot/prop (also removed) for that matter as a specialized use case, which was my understanding of the built-in use of the translucency.  It sounds more likely to cause confusion w/ a background color bleeding through, IMO.

I'm open to thoughts if there's more background; fwiw it's not a part of the current web shimmer or northstar skeleton interfaces

packages/experimental/Shimmer/src/Shimmer.types.ts
PPatBoyd: I referenced it in another comment but it's only implemented as a top line property/scenario in fluentUI iOS (if it's there for the native fluentui macos it isn't publicly documented).  I don't see the value in doing so when we're talking about adding alpha to colors and overlaying on top of an image.  We'd also be asserting our opinion about how that translucency should be applied to the gradient colors, which would need to be overridden for a different opinion.

It's also constraining in user control over the image relative to the shimmer e.g. aspect ratio; I don't believe we want to be in the business of managing that or extending our Shimmer API to expose those details.

packages/experimental/Shimmer/src/Shimmer.win32.tsx
PPatBoyd: what's the gain?

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: I'll do the followup in this thread. We don't have a FluentUI Apple macOS Shimmer, but that's mostly because we haven't needed one. I prototyped one a while back by porting the iOS code. 
The PR which added using a translucent shimmer in FluentUI Apple is here. I think it does a good job in explaining what the scenario was and why we added it. https://github.com/microsoft/fluentui-apple/pull/245

I'm not sure the decision making behind whether it belongs in FURN Shimmer or not, but it feels weird to me to take away that feature. I think for the purpose of expedience of this review, I'll leave it as a followup between you/Amanda/whoever has an opinion about what belongs and doesn't belong in the shimmer API. I'm not sure I have enough context to confirm.

packages/experimental/Shimmer/src/Shimmer.win32.tsx
PPatBoyd: trades off where I put the 3 `as any`  but nbd

packages/experimental/Shimmer/src/Shimmer.win32.tsx
Saadnajmi: Forgot to prefix my comment with nit.
Mostly keeping logic outside of the JSX. At one point Jason told me not to pass subobjects inside JSX as it forces js to recreate the object on every render. I'm not sure that optimization applies here.

packages/experimental/Shimmer/src/Shimmer.win32.tsx
Saadnajmi: Don't think I understand, what is the tradeoff? If there's a want to keep the any types that's fine. I just didn't see if on my first pass of this code.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -39,79 +38,65 @@ export const Shimmer = compose<ShimmerType>({
...
-              <Stop
-                offset="30%"
-                stopColor={uri ? memoizedShimmerData.gradientTintColor : memoizedShimmerData.shimmerTintColor}
-                stopOpacity={uri ? '0' : '1'}
-              />
+              <Stop offset="10%" stopColor={memoizedShimmerData.shimmerColor} stopOpacity={1} />
+              <Stop offset="20%" stopColor={memoizedShimmerData.shimmerWaveColor} stopOpacity={memoizedShimmerData.gradientOpacity} />
+              <Stop offset="30%" stopColor={memoizedShimmerData.shimmerColor} stopOpacity={1} />
             </AnimatedLinearGradient>
+            <ClipPath id="shimmerView">{rows}</ClipPath>
```
rurikoaraki, line 90: Was this meant to move into the Defs tag?

packages/experimental/Shimmer/src/Shimmer.win32.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { ClippingMaskProps, ShimmerType } from './Shimmer.types.win32';
+import { RCTNativeAnimatedShimmer } from './consts.win32';
+import { convertRectToSvgPath, convertCircleToSvgPath } from './SvgShapeToPath';
+import { withSlots } from '@fluentui-react-native/framework';
+import { assertNever } from 'assert-never';
+
+/** Absolute positioning is used to overlay the clipping mask on top of the shimmer wave. */
+const clippingMask: React.FunctionComponent<ClippingMaskProps> = (props: ClippingMaskProps) => {
+  return (
+    <Svg style={{ position: 'absolute' }} width="100%" height="100%">
```
rurikoaraki, line 16: nit: the style could be a const object instead of regenerated on every render.

packages/experimental/Shimmer/src/SvgShapeToPath.ts
```diff
@@ -0,0 +1,185 @@
...
+// segments of the rect if so, and will be reduced to the maximum values accordingly.
+function convertRoundedRectToSvgPath(
+  x: number,
+  y: number,
+  borderRadiusX: number,
+  borderRadiusY: number,
+  width: number,
+  height: number,
+): string {
+  console.log('Rect parameters provided: x: %f y: %f rx:%f ry:%f width:%f height:%f', x, y, borderRadiusX, borderRadiusY, width, height);
```
rurikoaraki, line 31: is this intended to stay in the code?

packages/experimental/Shimmer/src/Shimmer.types.ts
PPatBoyd: What I see being argued is that inverting the gradient opacity delta (shimmerColor-opacity >=< shimmerWaveColor-opacity) is better for an "image" shimmer element -- represented here with an icon or otherwise svg-compatible shape.  When you can customize the gradient colors, what's the difference?

The way the prop is now doesn't cover a lot -- You can only use one.  you can't make those shimmers shown because you don't have relative positioning at the moment -- if you check out the native code it cheats by walking the UI tree to manage giving a gradient to all the elements, which we can't do from JS. Or you do it through ImageProps crafted based on Shimmer implementation details; IMO an unkind coupling.  I'm making the assumption we're actually talking about raster images here (reason why is next paragraph), which will increase bundle sizes by theme colors * DPI, requiring coordinating resources w/ themes.  Warren Leung told me in another channel that using URI w/ SVG is considered a security hazard by Office [didn't get more info than that]

I'd rather add support for custom polygon/shape as an element type, unless the point is to offer color raster images which would be its own thing (not pictured, I haven't seen it before? and then I'd argue again not worth top-line support and maintenance).  This enables avoiding using resources/resource coordination/URI security concern, I imagine fits better/no worse than images/icons w/ the design pipeline, and can be made a top-line element type.

rurikoaraki (approved):
Looks good to me, althugh I'm not a Shimmer/component expert as a caveat

packages/experimental/Shimmer/src/SvgShapeToPath.ts
PPatBoyd: whoops -- nope ❌

packages/experimental/Shimmer/src/Shimmer.tsx
PPatBoyd: I did it on purpose but to the best of my knowledge it doesn't matter for practical purposes on a ClipPath since it isn't rendered as a raw element, only as a clipRule.

It's specifically important that AnimatedLinearGradient is in Defs so that it isn't rendered on top of being used as a fill for the Rect (though for _those_ practical purposes the rect is unnecessary AFAIK, but not changing it atm).  Elements in defs aren't rendered as elements w/ <Use/> elements.

Saadnajmi (approved):
Approved with the image conversation being a followup when Amanda is back in office to see if there was a need for it.

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -1,114 +1,144 @@
...
-   * Specifies the gradient angle, value should be anywhere between -1 and 1
-   * -1 to 1 maps to 90 degrees horizontally to ≈ 120 degrees
-   * @defaultValue '0'
+   * Defines the tip color of the wave which has a linear gradient.
+   * from shimmerColor to shimmerWaveColor to shimmerColor.
+   * @defaultValue '#E1E1E1'
    */
-  angle?: number;
+  shimmerWaveColor?: ColorValue;
+
```
Saadnajmi, line 125: Juuuust noticed. These cannot be color values because rn-svg doesn't support ColorValue. This has to be `string` or `string | number`

Saadnajmi (dismissed):

packages/experimental/Shimmer/src/Shimmer.types.ts
rurikoaraki: Is this being used in the svg itself? I thought it was the gradient, is that an svg?

PR #863, opened 2021-08-06
Enable hooking state on the stock Pressable component

JasonVMo:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

When we were on a version <0.63 I ported a bunch of the Pressability and Pressable code from core react-native into our repo. This was done because the libraries were not yet available but actually worked fine against the earlier versions. Now that we have a baseline of >0.63 the Pressable component is now available.

The Pressable component will internally create a PressabilityConfig and do all the gesture management. It also has the android_ripple capabilities which we do not have. As such we should sunset much of the code. This change has a few parts:

**Use Base Types (where possible)**

This changes a bunch of the types to match what is exposed via `PressableProps`. This is done to allow the helper hook functions `useHoverHelper`, `useFocusHelper`, and `usePressHelper` to work with either the existing code or the new hook function.

**`usePressableState` hook

This is a new function that does the appropriate prop decoration to get the current state (and prompt state changes at the parent level) that `useAsPressable` does. This one simply works against the stock `Pressable` component instead of a `View`. This leverages the fact that the stock `Pressable` internally will manage the `PressabilityConfig` object.

### Verification

This was done as part of me reworking experimental button. (Hopefully a PR for that will be out later today). I split this out into a separate PR. I did this first though and switched experimental button to use a Pressable internally then verified it worked in the tester.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
+++ b/packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
@@ -1,60 +1,30 @@
 import { RectOrSize } from './InternalTypes';
 import { BlurEvent, FocusEvent, PressEvent, MouseEvent } from './CoreEventTypes';
-import { ViewProps } from 'react-native';
+import { ViewProps, PressableProps } from 'react-native';
 
-export type PressabilityConfig = Readonly<{
+export type PressablePressProps = {
   /**
-   * Whether a press gesture can be interrupted by a parent gesture such as a
-   * scroll event. Defaults to true.
+   * Called when a press gestute has been triggered.
    */
-  cancelable?: boolean;
+  onPress?: PressableProps['onPress'];
 
   /**
-   * Whether to disable initialization of the press gesture.
+   * Called when the press is activated to provide visual feedback.
    */
-  disabled?: boolean;
+  onPressIn?: PressableProps['onPressIn'];
 
   /**
-   * Amount to extend the `VisualRect` by to create `HitRect`.
+   * Called when the press is deactivated to undo visual feedback.
    */
-  hitSlop?: ViewProps['hitSlop'];
+  onPressOut?: PressableProps['onPressOut'];
 
   /**
-   * Amount to extend the `HitRect` by to create `PressRect`.
+   * Called when a long press gesture has been triggered.
    */
-  pressRectOffset?: RectOrSize;
+  onLongPress?: PressableProps['onLongPress'];
+};
 
-  /**
-   * Whether to disable the systemm sound when `onPress` fires on Android.
-   **/
-  android_disableSound?: boolean;
-
-  /**
-   * Duration to wait after hover in before calling `onHoverIn`.
-   */
-  delayHoverIn?: number;
-
-  /**
-   * Duration to wait after hover out before calling `onHoverOut`.
-   */
-  delayHoverOut?: number;
-
-  /**
-   * Duration (in addition to `delayPressIn`) after which a press gesture is
-   * considered a long press gesture. Defaults to 500 (milliseconds).
-   */
-  delayLongPress?: number;
-
-  /**
-   * Duration to wait after press down before calling `onPressIn`.
-   */
-  delayPressIn?: number;
-
-  /**
-   * Duration to wait after letting up before calling `onPressOut`.
-   */
-  delayPressOut?: number;
-
+export type PressableFocusProps = {
   /**
    * Called after the element loses focus.
    */
@@ -64,7 +34,9 @@ export type PressabilityConfig = Readonly<{
    * Called after the element is focused.
    */
   onFocus?: (event: FocusEvent) => any;
+};
 
+export type PressableHoverProps = {
   /**
    * Called when the hover is activated to provide visual feedback.
    */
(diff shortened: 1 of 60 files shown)


Saadnajmi:
Thanks for this change @JasonVMo ! I was about to ask what I should do here now that we're on 0.63 and people want to use Pressable =) I'll review this when I can later today

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
```diff
@@ -1,60 +1,30 @@
...
-  delayPressIn?: number;
+  onPressOut?: PressableProps['onPressOut'];
 
   /**
-   * Duration to wait after letting up before calling `onPressOut`.
+   * Called when a long press gesture has been triggered.
    */
-  delayPressOut?: number;
+  onLongPress?: PressableProps['onLongPress'];
+};
```
Saadnajmi, line 25: Why do we still need these extra props? Are they not in core?

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
```diff
@@ -64,7 +34,9 @@ export type PressabilityConfig = Readonly<{
    * Called after the element is focused.
    */
   onFocus?: (event: FocusEvent) => any;
```
Saadnajmi, line 36: Note: I think it's an open action item that we need to add onFocus/onBlur to RN-macOS pressable. I'm not sure how that affects FURN downstream.

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
```diff
@@ -74,43 +46,93 @@ export type PressabilityConfig = Readonly<{
...
+  /**
+   * While the user API is onHoverIn the View event is onMouseEnter
+   */
+  onMouseEnter?: (event: MouseEvent) => any;
+
+  /**
+   * While the user API is onHoverOut the View event is onMouseLeave
+   */
+  onMouseLeave?: (event: MouseEvent) => any;
+};
```
Saadnajmi, line 61: Where is this ViewEvent coming from? From RNW/win32/macOS?

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
```diff
@@ -74,43 +46,93 @@ export type PressabilityConfig = Readonly<{
...
+
+export type PressabilityConfig = Readonly<
+  PressablePressProps &
+    PressableFocusProps &
+    PressableHoverProps & {
```
Saadnajmi, line 66: Is this not a flow type that will not get read by typescript? That's where I thought `ReadOnly<>` decorators came from.

Saadnajmi: Is there any chance of this config getting out of sync with RN-Core in a future version? I.E is this something we could just import from RN-Core?

packages/utils/interactive-hooks/src/useAsPressable.ts
```diff
@@ -1,19 +1,23 @@
...
+  PressablePropsExtended,
 } from './useAsPressable.types';
 import { usePressability } from './usePressability';
 
 /**
  * hover specific state and callback helper
  */
 // eslint-disable-next-line @typescript-eslint/ban-types
-function useHoverHelper<T extends object>(props: IWithPressableOptions<T>): [IWithPressableOptions<object>, IHoverState] {
+function useHoverHelper(props: PressableHoverProps): [PressableHoverProps, IHoverState] {
```
Saadnajmi, line 20: Do we still need the aslant disable comments here?

Saadnajmi: Same for below lines.

packages/utils/interactive-hooks/src/useAsPressable.ts
```diff
@@ -146,3 +150,21 @@ export function useHoverState<T extends object>(props: IWithPressableOptions<T>)
...
+ *
+ * @param props - props to pass to a Pressable component
+ * @returns - modified props to pass into the Pressable as well as the current state with regards to hover, focus, and press
+ */
+export function usePressableState(props: PressablePropsExtended): { props: PressableProps; state: IPressableState } {
+  const { onPressIn, onPressOut, onHoverIn, onHoverOut, onFocus, onBlur, ...rest } = props;
+  const [focusProps, focusState] = useFocusHelper({ onFocus, onBlur });
+  const [pressProps, pressState] = usePressHelper({ onPressIn, onPressOut });
+  const [hoverProps, hoverState] =
+    Platform.OS !== 'android' && Platform.OS !== 'ios' ? useHoverHelper({ onHoverIn, onHoverOut }) : [{}, {}];
```
Saadnajmi, line 167: nit: Maybe something like this?

```suggestion
const platformSupportsHover = (Platform.OS !== 'android' && Platform.OS !== 'ios');
const const [hoverProps, hoverState] = platformSupportsHover ? useHoverHelper({ onHoverIn, onHoverOut }) : [{}, {}];
```

packages/utils/interactive-hooks/src/useAsPressable.types.ts
```diff
@@ -29,3 +30,15 @@ export type IPressableHooks<T extends object> = {
...
   state: IPressableState;
 };
+
+/**
+ * The hover props on RN 0.63 don't exist in react-native, only on the desktop SKUs, but we need them for hooking
+ * mouse events on windows and macos. This adds them to the props type, stripping them from the base pressable type as
+ * future proofing for when we upgrade to 0.64+.
+ *
+ * Similarly the focus methods exist at the View level but are not exposed on pressable props in 0.63. As a result they
+ * need a similar treatment
```
Saadnajmi, line 40: Is this true for windows as well? I know it's not for macOS but I thought it was for windows?

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
JasonVMo: Think of the stock Pressable component as a thin shim around the Pressability library in react-native. The code I brought in before in the /Pressability directory was a port of that code to our project (and to TypeScript). I'm mostly leaving this part unchanged for now. Also PressabilityConfig is not quite the same as PressableProps, they intersect but aren't equal.

The long term plan would be:

1. Get this change in
2. Switch all components that use useAsPressable() on View components to usePressableState on Pressable components.
3. Remove these old files along with the useAsPressable hook.

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
JasonVMo: Likely the events never trigger but we should make sure that is on the 0.64 list.

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
JasonVMo: That's correct, though on 0.64 it may get an implementation on core at some point. The Pressability library has routing code for it, so the events are handled and work, this is simply the name of the events that get view implements.

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
Saadnajmi: The long term plan helps me understand the whole port of the Pressability library is going away, thank you. That explains to me why we still have these types.

packages/utils/interactive-hooks/src/Pressability/Pressability.types.ts
JasonVMo: TypeScript has a Readonly concept as well. I put it in to PressabilityConfig when I did the port for consistency and am just staying consistent with what I had before here. In terms of getting out of sync, most of the types in here are not exposed and are on the inner Pressability library, not Pressable. As we get rid of the old code the interfaces should shift to be based off of what is exposed on PressableProps as much as possible.

Saadnajmi (approved):

packages/utils/interactive-hooks/src/useAsPressable.ts
JasonVMo: Good catch

rurikoaraki (approved):

PR #881, opened 2021-08-10
RadioGroup (macOS): Remove font family setting to eliminate runtime error

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
RadioGroup.settings.ts: Remove the `fontFamily: 'inherit'` style declaration, which is not recognized on macOS and results in a runtime error when displaying the RadioGroup test module:

Screen Shot 2021-08-10 at 12 02 37 PM

### Verification
Verified that the error no longer appears in the test app.

Also checked that this doesn't affect the appearance or behavior of the test app in win32.  Win32 screenshots below.

| Screen Shot 2021-08-10 at 1 34 17 PM | Screen Shot 2021-08-10 at 1 37 09 PM |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):
I'm somewhat sure this is not recognized on the other platforms either. I think I saw it before and it seemed like something that would work in CSS / web but not react native. Thanks for catching it, and testing on win32!

chiuam (approved):

samuelfreiberg (approved):

PR #883, opened 2021-08-10
Patch up branding entries on Theme page on Test App

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Patching up applyBrand.ts to include other brand-related entries that haven't been added so that the Tester shows the right colors when the brand change. Note that this applyBrand.ts really only shows the theme properly for the Colorful theme - for other themes, we might not have the right mapping to the right entries. Since it's just code that runs in the test app, that might be ok short term, but long term it'd be nice to be able to move away from that. FWIW the Default brand will show the proper colors, but will show them using Office orange.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
```diff
@@ -156,10 +174,10 @@ export const applyBrand = (currentBrand: string): PartialTheme => {
           compoundBrandForeground1Pressed: ramp.AppShade20,
           brandForeground1: ramp.AppPrimary,
           brandForeground2: ramp.AppShade10,
-          brandBackground: ramp.AppPrimary,
-          brandBackgroundHover: ramp.AppShade10,
-          brandBackgroundPressed: ramp.AppShade30,
-          brandBackgroundSelected: ramp.AppShade20,
+          brandBackground1: ramp.AppPrimary,
```
samuelfreiberg, line 177: I'm a bit confused by this naming convention... Why add the "1" suffix?

rurikoaraki: The names for some of the entries are taken from the new alias tokens that design has decided on, some of which have numbers. The other issue here is that brandBackground already exists in a different subtype that makes up the Palette type, so in order to avoid the name clash I decided to append 1 to the brandBackground we're getting from the alias tokens. Since other alias tokens already have that naming convention, I figured this would be the best way to resolve the name conflict.

Saadnajmi: What's the other subtype where brandBackground exists? Is that the ControlColorTokens subtype? I think that may only be used by the experimental button and not by and clients if so.

rurikoaraki: Yeah it's the ControlColorTokens subtype. I can go change that instead if preferred.

Saadnajmi: Does experimental button still render fine with the default theme? I think that's the only control that should be affected

lenahong (approved):

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/theme/applyBrand.ts
rurikoaraki: @Saadnajmi So I'm looking at the code and I'm not entirely sure how the Experimental Button has worked in the default theme up til now. It has some coloring in master right now but I think that's an incidental thing from the name clash I'm trying to fix in this PR... Works fine in the Office theme though.

Guess it's time to patch the default theme.

rurikoaraki: Though I would want to do that in its own PR if that's alright

Saadnajmi: Oh yeah totally! I didn't mean that to be a blocking comment sorry

PR #885, opened 2021-08-11
RadioGroup (macOS): Remove automatic select on focus

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

#### Native Behavior
When moving focus around a native radio group on macOS, the selection will follow the keyboard focus when using arrow keys to navigate.  However, when using VO+arrow keys to move screen reader focus around, the selection does not follow the focus.  VoiceOver focus exists to help the user understand what controls are on the screen, and that inspection doesn't imply any intent to take an action: that's what the VO+space to activate a control is for.

When VoiceOver is off, the standard VO keyboard modifiers (control and option) do not affect the arrow key behavior: that is, command-option-arrow keys behaves the same as unmodified arrow keys, and moves the selection with the focus.  Furthermore, when VoiceOver is on, using unmodified arrow keys moves the selection around with the keyboard and VoiceOver focus, the same as when VoiceOver is off.

#### Existing FURN macOS Behavior
Seeking to emulate the native behavior where selection follows focus, we added an override that selects any focused radio button.  This does the right thing in 3 out of 4 cases: unmodified arrow keys when VoiceOver is on or off, and command-option-arrow keys when VoiceOver is off.  However it does the wrong thing in the critical case of VoiceOver navigation with command-option-arrow keys, and the selection continues to follow the focus.  This is bad for a VoiceOver user, since it complicates their navigation by activating the controls that they are focusing on, without being asked to do so.

#### Changes in this Pull Request
The cure was worse than the disease.  If we look at the behavior of native checkboxes in a focus zone, the arrow keys navigate between them but the spacebar needs to be pressed to change the state of a box.  This is also the default behavior of a FURN radio group if we don't override the behavior upon receiving keyboard focus.  This slight deviation from native behavior shouldn't be especially confusing to the user: spacebar to activate a focused control is standard.  And it fixes the problem with VoiceOver navigation, so that selecting a radio button is always an intentional action and not a side effect of moving focus.

This change removes the override.

#### VoiceOver off
| VoiceOver Off Before | VoiceOver Off After |

#### VoiceOver on
| VoiceOver On Before | VoiceOver On After |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


chiuam (approved):
minor: On native macOS, keyboard focus doesn't follow the VO cursor when navigating inside focus zone, should we adopt this behavior?

Saadnajmi (approved):
Thanks for the excellent writeup! I agree with your assessment that the difference of focus not automatically clicking will be a minor one for users to understand. I personally feel like that's an oversight from Apple, given how archaic/undocumented their own radio group control is. Thanks for the fix!

chripear:
> minor: On native macOS, keyboard focus doesn't follow the VO cursor when navigating inside focus zone, should we adopt this behavior?

Yes we should try to match native behavior, but as a much lower priority goal and not in this PR.  Eventually if we can make the VO cases work correctly we should also make the arrow keys change the selection to match native, but the current changes put us in a significantly better state than we were in.

Issue #888, opened 2021-08-11
Icon color not working correctly


lindaweng:
### For SVGs:
Icon color is currently being finally set within the `renderSvg` call. There, it is pulling the color from `svgIconProps.color` which is the same as `iconProps.svgSource.color`. In actuality, color is being passed in at `iconProps.color`. So `svgIconProps.color` is always undefined and thus the color of the icon defaults to white or black depending on light/dark mode
By changing `svgIconProps.color` to `iconProps.color`, the color that users pass in gets recognized and works. However, in the call to make the Icon, if the user doesn't pass in a color, the color that gets passed to `renderSvg` is `theme.colors.buttonText`. (Minor change there, in Icon component creation, `props.color` should be `rest.color` as `rest` is the most updated passed in props at that point.) This works fine for default or ghost buttons but for primary and fab styled buttons this causes the icon to blend in with the background.
For experimental button, the icon is set with a `tintColor` style which tints it to the accurate color for each button. To pull this color within Icon, it should be `iconProps.style.tintColor`. However, if we pull this color, we do not pull the color that the user passes in so the user is unable to customize the color of the icon. Also, the non-experimental button does not set a tint color so all icons in those buttons will be white/black depending on light/dark mode.

### For raster image (PNG) icons:
`tintColor` accurately tints the icon to the correct color. Currently icons in the non-experimental button are not being tinted to any color so they appear as their natural base color. There is no need to change any code.

### For font icons:
Similar to SVG icons in the fact that changing `iconProps.color` to `iconProps.style.tintColor` in `renderFontIcon` tints the icon to the appropriate color but ignores custom colors. Currently the icons blend into the background for fab and primary styled buttons.

### The fix:
The icon should check if the user passed in a color, if so, use that color. If the user did not pass in a color, check if a tint color exists. If it does, use the tint color. Finally if the tint color does not exist, then default to black/white depending on light/dark mode.
Also, adding a `tintColor` style prop to the icon in the non-experimental button may be nice so that the icons match the button theme if a custom color is not passed in.

### Current coloring:
Experimental button
<img width="123" alt="Screen Shot 2021-08-11 at 2 40 17 PM" src="https://user-images.githubusercontent.com/30802320/129107659-9a27e79f-05b9-4daa-a361-6671d2ecb364.png">
Non-experimental button
[...]

PR #889, opened 2021-08-11
Fix icons in experimental button and add fab

lindaweng:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Make icons show up in experimental buttons.
- Update the icon slot in experimental button from an Image to an Icon
- Delete `filterImageProps` as the icon is no longer an Image
- Add `createIconProps` to create the icon props if user passes in a raster image (PNG) or font icon
- Icon coloring may look odd: See https://github.com/microsoft/fluentui-react-native/issues/888

Create FAB (Floating Action Button) with experimental button
- Create `shadow-tokens.ts` so that the shadowStyles can be pulled out of the tokens and applied to the button
- Create a new token and prop called `fab` to apply the appropriate styles

FAB still needs improvements to exactly match the Figma designs
- Exact font sizing (currently smaller by 1 or 2)
- Currently hard coded to size large for button size
- Currently hard coded to accent style (subtle style not supported)

Added Experimental button test to Android
- Styles are slightly different compared to iOS when normal (see video)
  - Primary coloring is currently off due to the naming change from prefix `brand` to `branded`: See https://github.com/microsoft/fluentui-react-native/issues/893
- Shadows are made with elevation. They don't exactly match the iOS shadows, but are close enough
- PNG icons are not showing up: See https://github.com/microsoft/fluentui-react-native/issues/892

### Verification

- Created icons of different types (PNG, SVG, and font)
- Created a FAB with text and without text
- Compared Android and iOS versions side by side to ensure they look similar

| Including an icon in an experimental button did not show or create space for the icon on the button | Screen Shot 2021-08-11 at 3 31 45 PM |

Android vs iOS

https://user-images.githubusercontent.com/30802320/129270310-3d6df546-9977-45ff-9271-357f20d142f8.mov

Android Color Naming Issue
Screen Shot 2021-08-12 at 3 25 31 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+        />
+      ) : null}
+      {svgIconsEnabled ? (
+        <Button icon={{ svgSource: svgProps, width: 20, height: 20 }} content="SVG" style={commonTestStyles.vmargin} />
+      ) : null}
+      <Button primary icon={testImage} content="PNG" style={commonTestStyles.vmargin} />
+      <Button icon={testImage} content="PNG" style={commonTestStyles.vmargin} />
+    </View>
+  );
+};
```
Saadnajmi, line 52: Did you copy this from another test or handcraft it yourself?

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -21,9 +21,10 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
           backgroundColor: tokens.backgroundColor,
           ...borderStyles.from(tokens, theme),
           ...layoutStyles.from(tokens, theme),
+          ...shadowStyles.from(tokens, theme),
         },
       }),
-      ['backgroundColor', 'width', ...borderStyles.keys, ...layoutStyles.keys],
+      ['backgroundColor', 'width', ...borderStyles.keys, ...layoutStyles.keys, ...shadowStyles.keys],
```
Saadnajmi, line 27: @rurikoaraki @FalseLobster FYI since both of you have been touching shadows as of late

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -17,6 +17,48 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme
...
+        color: t.colors.brandPressedContent,
+        borderColor: t.colors.brandPressedBorder,
+        iconColor: t.colors.brandPressedIcon,
+      },
+      focused: {
+        backgroundColor: t.colors.brandFocusedBackground,
+        color: t.colors.brandFocusedContent,
+        borderColor: t.colors.brandFocusedBorder,
+        iconColor: t.colors.brandFocusedIcon,
+      },
```
Saadnajmi, line 60: @rurikoaraki you are changing some of these token names correct? from `brand` -> `branded` ?

packages/experimental/Button/src/Button.tsx
```diff
@@ -1,36 +1,44 @@
...
     // now return the handler for finishing render
     return (final: ButtonProps, ...children: React.ReactNode[]) => {
       const { icon, content, ...mergedProps } = mergeProps(button.props, final);
-
+      const marginBetween = {
+        marginLeft: 0,
+      };
+      if (icon && content) {
+        marginBetween.marginLeft = 10;
+      }
```
Saadnajmi, line 33: Is there a margin you can put into root that provides spacing between it's children instead of doing this conditional margin? Something like alignContent https://reactnative.dev/docs/flexbox#align-content ?

Saadnajmi:
Something interesting I learned here is that iOS and macOS support font-icons with code points.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
lindaweng: I copied it over from `ButtonVariantTestSection.tsx`. While developing, I put everything in that test but for organization, I separated out the icons into their own test section.

packages/experimental/Button/src/Button.tsx
lindaweng: No, because the root is only wide enough for the content and padding that is hard coded in. Using justifyContent (since it is in 'row' direction), there is no space to do space-around which seems like the closest thing I would want. Essentially everything is packed in as tight as can be without extra spacing to move things around. If there was a fixed width of 100 let's say and padding, icon, and content only takes up 60 out of that 100, the rest (40) would be left over for spacing adjustments.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+import TestSvg from './test.svg';
+
+export const ButtonIconTest: React.FunctionComponent = () => {
+  const fontBuiltInProps = {
+    fontFamily: 'Arial',
+    codepoint: 0x2663,
+    fontSize: 32,
+  };
+
+  const testImage = require('./icon_24x24.png');
```
Saadnajmi, line 16: if we changed the require(icon.png) to an import, could we get rid of the aslant disabling comment?

packages/experimental/Button/src/Button.tsx
Saadnajmi: OK, if we go this route, I would go with this suggestion to make it more of a const and less a mutable variable.
I'll think a bit more today if there's a better way to modify the style today. Otherwise the rest of the change looks good!

```suggestion
      const marginBetween = {
        marginLeft: (icon && content) ? 10 : 0,
      };
```

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
Saadnajmi: If it's the only require, can we move the disabling comment to just above this line? I think that will ensure it only applies to the one line instead of the whole file, in case more require() calls re added in the future.

Saadnajmi (approved):
Looks good on my end! I'd like someone else to take a look but I'll merge it by tomorrow if nobody gets a chance

packages/experimental/Button/src/ButtonTokens.ts
rurikoaraki: Yeah, brand -> branded here.

packages/utils/tokens/src/shadow-tokens.ts
```diff
@@ -0,0 +1,23 @@
+import { ColorValue } from 'react-native';
+import { IOperationSet } from '@uifabricshared/foundation-tokens';
+import { ITheme } from '@uifabricshared/theming-ramp';
+import { tokenBuilder } from './tokenBuilder';
+
+export interface IShadowTokens {
+  shadowColor?: ColorValue;
```
rurikoaraki, line 7: I think we have two separate colors for shadows with the new tokens - one is for the shadow around the entire component, and the other if for the shadow that's positioned to the lower right of the component

rurikoaraki: I guess this can be applied to both the key and ambient colors so this isn't an issue

packages/utils/tokens/src/shadow-tokens.ts
```diff
@@ -0,0 +1,23 @@
+import { ColorValue } from 'react-native';
+import { IOperationSet } from '@uifabricshared/foundation-tokens';
```
rurikoaraki, line 2: Just an FYI that uifabricshared packages are generally deprecated - we're trying to use other packages instead now. For the Theme type you can probably take it from @fluentui-react-native/framework instead

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -17,6 +17,48 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme
     fluid: {
       width: '100%',
     },
+    fab: {
+      borderRadius: 100, // big number for always rounded corners
+      shadowColor: '#000000',
```
rurikoaraki, line 22: nit: you could use the global Shadow tokens here if one fits the same values, but not blocking

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
lindaweng: I could move it, but I also copied it from the ButtonIconTest.tsx for the non-experimental button where it's put as the first line of the file. I just kept it as the first line to be consistent.

packages/experimental/Button/src/ButtonTokens.ts
Saadnajmi: I took a quick look, and it looks like the shape of the global tokens is different from the props in View. I.E:
```
// globalTokens.shadow.2
"shadow": {
  "2": [
      {
          "x": 0,
          "y": 0,
          "blur": 2,
          "color": "#0000001f"
      },
      {
          "x": 0,
          "y": 1,
          "blur": 2,
          "color": "#00000024"
      }
  ],
  ...
}
```
vs what RN / iOS wants:
```
style: {
  shadowColor: ColorValue,
  shadowOffset: {
    width: number,
    height: number,
  },
  shadowRadius: number,
  shadowOpacity: number,      
}
```

To use the globalTokens, we will probably need some appropriate conversion layer?

rurikoaraki: You should be able to pass in the color as-is, without needing opacity? or is opacity strictly needed? Blur is the same as radius, I think

Saadnajmi: Actually, on a second pass I think you are right. 
shadowColor -> color (first 6 hex values)
shadowOffset -> x / y
blur -> radius
shadowOpacity -> color (last 2 hex values)

Saadnajmi: @lindaweng If you can't get to that before the end of your internship, I'll log a bug on our end to use the global tokens as a followup

lindaweng: The global shadow tokens 8 does match the same values, but I would only be able to use it for shadowOffset and shadowRadius. They correspond to the x, y, and blur values in global shadow tokens. The color in global tokens combines the color and opacity whereas RN separates them into shadowColor and shadowOpacity. I'm not sure if it would make things more or less confusing with some props set with global shadow tokens and some not.

lindaweng: Obviously I didn't see the other comments before making mine, but I'm just not sure how to separate out the color into 2 different parts

lindaweng: I got the hex value splitting working. It's not the prettiest, but it works with the global shadow tokens.

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -2,7 +2,7 @@ import { Theme } from '@fluentui-react-native/framework';
 import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { ButtonTokens } from '.';
 
-export const buttonStates: (keyof ButtonTokens)[] = ['fluid', 'primary', 'ghost', 'hovered', 'focused', 'pressed', 'disabled'];
+export const buttonStates: (keyof ButtonTokens)[] = ['fab', 'fluid', 'primary', 'ghost', 'hovered', 'focused', 'pressed', 'disabled'];
```
chripear, line 6: The 'fab' acronym seems like it will be hard for subsequent contributors to figure out.  Is there something longer but more descriptive/self-evident that we could call it?

Saadnajmi: probably just `floating` , since fab stands for floating-action-button. It's an industry term though

lindaweng: Yeah, we'd also want to make sure clients understand what `fab` means. If we feel like `floating` is more descriptive and easier to understand for both devs and clients, we can switch to `floating`.

lindaweng: But if `fab` is common in the industry and people coming into contact with it would know the term, we can also stick with `fab`.

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -17,6 +18,49 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme
...
     },
+    fab: {
+      borderRadius: 100, // big number for always rounded corners
+      shadowColor: globalTokens.shadow[8][1].color.substr(0, 7),
+      shadowOffset: {
+        width: globalTokens.shadow[8][1].x,
+        height: globalTokens.shadow[8][1].y,
+      },
+      shadowRadius: globalTokens.shadow[8][1].blur,
+      shadowOpacity: Number('0x' + globalTokens.shadow[8][1].color.substr(7, 2)) / 255.0,
```
Saadnajmi, line 29: You could define this helper above (with the helpful inline comments)
```
const shadowStyleFromGlobalToken = (shadowToken: number) => {
  const shadow = globalTokens.shadow[shadowToken][1] // Grab the second shadow from the global token, as we only support displaying one shadow
  return {
    // iOS Shadow props
    shadowColor: shadow.color.substr(0, 7), // split out the color
    shadowOpacity: Number('0x' + shadow.color.substr(7, 2)) / 255.0, // split out the opacity
    shadowRadius: shadow.blur,
    shadowOffset: {
      width: shadow.x,
      height: shadow.y
    },
    // Android shadow props
    elevation: shadowToken 
  }
}
```

and then do:
```
fab: {
  ...shadowStyleFromGlobalToken(8),
  // rest of props
}

```

packages/experimental/Button/src/ButtonTokens.ts
Saadnajmi: Yeah I vote to keep it unless someone else complains

packages/experimental/Button/src/ButtonTokens.ts
Saadnajmi: You could define this helper above (with the helpful inline comments)
```
const shadowStyleFromGlobalToken = (shadowToken: number) => {
  const shadow = globalTokens.shadow[shadowToken][1] // Grab the second shadow from the global token, as we only support displaying one shadow
  return {
    // iOS Shadow props
    shadowColor: shadow.color.substr(0, 7), // split out the color
    shadowOpacity: Number('0x' + shadow.color.substr(7, 2)) / 255.0, // split out the opacity
    shadowRadius: shadow.blur,
    shadowOffset: {
      width: shadow.x,
      height: shadow.y
    },
    // Android shadow props
    elevation: shadowToken 
  }
}
```

and then do:
```
fab: {
  ...shadowStyleFromGlobalToken(8),
  // rest of props
}

```

Saadnajmi: @rurikoaraki to answer your earlier question, I thought shadowOpacity would not be needed but it doesn't render properly unless you provide it. So we still necessarily need it yes.

PR #897, opened 2021-08-16
react-native-windows bump to 0.64

dake3601:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Bumped react-native-windows to [0.64](https://github.com/microsoft/react-native-windows/releases/tag/react-native-windows_v0.64.0). The main motivation for this bump is because while working on the Tabs windows control we found a bug that [can't be backported](https://github.com/microsoft/react-native-windows/pull/8381) to 0.63.

### Verification

Manually ran the tester app and the E2E test for win32 and windows. 
Manually ran the macOS, iOS and Android tester app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/package.json
+++ b/apps/fluent-tester/package.json
@@ -73,7 +73,7 @@
     "react-native-macos": "^0.63.0",
     "react-native-svg": "^12.1.1",
     "react-native-svg-transformer": "^0.14.3",
-    "react-native-windows": "^0.63.0",
+    "react-native-windows": "^0.64.0",
     "react-test-renderer": "~16.13.1",
     "rnpm-plugin-windows": "^0.5.1-0",
     "ts-node": "^8.10.1",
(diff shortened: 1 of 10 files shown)


Saadnajmi:
Bumping one platform but not the others of the test apps has generally caused me issues in the past, because you can end up with multiple versions of react-native in the monorepo which can cause weird clashes with dependencies and path resolutions. We are generally limited to the lowest common denominator / what Office is currently using, which is 0.63. If we are to do this, I would say you have to be very thorough of testing that all the test apps on every platform still works. I know from personal experience that react-native-macos 0.63 doesn't play well with react-native 0.64 for example.

Saadnajmi (approved):

PR #898, opened 2021-08-16
Custom theming documentation

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding documentation for creating custom FURN themes.

Aim is to provide clear examples of how to add custom themes. Thanks to @FalseLobster for her examples in her AccChecker example, some of this documentation was modeled off of that example.

If there are use cases missing please let me know!

### Verification

Sending documentation to some partners as a test run

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/docs/pages/Theming/CustomTheme.md
+++ b/docs/pages/Theming/CustomTheme.md
@@ -39,13 +39,17 @@ For example, if you want to change the settings on the [Button component](https:
 ```ts
 components: {
   Button: {
-    backgroundColor: 'primaryButtonBackground', // Different semantic color
-    color: 'neutralBackground1', // Alias token
+    tokens: {
+      backgroundColor: 'primaryButtonBackground', // Different semantic color
+      color: 'neutralBackground1', // Alias token
+    },
     _overrides:{
-      hovered: {
-        backgroundColor: theme.host.colors['AppShade10'], // Get brand color from host
-        color: theme.host.colors['Gray96'], // Get gray from host
-        borderColor: theme.host.palette.Bkg, // Get entry from palette from host
+      tokens:{
+        hovered: {
+          backgroundColor: theme.host.colors['AppShade10'], // Get brand color from host
+          color: theme.host.colors['Gray96'], // Get gray from host
+          borderColor: theme.host.palette.Bkg, // Get entry from palette from host
+        },
       },
     }
   },
--- a/docs/pages/Theming/CustomTheme.md
+++ b/docs/pages/Theming/CustomTheme.md
@@ -2,7 +2,7 @@
 
 One of the powerful aspects of FURN theming is that it is heavily customizable. We recognize that users may have a need to change various parts of default themes, so we allow for several ways to specify overrides.
 
-You can create a custom theme by either building on top of an existing ThemeReference or building one from scratch. This then can be used by [passing it into our ThemeProvider](./Basics.md), which makes the theme available for all child components of the provider.
+You can create a custom theme by either building on top of an existing `ThemeReference` or building one from scratch. This then can be used by [passing it into our `ThemeProvider`](./Basics.md), which makes the theme available for all child components of the provider.
 
 ## Different aspects of customization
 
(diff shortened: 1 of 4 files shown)


docs/pages/Theming/CustomTheme.md
```diff
@@ -1,7 +1,104 @@
...
+    backgroundColor: 'primaryButtonBackground', // Different semantic color
+    color: 'neutralBackground1', // Alias token
+    hovered: {
+      backgroundColor: theme.host.colors['AppShade10'], // Get brand color from host
+      color: theme.host.colors['Gray96'], // Get gray from host
+      borderColor: theme.host.palette.Bkg, // Get entry from palette from host
+    },
+  },
+},
+```
```
Saadnajmi, line 51: nit: These overrides are for the experimental button, but we link the original button here. Also, could we just link button.settings/button.styling?

docs/pages/Theming/CustomTheme.md
```diff
@@ -1,7 +1,104 @@
...
+}
+```
+
+#### Spacing example
+
+```ts
+spacing: {
+  s1: '10px';
+}
+```
```
Saadnajmi, line 98: Nothing actually uses spacing yet, correct? Maybe just leave it out?

docs/pages/Theming/CustomTheme.md
rurikoaraki: Ah does this only work for the newer compose framework?

docs/pages/Theming/CustomTheme.md
rurikoaraki: Apparently the Stack does?

docs/pages/Theming/CustomTheme.md
Saadnajmi: The structure of the object you passed in is specific to the experimental buttons' settings, specifically the nested hover. It works with the old framework, the shape would just look different since you have stuff like `_precedences` in there which aren't in the new framework.

docs/pages/Theming/CustomTheme.md
Saadnajmi: Oh? TIL...

FalseLobster (approved):

docs/pages/Theming/CustomTheme.md
rurikoaraki: Ah okay my bad. I copied the structure from something else and changed the component without really thinking about it. I'll fix it up.

rurikoaraki: Should be fixed now, thanks for catching!

docs/pages/Theming/CustomTheme.md
```diff
@@ -34,23 +34,29 @@ components: {
...
     },
+    _overrides:{
+      tokens:{
+        hovered: {
+          backgroundColor: theme.host.colors['AppShade10'], // Get brand color from host
+          color: theme.host.colors['Gray96'], // Get gray from host
+          borderColor: theme.host.palette.Bkg, // Get entry from palette from host
+        },
+      },
+    }
```
Saadnajmi: I kinda liked the new framework's styling structure better than the old for documentation 😅 but this is probably more useful since the old button is what is (presumably) used in devmain more.

Saadnajmi (approved):

docs/pages/Theming/CustomTheme.md
Saadnajmi: An aside, a simpler .settings structure for themability I think was one of the goals of the new framework.

rurikoaraki: Yeah, I agree that the newer frameworks are better, but since the goal is to be friendly to users and the experimental button isn't exported right now, this seemed like a better example to use for the reason you mentioned.

PR #899, opened 2021-08-17
Update Apple Native Module naming convention (MSF -> FRN)

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

As per [Microsofts' Objective C Guide](https://microsoft.github.io/objc-guide/Classes/Naming.html), objective c classes by convention have a 3 letter prefix, so as to avoid clashes in the global namespace. We were previously using "MSF" as our prefix, since most of our native modules were wrappers around FluentUI Apple components, which also use MSF.  We now have completely custom controls and some wrappers around Appkit types, so "MSF" was starting to make less sense. To standardize, we will adopt the "FRN" prefix over "MSF", which should be more specific to this repo.

Other changes I made while bulk editing names:
- Renamed the names of pods in each .podspec from "FluentUI-React-Native-<>" to "FRN<>", aka the name of the actual control. The name here doesn't matter too much, but the old name felt way too wordy.

### Verification

CI Passing should ensure that I didn't miss a rename somewhere and things build 🙃.
I'll also re-enable the "NuGet Publish Dry Run" job temporarily to make sure the NuGet package still builds/packs.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


package.nuspec
```diff
@@ -10,38 +10,38 @@
   </metadata>
   <files>
     <!-- iOS device debug -->
-    <file src="DerivedData\Build\Products\Debug-iphoneos\FluentUI-React-Native-Avatar\libFluentUI-React-Native-Avatar.a" target="Debug-iphoneos"/>
-    <file src="DerivedData\Build\Products\Debug-iphoneos\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Debug-iphoneos"/>
-    <file src="DerivedData\Build\Products\Debug-iphoneos\FluentUI-React-Native-Date-Picker\libFluentUI-React-Native-Date-Picker.a" target="Debug-iphoneos"/>
+    <file src="DerivedData\Build\Products\Debug-iphoneos\FRNAvatar\libFRNAvatar.a" target="Debug-iphoneos"/>
+    <file src="DerivedData\Build\Products\Debug-iphoneos\FRNButton\libFRNButton.a" target="Debug-iphoneos"/>
+    <file src="DerivedData\Build\Products\Debug-iphoneos\FRNDate-Picker\libFRNDate-Picker.a" target="Debug-iphoneos"/>
```
Saadnajmi, line 15: guess I can remove this hyphen

package.nuspec
```diff
@@ -10,38 +10,38 @@
...
-    <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-Button\libFluentUI-React-Native-Button.a" target="Debug-macosx"/>
-    <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-FocusZone\libFluentUI-React-Native-FocusZone.a" target="Debug-macosx"/>
-    <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-MenuButton\libFluentUI-React-Native-MenuButton.a" target="Debug-macosx"/>
-    <file src="DerivedData\Build\Products\Debug\FluentUI-React-Native-Radio-Button\libFluentUI-React-Native-Radio-Button.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\FRNAvatar\libFRNAvatar.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\FRNButton\libFRNButton.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\RCTFocusZone\libRCTFocusZone.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\FRNMenuButton\libFRNMenuButton.a" target="Debug-macosx"/>
+    <file src="DerivedData\Build\Products\Debug\FRNRadio-Button\libFRNRadio-Button.a" target="Debug-macosx"/>
```
Saadnajmi, line 37: and the last hyphen here

packages/components/FocusZone/RCTFocusZone.podspec
```diff
@@ -3,7 +3,7 @@ require 'json'
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
 
 Pod::Spec.new do |s|
-  s.name         = 'FluentUI-React-Native-FocusZone'
+  s.name         = 'RCTFocusZone'
```
Saadnajmi, line 6: @chripear @chiuam I changed the podspecs' names to match the component they represent. This means FocusZone is an odd one out, the only one with the prefix "RCT". @HeyImChris has told me that RCT is reserved for React Native internal types, but we (FURN / rn-win32 ) have used it for FocusZone. How would you all feel about either:

1) Keeping the naming as is in this change, where the pod spec matches the component, and the internal objective C object is RCTFocusZone
2) Changing the podspec to "FRNFocusZone", and keeping the internal objective c object RCTFocusZone to keep binary compatibility
3) Changing both the podspec and the objective C type to "FRNFocusZone", introducing a divergence (win32 has a native module named "RCTFocusZone" while macOS has a native module "FRNFocusZone" and updating it downstream so that we use the consume the correct FRNFocusZone type?

Saadnajmi:
https://github.com/microsoft/fluentui-react-native/blob/364c4b60f9f6706a4b85d915b3d335fda99f56c2/packages/experimental/Expander/windows/ReactNativeExpander/ExpanderViewManager.cpp#L17

Windows copied our naming convention. Perhaps I should snuff it and change MSFExpander to FRNExpander while they only have one component?

HeyImChris (approved):
We'll need to find and notify the right client teams using these controls the next time we update devmain. I'd imagine this would break their usage of the controls?

Saadnajmi:
> We'll need to find and notify the right client teams using these controls the next time we update devmain. I'd imagine this would break their usage of the controls?

No, we just need to update the NuGet in devmain and fix fallout by renaming the libraries exported. Client teams should not be affected by this change as long as they have devmain in sync.

HeyImChris:
> > We'll need to find and notify the right client teams using these controls the next time we update devmain. I'd imagine this would break their usage of the controls?
> 
> No, we just need to update the NuGet in devmain and fix fallout by renaming the libraries exported. Client teams should not be affected by this change as long as they have devmain in sync.

Why won't client teams be impacted if, say, they're using `MSFButtonViewManager` somewhere in their JS but we renamed it?

Saadnajmi:
> > > We'll need to find and notify the right client teams using these controls the next time we update devmain. I'd imagine this would break their usage of the controls?
> > 
> > 
> > No, we just need to update the NuGet in devmain and fix fallout by renaming the libraries exported. Client teams should not be affected by this change as long as they have devmain in sync.
> 
> Why won't client teams be impacted if, say, they're using `MSFButtonViewManager` somewhere in their JS but we renamed it?

That class will never be accessed by clients in their JS afaik. It would _only_ be accessed by FURN components, so we would have to make sure the FURN NPM package and NuGet static libs stay in sync. It's not my favorite solution but we have to use NuGets instead of CocoaPods in devmain so 🤷🏾‍♂️

packages/components/FocusZone/RCTFocusZone.podspec
Saadnajmi: I'm gonna merge this and leave this as a followup question for expediency's sake. Sticking with 1)

PR #901, opened 2021-08-17
Enhancements for RadioGroup on macOS

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

After getting initial feedback from clients, I learned that it's quite cumbersome to add custom code to integrate the new package downstream. This change alleviates this issue by replacing the NativeRadioButton package with a macOS specific implementation that uses the same API. 

This change also adds more compatibility between RadioGroup and the native RadioButton:

- Rerouted native callback event to RadioGroup's onChange event (no more individual callbacks per radio button)
- Hooked up `defaultSelectedKey` prop in RadioGroup 

RadioButton Props supported on macOS:

- content
- buttonKey
- disabled

RadioGroup Props supported on macOS:

- label
- defaultSelectedKey
- selectedKey
- onChange
- IRadioGroupContext
  -  selectedKey
  - onChange

### Verification
Manually tested functionality in the test app, no visual changes.

https://user-images.githubusercontent.com/67026167/129824533-081680f3-b5d6-4956-b333-438f28fc0e19.mov

Ensured that `defaultSelectedKey` works as expected (option B is preselected) and RadioGroup's `onChange` is called.

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [X] Voiceover


packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -0,0 +1,56 @@
...
+  displayName: radioButtonName,
+  usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
+    const { content, buttonKey, disabled } = userProps;
+    const info = React.useContext(RadioGroupContext);
+    // Reroute the native component's onPress event to RadioGroup's onChange
+    const onPressRerouted = () => {
+      // Prevent calls to RadioGroup's onChange on the currently selected button
+      if (buttonKey != info.selectedKey) {
+        info.onChange && info.onChange(buttonKey);
+      }
```
Saadnajmi, line 22: The extra check is so that if the user clicks the same radio button 100 times, only one `oChange` callback is ever fired? Is that to match the win32 implementation?

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    const styleProps = useStyling(userProps);
+    const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
+      root: {
+        buttonKey: buttonKey,
+        content: content,
+        disabled: disabled,
+        onPress: onPressRerouted,
+        selected: info.selectedKey === buttonKey,
+        style: {
```
chiuam, line 33: For some reason style props aren't overridable from tests, same layout issue with menuButton? @Saadnajmi

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    const styleProps = useStyling(userProps);
+    const slotProps = mergeSettings<IRadioButtonSlotProps>(styleProps, {
+      root: {
+        buttonKey: buttonKey,
+        content: content,
+        disabled: disabled,
+        onPress: onPressRerouted,
+        selected: info.selectedKey === buttonKey,
```
Saadnajmi, line 32: nit, I would move the equality check to it's own line for readability, so that `const slotProps = ...` has less logic inside of it

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -0,0 +1,56 @@
...
+        style: {
+          // Fluent controls are designed to snap to a 4 px grid
+          marginLeft: 4,
+          marginTop: 4,
+          width: 150,
+          height: 20,
+        },
```
Saadnajmi, line 39: Do we actually need the margins?
Also 150 / 4 = 37.5, not a 4 px grid? The width is dependent on the label and will probably be overridden by client anyway so it's probably ok..

Saadnajmi: We should probably file a bug on investigating implementing RCTShadowView natively so we don't need to keep hardcoding these...

Saadnajmi (commented):
Can you also add a video / screenshot of the control working to the PR? Thanks!

packages/components/RadioGroup/src/RadioButton.macos.tsx
Saadnajmi: Do you mean passing it inline? Yeah, that doesn't work. I think you _have_ to call .customize() and pass in new settings...

apps/macos/src/Podfile.lock
```diff
@@ -490,6 +490,6 @@ SPEC CHECKSUMS:
   SwiftLint: 99f82d07b837b942dd563c668de129a03fc3fb52
   Yoga: 77d9987384e26b4aaf3e7a01808838cc2d4304f2
 
-PODFILE CHECKSUM: 15eda06aaa95b3f7091a3051b06322d719f25ff3
+PODFILE CHECKSUM: 2bf562c3be6637688b436c9b38272f364e9830f4
 
-COCOAPODS: 1.10.2
+COCOAPODS: 1.10.1
```
chripear, line 495: nit: Looks like you should update your cocoapods installation.

Saadnajmi: An aside on this: the version of cocoa pods from home-brew now works on m1 Macs without needing any rosetta aliasing! So I actually uninstalled the gem cocoa pods version and use the home-brew one instead now =)

packages/components/RadioGroup/src/RadioButton.macos.tsx
chiuam: The margins add paddings to the top and left of the radio button, I don't think it's relative to the button width? 
With 4px margins
image
Without margins
image

packages/components/RadioGroup/src/RadioButton.macos.tsx
chiuam: Correct, that's to match the win32 implementation. I believe that's the macOS behavior as well.

Saadnajmi (approved):

packages/components/RadioGroup/src/RadioButton.macos.tsx
chiuam: I filed a bug for the content size issue: https://office.visualstudio.com/APEX/_workitems/edit/5351152

apps/macos/src/Podfile.lock
chiuam: Fixed in commit 4378c5e17db2b6ca78558a9df1b64b040fe620a0

PR #903, opened 2021-08-17
Tabs Control for windows

dake3601:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Implemented Tabs Control for windows.

*Update:* I am having issues with Voiceover not announcing the selected TabsItem, I will follow up on this, but I am not sure if it will be fixed for this PR.

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +91,33 @@ export const Tabs = compose<TabsType>({
...
+        }
+        if (ev.nativeEvent.key === 'ArrowLeft') {
+          if (!(!isCircularNavigation && currTabItemIndex == 0)) {
+            newCurrTabItemIndex = (currTabItemIndex - 1 + length) % length;
+            state.context.selectedKey = state.context.enabledKeys[newCurrTabItemIndex];
+            data.onKeySelect(state.context.selectedKey);
+          }
+        }
+      }
+    };
```
Saadnajmi, line 114: Am I to understand this correctly and too that we basically implemented re-implemented focusZone with JS here?

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +91,33 @@ export const Tabs = compose<TabsType>({
...
+            data.onKeySelect(state.context.selectedKey);
+          }
+        }
+        if (ev.nativeEvent.key === 'ArrowLeft') {
+          if (!(!isCircularNavigation && currTabItemIndex == 0)) {
+            newCurrTabItemIndex = (currTabItemIndex - 1 + length) % length;
+            state.context.selectedKey = state.context.enabledKeys[newCurrTabItemIndex];
+            data.onKeySelect(state.context.selectedKey);
+          }
+        }
```
Saadnajmi, line 112: A gotcha is to make sure this does the right behavior in an RTL language, where the order of elements in a view are swapped

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +91,33 @@ export const Tabs = compose<TabsType>({
...
+    };
+
     const slotProps = mergeSettings<TabsSlotProps>(styleProps, {
-      root: { rest, ref: componentRef, accessibilityLabel: accessibilityLabel, accessibilityRole: 'tablist'},
+      root: { rest, ref: componentRef, accessibilityLabel: accessibilityLabel, accessibilityRole: Platform.OS !== 'windows' ? 'tablist' : null, ...pressable.props}, // Add windows role when RN is at >= 0.64
       label: { children: label },
-      container: { isCircularNavigation: isCircularNavigation, defaultTabbableElement: selectedTabsItemRef },
+      container: Platform.OS !== 'windows' ? { isCircularNavigation: isCircularNavigation, defaultTabbableElement: selectedTabsItemRef } : null,
+      stack: Platform.OS !== 'windows' ? null : {focusable: true, ref: focusZoneRef, onKeyDown: onKeyDown}
     });
```
Saadnajmi, line 121: why not make a Tabs.windows.ts? Or maybe a `windowsSlotProps` variable? It seems like there's a lot of platform checks that we could condense

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -137,8 +178,8 @@ export const Tabs = compose<TabsType>({
   slots: {
     root: View,
     label: Text,
-    container: FocusZone,
-    stack: { slotType: View, filter: filterViewProps },
+    container: Platform.OS !== 'windows' ? FocusZone : React.Fragment,
```
Saadnajmi, line 181: I'm curious if the non-focus zone implementation would "just work" for other platforms as well that have keyboard events as well, like macOS. 
Perhaps a better question: What is the functionality we change/lose in the windows version vs the UWP version?

packages/components/Tabs/src/TabsItem.settings.windows.ts
```diff
@@ -0,0 +1,79 @@
...
+      fontSize: 20,
+      borderWidth: 2,
+      borderColor: 'transparent',
+      borderRadius: 4,
+      indicatorColor: 'transparent',
+    },
+    root: {
+      accessible: true,
+      focusable: false,
+      // accessibilityRole: 'tab', // Add role when RN is at >= 0.64
```
Saadnajmi, line 19: why?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -43,6 +43,9 @@ export const TabsItem = compose<TabsItemType>({
     We check that there is an actual change in selection before forwarding the message to the user callback
     so that they aren't notified more than once for each click. */
     const changeSelection = () => {
+      if (Platform.OS === 'windows') {
+        info.focusZoneRef.current.focus();
+      }
```
Saadnajmi, line 48: I thought there is no FocusZone on windows so the ref would be null?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -64,7 +67,7 @@ export const TabsItem = compose<TabsItemType>({
       info: {
         ...pressable.state,
         selected: info.selectedKey === userProps.itemKey,
-        icon: !!icon,
+        icon: !!icon && Platform.OS !== 'windows',
```
Saadnajmi, line 70: perhaps add a comment + a bug number describing how windows doesn't support SVG icons yet? I assume that's why the check is here

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -139,7 +142,7 @@ export const TabsItem = compose<TabsItemType>({
   slots: {
     root: View,
     stack: { slotType: View, filter: filterViewProps },
-    icon: { slotType: Icon as React.ComponentType },
+    icon: Platform.OS !== 'windows' ? { slotType: Icon as React.ComponentType } : React.Fragment,
```
Saadnajmi, line 145: Why a Fragment?

packages/components/Tabs/src/TabsItem.settings.windows.ts
dake3601: There is a [bug ](https://github.com/microsoft/react-native-windows/pull/6728) with react-native-windows 0.63 that makes elements focusable if they have an accessibilityRole, and thus not allowing the expected behavior on UWP.

packages/components/Tabs/src/TabsItem.tsx
dake3601: It is the reference to the View (with JS keyboarding) that acts as a "focusZone". Should I change the name of the variable to something more clear?

Saadnajmi: What is the reference in the case of windows? Like, which slot? 
It kinda feels like we should just take the JS code you wrote to mimic a FocusZone, and put it in the FocusZone component as the windows implementation of FocusZone, and keep the logic inside Tabs simpler. I'm not sure how complicated of a change that is to implement though, especially since you all will finish your internship soon.

packages/components/Tabs/src/TabsItem.tsx
dake3601: Good idea! Yeah, that is the reason of the check.

packages/components/Tabs/src/TabsItem.tsx
dake3601: That check was added to make sure an icon is never rendered on UWP, but right now it seems unnecessary as it wont render due to a previous check on the component's state.

Will be removing the Fragment and changing the line back to:
```icon: { slotType: Icon as React.ComponentType },```

File: packages/components/Tabs/src/Tabs.tsx (hunk 1 above)
Comment (dake3601) [reply packages/components/Tabs/src/Tabs.tsx:114]: Yup, with a combination of setting components to focusable and this function, the needed behavior of the FocusZone for this Control is implemented. This is due to a lack of FocusZone on UWP.

File: packages/components/Tabs/src/TabsItem.tsx (hunk 7 above)
Comment (dake3601) [reply packages/components/Tabs/src/TabsItem.tsx:70]: I was not able to find a bug number, but I found the [Status](https://github.com/microsoft/fluentui-react-native/wiki/Status-of-Controls) of the Control. Currently, Icons are on backlog for windows.

File: packages/components/Tabs/src/TabsItem.settings.windows.ts (hunk 5 above)
Comment (Saadnajmi) [reply packages/components/Tabs/src/TabsItem.settings.windows.ts:19]: Thanks! That is very helpful to know =)

File: packages/components/Tabs/src/TabsItem.tsx (hunk 7 above)
Comment (Saadnajmi) [reply packages/components/Tabs/src/TabsItem.tsx:70]: I meant to open a bug on our GitHub issue tracker, sorry. Should have been clear.

File: packages/components/Tabs/src/Tabs.tsx (hunk 3 above)
Comment (dake3601) [reply packages/components/Tabs/src/Tabs.tsx:121]: Since UWP and win32 have a lot similarities, I thought that having one file for both would be more convenient. I will add the windowsSlotProps variable.

File: packages/components/Tabs/src/TabsItem.tsx (hunk 6 above)
Comment (dake3601) [reply packages/components/Tabs/src/TabsItem.tsx:48]: It is a reference to the stack slot.

Comment (dake3601) [reply packages/components/Tabs/src/TabsItem.tsx:48]: It might be a bit complicated since the JS code that mimics the FocusZone is ingrained to the Tabs Control. But it is an interesting proposition to consider.

File: packages/components/Tabs/src/TabsItem.tsx (hunk 7 above)
Comment (dake3601) [reply packages/components/Tabs/src/TabsItem.tsx]: Since there is already an Icon Control for the other platforms, would it be a "Feature Request" or a "New Component"?

File: packages/components/Tabs/src/Tabs.tsx (hunk 4 above)
Comment (dake3601) [reply packages/components/Tabs/src/Tabs.tsx:190]: Hmm, the main difference between the windows version and the UWP version is that on windows the keyboard focus hops from header (tabsItem) to header, while on the UWP version the focus stays on the stack when keyboarding.

File: packages/components/Tabs/src/Tabs.tsx (hunk 2 above)
Comment (dake3601) [reply packages/components/Tabs/src/Tabs.tsx:112]: Do you know if there is a way to determine if RTL is being used on UWP? Or someone that could know more about this?

File: packages/components/Tabs/src/TabsItem.tsx (hunk 7 above)
Comment (dake3601) [reply packages/components/Tabs/src/TabsItem.tsx]: Bug created: #935

File: packages/components/Tabs/src/Tabs.tsx [hunk 9]
```diff
@@ -10,7 +10,8 @@ import { settings } from './Tabs.settings';
 import { mergeSettings } from '@uifabricshared/foundation-settings';
 import { filterViewProps } from '@fluentui-react-native/adapters';
 import { foregroundColorTokens, textTokens, backgroundColorTokens } from '@fluentui-react-native/tokens';
-import { useSelectedKey } from '@fluentui-react-native/interactive-hooks';
+import { useSelectedKey, useAsPressable } from '@fluentui-react-native/interactive-hooks';
+import type { IKeyboardEvent } from '@office-iss/react-native-win32';
```
Saadnajmi, line 14: Nit: importing something from react-native-win32 in the platform agnostic tabs file means we either don't get the import on other platforms, or now other platforms like macOS / windows depend on react-native-win32. If possible, I would move this to a `Tabs.win32.tsx` specific file. It looks like you're using the type for the event handler at line 94, which I think you might also be fine to leave as the `any` type (I've seen other event handlers do that for better or worse).

packages/components/Tabs/src/Tabs.tsx
Saadnajmi: import {I18nManager} from "react-native";
if (I18nManager.isRTL) {

packages/components/Tabs/src/Tabs.settings.windows.ts
```diff
@@ -12,17 +12,18 @@ export const settings: IComposeSettings<TabsType> = [
...
       style: {
         flexDirection: 'row',
-        marginLeft: -14,
         marginTop: 6,
       },
     },
+    label: {
+      style: {
+        marginLeft: 10,
+      },
```
Saadnajmi, line 25: nit: marginStart instead of marginLeft so it flips with RTL

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +90,42 @@ export const Tabs = compose<TabsType>({
...
+          }
+        }
+      }
+    };
+
+    if (Platform.OS == 'windows') {
+      const slotProps = mergeSettings<TabsSlotProps>(styleProps, {
+        root: { ref: componentRef, accessibilityLabel: accessibilityLabel, accessibilityRole: 'tablist', ...pressable.props, ...rest},
+        label: { children: label },
+        stack: { focusable: true, ref: focusZoneRef, onKeyDown: onKeyDown},
```
Saadnajmi, line 119: I would move the onKeyDown function definition inside the platform.os check if it's only for UWP windows.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -66,13 +69,18 @@ export const Tabs = compose<TabsType>({
     // Stores views to be displayed
     const map = new Map<string, React.ReactNode[]>();
 
+    const pressable = useAsPressable({
+      ...rest,
+    });
+
```
Saadnajmi, line 75: Why is this only used by windows? as far as I can tell? Would this not be used by the other platforms too?

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +90,43 @@ export const Tabs = compose<TabsType>({
...
+            if (isCircularNavigation || !(currTabItemIndex == 0)) {
+              newCurrTabItemIndex = (currTabItemIndex - 1 + length) % length;
+              state.context.selectedKey = state.context.enabledKeys[newCurrTabItemIndex];
+              data.onKeySelect(state.context.selectedKey);
+            }
+          }
+        }
+      };
+
+      const slotProps = mergeSettings<TabsSlotProps>(styleProps, {
```
Saadnajmi, line 117: nit, I would still name this `windowsSlotProps` or something similar for clarity, along with a comment on how we don't have FocusZone there so we need the extra focusZoneRef / onKeyDown callback, along with maybe a bug tracking bringing FocusZone to UWP windows.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -137,8 +187,8 @@ export const Tabs = compose<TabsType>({
   slots: {
     root: View,
     label: Text,
-    container: FocusZone,
-    stack: { slotType: View, filter: filterViewProps },
+    container: Platform.OS !== 'windows' ? FocusZone : React.Fragment,
+    stack: View,
```
Saadnajmi, line 191: Curious, why did you remove the filterViewProps?

packages/components/Tabs/src/Tabs.tsx
Saadnajmi: I see. That feels like a side effect of not having FocusZone. I would still add a comment with a GitHub issue number tracking adding FocusZone for windows.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -64,7 +67,7 @@ export const TabsItem = compose<TabsItemType>({
       info: {
         ...pressable.state,
         selected: info.selectedKey === userProps.itemKey,
-        icon: !!icon,
+        icon: !!icon && Platform.OS !== 'windows', // Icons are on backlog for windows
```
Saadnajmi, line 70: Maybe add a GitHub issue number here as well tracking icon support for windows.

packages/components/Tabs/src/Tabs.tsx
dake3601: It is only used by windows as it was needed to implement the keyboarding behavior. For the spec'd behavior, I believe that the other platforms do not need it.  I will be moving it under the windows platform check to make it clear it is only for windows.

packages/components/Tabs/src/Tabs.tsx
dake3601: I tried renaming it to windowsSlotProps, but the tester app crashes if that variable is not called slotProps. I will add the comment on the props with a bug.

packages/components/Tabs/src/Tabs.tsx
dake3601: Because I needed the componentRef of the stack to focus on it and the filter removed that prop.

packages/components/Tabs/src/Tabs.tsx
Saadnajmi: oh, interesting. I didn't know about that thanks!

packages/components/Tabs/src/Tabs.tsx
Saadnajmi: super nit: "GH #964" instead of "GH 964" to match convention elsewhere and also make it more obvious it's a bug number

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -34,10 +34,20 @@ export interface TabsContextData {
...
+
   /**
    * A Map to store the corresponding view of each TabsItem
    */
   views?: Map<string, React.ReactNode[]> | null;
+
+  /**
+   * Reference to the Focus Container
+   */
+  focusZoneRef: React.RefObject<any> | null;
```
Saadnajmi, line 50: I would update the comment here as well to say this is because we don't have focus zone on windows

packages/components/Tabs/src/TabsItem.tsx
Saadnajmi: We talked offline, and it sounds like we're not trying to make a JS FocusZone for windows with this review. I would leave a comment with the GH issue number for this piece of code as well. Sorry for being pedantic, I tend to recommend over-documenting rather than under-documenting.

Saadnajmi (approved):
Left a bunch of comments about documentation, but looks good to me! I will definitely say I am not a windows engineer so I would like someone on the windows team to also review / approve this.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -64,7 +67,7 @@ export const TabsItem = compose<TabsItemType>({
       info: {
         ...pressable.state,
         selected: info.selectedKey === userProps.itemKey,
-        icon: !!icon,
+        icon: !!icon && Platform.OS !== 'windows', // Icons are on backlog for windows, GH 935
```
Saadnajmi, line 70: Update the other GH issue comments as well.  Also put the GH issue first, I.E
> GH #935, Icons are on backlog for windows
Not sure I have a good reason for putting the issue first except that it's the first thing the reader sees then, and is convention I've seen elsewhere.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +86,48 @@ export const Tabs = compose<TabsType>({
 
     const styleProps = useStyling(userProps, (override: string) => state[override] || userProps[override]);
 
+    if (Platform.OS == 'windows') {
```
samuelfreiberg, line 89: I'm wondering if this should just be a separate file completely - Tabs.windows.tsx. 

This is a lot of platform-specific code, and it makes this file a lot more complicated that it has to be. If there was just a line or two difference, it wouldn't matter. But this is a lot of Windows-specific logic. 

@ksiler @Saadnajmi @PPatBoyd Thoughts?

samuelfreiberg: Since we have Tabs.settings.**windows**.ts, I think it would make sense to split this into a .windows file

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -82,10 +86,48 @@ export const Tabs = compose<TabsType>({
...
+        ...rest,
+      });
+
+      const onKeyDown = (ev: any) => {
+        if (ev.nativeEvent.key === 'ArrowRight' || ev.nativeEvent.key === 'ArrowLeft') {
+          const length = state.context.enabledKeys.length;
+          const currTabItemIndex = state.context.enabledKeys.findIndex(x => x == state.context.selectedKey)
+          let newCurrTabItemIndex;
+          if (ev.nativeEvent.key === 'ArrowRight') {
+            if (isCircularNavigation || !(currTabItemIndex + 1 == length)) {
```
samuelfreiberg, line 100: Isn't the only case you need to handle `if (isCircularNavigation && currTabItemIndex+1==length)`?

Keyboarding for all other scenarios should already be handled by the underlying platform, correct? I would think keyboarding between the Tab Items would work already, just not the circularNav feature

samuelfreiberg (changes requested):

packages/components/Tabs/src/Tabs.tsx
Saadnajmi: Yeah, with us not splitting out the FocusZone, I think that makes sense.

Saadnajmi: FWIW, I think we pushed most of the platform specifics to the windows specific slot props here, and have well documented where else we have platform.os checks

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -113,7 +113,7 @@ export const TabsItem = compose<TabsItemType>({
         accessibilityPositionInSet: accessibilityPositionInSet ?? info.tabsItemKeys.findIndex(x => x == itemKey) + 1,
         accessibilitySetSize: accessibilitySetSize ?? info.tabsItemKeys.length,
         onAccessibilityAction: onAccessibilityAction,
-        focusable: Platform.OS === 'macos' ? !userProps.disabled : userProps.focusable ?? true,
+        focusable: Platform.OS === 'macos' ? !userProps.disabled : userProps.focusable ?? Platform.OS === 'windows' ? false : true,
```
Saadnajmi, line 116: You could use platform.select here to simplify this.

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,188 @@
...
+    return;
+  },
+  getTabId: (/* key:string, index: number*/) => {
+    return null
+  },
+  updateSelectedTabsItemRef: (/* ref: React.RefObject<any>*/) => {
+    return;
+  },
+  tabsItemKeys: [],
+  enabledKeys: [],
```
rurikoaraki, line 26: Two thoughts:

1. It doesn't seem like `enabledKeys` would be something that a TabItem/child of Tabs would be interested in knowing about. I guess TabsContext is starting to be used like a State here, but unless there's a reason to conflate the two I think it'd be better to keep state separate from things that child components need to know from an encapsulation standpoint.
2. To me it feels more complicated than necessary to store the enabled keys as a separate array. Since we have tabsItemKeys here, I think we could use that to determine which key to select when arrow keys are pressed by modifying the logic you already have to skip over disabled keys instead.

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    const onChangeTabId = React.useCallback((key: string, index: number) => {
+      if (getTabId) {
+        return getTabId(key, index);
+      }
+      return `${key}-Tab${index}`;
+    }, [getTabId]);
+
+    // Stores views to be displayed
+    const map = new Map<string, React.ReactNode[]>();
```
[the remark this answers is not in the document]
dake3601: On TabsItem.tsx, line 129:
```    context.views.set(info.key, children);```

File: packages/components/Tabs/src/Tabs.windows.tsx (hunk 21 above)
Comment (dake3601) [reply packages/components/Tabs/src/Tabs.windows.tsx:26]: 1. Agree, I could move it over to state.info to keep it on the state of Tabs, but out of TabsItems and the context.

2. That is something I had tried, but for that to work, the keyboarding logic would get more complicated and Tabs would need to know which TabsItems are disabled. For Tabs to know which TabsItems are disabled, the ideas that I had were that tabsItemKeys could be made up with objects with keys and disabled status or there could be another array with disabled status. Any recommendation on how to approach this?

Also, tabsItemKeys could not be replaced by enabledKeys, as tabsItemKeys is used for accessibility.

File: packages/components/Tabs/src/Tabs.windows.tsx (hunk 22 above)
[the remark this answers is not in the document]
Comment (rurikoaraki) [reply packages/components/Tabs/src/Tabs.windows.tsx:68]: Ah thanks. Missed it since it wasn't part of the changes.

File: packages/components/Tabs/src/Tabs.windows.tsx (hunk 21 above)
Comment (rurikoaraki) [reply packages/components/Tabs/src/Tabs.windows.tsx]: Ah, `tabsItemKeys` is only storing keys, right.

Since the current logic is pretty tied to having the selection tracker being a string/key instead of an index, switching to keeping the index to track the current selection instead seems like not a great idea. Given that, I'd probably lean towards having an array of tuples (`[string, boolean][]`) like you mentioned above, but the current logic also works and I think it's okay if you decide you'd rather keep it the way it is.

File: packages/components/Tabs/src/Tabs.windows.tsx [hunk 23]
```diff
@@ -0,0 +1,187 @@
...
+      getTabId,
+      componentRef = defaultComponentRef,
+      isCircularNavigation,
+      ...rest
+    } = userProps;
+
+    // This hook updates the Selected TabsItem and calls the customer's onTabsClick function. This gets called after a TabsItem is pressed.
+    const data = useSelectedKey(selectedKey || defaultSelectedKey || null, userProps.onTabsClick);
+
+    const [, setSelectedTabsItemRef] = React.useState(React.useRef<View>(null));
```
samuelfreiberg, line 51: `[, setSelectedTabsItemRef]` Was this a mistake? Or is this some syntax I've never seen before?

Saadnajmi: Related, I think it would look cleaner if the useRef hook wasn't inline like that. Especially for hooks, I think it's important to keep them top level and not nested to be sure they aren't called conditionally and such.

dake3601: It was on purpose, it omits the first element and just assigns the second one. This is because selectedTabsItemRef is not needed for the windows implementation.

samuelfreiberg: What do you mean by the second one? If `selectedTabsItemRef ` is not needed for the Windows implementation, does this provide any value?

dake3601: It does not, will be removing the hook and references to it.

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -38,11 +38,20 @@ export interface TabsContextData {
    * A Map to store the corresponding view of each TabsItem
    */
   views?: Map<string, React.ReactNode[]> | null;
+
+  /**
+   * GH #964, Reference to the Focus Container as there is no FocusZone on windows.
+   */
+  focusZoneRef?: React.RefObject<any> | null;
```
samuelfreiberg, line 45: I would add another line to the comments, something like: "Windows-Specific Prop`

packages/components/Tabs/src/Tabs.types.ts
```diff
@@ -38,11 +38,20 @@ export interface TabsContextData {
...
 }
 
 export interface TabsInfo {
   headersOnly?: boolean;
   label?: boolean,
+  /**
+   * Array of enabled keys in the group
```
samuelfreiberg, line 52: I would add another line to the comments, something like: "Windows-Specific Prop`

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,177 @@
...
+      getTabId,
+      componentRef = defaultComponentRef,
+      isCircularNavigation,
+      ...rest
+    } = userProps;
+
+    // This hook updates the Selected TabsItem and calls the customer's onTabsClick function. This gets called after a TabsItem is pressed.
+    const data = useSelectedKey(selectedKey || defaultSelectedKey || null, userProps.onTabsClick);
+
+    const onChangeTabId = React.useCallback((key: string, index: number) => {
```
samuelfreiberg, line 51: I'm not sure the name of this function is intuitive. The function is to get the tab id, right? The function name sounds like your changing the tabID. I'd change it to `findTabId` or something similar

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,177 @@
...
+      root: { ref: componentRef, accessibilityLabel: accessibilityLabel, accessibilityRole: 'tablist', ...pressable.props, ...rest},
+      label: { children: label },
+      stack: { focusable: true, ref: focusZoneRef, onKeyDown: onKeyDown},
+    });
+
+    return { slotProps, state };
+  },
+
+  render: (Slots: ISlots<TabsSlotProps>, renderData: TabsRenderData, ...children: React.ReactNode[]) => {
+    if (renderData.state == undefined) {
```
samuelfreiberg, line 115: `if (!renderData.state)`

packages/components/Tabs/src/Tabs.windows.tsx
dake3601: Yeah, making the change to ```findTabId``` on Tabs and Tabs.windows.

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,177 @@
...
+            return child.props.itemKey;
+          }
+        }
+      });
+    }
+
+    return (
+      <TabsContext.Provider
+        // Passes in the selected key and a hook function to update the newly selected tabsItem and call the client's onTabsClick callback
+        value={renderData.state.context}
```
samuelfreiberg, line 146: `renderData.state?.context` You check for `!renderData.state` above, but that doesn't mean renderData.state.context != nullptr

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,177 @@
...
+      >
+        <Slots.root>
+          {renderData.state.info.label && <Slots.label />}
```
samuelfreiberg, line 149: `renderData.state?.info?.label`

packages/components/Tabs/src/Tabs.windows.tsx
```diff
@@ -0,0 +1,177 @@
...
+      },
+      info: {
+        headersOnly: headersOnly ?? false,
+        label: !!label,
+      },
+    };
+
+    const styleProps = useStyling(userProps, (override: string) => state[override] || userProps[override]);
+
+    const pressable = useAsPressable({
```
samuelfreiberg, line 77: Nit: Make this into one line

samuelfreiberg (approved):

Issue #728, opened 2021-06-15
Support PlatformColor for Icon


Saadnajmi:
#### Platform
- [x] iOS
- [x] macOS
- [x] win32
- [x] windows
- [x] android
 
#### Describe the feature that you would like added

RN-SVG doesn't support RN 0.63 / the new PlatformColor API yet. This blocks us from creating any `@fluentui-react-native/icon` using a PlatformColor. This is a divergence from the rest of our color tokens that we would like to resolve. To do so, we'll need to update the RN-SVG repo

#### What component or utility would this be added to

Icon

#### Have you discussed this feature with our team, and if so, who

Harrie Shin, Ruriko Arakai

PPatBoyd:
@saadnajmi any update on rn-svg status, if we have a PR for them / it's in progress?

cross-linking issue: https://github.com/react-native-svg/react-native-svg/issues/1567

Saadnajmi:
I'll tag @ejlayne who probably best knows how it is prioritized. I don't have new information to share

rurikoaraki:
I think this is fixed, so closing

Saadnajmi:
Still open I think, talking with @lyzhan7 again

lyzhan7:
The error we get on iOS is "JSON value of type NSString cannot be converted to a UIColor"
This is from the Notification component, which passes DynamicColorIOS as token values that get used on some icons

Saadnajmi:
Any chance we could "exclude" windows from auto linking rn-svg so we can update the package and unblock this?

rurikoaraki:
@lawrencewin who is working on #2252

Saadnajmi:
@lyzhan7 FYI this should be unblocked

lyzhan7:
Have confirmed that the PR mentioned above fixes the iOS Notification error that was popping up

PR #906, opened 2021-08-18
Fix iconColor not applying to svg icons

### LINKED ISSUE #888: Icon color not working correctly

rurikoaraki:
Fixes #888 

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

When I made my ColorValue changes, I split out the color prop for the SVG icon into a separate prop, since RN SVG is still on an older version of RN and doesn't support ColorValue yet, only taking string. However I did this without really understanding how the token/slot/styling system worked and broke things for other components.

This change removes the color prop on the SvgIconProps and makes the svg icon use the color prop on IconProps as its base instead. I implemented a downgradeColor function which falls back to an appropriate color if the color is not a string.

I'm open to suggestions on better ways to gracefully handle the downgrade, but this won't really be fixed until RN SVG can handle ColorValues.

### Verification

| image | image |

These are actually both supposed to be red, but the customize for the iconColor token isn't working on the Button. Still looking into that.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Icon/src/Icon.tsx
+++ b/packages/experimental/Icon/src/Icon.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { IconProps, SvgIconProps, FontIconProps } from './Icon.types';
-import { Image, ImageStyle, Platform, View } from 'react-native';
+import { Image, ImageStyle, Platform, View, ColorValue } from 'react-native';
 import { Text } from '@fluentui-react-native/text';
 import { mergeStyles, useFluentTheme } from '@fluentui-react-native/framework';
 import { stagedComponent, mergeProps, getMemoCache } from '@fluentui-react-native/framework';
@@ -51,18 +51,13 @@ function renderFontIcon(iconProps: IconProps) {
 
 function renderSvg(iconProps: IconProps) {
   const svgIconProps: SvgIconProps = iconProps.svgSource;
-  const { width, height } = iconProps;
+  const { width, height, color } = iconProps;
   const viewBox = iconProps.svgSource.viewBox;
-  const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width: width, height: height }, [width, height])[0]);
+  const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width, height }, [width, height])[0]);
 
   // react-native-svg is still on 0.61, and their color prop doesn't handle ColorValue
   // If a color for the icon is not supplied, fall back to white or black depending on appearance
-  const theme = useFluentTheme();
-  const iconColor = svgIconProps.color
-    ? svgIconProps.color
-    : getCurrentAppearance(theme.host.appearance, 'light') === 'dark'
-    ? '#FFFFFF'
-    : '#000000';
+  const iconColor = downgradeColor(color);
 
   if (svgIconProps.src) {
     return (
@@ -106,3 +101,11 @@ export const Icon = stagedComponent((props: IconProps) => {
 });
 
 export default Icon;
+
+function downgradeColor(color: ColorValue): string {
+  if (typeof color === 'string') {
+    return color as string;
+  }
+
+  return getCurrentAppearance(useFluentTheme().host.appearance, 'light') === 'dark' ? '#FFFFFF' : '#000000';
+}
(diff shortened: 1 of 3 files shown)


packages/experimental/Icon/src/Icon.tsx
```diff
@@ -106,3 +101,11 @@ export const Icon = stagedComponent((props: IconProps) => {
...
 
 export default Icon;
+
+function downgradeColor(color: ColorValue): string {
+  if (typeof color === 'string') {
+    return color as string;
+  }
+
+  return getCurrentAppearance(useFluentTheme().host.appearance, 'light') === 'dark' ? '#FFFFFF' : '#000000';
+}
```
Saadnajmi, line 111: Maybe add a comment linking to our issue #728 ?

Saadnajmi (approved):

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -51,18 +51,13 @@ function renderFontIcon(iconProps: IconProps) {
 
 function renderSvg(iconProps: IconProps) {
   const svgIconProps: SvgIconProps = iconProps.svgSource;
-  const { width, height } = iconProps;
+  const { width, height, color } = iconProps;
```
ValentinaKozlova, line 54: I'll need to revert my changes because I also fixed it but in a different way :)

ValentinaKozlova (approved):

Saadnajmi:
Would this close https://github.com/microsoft/fluentui-react-native/issues/888 ?
Assuing we're tracking the ColorValue in SVG issue separately in #728

packages/experimental/Icon/src/Icon.tsx
rurikoaraki: Ah sorry, I missed that 😞

rurikoaraki:
@Saadnajmi Yeah it would close #888. I'll edit the comments.

packages/experimental/Icon/src/Icon.tsx
rurikoaraki: Done.

packages/experimental/Icon/src/Icon.tsx
ValentinaKozlova: It's ok ;)

PR #912, opened 2021-08-18
Tabs control for macos

sriya112:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Implemented the Tabs control for macos.

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/components/Tabs/src/TabsItem.settings.macos.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { IViewProps } from '@fluentui-react-native/adapters';
+
+export const tabsItemSelectActionLabel = 'Select a TabsItem';
+
+export const settings: IComposeSettings<TabsItemType> = [
+  {
+    tokens: {
+      color: 'bodyText',
+      variant: 'subheaderStandard',
+      fontSize: 13,
```
Saadnajmi, line 17: Do you have to specify font size? Is it not included with the variant?

packages/components/Tabs/src/TabsItem.settings.macos.ts
```diff
@@ -0,0 +1,87 @@
...
+        tokens: {
+          color: 'buttonTextHovered',
+          fontWeight: 'bold',
+        },
+        _overrides: {
+          selected: {
+            indicator: {
+              style:{
+                marginHorizontal: -1,
+              },
```
Saadnajmi, line 72: -1?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -70,7 +70,7 @@ export const TabsItem = compose<TabsItemType>({
       },
     };
 
-    const buttonRef = useViewCommandFocus(componentRef);
+    const buttonRef = Platform.OS !== 'macos' ?  useViewCommandFocus(componentRef) : componentRef;
```
Saadnajmi, line 73: I think we have an open bug to make UseViewCommandFocus work on macOS. Once we do so, I think you can undo the platform check here

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -110,6 +110,7 @@ export const TabsItem = compose<TabsItemType>({
         accessibilityPositionInSet: accessibilityPosInSet ?? info.tabsItemKeys.findIndex(x => x == itemKey) + 1,
         accessibilitySetSize: accessibilitySetSize ?? info.tabsItemKeys.length,
         onAccessibilityAction: onAccessibilityAction,
+        focusable: Platform.OS=='macos' ? !userProps.disabled : userProps.focusable,
```
Saadnajmi, line 113: Does focus not get disabled when disabled is set?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -56,7 +56,7 @@ export const TabsItem = compose<TabsItemType>({
     const pressable = useAsPressable({
       ...rest,
       onPress: changeSelectionWithFocus,
-      onFocus: changeSelection,
+      onFocus: Platform.OS == 'macos' ? null : changeSelection,
```
Saadnajmi, line 59: why null?

Saadnajmi:
A note: One bug we haven't solved yet is the prop "enableFocusRing" needs to be set to true for the system focus ring to show up on a control. None of FluentTester sets it, and we haven't yet found a way for it to be on by default. As a result, it's almost impossible to tell if something has focus, which is really unfortunate. Apologies if that has caused any confusion while you all worked on stuff!

packages/components/Tabs/src/TabsItem.tsx
Saadnajmi: But also, what does useViewCommandFocus do here?

packages/components/Tabs/src/TabsItem.settings.macos.ts
sriya112: It is included in the variant, but there is no variant that has a fontSize of 13 which is what is specified in the Figma design.

packages/components/Tabs/src/TabsItem.tsx
Saadnajmi: As of #919 , useViewCommandFocus is implemented for macOS. Try removing this platform check and let me know if it works! =)

packages/components/Tabs/src/Tabs.settings.macos.ts
```diff
@@ -0,0 +1,31 @@
...
+  {
+    tokens: {
+      color: 'menuItemText',
+      variant: 'bodySemibold',
+      fontWeight: 'bold',
+      fontSize: 14,
+    },
+    root: {
+      accessible: true,
+      // accessibilityRole: 'tablist', // Add role when RN is at >= 0.64
```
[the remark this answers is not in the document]
sriya112: You're right this is unnecessary here, it's removed now.

packages/components/Tabs/src/TabsItem.tsx
sriya112: It works now, thank you!

packages/components/Tabs/src/TabsItem.settings.macos.ts
rurikoaraki: nit: I would probably push back on design and tell them that 13 isn't part of the set of pre-determined typography, and get a new size. If macOS has different font sizes than other platforms and we just haven't gotten around to updating things on our end, that's fine, but in general we should be avoiding using non-standard sizes

Saadnajmi: For reference, I tried to follow the appe Human Interface Guidelines for typography when I made the apple theme and the typography variants https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/

Definitely not a blocker

packages/components/Tabs/src/TabsItem.tsx
sriya112: This makes it so that a disabled TabsItem cannot be focusable

Saadnajmi: Did you not need to do this for win32 or windows? If so, that sounds like it could be a bug we open on react-native-macos

packages/components/Tabs/src/TabsItem.tsx
sriya112: updated so there's no null now

packages/components/Tabs/src/TabsItem.tsx
sriya112: update: useViewCommandFocus doesn't seem to be working when tested again, talked offline about this and will be filing a bug

packages/components/Tabs/src/Tabs.settings.macos.ts
```diff
@@ -0,0 +1,31 @@
...
+      style: {
+        marginLeft: 7,
+      },
```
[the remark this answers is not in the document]
sriya112: Updated to 4

packages/components/Tabs/src/TabsItem.settings.macos.ts
sriya112: this makes the indicator grow on hover

sriya112: this part is supposed to work with animation, but animation is a work in progress right now

packages/components/Tabs/src/TabsItem.settings.macos.ts
rurikoaraki: @ksiler @sriya112 sounds good to me! Thanks.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -110,6 +110,7 @@ export const TabsItem = compose<TabsItemType>({
         accessibilityPositionInSet: accessibilityPosInSet ?? info.tabsItemKeys.findIndex(x => x == itemKey) + 1,
         accessibilitySetSize: accessibilitySetSize ?? info.tabsItemKeys.length,
         onAccessibilityAction: onAccessibilityAction,
+        focusable: Platform.OS=='macos' ? !userProps.disabled : userProps.focusable ?? true,
```
ksiler, line 113: nit: add space around "=="

packages/components/Tabs/src/Tabs.settings.macos.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    root: {
+      accessible: true,
+      style: {
+        marginLeft: 4,
+      },
+    },
+    label: {
+      style: {},
+    },
```
ksiler, line 25: Is an entry for label needed here if no styles are being applied?

packages/components/Tabs/src/Tabs.settings.macos.ts
```diff
@@ -0,0 +1,35 @@
...
+    stack: {
+      style: {
+        // marginLeft: -10,
```
ksiler, line 28: What's this for? I would remove commented styles if they're not being used.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -70,7 +70,7 @@ export const TabsItem = compose<TabsItemType>({
       },
     };
 
-    const buttonRef = useViewCommandFocus(componentRef);
+    const buttonRef = Platform.OS == 'macos' ? componentRef : useViewCommandFocus(componentRef);
```
Saadnajmi, line 73: nit: I think this should also be `===`
Also @chiuam fyi since she's investigating ref issues on macOS

Saadnajmi (approved):
Looks good, thanks for all the work! Be sure to file the bugs you found so we can look into it after you all have left =)

packages/components/Tabs/src/TabsItem.tsx
sriya112: fixed!

sriya112: here's the bug I filed for this: https://office.visualstudio.com/OC/_workitems/edit/5382203

packages/components/Tabs/src/TabsItem.tsx
sriya112: here's the bug I filed for this: https://office.visualstudio.com/OC/_workitems/edit/5382366

PR #919, opened 2021-08-19
Bring useViewCommandFocus to macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There's a hook called useViewCommandFocus a bunch of our components use that is only implemented for win32, and is a no-op on every other platform. Let's add it to macOS. I just copied the win32 version and changed "IViewWIn32" to "View"

Note that I think this hook doesn't need to exist, every standard react native ref has a `.focus()` method on it, so I don't know what functionality this adds.

### Verification

Not sure how to test this, we still don't have `enableFocusRing` set on most of our views =/ 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-interactive-hooks-126dc43e-9e24-47a7-9585-d348a939669a.json
+++ b/change/@fluentui-react-native-interactive-hooks-126dc43e-9e24-47a7-9585-d348a939669a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add macOS implementation of useViewCommandFocus",
+  "packageName": "@fluentui-react-native/interactive-hooks",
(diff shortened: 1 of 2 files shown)


chripear (commented):
Recommend testing with RadioGroup on macOS.  See if it affects the tab and arrow key behavior in any way.

chripear (approved):

Saadnajmi:
> Recommend testing with RadioGroup on macOS. See if it affects the tab and arrow key behavior in any way.

Just did. Looks the same to me. Let me know if you see any regressions

PR #922, opened 2021-08-19
RadioGroup (macOS): Focus selected button when tabbing into group

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
When tabbing into a RadioGroup, currently we focus on the closest radio button.  However, we want to emulate the native behavior in which we focus on the selected button (if there is one).  This change adds a special property to the FocusZone attached to a RadioGroup, which triggers some special behavior to find and focus a selected button.

### Verification
Used RadioGroup test pane in FluentTester app.

| Before | After |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
I think CXE solved this for win32 solved this by adding the method `updateSelectedButtonRef` to RadioGroupContext.
The PR is here: https://github.com/microsoft/fluentui-react-native/pull/680

I think the way it works is every time a RadioButton is selected, it can call `updateSelectedButtonRef` to tell RadioGroup "remember me as the `defaultTabbableElement`". If we have `defaultTabbableElement` is implemented for macOS, then we shouldn't need this new property?
^ Not that method might need the ref on each button to not be null, aka `useViewCommandFocus` to be implemented for macOS, aka #919 to be merged. 

How does that approach sound? Am I missing any context from investigations you have done?

chripear:
> I think CXE solved this for win32 solved this by adding the method `updateSelectedButtonRef` to RadioGroupContext.
> The PR is here: #680
> 
> I think the way it works is every time a RadioButton is selected, it can call `updateSelectedButtonRef` to tell RadioGroup "remember me as the `defaultTabbableElement`". If we have `defaultTabbableElement` is implemented for macOS, then we shouldn't need this new property?
> ^ Not that method might need the ref on each button to not be null, aka `useViewCommandFocus` to be implemented for macOS, aka #919 to be merged.
> 
> How does that approach sound? Am I missing any context from investigations you have done?

In my previous testing on win32, focus zones didn't seem to remember the last focused item in them.  I'll have to take another look.

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -6,26 +6,41 @@ import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
...
 const NativeRadioButtonView = ensureNativeComponent('FRNRadioButtonView');
 
 export const RadioButton = compose<IRadioButtonType>({
   displayName: radioButtonName,
   usePrepareProps: (userProps: IRadioButtonProps, useStyling: IUseComposeStyling<IRadioButtonType>) => {
-    const { content, buttonKey, disabled } = userProps;
+    const defaultComponentRef = React.useRef(null);
+    const { content, buttonKey, disabled, componentRef = defaultComponentRef } = userProps;
     const info = React.useContext(RadioGroupContext);
+    const buttonRef = useViewCommandFocus(componentRef);
```
Saadnajmi, line 19: @chiuam were talking offline about this and we're pretty sure this makes `buttonRef` null, even with my changes :/ You could try just doing `const buttonRef = componentRef` for the macOS version.

chripear:
> I think CXE solved this for win32 solved this by adding the method `updateSelectedButtonRef` to RadioGroupContext.
> The PR is here: #680
> 
> I think the way it works is every time a RadioButton is selected, it can call `updateSelectedButtonRef` to tell RadioGroup "remember me as the `defaultTabbableElement`". If we have `defaultTabbableElement` is implemented for macOS, then we shouldn't need this new property?
> ^ Not that method might need the ref on each button to not be null, aka `useViewCommandFocus` to be implemented for macOS, aka #919 to be merged.
> 
> How does that approach sound? Am I missing any context from investigations you have done?

Implemented `defaultTabbableElement` for macOS.  It is a work-alike with the previous implementation.

packages/components/RadioGroup/src/RadioButton.macos.tsx
chripear: Nah it totally works, the debugger shows a huge blob of JSON

chripear: nm I take it back, componentRef was the thing I had been logging.  buttonRef does seem to be empty.

Saadnajmi: Oh. I continue to be confused by this hook.

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -6,26 +6,40 @@ import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
...
       if (buttonKey != info.selectedKey) {
         info.onChange && info.onChange(buttonKey);
+        info.updateSelectedButtonRef && componentRef && info.updateSelectedButtonRef(componentRef);
       }
     };
 
+    /* We use the componentRef of the currently selected button to maintain the default tabbable
+    element in a RadioGroup. Since the componentRef isn't generated until after initial render,
+    we must update it once here. */
+    React.useEffect(() => {
```
Saadnajmi, line 31: Thanks for the comment!

packages/components/RadioGroup/src/RadioButton.macos.tsx
```diff
@@ -6,26 +6,40 @@ import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
...
+      if (buttonKey == info.selectedKey) {
```
Saadnajmi, line 32: nit, I think this should be `===` because javascript

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -281,6 +296,16 @@ - (NSView *)nextViewToFocusOutsideZone:(FocusZoneAction)action
...
 	}
+	else
+	{
+		// If the next view is in a FocusZone, focus the default tabbable element.
+		RCTFocusZone *nextFocusZone = GetFocusZoneAncestor(nextViewToFocus);
+		NSView *element = [nextFocusZone defaultTabbableElement];
+		if (element != nil)
+		{
+			nextViewToFocus = element;
+		}
```
Saadnajmi, line 307: Note: This is speculation (and I also know very little about the responder chain), so take it with many grains of salt, and don't consider it blocking. 

I wonder if the more "macOS" way to do this would have been to remember who the last responder was inside an override of "acceptsFirstResponder" / "resignFirstResponder" or something else to do with overriding the responder chain. I think if we did it that way, we would only have the reference to the last focused element in native and not in JS so it wouldn't work.

Saadnajmi (approved):
🎉

packages/components/RadioGroup/src/RadioButton.macos.tsx
chripear: Is just copypasta from RadioButton.tsx

Saadnajmi: Ah, I still appreciate it nonetheless

PR #929, opened 2021-08-22
Support  menu item images and style for MenuButton on macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change does 3 main things:
- Adds support for showing images in the MenuButton items. I had to filter out the icon prop to only be passed in if it was a PNG, as we don't support SVG's through the native module.
- Adds support for a user defined style prop. This is primarily to let the user pass in a custom height/width to fit their needs, like if they want an icon only MenuButton.
- Refactors the macOS MenuButton to do less menu item prop conversion on the native side, and more prop conversion on the JS side. This is in preparation for factoring out the native menu code into the macOS ContextualMenu component in the future. I created Bug #931 to track that work, and document in code where differences exist because we don't have a macOS ContextualMenu component.

I also made an empty change in Expander to trigger a yarn change file because I'm fairly sure the NPM Publish pipeline is still broken. 

### Verification

Ran the test app with my changes.

Screen Shot 2021-08-23 at 8 51 42 AM

The changes should not affect MenuButton on other platforms, but I will double check once I get access to a windows computer.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Note: I chose not to support the `primary` prop because that's not quite in the design of macOS MenuButtons (unlike win32). Primary style buttons usually indicate a default action and are associated with the "space" or "enter" keyEquivalent. Both of those don't quite apply to the MenuButton in FURN.
https://developer.apple.com/design/human-interface-guidelines/macos/buttons/push-buttons/

packages/components/MenuButton/macos/FRNMenuButtonManager.m
```diff
@@ -3,13 +3,14 @@
 
 @implementation RCTConvert (FRNMenuButtonAdditions)
 
-+ (NSMenuItem *)menuItem:(id)json
++ (NSMenuItem *)NSMenuItem:(id)json
```
harrieshin, line 6: for my own knowledge why to NSMenuItem?

harrieshin (approved):

packages/components/MenuButton/macos/FRNMenuButtonManager.m
Saadnajmi: The method names of RCTConvert usually match the object it returns. I also made it so that the JSON object passed in has keys that match the properties on NSMenuItem. So prior to this change, [RCTConvert menuItem:] took in JSON that was pretty specific to FURN MenuButton. After this change, it takes in a JSON that is much more generic, so I felt the name change appropriate.

I changed this method to be more generic in preparation for when I pull it out of MenuButton and into a standalone ContextMenu component.

PR #961, opened 2021-08-30
Dynamic TabsItem

dake3601:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

With this change, dynamic TabsItems are implemented. The behavior comes from [Show/Hide pivot item of fluentui](https://developer.microsoft.com/en-us/fluentui#/controls/web/pivot). 

The behavior is the following:
The selected tabs item will not change when the number of tabs items changes. If the selected tabs item is removed, the new first item will be selected.

This allows for TabsItems to be conditionally rendered in  and out of Tabs.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
Can you please be specific within the description of this PR? What exactly is a dynamic TabItem? How is this different than a normal TabItem.

It helps when reviewing to know exactly what this change is adding in. Thanks

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
```diff
@@ -177,6 +177,35 @@ const tabsSettingSelectedKey: React.FunctionComponent = () => {
   );
 };
 
+const tabsShowHideItem: React.FunctionComponent = () => {
+  const [showFirstItem, setshowFirstItem] = React.useState(true);
+
+  const toggleShowFirstItem = () => {
```
samuelfreiberg, line 183: Nit: Use React.useCallback() here

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -98,17 +98,25 @@ export const Tabs = compose<TabsType>({
 
     // Populate the tabsItemKeys array
     if (children) {
+      const enabledKeys = []
```
samuelfreiberg, line 101: I'm wondering if this whole `enabledKeys` logic is actually needed. The whole point of enabledKeys is to find the selectedKey, correct? 

Why not just use `data.selectedKey`?

dake3601: The point of enabledKeys is to find an available key to be selected as a new TabsItem if the currently selected has been hidden. For this scenario, I can't see how data.selectedKey would help find the other key that should be selected.

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -98,17 +98,25 @@ export const Tabs = compose<TabsType>({
...
-            renderData.state.context.selectedKey = child.props.itemKey;
+          if (!child.props.disabled) {
+            enabledKeys.push(child.props.itemKey);
           }
           return child.props.itemKey;
         }
       });
+
+      // Sets default selected tabItem
+      enabledKeys.forEach((child: string) => {
```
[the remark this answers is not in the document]
dake3601: Yeah, if a TabsItem is selected and then hidden, the first TabsItem will be selected. This behavior is the intended and was taken from fluentui's dynamic Pivot.

rurikoaraki (approved):

packages/components/Tabs/src/Tabs.tsx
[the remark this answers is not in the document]
rurikoaraki: Cool. The newest commit makes this much clearer as the intended behavior. Makes sense, thanks!

packages/components/Tabs/src/Tabs.tsx
```diff
@@ -98,17 +98,23 @@ export const Tabs = compose<TabsType>({
...
-          if (renderData.state.context.selectedKey == null && !child.props.disabled) {
-            renderData.state.context.selectedKey = child.props.itemKey;
+          if (!child.props.disabled) {
+            enabledKeys.push(child.props.itemKey);
           }
           return child.props.itemKey;
         }
       });
+
+      // Sets default selected tabsItem
```
rurikoaraki, line 114: nit: might be worth adding a comment that default selected item is the first item if the given selected item is hidden.

rurikoaraki (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Tabs/TabsTest.tsx
```diff
@@ -237,6 +266,10 @@ const tabsSections: TestSection[] = [
     name: 'Override Selected Key',
     component: tabsSettingSelectedKey,
   },
+  {
+    name: 'Show/Hide pivot item',
```
samuelfreiberg, line 270: Should this be "Tabs" instead of "Pivot"?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
     // Grabs the context information from Tabs (currently selected TabsItem and client's onTabsClick callback).
     const info = React.useContext(TabsContext);
 
-    /* There's a bug where the user callback is being called multiple times on one click.
-    We check that there is an actual change in selection before forwarding the message to the user callback
-    so that they aren't notified more than once for each click. */
-    const changeSelection = () => {
-      if (itemKey != info.selectedKey) {
+    const [focusState, setFocusState] = React.useState({
```
samuelfreiberg, line 41: Can you just use `pressable.state.focused`?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
...
+      focused: false,
+    });
+
+    const changeSelection = React.useCallback(() => {
+      componentRef.current.focus();
```
samuelfreiberg, line 46: `componentRef?.current?.focus()` is the safest way to do this. It checks if it's undefined first. The way it's currently implemented, if `componentRef == nullptr || .current == nullptr`, it will cause an error

samuelfreiberg (changes requested):

packages/components/Tabs/src/TabsItem.tsx
dake3601: I don't know how I could use pressable if I am defining it on line 62.

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
...
+      focused: false,
+    });
+
+    const changeSelection = React.useCallback(() => {
+      componentRef?.current?.focus();
+    }, [componentRef]);
+
+    const changeSelectionWithFocus = React.useCallback(() => {
+      setFocusState({ focused: true });
+      if (componentRef.current && !focusState.focused) {
```
samuelfreiberg, line 51: Wouldn't `focusState.focused = true` always? The line above, you set it to true.

dake3601: ```setFocusState({ focused: true });``` works asynchronously. The first time that function is called, even if focusState.focused is set to true, it will be false on evaluation (from onBlur/removeFocus). The next evaluations done (repeated focus bug), focusState.focused wiill be true and stop the user functions to be called more than once.

samuelfreiberg: Is `focusState.focused = true` supposed to go on line 51?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
...
         info.onTabsClick && info.onTabsClick(itemKey);
         info.getTabId && info.getTabId(itemKey, info.tabsItemKeys.findIndex(x => x == itemKey) + 1);
         info.updateSelectedTabsItemRef && componentRef && info.updateSelectedTabsItemRef(componentRef);
       }
-    };
+    }, [focusState, setFocusState]);
```
samuelfreiberg, line 56: I believe the dependency array should include componentRef and info

`[focusState, setFocusState, componentRef, info]`

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
...
 
-    // Ensure focus is placed on tabsItem after click.
-    const changeSelectionWithFocus = useOnPressWithFocus(componentRef, changeSelection);
+    const removeFocus = React.useCallback(() => {
+      setFocusState({ focused: false });
+    }, [focusState, setFocusState]);
```
samuelfreiberg, line 60: focusState is not needed in this dependency array, just `[setFocusState]`

packages/components/Tabs/src/TabsItem.tsx
samuelfreiberg, line 51: Why do we need to check for `componentRef.current` here? componentRef is an optional prop, correct? If the user doesn't pass one in, this will never change selection?

dake3601: It is an optional prop, but if the user doesn't pass one, componentRef is set to defaultComponentRef. This can be seen on line 24 and 29 of TabsItem.tsx.

samuelfreiberg: What does `componentRef.current` equal when the user doesn't pass one in?

samuelfreiberg: Wouldn't it be null? And then `componentRef.current` wouldn't be true?

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -39,24 +38,32 @@ export const TabsItem = compose<TabsItemType>({
...
+
+    const changeSelectionWithFocus = React.useCallback(() => {
+      setFocusState({ focused: true });
+      if (!focusState.focused) {
         info.onTabsClick && info.onTabsClick(itemKey);
         info.getTabId && info.getTabId(itemKey, info.tabsItemKeys.findIndex(x => x == itemKey) + 1);
         info.updateSelectedTabsItemRef && componentRef && info.updateSelectedTabsItemRef(componentRef);
       }
-    };
+    }, [focusState, setFocusState, componentRef, info]);
```
samuelfreiberg, line 56: Actually, this would also need `itemKey`. Sorry I missed that. Every var/function within a useCallback should be in the dependency array, because if one of them changes, the output of the function changes.

packages/components/Tabs/src/TabsItem.tsx
dake3601: If the user does not pass one, it would be equal to defaultComponentRef = React.useRef(null).
And it was evaluating for true as the functions bellow where being called.

dake3601: Changed it so only ```!focusState.focused``` gets evaluated on this if statement.

samuelfreiberg (approved):

PR #976, opened 2021-09-03
Add default values for alias tokens for each Office theme

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
This change:
- Adds all the brand colors for win32 into the pipeline
- Adds input files for each Office theme
- Updates win32 office theme creation to take new tokens into account
- Updates tokens test page to render win32 tokens differently

Still missing
- Brand colors taken from JSON files instead of Native Module
- Fixing up tester to have correct brand values for lighter/darker themes

### Verification

Colorful/White:
image

Dark Gray:
image

Black:
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-tokens/src/getOfficeAliasTokens.ts
```diff
@@ -0,0 +1,18 @@
+import * as colorfulAliasTokens from './colorful/tokens-alias';
+import * as darkGrayAliasTokens from './darkGray/tokens-alias';
+import * as blackAliasTokens from './black/tokens-alias';
+import * as hcAliasTokens from './highContrast/tokens-alias';
+
+export function getOfficeAliasTokens(officeTheme: string) {
+  if (officeTheme === 'White' || officeTheme === 'Colorful') {
+    return colorfulAliasTokens.default;
```
ksiler, line 8: White and Colorful theme use the same colors/palette?

rurikoaraki: Yep, they're exactly the same right now. That may change in the future, but that should just mean we would add a new input file for White theme, or something

packages/theming/theme-tokens/src/pipeline-input/token-input-black.json
```diff
@@ -0,0 +1,218 @@
...
+  "$schema": "https://raw.githubusercontent.com/microsoft/fluentui-token-pipeline/main/docs/token.schema.json",
+  "Meta": {
+    "FluentUITokensVersion": 0
+  },
+  "Set": {
+    "NeutralForeground2Brand": {
+      "Fill": {
+        "Color": {
+          "Hover": { "aliasOf": "Global.Color.Brand.Tint.40" },
+          "Pressed": { "aliasOf": "Global.Color.Brand.Tint.10" }
```
ksiler, line 11: Trying to understand what "Hover" and "Pressed" mean here for "NeutralForeground2Brand". If I set this alias token as the foreground color of a control, does this mean it will automatically change colors depending on the state (like hover and pressed) and I don't need to explicitly set a different color for hover and pressed states in my token/settings file?

ksiler: Oh nevermind, I think I see how it works. These are eventually split out into "neutralForeground2Hover" and "neutralForeground2Pressed" and that value for those changes based on the theme (black, dark, colorful, etc.)?

rurikoaraki: Unfortunately no, although that would be nice.

Design gave us the alias tokens as "neutralForeground2Brand", "neutralForeground2BrandHover", neutralForeground2BrandPressed", etc. This file is the file that gets fed into the pipeline, which expects the input to be formatted as:

```
<Root of alias token name>: {
    Fill: {
        Color: {
            Rest/Hover/Pressed/etc. as separate entries
        }
    }
}
```

This particular file is only overriding entries that are different from token-input-dark.json (the web dark mode version). The pipeline allows you to stack inputs, similar to how styles work in RN. That's why some entries are missing.

There's a separate place that takes care of mapping the output json structure to actual alias tokens stored on the Theme object, which was moved to mapPipelineToTheme. These convert the output to the alias token names given by design, so that when you're given redlines, you can reference those theme object entries to fill out token entries. Design will specify whether they want hover/pressed, or not.

ksiler: Ah okay 🙂 thanks for the explanation!

packages/theming/win32-theme/src/createBrandedThemeWithAlias.ts
```diff
@@ -1,32 +1,34 @@
 import { Theme, PartialTheme, AliasColorTokens } from '@fluentui-react-native/theme-types';
 
-export function createBrandedThemeWithAlias(theme: Theme): PartialTheme {
-  if (!theme.host.colors) {
+export function createBrandedThemeWithAlias(themeName: string, theme: Theme): PartialTheme {
+  if (themeName === 'HighContrast' || !theme.host.colors) {
     return {};
```
ksiler, line 5: No values for HC yet?

rurikoaraki: HC is in already (see https://github.com/microsoft/fluentui-react-native/tree/master/packages/theming/theme-tokens/src/highContrast) but we shouldn't apply brand colors in HC, so we skip this step.

rurikoaraki: Specifically this steps pulls the brand color swatches directly from the native module, which doesn't change when you switch to HC.

ksiler: Ah I see, I never noticed that but thinking about it makes sense. Thanks for the explanation :)

ksiler (approved):

PR #985, opened 2021-09-07
(Apple) Pack only ship slices of libraries in our NuGet

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We made a change downstream where we consume the NuGet to only ever link the ship slices of the libraries (even if building debug). As such, we can save some time and not build and pack them in our pipeline in the first place. I took this as an opportunity to simplify the NuGet Publish pipeline too.

Summary:
- Removed the build debug steps and the yarn bundle steps from the NuGet Publish job. That should cut the job time in half (!!!)
- Renamed `nuget-publish.yml -> apple-nuget-publish.yml`, and moved the entire NuGet publish job inside of it, so we can reference it from both the PR and Publish pipeline easily. Removed `apple-build-static-libs.yml` as a consequence.

### Verification

Ran the NuGet publish job as part of the PR job once to make sure it isn't broken. 
You can see the results here: https://dev.azure.com/ms/ui-fabric-react-native/_build/results?buildId=219165&view=logs&j=f22c9018-44f9-5836-07db-908da0f6bf23&t=44edb6c8-a9d2-5bdd-d017-7858e5b6b8b9

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
I forgot, downstream we still consume the iOS debug binaries. I have a code review to fix that, so I won't merge this until that one is merged.

harrieshin:
"We made a change downstream where we consume the NuGet to only ever link the ship slices of the libraries (even if building debug). " -> i am surprised by that. why?

package.nuspec
```diff
@@ -9,33 +9,16 @@
     <repository type="git" url="$repoUri$" commit="$commitId$"/>
   </metadata>
   <files>
-    <!-- iOS device debug -->
-    <file src="DerivedData\Build\Products\Debug-iphoneos\FRNAvatar\libFRNAvatar.a" target="Debug-iphoneos"/>
```
harrieshin, line 13: with empty debug target that is still ok? how would we ever debug native layers of RN?

Saadnajmi:
> "We made a change downstream where we consume the NuGet to only ever link the ship slices of the libraries (even if building debug). " -> i am surprised by that. why?

Saadnajmi:
> "We made a change downstream where we consume the NuGet to only ever link the ship slices of the libraries (even if building debug). " -> i am surprised by that. why?

HImanshu and Anand have both mentioned it as optimizations before. We don't have dsyms in the debug libraries, so their usefulness is limited. I suppose they realized we can save on NuGet size by just packing the ship binaries.

package.nuspec
Saadnajmi: We never had a full end to end debug story working for the FURN static libraries in devmain. If we wanted to, we would need to write a custom script to build the libraries locally, then side load them into devmain. That script could just build debug libraries with dsyms, and pack them into a "ship-$(PLATFORM_NAME)" directory inside the NuGet and devmain would be none the wiser when we side load that NuGet. 

In summary, I don't think this change would prevent us from debugging FURN native code in devmain in the future, we still need to side load with a custom script.

harrieshin (approved):

PR #993, opened 2021-09-08
Experimental Checkbox

ksiler:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Move Checkbox over to the new framework.

I'm looking to add snapshot tests in a future commit.

### Verification

Manually tested in Fluent Tester on win32, Windows, and macOS.

| Old Framework                                      | New Framework                                      |
|image| image |
| image | image |
| image | image |


packages/experimental/Checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+    // grab the styled slots
+    const Slots = useSlots(userProps, layer => Checkbox.state[layer] || userProps[layer]);
+    // now return the handler for finishing render
+    return (final: CheckboxProps, ...children: React.ReactNode[]) => {
+      const { label, ...mergedProps } = mergeProps(Checkbox.props, final);
+
+      // const boxAtEndStyle = {
+      //   marginStart: 4,
+      //   marginEnd: 0,
+      // };
```
Saadnajmi: remove commented out code?

packages/experimental/Checkbox/src/Checkbox.types.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface CheckboxTokens extends FontTokens, IForegroundColorTokens, IBackgroundColorTokens, IBorderTokens {
+  checkboxBackgroundColor?: ColorValue;
+  checkboxBorderColor?: ColorValue;
+  checkmarkColor?: ColorValue;
+  checkmarkVisibility?: number;
+  textBorderColor?: ColorValue;
+  checkboxBorderRadius?: number;
+  checkboxMarginStart?: number;
+  checkboxMarginEnd?: number;
+  checkboxVisibility?: number;
```
Saadnajmi, line 18: What does visibility do?

packages/experimental/Checkbox/src/Checkbox.types.ts
```diff
@@ -0,0 +1,111 @@
...
+   * A RefObject to access the IFocusable interface. Use this to access the public methods and properties of the component.
+   */
+  componentRef?: React.RefObject<IFocusable>;
+
+  /**
+   * Callback that is called when the checked value has changed.
+   */
+  onChange?: (isChecked: boolean) => void;
+
+  testID?: string;
```
Saadnajmi, line 69: I think we inherit this from ViewProps?

packages/experimental/Checkbox/src/Checkbox.types.ts
```diff
@@ -0,0 +1,111 @@
...
+   */
+  disabled?: CheckboxTokens;
+  boxAtEnd?: CheckboxTokens;
+  hovered?: CheckboxTokens;
+  focused?: CheckboxTokens;
+  pressed?: CheckboxTokens;
+  checked?: CheckboxTokens;
+}
+
+export interface CheckboxProps extends Omit<IViewProps, 'onPress'> {
```
Saadnajmi, line 31: why omit onPress?

packages/experimental/Checkbox/src/useCheckbox.ts
```diff
@@ -0,0 +1,80 @@
...
+  useAsPressable,
+  useKeyCallback,
+  useOnPressWithFocus,
+  useViewCommandFocus,
+  useAsToggle,
+} from '@fluentui-react-native/interactive-hooks';
+import { CheckboxProps, CheckboxInfo, CheckboxState } from './Checkbox.types';
+import { IPressableProps } from '@fluentui-react-native/pressable';
+
+export const useCheckbox = (props: CheckboxProps): CheckboxInfo => {
```
Saadnajmi, line 12: Maybe add some documentation on this method so when people hover over it, VS code / intelligence gives a short summary? 
Note: I only recently learned about how we can use hooks to provide the state of an _entire_ component with one hook, so this pattern was confusing to me at first. Now that I understand it, I think it's pretty cool :D

packages/experimental/Checkbox/src/useCheckbox.ts
```diff
@@ -0,0 +1,80 @@
...
+      accessible: true,
+      focusable: true,
+      accessibilityRole: 'checkbox',
+      accessibilityLabel: accessibilityLabel ?? label,
+      accessibilityState: { disabled: state.disabled, checked: state.checked },
+      accessibilityActions: [{ name: 'Toggle' }],
+      onAccessibilityAction: onAccessibilityAction,
+      onKeyUp: onKeyUpSpace,
+      ...props,
+    },
```
Saadnajmi, line 74: yay accessibility props!

packages/experimental/Checkbox/src/CheckboxTokens.macos.ts
```diff
@@ -0,0 +1,31 @@
...
+    checked: {
+      checkboxBorderColor: t.colors.checkboxBackground,
+      checkboxBackgroundColor: t.colors.checkboxBackground,
+      checkmarkVisibility: 1,
+    },
+    boxAtEnd: {
+      checkboxMarginStart: 4,
+      checkboxMarginEnd: 0,
+    },
+  } as CheckboxTokens);
```
Saadnajmi, line 31: Are the macOS tokens and styling actually different? Or are they inherited from the old checkbox? or did you choose them from Figma?

packages/experimental/Checkbox/src/Checkbox.types.ts
ksiler: Visibility = opacity, sometimes you have a checked checkbox that's disabled and the styles for that are the same as a checked checkbox but slightly dimmed

packages/experimental/Checkbox/src/Checkbox.types.ts
ksiler: Yup :) I'll remove it

packages/experimental/Checkbox/src/Checkbox.types.ts
ksiler: I copied this from the old checkbox - I think it's removed because we want them to use 'onChange' instead which is assigned to onPress when we use useAsPressable (see useCheckbox.ts, line 36).

packages/experimental/Checkbox/src/CheckboxTokens.macos.ts
ksiler: I couldn't find any Figma files for checkbox on macOS. If you know where they exist, send them my way :) 

The values for the tokens on macOS are slightly different. I pulled them from both the old checkbox and this file: https://github.com/microsoft/fluentui-react-native/blob/master/packages/theming/apple-theme/src/appleTheme.macos.ts#L49 which the old checkbox seemed to be picking up the styles from but I'm not sure how to hook it up to the new framework so I copied them over directly. @Saadnajmi do you know how this file works/if I can get it to work for the new checkbox?

Saadnajmi: https://www.figma.com/file/RKqH3UfoIZpO1ZZIlMSp0j/Fluent-vNext-macOS?node-id=1406%3A2386
I wouldn't give it too much weight though, designs' take for most macOS controls has been "use the system control".
Those values I think Tushar added, but are just the same structure as a settings file for the old framework. In effect, they override the default settings of the control. I used to put control specific overrides there because I thought all the "theme" info should be in the apple theme files instead of in a .<platform> file. It feels like we're not really doing that anymore, especially with the token pipeline.

The settings aren't being picked up from the apple theme because the names are different: "checkbox" vs "Checkbox". 
I kinda like having them not picked up though, it kinda feels like we don't want to do the thing where we put component settings in that one file, so I wouldn't change how you've done it =)

Saadnajmi: Actually, wait, both are "Checkbox". I'm not sure why those settings aren't messing with the settings here. Happy accident?

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -0,0 +1,162 @@
+import * as React from 'react';
+import { ExperimentalCheckboxTestPageId } from './consts';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { Checkbox } from '@fluentui-react-native/experimental-checkbox';
+import { useTheme } from '@fluentui-react-native/theme-types';
+import { View, TextInput, TextStyle } from 'react-native';
+import { commonTestStyles as commonStyles } from '../Common/styles';
+
+function onChangeUncontrolled(isChecked: boolean) {
+  console.log(isChecked);
```
rurikoaraki, line 10: Is this meant to be checked in?

packages/experimental/Checkbox/src/Checkbox.types.ts
Saadnajmi: Could we just call it opacity then?

packages/experimental/Checkbox/src/Checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const Slots = useSlots(userProps, layer => Checkbox.state[layer] || userProps[layer]);
+    // now return the handler for finishing render
+    return (final: CheckboxProps, ...children: React.ReactNode[]) => {
+      const { label, ...mergedProps } = mergeProps(Checkbox.props, final);
+
+      return (
+        <Slots.root {...mergedProps}>
+          {Checkbox.state.boxAtEnd && <Slots.content key="content">{label}</Slots.content>}
+          <Slots.checkbox>
+            <Slots.checkmark key="checkmark">✓</Slots.checkmark>
```
rurikoaraki, line 31: nit: It's a bit unfortunate that we do "unchecked" by rendering the checkmark with 0 opacity, since that's more costly from a rendering perspective than not rendering the checkmark. It would only be a small perf optimization though so not a blocker.

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
ksiler: the console.log line? it appears in our other checkbox test page so I was planning to check it in

packages/experimental/Checkbox/src/Checkbox.types.ts
ksiler: Sure. It would diverge from the old Checkbox tokens though. But I would prefer it to fall more in line with the name of the style prop it's changing.

packages/experimental/Checkbox/src/useCheckbox.ts
```diff
@@ -0,0 +1,86 @@
...
+  useAsPressable,
+  useKeyCallback,
+  useOnPressWithFocus,
+  useViewCommandFocus,
+  useAsToggle,
+} from '@fluentui-react-native/interactive-hooks';
+import { CheckboxProps, CheckboxInfo, CheckboxState } from './Checkbox.types';
+import { IPressableProps } from '@fluentui-react-native/pressable';
+
+/** Re-usable hook for FURN Checkbox.
```
rurikoaraki, line 12: nit: Comment should start on line after the comment opener

Also I think the syntax for props/returns is `@param props` and `@returns` in JSDocs

rurikoaraki (approved):

packages/experimental/Checkbox/src/useCheckbox.ts
ksiler: Thanks! Those are good callouts.

rurikoaraki (approved):

rurikoaraki (approved):

packages/experimental/Checkbox/src/Checkbox.tsx
ksiler: It just occurred to me that if we only render the checkmark when it's checked (instead of rendering it with 0 vs 1 opacity) then we can't have a checkbox where the checkmark is visible on hover (which is one of our examples in the Fluent Tester). For this reason, I may change the behavior back to what it was since I can't see how else someone would achieve that without messing with the opacity of an already rendered checkmark.

rurikoaraki: Yeah that's a good catch. Makes sense to me.

rurikoaraki (approved):

PR #995, opened 2021-09-10
Take brand colors from pipeline

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change is a follow up to #976.

This change pulls brand colors from the pipeline instead of the native module, so we are closer to having most things pull from the pipeline tokens as a source of truth.

I added brand colors for the "web" version (the base pipeline input file) so that the schema between the web and win32 versions of the pipeline input would match up. This probably isn't the best solution, however it allows the pipeline output for globals to be in one file as opposed to having brand colors be in separate files. This could be better from a pipeline output perspective so that pulling inputs is more straightforward, but I'm open to suggestions here.

I deleted Access an Publisher as I don't think those will have use cases and they don't have colors specified on web.

We match the correct brand tokens by taking the current AppPrimary off the native module. This is, I think, the best way we have available right now to get the host app. This isn't great - probably we should add a way to query for the host app or have that information available somehow - but at least for the short term it should suffice as a way to get the host app.

We also have the logic run on the tester now so that we have less copied code.

### Verification

Ran the tester on Office theme with various brands and Office themes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
+++ b/apps/fluent-tester/src/FluentTester/theme/CustomThemes.ts
@@ -39,7 +39,11 @@ export class TesterThemeReference extends ThemeReference {
   private baseTheme: ThemeReference;
 
   constructor() {
-    super(baseTheme, parent => applyTheme(parent, this._themeName, this.options.appearance), () => applyBrand(this._brand));
+    super(
+      baseTheme,
+      parent => applyTheme(parent, this._themeName, this.options.appearance),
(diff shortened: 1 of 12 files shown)


Saadnajmi (approved):
At some point I'll have to learn ho wot use all these tokens for the apple theme :')

PR #996, opened 2021-09-10
Allow theming module to be implemented as a turbo module

acoates-ms:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Using `TurboModuleRegistry.get('Theming')` instead of `NativeModules.Theming`, which will allow the theming module to be implemented as either a NativeModule, or a TurboModule.

Also changed a couple of uses of `NativeModule.UIManager`, to instead use the exported UIManger object, which has better type information, and will allow UIManager to be implemented as a TurboManager in the future too. 

### Verification

Booted the test app on win32 and went to the theming page.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
+++ b/packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
@@ -1,12 +1,15 @@
 import { OfficeThemingModule } from './officeThemingModule';
-import { NativeEventEmitter, NativeModules } from 'react-native';
+import { NativeEventEmitter } from 'react-native';
 import { fallbackGetPalette, fallbackOfficeModule } from './fallbackOfficeModule';
 
 declare module 'react-native' {
-  interface NativeModulesStatic {
-    Theming: OfficeThemingModule & EventSubscriptionVendor;
+  interface ITurboModuleRegistry {
+    get: (name: 'Theming') => OfficeThemingModule & EventSubscriptionVendor;
   }
+  const TurboModuleRegistry: ITurboModuleRegistry;
 }
+
+import { TurboModuleRegistry } from 'react-native';
 
 /**
  *  If we have a userAgent string, let's assume we're web debugging.  __DEV__ is for developer bundles.  Currently,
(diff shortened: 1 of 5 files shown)


packages/utils/interactive-hooks/src/useViewCommandFocus.macos.ts
```diff
@@ -22,19 +22,15 @@ export function useViewCommandFocus(
...
        * Add focus() as a callable function to the forwarded reference.
        */
       if (localRef) {
         localRef.focus = () => {
-          NativeModules.UIManager.dispatchViewManagerCommand(
-            findNodeHandle(localRef),
-            NativeModules.UIManager.getViewManagerConfig('RCTView').Commands.focus,
-            null,
-          );
+          UIManager.dispatchViewManagerCommand(findNodeHandle(localRef), UIManager.getViewManagerConfig('RCTView').Commands.focus, null);
```
Saadnajmi, line 33: Dumb question I've had that I haven't been able to answer elsewhere:
This `UseViewCommandFocus` hook seems redundant, since the standard ref from React should have a `.focus()` method on it. But if I remove the hook altogether, it seems imperative focus stops working on win32. Would you happen to know why?

packages/theming/win32-theme/src/NativeModule/getThemingModule.native.ts
```diff
@@ -19,7 +22,7 @@ function disableGetPalette(): boolean {
 }
 
 export function getThemingModule(): [OfficeThemingModule, NativeEventEmitter | undefined] {
-  const module = NativeModules && NativeModules.Theming;
+  const module = TurboModuleRegistry.get('Theming');
```
Saadnajmi, line 25: Would this be backwards compatible with Native Modules, or does this only work for TurboModules?

acoates-ms: https://github.com/facebook/react-native/blob/d8d9c28b482ed294770592ecb75bd6cc7f5c9da7/Libraries/TurboModule/TurboModuleRegistry.js#L21

Yes - it'll try to get a NativeModule if it exists for backwards compat.

packages/utils/interactive-hooks/src/useViewCommandFocus.macos.ts
acoates-ms: Sounds like a bug.  We should make sure that view.focus() does the right thing in win32.

rurikoaraki (approved):

PR #997, opened 2021-09-10
Bump Prettier and run it

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Our Prettier is pretty out of date at this point, and is tripping up on some syntax that they've since fixed.

This change bumps Prettier to latest, and reruns it on the repo to get us clean again. No functional changes.

### Verification

Builds

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):
Any reason dependabot isn't updating this for us?

rurikoaraki:
> Any reason dependabot isn't updating this for us?

Idk if/why it didn't bump to 1.19.1 for us, but everything past that was a major version bump, which we set package.json not to do automatically (the ~ makes it so that we accept any package that doesn't change the major version number). This also makes sense because major version bumps tend to signify breaking changes, so people tend to want to make those leaps on their own time.

PR #1024, opened 2021-09-22
Experimental Tabs (macOS + Windows)

ksiler:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Implement Tabs in the new framework on macOS and windows.

### Verification

Manually tested keyboarding, clicking, and voice over behavior for Tabs on macOS and windows.

image

image

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


apps/fluent-tester/src/FluentTester/testPages.macos.ts
```diff
@@ -128,6 +129,11 @@ export const tests: TestDescription[] = [
     component: TabsTest,
     testPage: HOMEPAGE_TABS_BUTTON,
   },
+  {
+    name: 'Tabs Experimental Test',
+    component: TabsExperimentalTest,
```
Saadnajmi, line 134: Most tests prefix "Experimental" (I.E: Experimental Button), while this one makes it a suffix. I prefer your style, but could we keep it consistent across them alll?

packages/experimental/Tabs/src/Tabs.styling.macos.ts
```diff
@@ -0,0 +1,41 @@
...
+export const stylingSettings: UseStylingOptions<TabsProps, TabsSlotProps, TabsTokens> = {
+  tokens: [defaultTabsTokens, tabsName],
+  slotProps: {
+    root: buildProps(
+      () => ({
+        style: {
+          display: 'flex',
+          minHeight: 32,
+          minWidth: 80,
+          marginLeft: 4,
```
ksiler, line 16: The only style that differs between macOS, win32, and windows for Tabs are margins (margin, marginTop, marginStart, etc.) which are placed on different slots for each platform. I'd like to reduce the amount of platform specific files we have  for this control but I didn't see how else to apply these styles w/o creating a separate file for each platform. If there's a better way to do this, let me know.

packages/experimental/Tabs/src/useTabs.ts
```diff
@@ -1,10 +1,16 @@
...
+import { TabsProps, TabsState, TabsInfo } from './Tabs.types';
 
-export const useTabs = (props: TabsProps): TabsState => {
-  // Attach the pressable state handlers.
+/**
+ * Re-usable hook for FURN Tabs.
+ * This hook configures tabs props and state for FURN Tabs.
+ *
+ * @param props user props sent to FURN Tabs
+ * @returns configured props and state for FURN Tabs
```
Saadnajmi, line 11: I don't think we actually use "FURN" anywhere in code. Could we just omit that word?

Saadnajmi (approved):

packages/experimental/Tabs/src/Tabs.types.ts
```diff
@@ -83,22 +89,25 @@ export interface TabsProps extends Pick<FocusZoneProps, 'isCircularNavigation' |
...
-    context?: TabsContextData;
-    info: TabsInfo;
-  };
+  state: TabsState;
 }
 export interface TabsSlotProps {
   root: React.PropsWithRef<IViewProps>;
   label: TextProps;
-  container: FocusZoneProps;
+  container?: FocusZoneProps;
```
ksiler, line 110: made this slot optional since windows doesn't use FocusZone

packages/experimental/Tabs/src/TabsItem.types.ts
```diff
@@ -14,6 +14,8 @@ export interface TabsItemTokens extends FontTokens, IBorderTokens {
    */
   indicatorColor?: string;
 
+  tabsItemOpacity?: number;
```
ksiler, line 17: Should we have platform specific tokens? This token is really only used for the Windows platform, though I can hook it up to macOS and win32 as well.

apps/fluent-tester/src/FluentTester/testPages.macos.ts
ksiler: Sure! Consistency is probably best.

packages/experimental/Tabs/src/useTabs.ts
ksiler: Sure

packages/experimental/Tabs/src/Tabs.styling.macos.ts
Saadnajmi: Is that just because windows doesn't have focusZone? I think that was the main reason slots were different for windows. Not sure about win32

ksiler: No, (unless I'm misunderstanding you) I don't think this has to do with focusZone at all. It's just that the different platforms have different styles for the Tabs controls w.r.t. margins on the root, label, and stack.

ksiler: Like macOS places margins on the root and the stack while windows places margins on the label and the stack and win32 doesn't set margins at all. We don't even reference the container slot (which is FocusZone) in this file.

Saadnajmi: Do they have the same slots, but different margins on each slot?
That seems weird, can't imagine the style would change that much....

I thought it was that windows vs win32 vs macOS had different slots altogether, my bad.

rurikoaraki: @ksiler I'd consider making a `getTabsMargins()` function which returns different values for each platform (so that file would be forked), and then use it to build the style using the spread operator. Then you could consolidate the rest of this file into one file.

packages/experimental/Tabs/src/TabsItem.types.ts
rurikoaraki: I think we can expose it to all platforms. Maybe someone will want it on win32 one day. We can rename it to make it obvious it's windows specific, or add a comment signaling it. RN has a similar issue with its props, so, I don't think it's that weird to have some platform-ness bleed through.

packages/experimental/Tabs/src/Tabs.styling.macos.ts
ksiler: Yeah, they all reference the same slots in the styling file but margins are placed on different slots. I just copied the styles over directly from Tabs on the old framework. There aren't official designs for Tabs on macOS or Windows so some of these styles are best guesses and may change in the future so maybe that's why it seems weird?

ValentinaKozlova (approved):

packages/experimental/Tabs/src/Tabs.styling.macos.ts
ksiler: @rurikoaraki Thanks for the suggestion! But I'm a bit confused on how that would work if I need to supply different margins to different slots. For example, macOS uses `marginLeft: 4` on the root slot and `marginTop: 6` on the stack slot. If I were to use `...getTabsMargins()` under each slot, how would I know which margin style to return?

rurikoaraki: @ksiler Oh I didn't realize multiple slots had different margins. You could make a function for each one (i.e. `getRootMargins()`, `getStackMargins()`) and define each as needed in a platform-forked file. So like, a `TabsMargins.ts` has the default values for `getRootMargins()` and `getStackMargins()` but `TabsMargins.windows.ts` has whatever you need for UWP specifically.

rurikoaraki: @ksiler I think that's up to you. The one downside of leaving it as is is that it's not clear which styles are common across platforms, so if someone tweaks one style, that wouldn't propagate to other components automatically and it's easy to miss that other platforms may need changing, too. But you're right that we might end up having more separate styling for each platform, so it's valid to leave it as is if you foresee that the styling will diverge a decent amount on other platforms.

ksiler: @rurikoaraki  I see, thanks for your perspective!

PR #1030, opened 2021-09-25
Style experimental button according to redlines

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

This change brings our experimental button into alignment with the fluent redlines.

There are several changes here, but the big ones are:
- Added a size token for the button. There are three preset sizes, small, medium, and large, for the button.
    - default is small for win32 and medium for UWP/web
- Added new states for the button for size
- Added some additional styling tokens according to the figma for things that we should let consumers adjust, like padding and spacing between elements
- Refactored styling files so that the color tokens are in their own files, and the style files don't need to be split per platform

This PR primarily focuses on making sure win32 aligns with redlines. It also focuses on SVG icons. Follow ups for UWP and other icon types are here and here, respectively (TODO: log those items)

Additionally there's some information pulled directly from global tokens which arguably belong on the theme object (typography, spacing). This item tracks adding those elements to the Theme object.

### Verification

Confirmed the styling with design.

Before:
image
image

After:
image
image
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -136,7 +136,8 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
         <ScrollView contentContainerStyle={fluentTesterStyles.testListContainerStyle}>
           {sortedTestComponents.map((description, index) => {
             return (
-              <StealthButton
+              <Button
```
rurikoaraki, line 139: Figured I could bring the tester up to our latest button here.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
```diff
@@ -9,7 +9,7 @@ export const ButtonIconTest: React.FunctionComponent = () => {
   const fontBuiltInProps = {
     fontFamily: 'Arial',
     codepoint: 0x2663,
-    fontSize: 32,
+    fontSize: 24,
```
rurikoaraki, line 12: This font size gives us approximately the size of icon we'd want for small/medium buttons

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
```diff
@@ -23,29 +23,19 @@ export const ButtonIconTest: React.FunctionComponent = () => {
 
   return (
     <View style={[stackStyle, commonTestStyles.view]}>
+      {svgIconsEnabled ? <Button icon={{ svgSource: svgProps }} style={commonTestStyles.vmargin} /> : null}
+      {svgIconsEnabled ? <Button primary icon={{ svgSource: svgProps }} style={commonTestStyles.vmargin} /> : null}
       <Button
-        icon={{ fontSource: fontBuiltInProps, width: 100, height: 100, color: '#060' }}
+        icon={{ fontSource: { ...fontBuiltInProps, fontSize: 32 }, color: '#060' }}
```
rurikoaraki, line 29: This was to leave some sort of customization test

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonIconTestSection.tsx
```diff
@@ -23,29 +23,19 @@ export const ButtonIconTest: React.FunctionComponent = () => {
...
-          icon={{ svgSource: svgProps, width: 20, height: 20, color: 'red' }}
-          content="SVG"
-          style={commonTestStyles.vmargin}
-        />
-      ) : null}
+      <Button icon={{ fontSource: fontBuiltInProps }} content="Font icon" style={commonTestStyles.vmargin} />
+      <Button primary icon={{ fontSource: fontBuiltInProps }} content="Font icon" style={commonTestStyles.vmargin} />
       {svgIconsEnabled ? (
-        <Button icon={{ svgSource: svgProps, width: 20, height: 20 }} content="SVG" style={commonTestStyles.vmargin} />
+        <Button primary icon={{ svgSource: svgProps, color: 'red' }} content="SVG" style={commonTestStyles.vmargin} />
```
rurikoaraki, line 36: I got rid of the width/height now that we have a default width/height for the icon.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonSizeTestSection.tsx
```diff
@@ -0,0 +1,28 @@
+import { Button, CompoundButton } from '@fluentui-react-native/experimental-button';
+import { SvgIconProps } from '@fluentui-react-native/icon';
+import * as React from 'react';
+import { Platform, View } from 'react-native';
+import { commonTestStyles, stackStyle } from '../Common/styles';
+import TestSvg from './test.svg';
+
+export const ButtonSizeTest: React.FunctionComponent = () => {
```
rurikoaraki, line 8: This is so we have a size example/test on the tester for the new size tokens

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -32,18 +46,34 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
       (tokens: ButtonTokens, theme: Theme) => ({
         style: {
           color: tokens.color,
+          marginLeft: tokens.spacingIconContent ?? 1,
+          marginRight: -1,
```
rurikoaraki, line 50: The negative margin here is to force the text to be laid out in the center of the button. There's an issue with the Segoe UI font such that the space around the text is asymmetric, so that the text is laid out off center in the text component:
image

There's some workarounds that web has used that I will look into, but this may be what we have to do in RN to make the font render well.

packages/experimental/Button/src/Button.tsx
```diff
@@ -3,12 +3,23 @@ import * as React from 'react';
...
 import { buttonName, ButtonType, ButtonProps } from './Button.types';
 import { Text } from '@fluentui-react-native/experimental-text';
-import { stylingSettings } from './Button.styling';
+import { stylingSettings, getDefaultSize } from './Button.styling';
 import { compose, mergeProps, withSlots, UseSlots } from '@fluentui-react-native/framework';
 import { useButton } from './useButton';
 import { Icon } from '@fluentui-react-native/icon';
 import { createIconProps } from '@fluentui-react-native/interactive-hooks';
 
+export const buttonLookup = (layer: string, state: ButtonState, userProps: ButtonProps): boolean => {
```
rurikoaraki, line 12: This was getting large and we would want to share it across the various button types, so I refactored it out into its own function

packages/experimental/Button/src/Button.tsx
```diff
@@ -21,21 +32,15 @@ export const Button = compose<ButtonType>({
...
     const iconProps = createIconProps(userProps.icon);
     // grab the styled slots
-    const Slots = useSlots(userProps, (layer) => button.state[layer] || userProps[layer]);
+    const Slots = useSlots(userProps, layer => buttonLookup(layer, button.state, userProps));
+
     // now return the handler for finishing render
     return (final: ButtonProps, ...children: React.ReactNode[]) => {
       const { icon, content, ...mergedProps } = mergeProps(button.props, final);
-      const marginBetween = {
-        marginLeft: icon && content ? 10 : 0,
```
rurikoaraki, line 29: The ButtonToken.ts file handles this now

packages/experimental/Button/src/Button.types.ts
```diff
@@ -1,34 +1,15 @@
...
 import { IconProps, IconSourcesType } from '@fluentui-react-native/icon';
 
 export const buttonName = 'Button';
+export type ButtonSize = 'small' | 'medium' | 'large';
 
-export interface ButtonTokens extends FontTokens, IBorderTokens, IShadowTokens {
-  /**
-   * Background color for the button
-   */
-  backgroundColor?: ColorValue;
```
rurikoaraki, line 15: The tokens I deleted can be handled by the other tokens that I added to the "extends" portion (IColorTokens, LayoutTokens). It just seemed better to make ButtonTokens inherit from them in case new ones are added in the future, it wouldn't be something the ButtonTokens has to add separately.

packages/experimental/Button/src/Button.types.ts
```diff
@@ -47,26 +28,22 @@ export interface ButtonTokens extends FontTokens, IBorderTokens, IShadowTokens {
...
   /**
    * The weight of the lines used when drawing the icon.
    */
   iconWeight?: number;
 
   /**
-   * Text to show on the Button.
+   * The width of the button.
    */
-  content?: string;
```
rurikoaraki, line 60: content and icon seemed weird to have as tokens since they're not influencing styling... but if someone feels otherwise I can add them back.

packages/experimental/Button/src/ButtonColorTokens.ts
```diff
@@ -0,0 +1,128 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { ButtonTokens } from './Button.types';
+import { shadowStyleFromGlobalToken } from './shadowStyle';
+
+export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme) =>
```
rurikoaraki, line 6: This is same as ButtonTokens.ts was before, minus layout-related tokens.

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -1,139 +1,65 @@
...
-        borderColor: t.colors.brandedPressedBorder,
-        iconColor: t.colors.brandedPressedIcon,
-      },
-      focused: {
-        backgroundColor: t.colors.brandedFocusedBackground,
-        color: t.colors.brandedFocusedContent,
-        borderColor: t.colors.brandedFocusedBorder,
-        iconColor: t.colors.brandedFocusedIcon,
+    medium: {
+      padding: globalTokens.spacing.sNudge - 1,
```
rurikoaraki, line 24: - 1 here is to adjust for the border taking up a pixel

packages/theming/theme-tokens/src/pipeline-input/token-input.json
```diff
@@ -416,43 +432,6 @@
       }
     },
 
-    "Square": {
```
rurikoaraki, line 419: These aren't being used so I figured I would remove them

packages/experimental/Button/src/ButtonTokens.ts
Saadnajmi: Is the border taking up a pixel only a win32 thing? AKA, can we assume this for other platforms?
Also should it be "globalTokens.stroke.width.thin" instead of just 1, in case the border width is not always 1?

packages/experimental/Button/src/CompoundButton/CompoundButton.styling.ts
```diff
@@ -79,6 +41,10 @@ export const stylingSettings: UseStylingOptions<CompoundButtonProps, CompoundBut
     content: buildProps(
       (tokens: CompoundButtonTokens, theme: Theme) => ({
         style: {
+          marginLeft: 1,
+          marginRight: -1,
+          marginBottom: 1,
+          marginTop: -1,
```
Saadnajmi, line 47: why the negative numbers here? Also Segoe UI workaround?

Saadnajmi (commented):
It looks nice on macOS/iOS as is too :')
Probably not up to redlines, but much better than before I think. I consider this a win!
Screen Shot 2021-09-26 at 9 23 29 AM
Screen Shot 2021-09-26 at 9 42 37 AM
)

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -32,18 +46,34 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
+
+export const getDefaultSize = (): ButtonSize => {
+  if (Platform.OS === 'windows') {
+    return 'medium';
+  } else if ((Platform.OS as any) === 'win32') {
+    return 'small';
+  }
+
+  return 'medium';
+};
```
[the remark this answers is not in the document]
rurikoaraki: Ah yeah... we could make win32 "default" but that seems weird, I'd rather stick to default as medium for now.

packages/experimental/Button/src/ButtonTokens.ts
rurikoaraki: @Saadnajmi I know that UWP operates the same way but yeah, not sure about other platforms. That will be a point where we could get some per-platform styling. I can preemptively add a function that doesn't do anything for android or Apple.

And yeah, you're right, we should be using the value of the constant.

Saadnajmi: I tried making borderRadius 10 on the apple theme for macOS, and all the buttons got bigger, so I guess this may be an actual platform difference :/ 
Good to keep in mind for future when we use the token pipeline tokens for experimental button on macOS and others!

packages/experimental/Button/src/Button.types.ts
```diff
@@ -79,6 +56,11 @@ export interface ButtonTokens extends FontTokens, IBorderTokens, IShadowTokens {
   subtle?: ButtonTokens;
   fluid?: ButtonTokens;
   fab?: ButtonTokens;
+  small?: ButtonTokens;
+  medium?: ButtonTokens;
+  large?: ButtonTokens;
+  hasContent?: ButtonTokens;
+  hasIcon?: ButtonTokens;
```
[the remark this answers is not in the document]
rurikoaraki: These tokens would affect how the button is styled if these states were active (which is controlled by that lookup function in button.tsx).

i.e. right now by default hasContent by itself doesn't do anything by default, but if you wanted to say change the backgroundColor of the button when the button has content, this token is how you would change that.

Mostly it's here to adjust styling of certain things when the button has content. We can do it in the Button render function instead, but I thought that this way would be better since then the styling is handled mostly in the same space.

packages/experimental/Button/src/ButtonColorTokens.win32.ts
```diff
@@ -3,25 +3,14 @@ import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { ButtonTokens } from './Button.types';
 import { shadowStyleFromGlobalToken } from './shadowStyle';
 
-export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme) =>
+export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme) =>
   ({
     backgroundColor: t.colors.neutralBackground1,
     color: t.colors.neutralForeground1,
     borderColor: t.colors.neutralStroke1,
     iconColor: t.colors.neutralForeground1,
```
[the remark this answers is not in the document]
rurikoaraki: Yeah, the plan is eventually all platforms would use these tokens. If I could, I would make it so that windows/win32 shared the same file, but I can only do one platform at a time with the forking. If I make two objects to share instead I believe it would load/create both objects depending on how I write the code so I figured this would be better even though there would be some code duplication. Let me know if you have other thoughts/ideas.

packages/experimental/Button/src/CompoundButton/CompoundButton.styling.ts
rurikoaraki: Yeah, same Segoe thing here.

rurikoaraki: I can potentially refactor this out

packages/experimental/Button/src/Button.types.ts
```diff
@@ -114,6 +98,9 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
 
   /** A floating action button  */
   fab?: boolean;
+
+  /** Sets style of button to a preset size style  */
+  size?: ButtonSize
```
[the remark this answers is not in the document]
rurikoaraki: ah the size thing here was from design, so it seems like, at least on web/win32, there's a sizing token to (some?) components. I can follow up with design if this will be a consistent thing.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonSizeTestSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const svgProps: SvgIconProps = {
+    src: TestSvg,
+    viewBox: '0 0 500 500',
+  };
+  const svgIconsEnabled = ['ios', 'macos', 'win32', 'android'].includes(Platform.OS as string);
+
+  return (
+    <View style={[stackStyle, commonTestStyles.view]}>
+      {svgIconsEnabled ? <Button size="small" icon={{ svgSource: svgProps }} style={commonTestStyles.vmargin} /> : null}
```
ValentinaKozlova, line 17: Should here be 3 conditions for svgIconsEnabled? Can it be one? And Button components can be in react fragment.

ValentinaKozlova (approved):

packages/experimental/Button/src/Button.types.ts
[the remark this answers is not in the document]
rurikoaraki: Sunmin said that if components have a "size," it'll be standardized to small/medium/large. So it'll be according to design/redlines, but we should be standardizing as we move to v1.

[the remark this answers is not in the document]
rurikoaraki: So I guess... we can close the issue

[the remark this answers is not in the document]
Saadnajmi: Does that mean that there will be a separate "size" global or control token that will eventually show up?
Or is that just... tbd? I was mostly curious if the idea from design is that "size" will be a theme level thing. But yeah closing the issue is fine, thanks for following up!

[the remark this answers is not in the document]
rurikoaraki: It would be considered a control token, I don't think it would be a global. The size is basically an option for particular clients to override the default size if they want for particular components. For example win32 might want to use a medium button in certain places.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonSizeTestSection.tsx
rurikoaraki: Yeah, this can be simplified. Thanks for catching!

rurikoaraki:
Here's UWP without further layout tweaks:

image
image

rurikoaraki:
Android:

image
image
image

Color tokens haven't been updated in Android yet so the colors probably aren't correct

packages/experimental/Button/src/Button.tsx
rurikoaraki: Added comment to the function.

apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonSizeTestSection.tsx
rurikoaraki: Did some cleanup

packages/experimental/Button/src/Button.styling.ts
rurikoaraki: I extracted the constant into a function and left a comment on the function.

packages/experimental/Button/src/ButtonTokens.ts
rurikoaraki: I extracted the constant out into a function which can give different values based on platform, but this file can be shared between the platforms. Not sure if this is the best way of handling this, but for now since is the main difference, I'll start with this and see if we need to adjust later.

packages/experimental/Button/src/CompoundButton/CompoundButton.styling.ts
rurikoaraki: Refactoring done.

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -46,10 +52,10 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
         style: {
           color: tokens.color,
-          marginLeft: tokens.spacingIconContent ?? 1,
-          marginRight: -1,
-          marginBottom: 1,
-          marginTop: -1,
+          marginLeft: tokens.spacingIconContent ?? getTextMarginLeftAdjustment(),
+          marginRight: getTextMarginRightAdjustment(),
+          marginBottom: getTextMarginBottomAdjustment(),
+          marginTop: getTextMarginTopAdjustment(),
```
Saadnajmi, line 58: Could we combine these into one with a spread operator?

packages/experimental/Button/src/Button.tsx
```diff
@@ -7,14 +7,21 @@ import { stylingSettings, getDefaultSize } from './Button.styling';
...
+ * @param userProps The props that were passed into the button
+ * @returns Whether the styles that are assigned to the layer should be applied to the button
+ */
+export const buttonLookup = (layer: string, state: IPressableState, userProps: ButtonProps): boolean => {
   return (
     state[layer] ||
     userProps[layer] ||
-    layer === userProps['size'] ||
-    (!userProps.fab && layer === getDefaultSize()) ||
+    (!userProps.fab && (layer === userProps['size'] || (!userProps['size'] && layer === getDefaultSize()))) ||
```
Saadnajmi, line 24: Why the special casing around FAB? FWIW, mobile does have multiple size fab buttons, so if the goal was to only have one size fab button I would not do that.

packages/utils/styling/src/getMarginAdjustment.win32.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Adjusts the margin of text so that it's centered within the layout area of the text.
+ * Due to how Segoe UI lays out, the text does not appeared centered within the text
+ * layout area without some fuding of margins.
+ * @returns Adjustment necessary for text
+ */
+export function getTextMarginRightAdjustment(): number {
+  return -1;
+}
```
Saadnajmi, line 39: I think we could combine all 4 into one function, not sure when you would need just one of them

packages/utils/styling/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@fluentui-react-native/styling-utils",
+  "version": "0.1.0",
```
Saadnajmi, line 3: Everytime we add a new package, the matrix of dependencies gets slightly more complex, which also means more change files and more things for clients to update. I thought this mores when I was thinking about "what is the minimum set of things we could put into the platform bundle?". While I wouldn't block this review on it, it might be worth seeing if we could reconcile a bunch of these packages we have together in the future.

Saadnajmi (approved):

packages/utils/styling/src/getMarginAdjustment.win32.ts
rurikoaraki: That's true. And if someone needed to override one field they could use spread operator or something. I'll do that.

packages/experimental/Button/src/Button.tsx
rurikoaraki: Mostly lack of redlines/focusing more on web/win32 redlines. I didn't want to accidentally mess with FAB in a weird way either, so in order to scope the change I excluded it.

packages/utils/styling/package.json
rurikoaraki: Would you mind opening an issue with this thought? Tagging pat/sammy/krystal in it would be good, too, since it has to do with platform bundle complexity

packages/experimental/Button/src/Button.tsx
Saadnajmi: The FAB as is written isn't exactly to redlines anyway, since things like size tokens didn't exist when we made it. So I wouldn't worry about it _too_ much but fair!
Figma for reference: https://www.figma.com/file/ArWQiU7BAlrDGNmbrGEXbA/[Specs]-Fluent-vNext-iOS-Components?node-id=2024%3A497

packages/utils/styling/package.json
Saadnajmi: That sounds good! I shall do that later today =)

packages/utils/styling/src/getMarginAdjustment.win32.ts
```diff
@@ -1,39 +1,11 @@
...
-/**
- * Adjusts the margin of text so that it's centered within the layout area of the text.
- * Due to how Segoe UI lays out, the text does not appeared centered within the text
- * layout area without some fuding of margins.
- * @returns Adjustment necessary for text
- */
-export function getTextMarginLeftAdjustment(): number {
-  return 1;
-}
+const margins = { marginTop: -1, marginBottom: 1, marginLeft: 1, marginRight: -1 };
```
Saadnajmi, line 1: Another thought (not review blocking): 
I wonder if information like this could live in a specific theme, rather than globally through an extra package. I have this thought because there's no guarantee that win32 means Segoe UI: you only really know you have Segoe if you have a theme that uses Segoe. With the apple theme, I had [component specific overrides](https://github.com/microsoft/fluentui-react-native/blob/78a02005cdd60b263b1c3f9ecec5187e9a4a9931/packages/theming/apple-theme/src/appleTheme.macos.ts#L15). With global/control/alias tokens, I feel like the role of themes is changing so I don't quite understand if doing stuff like component specific overrides makes sense anymore.

rurikoaraki: Hmm yeah, that makes sense. I'll open an issue to track this.

PR #1036, opened 2021-09-28
Delete NativeButton

chiuam:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

A few reasons for this change:

- The FUA crew is in the process of replacing the UIKit Button with the vNext Button using SwiftUI, so we should remove the dependency before it gets deprecated.
- Debatable but the JS button comes close to the native button (thanks to the recent efforts in adding tokens and styling), we can experiment with using the JS Button instead.

Not sure if we should keep the package for when we integrate the vNext button, but I'd say let's get rid of it so that we don't have a separate package for the same control. 

### Verification
yarn build

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

rdeassis (approved):

Saadnajmi (approved):

Saadnajmi:
Don't forget the Nuspec!

PR #1039, opened 2021-09-29
Add props to buildProps

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

We've been running into issues with slotProps/buildProps in compose v2. Specifically, it doesn't really allow for us to take the outer component's props into account when building slot props. This makes cases where we want to nest slots harder to do, since much of the outer portions of the render functions of nested slots rely on props that we wouldn't be able to pass along until we get to the inner portion of the render function of the outer component, thus making them unavailable for the inner slot.

Although tokensThatAreAlsoProps exists, for some of the props, like passing onClick along from an outer component to a slot that's a FURN Button, it didn't really make sense to make that a token. So I think the "right" change is to allow props to be passed into buildProps and be used to create slotProps, so that certain props are available for outer render time.

### Verification

Tests have been run and a new test has been added. Verified with the menu button changes that Valentyna is working on that this seems to resolve the issues we've been running into with onClick, and should work for propagating primary/ghost.

### Pull request checklist

- [ ] Automated Tests - working on it, getting this out to make sure that there's no large issue with this change.
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-framework-8d280009-d6bf-4d13-935f-c60be74ca8bb.json
+++ b/change/@fluentui-react-native-framework-8d280009-d6bf-4d13-935f-c60be74ca8bb.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add props to build props",
+  "packageName": "@fluentui-react-native/framework",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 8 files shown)


rurikoaraki:
Note to self: go over documentation and see if anything needs to be updated

lenahong:
So the introduction of the props parameter is specifically for props that are solely props and don't make sense as tokens to be handled by buildProps. Am I understanding correctly?

rurikoaraki:
> So the introduction of the props parameter is specifically for props that are solely props and don't make sense as tokens to be handled by buildProps. Am I understanding correctly?

Yeah, basically. There is redundancy with tokensThatAreAlsoProps in terms of functionality, but declaring some things as tokens didn't make sense from a conceptual standpoint.

ValentinaKozlova (approved):

rurikoaraki:
Just realized the cache keys aren't taking in the props into account, so working on that

PR #1045, opened 2021-09-30
Downgrade Text component

### LINKED ISSUE #1028: Windows Tester/Experimental Text: Cannot redefine non-configurable property 'margin'
### Environment Information

- **Package version(s)**: 
  - "@fluentui-react-native/experimental-text": "0.6.20"
  - "react-native-windows": "^0.64.0"
  - "react-native": "^0.63.2",
  - "@fluentui-react-native/tester": "^0.43.3",
- **OS version**: (fill this out if relevant)

- **Platform**:
  - [ ] iOS
    - **Xcode version**:
  - [ ] macOS
    - **Xcode version**:
  - [ ] win32
  - [x] windows
  - [ ] android

When clicking on some of the test pages in the Windows Tester (like Checkbox Test for example), an error related to Experimental Text pops up:

image

image

#### Actual behavior:

Error regarding non configurable property 'margin' pops up

#### Expected behavior:

Error does not pop up

Requested priority: High

Products/applications affected: (if applicable)

rurikoaraki:
Fixes #1028

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

When clicking on a test page in the tester, an error pops up from RN that tells us that we're changing the value of a prop at a time when we shouldn't be in the rendering phase. This is an issue that is specific to the UWP platform, because we're still using Chakra. It seems to be a timing issue - it's not something that we see causing issues with other engines on other platforms.

This error also occurs in the Separator component, but we worked around that by skipping the prop that was causing problems. But unfortunately since the style is integral to the Text component, we can't really do that for the Text component.

To unblock the Tester I'm going to downgrade the Text component to the non-experimental version.

### Verification

Tester works fine when I downgrade the Text component

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester/FluentTester.tsx
@@ -1,7 +1,7 @@
 import { Theme } from '@fluentui-react-native/framework';
 import { FocusTrapZone, Separator, StealthButton } from '@fluentui/react-native';
 import { Button } from '@fluentui-react-native/experimental-button';
-import { Text } from '@fluentui-react-native/experimental-text';
+import { Text } from '@fluentui-react-native/text';
 import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
 import * as React from 'react';
 import { ScrollView, View, Text as RNText, Platform, SafeAreaView, BackHandler } from 'react-native';


apps/fluent-tester/src/FluentTester/FluentTester.tsx
```diff
@@ -1,7 +1,7 @@
 import { Theme } from '@fluentui-react-native/framework';
 import { FocusTrapZone, Separator, StealthButton } from '@fluentui/react-native';
 import { Button } from '@fluentui-react-native/experimental-button';
-import { Text } from '@fluentui-react-native/experimental-text';
+import { Text } from '@fluentui-react-native/text';
```
ksiler, line 4: Is this change only needed here? Don't you need to downgrade text in the other test pages too?

rurikoaraki: I just tested the other pages and they seemed to be working for me. This change specifically targets the header on every page, so it does help with all pages.

rurikoaraki: I'll take another look to see if this can be fixed in a different way than downgrading the Text for just the test header.

ksiler: Okay, sounds good. I see how downgrading the Text header works for now (it's the same solution I came to when working in the UWP tester recently) but I recall still seeing the issue pop up intermittently after that 😥 (particularly in the Tabs Experimental test page). That's why I wonder if we need to downgrade Text in all test pages.

Saadnajmi:
Curious: Did the issues arrive when we moved Text to compressible over compose?

rurikoaraki:
> Curious: Did the issues arrive when we moved Text to compressible over compose?

Yeah, I don't think this was an issue before we moved to compressible

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/FluentTester.tsx
rurikoaraki: @ksiler if you're ok with this i'd like to check this change in because it makes the tester usable in UWP, and then tackle the intermittency and track it with #1046

PR #1054, opened 2021-10-06
Automated E2E Testing Reliability Improvements

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

In regards to [Reduce the chance of Intermittent failures within our CI for E2E Testing (Win32/Windows)](https://office.visualstudio.com/OC/_workitems/edit/5480039).

Currently, there are some intermittent timing issues within our CI regarding E2E testing. It seems that sometimes, WebDriverIO doesn't have the time it needs to search the given test app for specific components - leading to the test surpassing the Timeout and causing a failure.

Here, I'm increasing the timeouts to lower the risk of these failures happening. I'm also increasing the number of times the spec file retries from 1 to 3. Often times, if we do get a flaky test, it fixes itself after the 1st try. This will decrease the amount of false-negatives our engineers see. The net-time lost is only about 3 mins per spec run, so I believe it's very worth it.

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/win32/wdio.conf.js
+++ b/apps/win32/wdio.conf.js
@@ -167,9 +167,12 @@ exports.config = {
    */
   afterTest: function (test) {
     // if test passed, ignore, else take and save screenshot.
-    if (test.passed) {
-      return;
-    }
+    /* UPDATE: I want to take screenshots after every test to help gauge certain CI failures.
+     * I will re-enable this once I am able to determine what's causing some CI failures
+     */
+    // if (test.passed) {
+    //   return;
+    // }
 
     // get current test title and clean it, to use it as file name
     const fileName = encodeURIComponent(test.title.replace(/\s+/g, '-'));
--- a/apps/windows/wdio.conf.js
+++ b/apps/windows/wdio.conf.js
@@ -151,9 +151,12 @@ exports.config = {
    */
   afterTest: function (test /*, context*/) {
     // if test passed, ignore, else take and save screenshot.
-    if (test.passed) {
-      return;
-    }
+    /* UPDATE: I want to take screenshots after every test to help gauge certain CI failures.
+     * I will re-enable this once I am able to determine what's causing some CI failures
+     */
+    // if (test.passed) {
+    //   return;
+    // }
 
     // get current test title and clean it, to use it as file name
     const fileName = encodeURIComponent(test.title.replace(/\s+/g, '-'));


apps/win32/wdio.conf.js
```diff
@@ -58,6 +58,10 @@ exports.config = {
     defaultTimeoutInterval: jasmineDefaultTimeout,
   },
 
+  // The number of times to retry the entire specfile when it fails as a whole.
+  // Adding an extra retry will hopefully reduce the risk of engineers seeing a false-negative
```
ksiler, line 62: How many false-negatives are we currently seeing? Do we have plans to investigate why they're happening?

samuelfreiberg: Yes, that data is here: [https://office.visualstudio.com/OC/_workitems/edit/5480023](https://office.visualstudio.com/OC/_workitems/edit/5480023)

ksiler: Thanks! Just wanted to know if we had plans to investigate and tackle some of the root causes. I don't like the idea of rerunning the tests in hopes of it passing during a second or third run. Is this a permanent change or a temporary measure to help reduce the amount of failures we're seeing until we can find a better solution?

ksiler (approved):

apps/win32/wdio.conf.js
samuelfreiberg: I definitely understand where you're coming from, and it's a valid point. When running E2E tests that require automated interaction with apps, there's often some intermittent timing issues (especially when your running on a CI machine and not locally). These can come from slower than expected app interaction. Or, in this case, the Appium server might be running a bit slow on this CI computer. Also, when running these tests, there's a lot of communication between the 3 softwares (WebDriverIO as the client, Appium as the server, and WinAppDriver which is the driver). With all this communication, there might be some timing issues. 

Reasons are not exactly known yet, but increasing the timeouts is our best shot at fixing some of these (they probably should have been higher in the first place, but I implemented this a long time ago without much knowledge of possible errors :) ) 

However, I enabling screenshots after every test to try and pinpoint specific reasons for failure, hopefully then I can start tackling specific scenarios.

PR #1069, opened 2021-10-14
Get changelog for main fluentui/react-native package to have more detail

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change changes the beachball config to treat the fluentui/react-native package as a group of packages of the packages it re-exports. This will hopefully make it so that there's more detail in the changelog for the changes that cause the package to bump in version, which we can then use to create release notes.

The main challenge with release notes with FURN has been how to make the release notes fairly accessible (i.e. don't make users run around to individual packages to piece together what changes are included in a version bump of @fluentui/react-native) without including changes that aren't included in the package (ex. theming or framework changes don't directly affect the @fluentui/react-native package).

The beachball grouping config seems to be the solution here as it will include the changes of the packages that I list as part of the group into the larger package's changelog. Then I can use code similar to RNW to read out changes that affect the larger package from the changelog to generate release notes.

### Verification

Used `yarn beachball bump` which is the local version of `beachball publish` to test the output. The output of the @fluentui/react-native changelog was more verbose with these settings.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/beachball.config.js
+++ b/beachball.config.js
@@ -4,7 +4,7 @@ const path = require('path');
 module.exports = {
   disallowedChangeTypes: ['major'],
   hooks: {
-    prepublish: packagePath => {
+    prepublish: (packagePath) => {
       const packageJsonPath = path.join(packagePath, 'package.json');
       const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
       if (packageJson.onPublish) {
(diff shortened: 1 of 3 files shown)


beachball.config.js
```diff
@@ -4,14 +4,40 @@ const path = require('path');
...
+          '@fluentui-react-native/persona',
+          '@fluentui-react-native/persona-coin',
+          '@fluentui-react-native/pressable',
+          '@fluentui-react-native/radio-group',
+          '@fluentui-react-native/separator',
+          '@fluentui-react-native/text',
+          '@fluentui-react-native/interactive-hooks',
+          '@fluentui-react-native/menu-button',
+          '@fluentui-react-native/tabs',
+        ],
```
Saadnajmi, line 38: This is just all the packages in @fluentui/react-native? How do we uh.. make sure this stays in sync with the package.json? I.E: What if we add a new package there but forget to add it to this config?

rurikoaraki: Yeah I'll take some time to think about automating this. I did add a comment to the index.ts file but you're right, that's probably not enough here.

rurikoaraki: Though I'm a little not hopeful because I can only feed regex's into here (input is string or string[], so it's not expecting a function which can generate an array of strings)

beachball.config.js
```diff
@@ -4,14 +4,40 @@ const path = require('path');
...
+          '@fluentui-react-native/checkbox',
+          '@fluentui-react-native/contextual-menu',
+          '@fluentui-react-native/focus-trap-zone',
+          '@fluentui-react-native/focus-zone',
+          '@fluentui-react-native/link',
+          '@fluentui-react-native/persona',
+          '@fluentui-react-native/persona-coin',
+          '@fluentui-react-native/pressable',
+          '@fluentui-react-native/radio-group',
+          '@fluentui-react-native/separator',
```
samuelfreiberg, line 33: Shimmer? Or is this just non-experimental controls?

beachball.config.js
Saadnajmi: I didn't notice the comment. I think that should be good enough actually. I don't know it could be more automated than this.

beachball.config.js
Saadnajmi: I will say that I think the set included is somewhat arbitrary, and when we go to v1, we should probably revisit "what packages are in the main @fluentui/react-native set are exported"

Saadnajmi (approved):

beachball.config.js
rurikoaraki: Took some time to get it to work but I have the file parsing the index file to get the reexports now

Issue #1035, opened 2021-09-27
Replace uses of "UseAsPressable" with "usePressableState"


Saadnajmi:
Per this [discussion](https://github.com/microsoft/fluentui-react-native/pull/863#discussion_r684525586) in #863 , we should replace all of our uses of "useAsPressable" / View slots with "usePressableState" / Pressable slots.

As of the time of writing, here are all the things that use that hook:
- Button
- Checkbox 
- ContextualMenu
- Link
- RadioButton
- Tabs / TabItem

PR #1070, opened 2021-10-15
[Draft] Switch Button, Checkbox, and RadioButton from View to Pressable

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### TODO
- Test that Android indeed gets the ripple for free with Core Pressable, and if so remove our optional ripple slot from the android button
- Switch over the `onClick` prop of experimental button to `onPress`, to match Pressable. 

### Description of changes

This will handle part of #1035 

Before we were using React Native 0.63 (where Pressable was added), we ported over Pressable to our own repo, and created the hook `useAsPressable` to get similar functionality. Now that we are on RN 0.63, we should just use the Pressable coming from React Native Core (which has benefits like automatically adding the ripple for android), and remove our port. PR #863 introduced the hook `usePressableState` to help with this. 

The original plan was to switch all components that called `useAsPressable` to instead call `usePressableState`, switch all the root slots from `View` to `Pressable`, and then delete our port of Pressable. 

Unfortunately, after investigation, I found out that the RN Core Pressable does not support hover event handlers (I filed https://github.com/facebook/react-native/issues/32406 to track this), nor does it support keyboard focus events on win32. As such, we will continue to have to use our own port of Pressable for at least win32.

 As such the new plan is to have our package `@fluentui-react-native/pressable` export the RN Core Pressable for all platforms except win32, which will continue to use our port of `useAsPressable`. This way, we can get the benefits of RN Core press ability (like the ripple on Android), while also keeping our hover/focus/press functionality on all platforms. Once the proper upstream changes have been made in react-native and/or it's desktop ports to Pressable, we can revisit this.

This PR switches over Button, Checkbox, and RadioButton. I had issues with porting ContextualMenu, Link, and Tabs/tabItem, so I will leave those for a separate PR. 

### Verification

I tested on macOS that we still get the proper hover, focus, and press effects. I need to test on other platforms, namely win32 and android to make sure nothing else broke there. Ideally, there should be functionality change. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/RadioGroup/src/__snapshots__/RadioButton.test.tsx.snap
```diff
@@ -26,7 +26,8 @@ exports[`RadioButton component tests RadioButton default 1`] = `
   onBlur={[Function]}
   onClick={[Function]}
   onFocus={[Function]}
-  onPress={[Function]}
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
   onResponderGrant={[Function]}
```
Saadnajmi, line 31: not sure why onPress is disappearing from the snapshot. The radio buttons still seem to update for me onPress. This probably requires investigation.

Saadnajmi: Actually.. I think it's because onPress got absorbed by Pressability, and doesn't show up as a prop on the final rendered component anymore.

Saadnajmi:
Closing as this will be split into different PRs

PR #1091, opened 2021-10-20
Layout test page

rurikoaraki:
Just putting this here for people to play with and understand borders and padding....

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Layout/LayoutTestSection.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Layout/LayoutTestSection.tsx
@@ -1,0 +1,45 @@
+import { Button } from '@fluentui-react-native/experimental-button';
+import { Text } from '@fluentui-react-native/experimental-text';
+import * as React from 'react';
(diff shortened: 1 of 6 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/Layout/LayoutTestSection.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button } from '@fluentui-react-native/experimental-button';
+import { Text } from '@fluentui-react-native/experimental-text';
+import * as React from 'react';
+import { Picker, View } from 'react-native';
```
Saadnajmi, line 4: As an FYI, this is deprecated and is/will be removed from react-native altogether. I'm weary about introducing another use of it

rurikoaraki: Yeah, if we check this in I'll change this to something else. This just seemed like the easiest way to do what I wanted to achieve, which was to have a way to change borderWidth/padding and see the effect is has on our button

PR #1100, opened 2021-10-28
Callout macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds a native module for macOS Callout. This is a fairly large change, be warned.
Implementing Callout also means we have a functional ContextualMenu, but I would not call it done, there's a lot of testing / bug fixing I need to do before I can declare that. 
The focusOnMount/focusOnContainer props are mostly windows specific, so I did not implement those. It might be worth going through the types and marking "win32 only" and such.

### Verification

I tested showing the Callout with a target view, anchor Rect, and as a cotextual menu. I tested that it behaves properly when at a border of the screen. 

Screen Shot 2021-11-03 at 12 26 28 PM

https://user-images.githubusercontent.com/6722175/140180045-1902bf97-c207-4d38-ab9e-3b57a3a5393e.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 import * as React from 'react';
-import { Text, View, Switch } from 'react-native';
+import { Text, View, Switch, Platform } from 'react-native';
```
HeyImChris, line 3: Why do we need `Platform`?

apps/fluent-tester/src/FluentTester/testPages.macos.ts
```diff
@@ -33,11 +35,31 @@ export const tests: TestDescription[] = [
...
+    component: ContextualMenuTest,
+    testPage: HOMEPAGE_CONTEXTUALMENU_BUTTON,
+  },
   {
     name: 'Experimental Button',
     component: ExperimentalButtonTest,
     testPage: HOMEPAGE_BUTTON_BUTTONEXPERIMENTAL,
   },
+  {
+    name: 'Checkbox Test',
```
HeyImChris, line 54: Does it make sense to have the list be alphabetical so there's some organization to it?

packages/components/Callout/FRNCallout.podspec
```diff
@@ -3,19 +3,17 @@ require 'json'
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
 
 Pod::Spec.new do |s|
-  s.name         = 'FRNRadioButton'
```
HeyImChris, line 6: I'm a bit confused on this change. You're renaming radio button to be a callout?

packages/components/Callout/macos/Callout.swift
```diff
@@ -0,0 +1,310 @@
...
+				preconditionFailure("Invalid target react tag")
+			}
+			anchorView = targetView
+			updateCalloutFrameToAnchor()
+		}
+	}
+
+	@objc public var anchorRect: NSRect = .null {
+		didSet {
+			updateCalloutFrameToAnchor()
```
HeyImChris, line 19: Not trying to start a flame war but honest question, is this GitHub making the tabs look enormous or are those some Xcode settings we can tone down? Does this align with other .swift files in the repo?

packages/components/Callout/macos/Callout.swift
```diff
@@ -0,0 +1,310 @@
...
+		let calloutOrigin: NSPoint = {
+			var origin = NSPoint()
+
+			switch(directionalHint) {
+			case .minX:
+				origin.x = NSMinX(anchorScreenRect) - calloutFrame.size.width
+				origin.y = NSMaxY(anchorScreenRect) - calloutFrame.size.height
+				break
+			case .minY:
+				// TODO is this right?
```
HeyImChris, line 185: Probably shouldn't have a floating TODO in code (yes I see the irony as I probably added this in the original copy of this menu code). Also why no break in this case? Wouldn't the .maxX case just overwrite it anyway?

packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -0,0 +1,43 @@
...
+			mouseEventMonitor = NSEvent.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { (event) -> NSEvent? in
+					if (event.window != self) {
+							self.windowLifeCycleDelegate?.didDetectHitOutsideCallout(calloutWindow: self)
+					}
+					return event
+			})
+	}
+	
+	deinit {
+			NotificationCenter.default.removeObserver(self)
```
HeyImChris, line 34: Same tabbing issue- these tabs just seem monstrously wide but maybe that's just GH formatting being odd

packages/components/Callout/macos/RCTCalloutManager.h
```diff
@@ -0,0 +1,5 @@
+#import <React/RCTAssert.h>
```
HeyImChris, line 1: There's no class interface?

Why import all these files if we don't use them?

HeyImChris (changes requested):

packages/components/Callout/macos/Callout.swift
Saadnajmi: There's a personal GitHub setting I think where you can set the tab width. I have it set to 4. If you prefer, I can convert all my tabs to spaces and maybe we can add some listing to do that automatically

packages/components/Callout/FRNCallout.podspec
Saadnajmi: I think git got confused here, I'm making a new file but git thinks I'm renaming the old one. I'll see if I can make that go away

packages/components/Callout/macos/Callout.swift
Saadnajmi: You can also add a .editorconfig file at the root of the repo to tell GitHub how many spaces are in a tab. I can do that too with this change haha

packages/components/Callout/macos/Callout.swift
Saadnajmi: Ah, good catch. FWIW I tested this and this should be right. It's not in the original copy, the original copy just asserts we don't support minY.

As it turns out, Swift switch statements don't fall through, so we don't need breaks anywhere. I should actually remove them from the other statements

packages/components/Callout/macos/RCTCalloutManager.h
Saadnajmi: We need most of these. They're used in the Swift code (touch handler, UIManager, etc.). I can remove the import of RCTAssert though

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: It's sorted before it's displayed, but fair

apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
Saadnajmi: line 121. Basically, custom fonts don't work on macOS yet, so I had to do a platform check in some test code

apps/fluent-tester/src/FluentTester/testPages.macos.ts
Saadnajmi: fixed

packages/components/Callout/macos/CalloutWindow.swift
Saadnajmi: Yeah this was wrong on my end and I fixed it

Saadnajmi: But also.... I wonder if I even need to make this call.

Saadnajmi: Turns out as of macOS 11, I don't!

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -68,6 +69,59 @@ export const appleComponents = {
...
+    tokens: {
+      borderRadius: 5,
+    },
+  },
+  ContextualMenuItem: {
+    root: {
+      style: {
+        borderRadius: 5,
+      },
+    },
```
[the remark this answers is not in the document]
Saadnajmi: Done!

packages/components/Callout/src/Callout.tsx
```diff
@@ -32,6 +32,9 @@ export const Callout = compose<ICalloutType>({
         ref: calloutRef,
         ...(nativeTarget && { target: nativeTarget }),
         ...rest,
+        style: {
+          position: 'absolute',
+        },
```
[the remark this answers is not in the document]
Saadnajmi: Made it macOS specific. No longer relevant

packages/components/Callout/FRNCallout.podspec
Saadnajmi: Yeah so what happened is this PR is so old it was back when we still had a separate NativeRadioButton package (instead of just having it in the same radio button package), and while I was working on it that package (and it's podspec) got deleted. Git interpreted that as me renaming the podspec.

packages/components/Callout/macos/Callout.swift
```diff
@@ -289,15 +289,37 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
 		window.level = .popUpMenu
+		window.backgroundColor = .clear
+		window.isMovableByWindowBackground = false
+		
+		let visualEffect = NSVisualEffectView()
+		visualEffect.translatesAutoresizingMaskIntoConstraints = false
+		visualEffect.material = .menu
+		visualEffect.state = .active
+		visualEffect.wantsLayer = true
+		visualEffect.layer?.cornerRadius = 5.0
```
HeyImChris, line 306: Is this the default NSButton corner radius?

packages/components/Callout/macos/Callout.swift
```diff
@@ -289,15 +289,37 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
+
+
```
HeyImChris, line 308: nit- did you mean to have a double newline inside this method?

HeyImChris (approved):

packages/components/Callout/macos/RCTCalloutManager.h
```diff
@@ -0,0 +1,4 @@
+#import <React/RCTTouchHandler.h>
```
HeyImChris, line 1: Why are all of these in the header file? Can we just move them out to the .mm where they're needed? Also no interface needs to be defined here?

packages/components/Callout/macos/Callout.swift
Saadnajmi: Default menu corner radius, which I found by grabbing the macOS design templates and importing them into sigma: https://developer.apple.com/design/resources/

Saadnajmi: I suppose I should pull that out as a constant since it's a magic number

packages/components/Callout/macos/RCTCalloutManager.h
Saadnajmi: There is no .mm file, there are swift files, and I can't import objective C headers into Swift headers, except by adding them here.

packages/components/Callout/macos/Callout.swift
Saadnajmi: No, I'll fix that

packages/components/Callout/macos/Callout.swift
Saadnajmi: It's not easy to get the template into figma and find the corner radius, so here's a screenshot of me doing so. You can see "5" on the right

MicrosoftTeams-image

PR #1116, opened 2021-11-02
Fix disabled state for button

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Seems that the migration to the new framework dropped defining accessibilityState for the Button. Putting it back to ensure that the button announces the right state and acts properly. Focus now skips disabled experimental buttons.

### Verification
Played around with keyboarding. Keyboard focusable is off according to accessibility checker

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Button/src/useButton.ts
+++ b/packages/experimental/Button/src/useButton.ts
@@ -17,6 +17,7 @@ export const useButton = (props: ButtonProps): ButtonState => {
       accessibilityRole: 'button',
       onAccessibilityTap: props.onAccessibilityTap || props.onClick,
       accessibilityLabel: props.accessibilityLabel || props.content,
+      accessibilityState: { disabled: props.disabled },
       focusable: true,
       ref: useViewCommandFocus(componentRef),
       onKeyUp: onKeyUp,
--- a/packages/experimental/Button/src/useButton.ts
+++ b/packages/experimental/Button/src/useButton.ts
@@ -1,11 +1,12 @@
 import * as React from 'react';
 import { useAsPressable, useKeyCallback, useOnPressWithFocus, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
 import { ButtonProps, ButtonState } from './Button.types';
+import { memoize } from '@fluentui-react-native/framework';
 
 export const useButton = (props: ButtonProps): ButtonState => {
   // attach the pressable state handlers
   const defaultComponentRef = React.useRef(null);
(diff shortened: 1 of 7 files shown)


ValentinaKozlova (approved):

ksiler (approved):
Thanks for fixing this!

Saadnajmi (approved):

Saadnajmi:
Good catch!

packages/experimental/Button/src/useButton.ts
```diff
@@ -17,10 +18,16 @@ export const useButton = (props: ButtonProps): ButtonState => {
...
       onKeyUp: onKeyUp,
     },
     state: pressable.state,
   };
 };
+
+const getAccessibilityState = memoize(getAccessibilityStateWorker);
+function getAccessibilityStateWorker(disabled: boolean) {
+  return { disabled: disabled };
+}
```
Saadnajmi, line 33: Why do we need this?

rurikoaraki: To avoid recreating the object on every render if nothing has changed. There's a test for button that checks to make sure that if you rerender the button with no prop changes, there aren't any changes in the props using shallow compare (which would cause a rerender). This was needed to pass that test.

Saadnajmi: I see... could we maybe write a hook that is something like `useButtonAccessbilityProps` or something that handles all the accessibility props? It feels a little weird to have this new pattern for this one prop to achieve the type of caching I thought hooks were supposed to help us avoid. Not a review blocking comment

rurikoaraki: Sure, I'd be happy to try it out. I did try using some hooks (I think?) but those wouldn't pass the test (React.useMemo, React.useCallback).

Saadnajmi: Yeah I'm mostly guessing, obviously you are the one who is actually running the tests and such. I suppose I should look into this worker/memoize pattern

rurikoaraki: I tried wrapping useMemo in another function but it still wouldn't pass the test. I think for now I'll keep this as is. We can refactor and make it look more like a hook once we share this function with other components.

PR #1117, opened 2021-11-03
Integrating accessibilityLabel functionality for MenuButton

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

MenuButton has the prop accessibilityLabel, but it wasn't doing anything with it. With this change, we forward the accessibilityLabel to the inner Button component so that screen readers can properly read it.

### Verification

Manual with Narrator

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Do we also need to do this for experimental menu button?
Also, do we need to do something like #1116 for menu button?

ksiler:
> Do we also need to do this for experimental menu button?

Looks like!

> Also, do we need to do something like #1116 for menu button?

Looks like MenuButton is passing props like `disabled` to Button so #1116 should apply.

samuelfreiberg:
> Do we also need to do this for experimental menu button? Also, do we need to do something like #1116 for menu button?

Good catch, adding now

ksiler (approved):

packages/components/MenuButton/src/MenuButton.tsx
```diff
@@ -72,6 +72,7 @@ export const MenuButton = compose<MenuButtonType>({
       startIcon,
       componentRef: stdBtnRef,
       onClick: toggleShowContextualMenu,
+      accessibilityLabel,
```
Saadnajmi, line 75: I wonder if we should just do something like ...props to avoid this happening in the future.

Saadnajmi (approved):

packages/components/MenuButton/src/MenuButton.tsx
ksiler: That's a good suggestion and I don't see any issues with it. I think I'd prefer that.

samuelfreiberg: I'd agree. This should probably be the norm for our components moving forward. I'll change it

ValentinaKozlova (approved):

packages/components/MenuButton/src/MenuButton.tsx
samuelfreiberg: Changed in newest commit

PR #1120, opened 2021-11-04
Use "FRN" prefix over "RCT" for macOS Callout native code

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On Apple platforms, we often prefix objective C classnames with a 3 letter prefix because objective C has a global namespace, and we want to avoid naming clashes. Thus, each project usually chooses one, and React Native has "RCT" (I.E: RCTView, RCTBridge, etc..). We don't want to use "RCT" for our own native modules, because that is the React Native namespace. Instead, we have elected to use "FRN" for FluentUI React Native. 

Let's update Callout to use FRN over RCT, and make the appropriate changes in JS to pick up the change.

### Verification

The project building should be enough. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/src/Callout.tsx
```diff
@@ -3,12 +3,15 @@ import { backgroundColorTokens, borderTokens } from '@fluentui-react-native/toke
...
+import { findNodeHandle, Platform } from 'react-native';
 import { settings } from './Callout.settings';
 import { calloutName, ICalloutProps, ICalloutSlotProps, ICalloutType } from './Callout.types';
 import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
 
-const RCTCallout = ensureNativeComponent('RCTCallout');
+const NativeCalloutView = Platform.select({
+  macos: ensureNativeComponent('FRNCallout'),
+  default: ensureNativeComponent('RCTCallout'), // win32
+});
```
Saadnajmi, line 14: So.. I guess Platform.Select doesn't include win32 as a key.

chiuam (approved):

PR #1125, opened 2021-11-04
Win32 E2E Testing - Implementation Changes + Accessibility Testing Additions

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

**Win32 Phase 1 of 2**
_I'm splitting this PR into phases because the original PR I had planned to release had changed 105 files. A PR that big is unrealistic to review and is prone to mistakes slipping through. The difference between the phases is simply which components I'm implementing the E2E testing for. The next Phase will have all the controls not in this one. And although there are 50+ file changed in this one, there's a lot of small/repetitive changes and file name changes that inflate the number._

The PR has some major changes to the implementation of E2E testing on Win32 ([UWP to come](https://office.visualstudio.com/OC/_workitems/edit/5550799)). It also some Accessibility Testing for our controls.

1. Add tests to validate `accessibilityRole` and `accessibilityLabel` attributes ([in applicable controls, see status here](https://microsoft-my.sharepoint-df.com/:x:/p/safreibe/EWXUoASFsylLvvPRKPbnylEBEBEK92H1amn2e-CEoShBBg?e=s8A3Pb))
2. Change E2E Testing Infrastructure to a per-component basis.
With this change, each FURN component will now have it's own testing data. As the scope of our testing grows, the more component-specific tests we will have.
**Each component will have its own respective [Page Object/Spec files.](https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/E2E#create-a-new-page-object).**
**Additionally, this will make setting up testing cross-plat much easier. We already have the skeleton of the structure we want, now it's just a matter of making platform-specific files.**

**Smaller Changes**

- Improvements to _BasePage.ts_. This is an **extremely important** file, as every component's page object extends this. I've added more in-depth comments and a wider range of UI interaction functions. The goal is to have this class extendable cross-plat
- Creating E2E Testing specific test files for each component. On each test page in our app, there will now be a section called "E2E Testing" that will only be used by this testing infrastructure. This will separate the actual examples for partners and the scenarios we want to run with out automation.
- Rename BootTestPage.ts to NavigateAppPage.ts. This name makes more sense, as this class is responsible for navigating and clicking on each test page. I'm also moving this to the _/common/_ folder of E2E testing because all components will use this.
- Rename all page object files (let's use Button as an example) from , _ButtonTestPage.ts_ to _ButtonPageObject.ts_. The new name more accurately reflects that the class is a Page Object class rather than an actual test page in our app.
- In TabsItem and Button, we were incorrectly passing the `testID` prop to it's _content_ slot. This assigned that slot (a Text component) the testID, which we don't want. We want to interact with the Button/TabItem control itself, so we want to pass it to the root (which is already done via `...rest`
- Add some extra documentation on Page Objects

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki (approved):

apps/fluent-tester/src/E2E/common/consts.ts
```diff
@@ -0,0 +1,15 @@
+/* Accessibility Role Constants - https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/UI/Accessibility/ */
```
ksiler, line 1: The link here doesn't resolve for me. Where is it from?

samuelfreiberg: Whoops, missed the ending of the link. I will update in my newest commit

apps/fluent-tester/src/E2E/common/consts.ts
```diff
@@ -1,4 +1,4 @@
-/* Accessibility Role Constants - https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/UI/Accessibility/ */
+/* Accessibility Role Constants - https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/Accessibility/index.html */
```
ksiler, line 1: I'm not really sure what this link is for. Is it related to the accessibility testing api you used? If you're trying to show what these roles/control types are (or show the constants that are used), I'd recommend using the Microsoft Docs website: https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-controltypesoverview#current-ui-automation-control-types or https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-controltype-ids

PPatBoyd (approved):

apps/fluent-tester/src/E2E/common/consts.ts
samuelfreiberg: You're right. This is actually the old link I was using that's not valid anymore. I'll update it with one of the ones above Thanks!

ksiler (approved):

PR #1130, opened 2021-11-08
Adding macOS brand tokens to FURN token pipeline

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
This change adds mac tokens to the token pipeline to so that we can use them in our JS components. Since the brand colors for mac & iOS are still WIP, let's copy what win32 uses in the meantime. 

### Verification
Ran `build-tokens` script to generate alias tokens for mac on a windows machine
`yarn build` succeeded 
### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-tokens/src/global-macos/reactnative/tokens-aliases.json
```diff
@@ -0,0 +1,76 @@
...
+		"fontWeight": "400"
+	},
+	"display": {
+		"fontFamily": "\"Segoe UI\", Roboto, \"SF Pro Text\", \"Helvetica Neue\", Helvetica, Arial, sans-serif",
+		"fontLineHeight": 92,
+		"fontSize": 68,
+		"fontWeight": "600"
+	},
+	"headline": {
+		"fontFamily": "\"Segoe UI\", Roboto, \"SF Pro Text\", \"Helvetica Neue\", Helvetica, Arial, sans-serif",
```
Saadnajmi, line 35: This isn't used for macos yes?

Saadnajmi (approved):

Saadnajmi:
If we're copying what win32 is using, why do we need to have a separate . macos file? Do we expect the tokens to change in the future?

packages/theming/theme-tokens/src/global-macos/reactnative/tokens-aliases.json
chiuam: Sorry, adding this file was a mistake ._.

chiuam:
> If we're copying what win32 is using, why do we need to have a separate . macos file? Do we expect the tokens to change in the future?

@rurikoaraki and I had talked about this offline, we expect the tokens to change in the future so decided to have a separate file for mac from the get go.

rurikoaraki:
Just wondering if you're expecting to add alias tokens too? Our components reference alias tokens instead of global tokens directly.

packages/theming/theme-tokens/src/pipeline-input/token-input-light.macos.json
```diff
@@ -0,0 +1,134 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/fluentui-token-pipeline/main/docs/token.schema.json",
+  "Meta": {
+    "FluentUITokensVersion": 0
+  },
+  "Global": {
```
rurikoaraki, line 6: Since this is a global I would expect to see it in a global input file. Maybe move to the macos brand json file and just rename it to token-input-macos?

scripts/build-tokens.js
```diff
@@ -33,6 +33,20 @@ async function run() {
...
   fs.unlinkSync(path.join(process.cwd(), 'packages/theming/theme-tokens/src/global-win32/reactnative/tokens-controls.json'));
 
+  console.log('Generating macOS global tokens...');
+  child_process.execSync(
+    'yarn transform-tokens --in ./packages/theming/theme-tokens/src/pipeline-input/token-input.json --in ./packages/theming/theme-tokens/src/pipeline-input/token-input-brand.macos.json --out ./packages/theming/theme-tokens/src/global-macos --p reactnative',
+  );
+  fs.unlinkSync(path.join(process.cwd(), 'packages/theming/theme-tokens/src/global-macos/reactnative/tokens-aliases.json'));
+  fs.unlinkSync(path.join(process.cwd(), 'packages/theming/theme-tokens/src/global-macos/reactnative/tokens-controls.json'));
+
+  console.log('Generating macOS light mode tokens...');
```
rurikoaraki, line 43: Are you skipping dark mode for now?

chiuam:
> Just wondering if you're expecting to add alias tokens too? Our components reference alias tokens instead of global tokens directly.

took this offline, thank you @rurikoaraki for walking me through how alias tokens work with global tokens. I added light/dark mode alias tokens for macOS. For high contrast, we'll let the OS handle that. FYI @mischreiber @harrieshin

scripts/build-tokens.js
chiuam: forgot to push the latest commits, they're added in ef20d47cff4149350648dd84506fd113e1e1d78b

packages/theming/theme-tokens/src/pipeline-input/token-input-light.macos.json
chiuam: Agreed, I'll move it to the global tokens file. Though I think we want to keep the filename as token-input-light.macos so that light / dark tokens are differentiated?

rurikoaraki: Ah I meant you can move the font stuff to the token-input-brand.macos.json file, and then rename token-input-brand.macos to token-input.macos, or something, so that the file doesn't imply that it only holds brand information.

chiuam: gotcha! fixed in 39abe89c1689464cfce3c38bca793e021a0477b4

rurikoaraki (approved):

PR #1148, opened 2021-11-10
Let users override "useNativeDriver" for Shimmer / ActivityIndicator

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

A client team would like to use Shimmer & ActivityIndicator. However, we would want to make sure we set `useNativeDriver` for those applications properly to ensure the best performance based on their react-native version. Due to a bug with animations (fixed with https://github.com/facebook/react-native/pull/29585 and https://github.com/microsoft/react-native-macos/pull/763), we can't set the useNativeDriver prop to true unless we're on RN 0.66+ or RN-macOS 0.62+.

Let's add a check that sets `useNativeDriver` to it's proper value based on our RN version, and also add `useNativeDriver` as a prop to Shimmer and ActivityIndicator, should the client need to override it for whatever reason (lex: they patch the bug in their own repo, or update to a newer version of react-native). 

### Verification

Ran the iOS and macOS test app with console.logs of the what value `useNativeDriver` is set to, to make sure it was set properly.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/ActivityIndicator/src/ActivityIndicator.tsx
```diff
@@ -36,6 +36,26 @@ export const ActivityIndicator = compose<ActivityIndicatorType>({
...
+       * which is only available in React Native 0.66+, and React Native macOS 0.62+
+       */
+      const nativeVersion = Platform.constants.reactNativeVersion;
+      if (Platform.OS === 'macos' && nativeVersion.minor >= 62) {
+        return true;
+      } else if (nativeVersion.minor >= 66) {
+        return true;
+      } else {
+        return false;
+      }
```
Saadnajmi, line 56: This isn't a perfect check of whether we are running with react-native-macos or not: An iOS or Android app could be running with react-native-macos 0.63, and this would return false where it should return true. I couldn't think of a better way to do it though.

packages/experimental/Shimmer/src/Shimmer.tsx
```diff
@@ -36,6 +36,26 @@ export const Shimmer = compose<ShimmerType>({
...
+      if (props.useNativeDriver !== undefined) {
+        return props.useNativeDriver;
+      }
+      /**
+       * https://github.com/facebook/react-native/pull/29585
+       * In order for native driven animations to loop, React Native needs this fix,
+       * which is only available in React Native 0.66+, and React Native macOS 0.62+
+       */
+      const nativeVersion = Platform.constants.reactNativeVersion;
+      if (Platform.OS === 'macos' && nativeVersion.minor >= 62) {
```
chiuam, line 50: minimum version of rn-macos FURN supports is 0.63?

packages/experimental/Shimmer/src/Shimmer.types.ts
```diff
@@ -142,6 +142,11 @@ export interface ShimmerTokens extends IBackgroundColorTokens {
    * @defaultValue '100%' of the 'width' property
    */
   shimmerWaveWidth?: number | string;
+
+  /**
+   * overrides the useNativeDriver prop on the component's animation
+   */
+  useNativeDriver?: boolean;
```
chiuam, line 149: can we not expose it as a prop if we do the RN version check inside the component?

packages/experimental/Shimmer/src/Shimmer.tsx
Saadnajmi: True, but the check was technically ported to RN-macOS 0.62+, so I thought it was better to write it as such. 
I actually don't know if we ever hard defined minimum versions of RN FURN supports or such

packages/experimental/Shimmer/src/Shimmer.types.ts
Saadnajmi: I added the prop in case, say, a user is using some other fork of react-native or managed to patch react-native to include that fix, so they could override our default value.

packages/experimental/Shimmer/src/Shimmer.tsx
Saadnajmi: JK, we explicitly define it: https://github.com/microsoft/fluentui-react-native/blob/b8535f925f1c90edaeaee03f9a379382d786a704/packages/libraries/core/package.json#L59
I'll update this check

packages/experimental/Shimmer/src/Shimmer.types.ts
chiuam: I feel like we are trying to cover all use cases rather than take it one step at a time. I personally think exposing this as a prop can be risky as it could affect performance of the control.

Saadnajmi:
Talked offline, and I'm closing this in favor of a different solution

PR #1151, opened 2021-11-11
Use the native driver in  Shimmer / ActivityIndicator animations

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Previously, we had `useNativeDriver` set to `false` for our components with Animations (Shimmer and ActivityIndicator). This greatly impacts performance. We did this because React Native was missing a [fix](https://github.com/facebook/react-native/pull/29585) to allow natively driven animations to loop. 

Let's borrow a solution from here (https://medium.com/@benjamintodts/react-natives-animated-loop-invoking-a-callback-whenever-an-iteration-finishes-1c3581d38d54) to emulate a looped animation, and allow us to use the native driver. We basically run the animation once and then in the completion callback, run the animation again, effectively running it in a loop. I also took inspiration from react-native-paper's [ActivityIndicator](https://github.com/callstack/react-native-paper/blob/main/src/components/ActivityIndicator.tsx) to help rewrite the animation in a way we could stop it mid animation.  

### Verification

I ran the iOS test app and made sure the props on ActivityIndicator still worked, and that the Shimmer / ActivityIndicator animations ran smoothly.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


chiuam (approved):
Thank you for making this change, this is great improvement in performance (no more 60FPS). Though I'm curious whether or not pinging the native thread every animation loop is prod acceptable :)

Saadnajmi:
> Thank you for making this change, this is great improvement in performance (no more 60FPS). Though I'm curious whether or not pinging the native thread every animation loop is prod acceptable :)

That's fair. Assuming an animation ran for say, 1 second, it's still a 60x improvement so I thought it was worth it.

PR #1153, opened 2021-11-12
chore: enable no-export-all rule

tido64:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Enables the no-export-all rule for exports from external packages.

More on why `export *` is bad here: https://hackmd.io/Z021hgSGStKlYLwsqNMOcg

### Verification

Build should succeed.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-52e4c067-3c10-45af-b1c6-12a9c26b4d43.json
+++ b/change/@fluentui-react-native-52e4c067-3c10-45af-b1c6-12a9c26b4d43.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Don't `export *` from external packages",
+  "packageName": "@fluentui/react-native",
+  "email": "4123478+tido64@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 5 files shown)


Saadnajmi:
So this is only applying the rule to "@fluentui/react-native" and @fluentui-react-native/framework". There's a lot more packages that's are publicly exported than that. Was that intentional?

tido64:
> So this is only applying the rule to "@fluentui/react-native" and @fluentui-react-native/framework". There's a lot more packages that's are publicly exported than that. Was that intentional?

Yeah, we will only fix the ones that are exporting from external packages. I'm not sure we actually to need to fix internal ones yet. There are a couple of things I need to check before we go there.

Saadnajmi:
> > So this is only applying the rule to "@fluentui/react-native" and @fluentui-react-native/framework". There's a lot more packages that's are publicly exported than that. Was that intentional?
> 
> Yeah, we will only fix the ones that are exporting from external packages. I'm not sure we actually to need to fix internal ones yet. There are a couple of things I need to check before we go there.

I see, so this only applies to our public packages that are coalescing other public packages together. Thanks!

Saadnajmi (approved):

PR #1157, opened 2021-11-12
Add v1 Button migration and spec files

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding documentation for v1 for spec and migration.
Shamelessly taking the layout from the fluentui repo.

The specs/migration steps are a first-pass level, nothing is finalized yet. Figured I'd get the documentation started so that it's easier to add to as we finalize things.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
+# Button
```
rurikoaraki, line 1: This file will replace the existing file in the docs folder once we move to the new button. Don't want to confuse people when we're still shipping the old button.

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+| `<Button text="Hello, world" />`        | `<Button>Hello, world</Button>`                      |
+| `<PrimaryButton text="Hello, world" />` | `<Button appearance='primary'>Hello, world</Button>` |
+| `<StealthButton text="Hello, world" />` | `<Button appearance='subtle'>Hello, world</Button>`  |
+
+### Props that remain as is
+
+- `children`
+- `disabled`
+- `icon`
+- `onClick`
```
Saadnajmi, line 24: nit: I would vote to  rename this onPress, to better match pressable, but that's a followup for me on my other code review #1070 anyway

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+- `testID`
+- `tooltip`
+
+### Props no longer supported with an equivalent functionality in v1 Button
+
+- `componentRef` => Use regular `ref` instead
```
Saadnajmi, line 30: 🎉

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+- `content` => Pass the content as `children` instead
```
Saadnajmi, line 31: Isn't content text, where children is a view hierarchy?

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+- `endIcon` => Use `right` value for `iconPosition` prop and pass icon information into `icon` prop instead
+- `startIcon` => Use `left` value for `iconPosition` prop and pass icon information into `icon` prop instead
```
Saadnajmi, line 33: nit: I would say we pass start and end to iconPosition for better internationalization

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Props no longer supported without an equivalent functionality in v1 Button
+
+- Cannot use both `startIcon` and `endIcon` at the same time in v1
```
Saadnajmi, line 37: What will MenuButton do, which has both icons?

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Icon
+
+The `Button` component can include an `icon` that appears before or after its `children`. If an `icon` is provided without any `children`, then the `Button` becomes an icon-only `Button`.
+
+### Shape
+
+- shape="rounded": The button as rounded corners. This is the default is shape is not set.
+- shape="circular": The button has completely round corners. A button of equal width and height will be a circle.
+- shape="square": The button has right-angle corners.
```
Saadnajmi, line 43: Is this coming from web? Because I feel this can easily be accomplished with varying the cornerRadius / width

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
...
+
+The `Button` component supports different sizing with at least three different sizes: `small`, `medium` (default) and `large`.
+
+### Block
+
+The `Button` component can completely fill the width of its container.
+
+### Loading
+
+The `Button` component can be loading if it's waiting for another action to occur before allowing itself to be interacted with.
```
Saadnajmi, line 55: As in, it would have an ActivityIndicator inside of it?

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
...
+
+## API
+
+### Slots
+
+- `root` - The outer container representing the `Button` itself that wraps everything passed via the `children` prop.
+- `icon` - If specified, renders an `icon` either before or after the `children` as specified by the `iconPosition` prop.
+- `loader` - If specified, renders a `loader` before the `icon` and `children` while the `loading` flag is set to `true`.
```
Saadnajmi, line 63: Nit: (and I'm opening a can of worms), I somewhat feel the public API should just be the props and tokens of a component, and not slots. That's mostly because slots tend to change as design evolves, which makes it harder to be part of the public API. 

Granted, as things are now, clients have access to slots, so we should at least document them.

Saadnajmi (commented):
Realized a bunch of these comments might not be relevant because the doc is copied from web, so I stopped myself before I went further!

rurikoaraki:
> Realized a bunch of these comments might not be relevant because the doc is copied from web, so I stopped myself before I went further!

The general answer is, "we are trying to align with the web API to make implementation across fluentui and FURN a smoother experience." We apparently get a lot of feedback around how different the APIs are across the platforms, so while some of these are easy to accomplish, it would be a quick ease of use thing to add props surfacing some defaults. There are some props we're not carrying over due to inherit differences between web and native (namely accessibility) but we're aiming to make the porting process smoother since it's a common scenario for clients.

packages/experimental/Button/SPEC.md
rurikoaraki: I think? @ValentinaKozlova might have more context from design. I haven't seen redlines for this yet.

packages/experimental/Button/MIGRATION.md
rurikoaraki: Apparently web is making it so that content is now not a prop and you just pass in text as a child of the button instead... Not fully sure how that works so this might be a case where there's a difference due to technical limitation. @ValentinaKozlova was looking into this.

packages/experimental/Button/MIGRATION.md
rurikoaraki: We can bring this feedback to web. @ejlayne

packages/experimental/Button/SPEC.md
rurikoaraki: Yeah this is one of those ease-of-use things.

packages/experimental/Button/MIGRATION.md
rurikoaraki: The chevron will be a special case. The main argument against the additional flexibility that was introduced with having icons on both ends is that design isn't speccing for that case, so we can't guarantee that it will look nice/follow fluentui designs. (and also alignment with web). The chevron is specced to occupy a smaller amount of space so it's a special case.

packages/experimental/Button/MIGRATION.md
rurikoaraki: Yeah, agreed.

packages/experimental/Button/MIGRATION.md
rurikoaraki: Ah I forgot about that conversation. I think this will be a case of balancing alignment with web vs. what is intuitively native/RN, since I think web is named onClick to align with the HTML API.

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Props
+
+```ts
+export type ButtonProps = ComponentPropsCompat &
+  React.ButtonHTMLAttributes<HTMLElement> & {
```
ejlayne, line 69: @rurikoaraki Should we remove references to HTMLElement in our Button documentation? Or is there something here that's required even for React Native?

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,191 @@
...
+     * - 'subtle': Minimzes emphasis to blend into the background until hovered or focused.
+     * - 'transparent': Removes background and border styling.
+     */
+    appearance?: 'primary' | 'outline' | 'subtle' | 'transparent';
+
+    /**
+     * Icon slot that, if specified, renders an icon either before or after the `children` as specified by the
+     * `iconPosition` prop.
+     */
+    icon?: ShorthandPropsCompat<React.HTMLAttributes<HTMLElement>>;
```
ejlayne, line 83: Another HTMLElement reference, see comment above.

packages/experimental/Button/MIGRATION.md
ejlayne: Fair point, let's bring up with web folks.

packages/experimental/Button/SPEC.md
ValentinaKozlova: It's a good question. I asked the Design

packages/experimental/Button/SPEC.md
rurikoaraki: Ah yeah thanks for catching.

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,190 @@
...
+   * Icon slot that, if specified, renders an icon either before or after the `children` as specified by the
+   * `iconPosition` prop.
+   */
+  icon?: IconSourcesType;
+
+  // /**
+  //  * Loader slot that, if specified, renders a `loader` before the `icon` and `children` while the `loading` flag
+  //  * is set to `true`.
+  //  */
+  // loader?: ShorthandPropsCompat<React.HTMLAttributes<HTMLSpanElement>>;
```
ValentinaKozlova, line 88: Should loader be commented?

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,190 @@
...
+   * @default 'before'
+   */
+  iconPosition?: 'before' | 'after';
+
+  // /**
+  //  * A button can show a loading indicator if it is waiting for another action to happen before allowing itself to
+  //  * be interacted with.
+  //  * @default false
+  //  */
+  // loading?: boolean;
```
ValentinaKozlova, line 113: I think loading prop should be uncommented

rurikoaraki: Yeah, we can uncomment these. I think I left them commented since we weren't sure if we were doing it yet.

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Touch interaction
+
+The same behavior as above translated for touch events. This means that there is no equivalent for `mouseenter` and `mouseleave`, which makes it so that the hovered state cannot be accessed.
+
+## Accessibility
+
+### Expected behavior
+
+- Should default to render a native `button` element unless something else has been specified for the `root` slot, in which case `role="button"` should be added to it.
```
chiuam, line 188: Does this apply to v1 Button since it doesn't have any native modules?

rurikoaraki: Ah I think I should rewrite this for FURN. I think "native" here is referring to a button element in HTML. I didn't really edit this for RN

packages/experimental/Button/SPEC.md
ValentinaKozlova: I talked to Design. As I understood ActivityIndicator is mostly for other controls. But we also can use it in Button component.

Saadnajmi: There is a (very new, in the last year or so) precedent for ActivityIndicators inside buttons on iOS is why I asked. Previously, I would have thought that a strange design, till apple made it one of their officially supported button configurations 🤷 (https://developer.apple.com/documentation/uikit/uibutton/configuration/3750795-showsactivityindicator?changes=l_8)

I suppose how we actually achieve the loading effect is a detail, but thanks for satisfying my curiosity :)

ValentinaKozlova (approved):

rurikoaraki:
@ejlayne @Saadnajmi @chiuam Any other comments? Otherwise I will merge this at EOD.

packages/experimental/Button/SPEC.md
chiuam: Thank you!

chiuam (approved):

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+
+#### Cursor interaction
+
+- `mouseenter`: Should immediately change the styling of the `Button` so that it appears to be hovered.
+- `mouseleave`: Should immediately remove the hovered styling of the `Button`.
+- `mouseup`: If triggered while cursor is still inside of the `Button's` boundaries, then it should execute the `Button` and move focus to its target.
+
+#### Touch interaction
+
+The same behavior as above translated for touch events. This means that there is no equivalent for `mouseenter` and `mouseleave`, which makes it so that the hovered state cannot be accessed.
```
Saadnajmi, line 181: This is part of why I want to push for #1070 to make this API a lot simpler. It would be great if we could just communicate that a `Button` is a `Pressable`, which supports `PressableProps`, which includes hover / keyboard focus / touch / click interaction. This last bit of documentation to me feels a bit too web and less RN. 
I don't consider that blocking this review though, since this is all iteration.

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Accessibility
+
+### Expected behavior
+
+- Should default to adding `role="button"` to the root slot.
+- Should mix in the accessibility props expected for a `button` component.
+- Should be keyboard tabbable and focusable.
```
Saadnajmi, line 189: For the future, we could link the `useButton` hook to say exactly what accessibility props are added. Something like that.

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+```ts
+export type ButtonProps = Omit<IWithPressableOptions<ViewProps>, 'onPress'> & {
+  /**
+   * A button can have its content and borders styled for greater emphasis or to be subtle.
+   * - 'primary': Emphasizes the button as a primary action.
+   * - 'outline': Removes background styling.
+   * - 'subtle': Minimzes emphasis to blend into the background until hovered or focused.
+   * - 'transparent': Removes background and border styling.
+   */
+  appearance?: 'primary' | 'outline' | 'subtle' | 'transparent';
```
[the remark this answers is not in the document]
rurikoaraki: Uh let me just.... remove those now.... thanks for catching!

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+   * Icon slot that, if specified, renders an icon either before or after the `children` as specified by the
+   * `iconPosition` prop.
+   */
+  icon?: IconSourcesType;
+
+  // /**
+  //  * Loader slot that, if specified, renders a `loader` before the `icon` and `children` while the `loading` flag
+  //  * is set to `true`.
+  //  */
+  loader?: TBD;
```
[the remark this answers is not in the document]
rurikoaraki: TBD as in I don't know the type of the loader slot yet since that was still in discussion. I think I will change this to be ActivityIndicator

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+| Key     | Description                   |
+| ------- | ----------------------------- |
+| `Enter` | Executes the `Button` action. |
+| `Space` | Executes the `Button` action. |
+
+#### Cursor interaction
+
+- `mouseenter`: Should immediately change the styling of the `Button` so that it appears to be hovered.
+- `mouseleave`: Should immediately remove the hovered styling of the `Button`.
+- `mouseup`: If triggered while cursor is still inside of the `Button's` boundaries, then it should execute the `Button` and move focus to its target.
```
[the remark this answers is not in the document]
rurikoaraki: Ah yeah, I will change these, good call

packages/experimental/Button/SPEC.md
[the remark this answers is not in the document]
Saadnajmi: If we do that, I think it would be a simple change to have FURN ActivityIndicator fall back to the RN base ActivityIndicator for not iOS/Android. But that's a whole different thing =)

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,189 @@
...
+
+The following section describes the different states in which a `Button` can be throughout the course of interaction with it.
+
+#### Enabled state
+
+An enabled `Button` communicates interaction by having styling that invites the user to click/tap on it to trigger an action.
+
+#### Disabled state
+
+A disabled `Button` is non-interactive, disallowing the user to click/tap on it to trigger an action.
```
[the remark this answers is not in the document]
rurikoaraki: That's fair, I can combine these.

packages/experimental/Button/SPEC.md
rurikoaraki: Added link to useButton hook

packages/experimental/Button/SPEC.md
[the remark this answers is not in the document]
rurikoaraki: Done

rurikoaraki:
@SaadNajmi Any other feedback before I merge?

packages/experimental/Button/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+
+## STATUS: WIP 🚧
+
+This Migration guide is a work in progress and is not yet ready for use.
+
+## Migration from v0 Button
+
+### Component renames
+
+Common buttons now all map to `Button`:
```
Saadnajmi, line 11: Except FAB, and MenuButton

packages/experimental/Button/MIGRATION.md
Saadnajmi: Looking back through all the comments, I think this might just be a react vs react native difference, since web renders to HTML and presumably raw text is ok. It _might_ still with in react native but I would have thought we want a <Text> element. 
Point of discussion, not a review blocker. I'm interested to see what @ValentinaKozlova finds

packages/experimental/Button/MIGRATION.md
Saadnajmi: Similar platform difference that is a point of discussion

packages/experimental/Button/SPEC.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Interaction
+
+#### Keyboard interaction
+
+The following is a set of keys that interact with the `Button` component:
+
+| Key     | Description                   |
+| ------- | ----------------------------- |
+| `Enter` | Executes the `Button` action. |
```
Saadnajmi, line 163: "Action passed into onClick" maybe, since we don't have an action prop?

Saadnajmi (approved):
I don't consider any comment I just left as blocking. Thanks for doing this!

packages/experimental/Button/SPEC.md
rurikoaraki: Reworded to Executes the function passed into the `onClick` prop. Thanks!

packages/experimental/Button/MIGRATION.md
rurikoaraki: Reworded to specifically mention just Primary and Stealth buttons, which are the variants that exist on the older button component in FURN.

PR #1158, opened 2021-11-13
Move FabButton to be its own component

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Per PM request, moving FabButton to be its own button component and out of the common button. This allows it to be available for mobile platforms but not include the functionality for platforms that don't implement it.

New FabButton component was made using a lot of the base Button types (some refactored). FabButton has been stubbed out for non-mobile platforms but still uses compose so that it can be referenced in the same way across platforms, it just won't show up in desktop platforms.

### Verification

Verified that the button doesn't show in win32. In Android, the button still shows:

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Instead of copying two separate .android and .iOS FabButton files, what we made one .mobile file, and then from the .android/.iOS file we imported the .mobile file? That way we only have one copy of the code. I've been considering that for ActivityIndicator too, so I can have a desktop design and mobile design

rurikoaraki:
> Instead of copying two separate .android and .iOS FabButton files, what we made one .mobile file, and then from the .android/.iOS file we imported the .mobile file? That way we only have one copy of the code. I've been considering that for ActivityIndicator too, so I can have a desktop design and mobile design

Ah yeah, that should work. I did try just a plain .mobile earlier which didn't work. Probably the proper way to add those extensions is to fiddle with babel, but your workaround is something I'll try

packages/experimental/Button/src/FabButton/FabButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const FabButton = compose<FabButtonType>({
+  displayName: fabButtonName,
+  slots: {
+    root: View,
+    icon: Icon,
+    content: Text,
+  },
+  render: (_userProps: ButtonCoreProps, _useSlots: UseSlots<FabButtonType>) => {
+    return (_final: ButtonCoreProps, ..._children: React.ReactNode[]) => {
+      return null;
```
Saadnajmi, line 19: Maybe
``` jsx
return (
<Text>Not implemented for this platform</Text>
);
```

packages/experimental/Button/src/FabButton/FabButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+    };
+  },
+});
+
+export default FabButton;
```
Saadnajmi, line 24: FAB == Floating Action Button, so FAB Button is Floating Action Button Button. 
Maybe just name it FAB? Both react-native-elements and react-native-paper  (the #1 and #2 RN UI frameworks by usage) name is FAB too.
https://reactnativeelements.com/docs/fab
https://callstack.github.io/react-native-paper/fab.html

packages/experimental/Button/src/FabButton/FabButtonTokens.ts
```diff
@@ -0,0 +1,18 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { ButtonCoreTokens } from '../Button.types';
+
+export const defaultFabButtonTokens: TokenSettings<ButtonCoreTokens, Theme> = () =>
+  ({
+    borderRadius: globalTokens.corner.radius.circle, // big number for always rounded corners
```
Saadnajmi, line 8: Since this is a token now, we can probably remove the comment explainer

packages/experimental/Button/src/FabButton/FabButtonTokens.ts
```diff
@@ -0,0 +1,18 @@
...
+    // For large size
```
Saadnajmi, line 9: also this comment, since we now presumably support a size ramp?

Saadnajmi (changes requested):

packages/experimental/Button/src/FabButton/FabButton.tsx
rurikoaraki: I'd rather not since this will make it so that platforms that don't use the FAB but want to just have one copy of code across platforms suddenly get text where there shouldn't be any. I guess it's unlikely that this case would happen, but to me it seems odd for a component to suddenly generate text,

rurikoaraki:
> > Instead of copying two separate .android and .iOS FabButton files, what we made one .mobile file, and then from the .android/.iOS file we imported the .mobile file? That way we only have one copy of the code. I've been considering that for ActivityIndicator too, so I can have a desktop design and mobile design
> 
> Ah yeah, that should work. I did try just a plain .mobile earlier which didn't work. Probably the proper way to add those extensions is to fiddle with babel, but your workaround is something I'll try

Tried this out and it seems that it introduces a circular dependency. I'll try for some time to fix it on Monday

packages/experimental/Button/src/FabButton/FabButton.tsx
Saadnajmi: I see. It's either they get nothing rendering or random text. I was thinking maybe we have some simple <NYI> aka <Not Yet Implemented> type component, but that's probably out of scope for this review and you make a fair point.

packages/experimental/Button/src/FabButton/FabButtonTokens.ts
rurikoaraki: Done with latest commits.

packages/experimental/Button/src/FabButton/FabButtonTokens.ts
rurikoaraki: Done with latest commits.

packages/experimental/Button/src/FabButton/FabButton.tsx
rurikoaraki: Sounds good, done with latest commits.

packages/experimental/Button/src/FAB/FAB.android.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * @param layer The name of the state that is being checked for
+ * @param state The current state of the button
+ * @param userProps The props that were passed into the button
+ * @returns Whether the styles that are assigned to the layer should be applied to the button
+ */
+const buttonLookup = (layer: string, state: IPressableState, userProps: ButtonCoreProps): boolean => {
+  return state[layer] || userProps[layer] || (layer === 'hasContent' && userProps.content) || (layer === 'hasIcon' && userProps.icon);
+};
+
+export const FabButton = compose<FABType>({
```
Saadnajmi, line 25: Nit, rename FabButton here

packages/experimental/Button/src/FAB/FAB.ios.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * @param layer The name of the state that is being checked for
+ * @param state The current state of the button
+ * @param userProps The props that were passed into the button
+ * @returns Whether the styles that are assigned to the layer should be applied to the button
+ */
+const buttonLookup = (layer: string, state: IPressableState, userProps: ButtonCoreProps): boolean => {
+  return state[layer] || userProps[layer] || (layer === 'hasContent' && userProps.content) || (layer === 'hasIcon' && userProps.icon);
+};
+
+export const FabButton = compose<FABType>({
```
Saadnajmi, line 25: Nit: rename FabButton here

packages/experimental/Button/src/FAB/FAB.android.tsx
rurikoaraki: Thanks for catching these!

rurikoaraki:
> > > Instead of copying two separate .android and .iOS FabButton files, what we made one .mobile file, and then from the .android/.iOS file we imported the .mobile file? That way we only have one copy of the code. I've been considering that for ActivityIndicator too, so I can have a desktop design and mobile design
> > 
> > 
> > Ah yeah, that should work. I did try just a plain .mobile earlier which didn't work. Probably the proper way to add those extensions is to fiddle with babel, but your workaround is something I'll try
> 
> Tried this out and it seems that it introduces a circular dependency. I'll try for some time to fix it on Monday

The way I was able to get it to work was to move the code into some "core" file and then importing & reexporting it on the ios and android files. Not really liking the FABMobile name though so if you have any suggestions on that I'm open to them, except that the core file name can't be FAB.

Saadnajmi:
> > > > Instead of copying two separate .android and .iOS FabButton files, what we made one .mobile file, and then from the .android/.iOS file we imported the .mobile file? That way we only have one copy of the code. I've been considering that for ActivityIndicator too, so I can have a desktop design and mobile design
> > > 
> > > 
> > > Ah yeah, that should work. I did try just a plain .mobile earlier which didn't work. Probably the proper way to add those extensions is to fiddle with babel, but your workaround is something I'll try
> > 
> > 
> > Tried this out and it seems that it introduces a circular dependency. I'll try for some time to fix it on Monday
> 
> The way I was able to get it to work was to move the code into some "core" file and then importing & reexporting it on the ios and android files. Not really liking the FABMobile name though so if you have any suggestions on that I'm open to them, except that the core file name can't be FAB.

Maybe FABMobile is FABCore, and the default FAB  is "UnimplementedFAB"? Inside one of these files we can leave a comment saying FAB is only implemented for mobile?

Saadnajmi (changes requested):
Can we add a test page to the iOS /Android test app too? Ran this locally and noticed I couldn't actually test the component.

rurikoaraki:
> Can we add a test page to the iOS /Android test app too? Ran this locally and noticed I couldn't actually test the component.

Sorry about that, it should show up as part of the first section of button tests now.

ValentinaKozlova (approved):

Saadnajmi (approved):

PR #1163, opened 2021-11-16
Remove transparent and outline as appearance options for v1 Button

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Design has stated that outside of web, the transparent and outline styles aren't supported by win32. Until they need to be supported, I'm removing them as options from the appearance prop.

### Verification

Since it wasn't implemented, removing these should have no effect.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ValentinaKozlova (approved):

chiuam (approved):

packages/experimental/Button/src/Button.types.ts
```diff
@@ -69,7 +69,6 @@ export interface ButtonProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
    * A button can have its content and borders styled for greater emphasis or to be subtle.
    * - 'primary': Emphasizes the button as a primary action.
    * - 'subtle': Minimizes emphasis to blend into the background until hovered or focused.
-   * - 'transparent': NYI. Removes background and border styling.
```
Saadnajmi, line 72: This is similar to our previous "ghost/stealth" button style, correct? I think that is what the sidebar button for FluentTester are implemented as, so expect a visual change whenever we eventually do make the visual changes

rurikoaraki: No, ghost/stealth has been renamed to subtle, which has a transparent background but will color the background when the button is interacted with. Transparent will always have a transparent background - it changes the text color on interaction with the button.

Saadnajmi (approved):

PR #1165, opened 2021-11-16
Phase 2: Win32 E2E Testing - Implementation Changes + Accessibility Testing Additions

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

**Win32 Phase 1 of 2, an extension of [this PR](https://github.com/microsoft/fluentui-react-native/pull/1125).**

This contains E2E Testing Additions for FocusZone, FocusTrapZone, Separator, MenuButton, Theme, Icon, and ContextualMenu.

I'm splitting this PR into phases because the original PR I had planned to release had changed 105 files. A PR that big is unrealistic to review and is prone to mistakes slipping through. The difference between the phases is simply which components I'm implementing the E2E testing for. The next Phase will have all the controls not in this one. And although there are 50+ file changed in this one, there's a lot of small/repetitive changes and file name changes that inflate the number.

The PR has some major changes to the implementation of E2E testing on Win32 (UWP to come). It also some Accessibility Testing for our controls.

Add tests to validate accessibilityRole and accessibilityLabel attributes (in applicable controls, see status here)
Change E2E Testing Infrastructure to a per-component basis.
With this change, each FURN component will now have it's own testing data. As the scope of our testing grows, the more component-specific tests we will have.
Each component will have its own respective Page Object/Spec files..
Additionally, this will make setting up testing cross-plat much easier. We already have the skeleton of the structure we want, now it's just a matter of making platform-specific files.
Smaller Changes

Improvements to BasePage.ts. This is an extremely important file, as every component's page object extends this. I've added more in-depth comments and a wider range of UI interaction functions. The goal is to have this class extendable cross-plat
Creating E2E Testing specific test files for each component. On each test page in our app, there will now be a section called "E2E Testing" that will only be used by this testing infrastructure. This will separate the actual examples for partners and the scenarios we want to run with out automation.
Rename BootTestPage.ts to NavigateAppPage.ts. This name makes more sense, as this class is responsible for navigating and clicking on each test page. I'm also moving this to the /common/ folder of E2E testing because all components will use this.
Rename all page object files (let's use Button as an example) from , ButtonTestPage.ts to ButtonPageObject.ts. The new name more accurately reflects that the class is a Page Object class rather than an actual test page in our app.
In TabsItem and Button, we were incorrectly passing the testID prop to it's content slot. This assigned that slot (a Text component) the testID, which we don't want. We want to interact with the Button/TabItem control itself, so we want to pass it to the root (which is already done via ...rest
Add some extra documentation on Page Objects

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/FocusTrapZone/consts.ts
```diff
@@ -1,2 +1,10 @@
 export const HOMEPAGE_FOCUSTRAPZONE_BUTTON = 'Homepage_FocusTrapZone_Button';
 export const FOCUSTRAPZONE_TESTPAGE = 'FocusTrapZone_TestPage';
+
+/* E2E Testing FocusTrapZone 1 */
+export const FOCUSTRAPZONE_TEST_COMPONENT = 'FocusTrapZone_Test_Component'; // A component on each specific test page
+export const FOCUSTRAPZONE_ACCESSIBILITY_LABEL = 'E2E testing FocusTrapZone_Test_Component accessibility label';
```
ksiler, line 6: Do we really use accessibility labels on FTZs?

ksiler: Same question for FocusZone. Not sure how effective the test is if we never encounter the label (I don't believe we get focus on the FocusZone/TrapZone itself, just the focusable items it contains, right?). I don't think FocusZone even supports a11yLabel.

apps/fluent-tester/src/FluentTester/TestComponents/Separator/consts.ts
```diff
@@ -1,2 +1,10 @@
 export const HOMEPAGE_SEPARATOR_BUTTON = 'Homepage_Separator_Button';
 export const SEPARATOR_TESTPAGE = 'Separator_TestPage';
+
+/* E2E Testing Separator 1 */
+export const SEPARATOR_TEST_COMPONENT = 'Separator_Test_Component'; // A component on each specific test page
+export const SEPARATOR_ACCESSIBILITY_LABEL = 'E2E testing Separator accessibility label';
```
ksiler, line 6: Another question about whether testing a11y label makes sense here. Separator is strictly a visual component right (as in not interactable)?

apps/fluent-tester/src/E2E/FocusZone/pages/FocusZonePageObject.win.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  FOCUSZONE_TESTPAGE,
+  FOCUSZONE_TEST_COMPONENT,
+  HOMEPAGE_FOCUSZONE_BUTTON,
+} from '../../../FluentTester/TestComponents/FocusZone/consts';
+import { BasePage, By } from '../../common/BasePage';
+
+class FocusZonePageObject extends BasePage {
+  // OVERRIDE: We use isExisting() here instead of isDisplayed() because FocusTrapZone does not have any UI to it, it's simply
```
ksiler, line 9: nit: change FocusTrapZone to FocusZone in comment

apps/fluent-tester/src/FluentTester/TestComponents/FocusTrapZone/consts.ts
samuelfreiberg: Good catch, I agree FZ and FTZ don't need these constants, it doesn't make sense.  I'll remove in my next commit

apps/fluent-tester/src/FluentTester/TestComponents/Separator/consts.ts
samuelfreiberg: Again, I agree with this. I'll remove it

apps/fluent-tester/src/E2E/FocusZone/pages/FocusZonePageObject.win.ts
samuelfreiberg: Good catch, changing now

apps/fluent-tester/src/FluentTester/TestComponents/FocusTrapZone/consts.ts
```diff
@@ -3,8 +3,6 @@ export const FOCUSTRAPZONE_TESTPAGE = 'FocusTrapZone_TestPage';
 
 /* E2E Testing FocusTrapZone 1 */
 export const FOCUSTRAPZONE_TEST_COMPONENT = 'FocusTrapZone_Test_Component'; // A component on each specific test page
-export const FOCUSTRAPZONE_ACCESSIBILITY_LABEL = 'E2E testing FocusTrapZone_Test_Component accessibility label';
 
 /* E2E Testing FocusTrapZone 2 */
 export const FOCUSTRAPZONE_NO_A11Y_LABEL_COMPONENT = 'FocusTrapZone_No_A11y_label_Component';
```
ksiler, line 8: Are these constants used? `FOCUSTRAPZONE_NO_A11Y_LABEL_COMPONENT`, `FOCUSZONE_NO_A11Y_LABEL_COMPONENT`, and `SEPARATOR_NO_A11Y_LABEL_COMPONENT`?

samuelfreiberg: Not quite yet. I preloaded them in because they will be used in the near feature once I add more functional tests. I can, however, change the name, because it doesn't reflect properly what each component will be

ksiler (approved):

PR #1166, opened 2021-11-16
Add palette fallback to new win32 theme code

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Apparently the fallback code to default to a palette when none is provided wasn't ported to the new theme code.

Adding code to fall back to Menus_FluentSV palette by default, which aligns with the newest visuals. Also updating tester to use new theme.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
+++ b/apps/fluent-tester/src/FluentTester/theme/applyTheme.ts
@@ -55,7 +55,7 @@ const themingModule = getThemingModule()[0];
 export function applyTheme(parent: Theme, name: ThemeNames, appearance: ThemeOptions['appearance']): PartialTheme {
   switch (name) {
     case 'Office':
-      return themingModule ? createOfficeTheme({ appearance, paletteName: 'WhiteColors' }).theme : {};
+      return themingModule ? createOfficeTheme({ appearance, paletteName: 'Menus_FluentSV' }).theme : {};
     case 'Caterpillar':
       return applyCaterpillarTheme(parent);
     default:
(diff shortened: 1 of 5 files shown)


Saadnajmi (approved):
Feels weird to have the string "Menus_FluentSV" be in open source but 🤷🏾

rurikoaraki:
> Feels weird to have the string "Menus_FluentSV" be in open source but 🤷🏾

Yeah... we could get the Native Module to just send a default palette when nothing is set if you'd rather?

Saadnajmi:
> > Feels weird to have the string "Menus_FluentSV" be in open source but 🤷🏾
> 
> Yeah... we could get the Native Module to just send a default palette when nothing is set if you'd rather?

Not sure what's better, as in, I have no idea how much of a change that is, or if it matters if this string is out there. So... 🤷🏾

rurikoaraki:
> > > Feels weird to have the string "Menus_FluentSV" be in open source but 🤷🏾
> > 
> > 
> > Yeah... we could get the Native Module to just send a default palette when nothing is set if you'd rather?
> 
> Not sure what's better, as in, I have no idea how much of a change that is, or if it matters if this string is out there. So... 🤷🏾

It shouldn't be any worse than when we had 'WhiteColors' but I'll ask around. The change to get the native module to have a default instead shouldn't be terrible, but it would take longer to implement.

rurikoaraki:
Some discussion is happening so closing this out for now.

PR #1169, opened 2021-11-17
Revert "Moved Avatar (PersonaCoin) to framework v2."

Saadnajmi:
Reverts microsoft/fluentui-react-native#1161

This broke iOS FluentTester, unfortunately, and I didn't find a simple fix in the 40 ish minutes I was testing =/


Saadnajmi:
I _may_ have found a local fix so that we don't need to revert this. Will update shortly

Saadnajmi:
Closing in favor of #1170

PR #1170, opened 2021-11-17
Fix experimental avatar

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

#1161 broke iOS Avatar, which is what is currently shipping in production, so we don't want to do that. This was because we had "2" Avatars (the existing Native Avatar, and the new JS Avatar) that were shipping with the same name, and some conflicting types. The name clash caused some issues I'm not 100% sure about. I fixed it by renaming the JS based Avatar to "JSAvatar", along with all of it's props/tokens, and have the experimental avatar package now export 2 components: Avatar (the Native Avatar on iOS/macOS), and "JSAvatar".

Also, the JS Avatar doesn't need to be limited to win32: it will work on Android and web, so I added it to each test app's test pages. 

### Verification

Tested on iOS, macOS, and win32

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Hmm hoping that once we can reconcile the types to be consistent across platforms we don't need the JS prefix, but I'm good for this to go in to unblock things.

rurikoaraki (approved):

Saadnajmi:
> Hmm hoping that once we can reconcile the types to be consistent across platforms we don't need the JS prefix, but I'm good for this to go in to unblock things.

Oh yes, I forgot to mention that in the description. I would like the prefix to only be temporary until the APIS are aligned, and we don't need to export two packages. Good callout!

chiuam (approved):

Saadnajmi:
Works on win32!

PR #1174, opened 2021-11-18
[Contextual Menu macOS] Fix bug where two items are highlighted on click

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We had a weird bug where when you clicked on a ContextualMenuItem, two items would be highlighted. Looking deeper, it seems that keyboard focus is going to some other view, and the styling (background highlights) for the focus state looked the same as the styling for the press state. This meant that two items got highlighted, the pressed menu item and the focused menu item. 
It's unclear why keyboard focus is going elsewhere, but we can fix the visual issue by removing the focus style overrides, so that the focus state looks like the rest state, and only one menu item gets highlighted

### Verification

Before:

https://user-images.githubusercontent.com/6722175/142502592-d3ad7624-cf13-4397-8c85-876299ef8460.mov

After:

https://user-images.githubusercontent.com/6722175/142501490-b4628329-0aa5-4b7f-8591-244c6012c0bf.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


HeyImChris (approved):
Does this work with the keyboard too? (e.g. arrow key navigation, escape to close, etc.)

Saadnajmi:
> Does this work with the keyboard too? (e.g. arrow key navigation, escape to close, etc.)

Keyboard navigation currently doesn't work at all. That's a separate bug/feature I'm looking into.

PR #1175, opened 2021-11-19
Allow maxHeight prop to be set on menu and submenu for Scrollview

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- ContextualMenu.tsx
- Submenu.tsx
          - maxHeight set on these components allows users to set a height  for when the menu should add a scrollView

The user can set a maxHeight on how tall the want they menu, and the scrollViewer will appear if the menu gets longer. If they don't set a value, scrollViewer will not get added, and the menu just gets truncated at window edge (which is not a desirable behavior, but we do not yet have a redline on the maxHeight. The decision on whether we want a default maxHeight value or follow the web convention and set it at window's height is TBD)

TODO (Bugs to fix):
- With the ScrollView, when menu contains items with png or fontIcon icons, it mysteriously adds a margin at the bottom of the menu
- Width of menu is not dynamic with the menu item size. It was not an issue before the ScrollView was added. I will look into this further.

### Verification

menu with scrollView

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Common/iconExamples.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Common/iconExamples.tsx
@@ -13,7 +13,7 @@ export const fontProps: FontIconProps = Platform.select({
   macos: {
     fontFamily: 'Arial',
     codepoint: 0x2663,
-    fontSize: 32,
+    fontSize: 12,
   },
   default: {
     fontFamily: `Font Awesome 5 Free`,
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Common/iconExamples.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Common/iconExamples.tsx
@@ -1,0 +1,26 @@
+import { Platform } from 'react-native';
+import { SvgIconProps, FontIconProps } from '@fluentui-react-native/icon';
+import TestSvg from '../Button/test.svg';
+
+export const svgProps: SvgIconProps = {
+  src: TestSvg,
+  viewBox: '0 0 500 500',
+};
+
+const testTtf = require('../Button/Font Awesome 5 Free-Solid-900.otf');
+
+export const fontProps: FontIconProps = Platform.select({
+  macos: {
+    fontFamily: 'Arial',
+    codepoint: 0x2663,
+    fontSize: 32,
+  },
+  default: {
+    fontFamily: `Font Awesome 5 Free`,
+    fontSrcFile: testTtf,
+    codepoint: 0xf083,
+    fontSize: 12,
+  },
+});
+
+export const testImage = require('../Button/icon_24x24.png');
--- a/apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
@@ -108,6 +108,26 @@ const contextualMenu: React.FunctionComponent = () => {
           <ContextualMenuItem text="Disabled Menu Item" itemKey="3" disabled />
           <ContextualMenuItem text="MenuItem 4" itemKey="4" />
           <ContextualMenuItem text="MenuItem 5" itemKey="5" />
+          <ContextualMenuItem text="MenuItem 1" itemKey="1" />
+          <ContextualMenuItem text="MenuItem 2" itemKey="2" />
+          <ContextualMenuItem text="Disabled Menu Item" itemKey="3" disabled />
+          <ContextualMenuItem text="MenuItem 4" itemKey="4" />
+          <ContextualMenuItem text="MenuItem 5" itemKey="5" />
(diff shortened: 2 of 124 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
```diff
@@ -343,6 +344,159 @@ const IconContextualMenu: React.FunctionComponent = () => {
...
+          setShowMenu={toggleShowContextualMenu}
+          shouldFocusOnMount={focusOnMount}
+          shouldFocusOnContainer={focusOnContainer}
+          isBeakVisible={isBeakVisible}
+          maxHeight={200}
+        >
+          <ContextualMenuItem text="MenuItem 1" itemKey="1" />
+          <ContextualMenuItem text="MenuItem 2" itemKey="2" />
+          <SubmenuItem
+            icon={{ svgSource: svgProps, width: 12, height: 12 }}
```
ValentinaKozlova, line 469: We have the same iconProps in several places. Maybe it would be good to have one object for that.

lenahong: I moved it to Common folder!

apps/fluent-tester/src/FluentTester/TestComponents/Common/iconExamples.tsx
```diff
@@ -0,0 +1,26 @@
...
+    fontFamily: 'Arial',
+    codepoint: 0x2663,
+    fontSize: 32,
+  },
+  default: {
+    fontFamily: `Font Awesome 5 Free`,
+    fontSrcFile: testTtf,
+    codepoint: 0xf083,
+    fontSize: 12,
+  },
```
Saadnajmi, line 23: You can make macOS 12 too, while we're at it :P

packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -83,10 +84,15 @@ export const ContextualMenu = compose<ContextualMenuType>({
...
+
     return (
       <CMContext.Provider value={renderData.state.context}>
         <Slots.root>
-          <Slots.container>{children}</Slots.container>
+          <Slots.container>
+            <ScrollView contentContainerStyle={{ flexDirection: 'column', flexGrow: 1}} showsVerticalScrollIndicator={true}>
+              {children}
+            </ScrollView>
+          </Slots.container>
```
Saadnajmi, line 95: Any reason why we couldn't have just exposed the prop on `container` and let the developer decide if they wanna add a ScrollView as a child or not?

lenahong: That's a good question. I think this is a discussion we need to have, as to how much we want to leave to users to decide? The current implementation just requires the users to decide how tall they want a menu to be. If their items list gets too long for the height fixed menu, the scrollView will be added. What you are suggesting is, they can choose both the height of the menu AND whether they want to add a ScrollView at all in any case. 

I think Lineage folks need a quick fix so I'm implementing this without official redlines actually. We should table this for sure, and we should have this design discussion eventually.

Saadnajmi: Yeah, nothing _too_ against the implementation. The thought initially came up because macOS menus by default don't have scrollviews. macOS has different scrolling behavior for overflowing menus, but it's really custom such that I don't know I could implement it in RN to begin with and will probably just resort to a scrollview 😅.

This is definitely something that could be left for, say, V1 Callout/ContextualMenu =)

apps/fluent-tester/src/FluentTester/TestComponents/ContextualMenu/ContextualMenuTest.tsx
```diff
@@ -343,6 +319,153 @@ const IconContextualMenu: React.FunctionComponent = () => {
...
+            itemKey="4"
+            onHoverIn={toggleShowSubmenu}
+            componentRef={stdMenuItemRef}
+          />
+          {showSubmenu && (
+            <Submenu maxHeight={200} target={stdMenuItemRef} onDismiss={onDismissSubmenu} onShow={onShowSubmenu} setShowMenu={toggleShowSubmenu}>
+              <ContextualMenuItem text="MenuItem 1" itemKey="1" />
+              <ContextualMenuItem text="MenuItem 2" itemKey="2" />
+              <ContextualMenuItem text="MenuItem 1" itemKey="1" />
+              <ContextualMenuItem text="MenuItem 2" itemKey="2" />
```
Saadnajmi, line 449: Duplicated item keys?

Saadnajmi (approved):
The ScrollView causes minheight issues on macOS for some reason. For the time being, could we make it win32 only? Otherwise, thanks for the change!

ValentinaKozlova (approved):

PR #1189, opened 2021-11-24
Integrate macOS tokens for Button

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Mac tokens were added in #1130, let's consume them in Button.  I also updated the button token values to use Fluent Apple palette colors. 

Updating these token values:
| Token (light-macos) | Before                                       | After                                      |
|---|--------------------------------|---------------------------|
|neutralBackgroundDisabled | #f0f0f0 | #E0E0E0 |
|brandBackground.fillColorRest | #185abd | #0078D4 |
|brandBackground.fillColorHover | #1651aa | #106EBE |
|brandBackground.fillColorPressed | #0e336a | # 004578 |
|brandBackground.fillColorSelected | #13458f | #005A9E |

| Token (dark-macos) | Before                                       | After                                      |
|---|--------------------------------|---------------------------|
|neutralBackgroundDisabled | #141414 | #D6D6D6 |
|neutralStrokeDisabled | #424242 | #525252 |
|brandBackground.fillColorRest | #185abd | #0086F0 |
|brandBackground.fillColorHover | #1651aa | #0078D4 |
|brandBackground.fillColorPressed | #0e336a | #6CB8F6 |
|brandBackground.fillColorSelected | #13458f | #3AA0F3 |

### Verification

| Button | Before                                       | After                                      |
|---|--------------------------------|---------------------------|
| Light |image| image|
| Dark |image| image|
| High contrast light |image|image|
| High contrast dark | image| image|

| Experimental Button | Before                                       | After                                      |
|---|--------------------------------|---------------------------|
| Light | image| image|
| Dark | image | image|
| High contrast light | image | image|
| High contrast dark |image | image|

Primary button hover/press state

https://user-images.githubusercontent.com/67026167/146228956-8e844117-a6cc-41d6-8231-3197cc02f73b.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


harrieshin:
screen shot of dark mode, increase contrast mode?

harrieshin:
might be worth actually having before and after table to make sure things didn't change as intented

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
```diff
@@ -1,262 +1,262 @@
...
+    "fillColorSelected": "#383838"
+  },
+  "neutralBackground2": {
+    "fillColorRest": "#1f1f1f",
+    "fillColorHover": "#333333",
+    "fillColorPressed": "#141414",
+    "fillColorSelected": "#2e2e2e"
+  },
+  "neutralBackground3": {
+    "fillColorRest": "#555555",
```
harrieshin, line 107: explicitly write them in your PR description for history

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
```diff
@@ -1,262 +1,262 @@
...
+    "fillColorRest": "#d6d6d6",
+    "fillColorHover": "#ffffff",
+    "fillColorPressed": "#ffffff",
+    "fillColorBrandHover": "#d2e0f4",
+    "fillColorBrandPressed": "#2e6ac5",
+    "fillColorBrandSelected": "#d2e0f4",
+    "fillColorSelected": "#ffffff"
+  },
+  "neutralForeground3": {
+    "fillColorRest": "#FFFFFF",
```
harrieshin, line 149: explicitly all out this change in your pr description for history

packages/theming/theme-tokens/src/light-macos/reactnative/tokens-aliases.json
```diff
@@ -1,262 +1,262 @@
...
+    "fillColorSelected": "#ebebeb"
+  },
+  "neutralBackground2": {
+    "fillColorRest": "#fafafa",
+    "fillColorHover": "#f0f0f0",
+    "fillColorPressed": "#dbdbdb",
+    "fillColorSelected": "#e6e6e6"
+  },
+  "neutralBackground3": {
+    "fillColorRest": "#0000000C",
```
harrieshin, line 107: first time seeing this hex but ok

mischreiber: Actually that's a good point: I don't see this value in the global tokens; seems like it should be `#F5F5F5`. Any idea what happened here @chiuam @rurikoaraki?

Saadnajmi:
Since this touches the apple theme, can we verify that the non experimental button is also unaffected?

packages/theming/theme-tokens/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export { default as globalTokens } from './tokens-global';
+export { default as aliasTokens } from './light-macos/tokens-alias';
```
rurikoaraki, line 2: I would rather us expose alias tokens through an implementation of `getAliasTokens` per platform. That way we allow for the alias tokens to change depending on light/dark/HC mode, and there's only one point exposed to get the alias tokens. If we export alias tokens directly like this, anyone else who imports `aliasTokens` instead of using `getAliasTokens` would only get the macOS light version ones.

packages/theming/theme-tokens/src/light-macos/reactnative/tokens-aliases.json
rurikoaraki: @chiuam any ideas? It doesn't look like the input file was touched, so I'm not sure why the output files were changed

rurikoaraki (changes requested):
Would like to see us use `getAliasTokens` instead of exporting alias tokens directly

packages/theming/theme-tokens/src/index.ts
chiuam: @rurikoaraki Thank you for the suggestion, added `getAliasTokens` implementation for macOS in 80f44e276ce34b9db6a78deeae6350eeb5a4e24b. Please take a second look when you get a chance.

rurikoaraki: Thanks @chiuam! Is there a reason to keep this implementation separate from the other `getAliasTokens`? Could we make a `getAliasTokens.macos.ts` so that `getAliasTokens` changes what it does on macos automatically instead of exporting an additional function?

Office has its own version as the function signature has to be different from the base `getAliasTokens`, but it looks like the Mac one can match the signature, unless you think this will change in the future.

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
```diff
@@ -0,0 +1,15 @@
...
+import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+
+export function getMacOSAliasTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return macOSLightAliasTokens.default;
+  } else if (mode === 'dark') {
+    return macOSDarkAliasTokens.default;
+  } else if (mode === 'highContrast') {
+    return hcAliasTokens.default;
+  }
```
rurikoaraki, line 13: nit: we can `assertNever` here to make sure we get a build failure is a new AppearanceOption is added

packages/theming/theme-tokens/src/index.ts
chiuam: Oh yeah, I didn't think of that. Fixed in b0aa4517334d4780e2c24a1236028f4730394a74

chiuam: Removed aliasTokens export in 7492c9ef0313fe93d0017ba534cfbb827efa69e4

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
chiuam: Added, I also removed the default case as it is unreachable I think...

packages/theming/apple-theme/src/appleColors.macos.ts
```diff
@@ -259,21 +264,13 @@ function getFluentUIApplePalette(): FluentUIApplePalette {
...
-    neutralStroke2: '#00000026',
-    neutralForeground3: DynamicColorMacOS({
-      light: '#272727',
-      dark: '#FFFFFF',
-    }),
-    neutralBackground3: DynamicColorMacOS({
-      light: '#0000000C',
-      dark: '#555555',
-    }),
+    neutralBackgroundInverted: macOSAliasColorTokens.neutralBackgroundInverted,
```
chiuam, line 267: It's kind of redundant to redefine the colors here, but let's keep it for now until we have all the colors updated?

packages/theming/apple-theme/src/appleColors.macos.ts
```diff
@@ -332,40 +325,40 @@ export function fallbackApplePalette(): ThemeColorDefinition {
...
     primaryButtonTextDisabled: fluentUIApple.brandForegroundDisabled,
 
     accentButtonBackground: fluentUIApple.communicationBlue,
-    accentButtonText: fluentUIApple.neutralInverted,
+    accentButtonText: macOSAliasColorTokens.neutralForegroundInverted,
 
     menuBackground: 'transparent',
     menuDivider: applePlatform.separatorColor,
-    menuIcon: neutralForeground3, //GH:728 Icon doesn't support PlatformColor
+    menuIcon: macOSAliasColorTokens.neutralForeground3,
```
chiuam, line 361: @Saadnajmi is it appropriate to not use PlatformColor here and use alias tokens instead?

packages/theming/theme-tokens/src/index.ts
rurikoaraki: You're still exporting it here though, I think we can remove this, too

packages/theming/apple-theme/src/appleColors.macos.ts
Saadnajmi: Eventually the token pipeline should have support for platform color, such that this token _is_ a platform color. This specific token is used in svgs so it can't be platform color just yet though.

Saadnajmi: I would keep the comment with the bug number

packages/theming/theme-tokens/src/index.ts
chiuam: removed, for real this time :)

packages/theming/apple-theme/src/appleColors.macos.ts
chiuam: Makes sense, fixed in d2132188efc7e7c453b4eaccff05d9b3f96dd578

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
Saadnajmi: `assertNever` was added to check that we never reach theoretically unreachable cases, so I think we would want to add that.

Saadnajmi: jk, it's there. My bad

packages/theming/theme-tokens/src/index.ts
chiuam: ^I removed the wrong one, should be fixed now...

packages/theming/theme-tokens/src/light-macos/reactnative/tokens-aliases.json
chiuam: I copied the value from the apple theme because it looked a lot closer to the existing button, here's a visual comparison: 
| #F5F5F5                                       | #0000000C                                      |
|----------|--------|
| image | image |

chiuam: I'm not sure whether or not we should go with `#F5F5F5` for buttonBackground, though I do think the hint of grey looks more natural to me. @Saadnajmi thoughts?

chiuam: I think we should follow the [spec](https://www.figma.com/file/ArWQiU7BAlrDGNmbrGEXbA/%5BSpecs%5D-Fluent-vNext-iOS-Components?node-id=744%3A734) for now to move forward and revisit the button background colors when the mac tokens are more finalized.  I'll revert the changes I made to the token values.

Saadnajmi: I copied the apple theme background button color from Fluent UI Apple. I would vote we open a bug on the design token repo or something to fix the macOS token, and follow spec. Especially if both options look acceptable in production, which it seems to be

packages/theming/apple-theme/src/appleColors.macos.ts
Saadnajmi: There will probably be lots of redundancy till we're fully integrated. Good to keep track where they are so we know to remove them, especially for the sake of bundle size.

Saadnajmi: I think if we do move to fully design token based, then we would get rid of the apple theme altogether, and just have it be part of the "Fluent Theme" or "Default Theme", since we don't have native code for this theme anyway.

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
rurikoaraki, line 107: This difference is still here

Can the changes to these files be fully reverted? Or can we update the input files instead if these entries are supposed to change? This file is auto-generated, we should not be editing it directly (same with the light version of this file)

rurikoaraki: (same comment for neutralForeground3)

packages/theming/theme-tokens/src/index.ts
rurikoaraki: Thanks!

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
chiuam: I decided to revert the changes here to follow the spec and do a followup once mac tokens are more finalized. I'll update the PR description to include screenshots of before & after

Saadnajmi (changes requested):
Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad.  I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.

image

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
chiuam: I did `git checkout origin/master <tokens-aliases.json>`, that should revert the file to its original state.

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
```diff
@@ -1,262 +1,262 @@
 {
-	"body": {
```
rurikoaraki, line 2: I don't think this file should be changed, either. Could you revert the changes in this file as well?

chiuam:
> Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad. I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.
> 
> image

Good catch @Saadnajmi. I wonder if we can detect appearance changes without using DynamicColorMacOS, looking into this...

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
chiuam: Oops, forgot to revert this one

Saadnajmi:
> > Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad. I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.
> > image
> 
> Good catch @Saadnajmi. I wonder if we can detect appearance changes without using DynamicColorMacOS, looking into this...

You can use the FURN `getCurrentAppearance` API, which I think the apple theme uses for neutralForeground3

packages/theming/theme-tokens/src/dark-macos/reactnative/tokens-aliases.json
chiuam: Done, thanks for catching!

chiuam:
> > > Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad. I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.
> > > image
> > 
> > 
> > Good catch @Saadnajmi. I wonder if we can detect appearance changes without using DynamicColorMacOS, looking into this...
> 
> You can use the FURN `getCurrentAppearance` API, which I think the apple theme uses for neutralForeground3

Not sure if I'm missing something, but `getCurrentAppearance` doesn't pick up theme changes without re-bundling

rurikoaraki (approved):

chiuam:
> screen shot of dark mode, increase contrast mode?

added screenshots to the PR description, please take a second look when possible.

chiuam:
> > > screen shot of dark mode, increase contrast mode?
> > 
> > 
> > added screenshots to the PR description, please take a second look when possible.
> 
> The high contrast light mode looks off... why does the background go away / it have no border ._. Maybe it's because borderColor is set to transparent instead of matching the background? EDIT: It looks like other stuff doesn't work well on increased contrast either, like separator :(

I don't think it's the border color since it was transparent before? Probably it's the high contrast token values are off, though I couldn't find any background colors defined for high contrast in the spec...

Separator uses Button from fluentui/react-native so yeah it's affected as well...

harrieshin:
> > > > > Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad. I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.
> > > > > image
> > > > 
> > > > 
> > > > Good catch @Saadnajmi. I wonder if we can detect appearance changes without using DynamicColorMacOS, looking into this...
> > > 
> > > 
> > > You can use the FURN `getCurrentAppearance` API, which I think the apple theme uses for neutralForeground3
> > 
> > 
> > Not sure if I'm missing something, but `getCurrentAppearance` doesn't pick up theme changes without re-bundling on macOS
> 
> Turns out it was because I had `getCurrentAppearance` as a global variable, so it never got updated when theme changes.

would you please update your PR screenshot with the latest change?

harrieshin:
can you share screenshot on RTL as well please?

Saadnajmi:
> can you share screenshot on RTL as well please?

I think the (main) place RTL support is broken is with icons / SVGs. We don't have a way to say "flip this icon in RTL" yet as we do downstream in office. The two main components this affects is Icon and Shimmer. General layout of JS components should flip properly, as long as they use flex box correctly (i.e:  start/end instead of left/right).

For button, the concern would be setting an icon with directional meaning (like a chevron), and we can work around that by informing clients about our behavior, so they know to adjust properly. I don't expect this PR to break any RTL stuff, for what it's worth.

Saadnajmi (approved):
Approved with the caveat that we probably need to file a bug for and properly handle High Contrast on macOS separately, as neither our FluentUI Apple Button nor are FluentUI React Native button respond to High Contrast changes :')

chiuam:
> > > > > > Cloned this locally and the different background color is actually slightly off-putting to my eye. I wouldn't block on that, except it stays light colored in dark mode, which is bad. I'm not sure anyone is using the default button, but I think it would be best not to cause a regression.
> > > > > > image
> > > > > 
> > > > > 
> > > > > Good catch @Saadnajmi. I wonder if we can detect appearance changes without using DynamicColorMacOS, looking into this...
> > > > 
> > > > 
> > > > You can use the FURN `getCurrentAppearance` API, which I think the apple theme uses for neutralForeground3
> > > 
> > > 
> > > Not sure if I'm missing something, but `getCurrentAppearance` doesn't pick up theme changes without re-bundling on macOS
> > 
> > 
> > Turns out it was because I had `getCurrentAppearance` as a global variable, so it never got updated when theme changes.
> 
> would you please update your PR screenshot with the latest change?

@harrieshin Sure! Highlighted token color changes & updated screenshots to reflect those changes.

PR #1198, opened 2021-12-01
Mac/iOS NuGet pack script

chripear:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Add a new NuGet pack script for Apple platforms that will allow us to build and package FURN locally, for sideloading and debugging in consumers.

Features:
- Pretty good incrementalism.
- Runnable from any working directory.

Options:
- Help.  Everything you see here but in more detail.
- Verbosity in 4 levels.  Levels 1+ log all subprocess argument vectors, for easy debugging of the script itself.
- Cleaning in 4 levels, up to `git clean -dfx`.
- Prerequisites.  Bootstraps a new (or cleaned) git repo so that it can build the native libraries.
- Debug.  Normally we only package the Release/Ship flavor.  This option substitutes the Debug libraries instead, for a better local debugging experience.

### Verification
Ran the script with various options and confirmed that the contents of the nuget folder were as expected.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/xcconfig/publish_overrides.xcconfig
```diff
@@ -1,10 +1,10 @@
 // For publish builds, only provide line tables for symbolizing crashes
-CLANG_DEBUG_INFORMATION_LEVEL=line-tables-only
+CLANG_DEBUG_INFORMATION_LEVEL[config=Release]=line-tables-only
```
Saadnajmi, line 2: Why only release?

.ado/scripts/xcode_select_current_version.sh
```diff
@@ -7,4 +7,4 @@ else
 fi
 
 echo "Running command: sudo xcode-select --switch $XCODE_PATH"
-sudo xcode-select --switch "$XCODE_PATH"
\ No newline at end of file
+/usr/bin/sudo /usr/bin/xcode-select --switch "$XCODE_PATH"
```
Saadnajmi, line 10: Is this just a preference to be exact here? No problems with it, but was curious if it caused issues or not

scripts/nuget_pack_apple.sh
```diff
@@ -0,0 +1,388 @@
...
+# \param $1 configuration
+function build_and_copy_macos()
+{
+	build_and_copy_output \
+	"${git_root}/apps/macos/src/FluentTester.xcworkspace" \
+	FluentTester \
+	"$1" \
+	macosx \
+	"${git_root}/.ado/xcconfig/publish_overrides.xcconfig" \
+	${products_dir}/${1}/{"FRN*/libFRN*.a",RCTFocusZone/libRCTFocusZone.a}
```
Saadnajmi, line 326: Somewhat concerned we'll forget to add new native modules here in the future

.ado/xcconfig/publish_overrides.xcconfig
chripear: Because we want the default level of debug information for the Debug configuration.  For debugging.

.ado/scripts/xcode_select_current_version.sh
chripear: This is proactive.  I'm not aware of issues with `sudo` and `xcode-select` specifically, but `python` invoked without an absolute path certainly has caused us some problems.  Thus a general rule: we always want to be as specific as possible when running system tools, for predictable results that don't depend on the user's PATH or whatever else they might have installed.  The only exceptions I have made are where there's no official blessed installation of a particular tool, i.e. `gem`, `yarn` and `pod`.

Saadnajmi: You can see the "official" installed software for ADO macOS VMs here: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md

scripts/nuget_pack_apple.sh
chripear: Then a good future improvement would be to either invoke the same NuGet packing process that is invoked by the CI task, or  parse the .nuspec file to extract the source and destination info.

However, note that the use of a naming convention helps us a lot here.  If we adopted the name FRNFocusZone/libFRNFocusZone.a for that last library, then the same glob pattern would apply to all of the library outputs, and enforcing that naming convention for new components would ensure we don't have to update anything here.

.ado/scripts/xcode_select_current_version.sh
Saadnajmi: Though I'm not seeing paths listed either in there

.ado/xcconfig/publish_overrides.xcconfig
harrieshin: but that will make our nugget size for debug build much bigger which the size exception approvers won't like. can we have a different script or a setting that it is only used for local nugget package build but leave the CI alone?

.ado/scripts/xcode_select_current_version.sh
chripear: I'm calling this script directly from the nuget pack script to avoid having to double-specify the officially supported Xcode path.  So the audience is no longer limited to ADO machines.

scripts/nuget_pack_apple.sh
Saadnajmi: Agreed that RCTFocusZone should be FRNFocusZone, and I would argue win32 should make the same change. 
Hard to make that change here and downstream without breaking something though :')

chripear: Still, using the FRN* convention for new components will mean we don't have to update this script every time.

Saadnajmi: Yeah, I fully expect every new component to do so. Just no actual CI errors to enforce that. Guess I'll just live with it 🤷🏾

.ado/scripts/xcode_select_current_version.sh
Saadnajmi: Oh right, my bad

HeyImChris (approved):
This is awesome. @chripear have you actually validated debugging with this in Devmain? For instance, adding a breakpoint and verifying it gets hit in our downstream Office apps.

.ado/xcconfig/publish_overrides.xcconfig
chripear: Our NuGet is ship only.  This doesn't affect the package size.

chripear:
> This is awesome. @chripear have you actually validated debugging with this in Devmain? For instance, adding a breakpoint and verifying it gets hit in our downstream Office apps.

I haven't been able to do that yet.  I should be getting a Mac mini that can build Office tomorrow, so I could try testing it then.  @Saadnajmi also mentioned wanting to test the script locally before a final sign-off, so he might be able to do that easily as well.

Saadnajmi:
Tried validating side loading locally, and it seems Xcode didn't resolve the dsyms :/ 
The script looks great, so I'm sure I did something wrong, and I'm going to merge this anyway
Screen Shot 2021-12-08 at 4 10 34 PM

Saadnajmi (approved):

PR #1205, opened 2021-12-01
Support up/down arrow key navigation on macOS ContextualMenu

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

1) Adds a FocusZone to macOS ContextualMenu so the user can arrow key between the menu items.
2) Updates the styling so that the menu item highlights on keyboard focus, not mouse hover. mouse hover sets keyboard focus, so the user can still highlights menu items with their mouse
3) Styles the submenu chevron icon to be white when highlighted as well
4) Updates documentation to make it more clear what props are win32 only, and what are not. This also changes the base type of a few things from `ViewProps` to `IViewProps` (so that we can use enableFocusRing without issue)
5) Removes the scrollView / maxHeight prop from macOS ContextualMenu. this caused issued where max height behaved like min height, and there would be empty space. We still need to handle overflow on macOS, but we'll do that in a separate PR.
6) Sets the props so that disabled menu items are not accessible or focusable. This change affects win32 so I would like input if this is ok (ping @lenahong)

Remaining work (non exhaustive):
1) Left/Right arrow keys open submenus
2) escape closes the menu
3) down/up arrow key focus on the first/last menu item respectively when a menu is opened
4) Clicking a menu item still causes keyboard focus to move for some reason

### Verification

https://user-images.githubusercontent.com/6722175/144328299-b18cbe49-b8c5-4077-be8b-503036e6ab20.mov

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -107,6 +107,8 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
         onMouseEnter: onMouseEnter,
         onMouseLeave: onMouseLeave,
         accessibilityLabel: accessibilityLabel,
+        focusable: !disabled,
+        accessible: !disabled,
```
Saadnajmi, line 111: @lenahong these changes affect win32. Are these OK? I can make them macOS only if not.

packages/components/Callout/src/Callout.types.ts
```diff
@@ -97,8 +98,9 @@ export interface ICalloutTokens extends IBackgroundColorTokens, CalloutBorderTok
   minPadding?: number;
 }
 
-export interface ICalloutProps extends ViewProps, ICalloutTokens {
-  /*
+export interface ICalloutProps extends IViewProps, ICalloutTokens {
+  /**
+   * @platform win32
```
Saadnajmi, line 103: @rurikoaraki Something I've been meaning to do is add these tags for platform specific props. Just an FYI

harrieshin (approved):

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
Saadnajmi: Nevermind, I just made this macOS specific. But also added more accessbilityprops like AccessibilityState =)

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -103,10 +103,25 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
...
-        onKeyUp: onKeyUp,
-        onMouseEnter: onMouseEnter,
-        onMouseLeave: onMouseLeave,
+        onKeyUp,
+        onMouseEnter,
+        onMouseLeave,
+        accessible: true,
         accessibilityLabel: accessibilityLabel,
+        accessibilityRole: 'menuitem',
+        accessibilityState: { disabled: state.disabled, selected: state.selected },
```
Saadnajmi, line 112: Actually, @lenahong I just tested on win32, and setting accessibilityState here means that if the menu item is disabled, it will not receive keyboard focus. Is that acceptable behavior?

lenahong: I checked and that is the expected behavior on win32 as well! Thanks for checking with me.

PR #1216, opened 2021-12-06
Autolink Drawer Android Native Module to FURN Drawer

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [X] android

### Description of changes
This PR adds a bare native module with gradle build environment for FURN Android.

packages/experimental/Drawer/android/src/main/java/com/microsoft/frnandroid/drawer/DrawerModule.kt
     - Adds getName and getConstants methods to Drawer module
packages/experimental/Drawer/android/src/main/java/com/microsoft/frnandroid/drawer/DrawerPackage.kt
     - Adds the DrawerModule to drawer native module package
apps/android/package.json
     - Autolink gradle script looks into package.json to find the Native code to be added into the project
     
### Verification
FluentTester's Drawer logs from Drawer's getConstants method, accessed through React Native's NativeModules

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/android/package.json
```diff
@@ -24,6 +24,7 @@
     "@fluentui-react-native/tester": "^0.48.1",
     "@react-native-community/slider": "^4.1.12",
     "@react-native-picker/picker": "^1.16.7",
+    "@fluentui-react-native/experimental-drawer": "*",
```
rurikoaraki, line 27: nit: alphabetical please

packages/experimental/Drawer/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@fluentui-react-native/eslint-config-rules": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.63.7",
+    "@types/react-native": "^0.63.0",
+    "@uifabricshared/build-native": "^0.1.1",
+    "react": "16.13.1",
+    "react-native": "^0.63.4"
+  },
+  "peerDependencies": {
+    "react": ">=16.13.1",
+    "react-native": ">=0.63.4"
```
rurikoaraki, line 46: We've upgraded to RN 0.64, please update the RN versions you're listing in your dependencies. React has been bumped as well

packages/experimental/Drawer/jest.config.js
```diff
@@ -0,0 +1,2 @@
+const { configureReactNativeJest } = require('@uifabricshared/build-native');
```
rurikoaraki, line 1: I don't know if you need this file if you're not running any tests that depend on jest... (like snapshot tests or unit tests)

yarn.lock
```diff
@@ -1282,6 +1282,32 @@
...
+    tslib "^1.13.0"
+
+"@fluentui-react-native/theme-tokens@^0.9.2":
+  version "0.9.2"
+  resolved "https://registry.yarnpkg.com/@fluentui-react-native/theme-tokens/-/theme-tokens-0.9.2.tgz#09ef091c878338f8a80b0b00894d94a1abe9bce7"
+  integrity sha512-c95ntUKM6y6N1pmQdP7uSSYxbeiwAQJYvxapkcK7+WHFedVi5iIZCo7qC/q2zF5+H5xV76DaCJCzDYl9jXk75A==
+  dependencies:
+    "@fluentui-react-native/theme-types" ">=0.12.4 <1.0.0"
+    assert-never "^1.2.1"
+
```
Saadnajmi, line 1310: The presence of these (and other yarn.lock changes) implies your new package is using some version of "@fluentui-react-native/framework" that is out of sync with the rest of the monorepo, so instead of referring to the that dependency from your local clone of the monorepo, it's pulling it from NPM (I think). Might want to double check that.

packages/experimental/Drawer/src/Drawer.tsx
```diff
@@ -0,0 +1,41 @@
...
+    })),
+  },
+  render: (props: DrawerProps, useSlots: UseSlots<DrawerType>) => {
+    const { target, ...rest } = props;
+    const [nativeTarget, setNativeTarget] = React.useState<number | string | null>(null);
+
+    React.useLayoutEffect(() => {
+      if (typeof target === 'string') {
+        // Pass string type `target` directly
+        setNativeTarget(target);
```
Saadnajmi, line 27: Question for my own sake, I know this code was copied from Callout, but what is the actual scenario where a target is a string and not a  ref? I wasn't clear on that..

rurikoaraki: Could be a holdover of older ref behavior? You could use strings as ref in the past, though I believe that's been deprecated: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

(no idea if that's what it's for though, just a guess)

Saadnajmi: My guess too. @PPatBoyd is the one who added the behavior, so I guess I can ask him too.

packages/experimental/Drawer/android/.gitattributes
```diff
@@ -0,0 +1,6 @@
+#
+# https://help.github.com/articles/dealing-with-line-endings/
+#
+# These are explicitly windows files and should use crlf
```
Saadnajmi, line 4: Interesting. I'm curious is this was needed because something broke? Asking because we have some similar android files in react-native-macos that maybe we should add this attribute to..

packages/experimental/Drawer/android/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore Gradle project-specific cache directory
+.gradle
+
+# Ignore Gradle build output directory
+build
```
Saadnajmi, line 5: Could we push this up to the main repos gitignore via `**/build` or something like that, so that future android packages don't need to remember to add this gitignore?

packages/experimental/Drawer/android/gradle.properties
```diff
@@ -0,0 +1,6 @@
+FURNDrawer_kotlinVersion=1.5.31
+FURNDrawer_fluentuiVersion=0.0.21
+FURNDrawer_compileSdkVersion=29
+FURNDrawer_buildToolsVersion=29.0.3
+FURNDrawer_targetSdkVersion=29
+FURNDrawer_minSdkVersion=16
```
Saadnajmi, line 6: Because I don't actually know, what is the min sdk we support for Office, which probably informs what the min FURN sdk would be? 16 (Android 4.1) seems really low, and I think most things do 21+ (Android 5.0+)

lenahong: ah icic I will check the Office one and change it here

lenahong: I just checked hermes and the minSdk version they have there is 16. https://github.com/facebook/hermes/blob/a725c6e9740ac1153f34dd0b20303df592ef50bd/android/build.gradle

packages/experimental/Drawer/android/.gitattributes
lenahong: hmm I'm not actually why this was needed. I'm not familiar with .gitattributes files. What is this supposed to do?

packages/experimental/Drawer/src/Drawer.tsx
lenahong: I can address this in my next pr where I actually add the Drawer behavior. I'll remove this code bit for now

packages/experimental/Drawer/android/.gitattributes
Saadnajmi: Oh, I thought you added this file manually. Didn't realize it was autogenerated. Curious, how did you generate the template files for this native module? Might be good to update the [Contributing Guide](https://github.com/microsoft/fluentui-react-native/blob/master/CONTRIBUTING.md) with best practices afterwards.

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Windows and macOS have different line endings, so git usually has to make a decision on which one to commit to source. I _think_ the standard has been to commit to source linux/macOS line endings (lf) and then if a git repo 
is checked out on windows, update the line endings to crlf on checkout (and back to lf when pushing back to source). I guess since .bat is a windows specific file, it's never meant to be run with linux//macOS line endings (lf), so this file tells git not to convert .bat files.

.gitignore
```diff
@@ -99,4 +99,7 @@ typings/
 
 # VS Code debugging
 .vscode/.react/
-apps/*/.vscode/.react/
\ No newline at end of file
+apps/*/.vscode/.react/
+
+# Ignore Gradle build output directory
+**/build
```
Saadnajmi, line 105: (Sorry for the followup ask):

I looked at our partner repo react-native-test-app, and noticed they use a slightly more exact regex string, which I think might be slightly more exact / better so we don't unintentionally ignore stuff in the future:
`android/**/build/`
https://github.com/microsoft/react-native-test-app/blob/b7224ef3b9336f5bd853ddbc419f8d2129de9bba/.gitignore#L12

Since we have subdirectories, maybe something like this? I'm not sure if "**" encapsulates if it's a top level directory
```
android/**/build/
**/android/**/build/
```

packages/experimental/Drawer/android/.gitattributes
lenahong: https://docs.gradle.org/current/samples/sample_building_java_applications.html#run_the_init_task

I used the gradle init script to initialize the gradle project for Drawer. I can update the Contributing guide later! I might move the gradle project to a shared location in the components directory when we start creating native modules for other components as well

packages/experimental/Drawer/android/gradle.properties
Saadnajmi: It makes sense that hermes might need to support lower, but that's probably not equal to what office supports. I checked react-native-macos (which is what ends up getting used for Android I believe), and it looks like the minsdk is 21:
https://github.com/microsoft/react-native-macos/blob/f88f01a99e0f6fdd33b4e5c6e8ab882aba07b9ef/ReactAndroid/build.gradle#L392

Saadnajmi (approved):

packages/experimental/Drawer/android/gradle.properties
Saadnajmi: (Should preface with I'm not an android dev, just generally aware that minsdks being too low can and have caused problems for me in the past when updating FURN)

.gitignore
lenahong: Just looked up some documentation and `**/android/**/build/` is sufficient I think!

packages/experimental/Drawer/android/src/main/java/com/microsoft/frnandroid/drawer/DrawerPackage.kt
```diff
@@ -0,0 +1,29 @@
...
+class DrawerPackage: ReactPackage {
+
+    override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> {
+        val modules = ArrayList<NativeModule>()
+        modules.add(DrawerModule(reactContext))
+
+        return modules
+    }
+
+    /*override fun createViewManagers(reactContext: ReactApplicationContext): List<ViewManager> {
```
rurikoaraki, line 20: Is this commented code meant to be committed?

packages/experimental/Drawer/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "dependencies": {
+    "@fluentui-react-native/component-cache": "^1.3.0",
+    "@fluentui-react-native/framework": "0.7.0",
+    "tslib": "^1.13.0"
+  },
+  "devDependencies": {
+    "@fluentui-react-native/eslint-config-rules": "^0.1.1",
+    "@types/react-native": "^0.63.0",
```
rurikoaraki, line 30: types should be bumped to 0.64 as well

packages/experimental/Drawer/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@uifabricshared/build-native": "^0.1.1",
+    "immediate": "~3.0.5",
+    "lie" : "~3.3.0",
+    "pako" : "~1.0.2",
+    "react": "16.13.1",
+    "react-native": "^0.63.4",
```
rurikoaraki, line 36: We should bump this to the same version of react=native we're using for peer dependencies.
Same for the react version

rurikoaraki (approved):

rurikoaraki:
You might need to regen the yarn.lock file after your previous changes.

PR #1218, opened 2021-12-07
Add ref to v1 Button

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

The `fluentui` Button uses the built in `ref` functionality from `React`, and we would like to align to do the same. However, React doesn't allow `ref` as a prop on functional components, which our components (under all the layers of the framework) are. `fluentui` uses `React.forwardRef` to allow `ref`s to be passed to their components, which is what `React` advises as well. However because of our compose framework, it wasn't immediately obvious how to follow this behavior.

Turns out, we can just wrap the composed component in a `React.forwardRef`, and then export that component as our `Button` component. Outwardly, we have the expected ref prop available (as exemplified by the menu button), but internally we map that to an `innerRef` prop which allows us to pass the ref to the functional component, basically what `componentRef` was doing before.

Also `ref`s are allowed to be callback functions in `React`, but our API was limiting `ref`s to only being actual `ref` objects. At the moment I am just not firing the "focus on click" behavior if the ref is a function a function it doesn't allow for .focus() to be called on it, but I'm open to suggestions of other ways to handle this case.

### Verification

Verified focus-on-click logic is still active for the button.
Verified that menu button context menu still positioned correctly
Verified that the console is not logging any errors about invalid ref props.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/Button.types.ts
```diff
@@ -84,10 +84,7 @@ export interface ButtonCoreProps extends Omit<IWithPressableOptions<ViewProps>,
    */
   icon?: IconSourcesType;
 
-  /**
-   * A RefObject to access the IButton interface. Use this to access the public methods and properties of the component.
-   */
-  componentRef?: React.RefObject<IFocusable>;
+  innerRef: React.ForwardedRef<IFocusable>;
```
PPatBoyd, line 87: innerRef is still exposed to users, it seems?

rurikoaraki: Yep good catch.

packages/utils/interactive-hooks/src/useOnPressWithFocus.ts
```diff
@@ -11,11 +11,13 @@ export type OnPressWithFocusCallback = () => void;
...
  *         onPressWithFocus() - Callback to set focus after calling the userCallback for onPress
  */
-export function useOnPressWithFocus(focusRef: React.RefObject<any>, userCallback: OnPressCallback): OnPressWithFocusCallback {
+export function useOnPressWithFocus(focusRef: React.ForwardedRef<any>, userCallback: OnPressCallback): OnPressWithFocusCallback {
   const onPressWithFocus = React.useCallback(
     (args?: any) => {
       userCallback && userCallback(args);
-      focusRef?.current?.focus();
+      if (typeof focusRef !== 'function') {
+        focusRef?.current?.focus();
```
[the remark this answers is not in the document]
rurikoaraki: `focusRef` is of type `ForwardedRef`, which can either be `((instance: T | null) => void)` or `MutableRefObject<T | null>` (or `null`). Typescript will complain here that the function variant of `focusRef` does not have property `current`, so I added this check to tell typescript that, hey, if we fall in here, we can only be type `MutableRefObject<T | null>`.

packages/experimental/Button/src/Button.types.ts
rurikoaraki: Should be fixed with latest commits

packages/experimental/Button/src/Button.types.ts
```diff
@@ -73,7 +73,7 @@ export interface ButtonTokens extends ButtonCoreTokens {
   square?: ButtonTokens;
 }
 
-export interface ButtonCoreProps extends Omit<IWithPressableOptions<ViewProps>, 'onPress'> {
+export interface ButtonComposedCoreProps extends Omit<IWithPressableOptions<ViewProps>, 'onPress'> {
```
Saadnajmi, line 76: Why the name change?

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
```diff
@@ -47,4 +47,8 @@ export const CompoundButton = compose<CompoundButtonType>({
   },
 });
 
+export const CompoundButton = React.forwardRef<IFocusable, CompoundButtonProps>((props, ref) => (
+  <CompoundButtonComposed {...props} innerRef={ref} />
+));
```
Saadnajmi, line 52: oh, the name change because we're doing this

Saadnajmi: Would this work?
```js
const Button = React.forwardRef<IFocusable, CompoundButtonProps>((props, ref) => (
  compose<CompoundButtonType>({...})
)
```

compose<CompoundButtonType>({

ValentinaKozlova (approved):

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
rurikoaraki: I've tried this but I haven't gotten it to work, unfortunately. The return type of compose doesn't match what forwardRef expects. It's also not clear how to pass the props/ref args into compose...

Saadnajmi: Ah rip :/

Saadnajmi (approved):
Would be nice if we had less types (ButtonCore, ButtonCoreComposed, ButtonProps, etc...), or maybe something like ButtonPropsWithRef instead of ButtonPropsComposed to make it slightly less confusing. Otherwise, yay using forward ref 🎉🎉🎉

rurikoaraki:
> Would be nice if we had less types (ButtonCore, ButtonCoreComposed, ButtonProps, etc...), or maybe something like ButtonPropsWithRef instead of ButtonPropsComposed to make it slightly less confusing. Otherwise, yay using forward ref 🎉🎉🎉

Ah sure, I can rename the props to have "WithRef" instead. I'll think about whether we can consolidate some of the types

packages/experimental/Button/SPEC.md
```diff
@@ -99,13 +99,19 @@ export type ButtonProps = Omit<IWithPressableOptions<ViewProps>, 'onPress'> & {
...
+   * be interacted with.
+   * @default false
+   */
   loading?: boolean;
 
+  /**
+   * A RefObject to access the IButton interface. Use this to access the public methods and properties of the component.
+   * Note that this does not allow for a ref callback to be passed in.
+   */
+  ref?: React.RefObject<IFocusable>;
```
PPatBoyd, line 113: Will this be a breaking change?

rurikoaraki: Whoops, need to edit that sentence, since it's not true anymore.

It wouldn't be a breaking change, though - componentRef works the same way.

Issue #1033, opened 2021-09-25
Experimental button doesn't have correct border width in focused state


rurikoaraki:
- **Platform**:
  - [x] iOS
  - [x] macOS
  - [x] win32
  - [x] windows
  - [x] android

### Please provide a reproduction of the bug:

Use keyboard to focus on an experimental button

#### Actual behavior:

![image](https://user-images.githubusercontent.com/4602628/134753059-22988181-dcda-440b-8e7b-8bef07089953.png)

#### Expected behavior:

Border should be 2px wide. Height of button should not change

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: Normal

Products/applications affected: Tester

rurikoaraki:
Need to wait for changes to native side before we can fix this.

PR #1228, opened 2021-12-09
Enable native focus visuals for v1 button

### LINKED ISSUE #1162: Bump @office-iss/rex-win32
### Platform

- [ ] iOS
- [ ] macOS
- [X] win32
- [ ] windows
- [ ] android

### Describe the feature that you would like added

We should probably bump our version of @office-iss/rex-win32 to a build that includes the change to RN 0.64 in Office.

### What component or utility would this be added to

_No response_

### Have you discussed this feature with our team, and if so, who

_No response_

### Additional context/screenshots

_No response_

rurikoaraki:
Fixes #1162 and #1033

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Now that the work to enableFocusRing has been done on the native side, we can enable native focus visuals for win32 v1 Button. This change should not affect UWP or MacOS, as they have enableFocusRing set to true by default.

### Verification

Tested that focus visuals are working for win32 as intended. No element height changes or layout changes seen.

Still need to test on other DPIs and text scaling settings.

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/ButtonColorTokens.windows.ts
```diff
@@ -27,9 +27,10 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
       iconColor: t.colors.neutralForeground1Pressed,
     },
     focused: {
+      // Focused border matches background so latency betwen native and JS visuals isn't as noticable
       backgroundColor: t.colors.neutralBackground1Hover,
       color: t.colors.neutralForeground1Hover,
-      borderColor: t.colors.strokeFocus2,
+      borderColor: t.colors.neutralBackground1Hover,
```
Saadnajmi, line 33: You could also make it 'clear', which should make it transparent. I think the apple theme does that a few places.

packages/experimental/Button/src/ButtonColorTokens.windows.ts
```diff
@@ -27,9 +27,10 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
       iconColor: t.colors.neutralForeground1Pressed,
     },
     focused: {
+      // Focused border matches background so latency betwen native and JS visuals isn't as noticable
```
Saadnajmi, line 30: Is it noticeable?

packages/experimental/Button/src/ButtonTokens.ts
```diff
@@ -25,6 +28,9 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
       padding: globalTokens.spacing.xs - globalTokens.stroke.width.thin,
       borderWidth: globalTokens.stroke.width.thin,
       iconSize: 16,
+      focused: {
+        borderWidth: 0,
+      },
```
Saadnajmi, line 33: Effectively making the border disappear? Would this affect the following calculation?
` paddingHorizontal: globalTokens.spacing.l - globalTokens.stroke.width.thin,`

packages/experimental/Button/src/useButton.ts
```diff
@@ -19,6 +19,7 @@ export const useButton = (props: ButtonProps): ButtonState => {
       onAccessibilityTap: props.onAccessibilityTap || props.onClick,
       accessibilityLabel: props.accessibilityLabel || props.content,
       accessibilityState: getAccessibilityState(!!disabled || !!loading),
+      enableFocusRing: true,
```
Saadnajmi, line 22: Since this is in the default props, we can probably remove it from any macOS specific settings / stylings files? I can do that as a followup if you'd rather not touch that with this PR

packages/experimental/Button/src/ButtonColorTokens.windows.ts
rurikoaraki: Yeah, but I'm also looking for it and running the dev bundle.

packages/experimental/Button/src/useButton.ts
rurikoaraki: useButton is only in experimental and i don't see it being set explicitly in any mac files... maybe I'm missing something?

Saadnajmi: .... Thought this was a settings file. My bad.

packages/experimental/Button/src/ButtonTokens.ts
rurikoaraki: It didn't seem to, but I'll change the padding for explicitness

packages/experimental/Button/src/Button.styling.ts
```diff
@@ -24,6 +23,7 @@ export const buttonStates: (keyof ButtonTokens)[] = [
   'rounded',
   'circular',
   'square',
+  'focused',
```
Saadnajmi, line 26: I think you can get this + hovered + pressed by doing a spread of Pressable.states or something similar. But this is a string so I'm not sure

Saadnajmi (approved):

packages/experimental/Button/src/Button.styling.ts
rurikoaraki: That's true, although the ordering here actually matters so I'm not sure if adding a spread of all those states at the end is the best idea. I can do something like `...IFocusState.keys()` I think, so that we make sure we stay in sync with whatever that state is. I'll make that change separately, since other components would probably benefit from that change, too.

ValentinaKozlova:
Thank you for fixing this bug! :)

PR #1233, opened 2021-12-09
Adding MacOS E2E Testing

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

macos_testing

**This is part 1 - Part 2 will include enabling more code-reuse. I was running into a couple errors when attempting to put platform-specific logic (ie. importing {Platform} from 'react-native' that has blocked me for a couple days. Ideally, only 1 page-object should be used per component (across all platforms), but the errors I mentioned above have blocked me on that currently. I'm going to re-visit that once this is in (instead of blocking the whole feature on it).**

- Adding capability of running E2E automation tests on MacOS. It's using the same framework as Win32/Windows, but contains some different driver-specific syntax.
- Updating [WebDriverIO ](https://webdriver.io/)from v5 to v6. This is required for the new [Mac2 driver](https://appium.io/docs/en/drivers/mac2/) we're using from [Appium](https://appium.io/). 
- In having to make .macos specific files, I'm also fixing up other file names that should be windows specific (so adding .win to the file).

It's currently not integrated with the CI yet, only local running. I'm starting to configure that now and will update this PR as I go.

### Verification

Manual + Automated

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Thank you! I will check this out soon!

.ado/templates/e2e-testing-macos.yml
```diff
@@ -0,0 +1,18 @@
...
+    workingDirectory: apps/macos
+    displayName: "Run MacOS E2E Tests"
+    condition: succeeded()
+    continueOnError: true
+
+  # - template: e2e-publish-artifacts.yml
+  #   parameters:
+  #     applicationType: win32
+  #     platform: win32
+  #     buildArtifacts: variables['task.Build.status']
```
Saadnajmi, line 18: Nit: Remove comment or have it publish the artifacts somewhere? Were the artifacts useful for win32?

apps/fluent-tester/package-lock.json
```diff
@@ -0,0 +1,2013 @@
+{
```
Saadnajmi, line 1: We shouldn't need this file? Given we use `yarn` and not nom?

apps/macos/package-lock.json
```diff
@@ -0,0 +1,2013 @@
+{
```
Saadnajmi, line 1: same for this file

apps/macos/package.json
```diff
@@ -17,7 +17,8 @@
     "test": "fluentui-scripts jest",
     "macos": "react-native run-macos --scheme FluentTester --project-path src",
     "prettier": "fluentui-scripts prettier",
-    "prettier-fix": "fluentui-scripts prettier --fix true"
+    "prettier-fix": "fluentui-scripts prettier --fix true",
+    "e2etest": "wdio"
```
Saadnajmi, line 21: Say I wanted to in the future move all the platform specific test apps into one big test app: Would it be possible to add a platform flag to this command?

apps/win32/package.json
```diff
@@ -45,22 +45,23 @@
...
     "metro-config": "^0.66.2",
     "metro-react-native-babel-preset": "^0.66.2",
     "react-native-svg-transformer": "^0.14.3",
     "react-test-renderer": "17.0.1",
     "ts-node": "^8.10.1",
     "tsconfig-paths": "^3.9.0",
     "typescript": "3.8.3",
-    "webdriverio": "5.22.4"
+    "webdriverio": "^6.5.0",
+    "appium-mac2-driver": "0.14.1",
```
Saadnajmi, line 58: win32 needs the mac2 driver?

apps/windows/package.json
```diff
@@ -48,7 +42,14 @@
     "ts-node": "^8.10.1",
     "tsconfig-paths": "^3.9.0",
     "typescript": "3.8.3",
-    "webdriverio": "5.22.4"
+    "webdriverio": "^6.5.0",
+    "appium-mac2-driver": "0.14.1",
```
Saadnajmi, line 46: same comment about windows

apps/win32/package.json
samuelfreiberg: No, good catch. Removing for Win32/Windows now. Thanks

apps/windows/package.json
samuelfreiberg: No, good catch. Removing for Win32/Windows now. Thanks

apps/macos/package.json
samuelfreiberg: I'd have to learn a bit more about how this would work. So each platform would launch the same app? You can specify what app you want the test harness to deploy in each platforms WebDriverIO config file (wdio.conf,js)

Saadnajmi: Yeah, I've had the idea that instead of `apps/<platform>` to run each app, there should just be one `apps/FluentTester` app that runs on all platforms that `react-native-test-app` supports (iOS, Android, macOS, Windows). I've never quite gotten it working, so it's not really on my or anyone's roadmap.

apps/fluent-tester/package-lock.json
samuelfreiberg: Removing this - not sure why it was added :)

apps/macos/package-lock.json
samuelfreiberg: Removing this - not sure why it was added :)

Saadnajmi (approved):

PR #1234, opened 2021-12-09
Refactor useKeyCallback to useKeyUpProps / useKeyDownProps

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

( I would like to instead change react-native-macos upstream to not require this, but in the meantime...

macOS needs an extra prop `validKeysUp/Down` (the array of the keys that should trigger onKeyUp/Down) in order for key callbacks to work. To support this, I refactored the `useKeyCallback` hook (which previously returned a callback. into `useKeyProps` (which now returns an object with the correct platform specific props). In effect, the hooks name changed, and to use it you need to spread the returned object into your component rather than pass it into an `onKeyUp` prop. I also split it into `useKeyUpProps` and `useKeyDownProps`, previously `useKeyCallback` assumed you wanted `onKeyUp`. 

### Verification

Key events still work on macOS. I still need to verify on win32. windows is explicitly not supported, but we should probably just fix that...

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
useKeyCallback is exported (and used) publicly, so we should probably put a deprecation warning instead of removing it like I currently do.

rurikoaraki (approved):

PR #1240, opened 2021-12-11
Rename slots in ContexutalMenu from leftstack/rightstack to startstack/endstack

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change does 2 things, both of which should be a no-op on the final behavior.

1) Renames the left/right-stack slots on SubMenuItem to the more RTL friendly  start/end-stack. Note this is not a behavior change, just a name change. The placement was already flipping in RTL, the names make it more clear
2) Moves some accessibility props out of .settings files and directly into the main component file. This is mostly because I feel like .settings files should mostly encapsulate visual styling, and accessibility props should be shared so that one platform doesn't accidentally forget to add one.

### Verification

This is basically a no-op, and components should behave as they did before.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -80,7 +81,7 @@ export const ContextualMenu = compose<ContextualMenuType>({
   },
   settings: settings,
   slots: {
-    root: Callout,
+    root: View,
```
rurikoaraki, line 84: Is this change intentional?

Saadnajmi: No.... good catch

Saadnajmi: This is my super-not-so-secret trick for testing stuff without a native module, which should definitely not be committed.

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -122,38 +122,38 @@ export const SubmenuItem = compose<SubmenuItemType>({
   settings,
   render: (Slots: ISlots<SubmenuItemSlotProps>, renderData: SubmenuItemRenderData, ...children: React.ReactNode[]) => {
     const xml = `
-    <svg width="12" height="12" viewBox="0 0 2048 2048" >
+    <svg width="12" height="12" viewBox="0 0 2048 2048">
       <path class='OfficeIconColors_HighContrast' fill='currentColor' d='M 743 1767 l -121 -121 l 708 -707 l -708 -708 l 121 -121 l 828 829 z' />
-      <path class='OfficeIconColors_m22' fill='currentColor' d='M 743 1767 l -121 -121 l 708 -707 l -708 -708 l 121 -121 l 828 829 z' />
+      <path class='OfficeIconColors_m22'          fill='currentColor' d='M 743 1767 l -121 -121 l 708 -707 l -708 -708 l 121 -121 l 828 829 z' />
```
rurikoaraki, line 127: nit: Why the extra spaces here?

rurikoaraki (approved):

packages/components/ContextualMenu/src/SubmenuItem.tsx
Saadnajmi: Uh, in vscode it lines up the two svg elements and makes it more obvious they're the same path. Speaking of, @valentinakozlova why are these the same path?

ValentinaKozlova: @Saadnajmi , I don't know. I didn't add it for SubmenuItem. @lenahong , do you know why is there the same path?

Saadnajmi: Right, sorry, You did it for MenuButton, my bad =)

PR #1251, opened 2021-12-15
ContextualMenu fixes

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change does several code refactoring changes in ContextualMenu:

#### 1 - Always call both Context.onClick() & ContextualMenuItem.onClick()

 Previously, ContextualMenuItem / SubmenuItem would execute either their own onClick Handler, or the ContextualMenuItem's onClick Handler if they didn't have their own. This seemed like a bug, the ContextualMenu's onClick handler is required to update the ContextualMenu's selected key (which in turn helps set the accessibility state of which menu item is selected). I have changed it so that both the ContextualMenu AND ContextualMenuItem/SubmenuItem's onClick handlers are called every time.

#### 2 - Remove the need for `submenuitemHovered`

We used to track `submenuitemHovered` (to set some styling on the SubmenuItem) through the context. We would set it to `true` in a SubmenuItem's `onHoverIn`, and then `false` in _any other item's_ `onHoverIn`. This seemed also unnecessary, we can just set it to false in the same SubmenuItems' `onHoverOut`. Furthermore, if we completely internally manage it's `submenuitemHovered` state internally, there is no need to push it up to the context, so we can remove that code altogether!

#### 3- Clean up some onMouseEnter overrides

Previously, we would use the hook `useAsPressable` to manage hover/focus/press states on a menu item. The `useAsPressable` hook returns a bunch of handlers meant to be passed to a view, including `onMouse(Enter|Leave)`, `on(Focus|Blur)`, `On(PressIn|Press|LongPress|PressOut)`. The problem is we _also_ overrode the menu item's onMouseEnter handler, effectively cancelling out the one returned by `useAsPressable`. 

The reason for this is a little complex: We want to have hover effects show up right away onMouseEnter, but we also want to have a delay between the hover and when the submenu shows up. We achieved this by passing `useAsPressable` `onHoverIn` with an `onHoverInDelay` to handle showing the submenu (which would combine those into an onMouseEnter that executes the user provided callback after the specified delay), but then we _also_ override `onMouseEnter` to set `submenuItemHovered` (which we no longer need to do because of 2 ), and the pressable's returned `onMouseEnter` handler. Then, only because we passed the onMouseEnter handler _after_ the `...pressable.props` spread inside `<View>`, does the second onMouseEnter AKA our overridden one get actually passed in. 

This is super confusing, and I argue only works because of a quirk of javascript. Other repos have lint rules to prevent you from doing this exact kind of thing. I solve this by refactoring the code so we only ever explicitly pass one `onMouseEnter/onMouseLeave` handler to the final `<View>`. Note I would have much rather done _none_ of the overriding and simply had `onHoverIn` fire without a delay, and inside `onHoverIn`, showed the submenu after a delay. I can't do that without changing the external API though, so I'll leave that up for ContextualMenu V1 in the future. 

#### Open issues

On both my change's branch and master, I noticed that hovering off of a SubmenuItem does not collapse the submenu. The only way to dismiss the submenu is to hover onto another `ContextualMenuItem`. This is because (for whatever reason), `context.dismissSubmenu` inside `SubmenuItem` is empty, but not empty in `ContexutalMenuItem`. I need to learn more about contexts to fix this, and this change is already too big, so I'll leave it for another time. 

### Verification

Tested on macOS and win32 and I didn't see any regressions =)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/ContextualMenu.types.ts
```diff
@@ -3,31 +3,31 @@ import { IViewProps } from '@fluentui-react-native/adapters';
...
   isSubmenuOpen?: boolean;
-  /*
-   ** ContextualMenuItems will call this submenu dismissal when they are hovered
+
+  /**
+   * ContextualMenuItems will call this submenu dismissal when they are hovered
    */
   dismissSubmenu?: () => void;
-
-  setSubmenuItemHovered?: React.Dispatch<React.SetStateAction<boolean>>;
```
rurikoaraki, line 30: Since we're shipping this component right now, I would rather have this set as deprecated than to remove it outright, so that we don't introduce breaking changes.

rurikoaraki: Spoke offline, this isn't external so this should be ok.

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -74,38 +76,33 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
...
+    const onMouseLeaveModified = React.useCallback(
       (e) => {
-        pressable.props.onMouseLeave && pressable.props.onMouseLeave(e);
-        pressable.props.onBlur && pressable.props.onBlur(e);
-        e.stopPropagation();
+        pressablePropsModified.onBlur(e);
+        onMouseLeave && onMouseLeave(e);
       },
-      [pressable],
+      [onMouseLeave],
```
rurikoaraki, line 91: Does this need to track pressablePropsModified.onBlur, too?

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -42,58 +42,87 @@ export const SubmenuItem = compose<SubmenuItemType>({
...
+    const pressable = useAsPressable({
+      ...rest,
+      onPress: onItemPress,
+      onHoverIn: onItemHoverIn,
+      delayHoverIn: onHoverInDelay,
+      onHoverOut: onItemHoverOut,
+    });
+
+    /**
+     * GH #blah:
```
rurikoaraki, line 77: Should there be a number for this?

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
Saadnajmi: Yes, thanks!

packages/components/ContextualMenu/src/SubmenuItem.tsx
Saadnajmi: Ah, good catch. I put blah as an obvious-for-others placeholder if I forgot to update it with a real number

Saadnajmi:
Tested on win32 and the behavior seems similar on master and my branch.

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -42,58 +42,88 @@ export const SubmenuItem = compose<SubmenuItemType>({
...
-          context?.onDismissMenu();
-          context?.dismissSubmenu && context.dismissSubmenu();
-          onClick ? onClick() : context?.onItemClick(itemKey);
-          e.stopPropagation();
-        }
+        componentRef.current.focus();
+        onMouseEnter && onMouseEnter(e);
       },
-      [context, disabled, itemKey, onClick],
+      [onMouseEnter],
```
rurikoaraki, line 90: Does this not need to memoize on componentRef as well?

Saadnajmi: My feeling is no, because `componentRef.current` should be properly updated..
demoing on a mutable object seems wrong 😅

rurikoaraki (approved):
Seems good to me, but I'm not a ContextualMenu expert so if others have comments I'll defer to them

PR #1257, opened 2021-12-15
Remove content prop

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Fully remove the content prop from the Button. This includes

Removing content use from FAB. Instead use iconOnly to determine styling like other plats.
Add some logic to figure out the accessibilityLabel needed for the Button if one isn't provided. It would be pretty disruptive for us to add this as a required prop for Buttons with text, so instead add logic to figure it out.
Add content prop to MenuButton - we can deal with that component separately.

### Verification

Verified via Narrator that child text is being read.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/Button.tsx
```diff
@@ -47,10 +47,23 @@ const ButtonComposed = compose<ButtonType>({
...
-    return (final: ButtonPropsWithInnerRef, ...children: React.ReactNode[]) => {
-      const { icon, iconPosition, loading, ...mergedProps } = mergeProps(button.props, final);
+    return (final: ButtonPropsWithInnerRef, children: React.ReactNode[]) => {
+      const { icon, iconPosition, loading, accessiblityLabel, ...mergedProps } = mergeProps(button.props, final);
+      let childText = '';
+      if (accessiblityLabel === undefined) {
+        React.Children.forEach(children, (child) => {
+          if (typeof child === 'string') {
+            if (childText.length !== 0) {
+              childText += ' ';
```
rurikoaraki, line 57: Not sure if this raises i18n concerns... On the other hand, I think our stance for Buttons with more than one child is that you're kind of going into "Wild West" territory and we can't guarantee good experiences with that

Saadnajmi: Would this not add a space even if there is one child?
I would rather Wild West territory be that we only support one string, rather than try to handle more than one string incorrectly

rurikoaraki: Uh yep. That's supposed to be 1. Thanks

packages/experimental/Button/src/Button.types.ts
```diff
@@ -74,16 +74,16 @@ export interface ButtonTokens extends ButtonCoreTokens {
 }
 
 export interface ButtonCorePropsWithInnerRef extends Omit<IWithPressableOptions<ViewProps>, 'onPress'> {
-  /*
-   * Text to show on the Button.
-   */
-  content?: string;
```
ValentinaKozlova, line 80: Thanks for catching this

ValentinaKozlova (approved):

Saadnajmi (approved):

rurikoaraki:
Ughhh this change somehow is making the chevron disappear when we pass it into Button from MenuButton. How is this even possible

PR #1261, opened 2021-12-16
📦 applying package updates ***NO_CI***

dannyvv:


rurikoaraki (approved):

Saadnajmi (approved):

rurikoaraki:
Having this generate change files seeems... counterproductive? We should figure out a way to bypass that check for this PR.

rurikoaraki:
I guess the other way around it is to just tell yarn change that nothing is changing as part of this change? So that we bypass the check but nothing actually results out of it.

rurikoaraki (approved):

PR #1262, opened 2021-12-16
Fix Compound button hover color

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Let's set secondary content hover color the same as default content hover color.

### Verification
https://user-images.githubusercontent.com/67026167/146438174-af86ae2c-305a-4e0b-a913-d3163b36db76.mov

https://user-images.githubusercontent.com/67026167/146438237-e4b40b66-3f88-47c0-91c4-339d5df7cde7.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-apple-theme-5904ecaa-3e32-4b09-a624-13494f8c43a4.json
+++ b/change/@fluentui-react-native-apple-theme-5904ecaa-3e32-4b09-a624-13494f8c43a4.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Fix button hover color",
+  "packageName": "@fluentui-react-native/apple-theme",
+  "email": "email not defined",
+  "dependentChangeType": "none"
+}
--- a/packages/theming/apple-theme/src/appleColors.macos.ts
+++ b/packages/theming/apple-theme/src/appleColors.macos.ts
@@ -490,7 +490,7 @@ export function fallbackApplePalette(): ThemeColorDefinition {
     brandedPressedSecondaryContent: ColorWithSystemEffectMacOS(fluentUIApple.neutralForegroundInverted, 'pressed'),
 
     defaultDisabledSecondaryContent: fluentUIApple.brandForegroundDisabled,
-    defaultHoveredSecondaryContent: fluentUIApple.neutralForegroundInverted,
+    defaultHoveredSecondaryContent: fluentUIApple.neutralForeground3,
     defaultPressedSecondaryContent: fluentUIApple.neutralForegroundInverted,
 
     checkboxBackground: fluentUIApple.communicationBlue,


rurikoaraki (approved):

Saadnajmi:
Should the rest state of each text color also match?

chiuam:
> Should the rest state of each text color also match?

I actually thought about that, but decided to keep the rest state as is because I thought it was intentional to have slightly different text colors to differentiate between secondary and default. I don't think we have it specced so I'm not certain...

any thoughts? @rurikoaraki

rurikoaraki:
> > Should the rest state of each text color also match?
> 
> I actually thought about that, but decided to keep the rest state as is because I thought it was intentional to have slightly different text colors to differentiate between secondary and default. I don't think we have it specced so I'm not certain...
> 
> any thoughts? @rurikoaraki



> > Should the rest state of each text color also match?
> 
> I actually thought about that, but decided to keep the rest state as is because I thought it was intentional to have slightly different text colors to differentiate between secondary and default. I don't think we have it specced so I'm not certain...
> 
> any thoughts? @rurikoaraki

Oh I thought these were from a spec. From web redlines, the colors are supposed to not match. I'd also check the accessibility of the color change and make sure there's enough contrast between foreground and background after this change.

I noticed from this change that the macOS button doesn't directly use alias tokens to style the button unlike win32. It's probably fine since we don't have redlines for mac and this lets the buttons stay more mac-ish, but eventually we should hook the button up to alias tokens for styling.

Saadnajmi (approved):

Saadnajmi:
> > > Should the rest state of each text color also match?
> > 
> > 
> > I actually thought about that, but decided to keep the rest state as is because I thought it was intentional to have slightly different text colors to differentiate between secondary and default. I don't think we have it specced so I'm not certain...
> > any thoughts? @rurikoaraki
> 
> > > Should the rest state of each text color also match?
> > 
> > 
> > I actually thought about that, but decided to keep the rest state as is because I thought it was intentional to have slightly different text colors to differentiate between secondary and default. I don't think we have it specced so I'm not certain...
> > any thoughts? @rurikoaraki
> 
> I noticed from this change that the macOS button doesn't directly use alias tokens to style the button unlike win32. It's probably fine since we don't have redlines for mac and this lets the buttons stay more mac-ish, but eventually we should hook the button up to alias tokens for styling.

Agreed. It would also be good for bundle size (I have discovered) if we can remove the old set of hardcoded colors

chiuam:
> 

@rurikoaraki I verified the color contrast ratio for this change with CCA (Colour Contrast Analyser), seemed to pass the 4:5:1 requirement. 

RE: alias token hook up for Button - work item created for tracking https://office.visualstudio.com/APEX/_workitems/edit/5666577

PR #1268, opened 2021-12-17
ContextualMenu macOS - Override NSResponder.cancelOperation(sender) instead of keyDown to close Callout on escape key press

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Right after I checked in #1255 , @mischreiber pointed out we could have override `cancelOperation` (which NSWindow inherits from NSResponder) to get the same behavior of closing the window on escape key press. Presumably we will get other native behavior of closing a window for free from this, too. Best of all, we no longer need a Carbon header :D 

### Verification

Callout still dismisses on escape key press.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-callout-8e3b6be0-8bbc-434f-816b-a2363deea498.json
+++ b/change/@fluentui-react-native-callout-8e3b6be0-8bbc-434f-816b-a2363deea498.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "remove carbon header",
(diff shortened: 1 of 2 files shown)


mischreiber (approved):
Goodbye Carbon 👋

PR #1271, opened 2021-12-17
Remove content prop from Button take 2

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This is the change for #1257 plus a fix to make accessibilityLabels work again. Thanks @samuelfreiberg for finding the typo.

### Verification

Validated accessibilityLabel works correctly with Narrator.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/Button.tsx
```diff
@@ -47,10 +47,20 @@ const ButtonComposed = compose<ButtonType>({
...
+      const { icon, iconPosition, loading, accessibilityLabel, ...mergedProps } = mergeProps(button.props, final);
+      let childText = '';
+      if (accessibilityLabel === undefined) {
+        React.Children.forEach(children, (child) => {
+          if (typeof child === 'string') {
+            childText = child; // We only automatically support the one child string.
+          }
+        });
+      }
+      const label = accessibilityLabel ? accessibilityLabel : childText;
```
samuelfreiberg, line 60: I believe we can use the [Nullish Operator ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)here: `const label = accessibilityLabel ?? childText`

rurikoaraki: Thanks, done.

packages/experimental/Button/src/Button.types.ts
```diff
@@ -74,17 +74,17 @@ export interface ButtonTokens extends ButtonCoreTokens {
...
   /*
    * Source URL or name of the icon to show on the Button.
    */
   icon?: IconSourcesType;
 
+  /**
+   * Button contains only icon, there's no text content
+   * Must be set for button to style correctly when button has not content.
+   */
+  iconOnly?: boolean;
```
samuelfreiberg, line 86: So this is mutually exclusive to having text for the Button? If a user has both text + iconOnly=true, what happens?

rurikoaraki: Currently the result would be the styling would look a little weird.

We have a few bugs around mutually exclusive props that I'm hoping to clean up today or tomorrow.

rurikoaraki: Handled in #1283

packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
```diff
@@ -29,10 +29,24 @@ const CompoundButtonComposed = compose<CompoundButtonType>({
...
+      let childText = '';
+      if (accessibilityLabel === undefined) {
+        React.Children.forEach(children, (child) => {
+          if (typeof child === 'string') {
+            childText = child; // We only automatically support the one child string.
+          }
+        });
+      }
+
+      if (secondaryContent) {
```
samuelfreiberg, line 42: This can be moved in the above if statement (after line 39). If accessibilityLabel !== undefined, this logic doesn't matter, and thus should be moved (small perf reasons :) )

rurikoaraki: Thanks for catching, done.

samuelfreiberg (approved):

PR #1273, opened 2021-12-18
@rnx-kit package updates

afoxman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes
Update to the latest versions of `@rnx-kit` packages used by FURN.

The CLI, which is used to bundle the test apps, now does:

- cyclic package dependency detection
- duplicate package dependency detection
- type-checking

This exposed some duplicate packages and type errors that already existed in the monorepo. I've left duplicate package dependency detection on, but it won't fail the build. I've turned off type-checking, for now, until fixes can be made. I have some fixes which I'll propose in a separate PR, but more work will be needed.

### Verification

Builds and CI loop tests pass.

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #1275, opened 2021-12-18
Type fixes

afoxman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes
In PR #1273, `@rnx-kit` packages were updated, making it possible to turn on bundle-time type-checking. Doing so leads to a number of type errors in the monorepo and the test code, so it is currently off. This PR fixes most of the type issues that come up.

This is what is fixed:

- Add missing types package to interactive-hooks
- Enable JSON module resolution in theme-tokens, as it makes use of this capability and is required for proper type-checking
- Update useViewCommandFocus() on win32. It now returns a type that encompasses the union of a View (android/ios-combined) and IViewWin32 (win32 extensions to view)
- Make the JSAvatar's root slot component use IViewProps (plat-specific view props), rather than ViewProps (android/ios-combined props). This matches the JSAvatarProps definition which is based on IViewProps, no ViewProps
- Update the experimental Text control so it exports a Win32 Text control for Win32 bundles and a react-native Text control for all other platforms. This gives the user access to Win32-specific props in the type signatures and at runtime
- Add missing descriptor string for Windows TabsItem tests
- RadioButton and TabsItem expose two accessibility props (position in set, set size) across all platforms. These props are only exposed in the types for @office-iss/react-native-win32, not react-native. To compensate for that, add the missing props to each control with a comment explaining why they are in both places
- PressableUsage tests include keyboard events, but those only apply to desktop platforms (win32, macos). Fork the test code to reflect these different levels of support
- Shimmer doesn't appear to be supported on Win32, so add a stub control for it which does nothing

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.win32.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.win32.tsx
@@ -1,0 +1,10 @@
+import * as React from 'react';
+import { View, Text } from 'react-native';
+
+export const ShimmerTest: React.FunctionComponent = () => {
+  return (
+    <View testID="ScrollViewAreaForComponents">
+      <Text>{"Shimmer isn't supported on Win32."}</Text>
+    </View>
+  );
+};
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.desktop.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.desktop.tsx
@@ -3,29 +3,46 @@ import { Alert, Linking, View } from 'react-native';
 import { Stack } from '@fluentui-react-native/stack';
 import { Text } from '@fluentui/react-native';
 import { stackStyle } from '../Common/styles';
+import type { IKeyboardEvent, IHandledKeyboardEvent } from '@office-iss/react-native-win32';
 
 export const PressableUsage: React.FunctionComponent = () => {
   const _onPress = (): void => {
     Linking.openURL('https://www.microsoft.com');
   };
 
+  const _onKeyDown = (ev: IKeyboardEvent) => {
+    if (ev.nativeEvent.key === 'Enter' || ev.nativeEvent.key == 'Space') Linking.openURL('https://www.microsoft.com');
+  };
+
   const _onPress2 = (): void => {
     Alert.alert('Alert', 'Success!');
   };
+
+  const _onKeyDown2 = (ev: IKeyboardEvent) => {
+    if (ev.nativeEvent.key === 'Enter' || ev.nativeEvent.key == ' ') Alert.alert('Alert', 'Success!');
+  };
+
+  const handledNativeKeyboardEvents: IHandledKeyboardEvent[] = [{ key: 'Enter' }, { key: 'Space' }];
 
   return (
     <View>
       <Stack style={stackStyle} gap={5}>
         <Text variant={'bodyStandard'}>
           To learn more about microsoft, visit this{' '}
-          <Text variant={'bodyStandard'} color="blue" onPress={_onPress}>
+          <Text variant={'bodyStandard'} color="blue" keyDownEvents={handledNativeKeyboardEvents} onPress={_onPress} onKeyDown={_onKeyDown}>
             webpage
           </Text>{' '}
           for more details.
         </Text>
         <Text variant={'bodyStandard'}>
           Press{' '}
-          <Text variant={'bodyStandard'} color="blue" onPress={_onPress2}>
+          <Text
+            variant={'bodyStandard'}
+            color="blue"
+            keyDownEvents={handledNativeKeyboardEvents}
+            onPress={_onPress2}
+            onKeyDown={_onKeyDown2}
+          >
             here
           </Text>{' '}
           to view an alert.
--- a/apps/fluent-tester/src/FluentTester/TestComponents/TextExperimental/PressableUsage.desktop.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/TextExperimental/PressableUsage.desktop.tsx
@@ -3,29 +3,46 @@ import { Alert, Linking, View } from 'react-native';
 import { Stack } from '@fluentui-react-native/stack';
 import { Text } from '@fluentui-react-native/experimental-text';
 import { stackStyle } from '../Common/styles';
+import type { IKeyboardEvent, IHandledKeyboardEvent } from '@office-iss/react-native-win32';
 
 export const PressableUsage: React.FunctionComponent = () => {
   const _onPress = (): void => {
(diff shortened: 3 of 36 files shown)


packages/experimental/Text/src/Text.tsx
```diff
@@ -12,9 +12,10 @@ import {
   compressible,
   patchTokens,
 } from '@fluentui-react-native/framework';
-import { Text as RNText, ColorValue } from 'react-native';
+import { ColorValue } from 'react-native';
 import { ITextProps } from '@fluentui-react-native/adapters';
 import React from 'react';
+import { Text as RNText } from './TextComponent';
```
Saadnajmi, line 18: What was wrong with the previous approach? Just that we needed to import from 'react-native' for most platforms, but from '@office-iss/react-native-win32' for others? Would we need to do similar imports from 'react-native-windows' / 'react-native-macos' ?

apps/fluent-tester/src/FluentTester/TestComponents/Shimmer/ShimmerTest.win32.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { View, Text } from 'react-native';
+
+export const ShimmerTest: React.FunctionComponent = () => {
+  return (
+    <View testID="ScrollViewAreaForComponents">
+      <Text>{"Shimmer isn't supported on Win32."}</Text>
```
Saadnajmi, line 7: This is not true, Shimmer is explicitly supported on win32. Will need to figure out why the type check didn't work...

apps/fluent-tester/src/FluentTester/TestComponents/Text/PressableUsage.desktop.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const _onPress2 = (): void => {
+    Alert.alert('Alert', 'Success!');
+  };
+
+  const _onKeyDown2 = (ev: IKeyboardEvent) => {
+    if (ev.nativeEvent.key === 'Enter' || ev.nativeEvent.key == ' ') Alert.alert('Alert', 'Success!');
+  };
+
+  const handledNativeKeyboardEvents: IHandledKeyboardEvent[] = [{ key: 'Enter' }, { key: 'Space' }];
```
Saadnajmi, line 25: Because macOS uses the prop `validKeysDown` instead of `keyDownEvents`, this won't work on macOS. We have a hook ( `useKey(Up|Down)Props` from `@fluentui-react-native/interactive-hooks`) that returns the right props per platform that can be spread to the final component. I should make a bug to track having it return a `keyDownEvents` array for windows/win32, so that we can use it here.

I also plan to investigate if we can just make the macOS prop align names with the windows prop (`validKeysDown` -> `keyDownEvents`), but that's outside the scope of this PR for sure..

apps/fluent-tester/src/FluentTester/TestComponents/TextExperimental/PressableUsage.desktop.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react';
```
Saadnajmi, line 1: Why is this duplicated in `TestExperimental`?

packages/components/RadioGroup/src/RadioButton.types.ts
```diff
@@ -47,6 +47,12 @@ export interface IRadioButtonProps extends IPressableProps {
    */
   ariaSetSize?: number;
 
+  // Defined in IViewWin32, and duplicated here because RadioButton has already shipped supporting these properties. Rather than
+  // remove them from non-Windows platforms, they are being added here to align the types across all platforms. TypeScript will
+  // merge the prop definitions without complaint.
+  accessibilityPositionInSet?: number;
+  accessibilitySetSize?: number;
+
```
Saadnajmi, line 55: Do we not inherit these from IPressableProps on line 12? My impression was IPressableProps should inherit from IViewProps (from `@fluentui-react-native/adapters`, which should have the right props per platform).

packages/components/Tabs/src/TabsItem.settings.windows.ts
```diff
@@ -2,6 +2,8 @@ import { tabsItemName, TabsItemType } from './TabsItem.types';
 import { IComposeSettings } from '@uifabricshared/foundation-compose';
 import type { IViewProps } from '@fluentui-react-native/adapters';
 
+export const tabsItemSelectActionLabel = 'Select a TabsItem';
+
```
Saadnajmi, line 6: Where is this used? A hardcoded English string would not be good for internationalization if it's shipping with the control...

packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import { findNodeHandle, UIManager } from 'react-native';
+import { findNodeHandle, UIManager, View } from 'react-native';
 import { IViewWin32 } from '@office-iss/react-native-win32';
 
 const setAndForwardRef = require('./setAndForwardRef');
 
-export type IFocusable = IViewWin32;
+export interface IFocusable extends View, IViewWin32 {}
```
Saadnajmi, line 7: Could this just be IView from adapters?
I actually think there's no point to make this hook win32 and macOS only (all platforms should support imperative focus calls), so I wonder if we could just make it completely generic that way...

packages/components/Tabs/src/TabsItem.settings.windows.ts
Saadnajmi: Note: this is already in other platform forks, so this is a bug on us

packages/experimental/Text/src/TextComponent.win32.ts
```diff
@@ -0,0 +1 @@
+export { TextWin32 as Text } from '@office-iss/react-native-win32';
```
Saadnajmi, line 1: Note: Check if we actually need this, or if `Text` / `TextWin32` are the same

packages/theming/theme-tokens/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@uifabricshared/build-native/tsconfig.json",
   "compilerOptions": {
     "outDir": "lib",
+    "resolveJsonModule": true,
```
Saadnajmi, line 5: @rurikoaraki Talking to @afoxman about this, and we need this flag to import json properly

packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
Saadnajmi: Should be fine as the adapter

Saadnajmi:
Closing as this will be split into different PRs

PR #1282, opened 2021-12-21
Fix spacing around text when there's an icon

### LINKED ISSUE #1266: [Win32][Button] There's an extra space for Buttons with icon
**Steps:**
1. Create Button with icons from the left and right sides.
2. Open Experimental Button page in Tester.

**Actual result:**
If icon is from the left side, extra space will be from the right. And the same thing for right icons.
It's because of this code in Button.styling.ts
```
const spacingIconContent = tokens.spacingIconContent
    ? {
        marginLeft: tokens.spacingIconContent,
        marginRight: tokens.spacingIconContent,
      }
    : {};
```
image

**Expected result:**
There should be conditiion for position of icon. Currently both margins are added by default.

rurikoaraki:
Fixes #1266 

### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Currently the spacing between an icon and text is being applied to both sides of the text instead of just the side with the icon. This causes extra spacing on the side the icon is not on.

In order to fix this, a new token is added to separate spacing for an icon placed before vs. after the text. This is so that the correct margin is applied to the text when we apply the tokens to buildProps. The result is:

image
(background colors are not part of the change, they're there to make sure spacing is correct).

The other potential solution is to introduce a View in between the icon and the text with the correct width and avoid the token duplication, but while the code is arguably cleaner that way, extra elements do add some perf and it's probably less costly to calculate and apply the correct margins (I haven't actually measured this though).

Also you may notice that the spacing between the icon and the edge of the text is one pixel more than expected for the Medium and Large buttons. The M and the L letters don't have the fuzzy pixels on the left but the text does take up that space, so it looks like there's an extra pixel there.

### Verification

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/Button.tsx
```diff
@@ -28,7 +28,8 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
     layer === userProps['shape'] ||
     (!userProps['shape'] && layer === 'rounded') ||
     (layer === 'hasContent' && !userProps.iconOnly) ||
-    (layer === 'hasIcon' && (userProps.icon || userProps.loading))
+    (layer === 'hasIconAfter' && (userProps.icon || userProps.loading) && userProps.iconPosition === 'after') ||
+    (layer === 'hasIconBefore' && (userProps.icon || userProps.loading) && (!userProps.iconPosition || userProps.iconPosition === 'before'))
```
Saadnajmi, line 32: Is it possible to run into a state where you can have both `haveIconBefore` and `userProps.iconPosition === 'after'` ?

packages/experimental/Button/src/FAB/FAB.styling.ts
```diff
@@ -31,11 +31,11 @@ export const stylingSettings: UseStylingOptions<ButtonCorePropsWithInnerRef, But
         style: {
           color: tokens.color,
           ...getTextMarginAdjustment(),
-          ...(tokens.spacingIconContent && { marginLeft: tokens.spacingIconContent }),
+          ...(tokens.spacingIconContentBefore && { marginLeft: tokens.spacingIconContentBefore }),
```
Saadnajmi, line 34: Should be marginStart to account for RTL?

packages/experimental/Button/src/ToggleButton/ToggleButton.styling.ts
```diff
@@ -30,13 +29,10 @@ export const stylingSettings: UseStylingOptions<ToggleButtonPropsWithInnerRef, T
     content: buildProps(
       (tokens: ToggleButtonTokens, theme: Theme) => ({
         style: {
-          color: tokens.color,
-          ...getTextMarginAdjustment(),
-          ...(tokens.spacingIconContent && { marginLeft: tokens.spacingIconContent }),
```
Saadnajmi, line 35: likewise, marginStart

packages/experimental/Button/src/Button.tsx
rurikoaraki: It shouldn't be, since iconPosition can only be before, after, or undefined.

packages/experimental/Button/src/FAB/FAB.styling.ts
rurikoaraki: Yeah, I should switch to marginStart/end here, thanks for the catch. I think in reality, RTL will flip marginLeft/Right in RN so that marginLeft will be applied to the right in RTL, according to https://github.com/microsoft/react-native-windows/issues/4645#issuecomment-616890884.

packages/experimental/Button/src/Button.tsx
Saadnajmi: I guess the question was more, could we run into a state where the user provided prop and the layer string are out of sync?

rurikoaraki: I'm a little confused. The way the lookup function works is that it takes the list of states that we tell the component it can have and runs through each one asking if the state applies. Based on the states that returned true, the component will look at the token definitions to apply the correct styles for that state.

So this part of the condition basically asks if we're checking for the hasIconBefore state/condition, and if so, returns true is iconPosition is either undefined or set to "before". Hopefully that clears things up?

packages/experimental/Button/src/FAB/FAB.styling.ts
Saadnajmi: Oh.... that sounds wrong to me. I thought the whole point of marginStart/End was so that marginLeft is __always_ left, but marginStart flips. 

https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps#overview-changes-in-rn-for-rtl-support

Sounds like a RNW bug though if true

Saadnajmi: ....oh
https://github.com/microsoft/react-native-windows/issues/4645#issuecomment-616890884

packages/experimental/Button/src/Button.tsx
Saadnajmi: Oh, yes I misunderstood this function derp. It's going through every layer, so we'll never hit that case. My bad!

Saadnajmi (approved):

ValentinaKozlova (approved):

PR #1283, opened 2021-12-21
Make sure we handle Button props that don't work with each other

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Two changes:

- iconOnly shouldn't be applied when you have text content on the Button. We can't handle that case before the styling is applied due to children not being available when slotProps are built, so instead I added a warning to fire in dev mode when both occur on the same Button
- loading should override icon when it's set, per design.

### Verification

Tested cases in tester

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/CompoundButton/CompoundButton.tsx
```diff
@@ -29,7 +30,20 @@ const CompoundButtonComposed = compose<CompoundButtonType>({
 
     // now return the handler for finishing render
     return (final: CompoundButtonPropsWithInnerRef, ...children: React.ReactNode[]) => {
-      const { icon, secondaryContent, iconPosition, accessibilityLabel, ...mergedProps } = mergeProps(button.props, final);
+      const { icon, iconOnly, secondaryContent, iconPosition, loading, accessibilityLabel, ...mergedProps } = mergeProps(
+        button.props,
+        final,
+      );
+
+      const shouldShowIcon = !loading && icon;
```
ValentinaKozlova, line 38: We'll need to find a way how to avoid code duplication. Part with markup is similar in all Button components. Maybe we can move it in one file and reuse. It's just an idea, not for this PR

ValentinaKozlova (approved):

PR #1288, opened 2021-12-22
Fix RTL for ContextualMenu on macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Previously, ContextualMenu's would show up as completely blank on macOS if we ran FluentTester in RTL. This is because in our native code, the x coordinate for the view was off, so the actual view would not be within the bounds of the Callout. This is because within Callout, there is _a lot_ of manual positioning code that converts coordinates between different coordinate spaces of views/windows/screens. 

A simple fix is to make sure we reset the Callout's proxyView's (the view that holds the RN view hierarchy) frame's origin to zero, ensuring that it is always rendering within the Callout.  

Let's also change the directional hint in RTL to be the mirror of what it is manually via settings. In the future, we should implement the 'bottom_auto_edge' hint to flip in RTL. But really, in the future I think we should re-evaluate the directionalHint enum to be more RN Friendly / LTR-RTL agnostic.

### Verification

Ran FluentTester in RTL and verified that the ContextualMenu and submenus flipped.

Screen Shot 2021-12-22 at 12 37 13 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-callout-72255e1c-311b-440c-801d-68811c410a2c.json
+++ b/change/@fluentui-react-native-callout-72255e1c-311b-440c-801d-68811c410a2c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix RTL for ContextualMenu on macOS",
+  "packageName": "@fluentui-react-native/callout",
+  "email": "sanajmi@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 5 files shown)


HeyImChris (approved):

Saadnajmi:
@samuelfreiberg looks like your last PR is causing the win32 tests to fail. Would you happen to have more context here?

PR #1303, opened 2022-01-04
Fix for incorrect brand color in Experimental Button

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The primary button background color is incorrect in Experimental Button, this is a regression introduced in #1189 where the changes to update the brand colors were made in the generated files instead of the input files. In this change, I also made sure macOS alias tokens are mapped correctly to the [vNext spec](https://www.figma.com/file/OLIFtMJnKfD4FGcnNq9TVo/Fluent-vNext-iOS?node-id=744%3A734). 

### Verification

| image |image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/apple-theme/src/appleColors.macos.ts
```diff
@@ -330,7 +331,7 @@ export function fallbackApplePalette(): ThemeColorDefinition {
     inputPlaceholderText: applePlatform.placeholderTextColor,
 
     // Set the default button tokens to match the Acrylic Button style
-    buttonBackground: fluentUIApple.neutralBackground3,
+    buttonBackground: fluentUIApple.neutralBackground1,
```
Saadnajmi, line 334: We might be able to remove the comment on line 333. When I made the FURN macOS button, I copied the design from the FluentUI Apple macOS button. Now that we may be deviating from that and using design tokens instead, if the comment is not relevant we should remove it.

packages/theming/theme-tokens/src/generated/dark-macos/reactnative/tokens-aliases.json
```diff
@@ -20,35 +20,44 @@
...
   "brandBackground": {
-    "fillColorRest": "#185abd",
-    "fillColorHover": "#1651aa",
-    "fillColorPressed": "#0e336a",
-    "fillColorSelected": "#13458f"
+    "fillColorRest": "#0086f0",
+    "fillColorHover": "#1890f1",
+    "fillColorPressed": "#6cb8f6",
+    "fillColorDisabled": "#d6d6d6",
+    "fillColorSelected": "#3aa0f3"
```
Saadnajmi, line 27: Curious, how do these token updates happen? Are we just copying the latest from the token repo, or does the build-tokens script do that for us?

packages/theming/apple-theme/src/appleColors.types.macos.ts
```diff
@@ -215,6 +215,7 @@ interface FluentAppleButtonTokens {
   neutralBackgroundInverted: ColorValue;
   neutralForegroundInverted: ColorValue;
   neutralForeground2: ColorValue;
+  neutralBackground1: ColorValue;
   neutralBackground2: ColorValue;
```
Saadnajmi, line 219: Outside of this review, but if we do get a finalized design, I wonder if we need to update the FUA macOS button too.. probably for the FUA crew to do.

packages/theming/theme-tokens/src/pipeline-input/token-input-global.macos.json
```diff
@@ -0,0 +1,15 @@
...
+    "FluentUITokensVersion": 0
+  },
+  "Global": {
+    "Font": {
+      "Family": {
+        "Default": { "value": "\"Helvetica Neue\", Helvetica, Arial, \"Microsoft Sans Serif\"" },
+        "Monospace": { "value": "\"Courier New\"" },
+        "Numeric": { "value": "\"Helvetica Neue\", Helvetica, Arial, \"Microsoft Sans Serif\"" }
+      }
+    }
```
Saadnajmi, line 13: Shouldn't this be 'System', AKA Apple's San Fransisco font?

scripts/build-tokens.js
```diff
@@ -33,25 +33,16 @@ async function run() {
...
+    'yarn transform-tokens --in ./packages/theming/theme-tokens/src/pipeline-input/token-input.json --in ./packages/theming/theme-tokens/src/pipeline-input/token-input-global.macos.json --in ./packages/theming/theme-tokens/src/pipeline-input/token-input-brand-dark.macos.json --in ./packages/theming/theme-tokens/src/pipeline-input/token-input-dark.json --in ./packages/theming/theme-tokens/src/pipeline-input/token-input-dark.macos.json --out ./packages/theming/theme-tokens/src/generated/dark-macos --p reactnative',
   );
-  fs.unlinkSync(path.join(process.cwd(), 'packages/theming/theme-tokens/src/generated/dark-macos/reactnative/tokens-global.json'));
   fs.unlinkSync(path.join(process.cwd(), 'packages/theming/theme-tokens/src/generated/dark-macos/reactnative/tokens-controls.json'));
```
Saadnajmi, line 46: If this API matches the same one in python, anywhere we use `path.join()` we should be providing each directory as an arguement (`path.join('foo', 'bar', 'xyz')` instead of `path.join('foo', 'bar/xyz')`. That way, `Path.join()` will inject the correct path separators per OS ( "\" for windows, "/" for Darwin AKA macOS, idk for linux).

Saadnajmi (approved):

packages/theming/apple-theme/src/appleColors.macos.ts
chiuam: Oh... actually I tried to match it to the Acrylic style too so maybe worth keeping? also @harrieshin suggested to follow the macOS Button spec (without design tokens) to style the default button

packages/theming/theme-tokens/src/pipeline-input/token-input-global.macos.json
chiuam: Hmm I'm not sure, will need to revisit the font tokens... let me create a bug

packages/theming/theme-tokens/src/generated/dark-macos/reactnative/tokens-aliases.json
chiuam: They get generated when I include the mappings in the input files like token-input-dark.macos.json. I added them because the macOS/iOS tokens are different from the global tokens. @rurikoaraki please correct me if I'm wrong

scripts/build-tokens.js
chiuam: Thanks for the suggestion, this would motivate us to get the script working on macOS? :)

packages/theming/apple-theme/src/appleColors.macos.ts
Saadnajmi: Oh, if it's still the acrylic style than nvm. Comment still useful!

packages/theming/theme-tokens/src/generated/dark-macos/reactnative/tokens-aliases.json
rurikoaraki: Yeah currently the updates are manual so we change our input files and run the build-tokens script to regen the output if we need any changes. The fluentui-token-pipeline repo only has web tokens

rurikoaraki (approved):

PR #1310, opened 2022-01-05
Rename boxside on Checkbox to labelPosition

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

To align with fluentui's prop name, renaming "boxside" to "labelPosition" and using "before"/"after" instead of "start"/"end"

### Verification

Build and tests pass. Tested checkbox on tester app to ensure the checkbox is still positioned correctly.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Checkbox/src/useCheckbox.ts
+++ b/packages/experimental/Checkbox/src/useCheckbox.ts
@@ -24,7 +24,7 @@ export const useCheckbox = (props: CheckboxProps): CheckboxInfo => {
     accessibilityRole,
     checked,
     defaultChecked,
-    boxSide,
+    labelPosition,
     label,
     onChange,
     componentRef = defaultComponentRef,
@@ -54,7 +54,7 @@ export const useCheckbox = (props: CheckboxProps): CheckboxInfo => {
     ...pressable.state,
     disabled: !!props.disabled,
     checked: isChecked,
-    boxAtEnd: boxSide == undefined || boxSide == 'start' ? false : true,
+    boxAtEnd: labelPosition === 'after' ? true : false,
   };
 
   const onAccessibilityAction = React.useCallback(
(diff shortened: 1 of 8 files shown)


ValentinaKozlova (approved):

packages/experimental/Checkbox/src/useCheckbox.ts
```diff
@@ -54,7 +54,7 @@ export const useCheckbox = (props: CheckboxProps): CheckboxInfo => {
     ...pressable.state,
     disabled: !!props.disabled,
     checked: isChecked,
-    boxAtEnd: boxSide == undefined || boxSide == 'start' ? false : true,
+    boxAtEnd: labelPosition === 'after' ? true : false,
```
Saadnajmi, line 57: Intentional to not also change boxAtEnd? I guess that's a bigger change because a lot of settings refer to that.

Saadnajmi (approved):

packages/experimental/Checkbox/src/useCheckbox.ts
rurikoaraki: Yeah we can rename it, but since it's not something that's used by clients, we don't really need to rename it.

Saadnajmi: I think I slightly prefer it because labelPosition is easier to understand than "boxAtEnd", but it's not review blocking

rurikoaraki: Sure, done

Saadnajmi (approved):

PR #1319, opened 2022-01-05
Change how setAndForwardRef is exported

### LINKED ISSUE #1316: setAndForwardRef causing bundle crashes
A partner has asked that we change the way we export setAndForwardRef due to it causing crashes when running the bundle:

In setAndForwardRef.js , the function gets exported by using

module.exports = setAndForwardRef;
Since there is an import statement in the file, transpiling turns this into

module.exports = setAndForwardRef;
export {};
Webpack creates a module.exports via a readonly property. This leads to a crash when loading the bundle

The ask is for us to change how we export the function.

rurikoaraki:
Fixes #1316 

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Change how setAndForwardRef is exported. The module.exports assignment is apparently causing issues with bundles for partners. Changing export method to using the export keyword. This also means updating places where it is imported to using the import keyword instead of require.

I had to add any to a few types to get things to compile. This is a blocking issue so I'd rather get this fixed first and figure out the typing issues when this is not a blocking issue.

### Verification

Verified js doesn't generate the module.exports lines. Build passes. Loaded tester app, no crash, focus logic runs as expected.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
+++ b/packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import { findNodeHandle, UIManager } from 'react-native';
 import { IViewWin32 } from '@office-iss/react-native-win32';
-
-const setAndForwardRef = require('./setAndForwardRef');
+import { setAndForwardRef } from './setAndForwardRef';
 
 export type IFocusable = IViewWin32;
 /**
--- a/packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
+++ b/packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
@@ -18,11 +18,11 @@ export function useViewCommandFocus(
   /**
    * Set up the forwarding ref to enable adding the focus method.
    */
-  const focusRef = React.useRef<React.Component>();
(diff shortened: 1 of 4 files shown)


packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
```diff
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import { findNodeHandle, UIManager } from 'react-native';
 import { IViewWin32 } from '@office-iss/react-native-win32';
-
-const setAndForwardRef = require('./setAndForwardRef');
+import { setAndForwardRef } from './setAndForwardRef';
```
Saadnajmi, line 4: FWIW, I think this was still using `require()` because it was copied from RN Core

Saadnajmi (approved):

packages/utils/interactive-hooks/src/useViewCommandFocus.win32.ts
Saadnajmi: Thanks for the fix!

PR #1326, opened 2022-01-07
Add circular prop to checkbox

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Add `circular` prop to checkbox. This consists of adding a circular state and exposing a prop that users can set to make the checkbox circular.

The tokens will be revamped in the future, just trying to fill out the set of props for now.

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Checkbox/src/Checkbox.types.ts
+++ b/packages/experimental/Checkbox/src/Checkbox.types.ts
@@ -41,6 +41,8 @@ export interface CheckboxProps extends Omit<IViewProps, 'onPress'> {
 
   /**
    * Allows you to set the checkbox to have circular styling.
+   *
+   * @platform Android, iOS, windows, win32
    */
   circular?: boolean;
 
--- a/packages/experimental/Checkbox/src/Checkbox.types.ts
+++ b/packages/experimental/Checkbox/src/Checkbox.types.ts
@@ -24,7 +24,7 @@ export interface CheckboxTokens extends FontTokens, IForegroundColorTokens, IBac
    * States that can be applied to a checkbox
    */
   disabled?: CheckboxTokens;
-  boxAtEnd?: CheckboxTokens;
+  labelIsBefore?: CheckboxTokens;
   hovered?: CheckboxTokens;
   focused?: CheckboxTokens;
   pressed?: CheckboxTokens;
--- a/packages/experimental/Checkbox/src/Checkbox.types.ts
+++ b/packages/experimental/Checkbox/src/Checkbox.types.ts
@@ -55,7 +55,7 @@ export interface CheckboxProps extends Omit<IViewProps, 'onPress'> {
   /**
    * Allows you to set the checkbox to be at the before (start) or after (end) the label
    *
-   * @default: 'after'
+   * @default after
    */
   labelPosition?: 'before' | 'after';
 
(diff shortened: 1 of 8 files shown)


packages/experimental/Checkbox/src/Checkbox.types.ts
```diff
@@ -38,6 +39,11 @@ export interface CheckboxProps extends Omit<IViewProps, 'onPress'> {
    */
   checked?: boolean;
 
+  /**
+   * Allows you to set the checkbox to have circular styling.
+   */
+  circular?: boolean;
+
```
Saadnajmi, line 46: @chiuam can correct me if I'm wrong, but I think experimental checkbox on macOS is a native module so it can't become circular as of yet. Since that's the case, could we add a `@platform Android, iOS, windows, win32` flag to the end of the comment, or a note that it's not supported on macOS?

packages/experimental/Checkbox/src/CheckboxTokens.ts
```diff
@@ -37,8 +35,11 @@ export const defaultCheckboxTokens: TokenSettings<CheckboxTokens, Theme> = (t: T
     checked: {
       checkmarkOpacity: 1,
     },
-    boxAtEnd: {
+    labelIsBefore: {
       checkboxMarginStart: 4,
       checkboxMarginEnd: 0,
     },
+    circular: {
+      borderRadius: 9999,
```
Saadnajmi, line 43: We can't use `globalTokens.corner.radius.circle` here?

Saadnajmi (approved):

packages/experimental/Checkbox/src/CheckboxTokens.ts
rurikoaraki: Yeah I will convert this file to use globalTokens when I go through and make sure everything aligns to redlines.

packages/experimental/Checkbox/src/Checkbox.types.ts
rurikoaraki: Done

packages/experimental/Checkbox/src/Checkbox.types.ts
```diff
@@ -38,6 +39,13 @@ export interface CheckboxProps extends Omit<IViewProps, 'onPress'> {
    */
   checked?: boolean;
 
+  /**
+   * Allows you to set the checkbox to have circular styling.
+   *
+   * @platform Android, iOS, windows, win32
```
Saadnajmi, line 45: I don't think you need the extra space? Unless you saw something I didn't

rurikoaraki: Do you mean the extra line? I usually add them in as personal preference, but I can take it out.

Saadnajmi: Yes the extra line. I think it's fine to leave it then, as long as it shows up in intellisense.

rurikoaraki: Yeah it shows just fine in intellisense. I just find the extra line break to help me with parsing

PR #1332, opened 2022-01-10
Fix disabled button bug on macOS

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The disabled button is selectable on macOS, so let's not pass in `onClickWithFocus` to pressable if button is disabled.

### Verification
Ensured the disabled button is working as expected. 

Before

https://user-images.githubusercontent.com/67026167/148841745-66ba4550-b918-46c1-873c-ae5ea20a289c.mov

After

https://user-images.githubusercontent.com/67026167/148841708-e0adbf43-51e3-4cc0-a8ea-be30bb0393c2.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Button/src/useButton.ts
+++ b/packages/experimental/Button/src/useButton.ts
@@ -8,7 +8,7 @@ export const useButton = (props: ButtonPropsWithInnerRef): ButtonState => {
   const defaultRef = React.useRef(null);
   const { onClick, innerRef, disabled, loading, ...rest } = props;
   const ref = innerRef !== null ? innerRef : defaultRef;
-  const onClickWithFocus = useOnPressWithFocus(ref, onClick);
+  const onClickWithFocus = disabled ? null : useOnPressWithFocus(ref, onClick);
   const pressable = useAsPressable({ ...rest, onPress: onClickWithFocus });
   const onKeyUpProps = useKeyUpProps(onClick, ' ', 'Enter');
   const isDisabled = !!disabled || !!loading;
--- a/packages/experimental/Button/src/useButton.ts
+++ b/packages/experimental/Button/src/useButton.ts
@@ -8,7 +8,7 @@ export const useButton = (props: ButtonPropsWithInnerRef): ButtonState => {
   const defaultRef = React.useRef(null);
   const { onClick, innerRef, disabled, loading, ...rest } = props;
   const ref = innerRef !== null ? innerRef : defaultRef;
-  // Set focusRef to null if button is disabled to prevent getting keyboard focus. This is a workaround for GH:1336
+  // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
   const focusRef = disabled ? null : ref;
   const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
   const pressable = useAsPressable({ ...rest, onPress: onClickWithFocus });
--- a/packages/experimental/Button/src/useButton.ts
+++ b/packages/experimental/Button/src/useButton.ts
@@ -8,7 +8,9 @@ export const useButton = (props: ButtonPropsWithInnerRef): ButtonState => {
   const defaultRef = React.useRef(null);
   const { onClick, innerRef, disabled, loading, ...rest } = props;
   const ref = innerRef !== null ? innerRef : defaultRef;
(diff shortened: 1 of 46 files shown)


packages/experimental/Button/src/useButton.ts
```diff
@@ -8,7 +8,7 @@ export const useButton = (props: ButtonPropsWithInnerRef): ButtonState => {
   const defaultRef = React.useRef(null);
   const { onClick, innerRef, disabled, loading, ...rest } = props;
   const ref = innerRef !== null ? innerRef : defaultRef;
-  const onClickWithFocus = useOnPressWithFocus(ref, onClick);
+  const onClickWithFocus = disabled ? null : useOnPressWithFocus(ref, onClick);
```
Saadnajmi, line 11: This feels like it violates one of the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level): Don't call hooks conditionally. I would rather we unconditionally call the hook, pass in `disabled` as a prop to the hook, and only focus if `!disabled`. I also still feel like this is a workaround and might be worth leaving a comment in code explaining why we need it / the bug # to track why just setting `focusable` isn't enough

chiuam: Good point, we shouldn't call hooks conditionally to ensure the order of execution is the same in every component render.

How about we pass in null as `ref` if button is disabled so that the hook bypasses calling .focus()? (I think this is overriding `focusable` which explains why the disabled button gets keyboard focus) I prefer this approach because there's seems to be assumption that the component is pressable when `useOnPressWithFocus` is called.

Saadnajmi: That could work, though it still feels like a workaround I would leave a comment / bug number with. What do you mean `focusable` is overridden? The hook doesn't seem to touch how we define `focusable` (line 25).

chiuam: I meant to say that the setter for focusable is disregarded because we're calling .focus() in `useOnPressWithFocus`  if `ref` is valid.

Saadnajmi: IMO, calling `ref.focus()` should not focus on the view if `focusable: false`. That feels like the real bug to me.
Natively on macOS, it seems we are also checking for focusable when we call `acceptsFirstResponder` too https://github.com/microsoft/react-native-macos/blob/877d09aa862aea049cd58ec1545153c7574fc662/React/Views/RCTView.m#L1407

packages/experimental/Button/src/useButton.ts
```diff
@@ -8,7 +8,9 @@ export const useButton = (props: ButtonPropsWithInnerRef): ButtonState => {
   const defaultRef = React.useRef(null);
   const { onClick, innerRef, disabled, loading, ...rest } = props;
   const ref = innerRef !== null ? innerRef : defaultRef;
-  const onClickWithFocus = disabled ? null : useOnPressWithFocus(ref, onClick);
+  // Set focusRef to null if button is disabled to prevent getting keyboard focus. This is a workaround for GH:1336
```
Saadnajmi, line 11: Nit, In the repo we usually have the format `GH #<>: message`

Saadnajmi (approved):

packages/experimental/Button/src/useButton.ts
ksiler: Yeah that's weird. `useOnPressWithFocus` respected the disabled prop with no issue on win32. That workaround seems good for now though.

chiuam:
https://user-images.githubusercontent.com/67026167/149420667-0eb35a85-0163-4d75-b529-8a43974c85fe.mov

Verified that keyboard access skips the disabled/loading button as well. (this is tested on Monterrey using the full keyboard access feature)

packages/experimental/Button/src/useButton.ts
chiuam: Thank you for the feedback Saad and Krystal, I updated the [GH issue](https://github.com/microsoft/fluentui-react-native/issues/1336) to better explain the bug.

Issue #1339, opened 2022-01-12
Type Safety: Clean up some tsconfig.json issues


Saadnajmi:
All of our tsconfig.jsons extend off of `@uifabricshared/build-native/tsconfig.json`. We should change that name to something @fluentui-react-native. 

Also, some packages are missing a tsconfig.json, which means it is autogenerated. We should probably explicitly add one. 

For which ones should be added, see https://github.com/microsoft/fluentui-react-native/pull/1275

PR #1342, opened 2022-01-13
Remove references to uifabricshared

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change does two things:

1) Search and replace all strings "@uifabricshared/build-native" to "@fluentui-react-native/scripts". 

2) Renames the package at our root package.json from " uifabric-react-native" to "@fluentui-react-native/repo".
This is a private package mostly used for monorepo management. The public "@fluentui/react-native"  at libraries/core/package.json" if left untouched

Addresses most of #1339, I still need to add some more tsconfigs which I'll do separately.

### Verification

Only editing private packages, so CI passing should be enough.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/android/just.config.js
+++ b/apps/android/just.config.js
@@ -1,3 +1,3 @@
-const { preset } = require('@uifabricshared/build-native');
+const {preset} = require('@fluentui-react-native/scripts');
 
 preset();
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "uifabric-react-native",
+  "name": "@fluentui-react-native/repo",
   "version": "0.1.1",
   "private": true,
   "description": "",
--- a/scripts/src/webpack/webpack-resources.js
+++ b/scripts/src/webpack/webpack-resources.js
@@ -1,13 +1,5 @@
-const { just } = require('@uifabricshared/build-native');
-const {
-  basicWebpackConfig,
-  htmlOverlay,
-  webpackMerge,
-  tsOverlay,
-  fileOverlay,
-  displayBailoutOverlay,
-  stylesOverlay,
-} = just;
+const { just } = require('@fluentui-react-native/scripts');
+const { basicWebpackConfig, htmlOverlay, webpackMerge, tsOverlay, fileOverlay, displayBailoutOverlay, stylesOverlay } = just;
 
 const _isProduction = process.argv.indexOf('--production') > -1;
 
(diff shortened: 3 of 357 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "uifabric-react-native",
+  "name": "@fluentui-react-native/repo",
```
Saadnajmi, line 2: Calling this out specifically, as I kinda just chose an arbritrary name. Feel free to disagree / push back if you don't like it.

scripts/src/webpack/webpack-resources.js
```diff
@@ -18,17 +10,17 @@ const tsOverlayConfig = tsOverlay({
   loaderOptions: {
     transpileOnly: true,
     compilerOptions: {
-      "declaration": false,
-      "declarationMap": false
-    }
+      declaration: false,
+      declarationMap: false,
+    },
```
Saadnajmi, line 15: It seems prettier did some nontrivial changes here. I'm hoping it doesn't break anything as. web pack is only used for our web-tester iirc, which is unused.

apps/android/just.config.js
```diff
@@ -1,3 +1,3 @@
-const { preset } = require('@uifabricshared/build-native');
+const {preset} = require('@fluentui-react-native/scripts');
```
rurikoaraki, line 1: super nit: I think there should be spaces around "preset"? not sure why prettier didn't catch it.

Saadnajmi: Yeah, VS Code is removing the spaces I add...

Saadnajmi: But not in another just.config below... maybe there's a config in this specific package?

Saadnajmi: The android test app has a .pretterrc.js. Should I remove it?
```
module.exports = {
  bracketSpacing: false,
  jsxBracketSameLine: true,
  singleQuote: true,
  trailingComma: 'all',
};
```

scripts/src/webpack/webpack-resources.js
rurikoaraki: I think it just removed the quote marks because declaration and declarationMap are one word, so you don't need quotes for those for property names in objects.

apps/android/just.config.js
rurikoaraki: Ah interesting. Yeah I think the prettier should be consistent across the repo, but that can be its own PR

rurikoaraki (approved):

PR #1373, opened 2022-01-24
Default to `onKeyDown` on macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change introduces the hook `useKeyProps`, which redirects to `useKeyUpProps` on windows, and `useKeyDownProps` for macOS. 

On macOS, it is preferred to override `onKeyDown` rather than `onKeyUp`. This is not a rule, more a general guideline I have arrived at after some investigation. Here's what I found:

1. Most macOS system controls do their actions on key down
2. Windows supports keyboard input cancellation, while macOS does not
 - By "keyboard input cancellation", what I mean is you can do the following:
   - Focus on a button
   - Key down (doesn't perform the click just yet)
   - Tab to move focus away from a button
   - key up will no longer perform the click
 - This is similar to moving the mouse off of a button between `mouseDown` and `mouseUp` to cancel the click.
3. macOS will play a "bad keypress" feedback sound if you don't override keyDown
  - This is hard to explain, so see the attached video

https://user-images.githubusercontent.com/6722175/150856345-6e9b8ca5-db5e-48cb-8375-124eca233972.mov

### Verification

Will test on win32 and windows shortly (where overriding `onKeyUp` is preserved), but macOS test app handles keyboard events fine.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


apps/macos/app.json
```diff
@@ -2,5 +2,5 @@
   "name": "FluentTester",
   "displayName": "FluentTester",
   "components": [{ "appKey": "FluentTester" }],
-  "resources": ["dist/assets", "dist/index.macos.bundle"]
+  "resources": ["dist/assets", "dist/index.macos.jsbundle"]
```
Saadnajmi, line 5: Turns out I was wrong, and it was jsbundle here 🤷🏾

HeyImChris (approved):
Just to confirm here- I know NSMenus respond to key down events and would guess the others do too, but you've confirmed each control with its native counterpart on macOS does respond to key down and not key up?

Saadnajmi:
> Just to confirm here- I know NSMenus respond to key down events and would guess the others do too, but you've confirmed each control with its native counterpart on macOS does respond to key down and not key up?

I haven't done every exact counterpart to each FURN control I change, I more just did a general pass. I can actually go do that.

Saadnajmi:
> > Just to confirm here- I know NSMenus respond to key down events and would guess the others do too, but you've confirmed each control with its native counterpart on macOS does respond to key down and not key up?
> 
> I haven't done every exact counterpart to each FURN control I change, I more just did a general pass. I can actually go do that.

Went through all the equivalent macOS controls, and it is indeed keyDown

PR #1376, opened 2022-01-25
Publish Win32 FluentTester NuGet Package to Internal Office Feed

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In order to add test coverage for FURN in Devmain, we need to have the ability to pull the FluentTester bundle into Devmain. To achieve this goal, we need to publish the JS bundle to an internal Office feed as a NuGet package. Then, we can pull it in via Devmain.

Specifically, we are publishing it to this [React-Native Office Feed](https://dev.azure.com/ms/react-native/_packaging?_a=feed&feed=react-native). This is a private feed for Office (no public access), which is desired for us because this is for internal-testing purposes only.

I've tested this by making my own pipeline and validating the correct package is being pushed to the Office Feed (thanks @Saadnajmi )

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
```diff
@@ -0,0 +1,16 @@
...
+    <id>Microsoft.FluentUI.FluentTesterWin32</id>
+    <version>1.0.0</version>
+    <title>FluentUI-React-Native Win32 Bundle</title>
+    <authors>Microsoft Office CXE</authors>
+    <projectUrl>https://github.com/microsoft/fluentui-react-native.git</projectUrl>
+    <description>This package contains the React Native JS bundle of FluentUI-React-Native's Win32 Test App.</description>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+  </metadata>
+  <files>
+    <file src="..\dist\index.win32.bundle" target="win32" />
```
acoates-ms, line 13: Are the sourcemaps inline?  Or should you publish those too?

apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
```
acoates-ms, line 3: You should include this in the metadata:

```xml
  <metadata>
    ...
    <repository type="git"
      url="https://github.com/microsoft/fluentui-react-native.git"
      commit="$CommitId$" />
  </metadata>
```

Then add `-properties CommitId=$(Build.SourceVersion)` to the pack commandline.  Then it is easier to track where the nuget comes from.

apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
Saadnajmi: Also, is this not .jsbundle ?

acoates-ms: Only ios/macOS use `.jsbundle`.  I'm guessing because `.bundle` means something else on macOS.  All the other platforms use `.bundle` for their bundles.

acoates-ms (approved):

apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
Saadnajmi: Ah, makes sense! The existence of [NSBundle](https://developer.apple.com/documentation/foundation/nsbundle) explains that.

apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
samuelfreiberg: Changed in newest iteration - thanks for the advice

apps/win32/nuget/Microsoft.FluentUI.FluentTesterWin32.nuspec
samuelfreiberg: After bundling, I don't see any sourcemap files. I believe it's inline. Is there a way to double-check? How can I be sure? I've already tested out the published NuGet package with ReactTest and everything seems to work

PR #1377, opened 2022-01-25
Replace `ref` with `componentRef` on the Button to bring back `customize` and `compose`

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Turns out wrapping a component in `forwardRef` erases any exposing of APIs on that component. Whoops.
Reverting the change to add `ref` to bring back `customize` and `compose` functionality, Adding tests as well so that we get notified next time we break it.

Technically (as in like due to technical reasons) there isn't a way to have all of:
- A functional component (which prevents `ref` from being added as a prop to the component directly)
- Adding the `ref` prop through `forwardRef` (which will prevent us from exposing any functions on the component)
- Having callable functions on the component (which is what `compose` and `customize` are)
- Keeping the component returned a callable function (forwardRef returns a type derived from `ExoticNamedComponents` which explicitly does not allow for this behavior, but a callable component is needed for useSlots to work properly)

Following up with FluentUI on their new APIs since they don't have this issue, but in the meantime we go back to `componentRef` which allows us to support everything.

### Verification

image

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonHOCTestSection.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental/ButtonHOCTestSection.tsx
@@ -17,7 +17,7 @@ export const ButtonHOCTest: React.FunctionComponent = () => {
 
   return (
     <View style={[stackStyle, commonTestStyles.view]}>
-      <CustomButton style={commonTestStyles.vmargin} ref={buttonRef}>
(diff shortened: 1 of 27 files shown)


Saadnajmi (approved):
Sad, but thanks for the explanation!

chiuam (approved):

PR #1378, opened 2022-01-26
Button documentation edits

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding some more details to the Button documentation per feedback. Specifically I wrote some stuff about the customize and compose APIs and put it in the framework package close-ish to the code where it's defined. Happy to get any feedback about how I'm explaining those APIs.

I've tested out the code samples to ensure they work.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/SPEC.md
```diff
@@ -72,12 +76,18 @@ The `Button` component can be loading if it's waiting for another action to occu
 
 ### Slots
 
+The `Button` component has three slots, or parts. The slots behave as follows:
+
 - `root` - The outer container representing the `Button` itself that wraps everything passed via the `children` prop.
 - `icon` - If specified, renders an `icon` either before or after the `children` as specified by the `iconPosition` prop.
 - `loader` - If specified, renders a `loader` before `children` while the `loading` flag is set to `true` in place of an icon. This slot is mutually exclusive to the icon slot.
```
Saadnajmi, line 83: These don't match? In `Button.tsx` I see this?
```
  slots: {
    root: View,
    icon: Icon,
    content: Text,
  },
  ```
  
Glad we're documenting they exist!   In general though, I still feel Slots are hard to enforce as a public API... since we change the names a lot.

File: packages/experimental/Button/src/FAB/SPEC.md [hunk 2]
```diff
@@ -16,6 +16,8 @@ Basic examples:
 <FAB disabled>Text</FAB>
 ```
 
+More examples on the [Test pages for the Button](https://github.com/microsoft/fluentui-react-native/tree/master/apps/fluent-tester/src/FluentTester/TestComponents/ButtonExperimental). Instructions on running the tester app can be found [here](https://github.com/microsoft/fluentui-react-native/blob/master/apps/fluent-tester/README.md).
+
```
Saadnajmi, line 20: At the top of this page, can we expand somewhere that FAB === Floating Action Button ?

packages/framework/composition/README.md
```diff
@@ -1 +1,47 @@
-# @fluentui-react-native/use-styling
+# @fluentui-react-native/composition
```
Saadnajmi, line 1: Did the actual package name change or is this just a documentation update?

packages/framework/composition/README.md
```diff
@@ -1 +1,47 @@
...
+
+This package contains the `composeFactory` which all components created using the `compose` function under `@fluentui-react-native/framework` use to construct the component itself. The `composeFactory` returns a [staged component](https://github.com/microsoft/fluentui-react-native/tree/master/packages/framework/use-slot) with two additional functions, `customize` and `compose`. These functions allow for these components to have a degree of extensibility.
```
Saadnajmi, line 3: For the first words that people read in relation to compose, I feel there'a bit too much inner technical deatil (`composeFactory` / `stagedComponent`). I think it might be better if the first few sentences were a rougher overview. Something like "Compose is a way to make Higher Order Components that expose some methods (customize  / compose) to further style your components". Not that exactly.. but maybe save talking about `composeFactory` / `stagedComponent` for a "Behind the Scenes" section.

Happy to take alternative views though, and feel free to shoot the suggestion down!

packages/experimental/Button/SPEC.md
```diff
@@ -4,6 +4,10 @@
 
 The `Button` component enables users to trigger an action or event, such as submitting a form, opening a dialog, canceling an action, or performing a delete operation.
 
+## Requirements
+
+If using FURN's theming, the `Button` requires use of the `ThemeProvider` from `@fluentui-react-native/theme` to work properly with themes. Please see [this page](https://github.com/microsoft/fluentui-react-native/blob/master/docs/pages/Guides/UpdateThemeProvider.md) for information on updating your `ThemeProvider` if using the version from `@uifabricshared/theming-react-native`.
```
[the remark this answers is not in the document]
rurikoaraki: I figure since we'll probably keep the template the same between components, it wouldn't be too bad to remember to put this on top of every control. Since our controls will be a mixed bag until we get every control ported, I thought that having the prerequisite close to each control rather than in our docs folder somewhere would make it more visible.

packages/experimental/Button/SPEC.md
rurikoaraki: Thanks for catching, will change.

packages/framework/composition/README.md
rurikoaraki: I think this was a copy paste error, the actual name of the package is composition. use-styling is a different package.

packages/framework/composition/README.md
rurikoaraki: Good point, I can reword the first parts. Thanks for the feedback!

packages/experimental/Button/SPEC.md
Saadnajmi: My worry here is that we don't really have any way to enforce people to _not_ change the slots. We pretty consistently add/modify slots anytime we make a change to a component. An example: Adding a chevron to `SubmenuItem` needed new slots for a `leftStack` and `rightStack`, which I eventually renamed to `startStack` and `endStack`, which could easily have been refactored now since I haven't checked it recently. Really, any change to a component means you have to change _something_ in the slot props / definition, which would affect anyone using `customize`.

That monologue is irrelevant to the PR though. I just realized I never typed that out anywhere.  So I'll just tag @ejlayne to note the expressed thought, and leave it aside.

packages/framework/composition/README.md
rurikoaraki: Tried rewording, let me know if this is better

packages/experimental/Button/src/FAB/SPEC.md
rurikoaraki: Done

Saadnajmi (approved):

rurikoaraki:
Note to self: Add notes about where you can use customize/compose and also notes about how to use theme as an arg for customize or buildProps for compose

packages/experimental/Button/SPEC.md
```diff
@@ -72,12 +76,18 @@ The `Button` component can be loading if it's waiting for another action to occu
 
 ### Slots
 
+The `Button` component has three slots, or parts. The slots behave as follows:
+
 - `root` - The outer container representing the `Button` itself that wraps everything passed via the `children` prop.
 - `icon` - If specified, renders an `icon` either before or after the `children` as specified by the `iconPosition` prop.
-- `loader` - If specified, renders a `loader` before `children` while the `loading` flag is set to `true` in place of an icon. This slot is mutually exclusive to the icon slot.
+- `content` - If specified, renders the first entry of `children` as text.
```
ValentinaKozlova, line 83: Will we have a content prop for Button?

ValentinaKozlova (approved):

PR #1381, opened 2022-01-26
Add More Functional E2E Testing Win32/Windows

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

This PR adds Functional E2E Testing for Tabs, ContextualMenu, MenuButton, and RadioGroup. Specifically, clicking + keyboarding.

**Tabs**
- Click on a TabItem and validate the content changes correctly
- Keyboard navigate through the TabItem's and validate the content changes correctly

**Contextual Menu**
- Click and validate the ContextualMenu opens
- Type "Spacebar" to open and validate

**MenuButton**
- Click and validate the MenuButton opens
- Type "Spacebar" to open and validate

**RadioGroup**
- Click on RadioButtons and validate the Selection state changes
- Keyboard navigate through RadioButtons and validate selection state changes

**Additional Changes**
- Adding ComponentSelector enums to each page object. As testing gets more complicated (more UI elements to interact with), we need to provide an easy way for spec files to choose which UI elements they want to perform tasks on. This enum will help with that. Each entry in the enum will map to a specific UI element (ie. FirstRadioButton, SecondRadioButton, etc.)

### Verification

Manual + Automated

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/E2E/ContextualMenu/pages/ContextualMenuPageObject.win.ts
```diff
@@ -2,10 +2,55 @@ import {
...
   CONTEXTUALMENU_TEST_COMPONENT,
   HOMEPAGE_CONTEXTUALMENU_BUTTON,
+  CONTEXTUALMENUITEM_TEST_COMPONENT,
 } from '../../../FluentTester/TestComponents/ContextualMenu/consts';
 import { BasePage, By } from '../../common/BasePage.win';
 
+/* This enum gives the spec file an EASY way to interact with SPECIFIC UI elements on the page.
+ * The spec file should import this enum and use it when wanting to interact with different elements on the page. */
+export const enum ContextualMenuSelector {
+  ContextualMenu = 0, //this._primaryComponent
```
rurikoaraki, line 12: nit: the = 0 seems unnecessary, is there a particular reason for it?

rurikoaraki:
Thanks for adding all of these! This is awesome to see!

apps/fluent-tester/src/E2E/RadioGroup/pages/RadioGroupPageObject.win.ts
```diff
@@ -58,12 +104,22 @@ class RadioGroupPage extends BasePage {
   /***************/
   /* RadioButton *
   /**************/
-  get _radioButton() {
-    return By(RADIOBUTTON_TEST_COMPONENT);
+
+  /* The first RadioGroup has 4 radio buttons, all listed below */
+  get _firstRadioButton() {
```
rurikoaraki, line 109: I think the test pages would be more flexible if you could have a 

```
get _radioButtonById(id: string) {
    return By(id);
}
```

sort of thing instead. Then you wouldn't need to add a get function for each component you want to track.

It also seems to me that we could use the testIds to refer to components instead of the Selectors, is there a reason to use the Selectors over the testId strings? I can think of type safety being a reason, but otherwise it seems like it's making the process of referring to elements more complex.

apps/fluent-tester/src/E2E/ContextualMenu/pages/ContextualMenuPageObject.win.ts
samuelfreiberg: It's just a visual representation of the numeric value of each enum entry. ContextualMenu would be initialized with the value 0 regardless, but for people who aren't too familiar with JS I thought it might be helpful to have.

You can see [more here  ](https://www.typescriptlang.org/docs/handbook/enums.html). I was always taught this method so I've carried it with me :)

apps/fluent-tester/src/E2E/RadioGroup/pages/RadioGroupPageObject.win.ts
samuelfreiberg: I definitely see where you're coming from. The answer is: We don't want the tests themselves to be using the Test IDs. We're using the [Page Object Pattern](https://webdriver.io/docs/pageobjects) (recommended by WebDriverIO).

The goal of using Page Objects is to abstract any page information away from the actual tests. Ideally, you should store all selectors or specific instructions that are unique for a certain page in a page object, so that you still can run your test after you've completely redesigned your page.

I do see what you're saying about making it easier to track, though. That's why I added `getRadioButton(radioButtonSelector: RadioButtonSelector)` that makes it easier for the spec to communicate with the Page Object.

Another reason why we wouldn't necessarily want something like  
```
get _radioButtonById(id: string) {
    return By(id);
}
```
is because this will query the driver every single time we want to interact with the app, rather than just one time when the Page Object class gets initialized. It would effect the perf and opens the door for more intermittent failures (the more unnecessary times we query the driver, the higher chance a failure might happen.

But I do see what you're saying and I'd be open to having a discussion about pros/cons for each method.


Additionally, for using the TestIDs to refer to components instead of selectors. I'm a bit confused by the question, but I'll try to explain the functionality of both... Let me know if I missed the question.

The "Selectors" are the methods of querying the app driver for specific UI components. The selector method needs some type of unique id to query these elements.. Hence the TestID. We need the "Selectors" to query into the app to return the specific UI element. We need the TestID to specify which UI element we want.

rurikoaraki: I see, didn't realize that this was a suggested pattern and that there are perf reasons behind this. Thanks for the explanation!

rurikoaraki (approved):

PR #1382, opened 2022-01-26
Callout macOS: Fix focus flickering on click

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On macOS, clicking on a view inside a Callout would register the click in seemingly the wrong location, causing keyboard focus / highlight effects to flicker. Clicking on the 1st Menu item inside a ContextualMenu would sometimes move focus (and highlight) to the last (5th) menu item, while still showing the pressed effect on the 1st menu item. So... it would sometimes look like 2 menu items are highlighted when you click. Similarly, clicking the 2nd menu item would highlight the N-1'th (4th) menu item and so on. This was a very weird bug that stumped me for a long time.

It seems this is because macOS by default uses a flipped coordinate system ( (0,0) is the bottom left of the screen instead of the top left), and subclasses of NSView must override [NSView.flipped](https://developer.apple.com/documentation/appkit/nsview/1483532-flipped) to get the more standard origin-in-top-left coordinate space. React Native macOS does this for RCTView/RCTUIView/all of it's native views. 

Inside Callout, I had one view (NSVisualEffectView) that was not flipped. Once I subclassed it into a private `FlippedVisualEffectView`, I seem to no longer reproduce the weird focus flicker.

### Verification

Took a video of using Callout / ContextualMenu after my changes, and things behave as expected. Because this issue is intermittent, I would appreciate if someone else with macOS can test my change on their machine.

https://user-images.githubusercontent.com/6722175/151259224-60d76589-babd-4732-8762-6463729a1c43.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/Callout.swift
+++ b/packages/components/Callout/macos/Callout.swift
@@ -58,17 +58,6 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 	// MARK: RCTComponent Overrides
 
 	override func insertReactSubview(_ subview: NSView!, at atIndex: Int) {
-		/*
-		  If we attach a touch handler to the proxy view directly, the touch coordinates
-		  are incorrect and the wrong subview recieves the touch event, most likely
-		  because macOS screen coordinates have a flipped Y axis. The workaround is to
-		  attach a separate touch handler to each subview
-		 */
-		guard let touchHandler = RCTTouchHandler(bridge: bridge) else {
-			preconditionFailure("Callout could not create RCTTouchHandler")
-		}
-		touchHandler.attach(to: subview as? RCTUIView)
-
 		proxyView.insertReactSubview(subview, at: atIndex)
 	}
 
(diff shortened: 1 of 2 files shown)


packages/components/Callout/macos/Callout.swift
```diff
@@ -281,37 +270,53 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
-		visualEffect.material = .menu
-		visualEffect.state = .active
-		visualEffect.wantsLayer = true
-		visualEffect.layer?.cornerRadius = calloutWindowCornerRadius
-
 		guard let contentView = window.contentView else {
 			preconditionFailure("Callout window has no content view")
 		}
 
-		contentView.addSubview(visualEffect)
```
harrieshin, line 303: dumb question : why is it ok to remove the constraints on proxyview?

packages/components/Callout/macos/Callout.swift
```diff
@@ -281,37 +270,53 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 
 	// MARK: Private variables
 
+	/// The view the Callout is presented from, if anchorRect is nil.
 	private var anchorView: NSView?
 
-	private var proxyView = RCTView()
+	/// The  view we forward Callout's Children to. It's hosted within the CalloutWindow's
+	/// view heirarchy, ensuring our React Views are not placed in the main window.
+	private lazy var proxyView: NSView = {
```
harrieshin, line 278: since this ain't override var, why can't it be FlippedvisualEffectView for clarity? does it have to stay nsview?

packages/components/Callout/macos/Callout.swift
Saadnajmi: Before the view hierarchy looked like NSWindow->ContentView->VisualEffectView->proxyView.
Turns out the visualEffectView could just be the proxyView, so I didn't need the extra `RCTView`. And also, we no longer need those constraints.. somehow.

packages/components/Callout/macos/Callout.swift
Saadnajmi: It could be `FlippedVisualEffectView`. I think I just went for the more generic type since `FlippedVisualEffectView` is more of an implementation detail. There could be a future we break out FlippedVisualEffectView as it's own Native module, and we want some other type of view as the proxyView here.

packages/components/Callout/macos/Callout.swift
harrieshin: you are setting visualEffectView.translatesAutoresizingMaskIntoConstraints = false .... so who is setting the constraint? have you look at ui debugger to see if there are any ambiguous size?

Saadnajmi: There seems to not be. It looks like it inherited the size just fine. In general, there is no set size in the whole view hierarchy, so I think it's React Native is doing it's own layout pass where it is setting the frame on the Callout, which get's forwarded to the the proxyView frame (visualEffectView)

image

Saadnajmi: I thought about it more and have an answer. Now that `FlippedVisualEffectView` _is_ the proxy view, it's frame is set by React Native on line 69. React Native uses Yoga / Shadow Views to do Layout, so that's why we don't need constraints.

PR #1383, opened 2022-01-27
Migrate `Button` V2 to `component` folder

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes=

This PR moves the experimental button into the component button, as we are close to finishing the Button for V2. This PR will NOT be committed until messaging and the component has had final sign off from @ejlayne.

NOTE: It may be easier to review this commit-by-commit than all at once.

1. Old Button component is moved under a deprecated folder.
2. All experimental button files have been moved to the button package. The experimental button has been (temporarily) exported as `ButtonV2` to avoid name clashing with the already existing `Button`.
3. The experimental-button package now just reexports what it used to have from the button package. This is to minimize dropping changes we make to the V2 Button while migrating, and not have duplicate code.
4. Added deprecation messages to the experimental package and the old Button component:
    - Experimental package will have the message "The @fluentui-react-native/exprimental-button package is deprecated. The contents of this package have been moved to @fluentui-react-native/button. If you need to use the Button component from this package, please use ButtonV2 from @fluentui-react-native/button."
    - The old Button components will show in intellisense "This component is deprecated in favor of ButtonV2. This Button will be removed when the package moves to 1.0.0. At that point, ButtonV2 will be renamed to Button. Please see MIGRATION.md for details on how to move to the new Button."
5. I removed default exports for each control as I was worried that they would clash. This does not affect what gets exported from the package.
6. Merged the Button test pages for both buttons to one page. The old Button tests are now at the bottom of the Button test page. E2E tests have not been merged as I can't specify a test to look for more than two components.

### Verification

Verified that deprecation message only shows on debug, not release bundle
Verified that Button still loads out of experimental button. One test still pulls the button from the experimental-button package to ensure that it hasn't been broken.
Verified that deprecation intellisense shows up.
Verified that old Button is not affected via test page.
Verified E2E tests run on both versions of the button.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/package.json
```diff
@@ -28,19 +28,28 @@
...
     "@fluentui-react-native/pressable": ">=0.8.17 <1.0.0",
+    "@fluentui-react-native/styling-utils": ">=0.3.4 <1.0.0",
+    "@fluentui-react-native/theme-tokens": ">=0.11.6 <1.0.0",
     "@fluentui-react-native/text": ">=0.11.8 <1.0.0",
     "@fluentui-react-native/tokens": ">=0.11.5 <1.0.0",
+    "@fluentui-react-native/use-styling": ">=0.8.2 <1.0.0",
     "@uifabricshared/foundation-composable": ">=0.10.2 <1.0.0",
-    "@uifabricshared/foundation-settings": ">=0.11.2 <1.0.0"
+    "@uifabricshared/foundation-settings": ">=0.11.2 <1.0.0",
+    "tslib": "^2.3.1"
```
Saadnajmi, line 44: Why tslib?

Saadnajmi (approved):

packages/components/Button/package.json
rurikoaraki: It was part of the experimental-button package. I think we ran into some sort of issue before where something we were doing (I think spread?) depended on this version of tslib

ValentinaKozlova (approved):

PR #1386, opened 2022-01-29
Tester customize cleanup

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

A bunch of our tester examples misuse the customize() API by calling them inside a test component's render function. Doing this will result in the customized component being recreated on every render, causing wasted renders of the customized component. Moving the components out of the render functions or memoizing them in order to save on renders.

~Through this I discovered that a bunch of our Avatar customization tests have stopped working, so, I guess I'll fix those too? Those will come at a later commit and it's why this is still a draft~

Probably not worth blocking this on that, so will just publish this as is

### Verification

Verified that every test page is the same as before the changes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Thanks for doing this!
Could we add documentation somewhere to not call `customize` inside a render?

rurikoaraki:
> Thanks for doing this! Could we add documentation somewhere to not call `customize` inside a render?

Yeah, a note about that has been added to the customize documentation that I just checked in.

Saadnajmi (approved):

PR #1388, opened 2022-02-01
Fix toggle button text color

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

I noticed that the ToggleButton in checked state didn't have legible text, fixing the text match redlines.

This PR also fixes HC for UWP, which I only just realized causes crashes.

### Verification

Tested win32 and UWP in non-HC/HC scenarios

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-experimental-button-ade8b461-9bc0-493a-8e99-aea4ecdd5536.json
+++ b/change/@fluentui-react-native-experimental-button-ade8b461-9bc0-493a-8e99-aea4ecdd5536.json
@@ -1,0 +1,10 @@
+{
+  "type": "patch",
+  "comment": {
+    "title": "",
+    "value": ""
+  },
+  "packageName": "@fluentui-react-native/experimental-button",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@fluentui-react-native-theme-tokens-aeef413c-0cbf-4c4f-8679-50a00723ef38.json
+++ b/change/@fluentui-react-native-theme-tokens-aeef413c-0cbf-4c4f-8679-50a00723ef38.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Fix UWP HC",
+  "packageName": "@fluentui-react-native/theme-tokens",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/experimental/Button/src/ToggleButton/ToggleButtonColorTokens.win32.ts
+++ b/packages/experimental/Button/src/ToggleButton/ToggleButtonColorTokens.win32.ts
@@ -8,9 +8,9 @@ export const defaultToggleButtonColorTokens: TokenSettings<ToggleButtonTokens, T
     backgroundColor: t.colors.neutralBackground1Selected,
     borderColor: t.colors.neutralStroke1,
     primary: {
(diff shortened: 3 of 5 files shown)


Saadnajmi:
The last button in the last screenshot looks a little weird? Like the text is selected black?

packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
```diff
@@ -9,7 +9,7 @@ function processAliasTokens(aliasTokens: any) {
     for (const innerKey in aliasTokens[key]) {
       const entry = aliasTokens[key][innerKey];
       if (typeof entry === 'string' && entry.includes('PlatformColor')) {
-        const color = entry.substring(14, entry.length - 1);
+        const color = 'SystemColor' + entry.substring(14, entry.length - 1) + 'Color';
```
Saadnajmi, line 12: Curious, what does system color mean here?

samuelfreiberg (approved):

samuelfreiberg (commented):
Seconding what Saad mentioned

rurikoaraki:
> The last button in the last screenshot looks a little weird? Like the text is selected black?

@Saadnajmi @samuelfreiberg I'll look into it and see if setting the backgroundColor on it works, but I vaguely remember this being an issue when I worked on UWP.

XAML will automatically style parts of the UI in HC to have the appropriate color if no explicit color is provided. Normally that it good. However, because we build the button using "parts" like `Text`, `View`, etc. instead of using a built-in `Button` component, the coloring comes out that the `Text` component gets a black background because, well, that's the Window color that was defined for text in HC.

packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
rurikoaraki: It denotes that it's a brush "reserved"/built-in by the OS, as opposed to brushes that are user-defined.

The brush names for HC are documented here: https://docs.microsoft.com/en-us/windows/apps/design/accessibility/high-contrast-themes#contrast-colors

Saadnajmi: AH, gotcha! macOS has a bunch of similar colors https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors

Should the day come tokens have macOS PlatformColor's in them, I''' remember this is a place we can decode it!

rurikoaraki:
> > The last button in the last screenshot looks a little weird? Like the text is selected black?
> 
> @Saadnajmi @samuelfreiberg I'll look into it and see if setting the backgroundColor on it works, but I vaguely remember this being an issue when I worked on UWP.
> 
> XAML will automatically style parts of the UI in HC to have the appropriate color if no explicit color is provided. Normally that it good. However, because we build the button using "parts" like `Text`, `View`, etc. instead of using a built-in `Button` component, the coloring comes out that the `Text` component gets a black background because, well, that's the Window color that was defined for text in (this theme of) HC.

Tried this out and it seems like something with the Slots architecture is making it so that the text is not visible when the backgroundColor is set on Text when in HC specifically (FURN or RN). No idea what's going on with that, but I think that's something that will need to be looked into separately.

Saadnajmi (approved):

PR #1391, opened 2022-02-01
High contrast support for Button

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds high contrast support for the default button, specifically to increase contrast between button border and button background.

### Verification

https://user-images.githubusercontent.com/67026167/151919232-26af8b50-cccf-4a1e-8fef-548421a9e9de.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/apple-theme/src/createAppleTheme.macos.ts
```diff
@@ -2,15 +2,31 @@ import { ThemeReference } from '@fluentui-react-native/theme';
...
     appleThemeReference.invalidate();
   });
-
   return appleThemeReference;
 }
+
+function highContrastHandler() {
+  AccessibilityInfo.isHighContrastEnabled().then((isEnabled) => {
+    appleThemeReference.invalidate();
+    isHighContrastMode = isEnabled;
```
rurikoaraki, line 30: nit: I would set this before running invalidate, as the invalidate step depends on isHighContrastMode.

packages/theming/theme-tokens/src/getAliasTokens.macos.ts
```diff
@@ -1,14 +1,25 @@
 import * as macOSLightAliasTokens from './light-macos/tokens-alias';
 import * as macOSDarkAliasTokens from './dark-macos/tokens-alias';
+import * as macOSLightHCAliasTokens from './light-high-contrast-macos/tokens-alias';
+import * as macOSDarkHCAliasTokens from './dark-high-contrast-macos/tokens-alias';
 import * as hcAliasTokens from './highContrast/tokens-alias';
 import { AppearanceOptions } from '@fluentui-react-native/theme-types';
 import { assertNever } from 'assert-never';
+import { isHighContrastMode } from '../../apple-theme/src/createAppleTheme.macos';
```
rurikoaraki, line 8: I would prefer this get passed in as a variable than have the theme-tokens package import something from the apple-theme package. That way the theme-tokens package doesn't explicitly depend on any platform theme.

rurikoaraki (changes requested):
Please see comment about not importing from apple-theme

rurikoaraki:
Was this PR meant to be separate from the other tokens PR? I don't think this change will build on its own, is the plan to not commit this until the other PR is committed?

packages/experimental/Button/src/ButtonColorTokens.ts
```diff
@@ -17,13 +17,13 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
...
       color: t.colors.defaultHoveredContent,
-      borderColor: t.colors.defaultHoveredBorder,
+      borderColor: t.colors.buttonBorder,
       iconColor: t.colors.defaultHoveredIcon,
     },
     pressed: {
       backgroundColor: t.colors.defaultPressedBackground,
       color: t.colors.defaultPressedContent,
-      borderColor: t.colors.defaultPressedBorder,
+      borderColor: t.colors.buttonBorder,
```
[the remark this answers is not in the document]
chiuam: Hmm, let me take a look at those tokens.

packages/theming/apple-theme/src/createMacOSAliasTokens.ts
```diff
@@ -1,11 +1,10 @@
...
-import { memoize } from '@fluentui-react-native/memo-cache';
 import { mapPipelineToTheme } from '@fluentui-react-native/theming-utils';
 
 function createMacOSAliasTokensWorker(mode: AppearanceOptions): AliasColorTokens {
   const aliasTokens = getAliasTokens(mode);
   return mapPipelineToTheme(aliasTokens);
 }
 
-export const createMacOSAliasTokens = memoize(createMacOSAliasTokensWorker);
+export const createMacOSAliasTokens = createMacOSAliasTokensWorker;
```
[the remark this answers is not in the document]
chiuam: From my understanding of `memoize()`, it caches and returns the cached set of tokens if no changes are reflected. Since `Appearance` doesn't detect HC on macOS, the same set of tokens are returned when HC setting changes. So by removing memoize, we call getAliasTokens() on every theme change to handle this edge case.  In other words, without it getAliasTokens won't fire when HC changes

chiuam:
> Was this PR meant to be separate from the other tokens PR? I don't think this change will build on its own, is the plan to not commit this until the other PR is committed?

Yep, this PR shouldn't be checked in until #1390 is merged.

packages/theming/theme-tokens/src/getAliasTokens.macos.ts
chiuam: Yeah and I noticed that this is breaking the build too :/

packages/theming/apple-theme/src/createMacOSAliasTokens.ts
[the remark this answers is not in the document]
Saadnajmi: Would it be possible to add `isHighContrast` or the like as an argument to `createMacOSAliasTokensWorker` so that it's properly reflected in the cache? I worry about the performance cost of recreating all the tokens / re-rendering everything by removing the `memoize()`.

I also wonder if for macOS, we can include checking `isHighContrast` into `AppearanceOptions`, similar to what's being done in [getCurrentAppearance.windows.ts](https://github.com/microsoft/fluentui-react-native/blob/master/packages/theming/theming-utils/src/getCurrentAppearance.windows.ts). It looks like there's effort on that side to inlude HC as part of the react-native-windows `Appearance` API (https://github.com/microsoft/react-native-windows/issues/3701)

packages/theming/apple-theme/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './createAppleTheme';
 export * from './createMacOSAliasTokens';
+export { isHighContrastModeEnabled } from './createAppleTheme.macos';
```
rurikoaraki, line 3: Is this still being used outside the package?

chiuam: I think so, it gets passed down to `getMacOSAliasTokens` that lives in the theme-tokens package

rurikoaraki: That's the value of the variable though, the variable itself doesn't get imported into `getMacOsAliasTokens.ts`

packages/theming/apple-theme/src/appleHighContrast.macos.ts
```diff
@@ -0,0 +1,5 @@
+export let isHighContrast = false;
+
+export function setIsHighContrast(isHighContrast: boolean) {
```
rurikoaraki, line 3: nit: I think if we have a set function we should also make a get function. Also makes the code more encapsulated.

packages/theming/apple-theme/src/appleColors.macos.ts
```diff
@@ -58,7 +59,7 @@ export function getAppleSemanticPalette(): AppleSemanticPalette {
 function getFluentUIApplePalette(): FluentUIApplePalette {
   const appearance = Appearance.getColorScheme();
   const mode = getCurrentAppearance(appearance, 'light');
-  const macOSAliasColorTokens = createMacOSAliasTokens(mode);
+  const macOSAliasColorTokens = createMacOSAliasTokens(mode, isHighContrastEnabled);
```
rurikoaraki, line 62: Now that we have a getter function we can use it here, and not export isHighContrastEnabled directly.

rurikoaraki (approved):

packages/theming/apple-theme/src/createMacOSAliasTokens.ts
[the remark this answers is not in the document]
chiuam: Thanks for the suggestions, I added `memoize()` back in commit 510a5e7ca6193385b05f99f0e6e65a3503dc08a6.

Adding high contrast options to `AppearanceOptions` for macOS may be a bit tricky since it has light and dark HC modes... not sure if we want to incorporate platform specific stuff to cross-plat objects.

packages/theming/apple-theme/src/index.ts
chiuam: Talked offline, extracting the variable from createAppleTheme and implementing a setter for it did the trick. Thanks!!

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (isHighContrast) {
+      return macOSDarkHCAliasTokens.default;
+    } else {
+      return macOSDarkAliasTokens.default;
+    }
+  } else if (mode === 'highContrast') {
+    return hcAliasTokens.default;
+  } else {
+    assertNever(mode);
+  }
```
Saadnajmi, line 26: If we never want to hit the case `(mode === 'highContrast')`, then we could call `assertNever(mode)` for that case instead.

Nit: I wonder if we can consolidate the AppearanceOption mode highContrast / getCurrentAppearance with our new isHighContrast boolean...

packages/theming/apple-theme/src/createAppleTheme.macos.ts
```diff
@@ -2,15 +2,31 @@ import { ThemeReference } from '@fluentui-react-native/theme';
...
-
   return appleThemeReference;
 }
+
+function highContrastHandler() {
+  AccessibilityInfo.isHighContrastEnabled().then((isEnabled) => {
+    setIsHighContrast(isEnabled);
+    appleThemeReference.invalidate();
+  });
+}
```
Saadnajmi, line 32: I wonder if we could encapsulate this logic ( + the added logic in createAppleTheme) into a hook like `useHighContrastEnabled()` or something, similar to this hook I found: https://github.com/react-native-hooks/accessibility-info/blob/master/src/index.js

packages/theming/apple-theme/src/appleHighContrast.macos.ts
```diff
@@ -0,0 +1,9 @@
+export let isHighContrastEnabled = false;
+
+export function setIsHighContrast(isHighContrast: boolean) {
+  isHighContrastEnabled = isHighContrast;
+}
+
+export function getIsHighContrast() {
+  return isHighContrastEnabled;
+}
```
Saadnajmi, line 9: For my own knowledge, why did you have to separate this out to it's own file? It seems like it was to resolve a circular dependency?

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
chiuam: I think `assertNever` is meant to check for exhaustiveness. The appropriate check for catching that error should be `throw new Error`

packages/theming/apple-theme/src/createAppleTheme.macos.ts
chiuam: Not an expert on hooks but that seems like a cleaner approach. Though, aren't hooks meant to be reusable?

rurikoaraki: This is also not in a component, which is what hooks are for.

Saadnajmi: Ah yeah, the hook I linked relies on `useEffect` so it probably would not work here. Whelp.

packages/experimental/Button/src/ButtonColorTokens.ts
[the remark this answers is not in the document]
chiuam: Good point, I'm reverting these changes and updating the default border tokens to use transparentStroke.

Aside, it may be a good idea to have a separate Button colors tokens file for macOS.

packages/theming/apple-theme/src/appleHighContrast.macos.ts
chiuam: Yeah, I had to extract the dependency part so that the caller doesn't query from the same file.

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
rurikoaraki: Yeah `assertNever` is there to check for case exhaustiveness. I don't think it would be bad to add code to the highContrast case to throw an error or something, since MacOS doesn't expect to hit it.

packages/experimental/Button/src/ButtonColorTokens.ts
```diff
@@ -23,7 +23,7 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
     pressed: {
       backgroundColor: t.colors.defaultPressedBackground,
       color: t.colors.defaultPressedContent,
-      borderColor: t.colors.defaultPressedBorder,
+      borderColor: t.colors.defaultHoveredBorder,
```
rurikoaraki, line 26: Is this intended to change to Hovered for the `pressed` state?

chiuam: Oh no, thanks for catching :)

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
```diff
@@ -0,0 +1,26 @@
+import * as macOSLightAliasTokens from './light-macos/tokens-alias';
+import * as macOSDarkAliasTokens from './dark-macos/tokens-alias';
+import * as macOSLightHCAliasTokens from './light-high-contrast-macos/tokens-alias';
+import * as macOSDarkHCAliasTokens from './dark-high-contrast-macos/tokens-alias';
```
rurikoaraki, line 4: small nit: we could do `import macOSDarkHCAliasTokens  from './dark-high-contrast-macos/tokens-alias'` here instead of importing everything as macOSDarkHCAliasTokens. Then when you access macOSDarkHCAliasTokens you don't need the ".default" part.

rurikoaraki (approved):

Saadnajmi:
Oh yeah, how does it look in dark mode + high contrast?

chiuam:
> Oh yeah, how does it look in dark mode + high contrast?

No changes from the video recording attached in the PR description, any further concerns?

packages/theming/theme-tokens/src/getMacOSAliasTokens.ts
```diff
@@ -0,0 +1,26 @@
...
+      return macOSLightAliasTokens;
+    }
+  } else if (mode === 'dark') {
+    if (isHighContrast) {
+      return macOSDarkHCAliasTokens;
+    } else {
+      return macOSDarkAliasTokens;
+    }
+  } else if (mode === 'highContrast') {
+    throw new Error('highContrast is not a valid AppearanceOptions on macOS');
```
Saadnajmi, line 22: nit: To clarify that we _do_ indeed support high contrast, maybe we can say something like "High Contrast must be enabled in system preferences to get High contrast styling"

Mostly so that the first time someone gets this error message, they don't report a bug "macOS doesn't support high contrast" incorrectly.

Saadnajmi (approved):

PR #1400, opened 2022-02-03
Keyboarding improvements for ContextualMenu macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This changes fixes several bugs on macOS FocusZone and ContextualMenu that vastly improve keyboarding accessibility. One can now fully use ContextualMenu on macOS with just a keyboard (and not a mouse).

#### FocusZone (macOS) changes:

**1. Renames `defaultTabbableElement` macOS native property , to `defaultKeyView`.** 

The public JS API is still `defaultTabbableElement`, but the native module now sets a property with the name `defualtkeyView`. The name feels much more idiomatic to me for macOS. The public facing API of FocusZone does not change, so this shouldn't affect any downstream clients. 

**2. Have FocusZone focus on the first key view within itself, rather than it's `nextValidKeyView`.**

Previously, FocusZone would focus on its' `nextValidKeyView` if no `defaultTabbableElement` was set. I believe the intent was "focus on the first key view _inside_ the FocusZone". However, I believe this would actually focus on the first key view _outside_ the FocusZone. In the case of ContextualMenu (where there was no key view after the FocusZone), it meant focus went nowhere. Instead, I wrote a helper method `GetFirstKeyViewWithin(NSView *parentView)` that will recursively search for the first key view within a view's subviews, and set that as the "firstKeyView". 

**3. We now call `[super keyDown:Event]` if a FocusZone does not handle a keyboard event, instead of doing nothing.**

If the user pressed `Arrow(Left|Right)` in a vertical FocusZone or `Arrow(Up|Down)` in a horizontal FocusZone, then focusZone would ignore the key event. This is incorrect, it should call `[super keyDown:event]` in those cases so that it's superview / base class may be able to handle the keyboard event. This prevented us from letting `ArrowLeft` close a submenu. 

#### ContextualMenu  changes:

**1. Make scrollView and focusZone into slots**

Before these were created inside the staged render. This is probably bad because they would be recreated every render. FocusZone is an optional macOS only slot. 

**2. (macOS Only) Focus on the ContextualMenu's inner FocusZone after render**

Unlike Win32, Callouts / NSWindows on macOS do not automatically place focus on the first focuable item once presented. Instead, you have to natively set an NSWindows `initialFirstResponder` property to do that. 

So now, I _could_ add a new property to Callout called `initialFocusView`, have it internally map to `[NSWindow initialFirstResponder]`, and then have ContextualMenu set it's `initialFocusView` to it's inner FocusZone...
... or I could just use the hook `useLayoutEffect()` to imperatively focus on the inner focus zone after render. This solution was much less code and seemed simpler. I also needed a `setTimeout(..., 0)` in there for some reason 🤷🏾. I'll probably go through the path of _properly_ adding a new prop `initialFocusView` to Callout after this change goes in, since that seems  the more correct solution. 

**3. (macOS Only) ArrowLeft will close a submenu**

This lets you navigate in and out of submenus with just arrow keys. Note this won't work on win32 until`...useKeyProps` also overrides `key(Up|Down)` events for win32.

### Verification

In FluentTester, one can now use ContextualMenu fully with the keyboard. Focus is properly placed on the menu when it opens, and you can arrow key in and out of submenus.

https://user-images.githubusercontent.com/6722175/152287608-c6b3493a-3d89-46f7-9887-b3e4ec520ed3.mov

I verified that RadioGroup (the other control that uses FocusZone) still behaves as expected.
I still need to verify that my FocusZone changes don't affect downstream clients. Will do so once my local build finishes...

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/ContextualMenu/src/Submenu.tsx
```diff
@@ -70,7 +70,9 @@ export const Submenu = compose<SubmenuType>({
         setInitialFocus: shouldFocusOnMount,
       },
       container: Platform.select({
-        macos: {},
+        macos: {
+          ...useKeyDownProps(onDismiss, 'ArrowLeft'),
```
Saadnajmi, line 74: Note: This should flip in RTL. I'm thinking I'll do that in a separate change.

Saadnajmi: @ppatboyd Would this be a good to have on Win32? Having `ArrowLeft` close a submenu?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,19 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
...
+	for (NSView *view in [parentView subviews]) {
+		if ([view canBecomeKeyView]) {
+			return view;
+		}
+		
+		NSView *match = GetFirstKeyViewWithin(view);
+		if (match) return match;
+	}
+	return nil;
+}
```
Saadnajmi, line 59: Inspired by https://github.com/nicklockwood/ViewUtils/blob/master/ViewUtils/ViewUtils.m#L249

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,19 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
 	);
 }
 
+static NSView *GetFirstKeyViewWithin(NSView *parentView)
```
HeyImChris, line 48: Would it make sense to have a GetLastKeyViewWithin to get the last menu button for when a user launches a menu and presses ArrowUp?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,19 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
...
+{
+	for (NSView *view in [parentView subviews]) {
+		if ([view canBecomeKeyView]) {
+			return view;
+		}
+		
+		NSView *match = GetFirstKeyViewWithin(view);
+		if (match) return match;
```
HeyImChris, line 56: I think there's some Obj-c guidelines around this somewhere but we should use curly braces even for 1 line conditional actions. I think the main benefit is debugging and handling merge conflicts - we've been burnt before with merges that hit this

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,19 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
 	);
 }
 
+static NSView *GetFirstKeyViewWithin(NSView *parentView)
+{
+	for (NSView *view in [parentView subviews]) {
+		if ([view canBecomeKeyView]) {
+			return view;
```
HeyImChris, line 52: I haven't looked into Apple's algorithm here but thought I'd ask - do we know that Apple does basically a depth first search through the view hierarchy instead of breadth first to find the first view to use?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -352,6 +366,7 @@ - (void)keyDown:(NSEvent *)event
 		|| (focusZoneDirection == FocusZoneDirectionNone))
 	{
 		// do nothing
```
HeyImChris, line 368: Not doing nothing anymore :)

HeyImChris (changes requested):

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Yes. It would also be helpful to implement keyboard overrides for the 'Home' and 'End' keys (which I found out navigate to the beginning and end of a list .

The behavior of my ContextualMenu vs NSmenu is slightly different. In an NSMenu, you have to ArrowDown to focus on the first item. With FURN ContextualMenu, focus is _already_ on the first item, so no arrow key is required. That was a small enough difference I was willing to keep that change...

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: We actually don't know. The closest answer I could find is here https://cocoadev.github.io/KeyViewLoopGuidelines/

HeyImChris: Couldn't we just test it? Make a view hierarchy like so [v1: [v2: [v4], v3]] where v2 has canBecomeKeyView set to NO and it's YES for v3 and v4.

I'm not sure how common this would come up, but it would be notable difference worth at least considering even if we end up ignoring it

packages/components/FocusZone/macos/RCTFocusZone.m
HeyImChris: What adds to the complications is that for submenus the first item always _is_ in focus. Just the root menu it's not...

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: That's fair. I can write a test app, try this out, and update.

Saadnajmi: That's fair. I can write a test app, try this out, and update.

packages/components/ContextualMenu/src/Submenu.tsx
Saadnajmi: Tested. This won't work on win32 because the native behavior of arrow keys is to move focus. In order to supress that and _actually_ override the keypress, we'd need to specify the prop `keyDownEvents=[{code: 'ArrowLeft'}]`. That would be a good thing to add to `useKeyProps / useKeyUpProps / useKeyDownProps`....

Saadnajmi:
This PR (just by the change notes) is starting to grow. I'm thinking I may just break it apart into like.. 3-5 changes. Will think about that and update...

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -124,13 +139,14 @@ static inline BOOL IsHorizontalNavigationWithinZoneAction(FocusZoneAction action
...
 {
-	// Accept firstResponder on FocusZone itself in order to reassign it to defaultTabbableElement.
+	// Accept firstResponder on FocusZone itself in order to reassign it within the FocusZone.
 	return !_disabled;
 }
 
 - (BOOL)becomeFirstResponder
 {
-	return !_disabled && [[self window] makeFirstResponder:_defaultTabbableElement ?: [self nextValidKeyView]];
+	NSView *keyView = _defaultKeyView ?: GetFirstKeyViewWithin(self);
```
harrieshin, line 148: where do we actually set this property?

Saadnajmi: RCTFocusZoneManager.m, line 43. See my change in that file.

harrieshin: ah thanks! so... if nothing is to put focus on within itself, it will still put the firstreponder to the next one(sibling view)?

Saadnajmi: Yeah it defaultKeyViw is nil (as is the case with Contextualmenu), then it uses my helper function to focus on the first key view within itself.

Saadnajmi:
I decided to break apart this PR into 3-5 smaller PR's, to make it easier to review each change (and also separate macOS only changes from cross-platform changes). Expect a few new PR's shortly.

PR #1407, opened 2022-02-03
Automated E2E Testing - Improve Assert Tracking

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

There are two types of asserts we can see:
1. Some asserts take focus **away** from the app, and you **can't interact** with the app until you close the assert dialogue. 2
2. Other asserts do **not** take focus away from the app. It pops up, but still **allows for interaction** in RNTester.

This is an issue with automated E2E testing. When the **2nd type** of assert pops up, E2E testing continues because the assert dialogue doesn't take focus away from the app, and if the assert isn't a crashing one, the test passes successfully.

This is bad for two reasons:

- We can easily miss asserts while the test passes - opening the door for regressions.
- This affects functional testing. With the assert dialogue open, it can negatively interact with click/keyboarding events and cause tests to wrongfully fail.

This PR addresses this issue and improves assert tracking across our testing infrastructure.

**Improvements**
At the end of each test, we query the test driver for the number of window handles currently open associated with FluentTester. When all is working correctly, windowHandles = 1 (the only window open is FluentTester). However, if windowHandles > 1, we know an additional assert dialogue opened up.

With this information, we fail the test and perform some logic in the `afterTest()` hook that puts focus on the assert, takes a screenshot, and closes the assert dialogue and attempts to continue with any further tests.

Another small improvement this PR introduces is better error messages for specific failures.

### Verification

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/E2E/common/BasePage.win.ts
```diff
@@ -131,6 +131,23 @@ export class BasePage {
...
   }
 
+  /* FUNCTIONALITY: Determines if an Assert has fired. True if yes; false otherwise
+   *
+   * WHY/HOW: Unfortunately, some Asserts don't take focus away from the test app. Because of this, the test execution is able
+   * to continue without realizing an assert as popped up (unless the assert is a crashing one).
+   *
+   * In order to detect an assert, this function gets the number of window handles currently open within the test app.
+   * Only one instance SHOULD be open at a time (the app itself). If another instance of the app is open, we know an assert dialogue
+   * has popped up. With this information, we know to fail the test.
```
ksiler, line 141: Are we sure assert dialogs are the only case where we have another window handle open within the test app? I think we have some buttons and links that open other dialogs (like an alert). Those would probably count as a window handle, no?

image

samuelfreiberg: That's a great point. In these tests (we don't have any yet), we can alter the logic a bit to validate another window **does** show up. In all the tests where we **don't** expect a pop-up, we can use the current logic. So it can be on a per-test basis. I don't think there are many situations like this, maybe ~5? But it's a great point to bring up and I'll consider it when I'm writing tests like that.

samuelfreiberg: Unfortunately, assert dialogue's are Windows dialogue's and not NetUI. If it was NetUI, we could easily query the Assert itself. But being a Windows handle makes these situations more difficult to handle

apps/fluent-tester/src/E2E/CheckboxExperimental/specs/ExperimentalCheckbox.spec.win.ts
```diff
@@ -10,7 +10,7 @@ import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT, CHECKBOX_A11Y_ROLE } from '../../common
 describe('Experimental Checkbox Testing Initialization', () => {
   it('Wait for app load', () => {
     NavigateAppPage.waitForPageDisplayed(BOOT_APP_TIMEOUT);
-    expect(NavigateAppPage.isPageLoaded()).toBeTruthy();
+    expect(NavigateAppPage.isPageLoaded()).toBeTruthy(ExperimentalCheckboxPageObject.ERRORMESSAGE_APPLOAD);
```
rurikoaraki, line 13: >ExperimentalCheckboxPageObject [](http://example.com/codeflow?start=54&length=30)

SHould this be NavigateAppPage for consistency with other pages?

I'm guessing in general it doesn't matter where it comes from, that ExperimentalCheckboxPageObject is inheriting ERRORMESSAGE_APPLOAD from some BaseAppPage, but it seems like you're trying to not depend on the individual test pages for the first load test?

samuelfreiberg: Good catch! Changing it now

rurikoaraki (approved):

PR #1409, opened 2022-02-04
Add Scrollview and FocusZone slots to ContextualMenu

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is the first of a few changes that I'm breaking https://github.com/microsoft/fluentui-react-native/pull/1400 apart into.

This change refactors the subviews "ScrollView" and "FocusZone" into slots. Doing so allows us to set up and reference their state in `usePrepareProps`. This would be useful if I want to say, set a componentRef on FocusZone that I want to access later on. 

This change also makes scrollView a required slot, and adds it to the macOS contextualMenu. Turns out that's fine, things still behave as they did before. The FocusZone slot is made optional, as it is only needed on macOS (where we can't move focus with arrow keys by default). 

### Verification

This change should largely be a no-op, and CI's E2E tests should catch regressions (as they did in #1400).

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/ContextualMenu/src/ContextualMenu.tsx
+++ b/packages/components/ContextualMenu/src/ContextualMenu.tsx
@@ -73,6 +73,12 @@ export const ContextualMenu = compose<ContextualMenuType>({
           style: { maxHeight: maxHeight, width: maxWidth },
         },
       }),
+      scrollView: {
+        showsVerticalScrollIndicator: true,
+      },
+      focusZone: {
+        focusZoneDirection: 'vertical',
+      },
     });
 
     return { slotProps, state };
@@ -81,6 +87,8 @@ export const ContextualMenu = compose<ContextualMenuType>({
   slots: {
     root: Callout,
     container: View,
+    scrollView: ScrollView,
+    focusZone: FocusZone,
   },
   styles: {
     root: [backgroundColorTokens, borderTokens],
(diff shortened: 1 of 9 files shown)


packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -95,13 +103,7 @@ export const ContextualMenu = compose<ContextualMenuType>({
...
           <Slots.container>
-            {Platform.OS === 'macos' ? (
-              <FocusZone focusZoneDirection={'vertical'}>{children}</FocusZone>
-            ) : (
-              <ScrollView contentContainerStyle={{ flexDirection: 'column', flexGrow: 1 }} showsVerticalScrollIndicator={true}>
-                {children}
-              </ScrollView>
-            )}
+            <Slots.scrollView>{Platform.OS === 'macos' ? <Slots.focusZone>{children}</Slots.focusZone> : { children }}</Slots.scrollView>
           </Slots.container>
```
rurikoaraki, line 107: nit: I find this harder to parse/read than the format it was in before. Prettier probably won't let you split this into several lines?

Maybe assigning the internal part to a variable might help readability? ex

```tsx
const inner = Platform.OS === 'macos' ? <Slots.focusZone>{children}</Slots.focusZone> : { children };

...

<Slots.scrollView>{inner}</Slots.scrollView>
```

Saadnajmi: Yeah, Prettier won't. I like that suggestion!

rurikoaraki: I guess similar comment for Submenu

rurikoaraki (approved):
:shipit:

rurikoaraki (approved):

packages/components/ContextualMenu/src/ContextualMenu.tsx
Saadnajmi: Unfortunately, this optional rendering of a slot caused issues, presumable because `<Slots.focusZone>{children}</Slots.focusZone>` is of type `JSX.Element` while `{ children }` is of type `React.ReactNode[]`.

The E2E tests were failing with this error in the screenshot below. I was only able to fix it locally by just duplicating the entire tree inside an if/else 😢 

error

PR #1410, opened 2022-02-04
Several macOS FocusZone fixes

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is the one of a few changes that I'm breaking https://github.com/microsoft/fluentui-react-native/pull/1400 apart into.

This change makes several bug fixes to macOS FocusZone. 

**1. Renames defaultTabbableElement macOS native property to defaultKeyView.**

The public JS API is still defaultTabbableElement, but the native module now sets a property with the name defualtkeyView. The name feels much more idiomatic to me for macOS. The public facing API of FocusZone does not change, so this shouldn't affect any downstream clients.

**2. Have FocusZone focus on the first key view within itself, rather than it's nextValidKeyView.**

Previously, FocusZone would focus on its' nextValidKeyView if no defaultTabbableElement was set. I believe the intent was "focus on the first key view inside the FocusZone". However, I believe this would actually focus on the first key view outside the FocusZone. In the case of ContextualMenu (where there was no key view after the FocusZone), it meant focus went nowhere. Instead, I wrote a helper method GetFirstKeyViewWithin(NSView *parentView) that will recursively search for the first key view within a view's subviews, and set that as the "firstKeyView".

**3. We now call [super keyDown:Event] if a FocusZone does not handle a keyboard event, instead of doing nothing.**

If the user pressed Arrow(Left|Right) in a vertical FocusZone or Arrow(Up|Down) in a horizontal FocusZone, then focusZone would ignore the key event. This is incorrect, it should call [super keyDown:event] in those cases so that it's superview / base class may be able to handle the keyboard event. This prevented us from letting ArrowLeft close a submenu.

### Verification

The FocusZone test page still behaves as expected. Additionally, the RadioGroup and Tabs (both of which depend on FocusZone) test pages also behave as expected.

Pending: Test downstream integration into Office to make sure the changes don't break existing clients. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -45,6 +45,21 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
 	);
 }
 
+static NSView *GetFirstKeyViewWithin(NSView *parentView)
+{
+	for (NSView *view in [parentView subviews]) {
+		if ([view canBecomeKeyView]) {
+			return view;
+		}
+
+		NSView *match = GetFirstKeyViewWithin(view);
+		if (match) {
+			return match;
+		}
+	}
+	return nil;
+}
+
 static NSView *GetFirstResponder(NSWindow *window)
 {
 	NSResponder *responder = [window firstResponder];
@@ -124,13 +139,14 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 
 - (BOOL)acceptsFirstResponder
 {
-	// Accept firstResponder on FocusZone itself in order to reassign it to defaultTabbableElement.
+	// Accept firstResponder on FocusZone itself in order to reassign it within the FocusZone.
 	return !_disabled;
 }
 
 - (BOOL)becomeFirstResponder
 {
-	return !_disabled && [[self window] makeFirstResponder:_defaultTabbableElement ?: [self nextValidKeyView]];
+	NSView *keyView = _defaultKeyView ?: GetFirstKeyViewWithin(self);
+	return !_disabled && [[self window] makeFirstResponder:keyView];
 }
 
 - (NSView *)nextViewToFocusForCondition:(IsViewLeadingCandidateForNextFocus)isLeadingCandidate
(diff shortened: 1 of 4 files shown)


packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,21 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
...
+			return view;
+		}
+
+		NSView *match = GetFirstKeyViewWithin(view);
+		if (match) {
+			return match;
+		}
+	}
+	return nil;
+}
```
Saadnajmi, line 61: Comment from @HeyImChris last time: (https://github.com/microsoft/fluentui-react-native/pull/1400#discussion_r799040956): 
> Would it make sense to have a GetLastKeyViewWithin to get the last menu button for when a user launches a menu and presses ArrowUp?

Yes, though I'd rather implement that as a separate change..

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -45,6 +45,21 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
...
+{
+	for (NSView *view in [parentView subviews]) {
+		if ([view canBecomeKeyView]) {
+			return view;
+		}
+
+		NSView *match = GetFirstKeyViewWithin(view);
+		if (match) {
+			return match;
+		}
```
Saadnajmi, line 58: Re: @HeyImChris 's comment last time (https://github.com/microsoft/fluentui-react-native/pull/1400#discussion_r799043491):

Verify that this depth first search algorithm is accurate to what Apple does internally with a simple test app.

HeyImChris (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: I tested this with a simple test inside of RN-Tester. The yellow boxes are focusable. If it was depth-first, then the first yellow box would get focused. If it was breadth-first, then the second yellow box would get focused. In both cases (vertical and horizontal), the first yellow box was focused, implying depth first.

```
class KeyViewExample extends React.Component<{}> {
  render() {
    return (
      <View style={{padding: 10}}>
        <View style={styles.outer}>
          <View style={styles.middle}>
            <View style={styles.innerFocusable} focusable={true} />
          </View>
        </View>
        <View style={styles.outer}>
          <View style={styles.innerFocusable} focusable={true} />
        </View>
      </View>
    );
  }
}
```

Screen Shot 2022-02-04 at 11 32 49 AM
Screen Shot 2022-02-04 at 11 33 08 AM

PR #1411, opened 2022-02-04
ContextualMenu: Arrow Right/Left to open/close submenus

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is the one of the changes that I'm breaking https://github.com/microsoft/fluentui-react-native/pull/1400 apart into.

This change does two things that are needed to support opening and closing submenus with arrow keys inside a ContextualMenu. Note we won't get this behavior on macOS until a few other PRs land.

First, this change let's you close submenus with `ArrowLeft`. This is pretty neat, and matches native behavior of Contexual menus!

Second, this change fixes a regression introduced by https://github.com/microsoft/fluentui-react-native/pull/1373 , where on win32, SubmenuItems no longer opened their submenu with `ArrowRight`. The reason is that change defaulted all keyboard events on win32 to `onKeyUp`, but win32 moves focus with `onKeyDown`. In the case of SubmenuItem, `ArrowRight` moves focus to the next menu item before our `onKeyUp` override ever got a chance to open the submenu. 

However, the original implementation also had a fault. It turns out on windows / win32, in order for your `onKey(Up|Down)` override to suppress the default native behavior (like `Tab` or `ArrowRight` moves focus with keyDown), you must specify an additional prop `keyDownEvents`. The reason is quite complex, you can read about it [here](https://github.com/microsoft/react-native-windows/blob/main/vnext/proposals/active/keyboard-reconcile-desktop.md#declarative-properties). If you _don't_ specify keyDown events, then both your JS `onKeyDown` callback and the native XAML/NetUI behavior will happen asynchronously. In effect, I think we had a race condition that we happened to win before. Note we don't have this problem on macOS because the similar macOS prop `validKeysDown` is set and suppresses native behavior. 

I fix the above issue by having our hook `useKey(Up|Down)Props` also specify `key(Up|Down)Events`. Note this changes the behavior of _all_ FURN win32 components that had keyboard overrides. It's arguable whether we _always_ want to suppress the native behavior, but it feels like we were operating on that assumption anyway since nobody seems to know about this prop 🤷🏾. At least now, the behavior is consistent across platforms. 

### Verification

I don't know how to screen record on Windows 11 (the one built into Xbox Game bar doesn't record floating windows on top of FluentTester)... so I'll describe what I see.

On Win32, ArrowRight opens submenus again once more, and ArrowLeft closes submenus.

To test native keyboard behavior suppression, I made a small test. I made a button, and override `Tab` on key down to launch an alert. Previously, both the alert would fire and focus would move off the button. Now, just the alert fires. 

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/utils/interactive-hooks/src/useKeyProps.ts
```diff
@@ -1,13 +1,16 @@
...
 
 export type KeyPressProps = {
+  onKeyDown?: KeyCallback;
+  validKeysDown?: string[]; // macOS
+  keyDownEvents?: IHandledKeyboardEvent[]; // win32
   onKeyUp?: KeyCallback;
   validKeysUp?: string[];
-  onKeyDown?: KeyCallback;
-  validKeysDown?: string[];
+  keyUpEvents?: IHandledKeyboardEvent[]; // win32
```
rurikoaraki, line 13: Is the comment here supposed to say macos?

rurikoaraki: Oh, nvm. I think I understand

Saadnajmi: I did miss a macOS for validKeysUp. so that's fair.

What's fun is windows has slightly different type for their `keyDownEvents` prop. I intentionally ommitted adding that to not deal with that for now :D

packages/utils/interactive-hooks/src/useKeyProps.win32.ts
```diff
@@ -37,20 +37,26 @@ export function getKeyCallbackWorker(userCallback?: KeyCallback, ...keys: string
 function getKeyUpPropsWorker(userCallback: KeyCallback, ...keys: string[]): KeyPressProps {
   return {
     onKeyUp: getKeyCallbackWorker(userCallback, ...keys),
+    keyUpEvents: keys.map((keyCode) => {
+      return { key: keyCode };
+    }),
```
rurikoaraki, line 42: suggestion: maybe add a comment on what this is doing/why?

rurikoaraki: I guess there's a comment in Submenu so maybe it's unnecessary

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -111,10 +111,11 @@ export const SubmenuItem = compose<SubmenuItemType>({
       icon: !!icon,
     };
 
-    /*
-     * SubmenuItem launches the submenu onMouseEnter event. For keyboarding, submenu should be launched with Spacebar, Enter, or right arrow.
+    /**
+     * SubmenuItem launches the submenu onMouseEnter event. submenu should be launched with Spacebar, Enter, or right arrow.
```
rurikoaraki, line 115: >s [](http://example.com/codeflow?start=60&length=1)

Super nit: capitalization

Saadnajmi: ..uh.. which S?

rurikoaraki: Ah sorry:

"event. submenu"

packages/utils/interactive-hooks/src/useKeyProps.win32.ts
Saadnajmi: It's turning `['a', 'b', 'c']` into `[{key: 'a'}, {key: 'b'}, {key: 'c}]`. I figured that was simple enough that it didn't need a comment.

The reason for the array of objects is to support key combinations. You might want Ctrl+Enter (`{ ctrlKey: true, key: 'Enter'}`) to be suppressed, while Shift+Enter passes through.

rurikoaraki: Oh I meant the general intent of supplying the events to the prop so that they don't get listened to on the native side, not the array mapping

rurikoaraki:
I'd probably make sure the behavior seems good from someone with more knowledge of context menus, but otherwise this change seems good to me.

rurikoaraki (approved):
:shipit:

packages/utils/interactive-hooks/src/useKeyProps.win32.ts
Saadnajmi: Ah. That would be a super simple one liner that avoids needing to find this PR and read my paragraphs of PR notes. Nice catch!

rurikoaraki (approved):

harrieshin:
i think you said in the previous notes, in RTL it is still right arrow?

Saadnajmi:
> i think you said in the previous notes, in RTL it is still right arrow?

Good catch. This change does not handle switching the keyboard override for RTL. That would still need to be a followup. 
I still want to keep that a separate fix. An RTL fix would be dependent on this change going in, and there's enough of a win32/macOS change in this PR already that I want to validate this is OK first.

PR #1417, opened 2022-02-04
ContextualMenu macOS: Focus on menu after render

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is the one of the changes that I'm breaking https://github.com/microsoft/fluentui-react-native/pull/1400 apart into.

This is a macOS only change.  This changes adds a layoutEffect to ContextualMenu to focuses on it's FocusZone slot, effectively placing focus on the menu. 

Unlike Win32, Callouts / NSWindows on macOS do not automatically place focus on the first focusable item once presented. Instead, you have to natively set an NSWindows' `initialFirstResponder` property to do that. 

So now, I could add a new property to Callout called initialFocusRef, have it internally map to [NSWindow initialFirstResponder], and then have ContextualMenu set it's initialFocusRef to it's FocusZone slot...
... or I could just use the hook useLayoutEffect() to imperatively focus on the inner focus zone after render. This solution was much less code and seemed simpler. I also needed a setTimeout(..., 0) in there for some reason 🤷🏾. 

I'll probably go through the path of properly adding a new prop initialFocusRef to Callout after this change goes in, since that seems the more correct solution. I would also want to take some time and reconcile the win32 props (`setInitalFocus/focusOnMount/focusOnContainer`) and see if we can support them on macOS. 

### Verification

This is the current behavior. We don't get `ArrowLeft` to close submenu yet, as that depends on #1411 to land.

https://user-images.githubusercontent.com/6722175/152618264-3842223f-3e21-40eb-a9b1-7b0e16a23ea8.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/ContextualMenu/src/ContextualMenu.tsx
+++ b/packages/components/ContextualMenu/src/ContextualMenu.tsx
@@ -12,7 +12,7 @@ import {
 } from './ContextualMenu.types';
 import { settings } from './ContextualMenu.settings';
 import { IUseComposeStyling, compose } from '@uifabricshared/foundation-compose';
-import { useSelectedKey } from '@fluentui-react-native/interactive-hooks';
+import { IFocusable, useSelectedKey } from '@fluentui-react-native/interactive-hooks';
 import { mergeSettings } from '@uifabricshared/foundation-settings';
 import { backgroundColorTokens, borderTokens } from '@fluentui-react-native/tokens';
 import { Callout } from '@fluentui-react-native/callout';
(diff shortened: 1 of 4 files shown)


packages/components/ContextualMenu/src/ContextualMenu.tsx
```diff
@@ -34,6 +34,21 @@ export const ContextualMenu = compose<ContextualMenuType>({
   usePrepareProps: (userProps: ContextualMenuProps, useStyling: IUseComposeStyling<ContextualMenuType>) => {
     const { setShowMenu, maxHeight, maxWidth, shouldFocusOnMount = true, shouldFocusOnContainer = false, ...rest } = userProps;
 
+    /**
+     * On macOS, focus isn't placed by default on the first focusable element. We get around this by focusing on the inner FocusZone
```
HeyImChris, line 38: Interesting that the OS behavior for menus actually isn't to put focus in the first item of the root menu. This fix is an awesome improvement over the current state but ideally wouldn't we want the root menu to only get focus on the first key up/down press?

May be worth noting that OS difference if we do end up not aligning with NSMenus there.

HeyImChris: Definitely makes sense for submenus though as that does mimic NSMenu submenu keyboard navigation/accessibility.

HeyImChris (approved):

packages/components/ContextualMenu/src/ContextualMenu.tsx
Saadnajmi: Yes, @mischreiber brought up the same point to me as well. I have an idea of how we might emulate that behavior (Have a custom container view for the menu with overrides for ArrowUp and ArrowDown, and focus on that instead). 

I _may_ file that as a separate bug under "polish", since this specific big has been lurking for a while. I can update the comment here though!

Saadnajmi: I went with filing a bug in ADO and not updating the comment because I couldn't think of a way to word it without getting _real_ specific (in a way that wouldn't help say, a new windows developer looking at this who doesn't know what an NSMenu is)

PR #1418, opened 2022-02-05
Add `smallerMargins` prop to experimental checkbox

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Note: Not really sure about this name if anyone has better suggestions I'd be open to them.

Win32 has checkboxes with smaller spacing around/between the elements of the checkbox. This is not a specification on other endpoints, so we're on our own with the prop name here.

Adding a prop which enables this smaller-spacing mode of the checkbox. Added a test for it, too.

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
For naming, what about making a size enum and having this be the "small" size?

rurikoaraki:
Removing this prop as we don't plan on supporting this scenario

PR #1428, opened 2022-02-08
ContextualMenu: Don't dismiss menu on SubmenuItem click

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is a small change to the shared JS of ContextualMenu, and a large change to the native macOS Callout code.

**Shared (macOS/Win32) changes **

The shared JS change is that we no longer dismiss the menu and submenu inside SubmenuItem's `onPress`. This affects both macOS and win32, and fixes a bug we saw downstream.

** macOS only changes **

In the macOS Callout code, we now add submenus as childwindows of the NSWindow they are presented from. We can do this easily because each "CalloutWindow" class is created from a placeholder "Callout" view (which itself is in a window). So, everytime we create a CalloutWindow, we can just look our Callout Views' window and add ourselves as a child 🥳.. 

Now that we _have_ a parent/child relationship, we can update our "dismiss on click outside window" code to check if our click was in a child window (AKA: We are a Menu and the click was in a submenu), or in a parent window (AKA: We are a submenu and the click was in a menu). The end result is less unintentional dismissal of menus! 🥳

Some other minor changes:
  - Replaced `window.close()` with `window.orderOut()`. 
    - I mostly did this so we don't unintentionally release on close, and because downstream code in Office did the same.
  - Remove the mouseEventMonitor on Callout dismissal
     - I noticed that because we don't deallocate our windows when they close, we were still monitoring for mouse events. Leaking NSEvent monitors seemed like a bad idea, so now we're more careful to remove them when the window is closed.  

### Verification

I still don't know how to record my screen on win32, but I did verify that the Submenu wouldn't dismiss when we clicked on a SubmenuItem. I would still want a win32 engineer from CXE ( @lenahong ? ) to verify this change is fine.

On macOS, you can see the behavior with the childwindow support in the video below. Notice the Menu window moves with the main window when the main window moves. Also notice clicking on the submenu item does not dismiss the submenu window. Also notice clicking on a disabled menu item in the submenu window does not dismiss the main menu window.

https://user-images.githubusercontent.com/6722175/153095306-14ee95ab-3391-4136-8a59-d759f89b2c25.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -22,11 +22,36 @@ class CalloutWindow: NSWindow {
...
+			if (clickedWindow.isEqual(to: self) ) {
+				shouldDismissCallout = false
+			} else if (clickedWindow.isMainWindow) {
+				shouldDismissCallout = true
+			} else if (clickedWindow.isEqual(to: self?.parent) && self?.parent is CalloutWindow) {
+				// We are a child window of a Callout (e.g: a ContextualMenu Submenu), and the click happened in our parent
+				shouldDismissCallout = false
+			} else {
+				// Check if the click happened in any of our child windows (e.g: our submenus)
+				if let childWindows = self?.childWindows {
```
harrieshin, line 42: childWindows ... with a "s" it sounds like it is an array?

Saadnajmi: It is an array. You can have multiple childwindows.

packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -22,11 +22,36 @@ class CalloutWindow: NSWindow {
...
+
+			if (clickedWindow.isEqual(to: self) ) {
+				shouldDismissCallout = false
+			} else if (clickedWindow.isMainWindow) {
+				shouldDismissCallout = true
+			} else if (clickedWindow.isEqual(to: self?.parent) && self?.parent is CalloutWindow) {
+				// We are a child window of a Callout (e.g: a ContextualMenu Submenu), and the click happened in our parent
+				shouldDismissCallout = false
+			} else {
+				// Check if the click happened in any of our child windows (e.g: our submenus)
```
Saadnajmi, line 41: I feel a little iffy mentioning Menus/submenus inside Callout (since Callout doesn't know / shouldn't know about what it's hosting), but this was the best way I could describe these cases.

packages/components/Callout/macos/CalloutWindow.swift
harrieshin: ah thansk. something about line 42 and line 43 doesn't feel very swifty. self?.childWindws.contain(clickedwindow) ?? true  ?

Saadnajmi: That's fair. The checks with optional Bools (`Bool?`) trip me up. I tried using the `if let` to be more swifty, but I think the nil coalescing operator might be cleaner overall!

packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -47,7 +68,12 @@ class CalloutWindow: NSWindow {
 
 	@objc private func dismissCallout() {
 		lifeCycleDelegate?.calloutWillDismiss(window: self)
-		close()
+
+		if let monitor = mouseEventMonitor {
+			NSEvent.removeMonitor(monitor)
+		}
+
+		orderOut(self)
```
harrieshin, line 76: Calling orderOut: "causes the window to be removed from the screen, but does not cause it to be released." so when does it actually release?

packages/components/Callout/macos/CalloutWindow.swift
Saadnajmi: It ended up needing to be `self?.childWindows.contains(clickedwindow) ?? false ` but I like this better! Thanks =)

packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -22,11 +22,32 @@ class CalloutWindow: NSWindow {
...
 
-		// Dismiss the Callout if the user touched/clicked outside the bounds of the callout.
+		// Dismiss the Callout if the user clicked outside the Callout Window or any of our child windows.
 		mouseEventMonitor = NSEvent.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { [weak self] (event) -> NSEvent? in
-			if (event.window != self) {
+			guard let clickedWindow = event.window else {
+				return event
+			}
+
+			var shouldDismissCallout = false
```
harrieshin, line 31: should this be true then you don't need line 43?

harrieshin (approved):

packages/components/Callout/macos/CalloutWindow.swift
Saadnajmi: I could probably also chop off line 36 and 36. I'll sleep on it, my brain is tired of looking at long if/else trees 🤣

packages/components/Callout/macos/CalloutWindow.swift
Saadnajmi: Presumably when the placeholder view `Callout` (What is defined in `Callout.Swift`) that owns the window is deallocated, since Callout is one of it's private variables. React Native manages the ownership of the placeholder view. I wanted the window to be allocated as long as it's placeholder view was allocated.

packages/components/Callout/macos/CalloutWindow.swift
Saadnajmi: Actually.. seems to work.

packages/components/Callout/macos/CalloutWindow.swift
```diff
@@ -22,11 +22,29 @@ class CalloutWindow: NSWindow {
...
+			guard let clickedWindow = event.window else {
+				return event
+			}
+
+			var shouldDismissCallout = true
+
+			if (clickedWindow.isEqual(to: self) ) {
+				// The click happened in our own window. 
+				shouldDismissCallout = false
+			} else if (clickedWindow.isEqual(to: self?.parent) && self?.parent is CalloutWindow) {
```
harrieshin, line 36: again super nit but would this make more swift like? 
 if let parentWindw = self?.parent as? CalloutWindow, clickedwindow.isequal(to: parentWindow).

Saadnajmi: `if let` is usually used to resolve an optional, and in this case we never use the resulting `parentWindow` variable. 
Maybe something like this?
``` 
} else if (clickedWindow.isEqual(to: self?.parent as? CalloutWindow)) {
```

Saadnajmi: https://developer.apple.com/swift/blog/?id=23 This blog is telling me that if the conversion`self?.parent as? CalloutWindow` fails, it evaluates to nil, so I think this is OK?

> let animal: Animal = Cat()
animal as? Dog	// evaluates to nil
animal as! Dog	// triggers a runtime error

Saadnajmi: An aside... I think this logic only works for one level of submenus. If someone had 2 levels of submenus (Menu -> Submenu -> Submenu) I think it breaks :/

PR #1433, opened 2022-02-10
Update button tokens

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Major visual changes:

1. Added border to default button
2. Added border to primary button when in high contrast mode
3. Removed hover effects
4. Used a lighter shade for the default button background color when in dark mode

### Verification

| image |image |
| image | image |
| image | image |
| image |image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Button/src/ButtonColorTokens.macos.ts
```diff
@@ -0,0 +1,95 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { ButtonTokens } from './Button.types';
+
+export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme) =>
```
Saadnajmi, line 5: Do we need to fork the application of the tokens by platform too? I thought that heading separate macOS alias tokens meant we wouldn't need a separate macOS file here?

packages/experimental/Button/src/ButtonColorTokens.macos.ts
```diff
@@ -0,0 +1,95 @@
...
+  ({
+    backgroundColor: t.colors.defaultBackground,
```
rurikoaraki, line 7: Not blocking/doesn't need to be addressed soon, but we should eventually move away from using the old semantic colors and hooking into alias tokens

DanielMorM (approved):
The theme looks much better and aligned to macOS style. Thanks for fixing this, Amanda!

packages/experimental/Button/src/ButtonColorTokens.macos.ts
chiuam: Oh, I added this file because I wanted to define a different set of tokens from the default file

Saadnajmi: I thought that if we wanted a different set, then we could change the macOS alias tokens to whatever we needed, and at this level the tokens would stay the same? 
I guess, what were the specific changes that needed this file to differ?

harrieshin:
talked offline mostly but based on the screenshot i see, we could adjust the followings:
(1) fontweight of all the button titles -> its too bold
(2)"primary disabled" darkmode/highcontrast border color --> doesn't seem like it should be branding color 
(3) high contrast, light mode , "block primary" button doesn't have dark borders but "primary compound" border has black borders -->both of them should have black borders?

packages/experimental/Button/src/ButtonColorTokens.macos.ts
chiuam: Tokens should stay the same at this level. I like that I get to customize / style the button tokens without changing the definitions of the control tokens, which could potentially affect other components outside of Button.

chiuam:
> talked offline mostly but based on the screenshot i see, we could adjust the followings: (1) fontweight of all the button titles -> its too bold (2)"primary disabled" darkmode/highcontrast border color --> doesn't seem like it should be branding color (3) high contrast, light mode , "block primary" button doesn't have dark borders but "primary compound" border has black borders -->both of them should have black borders?

Thank you for the suggestions, screenshots are updated :)

packages/experimental/Button/src/ButtonColorTokens.macos.ts
Saadnajmi: If the tokens should stay the same at this level, why do we need a `ButtonColorTokens.macos.ts` file in addition to `ButtonColorTokens.ts` ?

packages/theming/theme-types/src/Color.types.ts
```diff
@@ -533,6 +535,7 @@ export interface AliasColorTokens {
   brandBackground: ColorValue;
   brandBackgroundHover: ColorValue;
   brandBackgroundPressed: ColorValue;
+  brandBackgroundDisabled: ColorValue;
```
rurikoaraki, line 538: Is this color different from `neutralBackgroundDisabled`? For other platforms the disabled color has been the same across all backgrounds, which is why there's only one `neutralBackgroundDisabled` token

packages/experimental/Button/src/ButtonColorTokens.macos.ts
chiuam: If we apply alias tokens directly here, each platform may have different sets of alias tokens for Button. Adding this file is necessary to differentiate custom styling per platform. (similar to win32)

chiuam:
> I'm a little surprised that subtle buttons don't have borders in MacOS' high contrast mode? Windows HC has borders for all buttons.

Actually, I think the subtle button should have a border when in high contrast mode. Let me go add it

packages/experimental/Button/src/ButtonColorTokens.macos.ts
Saadnajmi: OK, I see that there is a win32 and windows specific file. This is where the alias tokens are assigned to the control tokens, OK.

Saadnajmi:
I cloned this locally and it seems the font also changed? Do we need a ButtonFontTokens.macos file?
Also, the "Customized Button with Ref" still has a hover effect that is quite different from the rest state.

Screen Shot 2022-02-14 at 2 01 56 PM

chiuam:
> I cloned this locally and it seems the font also changed? Do we need a ButtonFontTokens.macos file? Also, the "Customized Button with Ref" still has a hover effect that is quite different from the rest state.
> 
> Screen Shot 2022-02-14 at 2 01 56 PM

Thanks for testing! I did add that file in 3d9de1bc3056fc4a194d5e258033cfe9868dd468. I believe the font change is expected since it uses the experimental button?

rurikoaraki:
> I cloned this locally and it seems the font also changed? Do we need a ButtonFontTokens.macos file? Also, the "Customized Button with Ref" still has a hover effect that is quite different from the rest state.
> 
> Screen Shot 2022-02-14 at 2 01 56 PM

Yeah I only overrode the color of the rest state, hover/pressed will revert back to default colors

Saadnajmi:
> > I cloned this locally and it seems the font also changed? Do we need a ButtonFontTokens.macos file? Also, the "Customized Button with Ref" still has a hover effect that is quite different from the rest state.
> > Screen Shot 2022-02-14 at 2 01 56 PM
> 
> Thanks for testing! I did add that file in [3d9de1b](https://github.com/microsoft/fluentui-react-native/commit/3d9de1bc3056fc4a194d5e258033cfe9868dd468). I believe the font change is expected since it uses the experimental button?

Ah, didn't notice that sorry! Interesting, It does seem to be the apple theme  variant. Size 13, font "System"...

Screen Shot 2022-02-14 at 2 33 48 PM

Saadnajmi (approved):

Saadnajmi (approved):
I'm still a little confused about at which level the application of tokens should be platform specific and which it should be shared, but overall looks great! Thanks for doing this!

rurikoaraki (approved):

PR #1446, opened 2022-02-16
Update just-scripts to latest

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Updating just-scripts to latest version. This mostly brought up errors for tasks that aren't set up properly because they aren't used (like bundle).

- Removed the test files from the tester folders as they don't seem to do anything or test anything relevant
- Removed bundle command from non-test packages are they are not set up to be bundled
- Removed code-style command as we never use it - use `yarn prettier` and `yarn lint/test` instead
- Removed linting option from some packages that don't have code to lint.

# Verification

Ran the following locally:
- build
- bundle
- clean
- lint
- prettier
- prettier-fix
- test

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


harrieshin (approved):

Saadnajmi:
A dumb gut check: I noticed your fork is 4 commits behind the current master (https://github.com/rurikoaraki/fluentui-react-native). Beachball compares your branch to your remote `origin` when making change files. If `origin` is behind (while your branch is not) it may make unnecessary changefiles. I felt like there was a lot more change files than I expected in this PR so it may be worth double checking that.

rurikoaraki:
> A dumb gut check: I noticed your fork is 4 commits behind the current master (https://github.com/rurikoaraki/fluentui-react-native). Beachball compares your branch to your remote `origin` when making change files. If `origin` is behind (while your branch is not) it may make unnecessary changefiles. I felt like there was a lot more change files than I expected in this PR so it may be worth double checking that.

I don't think the change files are excessive? I've removed a line from the package.json of almost every package we have...

Saadnajmi:
> > A dumb gut check: I noticed your fork is 4 commits behind the current master (https://github.com/rurikoaraki/fluentui-react-native). Beachball compares your branch to your remote `origin` when making change files. If `origin` is behind (while your branch is not) it may make unnecessary changefiles. I felt like there was a lot more change files than I expected in this PR so it may be worth double checking that.
> 
> I don't think the change files are excessive? I've removed a line from the package.json of almost every package we have...

That's fair.. my comment is a check that goes off in my head everytime I see a lot of change files in my own PR's. I didn't actually map all your change files to changes so it's probably fine

PR #1453, opened 2022-02-17
Fix iconColor token being no-op

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change corrects the mapping of the `color` prop in iconProps.

### Verification

Added a custom button as a test case, verified the token works for SVGs, font icons, and PNGs.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonSizeTestSection.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonSizeTestSection.tsx
@@ -11,12 +11,12 @@ export const ButtonSizeTest: React.FunctionComponent = () => {
     viewBox: '0 0 500 500',
   };
   const svgIconsEnabled = ['ios', 'macos', 'win32', 'android'].includes(Platform.OS as string);
-
+  const CustomButton = Button.customize({ iconColor: 'yellow' });
   return (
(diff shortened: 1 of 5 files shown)


HeyImChris (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonSizeTestSection.tsx
```diff
@@ -11,12 +11,12 @@ export const ButtonSizeTest: React.FunctionComponent = () => {
     viewBox: '0 0 500 500',
   };
   const svgIconsEnabled = ['ios', 'macos', 'win32', 'android'].includes(Platform.OS as string);
-
+  const CustomButton = Button.customize({ iconColor: 'yellow' });
```
rurikoaraki, line 14: 1. Could we put this in a separate test (maybe the HOC page) so that this test stays purely about showing sizes of buttons?
2. Creating customized components should be outside of any component render functions, otherwise the component is recreated on every render.

rurikoaraki (changes requested):
See comment

rurikoaraki (approved):

PR #1474, opened 2022-02-24
Adding E2E Tests that safeguard our native testing implementation

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On the Native side, we use the ScrollView containing all the test page button's to dynamically pull all the valid test pages + automation IDs (instead of hardcoding a component test list on the native side, this makes it so native always stays in sync with JS).

We need to safeguard this, though. The ScrollView testID cannot change (that's hardcoded into Native because we don't have a way of sharing testIDs). We also want to safeguard the UI structure of the ScrollView being the direct parent of the test page buttons (Native looks at all the Children of the ScrollView to grab the automation IDs).

These tests ensure these conditions hold true, and ensure no accidental changes are made to these patterns (if changes are made, it will break native tests).

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
```diff
@@ -0,0 +1,45 @@
...
+class NativeTestingPageObject extends BasePage {
+  waitForScrollViewDisplayed(timeout?: number): void {
+    browser.waitUntil(
+      () => {
+        return this.doesScrollViewParentExist();
+      },
+      {
+        timeout: timeout ?? this.waitForPageTimeout,
+        timeoutMsg:
+          'The ScrollView containing all test page buttons was not found. Did you remove the ScrollView or change the testID? The testID must stay constant to align with our native code.',
```
PPatBoyd, line 12: we should likely make this common in time and not win32-only but regardless -- since we're codifying a policy rule I'd remove the reference to native code and more ambiguously assert something like `the testID is fixed for testing purposes` since the source of the policy isn't *especially relevant

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  doesScrollViewParentExist(): boolean {
+    return this._scrollViewParent.isExisting();
+  }
+
+  validateScrollViewChildren() {
+    // Gets all the children
+    const testChildren = this._scrollViewParent.$$('//*');
+
+    // All the button children have the same format. We want to find these buttons
```
PPatBoyd, line 26: nit would tweak to something like // Ensure the AutomationId properties of the button children match the defined testing format

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
```diff
@@ -0,0 +1,45 @@
...
+    const reg = new RegExp('Homepage_[a-zA-Z]*_Button');
+    const buttonList = [];
+
+    testChildren.forEach((child) => {
+      const autoId = child.getAttribute('AutomationId');
```
PPatBoyd, line 31: Dedupe the getAttribute call, but also this fn will return true if any of the buttons match, not all of the buttons match, right?  I think we should be reversing the two conditions e.g.

`forEach {
...
if (!autoId || !autoId.match(reg))   { return false; }
...
}`

`return testChildren.length > 0; // and still check that we _do_ have buttons matching the format`

apps/fluent-tester/src/E2E/common/NativeTesting/specs/NativeTesting.spec.win.ts
```diff
@@ -0,0 +1,28 @@
...
+import NativeTestingPageObject from '../pages/NativeTestingPageObject.win';
+import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT } from '../../../common/consts';
+
+describe('Native Safety Check Testing Initialization', () => {
+  it('Wait for app load', () => {
+    NavigateAppPage.waitForPageDisplayed(BOOT_APP_TIMEOUT);
+    expect(NavigateAppPage.isPageLoaded()).toBeTruthy(NavigateAppPage.ERRORMESSAGE_APPLOAD);
+  });
+
+  // The Scrollview in testing is the one that contains all the buttons that navigate to each component's
```
PPatBoyd, line 11: nit Scrollview -> ScrollView
nit here I would also more opaquely assert something to the effect of:
- For testing purposes we require that the root view contains a non-empty, immutable scrollview of buttons that navigate to test pages

sidebar -- depending on future design or for small-screen purposes I think we're likely to make changes in this `codified test requirements` space to enable a nicely organized, high volume of tests e.g. a nested test list to avoid all tests and examples being opened on the same page if we have dozens.  Not an issue at the moment though

change/@fluentui-react-native-tester-a0ed0eb4-ec38-4450-ae20-c0a1773ba265.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Adding tests to ensure JS + Native stay in sync",
```
PPatBoyd, line 3: nit `Add tests to codify testing requirements` I'd avoid references to `native` that could be misleading, an opaque testing requirement is sufficient and we're running the "open all component test pages" test here in the repo too -- native isn't a necessary consideration it's just a policy decision

apps/fluent-tester/src/E2E/common/NativeTesting/specs/NativeTesting.spec.win.ts
```diff
@@ -0,0 +1,28 @@
...
+  //    1) Continues to parent all the test buttons
+  //    2) Keeps testID='SCROLLVIEW_TEST_ID'
+  it('Validate the ScrollView containing all the navigational buttons exists', () => {
+    NativeTestingPageObject.waitForScrollViewDisplayed(PAGE_TIMEOUT);
+    expect(NativeTestingPageObject.doesScrollViewParentExist()).toBeTruthy();
+  });
+
+  // In addition to existing, we want to ensure the Children of the ScrollView stay intact. The children are the buttons that
+  // navigate to each test page. Validate these children exist with the proper automation IDs.
+  it('Validate the children of ScrollView are the buttons that navigate to each test page WITH proper automationId', () => {
```
PPatBoyd, line 24: nit use canonical sentence casing, save fully-capitalized character sequences for acronyms and initialisms (FURN :) )

PPatBoyd: `with` and also above @ `vital`

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
PPatBoyd: sidebar wondering for our test in this repo that opens all of the test pages if we should have a mechanism for verifying that clicking each button child actually generated the expected page change... but maybe that's unnecessary red tape

PPatBoyd (changes requested):

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
samuelfreiberg: Each Platform's native code will be different, so I believe we'll probably need platform-specific files just like this one for Mac, ios, etc.

Also, I changed the error message to something more ambiguous.

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
samuelfreiberg: Changing in newest commit

apps/fluent-tester/src/E2E/common/NativeTesting/specs/NativeTesting.spec.win.ts
samuelfreiberg: Changing both. For the assert, I'm moving it to the PageObject in the `waitForPageDisplayed()` function, as that is where the failure would occur.

And I agree, that sounds good

apps/fluent-tester/src/E2E/common/NativeTesting/specs/NativeTesting.spec.win.ts
samuelfreiberg: Changing in newest iteration

change/@fluentui-react-native-tester-a0ed0eb4-ec38-4450-ae20-c0a1773ba265.json
samuelfreiberg: Got it, changing in newest commit

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
samuelfreiberg: You're right. Initially, I wanted to see if the regex worked and output all the contents of the array, but I forgot to change it back. However, I'm changing it to something a bit different... If one button matches the regex, we return true immediately.


In regards to your second statement, if I'm understanding correctly, we do already have that (to an extent). When clicking on each button, we wait for the page load and grab a UI element specific to that page to validate the correct content is being loaded.

ksiler: > If one button matches the regex, we return true immediately.

Why wouldn't you check all the buttons?

samuelfreiberg: It's a good question. But what would the success/failure condition be like?

If we want to validate every single component, then we have to hardcode a component list somewhere. And that list will have to be upkept when new controls are made, or name changes, etc.

I think the goal here is to simply check that the **UI Structure** is maintained. So not checking every individual component, just making sure no engineer changes the UI structure, because that would mess up native

ksiler: I see, that makes sense, and we could keep it as is, but shouldn't we also be checking that each button's automationID matches the expected format? I don't know if there's any other safeguards in place for that. I'm just wondering how we could catch those mistakes to ensure we don't unintentionally remove or exclude a test page from being tested if the format is wrong.

I understand

rurikoaraki (approved):

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
samuelfreiberg: I do think that's a good idea. That'll make it so no one changes those testID values as well. Changed in my newest commit

ksiler (approved):

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
PPatBoyd: might be desktop vs mobile but yep :)

apps/fluent-tester/src/E2E/common/NativeTesting/specs/NativeTesting.spec.win.ts
```diff
@@ -0,0 +1,28 @@
...
+
+describe('Native Safety Check Testing Initialization', () => {
+  it('Wait for app load', () => {
+    NavigateAppPage.waitForPageDisplayed(BOOT_APP_TIMEOUT);
+    expect(NavigateAppPage.isPageLoaded()).toBeTruthy(NavigateAppPage.ERRORMESSAGE_APPLOAD);
+  });
+
+  // The Scrollview in testing is the one that contains all the buttons that navigate to each component's
+  // test page. This is VITAL for our native testing, as we grab this Scrollview on the native side and
+  // to climb down the UI tree and find all of FURN's testable components.
```
PPatBoyd, line 13: nit the the previous nit about casing should've included this comment, whoops

PPatBoyd (approved):

PR #1475, opened 2022-02-25
Add native event info to onChange in Checkbox V1 and onClick for Button V1

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Changing the signature of the onChange and onClick callbacks to have an e (event) arg. This arg is one of types GestureResponderEvent, KeyEvent, or AccessibilityActionEvent, since we use this callback to respond to all 3 of click, key press, and accessibility invocations.

This allows users to get information about the events that come through and has the bonus of aligning our API to FluentUI JS better.

## Verification

Build passes, UTs pass. Tested on tester and ensured that event information is being filled in correctly.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
@@ -38,7 +38,8 @@ export const ButtonHOCTest: React.FunctionComponent = () => {
       {svgIconsEnabled && <CustomIconButton icon={{ svgSource: svgProps }}>Customized Icon Button</CustomIconButton>}
       <Button
         style={commonTestStyles.vmargin}
-        onClick={() => {
+        onClick={(e) => {
+          console.log(e.timeStamp);
           if (buttonRef.current) {
             buttonRef.current.focus();
           }
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import { Platform, View } from 'react-native';
 import { commonTestStyles, stackStyle } from '../Common/styles';
 import TestSvg from './test.svg';
+import { InteractionEvent, isGestureResponderEvent } from '@fluentui-react-native/interactive-hooks';
 
 const CustomText = Text.customize({ fontSize: 'header', color: 'hotpink' });
 const CustomButton = Button.customize({ backgroundColor: 'pink' });
@@ -38,8 +39,13 @@ export const ButtonHOCTest: React.FunctionComponent = () => {
       {svgIconsEnabled && <CustomIconButton icon={{ svgSource: svgProps }}>Customized Icon Button</CustomIconButton>}
       <Button
         style={commonTestStyles.vmargin}
-        onClick={(e) => {
+        onClick={(e: InteractionEvent) => {
           console.log(e.timeStamp);
+
+          if (isGestureResponderEvent(e)) {
+            console.log('I was clicked!');
+          }
+
           if (buttonRef.current) {
             buttonRef.current.focus();
           }
--- a/apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
@@ -3,12 +3,12 @@ import { EXPERIMENTAL_CHECKBOX_TESTPAGE } from './consts';
 import { Test, TestSection, PlatformStatus } from '../Test';
 import { Checkbox } from '@fluentui-react-native/experimental-checkbox';
 import { useTheme } from '@fluentui-react-native/theme-types';
-import { View, TextInput, TextStyle, AccessibilityActionEvent, GestureResponderEvent } from 'react-native';
+import { View, TextInput, TextStyle } from 'react-native';
 import { commonTestStyles as commonStyles } from '../Common/styles';
 import { E2ECheckboxExperimentalTest } from './E2ECheckboxExperimentalTest';
-import { KeyPressEvent } from '@fluentui-react-native/interactive-hooks';
+import { CallbackEvent } from '@fluentui-react-native/interactive-hooks';
 
-function onChangeUncontrolled(_e: GestureResponderEvent | KeyPressEvent | AccessibilityActionEvent, isChecked: boolean) {
+function onChangeUncontrolled(_e: CallbackEvent, isChecked: boolean) {
   console.log(isChecked);
 }
 
--- a/apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
@@ -6,9 +6,9 @@ import { useTheme } from '@fluentui-react-native/theme-types';
 import { View, TextInput, TextStyle } from 'react-native';
 import { commonTestStyles as commonStyles } from '../Common/styles';
 import { E2ECheckboxExperimentalTest } from './E2ECheckboxExperimentalTest';
-import { CallbackEvent } from '@fluentui-react-native/interactive-hooks';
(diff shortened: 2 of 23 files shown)


packages/utils/interactive-hooks/src/useAsToggleWithEvent.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { AccessibilityActionEvent, GestureResponderEvent } from 'react-native';
+import { KeyPressEvent } from '.';
+
+export type OnToggleWithEventCallback = (e: GestureResponderEvent | KeyPressEvent | AccessibilityActionEvent, value?: boolean) => void;
+export type OnChangeWithEventCallback = (e: GestureResponderEvent | KeyPressEvent | AccessibilityActionEvent) => void;
+
+/* Re-usable hook for toggle components.
```
rurikoaraki, line 8: I made a copy of this to be able to leave the version used in the old checkbox alone. If you have suggestions on how to handle this better, let me know.

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
```diff
@@ -38,7 +38,8 @@ export const ButtonHOCTest: React.FunctionComponent = () => {
       {svgIconsEnabled && <CustomIconButton icon={{ svgSource: svgProps }}>Customized Icon Button</CustomIconButton>}
       <Button
         style={commonTestStyles.vmargin}
-        onClick={() => {
+        onClick={(e) => {
+          console.log(e.timeStamp);
```
Saadnajmi, line 42: This _might_ be a windows only prop. I can try testing this on macOS and see what happens.

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -3,11 +3,12 @@ import { EXPERIMENTAL_CHECKBOX_TESTPAGE } from './consts';
...
 import { Checkbox } from '@fluentui-react-native/experimental-checkbox';
 import { useTheme } from '@fluentui-react-native/theme-types';
-import { View, TextInput, TextStyle } from 'react-native';
+import { View, TextInput, TextStyle, AccessibilityActionEvent, GestureResponderEvent } from 'react-native';
 import { commonTestStyles as commonStyles } from '../Common/styles';
 import { E2ECheckboxExperimentalTest } from './E2ECheckboxExperimentalTest';
+import { KeyPressEvent } from '@fluentui-react-native/interactive-hooks';
 
-function onChangeUncontrolled(isChecked: boolean) {
+function onChangeUncontrolled(_e: GestureResponderEvent | KeyPressEvent | AccessibilityActionEvent, isChecked: boolean) {
```
Saadnajmi, line 11: Do we want to define our own type for this intersection? Or leave it explicit and let React Native Core/windows/macOS deal with creating a type we can later use?

packages/components/Button/src/Button.types.ts
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import { ViewProps, ViewStyle, ColorValue } from 'react-native';
+import { ViewProps, ViewStyle, ColorValue, GestureResponderEvent, AccessibilityActionEvent } from 'react-native';
 import { TextProps } from '@fluentui-react-native/experimental-text';
 import { FontTokens, IBorderTokens, IColorTokens, IShadowTokens, LayoutTokens } from '@fluentui-react-native/tokens';
-import { IFocusable, IPressableHooks, IWithPressableOptions } from '@fluentui-react-native/interactive-hooks';
+import { IFocusable, IPressableHooks, IWithPressableOptions, KeyPressEvent } from '@fluentui-react-native/interactive-hooks';
```
Saadnajmi, line 5: It's probably good feedback to RN-macOS / windows that there should be a common "KeyPresEvent" type so that we can import it like the others from line 2

packages/utils/interactive-hooks/src/useAsToggleWithEvent.ts
Saadnajmi: Maybe once the old one is deprecated, we could rename this one to just `useAsToggle` ? `WithEvent` feels a bit unnecessary. I'm not sure my suggestion isn't a breaking change though.

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
rurikoaraki: I think leaving it explicit is better for clients to understand what kind of event info comes through, but copy/pasting this everywhere is pretty cumbersome... so yeah I think defining our own union type would be good. Let me figure out where that should live...

packages/utils/interactive-hooks/src/useAsToggleWithEvent.ts
rurikoaraki: I think we do technically ship this package so it could be breaking? I don't think it's used a lot outside of our repo though. Maybe it's fine as long as we bump the major version

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
Saadnajmi: I would imagine interactive hooks, but 🤷🏾‍♂️

apps/fluent-tester/src/FluentTester/TestComponents/Button/ButtonHOCTestSection.tsx
rurikoaraki: Ugh it's probably timestamp on other platforms and timeStamp on win32 or something like that. /shakesfist

rurikoaraki: Oh it's timeStamp on the outer event and timestamp on the inner native event.

apps/fluent-tester/src/FluentTester/TestComponents/CheckboxExperimental/CheckboxTest.tsx
rurikoaraki: Done in latest commits

packages/utils/interactive-hooks/src/events.types.ts
```diff
@@ -0,0 +1,4 @@
+import { AccessibilityActionEvent, GestureResponderEvent } from 'react-native';
+import { KeyPressEvent } from '.';
+
+export type CallbackEvent = GestureResponderEvent | KeyPressEvent | AccessibilityActionEvent;
```
rurikoaraki, line 4: Names are hard 😓if people have a better idea feel free to suggest it!

FalseLobster: One question I have is: do we have a recommended way for our users to narrow the type?

FalseLobster (approved):

packages/utils/interactive-hooks/src/events.types.ts
FalseLobster: Also, perhaps Input/IOEvent? It's slightly less generic than Callback

Saadnajmi: I think ultimately this should be the `PressEvent` type coming from Pressable. So this intermediate name is fine to me.

rurikoaraki: > One question I have is: do we have a recommended way for our users to narrow the type?

@FalseLobster Narrowing the type would follow whatever you would usually do via TypeScript. Something like `if (e instanceOf GestureResponderEvent)`. I can add an example in a test case.

> Also, perhaps Input/IOEvent? It's slightly less generic than Callback

I think "Input" has its own connotations (it makes me think of typing due to the TextInput RN control). Maybe `InteractionEvent`?

rurikoaraki: Well it turns out that since they're interfaces you can't use "instanceof". So I'm making type guard functions.

rurikoaraki: @FalseLobster Okay type guard functions + unit tests have been added

Saadnajmi (approved):

PR #1494, opened 2022-03-04
Bump eslint-plugin to 0.2.11 and fix hook-related errors

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

`@rnx-kit/eslint-plugin` enabled eslinting for react-hooks. This PR bumps the package to the latest version and fixes all linting errors.

I added a few ignore lines which should be fixed eventually:

- In the Navigation2DFocusZone of FocusZone test
- older version of TabItem component using useContext in the render function (this one might be okay but I'm confused about how context is being used in this control

### Verification

All UTs pass. Will so a pass over components like MenuButton and Tabs in tester.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
@@ -121,7 +121,7 @@ export const Slider: React.FunctionComponent<ISliderProps> = (props: ISliderProp
         setThumbLocation(initialThumbLocation);
       });
     }
-  }, [ref.current, initialValue, maximum, minimum]);
+  }, [initialValue, maximum, minimum]);
 
   return (
     <View style={[userStyle, styles.root]}>
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
@@ -121,7 +121,7 @@ export const Slider: React.FunctionComponent<ISliderProps> = (props: ISliderProp
         setThumbLocation(initialThumbLocation);
       });
     }
-  }, [initialValue, maximum, minimum]);
+  }, [ref, initialValue, maximum, minimum]);
 
   return (
     <View style={[userStyle, styles.root]}>
--- a/docs/pages/Theming/ColorTokens/UsageWithComponentTokens.md
+++ b/docs/pages/Theming/ColorTokens/UsageWithComponentTokens.md
@@ -34,7 +34,7 @@ export const Foo = compose<TabItemType>({
       ['color', ...fontStyles.keys],
     ),
   },
-  render: (userProps: TProps, useSlots: UseSlots<TType>) => {
+  useRender: (userProps: TProps, useSlots: UseSlots<TType>) => {
     // Grab the styled slots.
     const Slots = useSlots(userProps);
     // Return the handler to finish render.
--- a/packages/components/MenuButton/src/MenuButton.tsx
+++ b/packages/components/MenuButton/src/MenuButton.tsx
@@ -16,6 +16,7 @@ import {
   MenuButtonType,
   MenuButtonRenderData,
   MenuButtonState,
+  MenuButtonItemProps,
 } from './MenuButton.types';
 
 export const MenuButton = compose<MenuButtonType>({
@@ -33,30 +34,6 @@ export const MenuButton = compose<MenuButtonType>({
     const toggleShowContextualMenu = useCallback(() => {
       setShowContextualMenu(!showContextualMenu);
     }, [showContextualMenu, setShowContextualMenu]);
-
-    const menuItemsUpdated = menuItems.map((item) => {
-      if (item.hasSubmenu) {
-        const [showSubmenu, setShowSubmenu] = useState(false);
-
-        const toggleShowSubmenu = React.useCallback(() => {
-          setShowSubmenu(!showSubmenu);
-        }, [showSubmenu, setShowSubmenu]);
-
-        const onDismissSubmenu = React.useCallback(() => {
-          setShowSubmenu(false);
-        }, [setShowSubmenu]);
-        const { onHoverIn = toggleShowSubmenu, submenuProps = {}, ...restItems } = item;
-        const { onDismiss = onDismissSubmenu, setShowMenu = toggleShowSubmenu, ...restSubmenuProps } = submenuProps;
-        const menuItemUpdated = {
-          ...restItems,
-          onHoverIn,
-          showSubmenu: item.showSubmenu ? showSubmenu : undefined,
-          submenuProps: { ...restSubmenuProps, onDismiss, setShowMenu },
-        };
-        return menuItemUpdated;
-      }
-      return item;
-    });
 
     const state: MenuButtonState = {
       context: {
@@ -87,7 +64,7 @@ export const MenuButton = compose<MenuButtonType>({
         ...contextualMenu,
       },
       contextualMenuItems: {
-        menuItems: menuItemsUpdated,
+        menuItems,
       },
     });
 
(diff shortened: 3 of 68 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
```diff
@@ -121,7 +121,7 @@ export const Slider: React.FunctionComponent<ISliderProps> = (props: ISliderProp
         setThumbLocation(initialThumbLocation);
       });
     }
-  }, [ref.current, initialValue, maximum, minimum]);
```
Saadnajmi, line 124: Did eslint actually complain about this, or did you just notice it?

It makes sense to me to not memo on a ref, but just curious.

docs/pages/Theming/ColorTokens/UsageWithComponentTokens.md
```diff
@@ -34,7 +34,7 @@ export const Foo = compose<TabItemType>({
       ['color', ...fontStyles.keys],
     ),
   },
-  render: (userProps: TProps, useSlots: UseSlots<TType>) => {
+  useRender: (userProps: TProps, useSlots: UseSlots<TType>) => {
```
Saadnajmi, line 37: For my own understanding, this line defines a hook named `useRender` that is called inside the resulting components render? I had always assumed that this line created the _actual_ render function, not something used inside render.

packages/components/MenuButton/src/MenuButton.tsx
```diff
@@ -34,30 +35,6 @@ export const MenuButton = compose<MenuButtonType>({
...
-          ...restItems,
-          onHoverIn,
-          showSubmenu: item.showSubmenu ? showSubmenu : undefined,
-          submenuProps: { ...restSubmenuProps, onDismiss, setShowMenu },
-        };
-        return menuItemUpdated;
-      }
-      return item;
-    });
-
```
Saadnajmi, line 60: Why did this go away?

apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
rurikoaraki: Hm. I probably removed it because the linter said the useEffect wasn't depending on it, but now I see that it is. I'll take another look

packages/components/MenuButton/src/MenuButton.tsx
rurikoaraki: It's being handled by the Submenu component below on line 127.

The issue is that hooks cannot be called inside callbacks, so I needed to move all the hook code into its own component.

apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
rurikoaraki: The linter did complain I think saying that the ref wasn't referenced in the hook, but now that I look, it does, so I'll have to take another look at this.

docs/pages/Theming/ColorTokens/UsageWithComponentTokens.md
rurikoaraki: We call this function inside the component's render function. You can see it being called here:

https://github.com/microsoft/fluentui-react-native/blob/ae8453d3911de46140bd727559c787f56de67b67/packages/framework/composition/src/composeFactory.ts#L67

packages/components/MenuButton/src/MenuButton.tsx
rurikoaraki: This code is being handled by the SubMenu component on line 127. The issue is that hooks cannot be called directly inside of callbacks, so I needed to move the code inside a component in order for this to work.

Saadnajmi: OK. Locally testing this seems fine, so yay!

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
rurikoaraki: I think the issue is that it can't depend on a property of an object, just an object, so I opted to pass in ref instead and it's fine with that.

Saadnajmi: OK. FWIW, our sliders never worked to begin with, iirc.

packages/components/MenuButton/src/MenuButton.tsx
PPatBoyd: > looks like this logic was merged until the function component SubMenuItem below

zzzz I seriously detest the ```start review``` functionality of leaving everything pending and not updating any comments mid-review.

packages/components/MenuButton/src/MenuButton.tsx
```diff
@@ -156,4 +124,30 @@ export const MenuButton = compose<MenuButtonType>({
...
+
+  const { showSubmenu = showSubmenuState, submenuProps, componentRef, submenuItems, onHoverIn = toggleShowSubmenu, ...restItems } = props;
+  const { onDismiss = onDismissSubmenu, setShowMenu = toggleShowSubmenu, ...restSubmenuProps } = submenuProps;
+
+  return (
+    <React.Fragment>
+      <SubmenuItem componentRef={componentRef} onHoverIn={onHoverIn} {...restItems} />
+      {showSubmenu && (
+        <Submenu target={componentRef} onDismiss={onDismiss} setShowMenu={setShowMenu} {...restSubmenuProps}>
+          {submenuItems?.map((submenuItem) => (
```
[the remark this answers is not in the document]
rurikoaraki: It shouldn't unless something has changed - the "key" prop should make it so that React will "map" the generated items to what already exist and only replace the elements that have been changed.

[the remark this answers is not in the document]
rurikoaraki: https://reactjs.org/docs/lists-and-keys.html#keys

packages/components/Tabs/src/TabsItem.tsx
```diff
@@ -71,16 +71,20 @@ export const TabsItem = compose<TabsItemType>({
...
+     * We use the componentRef of the currently selected tabsItem to maintain the default tabbable
+     * element in Tabs. Since the componentRef isn't generated until after initial render,
+     * we must update it once here.
+     * Since this is meant to only be run once, surpressing lint error
+     */
     React.useEffect(() => {
       if (itemKey == info.selectedKey) {
         info.updateSelectedTabsItemRef && componentRef && info.updateSelectedTabsItemRef(componentRef);
       }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
rurikoaraki: Tbh I was just trying to preserve the current behavior 😓 Let me take a closer look at this.

apps/fluent-tester/src/FluentTester/TestComponents/Common/Slider.tsx
```diff
@@ -121,7 +121,7 @@ export const Slider: React.FunctionComponent<ISliderProps> = (props: ISliderProp
         setThumbLocation(initialThumbLocation);
```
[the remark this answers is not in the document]
rurikoaraki: It may have been one of those things to shut up typescript. The "!" after a variable tells typescript "I know this /can/ be undefined but I am telling you that we guarantee a valid value here." The type of minimum/maxium/initialValue is probably `umber | undefined` so TS would complain here without the "!". "!" doesn't actually enforce any value so if it/is/ undefined we'd probably crash or do something weird instead.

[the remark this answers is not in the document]
rurikoaraki: (Probably the more "proper" way is to instantiate a new variable of type number. We shouldn't be directly modifying props anyway.)

packages/components/Tabs/src/TabsItem.tsx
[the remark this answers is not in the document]
rurikoaraki: So I think this is just to set up the initial value. There's code to handle when it changes through the onFocus handler earlier in the file, which is why it's set to only run once. Could be cleaner to let this useEffect always run when info.selectedKey changes. but as it is it should work okay

PR #1500, opened 2022-03-07
Fixes Experimental Expanders WinUI Dependency

TatianaKapos:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x]  windows
- [ ] android

### Description of changes

Fixes Experimental Expanders WinUI dependency that breaks on RNW .67
Resolves https://github.com/microsoft/react-native-gallery/issues/183

### Verification
Tested locally, no tests currently for experimental expander

### DIFF
--- a/change/@fluentui-react-native-experimental-expander-5b655709-3184-408e-ba74-75534899d4e0.json
+++ b/change/@fluentui-react-native-experimental-expander-5b655709-3184-408e-ba74-75534899d4e0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fixes winUI dependency",
+  "packageName": "@fluentui-react-native/experimental-expander",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


Saadnajmi (approved):

HeyImChris (approved):

rurikoaraki:
@TatianaKapos please run `yarn change` and push the commit it generates to allow the PR build to pass. Thanks!

TatianaKapos:
@Saadnajmi Sorry for the late reply, took some time to look into the example app and get it running for windows!

But Expander already depended on WinUI 2.6, this change makes it so ExperimentalFeatures.props can decide which versions of the Windows SDK and WinUI 2.x to use. Interestingly this doesn't help the example app as I expected it to do, the error might not depend on versioning

rurikoaraki:
@Saadnajmi are we okay with merging this?

Saadnajmi:
> @Saadnajmi are we okay with merging this?

Yeah, I'm not sure if we ever got Expander working inside FURN itself. I think it had to be used in a standalone test app. So any improvement to the control is fine to me =)

PR #1521, opened 2022-03-18
ContextualMenu macOS: Improve lifecycle management

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We had a bug in ContextualMenu for macOS where you could end up in a weird state if you open/close/reopen a menu:
image

After... a LOT of debugging, I found out the solution was to make sure on the native side (`Callout.swift`) we dismiss our CalloutWindows' subwindows (submenus) before itself. Along the way, I made a lot of changes/improvements to the object lifecycle of Callout / CalloutWindow / their NSEvent monitors. I shall describe the changes below:

1.  NSEventMonitor improvements. 

We create an [NSEvent localMonitor](https://developer.apple.com/documentation/appkit/nsevent/1534971-addlocalmonitorforeventsmatching) to watch for clicks outside the CalloutWindow & it's children that should dismiss the Callout. The API is quite old, and if you read the fine print for [removeMonitor](https://developer.apple.com/documentation/appkit/nsevent/1533709-removemonitor), you **must** call `removeMonitor` only once per `addLocalMonitor` call, or else you can run into an over-release. This is true with ARC, and with Swift. Even doing an `if let monitor = mouseEventMonitor` check could cause a BAD_ACCESS crash if `monitor` was already released. 

To get around this, I created a private helper class `guardedEventMonitor` which keeps track of whether we allocated a monitor already or not, to prevent against an over release. It's pretty simple, but has worked well!

2. Move most CalloutWindow presentation management to CalloutView

We have a "proxy" CalloutView that owns a CalloutWindow, and forwards it's props / children to the window. As our API is currently written, as long as CalloutView is alive and allocated, CalloutWindow should be as well. Previously,  both CalloutView and CalloutWindow could dismiss CalloutWindow. CalloutView would listen for if JS wanted to dismiss the Callout, while CalloutWindow listened for mouse/keyboard/app-did-resign-active events that should dismiss the CalloutWindow. This got a little confusing for the flow of what methods / callbacks were invoked each time a window is dismissed. This led to bugs where the mouseEventMonitor would still be active and listening for mouse events, even if the Callout was dismissed. Worse, we could forget to remove them, and have new monitors allocated every time we opened a new Callout 😢 .

To simplify this, I moved as much of the Callout show/dismissal responsibility to the CalloutView as possible. This meant moving the NSEventMonitor from _CalloutWindow_ to _CalloutView_. We now also only allocate and remove the monitor while the window is shown, so that they don't lie around after we close the window. I also moved the `NSApplication.didResignActiveNotification` notification listener to CalloutView, and remove it in DismissCallout. 

The one thing I could not move was `cancelOperation` which listens for the "Escape" key press to dismiss the Callout. That has to stay in CalloutWindow.

3. Dismiss Children Callout windows first

This is the one that actually fixed the bug. If we dismissed the top level Callout _before_ the children, then the children CalloutViews/CalloutWindows would not de-allocate. This meant the child window (in our case a submenu) would stick around hidden, and not show up till the next time you opened the menu. 

The fix was simple: in `CalloutView.dismissCallout()`, dismiss all your children Callouts _before_ you dismiss yourself and voila! Everything is fixed! 😁 

Other minor fixes:

- Renamed displayName: submenuItem -> SubmenuItem
    - Every other display name was capitalized, and this affects what shows up in the React tree.  Felt like a simple typo
- Made onShow/onDismiss direct events
    - Now that I know the difference, these don't feel like they should be bubbling events. 
- Support more submenu props
    - Totally just forgot to do this earlier. More reason to not duplicate Menu/Submenu and MenuItem/SubmenuItem code..

### Verification

For the simple case of the bug: I can open/close submenus as much as I want and they don't show up on top of the main menu anymore!

https://user-images.githubusercontent.com/6722175/158916366-2199bfc8-cd70-42fd-97b7-f21e00015fc5.mov

For the complex case of lifecyle management, I put _ a lot_ of breakpoints  to trace order flow. I set them to print a message in console, and continue execution. The video below does a good job of showing the order flow. The important thing is: no step is hit twice or skipped, ensuring we don't have zombie event monitors or zombie windows laying around. 

https://user-images.githubusercontent.com/6722175/158916326-b8995532-8fd3-4679-b33f-c9bcf485e071.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/Callout.swift
+++ b/packages/components/Callout/macos/Callout.swift
@@ -361,20 +361,12 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		if let parentWindow = self.window {
 			parentWindow.addChildWindow(window, ordered: .above)
 		}
-		if (self.window as? CalloutWindow != nil) {
-			window.isSubwindow = true
-		}
-
 		return window
 	}()
 	
 	private var mouseEventMonitor = GuardedEventMonitor()
 	
 	private var isCalloutWindowShown = false
-	
-	deinit {
-		if (calloutWindow.isSubwindow) {}
-	}
 }
 
 /// React Native macOS uses a flipped coordinate space by default. Let's stay consistent and
--- a/packages/components/Callout/macos/Callout.swift
+++ b/packages/components/Callout/macos/Callout.swift
@@ -82,39 +82,39 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		guard !isCalloutWindowShown else {
 			return
 		}
-		
+
 		updateCalloutFrameToAnchor()
 		calloutWindow.makeKeyAndOrderFront(self)
 		
 		// Dismiss the Callout if the window is no longer active.
 		NotificationCenter.default.addObserver(self, selector: #selector(dismissCallout), name: NSApplication.didResignActiveNotification, object: nil)
-		
+
 		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { [weak self] (event) -> NSEvent? in
 			guard let clickedWindow = event.window else {
-   				return event
-   			}
-   
-   			var shouldDismissCallout = true
-   
-   			// Did the click happened in our own window?
-   			if (clickedWindow.isEqual(to: self) ) {
-   				shouldDismissCallout = false
-   
-   			// Are we a child window of a Callout (e.g: a ContextualMenu submenu), where the click happened in our parent?
-   			} else if (clickedWindow.isEqual(to: self?.calloutWindow.parent as? CalloutWindow)) {
-   				shouldDismissCallout = false
-   
-   			// Did the click happened in any of our child windows (e.g: our submenus)?
-   			} else if (self?.calloutWindow.childWindows?.contains(clickedWindow) ?? false) {
-   				shouldDismissCallout = false
-   			}
-   
-   			if (shouldDismissCallout) {
-   				self?.dismissCallout()
-   			}
-   
-   			return event
-   		})
+					return event
+				}
+
+				var shouldDismissCallout = true
+
+				// Did the click happened in our own window?
+				if (clickedWindow.isEqual(to: self) ) {
+					shouldDismissCallout = false
+	
+				// Are we a child window of a Callout (e.g: a ContextualMenu submenu), where the click happened in our parent?
+				} else if (clickedWindow.isEqual(to: self?.calloutWindow.parent as? CalloutWindow)) {
+					shouldDismissCallout = false
+
+				// Did the click happened in any of our child windows (e.g: our submenus)?
+				} else if (self?.calloutWindow.childWindows?.contains(clickedWindow) ?? false) {
+					shouldDismissCallout = false
+				}
+	
+				if (shouldDismissCallout) {
+					self?.dismissCallout()
+				}
+
+				return event
+			})
 		
 		isCalloutWindowShown = true
 		
@@ -369,39 +369,6 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 	private var isCalloutWindowShown = false
 }
 
-/// React Native macOS uses a flipped coordinate space by default. Let's stay consistent and
-/// ensure any views hosting React Native views are also flipped. This helps RCTTouchHandler
-/// register clicks in the right location.
-private class FlippedVisualEffectView: NSVisualEffectView {
-	override var isFlipped: Bool {
-		return true
-	}
-}
 
-
-/// NSEvent localMonitors are an _old_ API that requires you to do some  memory management yourself.
-/// You _must_ call `removeMonitor`exactly once per monitor created from. This helper class enforces
-/// that with the use of an extra boolean to track whether we have already allocated/released our monitor.
-/// https://developer.apple.com/documentation/appkit/nsevent/1533709-removemonitor
-private class GuardedEventMonitor: Any {
-	
-	private var isMonitorAllocated = false
-	private var allocatedMonitor: Any?
-	
-	
-	func addLocalMonitorForEvents(matching mask: NSEvent.EventTypeMask, handler block: @escaping (NSEvent) -> NSEvent?) {
-		if (!isMonitorAllocated) {
-			allocatedMonitor = NSEvent.addLocalMonitorForEvents(matching: mask, handler: block)
-			isMonitorAllocated = true
-		}
-	}
-	
-	func removeMonitor() {
-		if (isMonitorAllocated) {
-			NSEvent.removeMonitor(allocatedMonitor as Any)
-			isMonitorAllocated = false
-		}
-	}
-}
 
 private var calloutWindowCornerRadius: CGFloat = 5.0
--- a/packages/components/Callout/macos/Callout.swift
+++ b/packages/components/Callout/macos/Callout.swift
@@ -26,9 +26,9 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
-	@objc public var onShow: RCTBubblingEventBlock?
+	@objc public var onShow: RCTDirectEventBlock?
 
-	@objc public var onDismiss: RCTBubblingEventBlock?
+	@objc public var onDismiss: RCTDirectEventBlock?
 
 	public weak var bridge: RCTBridge?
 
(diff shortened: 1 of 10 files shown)


packages/components/Callout/macos/Callout.swift
```diff
@@ -79,14 +79,68 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
+		
+		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { [weak self] (event) -> NSEvent? in
+			guard let clickedWindow = event.window else {
+   				return event
+   			}
+   
+   			var shouldDismissCallout = true
+   
+   			// Did the click happened in our own window?
+   			if (clickedWindow.isEqual(to: self) ) {
```
harrieshin, line 100: hmmm if all we are doing is setting shoulddismisscallout maybe we should have OR operators in one if statements? amI missing something?

packages/components/Callout/macos/Callout.swift
```diff
@@ -321,4 +378,30 @@ private class FlippedVisualEffectView: NSVisualEffectView {
 	}
 }
 
+
+/// NSEvent localMonitors are an _old_ API that requires you to do some  memory management yourself.
+/// You _must_ call `removeMonitor`exactly once per monitor created from. This helper class enforces
+/// that with the use of an extra boolean to track whether we have already allocated/released our monitor.
+/// https://developer.apple.com/documentation/appkit/nsevent/1533709-removemonitor
+private class GuardedEventMonitor: Any {
```
harrieshin, line 386: nit should we prefix with FRN or something just in case?

packages/components/Callout/macos/Callout.swift
```diff
@@ -321,4 +378,30 @@ private class FlippedVisualEffectView: NSVisualEffectView {
...
+	
+	
+	func addLocalMonitorForEvents(matching mask: NSEvent.EventTypeMask, handler block: @escaping (NSEvent) -> NSEvent?) {
+		if (!isMonitorAllocated) {
+			allocatedMonitor = NSEvent.addLocalMonitorForEvents(matching: mask, handler: block)
+			isMonitorAllocated = true
+		}
+	}
+	
+	func removeMonitor() {
```
harrieshin, line 399: do we need to call this removeMonitor in deinit() as well?

packages/components/Callout/macos/Callout.swift
harrieshin: another nit: can we create a new file for this private class?

Saadnajmi: This is Swift, so we don't have that namespace clashing consideration to deal with.

packages/components/ContextualMenu/src/SubmenuItem.types.ts
```diff
@@ -6,7 +6,7 @@ import { ContextualMenuItemProps, ContextualMenuItemTokens, ContextualMenuItemSt
 import { IconProps } from '@fluentui-react-native/icon';
 import { XmlProps } from 'react-native-svg';
 
-export const submenuItemName = 'submenuItem';
+export const submenuItemName = 'SubmenuItem';
```
harrieshin, line 9: might as well camelcase the const var  as well?

packages/components/Callout/macos/Callout.swift
Saadnajmi: In one of my implementations while debugging, I had a version of that. Theoretically it should be safe with my guarding. I can try it out and see what happens.

packages/components/Callout/macos/Callout.swift
Saadnajmi: I think I did it this way for readability. It's easier for me to read these `else if`'s as cases, and easier for me to add more cases in the future (for say, L2 submenus which are currently not supported..)

packages/components/Callout/macos/Callout.swift
harrieshin: with the zombie or memory leak profiling should help unittest it?

packages/components/Callout/macos/Callout.swift
Saadnajmi: Would we want another file for the FlippedVisualEffectView above as well?

packages/components/ContextualMenu/src/SubmenuItem.types.ts
Saadnajmi: I'm not sure what the convention is. It seems the var name has been camelCase, but the string value should be UpperCamelCase because this name is special / used in the React tree where everything else is UpperCamelCase

packages/components/Callout/macos/Callout.swift
```diff
@@ -321,4 +378,30 @@ private class FlippedVisualEffectView: NSVisualEffectView {
...
 }
 
+
+/// NSEvent localMonitors are an _old_ API that requires you to do some  memory management yourself.
+/// You _must_ call `removeMonitor`exactly once per monitor created from. This helper class enforces
+/// that with the use of an extra boolean to track whether we have already allocated/released our monitor.
+/// https://developer.apple.com/documentation/appkit/nsevent/1533709-removemonitor
+private class GuardedEventMonitor: Any {
+	
+	private var isMonitorAllocated = false
```
harrieshin, line 388: and we believe boolean is good rather than an actual counter because one callout view should only have monitor...so it should always be +1 -1 ...?

Saadnajmi:
> super nit: your description of pr for number 1 is great. lets move it to apple ux guide?

@harrieshin I don't know if my helper class works for the general case. It works if you only allocate one monitor, not several. `NSEvent.h` also mentions this line in the API description:

```...    In either case, the return value of the API is a retained NSObject. You typically do not need to retain and release the event monitor yourself, since the implementation will retain it while needed. ```

So it sounds like they _should_ be destroyed, even though I just tested it and found it lived past the deinit of CalloutView.

To put this in aka.ms/apple-ux-guide , I think I would want to create a page called "Event Monitoring", explain you should first try overriding `NSResponder` methods like `mouseDown / keyDown / becomeFirstResponder`, and only use localMonitor if the previous suggestions don't fit your needs. Only then, would you run into the case where my helper class is useful.

File: packages/components/Callout/macos/Callout.swift (hunk 5 above)
Comment (Saadnajmi) [reply packages/components/Callout/macos/Callout.swift:388]: Yes. Multiple monitors would mean multiple blocks executed, which seems bad. I also thought I read somewhere if you call `addLocalMonitorForEvents(matching mask: handler:)` with the same mask twice, you get the same monitor twice. I haven't been able to prove that in code though..

In another case, maybe there are two different types of NSEvents you're listening for (mouseDown, keyDown ?) at which point you could maybe have two monitors. We don't need that for our case.

File: packages/components/Callout/macos/Callout.swift (hunk 3 above)
Comment (Saadnajmi) [reply packages/components/Callout/macos/Callout.swift]: The zombie tool is telling me that no `guardedEventMonitors` are sticking around after I close a menu. Now that `removeMonitor` is called in deinit, I'm pretty confident no underlying NSEventMonitors are sticking around either.

**Review by harrieshin — APPROVED** [review]

File: packages/components/ContextualMenu/src/SubmenuItem.types.ts (hunk 4 above)
Comment (Saadnajmi) [reply packages/components/ContextualMenu/src/SubmenuItem.types.ts:9]: I think we are following proper convention here. vars are lower case, types/components are uppercase

File: packages/components/Callout/macos/GuardedEventMonitor.swift [hunk 6]
```diff
@@ -0,0 +1,30 @@
+
+
+/// NSEvent localMonitors are an _old_ API that requires you to do some  memory management yourself.
```
anandrajeswaran, line 3: What's the gap in the NSEvent API? I think you are always fine if you just nil out your reference to the monitor after you remove it, right? At least that's what I've done elsewhere.

And this wrapper does provide some potential for confusion and misuse - `addLocalMonitorForEvents`  no-ops if there already is a monitor, even if you send in a different mask.

Saadnajmi: Yep. Both of these are things I realized / people pointed out to me right after I merged. I didn't realize you could just nil out the reference and that would be sufficient. I rewrote the wrapper to do so in https://github.com/microsoft/fluentui-react-native/commit/b772dfe4542b7032f216bf5c2faa807c695fa2a7

There I added an extra comment that this internal class is only ever meant to be used for 1 monitor at a time, and I changed it so that we just nil out the monitor rather than the boolean check.


`addLocalMonitor` will never let you make more than 1 monitor per application? That's interesting...

anandrajeswaran: Sorry I meant that your wrapper function `addLocalMonitorForEvents` (not the system API) was the one that is confusing in that it would do nothing if you called it with a different mask because of the `if (!isMonitorAllocated)` check. Personally I would tend towards not having the wrapper. I think nil'ing out after removing is a pretty common pattern and it seems better to force people to remember that pattern than add our own abstraction on top that has other gotchas/patterns people would need to familiarize themselves with.

PR #1523, opened 2022-03-18
Disable focus on RadioButton when disabled

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There's a native change on win32 I'm working on that will prevent accessibilityState.disabled=true from disabling the control and instead only report the control as disabled from an accessibility perspective. As a result of this change, I noticed there was a bug on RadioButton which allowed focus to be placed on the control when it was marked as disabled. Other controls (like Button and Checkbox) mark focusable = false when disabled is true.

### Verification

Manually tested.

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


chiuam (approved):

harrieshin:
isn't there a value for screen reader focus to focus on disable item to read its text and state for visual impair users?

Saadnajmi:
> isn't there a value for screen reader focus to focus on disable item to read its text and state for visual impair users?

I'm not sure if windows screen readers are supposed to read disabled items, but I ran into a similar issue with macOS and menu items. We wanted disabled menu items to be not focusable, but read by screen reader. I found a simple enough fix and special cased the behavior for macOS. If we want this for other controls, we could use the same pattern/hook I have in this PR: https://github.com/microsoft/fluentui-react-native/pull/1435

rurikoaraki:
> > isn't there a value for screen reader focus to focus on disable item to read its text and state for visual impair users?
> 
> I'm not sure if windows screen readers are supposed to read disabled items, but I ran into a similar issue with macOS and menu items. We wanted disabled menu items to be not focusable, but read by screen reader. I found a simple enough fix and special cased the behavior for macOS. If we want this for other controls, we could use the same pattern/hook I have in this PR: #1435 (Could be a followup from our side after this PR)

@harrieshin @Saadnajmi On Windows there are ways to get screen readers to read disabled items even if they aren't tabbable/focusable. Focusable is a separate setting from accessible.

rurikoaraki (approved):

Saadnajmi:
> > > isn't there a value for screen reader focus to focus on disable item to read its text and state for visual impair users?
> > 
> > I'm not sure if windows screen readers are supposed to read disabled items, but I ran into a similar issue with macOS and menu items. We wanted disabled menu items to be not focusable, but read by screen reader. I found a simple enough fix and special cased the behavior for macOS. If we want this for other controls, we could use the same pattern/hook I have in this PR: #1435 (Could be a followup from our side after this PR)
> 
> @harrieshin @Saadnajmi On Windows there are ways to get screen readers to read disabled items even if they aren't tabbable/focusable. Focusable is a separate setting from accessible.

Ok. They might _also_ be different things on macOS (I only recently realized the voiceover cursor/screen reader ring is different from the keyboard focus ring). I'll double check that.

chiuam:
> > > > isn't there a value for screen reader focus to focus on disable item to read its text and state for visual impair users?
> > > 
> > > 
> > > I'm not sure if windows screen readers are supposed to read disabled items, but I ran into a similar issue with macOS and menu items. We wanted disabled menu items to be not focusable, but read by screen reader. I found a simple enough fix and special cased the behavior for macOS. If we want this for other controls, we could use the same pattern/hook I have in this PR: #1435 (Could be a followup from our side after this PR)
> > 
> > 
> > @harrieshin @Saadnajmi On Windows there are ways to get screen readers to read disabled items even if they aren't tabbable/focusable. Focusable is a separate setting from accessible.
> 
> Ok. They might _also_ be different things on macOS (I only recently realized the voiceover cursor/screen reader ring is different from the keyboard focus ring). I'll double check that.

@Saadnajmi I think they are. You can synchronize / desynchronize keyboard focus and VoiceOver cursor in VoiceOver utility:
image

Saadnajmi:
> 

OK, so trying that out, I see that on macOS, the voiceover cursor doesn't depend on `focusable`, unless you have that option @chiuam circled selected. This tells me that my fix in #1435 was wrong and how @ksiler is doing it here is correct. I'll revert my macOS specific logic in ContextualMenu in my next PR. Thanks everyone!  @harrieshin fyi

Saadnajmi (approved):

PR #1526, opened 2022-03-19
ContextualMenu: more fixes

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

A few unrelated changes to ContextualMenu:

1. (macOS only) Support any number of submenus. 

We have native code that detects clicks outside the CalloutWindow, and determines if we should dismiss the Callout. We did a special case if we were a submenu and we clicked on our parent menu, and vice versa. This special case didn't work if we had more levels of submenus. I simplified the "should click dismiss callout" code a bunch so that it just recursively checks if the click happened in _any_ parent or child Callout. This means that we now support any number of submenus.

2.(macOS only) Don't set focusable on menuitem/submenuitem based on screenreader status3. 

It turns out VoiceOver on macOS doesn't check if a view is focusable before reading it out. This is because the VoiceOver cursor is different from the keyboard focus ring. This reverts my earlier change in #1435 .

3. (macOS only) Simplify GuardedEventMonitor a bit4. 

Turns out we don't need the extra boolean. We just need to set our monitor ivar to nil after we remove it. 🤷🏾‍♂️.

4. Setting maxWdith on ContextualMenu sets maxWidth, not width.

This seemed like a typo. Not sure why were weren't already doing this.

5. Added more menu items to the "ContextualMenu Scrollview" test

Menu Items are smaller on macOS.. so we didn't have enough to fill the max height and cause a scrollbar to appear. Adding more menu items should just fix that.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/Callout.swift
+++ b/packages/components/Callout/macos/Callout.swift
@@ -85,39 +85,47 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 
 		updateCalloutFrameToAnchor()
 		calloutWindow.makeKeyAndOrderFront(self)
-		
+
 		// Dismiss the Callout if the window is no longer active.
 		NotificationCenter.default.addObserver(self, selector: #selector(dismissCallout), name: NSApplication.didResignActiveNotification, object: nil)
 
-		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { [weak self] (event) -> NSEvent? in
-			guard let clickedWindow = event.window else {
-					return event
+		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseDown, handler: { [weak self] (event) -> NSEvent? in
+			func isClickInsideWindowHeirarchy(window: NSWindow?, event: NSEvent) -> Bool {
+				guard let window = window else {
+					return false
+				}
+				guard let clickedWindow = event.window else {
+					return false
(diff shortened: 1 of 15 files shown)


packages/components/Callout/macos/Callout.swift
```diff
@@ -85,39 +85,47 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
-		
+
 		// Dismiss the Callout if the window is no longer active.
 		NotificationCenter.default.addObserver(self, selector: #selector(dismissCallout), name: NSApplication.didResignActiveNotification, object: nil)
 
-		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseUp, handler: { [weak self] (event) -> NSEvent? in
-			guard let clickedWindow = event.window else {
-					return event
+		mouseEventMonitor.addLocalMonitorForEvents(matching: .leftMouseDown, handler: { [weak self] (event) -> NSEvent? in
+			func isClickInsideWindowHeirarchy(window: NSWindow?, event: NSEvent) -> Bool {
```
harrieshin, line 93: misspelled hierarchy

harrieshin (approved):

PR #1534, opened 2022-03-23
Ignore 429 error from markdown-link-check

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Getting annoyed that PRs fail when we hit 429s (too many requests) which isn't something we can really control. Setting markdown-link-check to ignore 429 error by just telling it that it's a success code.

Also converting a bunch of our links to not use direct urls to minimize this issue.

### Verification

CI should pass

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Do we still need to include 429 as an alive status code if we converted everything to be relative URLS?
I kinda wanna leave a comment saying why we include 429 as "alive" but I don't know if that's possible in that JSON

Saadnajmi (approved):

Saadnajmi:
Alternatively / in addition to this fix, I think it's possible to make the Markdown CI check optional? I think you need admin access to the repo to do so though.

rurikoaraki:
> Do we still need to include 429 as an alive status code if we converted everything to be relative URLS? I kinda wanna leave a comment saying why we include 429 as "alive" but I don't know if that's possible in that JSON

Yeah, because we have other URLs that aren't part of the repo in our documentation, so those can potentially fail.

My experience with optional checks is that we don't notice failures because GH will show it as a green checkmark even if there's a failure, so I'd rather not do that. Unless previous optional failures were done in a different way.

PR #1535, opened 2022-03-23
Combine platform specific test apps (except win32)

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [x] windows
- [x] android

### Description of changes

We have separate test apps for each platform, with a shared JS package as source.
```
apps
├── fluent-tester (shared JS)
├── android
├── ios
├── macos
└── windows
```
This is a weird structure that nobody else has. It's a relic of when I set up the test apps and was still learning React Native. 
Most just put everything in one package: 
```
apps
├── fluent-tester (shared JS + platform native code)
```

Now that iOS, Android, macOS, and Windows are all managed by [react-native-test-app](https://github.com/microsoft/react-native-test-app), let's just combine them. There are a lot of advantages of maintainability with having one package:
- Less ability to have duplicate dependencies (like separate versions of React Native)
- Easier to update with React Native version upgrades
- Less code overall

### Verification

Locally, I can run the Android, iOS, macOS, and windows test apps. All the CI is passing, and E2E tests should be running for windows  & win32.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Note to self before check-in:

- Verify our `react-native.config.js` file. There's a difference between what we had that heavily used rnx-kit, and what we have in this PR which is mostly coming from RNTA.
- The Apple PR step "Pre-boot simulator" doesn't run on iOS. Figure out why.

tido64:
> * iOS and macOS (still) have a problem with autolinking, so I still need to manaually specify each podspec in the test app's podfile. We should file a bug to track fixing this (Especially if it affects downstream users).

I think you might have to add declare all native modules manually like here: https://github.com/react-native-webview/react-native-webview/blob/a7d2c66542eedef8a45dc9d0adee47178c30201c/react-native.config.js#L13-LL18. Autolinking currently does not support monorepos.

> * Android needs the `bundle:android` step, as far as I can tell. `Gradle` doesn't like it when the bundle / resources are missing.

Do you have more details here? It shouldn't be necessary to run `bundle:android` first.

tido64:
> That seems to contradict the README for auto linking that says auto linking in a mono repo is supported. https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-use-autolinking-in-a-monorepo . I'll try that out, I prefer that solution to manually linking the pods in our podfile.

From personal experience, I've had to add a `dependencies` section to help autolinking in [AsyncStorage](https://github.com/react-native-async-storage/async-storage/blob/master/react-native.config.js#L11), [NetInfo](https://github.com/react-native-netinfo/react-native-netinfo/blob/564a8e257f12194140ecbb6f1053d53003b669cf/react-native.config.js#L23), and [WebView](https://github.com/react-native-webview/react-native-webview/blob/master/react-native.config.js#L14). Might be worth investigating at some point.

Saadnajmi:
> WebView

A dumb question: Would we add a `react-native.config.js` file to every package in FURN with a native module? Would we also want to put a path to fluent-tester in the `projects` section of the config? I'll play around with it when I get a chance and see how much is necessary.

tido64:
> > WebView
> 
> A dumb question: Would we add a `react-native.config.js` file to every package in FURN with a native module? Would we also want to put a path to fluent-tester in the `projects` section of the config? I'll play around with it when I get a chance and see how much is necessary.

If autolinking works outside this repo, I think it's only needed by the test apps.

Saadnajmi:
Latest round of updates

Fixed:
1) Issue with SVGs not working fixed by updating our metro config & adding back the svg-transformer. 
2) Issue with Android and Windows not building solved by removing Drawer & Expander tests

Open issues:
1) Autolinking for Apple still doesn't work. I tried manually specifying dependencies in each packages' and in fluent-testers' `react-native.config.js` to no avail. I'll keep my current workaround of adding the pods manually and perhaps open an upstream issue later.
2) `yarn iOS/macOS` doesn't start the packager anymore. We had a custom build phase to do that before that I removed. This seems to be causing issues when you first boot the app / switch "Load JS Bundle -> Load from dev server". The fix will probably be to add back the custom build phase & maybe open a feature request to RNTA / the CLI to auto-start the packager.
3) Windows E2E tests are failing. We narrowed it down to an exception in native RNW code no longer being caught, so the app crashes instead of proceeding gracefully. I'm not sure why moving the  test app changes native exception handling behavior.. so 🤷🏾‍♂️.

tido64 (commented):
On Windows, I'm seeing two issues. One is:

Screenshot 2022-07-22 at 11 36 12

Which should be fixed with:

```diff
diff --git a/apps/fluent-tester/src/TestComponents/Tokens/TokenTest.tsx b/apps/fluent-tester/src/TestComponents/Tokens/TokenTest.tsx
index 5267d9050..b41844bb9 100644
--- a/apps/fluent-tester/src/TestComponents/Tokens/TokenTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Tokens/TokenTest.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { FlatList, View, ViewStyle, StyleSheet, ColorValue } from 'react-native';
+import { ColorValue, ScrollView, StyleSheet, View, ViewStyle } from 'react-native';
 import { useTheme, Theme } from '@fluentui-react-native/theme-types';
 import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
 import { getCurrentAppearance } from '@fluentui-react-native/theming-utils';
@@ -88,7 +88,7 @@ const AliasTokensSwatchList: React.FunctionComponent = () => {
     <View style={[commonTestStyles.view]}>
       <Text>Alias Color Tokens from Token Pipeline</Text>
       <View style={themedStyles.stackStyle}>
-        <FlatList data={aliasTokensAsArray} renderItem={renderSwatch} />
+        <ScrollView>{aliasTokensAsArray.map((item) => renderSwatch({ item }))}</ScrollView>
       </View>
     </View>
   );
```

The other is a crash if you click on anything in the sidebar. This is what's causing the Windows tests to fail. I tried to debug in Visual Studio and I see that a `null` command is passed to a node:

Screenshot 2022-07-22 at 12 07 13

I don't know why this starts to happen now though. We should ask someone with more intimate knowledge of `react-native-windows`.

.ado/azure-pipelines.yml
```diff
@@ -36,43 +36,39 @@ jobs:
...
         displayName: 'verify API and Ensure Changed Files'
 
-  # Windows bundling and end to end testing
-  - job: WindowsPR
-    displayName: Windows PR
+  - job: AndroidPR
+    displayName: Android PR
     pool:
-      vmImage: 'windows-2019'
+      vmImage: 'macos-11'
```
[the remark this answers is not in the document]
tido64: From [react-native-test-app](https://github.com/microsoft/react-native-test-app/actions/runs/2717542212):
image

harrieshin, line 42: why we using macos11? so old and it might not have all the latest security stuff.

Saadnajmi: This is probably a relic from pre EO work. Can update.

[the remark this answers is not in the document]
Saadnajmi: Good catch! Let's save the macOS VMs for apple builds :)

Saadnajmi: resolved!

.ado/azure-pipelines.yml
```diff
@@ -36,43 +36,39 @@ jobs:
           yarn check-for-changed-files
         displayName: 'verify API and Ensure Changed Files'
 
-  # Windows bundling and end to end testing
-  - job: WindowsPR
-    displayName: Windows PR
+  - job: AndroidPR
+    displayName: Android PR
     pool:
       vmImage: 'windows-2019'
```
Saadnajmi, line 42:
```suggestion
       vmImage: 'macos-11'
```

.ado/azure-pipelines.yml
Saadnajmi: Turns out the Android CI hangs without this. Bringing this back and I'll file an issue for later :/

tido64 (approved):
Awesome job 🎉

harrieshin (approved):

mischreiber (approved):

rurikoaraki (approved):

PR #1544, opened 2022-03-28
Update Appium to 2.0.0

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Upgrade Appium and @appium/mac2-driver to latest versions. This will update some very out of date packages and resolve some security issues.

# Verification

CI

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
I wonder if the integration with WebDriverIO will change (in addition to the drivers that Appium uses).

Is there any blog posts or docs that outline the changes with WebDriverIO?

rurikoaraki:
Looks like Appium is recommending usage of WebDriverIO. But wdio itself hasn't updated to use/test with the latest Appium...

I actually have it running (though failing) locally, so I think we'll be okay without having to update wdio at the same time.

.ado/templates/e2e-testing-win32.yml
```diff
@@ -1,34 +1,39 @@
 steps:
+  - script: |
+      yarn bundle
+    workingDirectory: apps/win32
+    displayName: 'yarn bundle'
+
+  - script: |
```
samuelfreiberg, line 7: Is this in additional to WinAppDriver?

package.json
```diff
@@ -77,6 +77,7 @@
     "metro-symbolicate": "^0.66.2",
     "metro-transform-plugins": "^0.66.2",
     "metro-transform-worker": "^0.66.2",
+    "plist": "^3.0.5",
```
samuelfreiberg, line 80: What's this for?

samuelfreiberg:
Can you share any links you used to help migrate to Appium V2? I'd appreciate learning a bit more

package.json
rurikoaraki: The latest version of Appium pulls in plist 3.0.4 which has a security issue. Since their dependency syntax specifically asks for 3.0.4, the package with the security issue will be pulled in even though there's a version with the fix out. This line just says "if someone asks for plist force them to resolve to version ^3.0.5

.ado/templates/e2e-testing-win32.yml
rurikoaraki: This installs WinAppDriver (and probably some things which appium uses to talk to WinAppDriver). 2.0 separated the drivers out from the main package to subpackages and you need to explicitly ask appium to install them to run drivers on a particular endpoint. https://github.com/appium/appium/blob/2.0/docs/en/advanced-concepts/migrating-to-appium-2.0.md#bangbang-installing-drivers-during-setup for more info

rurikoaraki:
> Can you share any links you used to help migrate to Appium V2? I'd appreciate learning a bit more

https://github.com/appium/appium/blob/2.0/docs/en/advanced-concepts/migrating-to-appium-2.0.md

was the main page I used for migration specifically

.ado/templates/e2e-testing-win32.yml
Saadnajmi: Sounds like macOS will need to install the `xcuitest` driver too. Since we don't have E2E tests in CI, that's not blocking here, but worth calling out. Maybe we need to update our documentation on how to locally run E2E tests now?

apps/fluent-tester/package.json
```diff
@@ -73,8 +73,8 @@
     "@wdio/local-runner": "^6.5.0",
     "@wdio/spec-reporter": "^6.5.0",
     "@wdio/sync": "^6.5.0",
-    "appium": "1.20.0",
-    "appium-mac2-driver": "0.14.1",
+    "appium": "^2.0.0-beta.27",
```
Saadnajmi, line 76: Does "^" work when there's a "-beta" in the name?

apps/win32/wdio.conf.js
```diff
@@ -1,6 +1,5 @@
 const path = require('path');
 const fs = require('fs');
-const rimraf = require('rimraf');
```
Saadnajmi, line 3: Do we not need rimraf at all anymore?

package.json
```diff
@@ -24,7 +24,7 @@
     "change": "beachball change",
     "check-for-changed-files": "cd scripts && yarn fluentui-scripts checkForModifiedFiles",
     "checkchange": "beachball check --changehint \"Run 'yarn change' to generate a change file\"",
-    "depcheck": "rnx-dep-check && lage depcheck",
+    "depcheck": "lage depcheck",
```
Saadnajmi, line 27: This seems like a regresssion?

package.json
Saadnajmi: This is on top of the need for simple-plist? Dang..

package.json
rurikoaraki: ugh I keep doing this sorry. I have to delete it in order to get depcheck working locally.

apps/win32/wdio.conf.js
rurikoaraki: Seems like it's used to run a command on package.json

package.json
Saadnajmi: Curious, what breaks locally? We could also update our lage config / scripts to just... call rnx-dep-check separately

rurikoaraki: It fails partway and tells us we should be using different versions of react things. Which is fine, but it stops the rest of the command from running when running locally so it doesn't get to the lage depcheck part.

apps/fluent-tester/package.json
rurikoaraki: From https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004:

>^1.2.3-beta.2 := >=1.2.3-beta.2 <2.0.0 Note that prereleases in the 1.2.3 version will be allowed, if they are greater than or equal to beta.2. So, 1.2.3-beta.4 would be allowed, but 1.2.4-beta.2 would not, because it is a prerelease of a different [major, minor, patch] tuple.

So seems like it'll work as long as the tuple doesn't change

package.json
ksiler: Just wondering - is there a reason why this is listed in the root package.json versus the package.json under apps/win32 and apps/windows? Does "resolutions" only work in the root package.json?

rurikoaraki: Nothing in the docs says that it won't work outside of the root package.json, but my guess was that it was put here since then you only have to specify it once to have it apply to the other packages.

.ado/templates/e2e-testing-win32.yml
rurikoaraki: Ah yeah good call, updated the Windows directions.

apps/fluent-tester/src/E2E/README.md
```diff
@@ -39,7 +39,9 @@ and drag & drop the **XCode Helper** app to **Security & Privacy -> Privacy -> A
 2. Bundle the test app (pick specific platform you want to test, we'll use win32)
    - C:\repo\fluentui-react-native>`cd apps\win32`
    - C:\repo\fluentui-react-native\apps\win32> `yarn bundle`
-3. Run E2E tests
+3. Hook up appium to the WinAppDriver
+   - C:\repo\fluentui-react-native\apps\win32> `yarn appium driver install windows`
```
Saadnajmi, line 43: You only have to do this once, right? Then it's installed on your system for good?

rurikoaraki: I think as long as you don't nuke your node_modules you should be okay

rurikoaraki:
@samuelfreiberg @ksiler Any additional comments? Otherwise I'll merge EOD.

.ado/templates/e2e-testing-win32.yml
```diff
@@ -1,34 +1,39 @@
...
+      yarn appium driver install windows@^2.0.1
+    workingDirectory: apps/win32
+    displayName: 'yarn appium'
```
samuelfreiberg, line 10: Same comment as above

samuelfreiberg (approved):

ksiler (approved):

PR #1551, opened 2022-03-30
chore: enable no-export-all on index files

tido64:
Some bundlers, such as esbuild, are not able to tree-shake nested `export *` statements. Fixed them using the `no-export-all` ESLint rule.

### Platforms Impacted

- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Some bundlers, such as esbuild, are not able to tree-shake nested `export *` statements.

The Fluent team has also recently turned this on: https://github.com/microsoft/fluentui/pull/22073

Fixed using the `no-export-all` ESLint rule.

### Verification

Functionally, nothing has changed.

### DIFF
--- a/apps/fluent-tester/src/index.ts
+++ b/apps/fluent-tester/src/index.ts
@@ -1,1 +1,1 @@
-export * from './FluentTester';
+export { FluentTesterApp } from './FluentTester';
--- a/change/@fluentui-react-native-adapters-c264b16f-5068-43f6-a8d0-c48bca3b27ff.json
+++ b/change/@fluentui-react-native-adapters-c264b16f-5068-43f6-a8d0-c48bca3b27ff.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Ban `export *` in index files for better tree-shakeability",
+  "packageName": "@fluentui-react-native/adapters",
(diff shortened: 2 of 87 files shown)


apps/fluent-tester/src/index.ts
```diff
@@ -1 +1 @@
-export * from './FluentTester';
+export { FluentTesterApp } from './FluentTester';
```
rurikoaraki, line 1: Seems to be missing FluentTesterProps, is this intentional?

rurikoaraki: Nvm, misread the file

rurikoaraki:
Out of curiosity, some index files are skipped, is that because they're not top-level files?

tido64:
> Out of curiosity, some index files are skipped, is that because they're not top-level files?

This is the rule I added:
https://github.com/microsoft/fluentui-react-native/blob/0fd96baf569e9993568e1bf08748a5696d034244/packages/framework/eslint-config-rules/eslintrc.js#L19-L22

It should only catch the `main` file. Index files that are internal in the package are skipped for now.

rurikoaraki (approved):

PR #1558, opened 2022-04-01
FocusZone macOS: Handle edge case where FocusZone has no elements

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, keyboard navigation breaks when there's an empty FocusZone in a view, that's because a nil key view is being passed in `becomeFirstResponder` that causes the key view loop to reset. This change fixes that by rejecting `acceptsFirstResponder` if the subview count of the current view is less than one.

Note: This logic will break if there are non-focusable elements inside a FocusZone.

### Verification
Added a test case for an empty FocusZone and traversed through (forward & backward directions) all test cases in the FocusZone test page. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/FocusZoneTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/FocusZoneTest.tsx
@@ -51,6 +51,9 @@ const DirectionalFocusZone: React.FunctionComponent = () => {
         <Checkbox label="Option D" />
       </FocusZone>
       <Button content="Outside FocusZone" />
+      <Text variant="headerSemibold">FocusZone with no elements</Text>
+      <FocusZone></FocusZone>
+      <Button content="Outside FocusZone" />
     </View>
   );
 };
--- a/change/@fluentui-react-native-focus-zone-3fe5bc1e-2a0c-4cbf-bdca-12b02be0e803.json
+++ b/change/@fluentui-react-native-focus-zone-3fe5bc1e-2a0c-4cbf-bdca-12b02be0e803.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for FocusZone",
+  "packageName": "@fluentui-react-native/focus-zone",
+  "email": "amchiu@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@fluentui-react-native-tester-1b7c37c9-ebf6-4c8b-9d1c-6c00f632509f.json
+++ b/change/@fluentui-react-native-tester-1b7c37c9-ebf6-4c8b-9d1c-6c00f632509f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for FocusZone",
+  "packageName": "@fluentui-react-native/tester",
+  "email": "amchiu@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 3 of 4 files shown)


PPatBoyd:
`Note: This logic will break if there are non-focusable elements inside a FocusZone.`
I don't know our canonical FocusZone scenarios but find this a little concerning.  A FocusZone with 0 elements seems more niche case than a FocusZone with no focusable elements.  A potential example might be a disabled RadioGroup?

chiuam:
> `Note: This logic will break if there are non-focusable elements inside a FocusZone.` I don't know our canonical FocusZone scenarios but find this a little concerning. A FocusZone with 0 elements seems more niche case than a FocusZone with no focusable elements. A potential example might be a disabled RadioGroup?

What's the behavior of a disabled RadioGroup like? I couldn't find a prop for that in the API. As for a disabled FocusZone, the elements inside are focusable and can be traversed by tabbing. 

Can a FocusZone have no focusable elements?

PPatBoyd:
> > `Note: This logic will break if there are non-focusable elements inside a FocusZone.` I don't know our canonical FocusZone scenarios but find this a little concerning. A FocusZone with 0 elements seems more niche case than a FocusZone with no focusable elements. A potential example might be a disabled RadioGroup?
> 
> What's the behavior of a disabled RadioGroup like? I couldn't find a prop for that in the API. As for a disabled FocusZone, the elements inside are focusable and can be traversed by tabbing.
> 
> Can a FocusZone have no focusable elements?

w.r.t. the FURN RadioGroup it'd be a RadioGroup that only has `disabled` RadioButtons; the group itself doesn't have the prop.  In the case of a RadioGroup existing where all RadioButtons are disabled, I'd expect focus to pass over the entire set; regardless of wrapping FocusZone.

A FocusZone with no focusable elements somewhat implies `focusable` being enabled and disabled, the RadioGroup example works here, but it holds as an issue for more generic components that may not have a direct relationship to focusable (or not-focusable) child elements or their children.  For the second I'm picturing a list of some kind but can't put together a good example quickly, more working off of orthogonality between components

chiuam:
> > > `Note: This logic will break if there are non-focusable elements inside a FocusZone.` I don't know our canonical FocusZone scenarios but find this a little concerning. A FocusZone with 0 elements seems more niche case than a FocusZone with no focusable elements. A potential example might be a disabled RadioGroup?
> > 
> > 
> > What's the behavior of a disabled RadioGroup like? I couldn't find a prop for that in the API. As for a disabled FocusZone, the elements inside are focusable and can be traversed by tabbing.
> > Can a FocusZone have no focusable elements?
> 
> w.r.t. the FURN RadioGroup it'd be a RadioGroup that only has `disabled` RadioButtons; the group itself doesn't have the prop. In the case of a RadioGroup existing where all RadioButtons are disabled, I'd expect focus to pass over the entire set; regardless of wrapping FocusZone.
> 
> A FocusZone with no focusable elements somewhat implies `focusable` being enabled and disabled, the RadioGroup example works here, but it holds as an issue for more generic components that may not have a direct relationship to focusable (or not-focusable) child elements or their children. For the second I'm picturing a list of some kind but can't put together a good example quickly, more working off of orthogonality between components

Thanks Pat! I'm closing this PR and will follow up with a more well-rounded fix.

PR #1566, opened 2022-04-07
Handle FocusZone edge cases

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, keyboard navigation breaks when these two cases happen:
1. FocusZone is empty
2. FocusZone has no focusable elements

That's because `becomeFirstReesponder` is false when a nil key view is being passed in which causes the key view loop to get reset.

Let's handle said edge cases by rejecting firstResponder if FocusZone is empty or has no focusable elements.

### Verification
Added new test cases and traversed through (forward & backward directions) them in the FocusZone test page.

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -157,10 +157,32 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 	return nil;
 }
 
+/// Bypass FocusZone if it's empty or has no focusable elements
+static BOOL ShouldSkipFocusZone(NSView *view)
+{
+    if([view isKindOfClass:[RCTFocusZone class]])
+    {
+        // FocusZone has no elements
+        if ([[view subviews] count] < 1)
+        {
+            return YES;
+        }
+        
+        NSView* keyView = GetFirstKeyViewWithin(view);
+        // FocusZone has no focusable elements
+        if (keyView == nil)
+        {
+            return YES;
+        }
+    }
+
+    return NO;
+}
+
 - (BOOL)acceptsFirstResponder
 {
 	// Accept firstResponder on FocusZone itself in order to reassign it within the FocusZone.
-	return !_disabled;
+    return !_disabled && !ShouldSkipFocusZone(self);
 }
 
 - (BOOL)becomeFirstResponder
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -49,6 +49,11 @@ static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSP
 /// This function does not take into account the geometric position of the view.
 static NSView *GetFirstKeyViewWithin(NSView *parentView)
 {
+    if ([[parentView subviews] count] < 1)
+    {
+        return nil;
+    }
+
 	for (NSView *view in [parentView subviews]) {
 		if ([view canBecomeKeyView]) {
 			return view;
@@ -162,14 +167,8 @@ static BOOL ShouldSkipFocusZone(NSView *view)
 {
     if([view isKindOfClass:[RCTFocusZone class]])
     {
-        // FocusZone has no elements
-        if ([[view subviews] count] < 1)
-        {
-            return YES;
-        }
-        
-        NSView* keyView = GetFirstKeyViewWithin(view);
-        // FocusZone has no focusable elements
+        NSView *keyView = GetFirstKeyViewWithin(view);
+        // FocusZone is empty or has no focusable elements
         if (keyView == nil)
         {
             return YES;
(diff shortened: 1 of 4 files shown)


packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -157,10 +157,32 @@ static inline BOOL IsHorizontalNavigationWithinZoneAction(FocusZoneAction action
...
+            return YES;
+        }
+    }
+
+    return NO;
+}
+
 - (BOOL)acceptsFirstResponder
 {
 	// Accept firstResponder on FocusZone itself in order to reassign it within the FocusZone.
```
harrieshin, line 184: it seems like the comment should be updated?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -157,10 +157,32 @@ static inline BOOL IsHorizontalNavigationWithinZoneAction(FocusZoneAction action
...
+{
+    if([view isKindOfClass:[RCTFocusZone class]])
+    {
+        // FocusZone has no elements
+        if ([[view subviews] count] < 1)
+        {
+            return YES;
+        }
+        
+        NSView* keyView = GetFirstKeyViewWithin(view);
```
harrieshin, line 171: super nit: * location

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -157,10 +157,32 @@ static inline BOOL IsHorizontalNavigationWithinZoneAction(FocusZoneAction action
 	return nil;
 }
 
+/// Bypass FocusZone if it's empty or has no focusable elements
+static BOOL ShouldSkipFocusZone(NSView *view)
+{
+    if([view isKindOfClass:[RCTFocusZone class]])
+    {
+        // FocusZone has no elements
+        if ([[view subviews] count] < 1)
```
harrieshin, line 166: this isn't combinable with line 171. if there isn't a FirstKeyviewwithin the view, there must be no elements in the focuszone?

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: I think this comment states the justification for overriding `acceptsFirstResponder`, though I think putting it above the function header would make more sense.

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: That's correct. My rationale behind adding that check was for readability and avoid calling `GetFirstKeyViewWithin` on an empty FocusZone.

Saadnajmi (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: We could move the "check is empty logic" inside `GetFirstKeyViewWithin` too. But I think that function should be relatively cheap since it will do nothing if there are no subviews.

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: The comment was added when Chris Pearson added support for `defaultTabbableElement` (before I think we just didn't override it). I also think it's worth updating the comment to clarify what cases we do/don't accept first responder.

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: I went with adding the "is empty" check inside `GetFirstKeyViewWithin`. Thought it'd benefit every caller that way

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -157,10 +162,26 @@ static inline BOOL IsHorizontalNavigationWithinZoneAction(FocusZoneAction action
 	return nil;
 }
 
+/// Bypass FocusZone if it's empty or has no focusable elements
+static BOOL ShouldSkipFocusZone(NSView *view)
+{
+    if([view isKindOfClass:[RCTFocusZone class]])
+    {
+        NSView *keyView = GetFirstKeyViewWithin(view);
+        // FocusZone is empty or has no focusable elements
```
harrieshin, line 171: now this comment is redundant from the function's documentation because you hae improved the getfirstkeyviewwithin()

harrieshin (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: I think it's better to state the possibilities here so that we know all the cases why we want to reject first responder. Will prefer to leave it as is.

PR #1570, opened 2022-04-07
Svg platform color

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Branch with removing fallback code for PlatformColors for SVGs

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
This is a draft PR so that the changes can be tested against in win32

PPatBoyd:
Is there already an affordance on the test page to use a PlatformColor?

rurikoaraki:
> Is there already an affordance on the test page to use a PlatformColor?

Checkboxes in HC 😁

But uh no probably not. I'll add a test case to the icon page if there isn't one already.

rurikoaraki:
> Any update on this PR?

Kinda waiting on 0.66 since we would need rex-win32 to be updated for this to work properly in the tester we have, and rex-win32 is now on 0.66...

Checking this in as is would show false bugs in HC so I would rather avoid that. We could just bump rex-win32 since running FURN against latest doesn't seem to cause problems?

Saadnajmi:
Nah, waiting on 0.66 makes sense. I would rather have more pressure to get that PR merged :P

Saadnajmi:
> > Any update on this PR?
> 
> @Saadnajmi  Kinda waiting on 0.66 since we would need rex-win32 to be updated for this to work properly in the tester we have, and rex-win32 is now on 0.66...
> 
> Checking this in as is would show false bugs in HC so I would rather avoid that. We could just bump rex-win32 since running FURN against latest doesn't seem to cause problems?

Wait.. is not bumping rex the reason that win32 fails on my 0.66 merge PR...

samuelfreiberg (approved):

apps/fluent-tester/src/TestComponents/Icon/IconTest.tsx
```diff
@@ -78,6 +78,7 @@ const Icons: React.FunctionComponent = () => {
             shouldShowDataUri ? <Icon svgSource={svgD20DataUriProps} width={100} height={100} color="#7a7" /> : null
           }
           <Icon svgSource={svgUriProps} width={100} height={100} color="red" />
+          <Icon svgSource={svgProps} width={100} height={100} color={PlatformColor('WindowText')} accessibilityLabel="Wheelchair" />
```
Saadnajmi, line 81: Can we make sure this is an platform.os check? It'll crash on other platforms I think

rurikoaraki: Oh yep, sorry about that

Saadnajmi: It would be nice to pull this out and test multiple platforms' platform colors, but we can do that later!

Saadnajmi (approved):

PR #1511, opened 2022-03-11
Update to React Native 0.66

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

CAUTION: Don't merge until downstream dependencies are updated to RN 0.66 _first_

This change updates the entire repo to React Native 0.66. Because this is a fairly large change, I tried to have each individual commit to be easy-to-read steps. 

Working:
- iOS and macOS test apps boot fine and seem to work

Open issues:
- We use "Picker" from React Native in several places. In RN 0.66, it is removed from Core and must be consumed from the community module. Furthermore, I think Picker just doesn't exist anymore for react-native-win32 so we need to move all usage of that in our repo away.
- Windows crashes on boot. Probably because of the reason above, but we can autolink the community module for windows at least.

### Verification

Ran the iOS and macOS test app. Still need to run Android, Windows, Win32

### Pull request checklist

This PR has considered (when applicable):
- [ ] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #1576, opened 2022-04-09
Remove unused deps

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Removing some devDeps that don't seem to be used by the Drawer component. The deps are pulling in a version of string_decoder that is discouraged from being used.

### Verification

Re-ran "gradle init" to make sure that it's not adding back the devDependencies
Built and booted Android FURN test app. Unfortunately Drawer was also broken without my changes, but given that the change builds and these are dev dependencies, it shouldn't affect the FURN test app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

Saadnajmi:
Thanks, fixing up Drawer dependencies should also help with #1511 and/or #1535

Issue #1336, opened 2022-01-11
Experimental Button: disabled button is pressable


chiuam:
PR #1332 made a workaround to stop disabled button from getting keyboard focus. A better fix may be figuring out why onPress is working in disabled state and perhaps disable it.

chiuam:
Update: investigate why `focusable` get disregarded when useOnPressWithFocus is called to set focus on a component.

PR #1587, opened 2022-04-12
Disable pressable behavior on button when button is disabled

ksiler:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

There's a native change on win32 I'm working on that will prevent accessibilityState.disabled=true from disabling the control and instead only report the control as disabled from an accessibility perspective. When testing this change, I noticed there was a bug where you could still click on a disabled button and have it perform the onClick behavior. To prevent this, I passed the disabled prop to the useAsPressable hook in useButton.js so pressable behavior is disabled when disabled=true. 

### Verification

Manually tested behavior in FluentTester

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):
Good catch!

samuelfreiberg (approved):

packages/components/Button/src/useButton.ts
```diff
@@ -11,7 +11,7 @@ export const useButton = (props: ButtonProps): ButtonState => {
   // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
   const focusRef = disabled ? null : componentRef;
   const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
-  const pressable = useAsPressable({ ...rest, onPress: onClickWithFocus });
+  const pressable = useAsPressable({ ...rest, disabled, onPress: onClickWithFocus });
```
Saadnajmi, line 14: I wonder if we should just pass all the props to it to avoid problems like this in the future, and for other controls? I think this fix is fine though.

packages/components/Button/src/useButton.ts
```diff
@@ -11,7 +11,7 @@ export const useButton = (props: ButtonProps): ButtonState => {
   // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
   const focusRef = disabled ? null : componentRef;
```
PPatBoyd, line 12: driveby comment, why are we only filling the ref when the button is enabled?  The focus() function can no-op and avoid state changes that may invoke re-renders of parent components.  Should we open an issue to change this?

packages/components/Button/src/useButton.ts
```diff
@@ -11,7 +11,7 @@ export const useButton = (props: ButtonProps): ButtonState => {
...
   const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
-  const pressable = useAsPressable({ ...rest, onPress: onClickWithFocus });
+  const pressable = useAsPressable({ ...rest, disabled, onPress: onClickWithFocus });
   const onKeyUpProps = useKeyProps(onClick, ' ', 'Enter');
   const isDisabled = !!disabled || !!loading;
```
PPatBoyd, line 16: we can end up with disabled != isDisabled; should we evaluate isDisabled sooner and use it everywhere, or is there a nuanced need for the conflict to be possible?

packages/components/Button/src/useButton.ts
PPatBoyd: +1; while we need to setup onPress seemingly the initial props parameters can be passed through in one

packages/components/Button/src/useButton.ts
rurikoaraki: Yeah, we should move isDisabled up and pass it in instead of just disabled.

packages/components/Button/src/useButton.ts
rurikoaraki: The issue that the comment links to notes that this is a workaround: https://github.com/microsoft/fluentui-react-native/issues/1336 and the initial PR, https://github.com/microsoft/fluentui-react-native/pull/1332

rurikoaraki: Basically, something on the MacOS' onPress hookup seems to not be working properly so they do this instead in the meantime.

rurikoaraki (changes requested):
Can we pass in isDisabled instead to keep click behavior consistent?

packages/components/Button/src/useButton.ts
PPatBoyd: reading comprehension failure with a high contrast theme disabling syntax highlighting, unfortunate

packages/components/Button/src/useButton.ts
ksiler: I updated usePressable to use props in the newest iteration.

packages/components/Button/src/useButton.ts
```diff
@@ -7,13 +7,13 @@ import { AccessibilityState } from 'react-native';
...
   const defaultComponentRef = React.useRef(null);
-  const { onClick, accessibilityState, componentRef = defaultComponentRef, disabled, loading, ...rest } = props;
+  const { onClick, accessibilityState, componentRef = defaultComponentRef, disabled, loading } = props;
+  const isDisabled = !!disabled || !!loading;
   // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
-  const focusRef = disabled ? null : componentRef;
+  const focusRef = isDisabled ? null : componentRef;
   const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
-  const pressable = useAsPressable({ ...rest, onPress: onClickWithFocus });
+  const pressable = useAsPressable({ ...props, onPress: onClickWithFocus });
```
rurikoaraki, line 15: I think you'll need to specifically override disabled and pass disabled: isDisabled here

packages/components/Button/src/useButton.ts
ksiler: I updated this in the newest iteration.

packages/components/Button/src/useButton.ts
ksiler: Ah right, good catch.

rurikoaraki (approved):

rurikoaraki:
@ksiler Hmmm I don't see that behavior myself... Can you make sure you've rebuilt with your latest changes?

(Also jsyk my GH username includes my last name 😅)

ksiler:
@rurikoaraki whoops sorry! Thanks for catching that. I think I'm seeing a difference because I'm applying a native change that hasn't been checked in yet.

ksiler:
I'm getting this error when running jest tests under packages/components/menubutton:

`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

Seems that it might be related to [jest v26](https://github.com/facebook/jest/issues/9980) but I'm still investigating the issue. Running `yarn jest --clear-cache` did not help.

ksiler:
Could not figure out the jest test issue :( but it does seem to be stemming from using `...props` vs. `...rest, disabled: IsDisabled` so I'm changing it back to the latter for now.

rurikoaraki (approved):

packages/utils/interactive-hooks/src/Pressability/Pressability.ts
```diff
@@ -395,7 +395,8 @@ export class Pressability {
         ? null
         : {
             onMouseEnter: (event: MouseEvent): void => {
-              if (isHoverEnabled()) {
+              const { disabled } = this._config;
+              if (!disabled && isHoverEnabled()) {
```
rurikoaraki, line 399: @SaadNajmi this is a short term workaround so that Krystal can make progress on her changes. I have an item logged to make Loading button not change visuals on hover and remove this, so no worries about bringing it upstream.

PR #1607, opened 2022-04-22
Menu prototype/initial impl

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Doing initial work for Menu and seeing if the current design will pan out.

This change implements Menu, MenuTrigger, and MenuPopover:
- Menu sets up context which its children will use to keep track of a lot of state. It also controls whether the Callout is showing
- MenuPopover is just setting up a Callout wrapper around its children
- MenuTrigger will override onClick on its children to call setOpen on the context which will then affect open state of the callout.

Still kind of want to look into React.cloneElement usage to refresh myself on what the perf implications are, but sending this out since I have a working prototype.

### Verification

Test page!
Animation

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/package.json
```diff
@@ -27,7 +27,8 @@
   },
   "dependencies": {
     "@fluentui-react-native/adapters": ">=0.8.5 <1.0.0",
-    "@fluentui-react-native/framework": "0.7.23",
+"@fluentui-react-native/framework": "0.7.23",
+    "@fluentui-react-native/interactive-hooks": ">=0.15.3 <1.0.0",
```
Saadnajmi, line 31: spacing is off?

packages/experimental/Menu/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    let menuTrigger = undefined;
+    let menuPopover = undefined;
+    if (childrenArray.length === 2) {
+      menuTrigger = childrenArray[0];
+      menuPopover = childrenArray[1];
+    } else if (childrenArray.length === 1) {
+      menuPopover = childrenArray[0];
+    }
```
Saadnajmi, line 34: In this case, what is rendered on line 38?

Saadnajmi (approved):

packages/experimental/Menu/package.json
rurikoaraki: Uh yeah idk how prettier didn't catch that

packages/experimental/Menu/src/Menu/Menu.tsx
rurikoaraki: Just the popover. Since the first child is undefined, it shouldn't render anything if left blank.

I can take out this case for now and add it back when custom target is implemented?

Saadnajmi: Oh, interesting. Up to you whether to leave it or not, I was just curious

rurikoaraki: I'll remove it for now. No need to make things more complicated I guess.

PR #1610, opened 2022-04-22
Implement target behavior for Menu

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding target behavior to Menu. This change sets up setting a ref on the MenuTrigger and then using that ref as the target on the Callout in MenuPopover

### Verification

menu_target

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/Menu/Menu.tsx
+++ b/packages/experimental/Menu/src/Menu/Menu.tsx
@@ -13,7 +13,7 @@ export const Menu = stagedComponent((props: MenuProps) => {
     const childrenArray = React.Children.toArray(children) as React.ReactElement[];
 
     if (__DEV__) {
-      if (childrenArray.length === 2) {
+      if (childrenArray.length !== 2) {
         // eslint-disable-next-line no-console
         console.warn('Menu must contain two children');
       }
--- a/packages/experimental/Menu/src/Menu/useMenu.ts
+++ b/packages/experimental/Menu/src/Menu/useMenu.ts
@@ -3,10 +3,12 @@ import { MenuProps, MenuState } from './Menu.types';
 
 export const useMenu = (props: MenuProps): MenuState => {
   const [open, setOpen] = useMenuOpenState(props);
+  const triggerRef = React.useRef(null);
 
   return {
     open,
     setOpen,
+    triggerRef,
   };
 };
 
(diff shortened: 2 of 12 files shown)


Saadnajmi (approved):

lenahong (approved):

packages/experimental/Menu/src/Menu/Menu.tsx
```diff
@@ -13,7 +13,7 @@ export const Menu = stagedComponent((props: MenuProps) => {
     const childrenArray = React.Children.toArray(children) as React.ReactElement[];
 
     if (__DEV__) {
-      if (childrenArray.length === 2) {
+      if (childrenArray.length !== 2) {
         // eslint-disable-next-line no-console
         console.warn('Menu must contain two children');
```
PPatBoyd, line 18: does console.warn turn into a logbox warn?

packages/experimental/Menu/src/Menu/useMenu.ts
```diff
@@ -3,10 +3,12 @@ import { MenuProps, MenuState } from './Menu.types';
 
 export const useMenu = (props: MenuProps): MenuState => {
   const [open, setOpen] = useMenuOpenState(props);
+  const triggerRef = React.useRef(null);
```
PPatBoyd, line 6: quick check -- does changing the target at runtime cause the appropriate re-render/move?

packages/experimental/Menu/src/Menu/Menu.tsx
rurikoaraki: It used to, not sure if that's still the case

packages/experimental/Menu/src/Menu/useMenu.ts
rurikoaraki: Not sure, but I think it should? It's kind of hard to test without coding this out more since right now target is hardcoded just to the trigger that MenuTrigger wraps. Since this is a prototype just to make sure that we can build out the structure we want, I didn't add more functionality than necessary to prove that target will work.

PPatBoyd: no problem! I'm not specifically concerned, it's something of a "technically correct" scenario for menus because there isn't a reason to be moving around vs closing.  I've tried to mind ref runtime behaviors since they aren't state on their own (useRef) and don't cause re-renders which has caught me a couple times.

PR #1611, opened 2022-04-25
ContextualMenu - Only show scrollbars if we specify maxHeight/maxWidth

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Note: This PR requires https://github.com/microsoft/react-native-macos/pull/1135 to land and get back ported as a prerequisite

On macOS, there are 2 types of scrollbars that you can switch between in user preferences:
- Legacy: Scrollbars that appear at the trailing end of the content view in a dedicated space. 
- Overlay: iOS style scrollbars that are overlaid on top of the content view.

The former legacy style scrollbars add some margins to the scroll views' shadow view in order to account for the scroll bars. There is a race condition during layout of our ContextualMenu where the margins may show up even if we don't have a need for the scrollbars because all the content is visible. This was rather awkward, and a pretty big visual bug in ContextualMenu.

https://user-images.githubusercontent.com/6722175/165080842-e47d5364-a865-4267-8e82-e4e9127a3433.mov

We can fix this by telling ContextualMenu to only show scroll bars when we need them, AKA, when the user specified a maxHeight for vertical scrollbars, or maxWidth for horizontal scrollbars. 

### Verification

I tested a few different menus on our ContextualMenu test page, including one that wants a scrollbar and one that doesn't. Apart from that very last submenu, things seem to be laying out correctly. 

https://user-images.githubusercontent.com/6722175/165081373-b244bc1f-803c-433a-ab9d-78f7289bfa27.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/macos/src/Podfile.lock
+++ b/apps/macos/src/Podfile.lock
@@ -9,17 +9,17 @@ PODS:
     - React-Core (= 0.64.30)
     - React-jsi (= 0.64.30)
     - ReactCommon/turbomodule/core (= 0.64.30)
-  - FRNAvatar (0.14.12):
+  - FRNAvatar (0.14.15):
     - MicrosoftFluentUI (= 0.3.0)
     - MicrosoftFluentUI/Avatar_ios (= 0.3.0)
     - React
-  - FRNCallout (0.19.39):
+  - FRNCallout (0.19.43):
     - React
-  - FRNCheckbox (0.10.11):
+  - FRNCheckbox (0.10.16):
     - React
-  - FRNMenuButton (0.7.49):
+  - FRNMenuButton (0.7.55):
     - React
-  - FRNRadioButton (0.14.37):
+  - FRNRadioButton (0.14.42):
     - React
   - glog (0.3.5)
   - MicrosoftFluentUI (0.3.0):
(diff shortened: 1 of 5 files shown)


harrieshin (approved):

PPatBoyd:
@Saadnajmi and if the context menu takes up the entire height of the screen, e.g. in a min-resolution max-text scale scenario?

Saadnajmi:
> @Saadnajmi and if the context menu takes up the entire height of the screen, e.g. in a min-resolution max-text scale scenario?

That's why we added support for a scroll view. It was added because our partners have that scenario, and adding a scroll view + max height prop to limit the size and make sure it never gets too big.

Saadnajmi:
@PPatBoyd I tested it locally. I made a long list with 36 menu items (I just copy-pasted the original 12 items over twice), set `showsVerticalIndicator=true`, and removed the maxHeight/maxWidth. On win32, it seems that the Callout will just clip, and not show a scrollbar. It seems that you do need to have the max height set, which is probably why @lenahong added it to begin with. So I think this change is still valid on win32.

image

PR #1615, opened 2022-04-25
Add Text props and test new Variant types

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- Added new props to align with web Text
- Defined alias tokens for new variant types
- Added standard tests for new variants

### Verification

new variants

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Any idea where the files under the ".idea" folder are coming from?

apps/fluent-tester/src/FluentTester/TestComponents/TextExperimental/StandardUsage.tsx
```diff
@@ -9,7 +9,7 @@ export const StandardUsage: React.FunctionComponent = () => {
   return (
     <View>
       <Stack style={stackStyle} gap={5}>
-        <Text variant="captionStandard" testID={HOMEPAGE_EXPERIMENTAL_TEXT_BUTTON}>
+        {/*<Text variant="captionStandard" testID={HOMEPAGE_EXPERIMENTAL_TEXT_BUTTON}>
```
rurikoaraki, line 12: Do we still need these commented out lines?

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -40,6 +40,20 @@ function _defaultTypography(): Typography {
       heroSemibold: { face: 'primary', size: 'hero', weight: 'semiBold' },
       heroLargeStandard: { face: 'primary', size: 'heroLarge', weight: 'regular' },
       heroLargeSemibold: { face: 'primary', size: 'heroLarge', weight: 'semiBold' },
+      //v1 text variants
+      caption1: { face: globalTokens.font.family['base'] as FontFamily, size: globalTokens.font.size['100'] as FontSize, weight: globalTokens.font.weight.regular as FontWeightValue },
```
rurikoaraki, line 44: In the future we should try to pull these from alias tokens so that these can change per platform without us needing to check for the current platform.

rurikoaraki: And then we could collapse this and the win32 variants code for the FluentUI v2 variants into one

packages/experimental/Text/src/Text.types.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const textName = 'Text';
+
+/**
+ * Text tokens, these are the internally configurable values for Text elements. In particular these
+ * drive decisions on how to build the styles
+ */
+ export interface TextTokens extends FontTokens {
+  /** foreground text color */
+  color?: ColorValue;
```
rurikoaraki, line 17: I think IForegroundColorTokens would cover the color prop, so you could have TextTokens extend IForegroundCOlorTokens as well.

packages/experimental/Text/src/Text.types.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Text props, based off of the standard react-native TextProps with some new extensions
+ */
+export type TextProps<TBase = ITextProps> = TBase &
+  FontVariantTokens & {
+    /** foreground text color */
+    color?: ColorValue;
+
+    /** alternate mode for disabled look and feel */
+    disabled?: boolean;
```
rurikoaraki, line 34: I think we decided that this could be removed

packages/experimental/Text/src/Text.types.ts
```diff
@@ -0,0 +1,112 @@
...
+export type TextFont = 'base' | 'monospace' | 'numeric';
+export type TextSize = 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900 | 1000;
+export type TextWeight = 'regular' | 'medium' | 'semibold' | 'bold';
+
+/**
+ * Text props, based off of the standard react-native TextProps with some new extensions
+ */
+export type TextProps<TBase = ITextProps> = TBase &
+  FontVariantTokens & {
+    /** foreground text color */
```
rurikoaraki, line 30: Can the commenting style be changed to be consistent with the other comments?

packages/experimental/Text/src/Text.tsx
```diff
@@ -1,67 +1,25 @@
...
-export type TextProps<TBase = ITextProps> = TBase &
-  FontVariantTokens & {
-    /** foreground text color */
-    color?: ColorValue;
-
-    /** whether or not this text should be presented as disabled */
-    disabled?: boolean;
-  };
-
-const useTextTokens = buildUseTokens<TextTokens>(
```
rurikoaraki, line 45: How is this building without useTextTokens?

Also. it seems like useTextTokens is supplying some defaults and props can be used to override them, so I think it would make sense to keep it and use it to fill defaults from tokens.

rurikoaraki (changes requested):
See comments

packages/experimental/Text/src/Text.tsx
lenahong: I actually moved it to the newly created Text.types file. Do you think it makes more sense to be placed here?

rurikoaraki: A .types file should only contain types or interface declarations. I don't think it should hold any functional code.
If you'd like the styling to be in its own file, we have Component.styling.ts ComponentTokens.ts in other components that I think would fit this better.

rurikoaraki (approved):

Saadnajmi:
> Any idea where the files under the ".idea" folder are coming from?

Post PR merge comment: these are probably coming from Android Studio. `.idea` is generated by the IntelliJ IDEA, the IDE which Android Studio is based off.

PR #1634, opened 2022-05-03
Initial Menu styling

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add initial Menu styling according to redlines.

MenuPopover:
- Add 1px border with color neutralStrokeAccessible (these should probably be tokenized in a later PR)

MenuList:
- New component which is a View that wraps menu items
- 4px padding (all sides non-win32, vertical only win32)
- 2 px gap between items non-win32, no gap between items win32 (NYI right now)

MenuItem:
- Add interactivity states
- Add tokens. Foreground, height, borderRadius, padding are different between non-win32 and win32.

### Verification

| menu_styling_before | menu_styling |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,24 @@
...
+  displayName: menuListName,
+  ...stylingSettings,
+  slots: {
+    root: View,
+  },
+  useRender: (userProps: MenuListProps, useSlots: UseSlots<MenuListType>) => {
+    const Slots = useSlots(userProps);
+
+    return (_final: MenuListProps, children: React.ReactNode) => {
+      // TODO add gap between children
```
rurikoaraki, line 18: Going to implement this in separate PR since this is already pretty big

packages/experimental/Menu/src/MenuList/MenuList.styling.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const stylingSettings: UseStylingOptions<MenuListProps, MenuListSlotProps, MenuListTokens> = {
+  tokens: [defaultMenuListTokens, menuListName],
+  slotProps: {
+    root: buildProps(
+      (tokens: MenuListTokens, theme: Theme) => ({
+        style: {
+          backgroundColor: tokens.backgroundColor,
+          display: 'flex',
+          gap: tokens.gap,
```
Saadnajmi, line 14: gap doesn't exist on react-native, unfortunately. There's been an open PR in yoga to add it that seems to have not gained traction.

packages/experimental/Menu/src/MenuItem/MenuItemTokens.ts
```diff
@@ -0,0 +1,29 @@
+import { FontWeightValue, Theme } from '@fluentui-react-native/framework';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { MenuItemTokens } from './MenuItem.types';
+
+export const defaultMenuItemTokens: TokenSettings<MenuItemTokens, Theme> = (t: Theme): MenuItemTokens => ({
+  backgroundColor: t.colors.neutralBackground1,
```
Saadnajmi, line 7: Are these "defaults" just the web tokens?

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: The "Stack" component I think exists to handle gap?

packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
```diff
@@ -1,14 +1,19 @@
...
 export const MenuPopover = stagedComponent((props: MenuPopoverProps) => {
   const state = useMenuPopover(props);
+  const theme = useFluentTheme();
 
   return (_rest: MenuPopoverProps, children: React.ReactNode) => {
-    return <Callout target={state.triggerRef}>{children}</Callout>;
+    return (
+      <Callout target={state.triggerRef} borderWidth={1} borderColor={theme.colors.neutralStrokeAccessible}>
+        {children}
+      </Callout>
```
Saadnajmi, line 15: Any reason this isn't a Slot with slotProps?

packages/utils/tokens/src/layout-tokens.ts
```diff
@@ -5,16 +5,27 @@ import { tokenBuilder } from './tokenBuilder';
 
 export interface LayoutTokens {
   minWidth?: ViewStyle['minWidth'];
+  maxWidth?: ViewStyle['maxWidth'];
   minHeight?: ViewStyle['minHeight'];
   padding?: ViewStyle['padding'];
   paddingHorizontal?: ViewStyle['paddingHorizontal'];
+  paddingVertical?: ViewStyle['paddingVertical'];
```
Saadnajmi, line 12: nit: This package pulls in from `@uifabricshared/foundation-tokens` on line 2, which I think is deprecated but used in some places? I saw it used in Floating Action Button. Is that something we need to stop referencing or is it OK?

packages/experimental/Menu/src/MenuList/MenuList.styling.ts
rurikoaraki: Yeah this is me jamming it into style for now because the compiler doesn't complain lol. I'm working locally on getting it to work, but this will move in the future. I suppose I can take it out for now.

packages/experimental/Menu/src/MenuItem/MenuItemTokens.ts
rurikoaraki: Yeah, the ones in this file are from web.

packages/utils/tokens/src/layout-tokens.ts
rurikoaraki: Hmmm I'll look into it.

packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
rurikoaraki: laziness?

Actually I find it unlikely that someone would want to swap out the Callout with something else, especially since it handles the native popover logic/functionality, so I was wondering if it was worth making this a slot.

packages/experimental/Menu/src/MenuList/MenuList.tsx
rurikoaraki: Oh that's true, I could just use Stack instead of reimplementing it. Hmm.

Saadnajmi: Can still be done separately, just pointing it out as an initial venue to explore =)

packages/experimental/Menu/src/MenuList/MenuList.styling.ts
Saadnajmi: I'd prefer taking it out, so as to not mislead people that this prop exists.

packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
Saadnajmi: I see. I've defaulted to just making everything a slot fit the sake of following patterns.

A real example of when you would want to swap out a Callout: One might want to replace it with just a View with absolute position (AKA, a floating view), for a platform that doesn't support Callout / more styling options / etc. I did that a lot as a "development shortcut" when Callout wasn't implemented on macOS, and will probably repeat the same for iOS.

 
Or maybe an inline menu 🤷 I think I saw an example of that in web.

Saadnajmi: I see. I've defaulted to just making everything a slot fit the sake of following patterns.

A real example of when you would want to swap out a Callout: One might want to replace it with just a View with absolute position (AKA, a floating view), for a platform that doesn't support Callout / more styling options / etc. I did that a lot as a "development shortcut" when Callout wasn't implemented on macOS, and will probably repeat the same for iOS

 
Or maybe an inline menu 🤷 I think I saw an example of that in web.

packages/experimental/Menu/src/MenuList/MenuList.tsx
rurikoaraki: Removing gap for now.

packages/utils/tokens/src/layout-tokens.ts
rurikoaraki: Hm yeah I think that'll be its own thing. https://github.com/microsoft/fluentui-react-native/issues/1639

Saadnajmi (approved):

packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
rurikoaraki: For inline menus the intent is that you just use MenuList, and leave out the Popover wrapping. So those cases should be fine.

Saadnajmi: Yeah, re-reading this I realize they ship MenuList separately for the case of inline menus. So slots / this slots-as-jsx stuff both kinda do similar things I realize

PR #1643, opened 2022-05-07
Callout do not take pointer capture prop

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add the doNotTakePointerCapture boolean property to the Callout properties interface, which defines event redirection behaviors for pointer events relative to the Callout control and its parent surface.
 
Pointer events (i.e. mouse, pen, touch) often have the effect of initiating and dismissing Callout controls. Typically the instigating events are click events, but when using a mouse may also include pointer enter and leave events i.e. hover.  Once opened, many Callout scenarios desire modal-like interactions, where the user must interact with or relative to the Callout before interacting with other controls again. A common example is context menus, where pointer events away from the menu may dismiss the context menu.  Such an interaction can initiate unintended actions such as invoking a button or otherwise pointer-responsive surface.  The modal-like behavior is the most common case of Callout usage, and having "safe" areas to press-dismiss the Callout without invoking additional actions is useful.

This modal-like behavior, where true modal behavior would specifically require interacting with the modal component until it is dismissed, has implications for canonical event routing in React-Native.  Canonically all pointer events are passed in a Capture phase from the root element to the target element, then bubbled back from the target element to the root element. This structure allows many interaction patterns to be implemented that may include stopping event routing in each phase. This event routing can disrupt the modal-like behavior desired by a Callout control, which in native frameworks may take advantage of a "take event capture" feature -- making the Callout the first handler of the pointer event.  The event may optionally be forwarded afterwards to the canonical event routing pattern for further handling (and often is not forwarded).

One impact of "taking event capture" is the implicit behavior for mouse events on the parent surface of the Callout -- if event capture is taken by the Callout, paired events may fire such as a mouse leave event.  This makes sense if we think about the event behavior after the Callout has taken capture; mouse enter and mouse leave events would only propagate to the parent surface if the Callout event handling decided to forward said events (often not).  As paired events, however, there should not be a mouse enter event without a corresponding mouse leave event (or vice versa).  Since mouse enter and mouse leave events would not naturally fire on the parent surface when the Callout has pointer capture, the parent surface should receive a mouse leave event when the Callout takes capture.

Coming back to the start of this comment, some Callout scenarios may be initiated by mouse enter/leave events.  A common example is tooltips, which are supported on lower-level React-Native components, but such behavior is not exclusive to tooltips. The Callout control is generally used for more advanced scenarios, as mentioned before often desiring modal-like behavior and accordingly the default behavior of the Callout control is to take pointer capture.  Accounting for the paired mouse enter/leave events in the paragraph above, it is difficult to have a mouse enter/leave-invoked Callout that takes pointer capture -- without significant workarounds, the Callout would open and close cyclically from repeated mouse enter/leave events.  There are other scenarios where taking pointer capture is not desirable, such as a semi-persistent Callout; to support these scenarios the doNotTakePointerCapture property is offered to specify if the Callout should not take pointer capture when it opens.

While pointer capture is generally managed in native frameworks via imperative APIs, managing pointer capture is offered as a property that only affects the Callout's on-show behavior when it would typically take pointer capture.  This design implies that changing this property after the Callout has been shown has no effect -- similar to the Callout's onShow callback.  Doing so simplifies Callout usage for developers, not needing to mind native platform pointer capture that may be complex particularly when React-Native is integrated into a larger app. Pointer capture may be lost (and regained) but is otherwise abstracted away from the developer.  More advanced control of pointer capture would need to be provided by a native module independent of any particular control.

When the Callout is closed, pointer capture is released by the Callout and further handled by the native platform -- generally returning pointer capture to the parent surface.

### Verification

Tested manually with test page additions

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/src/Callout.types.ts
```diff
@@ -110,6 +110,56 @@ export interface ICalloutProps extends IViewProps, ICalloutTokens {
    */
   componentRef?: React.RefObject<IFocusable>;
 
+  /**
+   * Defines event redirection behaviors for pointer events relative to the Callout control.
```
rurikoaraki, line 114: I wonder if some of this is better in a separate architecture file which can be linked to instead of in the code itself? I guess it's better that the note is as close to code as possible..

rurikoaraki:
This is actually something I need for one of the props on the new Menu, so great timing!

packages/components/Callout/src/Callout.types.ts
PPatBoyd: probably -- will open a task to do this

PR #1656, opened 2022-05-12
Shadow tokens theme integration

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Adding shadow tokens to Theme object for each platform. Notably the Android and iOS platforms have mocked out shadows since integration w/ FluentV2 is pretty minor.

1. Theme object now has "shadows" property which holds the shadow token values. Generally accessed off the theme as:
`theme.shadows.shadow2`. Each shadow has an ambient and key portion, and each of those has x/y offset, blur, and color.
2. Added mapPipelineToShadows which takes token pipeline output and translates it into the structure which can be put on the theme object.
3. Each platform then fills out their own Theme object information by accessing the relevant shadowTokens and passing them into mapPipelineToShadows. This is wrapped in a memo so that it doesn't run more often than necessary.

### Verification

Once token package is updated to actually have shadow tokens, test that shadow tokens are showing on theme object properly and merging is happening properly

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Viewing commit-by-commit might be easier to follow for this PR - sorry for the size.

rurikoaraki:
note to self: go change shadow tokens on FAB

rurikoaraki:
I deleted the FABColorTokens.win32/windows files since they shouldn't be needed - FAB only renders on mobile

Saadnajmi (approved):

packages/theming/theme-types/src/Shadow.types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Base structure of a shadow style, which has x and y offset, blur amount, and color specified
+ */
+export interface ShadowStyle {
+  x: number;
+  y: number;
+  blur: number;
+  color: ColorValue;
+}
```
Saadnajmi, line 11: Now that I've plated around with this a little more, the naming here kinda conflicts with the type  [ShadowStyleIOS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4f4f4a891a67881be36fa0c1d41d809f60afcdbb/types/react-native/index.d.ts#L3498), which is the platform specific iOS (and apparently macOS and win32) props the final View needs. 

It kinda feels like this should be called ShadowToken.. except that is what we call line 18. Maybe call line 18 ShadowSet? Not sure if the name "ShadowStyle" is also inherited from the pipeline or not.

rurikoaraki: We can rename these, it makes sense not to clash with ShadowStyleIOS

rurikoaraki: I went with `ShadowValue` as the new name which aligns with the naming scheme we have for `Variant`. I wanted to keep `ShadowToken` as is because a shadow token is specified with both ambient and key parts, so I think it's appropriate for the Token to be the type which encompasses both.

PR #1657, opened 2022-05-12
Implement checkmark styling for MenuItem and MenuItemCheckbox

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change just adds styling for MenuItem and MenuItemCheckbox to have a checkmark. The actual toggling & keeping track of checked vs. unchecked behavior is not part of this PR and will be a separate change.

`hasCheckmarks` is added as a prop to `Menu` and `MenuList`. This value gets added to the menu context which MenuItemX then uses to determine if it should apply styling for checkmarks, namely giving space for alignment & showing the checkmark.

MenuItem adds some space & padding.
MenuItemCheckmark shows a checkmark. The checkmark was copied from https://github.com/microsoft/fluentui-system-icons/blob/master/assets/Checkmark/SVG/ic_fluent_checkmark_12_regular.svg

### Verification

checkmarks

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.types.ts
+++ b/packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.types.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { ViewProps } from 'react-native';
+import { ColorValue, ViewProps } from 'react-native';
+import { SvgProps, XmlProps } from 'react-native-svg';
 import type { IViewProps } from '@fluentui-react-native/adapters';
 import { TextProps } from '@fluentui-react-native/experimental-text';
 import { IFocusable, InteractionEvent, IPressableHooks, IWithPressableOptions } from '@fluentui-react-native/interactive-hooks';
@@ -8,6 +9,13 @@ import { FontTokens, IBorderTokens, IColorTokens, LayoutTokens } from '@fluentui
 export const menuItemCheckboxName = 'MenuItemCheckbox';
 
 export interface MenuItemCheckboxTokens extends LayoutTokens, FontTokens, IBorderTokens, IColorTokens {
+  checkmarkColor?: ColorValue;
+  checkmarkPadding?: number;
+  checkmarkSize?: number;
+  checkmarkVisibility?: number;
+  gap?: number;
+
+  checked?: MenuItemCheckboxTokens;
   disabled?: MenuItemCheckboxTokens;
   focused?: MenuItemCheckboxTokens;
   hovered?: MenuItemCheckboxTokens;
@@ -28,20 +36,18 @@ export interface MenuItemCheckboxProps extends Omit<IWithPressableOptions<ViewPr
   componentRef?: React.RefObject<IFocusable>;
 
   /**
-   * If the menu item is a trigger for a submenu
-   */
-  hasSubmenu?: boolean;
-
-  /**
    * A callback to call on button click event
    */
   onClick?: (e: InteractionEvent) => void;
 }
 
-export type MenuItemCheckboxState = IPressableHooks<MenuItemCheckboxProps & React.ComponentPropsWithRef<any>>;
+export interface MenuItemCheckboxState extends IPressableHooks<MenuItemCheckboxProps & React.ComponentPropsWithRef<any>> {
+  hasCheckmarks?: boolean;
+}
 
 export interface MenuItemCheckboxSlotProps {
   root: React.PropsWithRef<IViewProps>;
+  checkmark?: SvgProps | XmlProps;
   content?: TextProps;
 }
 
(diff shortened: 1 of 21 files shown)


Saadnajmi:
is the spec to have `hasCheckmarks` be an actual prop? or is that just a placeholder to demonstrate styling?

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -11,7 +11,7 @@ export const useMenuItemCheckbox = (props: MenuItemCheckboxProps): MenuItemCheck
   const { onClick, accessibilityState, componentRef = defaultComponentRef, disabled, ...rest } = props;
   const pressable = useAsPressable({ ...rest, disabled, onPress: onClick });
   const onKeyProps = useKeyProps(onClick, ' ', 'Enter');
-  const hasSubmenu = useMenuContext().isSubmenu;
```
Saadnajmi, line 14: Feels like this shouldn't have gone away?

packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.types.ts
```diff
@@ -27,21 +35,19 @@ export interface MenuItemCheckboxProps extends Omit<IWithPressableOptions<ViewPr
...
 }
 
-export type MenuItemCheckboxState = IPressableHooks<MenuItemCheckboxProps & React.ComponentPropsWithRef<any>>;
+export interface MenuItemCheckboxState extends IPressableHooks<MenuItemCheckboxProps & React.ComponentPropsWithRef<any>> {
+  hasCheckmarks?: boolean;
+}
 
 export interface MenuItemCheckboxSlotProps {
   root: React.PropsWithRef<IViewProps>;
+  checkmark?: SvgProps | XmlProps;
```
Saadnajmi, line 50: Dumb question: Why does this need to be a union type?

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
rurikoaraki: No this is by design. The MenuItemCheckbox doesn't support submenus. (Doesn't really make sense for a submenu trigger to be a toggle-able button)

rurikoaraki:
> is the spec to have `hasCheckmarks` be an actual prop? or is that just a placeholder to demonstrate styling?

Hm well I thought it was an actual prop but I guess not. Will fix.

packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.types.ts
rurikoaraki: The union type I coped from somewhere else. Should probably figure out if it needs to be a union.

We could make it an Icon slot instead but since it's just for the checkmark (icon is a separate slot) I think it's okay for it to be more limited to start. If people want other icon types to be supported we can change it to Icon type.

ValentinaKozlova: I'm not sure that it should be union here. I don't remember why I made it union. Maybe I expected that bug with SVG icons would be fixed earlier. Anyway, I think now it's better to have here only XmlProps. I found this commit:
https://github.com/microsoft/fluentui-react-native/pull/969/commits/4112cbf8dba50d202e56fe821a5a3a258ff82a25

ValentinaKozlova (approved):

rurikoaraki:
> > is the spec to have `hasCheckmarks` be an actual prop? or is that just a placeholder to demonstrate styling?
> 
> Hm well I thought it was an actual prop but I guess not. Will fix.

@Saadnajmi Oh actually yeah it's a prop: 

https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-menu/Spec.md#menulist
and
https://github.com/microsoft/fluentui/blob/e74b180720cb9aec8df21753a55449849041af82/packages/react-components/react-menu/src/stories/MenuBestPractices.md
and
https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-menu/src/stories/MenuAligningWithSelectableItems.stories.tsx

packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.types.ts
rurikoaraki: `XmlProps` extends `SvgProps` so we don't need the union, just using `XmlProps` should be enough

Saadnajmi (approved):

Dewsk (approved):

PR #1659, opened 2022-05-13
V2 Text Initial implementation

lenahong:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes
- Defining and supporting new props to align with v2 Text
- Added variant types to use as stand alone
- Added tests for new props and variants in FURN tester

### Verification

Manually tested and more will be added after bug bash

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/TextExperimental/StandardUsage.tsx
```diff
@@ -1,30 +1,85 @@
...
+      </View>
+    );
+  };
+
+  switch (Platform.OS) {
+    case 'android':
+      return <V1Usage />;
+    case 'ios':
+      return <V1Usage />;
+    case 'macos':
```
rurikoaraki, line 75: nit: You could collapse the individual cases like:

```
case 'android':
case 'ios':
case 'macos':
    return <V1Usage />;
````

File: packages/utils/test-tools/package.json [hunk 2]
```diff
@@ -23,7 +23,8 @@
     "directory": "packages/utils/test-tools"
   },
   "dependencies": {
-    "@fluentui-react-native/theme-types": ">=0.14.3 <1.0.0"
+    "@fluentui-react-native/theme-types": ">=0.14.1 <1.0.0",
+    "@fluentui-react-native/theme-tokens": ">=0.16.1 <1.0.0"
```
rurikoaraki, line 27: >@fluentui-react-native/theme-tokens" [](http://example.com/codeflow?start=5&length=36)

I'm a bit surprised that this package needs theme-tokens,  esp since it seems like there' no other changes in here. What's it used for?

rurikoaraki:
>     return {};

If you want to add fontLineHeight as part of the font styles you will probably want to add it here as well

---
Refers to: packages/utils/tokens/src/text-tokens.ts:31 in 95bdd22. [](commit_id = 95bdd22bd83060743fc54e831c81783647e6047e, deletion_comment = False)

packages/experimental/Text/src/Variants.ts
```diff
@@ -0,0 +1,71 @@
...
+import { createFontAliasTokens } from '@fluentui-react-native/win32-theme';
+
+const variantsWin32: Partial<Variants> = createFontAliasTokens('light');
+
+export const Caption1 = Text.customize({
+  fontFamily: variantsWin32.caption1.face,
+  fontSize: variantsWin32.caption1.size,
+  fontWeight: variantsWin32.caption1.weight,
+});
+export const Body1 = Text.customize({
```
rurikoaraki, line 12: super nit: can we have a line between these?

packages/theming/win32-theme/src/createFontAliasTokens.ts
```diff
@@ -0,0 +1,11 @@
+import { getOfficeAliasTokens } from './getOfficeAliasTokens';
+import { Variants } from '@fluentui-react-native/theme-types';
+import { memoize } from '@fluentui-react-native/memo-cache';
+import { mapFontPipelineToTheme } from '@fluentui-react-native/theming-utils';
+
+function createFontAliasTokensWorker(officeTheme: string): Partial<Variants> {
```
rurikoaraki, line 6: >officeTheme: string [](http://example.com/codeflow?start=37&length=19)

Would the font tokens change between themes? Maybe we don't take a theme as input, and just hardcode fetching one specific alias token file here?

packages/experimental/Text/src/Variants.ts
```diff
@@ -0,0 +1,71 @@
+import { Variants } from '@fluentui-react-native/theme-types';
+import { Text } from './Text';
+import { createFontAliasTokens } from '@fluentui-react-native/win32-theme';
+
+const variantsWin32: Partial<Variants> = createFontAliasTokens('light');
```
rurikoaraki, line 5: >('light'); [](http://example.com/codeflow?start=62&length=10)

win32 doesn't use light so I'm surprised this works. Should be one of Colorful/White/DarkGray/Black/HighContrast

packages/experimental/Text/src/Variants.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const Caption1 = Text.customize({
+  fontFamily: variantsWin32.caption1.face,
```
rurikoaraki, line 8: >fontFamily [](http://example.com/codeflow?start=2&length=10)

I wonder if it would be simpler to use variant here instead of specifying fontFamily/fontSize/fontWeight separately?

packages/experimental/Text/src/Variants.ts
rurikoaraki, line 5: >variantsWin32 [](http://example.com/codeflow?start=6&length=13)

I would've expected to get this information from the theme object, is there a reason we can't do that? I think customize takes a function like:

```
Text.customize((t: Theme) => ({
  fontFamily: t.typography.variants.caption1.face,
}));
```

rurikoaraki: If users override the typography variants on the Theme object in order to change the font styles then that won't be respected if we integrate the font alias tokens this way

packages/experimental/Text/src/Text.tsx
```diff
@@ -1,83 +1,91 @@
...
   );
 
   // return a continuation function that allows this text to be compressed
   return (extra: TextProps, children: React.ReactNode) => {
-    const mergedProps = { ...rest, ...extra, style: mergeStyles(tokenStyle, props.style, extra?.style) };
+    const mergedProps = {
+      ...rest,
+      ...extra,
+      style: mergeStyles(tokenStyle, props.style, extra?.style),
+      numberOfLines: truncate ? 1 : wrap ? 0 : 1,
```
rurikoaraki, line 87: >truncate ? 1 : wrap ? 0 : 1, [](http://example.com/codeflow?start=21&length=28)

Could this be simplified to `wrap ? 0 : 1` and leave `truncate` out?

packages/experimental/Text/src/Text.tsx
```diff
@@ -1,83 +1,91 @@
...
+      break;
+  }
+  // override tokens from props
+  [tokens, cache] = patchTokens(tokens, cache, {
+    color,
+    variant,
+    fontSize: globalTokens.font.size[size],
+    fontWeight: globalTokens.font.weight[weight] as FontWeightValue,
+    fontStyle: props.italic ? 'italic' : 'normal',
+    textAlign: align == 'start' ? 'left' : align == 'end' ? 'right' : align,
```
rurikoaraki, line 60: > 'left'  [](http://example.com/codeflow?start=33&length=8)

Is this RTL-sensitive?

rurikoaraki (changes requested):
:clock1:

rurikoaraki:
If there's any changes to yarn.lock please make sure to commit that as well

packages/utils/test-tools/package.json
lenahong: Oh my bad. I think I added this when trying to define the default variants for the mockTheme as well but didn't make those changes after all. Removing!

packages/experimental/Text/src/Variants.ts
lenahong: Yeah, eventually we do want to do that. When we deprecate old variants and support v2 variant alias tokens for all platforms, we should define define V2 variants
 in defaultTypography and use the variant token to customize here. Right now, the issue is that I cannot define default token values for just win32 without causing an error on the other platforms I just made a note of that in this file as well as a comment.

packages/experimental/Text/src/Variants.ts
lenahong: Ah you're right. I'll have it pull from Theme object. createFontAliasTokens will eventually be used to set the defaultTypography tokens for win32

packages/experimental/Text/src/Text.tsx
lenahong: Updated code to check for rtl

packages/experimental/Text/src/Text.tsx
```diff
@@ -1,84 +1,95 @@
...
   const theme = useFluentTheme();
   // get the tokens from the theme
   let [tokens, cache] = useTokens(theme);
 
-  // apply states like disabled if specified in props
-  [tokens, cache] = applyTokenLayers(tokens, ['disabled'], cache, (state) => props[state]);
-  // override variant or color from props
-  [tokens, cache] = patchTokens(tokens, cache, { color, variant });
+  const textAlign = I18nManager.isRTL
+    ? align == 'start'
```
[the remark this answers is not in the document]
lenahong: since we need to check for exact strings, I think === is fine

packages/experimental/Text/src/Text.tsx
```diff
@@ -1,84 +1,95 @@
...
+    italic,
+    size,
+    strikethrough,
+    style,
+    truncate = false,
+    underline,
+    variant,
+    weight,
+    wrap = true,
+    onKeyDown,
```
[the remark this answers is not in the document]
lenahong: so can I just just revert this and have this prop pass through with '...rest' or should I do a platform check and pass it in to mergedProps only for win32/windows?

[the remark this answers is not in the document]
Saadnajmi: For now I would do a platform check. It would be a breaking change for macOS and I think this PR will merge faster than we can fix RN-macOS.

packages/experimental/Text/src/Text.tsx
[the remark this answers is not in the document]
Saadnajmi: but we're using `==` here?

rurikoaraki (approved):

PR #1677, opened 2022-05-20
Implement MenuItemRadio

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

MenuItemRadio support radio-button like selection. If a MenuItemRadio item is selected, selection is cleared from other MenuItemRadio/Checkbox items within the same MenuList. (This is a bug - I have an item filed to fix this in a separate PR)

~A lot of the visuals/logic for MenuItemRadio was copied from MenuItemCheckbox, I have an item to refactor all the MenuItem* components to reduce duplicate code.~ Using compose on MenuItemCheckbox so there's less duplication.

Main things are that the onClick (or equivalent) action will call selectRadio off the context, which in useMenuList sets the current selected component to the MenuItemRadio which was selected.

### Verification

radio

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/src/MenuItemRadio/useMenuItemRadio.ts
```diff
@@ -0,0 +1,90 @@
...
+    onAccessibilityAction,
+    ...rest
+  } = props;
+  const context = useMenuListContext();
+  const checked = context.checked?.[name];
+  const selectRadio = context.selectRadio;
+
+  const toggleChecked = React.useCallback(
+    (e: InteractionEvent) => {
+      selectRadio(e, name, !checked);
```
rurikoaraki, line 34: Here's the main difference from MenuItemCheckbox

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
```diff
@@ -0,0 +1,30 @@
...
+      </svg>`;
+
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.checkmark xml={chevronXml} />
+          {mergedProps.content && <Slots.content>{mergedProps.content}</Slots.content>}
+        </Slots.root>
+      );
+    };
+  },
```
Saadnajmi, line 29: What's different between the `useRenders` of MenuItemRadio and MenuItemCheckbox? Or is it that using `.compose` means you have to specify it twice?

packages/experimental/Menu/src/MenuItemRadio/useMenuItemRadio.ts
Saadnajmi: Would it be worth sharing the same hook and somehow passing a boolean`isRadio`? Either through the Context or the prop list? I'm ok if it makes sense to duplicate code, jus thought I'd ask.

Saadnajmi (approved):

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
rurikoaraki: useMenuItemRadio vs. useMenuItemCheckbox is the only difference.

But yeah, I have to re-specify the useRender again if I want to change it. I could refactor this too and maybe pass in a hook to run for state?

packages/experimental/Menu/src/MenuItemRadio/useMenuItemRadio.ts
rurikoaraki: Sure, we can do that. It'd be something that I'd have to pass in from the component though, not the context or prop list.

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
Saadnajmi: I feel like to do that, you would need to pass the resulting `useMenuItemCheckbox` hook an extra boolean or prop that does not match the final component API. AKA, the props passed into `MenuItemRadio / MenuItemCheckbox` would not match the props passed into `useMenuItemCheckbox`, since we need at least 1 extra one to specify if it's a radio or not. Is that OK? I feel like the intention of these state hooks we have is to have their input be equal to the final components input?

rurikoaraki: >  I feel like the intention of these state hooks we have is to have their input be equal to the final components input?

Dunno if I've ever heard of that. Is that a general pattern with state hooks?

Saadnajmi: If it's not, then I guess it's fine

rurikoaraki: I don't think it's an explicit rule, more of a general pattern? Anyway, I think I can work around it.

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const MenuItemRadio = MenuItemCheckbox.compose({
+  displayName: menuItemRadioName,
+  useRender: (userProps: MenuItemCheckboxProps, useSlots: UseSlots<MenuItemCheckboxType>) => {
+    const menuItem = useMenuItemRadio(userProps);
+    const Slots = useSlots(userProps, (layer): boolean => menuItem.state[layer]);
+
+    return (final: MenuItemCheckboxProps) => {
+      const mergedProps = mergeProps(menuItem.props, final);
+      const chevronXml = `
```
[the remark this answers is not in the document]
rurikoaraki: @lenahong I'll fix all the SVGs I added in its own PR, since there's quite a few of them

packages/experimental/Menu/src/MenuItemRadio/useMenuItemRadio.ts
rurikoaraki: Done w/ latest commit

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
rurikoaraki: I tried some refactoring, is it better? Or should I revert the .tsx files?

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -14,7 +14,33 @@ import { useMenuListContext } from '../context/menuListContext';
 const defaultAccessibilityActions = [{ name: 'Toggle' }];
 
 export const useMenuItemCheckbox = (props: MenuItemCheckboxProps): MenuItemCheckboxState => {
-  // attach the pressable state handlers
+  const { name } = props;
+  const context = useMenuListContext();
+  const checked = context.checked?.[name];
```
Saadnajmi, line 19: What's with the brackets around [name] ?

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -14,7 +14,33 @@ import { useMenuListContext } from '../context/menuListContext';
...
+
+const getAccessibilityState = memoize(getAccessibilityStateWorker);
+function getAccessibilityStateWorker(disabled: boolean, checked: boolean, accessibilityState?: AccessibilityState) {
+  if (accessibilityState) {
+    return { disabled, checked, ...accessibilityState };
+  }
+  return { disabled, checked };
+}
+
+export const useMenuCheckboxInteraction = (
```
Saadnajmi, line 40: So the new approach was to factor common code between `useMenuItemCheckbox / useMenuItemRadio` into this function and then call both from each? Cool!
Could we also add a simple one line comment saying so for the next dev looking at this code?

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -14,7 +14,33 @@ import { useMenuListContext } from '../context/menuListContext';
...
+  return useMenuCheckboxInteraction(props, toggleChecked);
+};
+
+const getAccessibilityState = memoize(getAccessibilityStateWorker);
+function getAccessibilityStateWorker(disabled: boolean, checked: boolean, accessibilityState?: AccessibilityState) {
+  if (accessibilityState) {
+    return { disabled, checked, ...accessibilityState };
+  }
+  return { disabled, checked };
+}
```
Saadnajmi, line 38: super nit: line 32 should be after line 33-38? Not sure it actually matters.

Saadnajmi (approved):

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
rurikoaraki: We want to access the property with the name that's the value of the name variable, not a property name.

i.e. if name is itemOne, we want to access `checked.itemOne`, which `checked.[name]` lets us do, not `checked.name`

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
rurikoaraki: Generally doesn't matter. I've usually done memo before the function and not after.

Saadnajmi: Mk, I'll resolve this then.

lenahong (approved):

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
rurikoaraki: Added a header to the function which should add context

packages/experimental/Menu/src/MenuItemRadio/MenuItemRadio.tsx
Saadnajmi: (Post merge comment) I think it's better? Hard to say without locally cloning the PR myself, since both versions are pretty involved changes =) . I'll leave that as a problem for the next time I have to touch the code :D

PR #1680, opened 2022-05-21
Implement gap token on MenuList

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

MenuList has a gap between items on web. Implementing that token on MenuList.

Stack would be handy to use in this situation, since it has a gap prop. Unfortunately the Stack component that's in FURN is not working on the v1 framework (no idea why) and I'd like to avoid using v0 components where possible. It would make sense to transition to using a Stack component for this in the future but for now, the gap logic was fairly easy to implement so I just added a hidden MenuStack component.

### Verification
gap set to something like 6:

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -1,18 +1,45 @@
...
-import { useMenuListContextValue } from './useMenuContextValue';
+import { useMenuListContextValue } from './useMenuListContextValue';
+import { IViewProps } from '@fluentui-react-native/adapters';
+
+const MenuStack = stagedComponent((props: React.PropsWithRef<IViewProps> & { gap?: number }) => {
+  const { gap, ...rest } = props;
+  return (final: React.PropsWithRef<IViewProps> & { gap?: number }, children: React.ReactNode) => {
+    if (gap && gap > 0 && children) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore - TODO, fix typing error
```
Saadnajmi, line 17: Actual TODO or fine to keep?

Saadnajmi (approved):

packages/experimental/Menu/src/MenuList/MenuList.tsx
rurikoaraki: I copied this from Stack logic. I'll log an issue and replace the TODO.

packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -1,18 +1,45 @@
...
+      children = React.Children.map(children, (child: React.ReactChild, index: number) => {
+        if (React.isValidElement(child) && index > 0) {
+          return React.cloneElement(
+            child,
+            mergeProps(child.props, {
+              style: {
+                ...{ marginTop: gap },
```
ValentinaKozlova, line 24: Why do you need here this syntax? Why not `style: { marginTop: gap }`?

ValentinaKozlova (approved):

PR #1681, opened 2022-05-21
Focus first element on popover open

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Focus should move to the first element when the popover appears.
Easy fix, need to add "setInitialFocus" to the wrapped Callout

### Verification

focus

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-menu-f53969f8-621e-4877-b705-4830ea208547.json
+++ b/change/@fluentui-react-native-menu-f53969f8-621e-4877-b705-4830ea208547.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Focus first element when popover opens",
+  "packageName": "@fluentui-react-native/menu",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
+++ b/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
@@ -16,6 +16,7 @@ export const MenuPopover = stagedComponent((props: MenuPopoverProps) => {
         target={state.triggerRef}
         onDismiss={state.onDismiss}
         dismissBehaviors={state.dismissBehaviors}
+        setInitialFocus={true}
       >
         {children}
       </Callout>
--- a/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
+++ b/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
@@ -3,16 +3,12 @@ import { stagedComponent, useFluentTheme } from '@fluentui-react-native/framewor
 import { Callout } from '@fluentui-react-native/callout';
 import { menuPopoverName, MenuPopoverProps } from './MenuPopover.types';
 import { useMenuPopover } from './useMenuPopover';
-import { Platform } from 'react-native';
 
 export const MenuPopover = stagedComponent((props: MenuPopoverProps) => {
   const state = useMenuPopover(props);
   const theme = useFluentTheme();
 
   return (_rest: MenuPopoverProps, children: React.ReactNode) => {
-    // Initial focus behavior differs per platform, Windows platforms move focus
-    // automatically onto first element of Callout
-    const setInitialFocus = Platform.OS === ('win32' as any) || Platform.OS === 'windows';
     return (
       <Callout
         borderWidth={1}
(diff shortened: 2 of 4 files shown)


Saadnajmi:
This is unfortunately one of the quirks the macOS Callout doesn't do well. I had to basically hardcode the behavior in ContextualMenu:  https://github.com/microsoft/fluentui-react-native/blob/36ea0d81629f9326eb8c23912d9949381014dade/packages/components/ContextualMenu/src/ContextualMenu.tsx#L44 . Probably should have an action item on me to investigate/fix that.

Related: Is there not a `shouldFocusOnMount` prop equivalent in V1 menu? In the old ContextualMenu, it was programmatic whether you placed focus when the Callout was shown, and `shouldFocusOnContainer` determined if we focus on the first menu item, or the containing view instead. I guess we just got rid of both of those levels of control? (fine by me, though macOS and win32 differ slightly in expectations)

Saadnajmi (approved):
Regardless of the quirks I mentioned, this change should be fine

rurikoaraki:
> This is unfortunately one of the quirks the macOS Callout doesn't do well. I had to basically hardcode the behavior in ContextualMenu:
> 
> https://github.com/microsoft/fluentui-react-native/blob/36ea0d81629f9326eb8c23912d9949381014dade/packages/components/ContextualMenu/src/ContextualMenu.tsx#L44
> 
> . Probably should have an action item on me to investigate/fix that.
> Related: Is there not a `shouldFocusOnMount` prop equivalent in V1 menu? In the old ContextualMenu, it was programmatic whether you placed focus when the Callout was shown, and `shouldFocusOnContainer` determined if we focus on the first menu item, or the containing view instead. I guess we just got rid of both of those levels of control? (fine by me, though macOS and win32 differ slightly in expectations)

Oh interesting, didn't know about the differences here. I can expose `shouldFocusOnMount` as a prop for V1 Menu and add some default behavior per platform? The Menu spec does call out that these props are not on the Menu anymore, but we can change that if there are platform expectations.

Saadnajmi:
> > This is unfortunately one of the quirks the macOS Callout doesn't do well. I had to basically hardcode the behavior in ContextualMenu:
> > https://github.com/microsoft/fluentui-react-native/blob/36ea0d81629f9326eb8c23912d9949381014dade/packages/components/ContextualMenu/src/ContextualMenu.tsx#L44
> > 
> > . Probably should have an action item on me to investigate/fix that.
> > Related: Is there not a `shouldFocusOnMount` prop equivalent in V1 menu? In the old ContextualMenu, it was programmatic whether you placed focus when the Callout was shown, and `shouldFocusOnContainer` determined if we focus on the first menu item, or the containing view instead. I guess we just got rid of both of those levels of control? (fine by me, though macOS and win32 differ slightly in expectations)
> 
> @Saadnajmi Oh interesting, didn't know about the differences here. I can expose `shouldFocusOnMount` as a prop for V1 Menu and add some default behavior per platform? The Menu spec does call out that these props are not on the Menu anymore, but we can change that if there are platform expectations.

I would keep it out for now. I don't want to expose those props if we get rid of them :P. We can instead just have custom code per platform to do what we need. We can talk more offline about this too if need be, I think I might have some assumptions that are wrong

rurikoaraki:
> > > This is unfortunately one of the quirks the macOS Callout doesn't do well. I had to basically hardcode the behavior in ContextualMenu:
> > > https://github.com/microsoft/fluentui-react-native/blob/36ea0d81629f9326eb8c23912d9949381014dade/packages/components/ContextualMenu/src/ContextualMenu.tsx#L44
> > > 
> > > . Probably should have an action item on me to investigate/fix that.
> > > Related: Is there not a `shouldFocusOnMount` prop equivalent in V1 menu? In the old ContextualMenu, it was programmatic whether you placed focus when the Callout was shown, and `shouldFocusOnContainer` determined if we focus on the first menu item, or the containing view instead. I guess we just got rid of both of those levels of control? (fine by me, though macOS and win32 differ slightly in expectations)
> > 
> > 
> > @Saadnajmi
> >             
> >             Saad Najmi
> >             FTE Oh interesting, didn't know about the differences here. I can expose `shouldFocusOnMount` as a prop for V1 Menu and add some default behavior per platform? The Menu spec does call out that these props are not on the Menu anymore, but we can change that if there are platform expectations.
> 
> I would keep it out for now. I don't want to expose those props if we get rid of them :P. We can instead just have custom code per platform to do what we need. We can talk more offline about this too if need be, I think I might have some assumptions that are wrong

@SaadNajmi done with latest commit.

PR #1685, opened 2022-05-23
Add delay to hover behavior

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds a default delay value for the onHoverIn/onHoverOut behaviors. It also adds the onHoverOut behavior which uses the newer doNotTakePointerCapture prop on the Callout (Note that the behavior is not visible yet and is buggy with the built-in version of rex-win32, we have to update it to get the correct behavior, which is currently gated behind 0.66 update).

Delay values were taken from the ContextMenu.

### Verification

hover

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
+++ b/packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
@@ -1,6 +1,7 @@
 import { useMenuContext } from '../context/menuContext';
 import { InteractionEvent } from '@fluentui-react-native/interactive-hooks';
 import { MenuTriggerProps } from './MenuTrigger.types';
+import { Platform } from 'react-native';
 
 export const useMenuTrigger = (_props: MenuTriggerProps) => {
   const context = useMenuContext();
@@ -10,9 +11,20 @@ export const useMenuTrigger = (_props: MenuTriggerProps) => {
   const openOnHover = context.openOnHover;
   const triggerRef = context.triggerRef;
 
+  const delayHover = Platform.select({
+    macos: 100,
+    default: 500, // win32
+  });
+
   const onHoverIn = (e: InteractionEvent) => {
     if (openOnHover) {
       setOpen(e, true /* isOpen */);
+    }
+  };
+
+  const onHoverOut = (e: InteractionEvent) => {
+    if (openOnHover) {
+      setOpen(e, false /* isOpen */);
     }
   };
 
@@ -20,5 +32,5 @@ export const useMenuTrigger = (_props: MenuTriggerProps) => {
     setOpen(e, !open);
   };
 
-  return { onClick, onHoverIn, componentRef: triggerRef };
+  return { onClick, onHoverIn, onHoverOut, componentRef: triggerRef, delayHoverIn: delayHover, delayHoverOut: delayHover };
 };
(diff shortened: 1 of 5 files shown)


packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -10,15 +11,26 @@ export const useMenuTrigger = (_props: MenuTriggerProps) => {
...
+      setOpen(e, false /* isOpen */);
+    }
+  };
+
   const onClick = (e: InteractionEvent) => {
     setOpen(e, !open);
   };
 
-  return { onClick, onHoverIn, componentRef: triggerRef };
+  return { onClick, onHoverIn, onHoverOut, componentRef: triggerRef, delayHoverIn: delayHover, delayHoverOut: delayHover };
```
Saadnajmi, line 35: Just to double check: `delayHoverIn/Out` are forwarded to `useAsPressable`, and this works for submenus too? I remember I had to do a lot of working around that in ContextualMenu

packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -10,15 +11,26 @@ export const useMenuTrigger = (_props: MenuTriggerProps) => {
   const openOnHover = context.openOnHover;
   const triggerRef = context.triggerRef;
 
+  const delayHover = Platform.select({
+    macos: 100,
+    default: 500, // win32
+  });
```
Saadnajmi, line 17: These delays were for when a submenu is opened off a menu, not for the main menu. I don't know we'd always want the delay for the main menu on macOS. Is that a windows / win32 requirement?

rurikoaraki: There's a short delay for opening a menu on hover in win32 though it was for submenus so I don't know if it's a strict requirement. Let me ask PM.

packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
Saadnajmi: I guess submenus is out of scope for this PR. The old ContextualMenu had quite a bit of hardcoding around onHoverIn to present submenus so I mostly wanted to make sure we avoided that. I'll close this comment as out of scope

lenahong (approved):

packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
Saadnajmi: I asked if there was any example of an "openOnHover" scenari for a ContextMenu on macOS, and we didn't know of any. So at least on macOS, we don't have an opinion / prior art to make an opinion. Since this delay will (presumably?) be reused for submenu, I guess it's OK

Saadnajmi (approved):

packages/experimental/Menu/src/MenuTrigger/useMenuTrigger.ts
rurikoaraki: win32 doesn't really have a specific scenario for it either. it was part of the spec, so in it went 🤷🏻‍♀️

Anyway PM says that it'd probably be useful to have so we'll add it for now. We can make it win32 only if we need to.

Saadnajmi: Eh.. I feel like there's a difference between "This is not supported on the platform" (engineering) vs "This is not a platform paradigm" (design). I personally prefer to mark props as win32/macOS only if they truly aren't supported. So I think it's OK.

PR #1686, opened 2022-05-24
Change Submenu chevron to use an embedded svg that can flip rtl and recolor

lenahong:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes
Previously, the chevron to indicate the submenu item was implemented as an Svgxml. This prevented the chevron color to be directly set by the Theme object. The current method allows the color to be read from the incoming color prop and will appropriately adapt to dark/hc mode as well.

### Verification
chevron dark mode

### Pull request checklist

- [X] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover
- [X] Internationalization and Right-to-left Layouts


.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "CurrentProjectSetting": "5upConfig"
```
rurikoaraki, line 2: Don't think we need these .vs files... Maybe they should be added to .gitignore?

Same with .suo, Browse.VC.db, and slnx.sqlite

packages/components/ContextualMenu/src/@types/ContextualMenu/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '*.svg' {
+  import { SvgProps } from 'react-native-svg';
```
rurikoaraki, line 2: Could you explain what this does/what it's needed for?

Saadnajmi: I want to say this and the metro config should not be needed. Generally that's set up per app, not per control. These files are a bunch of stuff to get metro to recognize svg files as source instead of assests, which react-native-svg needs to run

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,6 @@
+{
```
Saadnajmi, line 1: Delete this file too?

packages/components/ContextualMenu/src/SubmenuItem.tsx
```diff
@@ -119,6 +119,19 @@ export const SubmenuItem = compose<SubmenuItemType>({
...
+        <Svg width={props.width} height={props.height} viewBox="0 0 2048 2048" color={props.color}>
+          <G transform={I18nManager.isRTL ? 'translate(2048, 0) scale(-1, 1)' : ''}>
+            <Path fill="currentColor" d="M 743 1767 l -121 -121 l 708 -707 l -708 -708 l 121 -121 l 828 829 z"></Path>
+          </G>
+        </Svg>
+      );
+    };
+    const svgProps: SvgIconProps = {
+      src: chevronSvg,
+    };
```
Saadnajmi, line 133: Thanks for this fix! I really like the encapsulation into an IconProps to be really nice :) 
Out of scope of this PR, I wonder: How would you feel if we introduced something like a "Flips in RTL" prop to Icon in the future, so that Icon would then be in charge of checking `I18nManager` and adding a transform? I'm not sure it's feasible with SVGs but it's been on my mind a bit as a generic solution.

packages/components/ContextualMenu/src/@types/ContextualMenu/index.d.ts
lenahong: Ah this change is no longer necessary since we're not using a separate svg file.

rurikoaraki (approved):

PR #1687, opened 2022-05-24
Datepicker - add ability to update the reference start/end dates

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Ask from a client team to be able to set the reference start/end dates (range of possible dates the calendar allows, not to be confused with "date range" which refers to the dates a user can select). 

The default reference start date is Jan 1, 3 years before the current year. The default reference end date is 10 years after the default reference start date. Ex. For May 23, 2022, the default range is Jan 1, 2019 - Dec 21, 2029.

This change allows the default Calendar Configuration to be overridden - which includes the reference start and end dates, as well as firstWeekday (usually begins on Sunday or Monday). Note that overriding the default calendar configuration affects **all** FURN datepickers. Additional work will be required if we want to support different calendar configurations for different datepickers.

### Verification

Added a button at the bottom of the FluentTester DatePicker test page that updates the default reference start/end dates to be from 1999 to 2000.

| Before | After |
| image | image | image
 

| Before (earliest possible start date you could scroll to) | After (earliest possible start date you could scroll to) |
| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

Notes:
- Not sure if using a button in the FluentTester datepicker page to change the default reference start end dates is best. Maybe a toggle would be better?
- This allows users to set the default "CalendarConfiguration", but this PR only exposes referenceStartDate and referenceEndDate. Do we also want to expose firstWeekday?
- With this change it is easier to get into a state where the selected date is outside the reference start/end dates, ex if the default start date is already out of the range, and you try to select the default date. There is an assert in datetimepickerviewdatasource (line 308, assert(dateIndex >= 0 && dateIndex < numberOfDates, "index out of bound")) that gets hit when this happens. Do we want to do anything about this?

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/NativeDatePicker/NativeDatePickerTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/NativeDatePicker/NativeDatePickerTest.tsx
@@ -1,4 +1,5 @@
 import { Button } from '@fluentui-react-native/button';
+import { Checkbox } from '@fluentui/react-native';
 import { NativeDatePicker } from '@fluentui-react-native/experimental-native-date-picker';
 import { NATIVEDATEPICKER_TESTPAGE } from './consts';
 import { PlatformStatus, Test, TestSection } from '../Test';
@@ -22,6 +23,10 @@ const NativeDatePickerMainTest: React.FunctionComponent = () => {
     endDate: NativeDatePicker.parseISOString('2020-03-07T11:55:00.000Z'),
     referenceStartDate: NativeDatePicker.parseISOString('1999-01-01T00:00:00.000Z'),
     referenceEndDate: NativeDatePicker.parseISOString('2000-01-01T00:00:00.000Z'),
+    firstWeekday: 2,
+    defaultReferenceStartDate: new Date(new Date().getFullYear() - 3, 1, 1),
+    defaultReferenceEndDate: new Date(new Date().getFullYear() + 7, 1, 1),
+    defaultFirstWeekday: 1,
   };
   const titles = {
     startTitle: 'Start Title',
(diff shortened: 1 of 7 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/NativeDatePicker/NativeDatePickerTest.tsx
```diff
@@ -36,14 +38,14 @@ const NativeDatePickerMainTest: React.FunctionComponent = () => {
...
+      <Text variant="headerStandard">Selected Start Date/Time</Text>
       <Text variant="subheaderStandard">
         {startDate?.toString()} {'\n'}
       </Text>
 
-      <Text variant="headerStandard">End Date/Time</Text>
+      <Text variant="headerStandard">Selected End Date/Time</Text>
       <Text variant="subheaderStandard">
-        {endDate?.toString()} {'\n'}
+        {endDate?.toString() ?? 'N/A'} {'\n'}
```
lyzhan7, line 48: These changes aren't really related to being able to update the reference start/end dates - I made the labels more clear since "start date" is kind of ambiguous and also it was strange that if there was no end date, no placeholder text would show up. I can move this into another PR too

Saadnajmi:
Awesome, This looks great!  
A few notes:

- We generally don't mention specific partners or link internal work items in open source PRs. More often, we say "A client team asked..." =)
- The <Switch> control would work well as a toggle. We use it elsewhere.
- Might as well add it? Since it's the only other property you could configure
- I'm inclined to say it's ok to have that error, since that feels like a "The API was misused" error. If we don't like it, we could try catching the assert but I'm ok with leaving it as is.

Saadnajmi (approved):

lyzhan7:
@Saadnajmi Thank you for your notes - updated the description, and decided a checkbox made more sense than a toggle since there was a visible difference when it was on or off. Added firstWeekday to it as well.

Updated screenshot:
image

Saadnajmi:
> @Saadnajmi Thank you for your notes - updated the description, and decided a checkbox made more sense than a toggle since there was a visible difference when it was on or off. Added firstWeekday to it as well.
> 
> Updated screenshot: image

I think the iOS designs for Checkbox are not very finalized (iirc, the chevron is like super tiny?) and also I think the Apple HIG _technically_ wants to use Switches over checkboxes in most cases. This is me being picky, but would you actually change that to a Switch? I know @harrieshin mentioned before we shouldn't really use Checkbox on iOS much at all.

lyzhan7:
Updated screenshot - using switch, removed firstWeekday

Screen Shot 2022-05-24 at 4 09 40 PM

Saadnajmi (approved):

PR #1689, opened 2022-05-24
Implement Submenu properly for Menu V1

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change shores up the submenu implementation.

1. Submenu triggers adds openOnHover behavior by default
2. MenuPopover changes its `directionalHint` if they are a submenu in an internationalization-friendly way
3. A menuTriggerContext allows for MenuItems to identify if they are a submenu trigger. The chevron is changed to be a right-facing chevron which is flipped when in RTL. (Unfortunately the transform doesn't apply unless we're on the latest version of win32 native, so either we duplicate the icon or we hold back on Menu until 0.66 is done).

### Verification

submenu

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/src/MenuItem/MenuItem.tsx
```diff
@@ -23,8 +23,8 @@ export const MenuItem = compose<MenuItemType>({
     return (final: MenuItemProps) => {
       const mergedProps = mergeProps(menuItem.props, final);
       const chevronXml = `
-      <svg width="12" height="16" viewBox="0 0 11 6">
-        <path fill='currentColor' d='M0.646447 0.646447C0.841709 0.451184 1.15829 0.451184 1.35355 0.646447L5.5 4.79289L9.64645 0.646447C9.84171 0.451185 10.1583 0.451185 10.3536 0.646447C10.5488 0.841709 10.5488 1.15829 10.3536 1.35355L5.85355 5.85355C5.65829 6.04882 5.34171 6.04882 5.14645 5.85355L0.646447 1.35355C0.451184 1.15829 0.451184 0.841709 0.646447 0.646447Z' />
+      <svg>
+        <path fill='currentColor' d='M4.64645 2.14645C4.45118 2.34171 4.45118 2.65829 4.64645 2.85355L7.79289 6L4.64645 9.14645C4.45118 9.34171 4.45118 9.65829 4.64645 9.85355C4.84171 10.0488 5.15829 10.0488 5.35355 9.85355L8.85355 6.35355C9.04882 6.15829 9.04882 5.84171 8.85355 5.64645L5.35355 2.14645C5.15829 1.95118 4.84171 1.95118 4.64645 2.14645Z' />
```
rurikoaraki, line 27: All the SVGs will be cleaned up in a separate PR.

packages/experimental/Menu/src/Menu/useMenu.ts
```diff
@@ -9,8 +9,10 @@ export const useMenu = (props: MenuProps): MenuState => {
   const isSubmenu = context.triggerRef !== null;
   const isControlled = typeof props.open !== 'undefined';
   const [open, setOpen] = useMenuOpenState(isControlled, props);
+  const openOnHover = isSubmenu;
```
Saadnajmi, line 12: We don't want to do something like `const openOnHover = props.openOnHover || isSubmenu` ? or `(props.openOnHover !=  undefined) ? props.openOnHover : isSubmenu`

packages/experimental/Menu/src/Menu/useMenu.ts
```diff
@@ -9,8 +9,10 @@ export const useMenu = (props: MenuProps): MenuState => {
...
 
   return {
+    openOnHover,
     ...props,
     open,
```
Saadnajmi, line 17: I think if we do it this order and `props` contains a key named `openOnHover`, it will overwrite the value you provide on like 15. Is that intentional?

packages/experimental/Menu/src/MenuItem/MenuItem.types.ts
```diff
@@ -10,6 +10,8 @@ export const menuItemName = 'MenuItem';
 
 export interface MenuItemTokens extends LayoutTokens, FontTokens, IBorderTokens, IColorTokens {
   checkmarkSize?: number;
+  chevronPadding?: number;
+  chevronSize?: number;
```
Saadnajmi, line 14: Do these apply to the submenuIndicator?

packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -13,10 +13,23 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
...
+const getDirectionalHint = (isSubmenu: boolean, isRtl: boolean): DirectionalHint | undefined => {
+  if (!isSubmenu) {
+    return undefined;
+  }
+
+  if (isRtl) {
+    return 'leftTopEdge';
+  }
+
+  return 'rightTopEdge';
```
Saadnajmi, line 34: I might have hardcoded this switch in native code for macOS Callout. I'll have to double check these two don't conflict with each other.

packages/experimental/Menu/src/MenuTrigger/MenuTrigger.types.ts
```diff
@@ -1,7 +1,18 @@
-import type { IViewProps } from '@fluentui-react-native/adapters';
+import { InteractionEvent, IWithPressableOptions } from '@fluentui-react-native/interactive-hooks';
+import { ViewProps } from 'react-native';
 
 export const menuTriggerName = 'MenuTrigger';
 
-export interface MenuTriggerProps extends Omit<IViewProps, 'onPress'> {}
+export interface MenuTriggerProps extends Omit<IWithPressableOptions<ViewProps>, 'onPress'> {
+  /**
+   * A RefObject to access the IButton interface. Use this to access the public methods and properties of the component.
```
Saadnajmi, line 8: Where is "Button" defined? Is that comment still relevant?

packages/experimental/Menu/src/MenuTrigger/MenuTrigger.types.ts
```diff
@@ -1,7 +1,18 @@
...
+   */
+  componentRef?: React.RefObject<React.Component>;
 
-export interface MenuTriggerState {}
+  onClick?: (e: InteractionEvent) => void;
+}
+
+export interface MenuTriggerState {
+  props: MenuTriggerProps;
+  isSubmenu: boolean;
```
Saadnajmi, line 17: This means that the menu trigger will host a submenu if `isSubmenu` is true? And we need to store that to know if we want to override `openOnHover` by default?

packages/experimental/Menu/src/Menu/useMenu.ts
rurikoaraki: The `...props` below will provide openOnHover behavior if someone wants it for a parent menu or override a submenu's default openOnHover if someone wants to disable that behavior. But this is probably unclear, I can add a comment here on intended behavior.

packages/experimental/Menu/src/Menu/useMenu.ts
rurikoaraki: See answer to comment above

packages/experimental/Menu/src/MenuItem/MenuItem.types.ts
rurikoaraki: Yeah. I can rename the tokens to be that instead.

packages/experimental/Menu/src/MenuTrigger/MenuTrigger.types.ts
rurikoaraki: Yeah, the MenuTrigger will host a submenu if isSubmenu is true. This is actually needed to figure out if a submenu indicator should be shown on a MenuItem. Just knowing isSubmenu from the larger menu context isn't enough, because then all MenuItems in the submenu (including the ones in the submenu's popover) will get a submenu indicator. So the MenuItem needs to know both that it's in a submenu and that it's a trigger for a submenu.

packages/experimental/Menu/src/Menu/useMenu.ts
Saadnajmi: Yeah I realized that after finishing the review. I think if we want to keep the "use the spread to override the default" we should add a comment. I'm biased because I don't think it's obvious to many that spreads can overwrite previous props. Feel free to disagree and keep it as is if you prefer the brevity.

packages/experimental/Menu/src/MenuItem/MenuItem.types.ts
Saadnajmi: I was actually going to ask if we should rename the submenuIndicator slot to "chevron". Since we don't have separate MenuItem and SubmenuItem controls anymore, I think the decriptive name of "submenuIndicator" is slightly better, so renaming the control tokens would be nice.

packages/experimental/Menu/src/MenuTrigger/MenuTrigger.types.ts
Saadnajmi: Could we name it something like "hasSubmenu" or "hostsSubmenu" instead? I feel like "isSubmenu" feels like a relic of when we had a separate SubmenuItem control. If there's a precedent for "isSubmenu" in Fluent web, then I guess it's fine.

rurikoaraki: Sure. The naming scheme is from FluentUI but it's not a big deal to change things internally.

Saadnajmi (approved):

PR #1696, opened 2022-05-26
Icon accesibility update

warren-ms:
Add accessibilityLabel as an IconProp and set the right props to make Icon accessibile in narrator

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Icon/src/Icon.tsx
+++ b/packages/experimental/Icon/src/Icon.tsx
@@ -16,7 +16,8 @@ function renderRasterImage(iconProps: IconProps) {
     rasterImageStyleCache({ width: width, height: height, tintColor: color }, [width, height, color])[0],
   );
 
-  return <Image source={iconProps.rasterImageSource.src} style={style} />;
+  const accessibilityLabel = iconProps.accessibilityLabel ?? 'Image';
+  return <Image source={iconProps.rasterImageSource.src} style={style} accessible={true} accessibilityRole="image" accessibilityLabel={accessibilityLabel}/>;
 }
 
 function fontFamilyFromFontSrcFile(fontSrcFile: string, fontFamily: string): string {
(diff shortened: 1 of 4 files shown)


packages/experimental/Icon/src/Icon.tsx
```diff
@@ -16,7 +16,8 @@ function renderRasterImage(iconProps: IconProps) {
     rasterImageStyleCache({ width: width, height: height, tintColor: color }, [width, height, color])[0],
   );
 
-  return <Image source={iconProps.rasterImageSource.src} style={style} />;
+  const accessibilityLabel = iconProps.accessibilityLabel ?? 'Image';
```
Saadnajmi, line 19: This string isn't localized, so I'd rather not have the default and instead do something like set the accessibilityRole to "image"

Saadnajmi (changes requested):

packages/experimental/Icon/src/Icon.tsx
rurikoaraki: Yeah +1 on not having any hardcoded non-localized strings

rurikoaraki: Removed (Warren is OOF so I'm taking this over as it's needed urgently)

ksiler (approved):

packages/experimental/Icon/src/Icon.tsx
```diff
@@ -16,7 +16,15 @@ function renderRasterImage(iconProps: IconProps) {
...
   );
 
-  return <Image source={iconProps.rasterImageSource.src} style={style} />;
+  return (
+    <Image
+      source={iconProps.rasterImageSource.src}
+      style={style}
+      accessible={true}
+      accessibilityRole="image"
+      accessibilityLabel={iconProps.accessibilityLabel}
```
Saadnajmi, line 25: Post checkin comment: To avoid this in the future, we could do something where we pass in `...props` so that the user can specify any prop. I think that's a pattern we use elsewhere to avoid similar bugs.

PR #1699, opened 2022-05-26
Menu V1 keyboarding

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change shores up some keyboarding behavior on the Menu. Spec that has been implemented is:

- Up and down inside of the menu that has focus should navigate vertically between keyboard focusable elements (note: focus should never land on the "split" part of a split button this way)
- Up and down should circularly navigate through a list of menu items by default, e.g. if I hit down at the bottom of a menu, it should navigate to the first item.
- Right key on a menu item that has an available submenu or split button should invoke the submenu, and move focus into it onto the first focusable item (flipped for RTL)
- Left key when in a submenu should close the submenu and place focus back on the parent menu's item that opened the submenu. (flipped for RTL)
- ESC on a menu should close the currently focused menu. If you are in a submenu and hit ESC, you should only close the current submenu and return focus to the parent's element where focus was previously.
- Space and Enter key execute the action that is currently focused in the menu.

Additional fixes:
- Disabled visuals are now applied when disabled is passed in as a prop
- Close the Menu on MenuItem invoke

### Verification

Used tester to validate keyboard behavior

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { AccessibilityState } from 'react-native';
 import { MenuItemProps, MenuItemState } from './MenuItem.types';
 import { memoize } from '@fluentui-react-native/framework';
-import { useAsPressable, useKeyProps } from '@fluentui-react-native/interactive-hooks';
+import { InteractionEvent, useAsPressable, useKeyProps } from '@fluentui-react-native/interactive-hooks';
 import { useMenuContext } from '../context/menuContext';
 import { useMenuListContext } from '../context/menuListContext';
 import { useMenuTriggerContext } from '../context/menuTriggerContext';
@@ -11,8 +11,17 @@ export const useMenuItem = (props: MenuItemProps): MenuItemState => {
   // attach the pressable state handlers
   const defaultComponentRef = React.useRef(null);
   const { onClick, accessibilityState, componentRef = defaultComponentRef, disabled, ...rest } = props;
-  const pressable = useAsPressable({ ...rest, disabled, onPress: onClick });
-  const onKeyProps = useKeyProps(onClick, ' ', 'Enter');
+
+  const setOpen = useMenuContext().setOpen;
+  const onInvoke = React.useCallback(
+    (e: InteractionEvent) => {
+      onClick && onClick(e);
+      setOpen(e, false /*isOpen*/);
+    },
+    [onClick, setOpen],
+  );
+  const pressable = useAsPressable({ ...rest, disabled, onPress: onInvoke });
+  const onKeyProps = useKeyProps(onInvoke, ' ', 'Enter');
   const isTrigger = useMenuTriggerContext();
   const hasSubmenu = useMenuContext().isSubmenu && isTrigger;
   const hasCheckmarks = useMenuListContext().hasCheckmarks;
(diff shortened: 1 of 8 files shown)


packages/experimental/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -1,34 +1,71 @@
...
+        ) {
+          return;
+        }
+        if (
+          isKeyPressEvent(e) &&
+          isInSubmenu &&
+          ((I18nManager.isRTL && e.nativeEvent.key === 'ArrowLeft') || (!I18nManager.isRTL && e.nativeEvent.key === 'ArrowRight'))
+        ) {
+          return;
+        }
```
ValentinaKozlova, line 39: If we have just `return` in both conditions maybe it's better to put it to one? And long conditions with RTL put to variables? I'm just wondering. Not sure if it's better

ValentinaKozlova (approved):

packages/experimental/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -1,34 +1,74 @@
...
+      }
+      if (
+        isKeyPressEvent(e) &&
+        isInSubmenu &&
+        ((isRtl && e.nativeEvent.key === 'ArrowLeft') || (!isRtl && e.nativeEvent.key === 'ArrowRight'))
+      ) {
+        return;
+      }
+
+      onClick && onClick(e);
```
Saadnajmi, line 45: `onClick?.(e)` also works here I think

packages/experimental/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -1,34 +1,74 @@
...
+      }
+    },
+    [disabled, hasSubmenu, isInSubmenu, onClick, setOpen],
+  );
+
+  const pressable = useAsPressable({ ...rest, disabled, onPress: onInvoke });
+  const keys = isSubmenu ? submenuTriggerKeys : triggerKeys;
+
+  // Explicitly override onKeyDown to override the native windows behavior of moving focus with arrow keys.
+  const onKeyDownProps = useKeyDownProps(onInvoke, ...keys);
```
Saadnajmi, line 57: Nit: Also native macOS behavior, so we can just take out "windows" =)

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -14,16 +14,18 @@ import { useMenuListContext } from '../context/menuListContext';
 const defaultAccessibilityActions = [{ name: 'Toggle' }];
 
 export const useMenuItemCheckbox = (props: MenuItemCheckboxProps): MenuItemCheckboxState => {
-  const { name } = props;
+  const { disabled, name } = props;
```
Saadnajmi, line 17: Do we want to also pull out `...rest`  and pass it into a root slot somewhere below?

packages/experimental/Menu/src/MenuItemRadio/useMenuItemRadio.ts
```diff
@@ -5,16 +5,18 @@ import { MenuItemCheckboxProps, MenuItemCheckboxState } from '../MenuItemCheckbo
 import { useMenuCheckboxInteraction } from '../MenuItemCheckbox/useMenuItemCheckbox';
 
 export const useMenuItemRadio = (props: MenuItemCheckboxProps): MenuItemCheckboxState => {
-  const { name } = props;
+  const { disabled, name } = props;
```
Saadnajmi, line 8: similar comment here

Saadnajmi (approved):

packages/experimental/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
rurikoaraki: That should be taken care of in useMenuCheckboxInteraction on line 50.

PR #1709, opened 2022-06-01
Fix hover-related issues with Menu

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change:
1. Fixes behavior where hover color doesn't apply until after the delay passes. This is because pressable's hover state doesn't get set until after the timeout passes in Pressable.ts, and that state drives the coloring of the button/menuItem. The fix to this is to handle the delay in the component's hoverIn/hoverOut instead of passing the delay to useAsPressable.
2. Fixes behavior where the menu will close prematurely if the menu is set to open on hover and you move the mouse into the menu (where you would expect it to stay open). This is because the timeout to close the menu is set once the trigger is hovered out of, and there isn't anything stopping it from executing when the mouse enters the popover. The fix to this (thanks @PPatBoyd) is to cancel the timeout on the trigger once you hover into the popover. This can be done by storing the timer on the menu context - then both the trigger and the popover have access to the timer.
3. The Popover now needed to handle mouse events in order to clear timers on hover in and close the menu on hover out. This necessitated an additional View in the MenuPopover under the Callout since the Callout doesn't handle mouse events. The same clearing of the timeout needs to be done in the opposite direction (moving mouse from popover to trigger shouldn't cause the menu to close), so the popover's timer is also on the context.
4. Submenus add an additional layer of timeout complexity since there's two timeouts that get triggered when you go from a popover to a submenu's popover - the close submenu timeout that gets set on hovering out of the submenu's trigger (which is handled by 2), and the close menu timeout that gets set on hovering out of the menu's popover. To fix this, we store the parent's popover timer when we create the submenu on the context, and cancel it on hovering into a menu's popover so that the menu doesn't close prematurely.

Like I said, shenanigans.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
+++ b/packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { stagedComponent, useFluentTheme } from '@fluentui-react-native/framework';
+import { mergeProps, stagedComponent, useFluentTheme } from '@fluentui-react-native/framework';
 import { Callout } from '@fluentui-react-native/callout';
 import { menuPopoverName, MenuPopoverProps } from './MenuPopover.types';
 import { useMenuPopover } from './useMenuPopover';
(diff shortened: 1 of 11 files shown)


packages/experimental/Menu/src/MenuPopover/MenuPopover.tsx
```diff
@@ -21,7 +22,9 @@ export const MenuPopover = stagedComponent((props: MenuPopoverProps) => {
         directionalHint={state.directionalHint}
         doNotTakePointerCapture={state.doNotTakePointerCapture}
       >
-        {children}
+        <View onMouseEnter={state.onMouseEnter} onMouseLeave={state.onMouseLeave}>
```
rurikoaraki, line 25: This will change to adaptable view, gotta make this slots first.

rurikoaraki: Not a pressable/View as pressable as I don't care about pressable state, just tracking mouse in/out

rurikoaraki: Decided to just use React.createElement, slots/compose seemed overkill

packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -3,25 +3,64 @@ import { I18nManager, Platform } from 'react-native';
...
+  }, [parentPopoverHoverOutTimer, popoverHoverOutTimer, triggerHoverOutTimer]);
+  const onMouseLeave = React.useCallback(() => {
+    if (!openOnHover) {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      setOpen(undefined, false /* isOpen */);
+    }, 500);
+    console.log('popoverout');
```
rurikoaraki, line 45: Will remove

PR #1714, opened 2022-06-02
Menu v1: macOS tokens

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

1) Adds the styling tokens to get Menu V1 as close as ContextualMenu as I could. Note that a lot of tokens were hardcoded rather than coming from the pipeline

~~2) Adds a FocusZone slot to MenuList which is only used on macOS. We need this to move selection with arrow keys~~
EDIT: Removing this for now

3) Changes MenuItem behavior so that onHoverIn sets keyboard focus. This is to ensure we only ever have one menu item highlighted even if both the mouse and keyboard focus are on a menu at once.

### Verification

Tested on macOS. It's pretty close :) 

https://user-images.githubusercontent.com/6722175/171732433-a96e313a-1ac8-4ab7-89a0-5de0e910fbd3.mov

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Menu/src/MenuTrigger/MenuTrigger.tsx
```diff
@@ -22,7 +22,7 @@ export const MenuTrigger = stagedComponent((props: MenuTriggerProps) => {
     // state in the inner render so we can access the child component and its props.
     const child = childrenArray[0];
     const revisedState = getRevisedState(menuTrigger, child.props);
-    const revised = React.cloneElement(child, revisedState);
+    const revised = React.cloneElement(child, revisedState.props);
```
rurikoaraki, line 25: One of my PRs might fix this too, so it's possible there'll be a conflict here in the future.

Saadnajmi: I can wait for your PR's to merge and undo this bit then

packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +35,53 @@ export const MenuList = compose<MenuListType>({
...
-      return (
-        <MenuListProvider value={contextValue}>
-          <Slots.root>{children}</Slots.root>
-        </MenuListProvider>
-      );
+      if (Platform.OS === 'macos') {
+        return (
+          <MenuListProvider value={contextValue}>
+            <Slots.root>
+              <Slots.focusZone
```
rurikoaraki, line 65: I think it would be cleaner/less duplicate to so something like

```tsx
const content = Platform.OS === 'macos' ? (
         <Slots.focusZone
                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                // @ts-ignore
                enableFocusRing={false}
                componentRef={focusZoneRef}
                defaultTabbableElement={focusZoneRef}
                focusZoneDirection={'vertical'}
              >
                {children}
              </Slots.focusZone>
) : (
    {children}
)

return (
<MenuListProvider value={contextValue}>
            <Slots.root>
                {content}
            </Slots.root>
</MenuListProvider>
);
```

Less duplication of the render code which means less likely that things will be missed if we need to add things here.

packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +35,53 @@ export const MenuList = compose<MenuListType>({
...
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
```
rurikoaraki, line 67: Can we add a comment on why this is needed

packages/experimental/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +35,53 @@ export const MenuList = compose<MenuListType>({
...
     const menuList = useMenuList(userProps);
     const contextValue = useMenuListContextValue(menuList);
     const Slots = useSlots(menuList);
 
+    /**
+     * On macOS, focus isn't placed by default on the first focusable element. We get around this by focusing on the inner FocusZone
+     * hosting the menu. For whatever reason, to get the timing _just_ right to actually focus, we need an additional `setTimeout`
+     *  on top of the `useLayoutEffect` hook.
+     */
+    const focusZoneRef = React.useRef<IFocusable>(null);
```
rurikoaraki, line 50: I think it would be better to group all the hook code under useMenuList, so that it's all in one place to find.

packages/experimental/Menu/src/MenuList/MenuList.types.ts
```diff
@@ -1,10 +1,11 @@
 import type { IViewProps } from '@fluentui-react-native/adapters';
+import { FocusZoneProps } from '@fluentui-react-native/focus-zone';
 import { InteractionEvent } from '@fluentui-react-native/interactive-hooks';
-import { IBackgroundColorTokens, LayoutTokens } from '@fluentui-react-native/tokens';
+import { IBackgroundColorTokens, IBorderTokens, LayoutTokens } from '@fluentui-react-native/tokens';
 
 export const menuListName = 'MenuList';
 
-export interface MenuListTokens extends LayoutTokens, IBackgroundColorTokens {
+export interface MenuListTokens extends LayoutTokens, IBorderTokens, IBackgroundColorTokens {
```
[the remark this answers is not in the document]
Saadnajmi: Hmm, I think I added this for borderRadius but I do see that we don't actually set that as a token. Padding was what I mostly cared about, which we should get from LayoutTokens. I can revert this change.

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: I think I tried that before and ran into weird typing errors. I'll try once more.

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: Sure. On macOS, FocusZone inherits from all of ViewProps because natively, `FRNFocusZone` is a subclass of `RCTView` (the native class that backs `<View>`). As such, we can set `enableFocusRing` on it. I think on win32, FocusZone does _not_ inherit all of View props, and thus I didn't extend FocusZoneProps to include all of ViewProps. 

Something to revisit if we ever do FocusZoneV1 I suppose. I'll update the comment here.

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: Right. I think I got tripped up copying code from `usePrepareProps` in the old compose framework to here. I'll update accordingly. I think the `useLayoutEffect` call might have to stay in the render of MenuList though?

packages/experimental/Menu/src/MenuTrigger/MenuTrigger.tsx
rurikoaraki: Latest merged PR addresses this issue.

packages/experimental/Menu/src/MenuList/MenuList.tsx
rurikoaraki: Why's that? Is there a rule that it has to be in a tsx file?

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: This worked, thank you!

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: I was wrong. I thought it had to be in render but I guess you can call hooks from hooks.

packages/experimental/Menu/src/MenuList/MenuList.tsx
Saadnajmi: Updated. An aside: It seems even win32 FocusZone should inherit ViewProps natively so we can perhaps change the type. Feels out of scope for this PR so I'll address that separately.

rurikoaraki (approved):

Saadnajmi:
Removing the FocusZone changes for now, as they were breaking CI and I want to get part of this change in bc it's been around for a while.

rurikoaraki (approved):

PR #1724, opened 2022-06-07
Fix some Menu Radio bugs

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR fixes 3 bugs with MenuRadio and makes some API changes:
1. Selecting a radio item clears out selection from all items including checkbox items
2. onCheckedChange was not stating all checked state when radio item was selected
3. Selection state doesn't persist between Menu sessions

To fix 1, I added an array to track all current radio items and functions to add or remove items from that list. The array may be a problem with longer lists, may want to change it into a lookup table in the future. If an item is not in that list, its selection is not changed when running selectRadio.
To fix 2, I changed the onCheckedChange API to return a list of currently selected controls when a control changes, to cover cases where selection changes via toggle and selection changes via radio selection (which would be multiple events). We can change this to fire individual events in the future for each item, but it's more difficult to implement compared to returning the current state of checked components, and this API somewhat aligns with web as well.
To fix 3, currently selected items are preserved at the Menu level as well as the MenuList level so that selection is preserved across Menu sessions.

### Verification

radio_item_fix

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Will add more description and screenshots later

ValentinaKozlova (approved):

ksiler:
Not sure if you're looking to address this in this PR or another but it seems like the accessibilityState isn't being updated correctly when a new item is checked and another is unchecked. Narrator correctly announces the new checked item as being "checked" but when you place focus on the last checked item is thinks that is still "checked" too. It's not until you close the menu and reopen it that the correct checked state is reflected on all items.

rurikoaraki:
> Not sure if you're looking to address this in this PR or another but it seems like the accessibilityState isn't being updated correctly when a new item is checked and another is unchecked. Narrator correctly announces the new checked item as being "checked" but when you place focus on the last checked item is thinks that is still "checked" too. It's not until you close the menu and reopen it that the correct checked state is reflected on all items.

Thanks for finding this, I'll take a look

rurikoaraki:
> > Not sure if you're looking to address this in this PR or another but it seems like the accessibilityState isn't being updated correctly when a new item is checked and another is unchecked. Narrator correctly announces the new checked item as being "checked" but when you place focus on the last checked item is thinks that is still "checked" too. It's not until you close the menu and reopen it that the correct checked state is reflected on all items.
> 
> Thanks for finding this, I'll take a look

@ksiler So the bug shows up in our version of rex-win32 but with the latest native changes the bug is fixed. So I think this will correct on its own once we're able to update our version of rex/native code in the tester.

ksiler:
> > > Not sure if you're looking to address this in this PR or another but it seems like the accessibilityState isn't being updated correctly when a new item is checked and another is unchecked. Narrator correctly announces the new checked item as being "checked" but when you place focus on the last checked item is thinks that is still "checked" too. It's not until you close the menu and reopen it that the correct checked state is reflected on all items.
> > 
> > 
> > Thanks for finding this, I'll take a look
> 
> @ksiler So the bug shows up in our version of rex-win32 but with the latest native changes the bug is fixed. So I think this will correct on its own once we're able to update our version of rex/native code in the tester.

Good to know, thanks for checking!

PR #1732, opened 2022-06-07
Pin versions of components in core package

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Our core package's package.json allows for any type past the base version to be installed as a depenency, so that would be why it's fetching the latest version.
 
....which seems like a bug, because then we can't confidently direct people on which versions of theming/framework to use

Beachball should keep things up to date for us so pinning the version should be safe.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Maybe I should pin versions in other packages too?

FalseLobster (approved):

PR #1733, opened 2022-06-07
Close all menus when submenu item clicked

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding an optional "bubble" option which will call setOpen on parent menus with the same args until you hit the root menu. In this case, bubble is set to true on the setOpen call in the MenuItem, which will close all parent menus when a submenu item is invoked.

This then led to flickering behavior for the keyboard as the keyboard event would get sent to the trigger as well, causing it to reopen the menu. To prevent this, I added a delay or cooldown buffer of 150ms, during which the menu doesn't reopen if it was closed within the last 150ms. This also fixes the issue where the menu reopens if you click on the trigger button to dismiss it.

### Verification

| submenu_close_bug | close_submenu_all |
| click_close_bug | close_menu_trigger_frix |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/Menu/useMenu.ts
+++ b/packages/experimental/Menu/src/Menu/useMenu.ts
@@ -8,7 +8,7 @@ export const useMenu = (props: MenuProps): MenuState => {
   const context = useMenuContext();
   const isSubmenu = context.triggerRef !== null;
   const isControlled = typeof props.open !== 'undefined';
-  const [open, setOpen] = useMenuOpenState(isControlled, props);
+  const [open, setOpen] = useMenuOpenState(isControlled, props, context.setOpen);
 
   // Default behavior for submenu is to open on hover
   // the ...props line below will override this behavior for a submenu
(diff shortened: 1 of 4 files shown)


ValentinaKozlova (approved):

packages/experimental/Menu/src/Menu/useMenu.ts
```diff
@@ -33,24 +40,38 @@ export const useMenu = (props: MenuProps): MenuState => {
...
+          lastCloseTimestamp = Date.now();
+        }
       }
       if (onOpenChange && openPrev !== isOpen) {
         onOpenChange(e, isOpen);
       }
+
+      if (bubble && parentSetOpen) {
+        parentSetOpen(e, isOpen, bubble);
+      }
```
Saadnajmi, line 72: Post PR review: I wonder if we could have made an actual custom bubbling React Event, so that we can get the bubbling behavior for free? This is fine though

rurikoaraki: I'm not sure what you're suggesting, are you talking about like adding a new useBubble hook?

Saadnajmi: https://reactjs.org/docs/events.html

I'm suggesting something like creating an event named "onMenuDismiss" that is fired from the submenu, and bubbles up to the root menu. then we could have an event handler prop like 
```
onMenuDismiss={() => {
  if (rootMenu) { e.stopPropogation; }//stop it from bubbling if we're the top level menu.
  menu.dismisss(); // dismiss ourselves
  }
```

PR #1737, opened 2022-06-10
Focus should move on hovering over Menu Items

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR attempts to address issues with focus moving while hovering through items.

Unfortunately, React doesn't differentiate between mouse and keyboard hovering, so we have to resort to tricks to get the visuals we want on win32, which is that the focus ring should not show when mousing over items but should show when keyboarding through them.

The way I attempted to address this is by setting enableFocusRing to match isHovered, and then use a useEffect call to focus on a component when the hover state changes. Calling for focus changes in onHoverIn/onHoverOut calls focus too soon and so it doesn't quite work the way we'd like.

It does lead to some weird behaviors though so I'm passing this through PM before committing.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
@@ -2,7 +2,13 @@ import * as React from 'react';
 import { AccessibilityState, I18nManager } from 'react-native';
 import { MenuItemProps, MenuItemState } from './MenuItem.types';
 import { memoize } from '@fluentui-react-native/framework';
-import { InteractionEvent, isKeyPressEvent, useAsPressable, useKeyDownProps } from '@fluentui-react-native/interactive-hooks';
+import {
+  InteractionEvent,
+  isKeyPressEvent,
+  useAsPressable,
+  useKeyDownProps,
+  useViewCommandFocus,
+} from '@fluentui-react-native/interactive-hooks';
 import { useMenuContext } from '../context/menuContext';
 import { useMenuListContext } from '../context/menuListContext';
 import { useMenuTriggerContext } from '../context/menuTriggerContext';
@@ -20,6 +26,7 @@ export const useMenuItem = (props: MenuItemProps): MenuItemState => {
   const isInSubmenu = isSubmenu && !isTrigger;
 
   const setOpen = useMenuContext().setOpen;
+
   const onInvoke = React.useCallback(
     (e: InteractionEvent) => {
       if (disabled) {
@@ -51,11 +58,20 @@ export const useMenuItem = (props: MenuItemProps): MenuItemState => {
   );
 
   const pressable = useAsPressable({ ...rest, disabled, onPress: onInvoke });
+  const itemRef = useViewCommandFocus(componentRef);
   const keys = isSubmenu ? submenuTriggerKeys : triggerKeys;
 
   // Explicitly override onKeyDown to override the native behavior of moving focus with arrow keys.
   const onKeyDownProps = useKeyDownProps(onInvoke, ...keys);
   const hasCheckmarks = useMenuListContext().hasCheckmarks;
+
+  React.useEffect(() => {
+    if (pressable.state.hovered) {
+      componentRef?.current?.focus();
+    } else {
+      componentRef?.current?.blur();
+    }
+  }, [pressable.state.hovered, componentRef]);
 
   return {
     props: {
@@ -65,9 +81,9 @@ export const useMenuItem = (props: MenuItemProps): MenuItemState => {
       onAccessibilityTap: props.onAccessibilityTap || onInvoke,
       accessibilityLabel: props.accessibilityLabel || props.content,
       accessibilityState: getAccessibilityState(disabled, accessibilityState),
-      enableFocusRing: true,
+      enableFocusRing: !pressable.state.hovered,
       focusable: true,
-      ref: componentRef,
+      ref: itemRef,
       ...onKeyDownProps,
     },
     state: pressable.state,
--- a/change/@fluentui-react-native-menu-e1449a6c-0102-4aa8-9a1e-bc2fb3e66328.json
+++ b/change/@fluentui-react-native-menu-e1449a6c-0102-4aa8-9a1e-bc2fb3e66328.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "A fix for hover focus?",
+  "packageName": "@fluentui-react-native/menu",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}


Saadnajmi:
> Calling for focus changes in onHoverIn/onHoverOut calls focus too soon and so it doesn't quite work the way we'd like.

Why is that? I removed that change from #1714 since it seems it would conflict with this change. What about the timing is wrong?

Saadnajmi:
> Calling for focus changes in onHoverIn/onHoverOut calls focus too soon and so it doesn't quite work the way we'd like.

Why is that? I removed that change from #1714 since it seems it would conflict with this change. What about the timing is wrong?

Saadnajmi:
> Unfortunately, React doesn't differentiate between mouse and keyboard hovering

Sorry I've asked this before, but I'll ask again: What is this extra state in win32? From my understanding, there are 3 states a View can be in:

- Hovered (AKA, mouse is on top of the control. This is an addition to React Native Core in the desktop forks. This ice usuallly reflected with the control getting highlighted)
- Focused (AKA, Keyboard focus. The user moved focus to the view via pressing tab / arrow keys / whatever else. This is usually reflected with a "focus ring" or the control being highlighted in some way)
- Pressed / Selected (AKA, user clicked / touched the control. This ones easy).

I remember you mentioned that win32, in addition to keyboard focus has "mouse focus". Just to be sure, that's _not_ the same thing as hover, correct? Your comment also seems to say there is a "keyboard hover" state. Is that the same thing as keyboard focus?

Sorry for all the questions. I'm hoping it'll help me (and others) better understand platform semantics across macOS and win32.

Saadnajmi:
> Unfortunately, React doesn't differentiate between mouse and keyboard hovering

Sorry I've asked this before, but I'll ask again: What is this extra state in win32? From my understanding, there are 3 states a View can be in:

- Hovered (AKA, mouse is on top of the control. This is an addition to React Native Core in the desktop forks. This ice usuallly reflected with the control getting highlighted)
- Focused (AKA, Keyboard focus. The user moved focus to the view via pressing tab / arrow keys / whatever else. This is usually reflected with a "focus ring" or the control being highlighted in some way)
- Pressed / Selected (AKA, user clicked / touched the control. This ones easy).

I remember you mentioned that win32, in addition to keyboard focus has "mouse focus". Just to be sure, that's _not_ the same thing as hover, correct? Your comment also seems to say there is a "keyboard hover" state. Is that the same thing as keyboard focus?

Sorry for all the questions. I'm hoping it'll help me (and others) better understand platform semantics across macOS and win32.

packages/experimental/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -51,12 +58,21 @@ export const useMenuItem = (props: MenuItemProps): MenuItemState => {
...
   const onKeyDownProps = useKeyDownProps(onInvoke, ...keys);
   const hasCheckmarks = useMenuListContext().hasCheckmarks;
 
+  React.useEffect(() => {
+    if (pressable.state.hovered) {
+      componentRef?.current?.focus();
+    } else {
+      componentRef?.current?.blur();
+    }
+  }, [pressable.state.hovered, componentRef]);
```
Saadnajmi, line 74: Do we also need this for MenuItemCheckBox / MenuItemRadio?

rurikoaraki: Yeah, we do. Thanks for catching.

rurikoaraki:
> > Unfortunately, React doesn't differentiate between mouse and keyboard hovering
> 
> Sorry I've asked this before, but I'll ask again: What is this extra state in win32? From my understanding, there are 3 states a View can be in:
> 
> * Hovered (AKA, mouse is on top of the control. This is an addition to React Native Core in the desktop forks. This ice usuallly reflected with the control getting highlighted)
> * Focused (AKA, Keyboard focus. The user moved focus to the view via pressing tab / arrow keys / whatever else. This is usually reflected with a "focus ring" or the control being highlighted in some way)
> * Pressed / Selected (AKA, user clicked / touched the control. This ones easy).
> 
> I remember you mentioned that win32, in addition to keyboard focus has "mouse focus". Just to be sure, that's _not_ the same thing as hover, correct? Your comment also seems to say there is a "keyboard hover" state. Is that the same thing as keyboard focus?
> 
> Sorry for all the questions. I'm hoping it'll help me (and others) better understand platform semantics across macOS and win32.

@ksiler would probably have a better explanation? But mouse focus differs from hover. I forget if mouse focus usually moves with hover or if it usually only moves in response to mouse click events, but the way I think of it is, there's one control that has "focus". "mouse" vs "keyboard" denotes how the control got focus, i.e. whether it was clicked or tabbed to. Depending on the source of focus, we may show different visuals. So clicking on a control doesn't invoke focus borders on controls but it does now have mouse focus. If you then pressed tab, focus would move to the next element and not start at the first element, because focus was moved to the clicked control, and show thick borders, because tab denotes that it now has keyboard focus.

ksiler:
> > > Unfortunately, React doesn't differentiate between mouse and keyboard hovering
> > 
> > 
> > Sorry I've asked this before, but I'll ask again: What is this extra state in win32? From my understanding, there are 3 states a View can be in:
> > 
> > * Hovered (AKA, mouse is on top of the control. This is an addition to React Native Core in the desktop forks. This ice usuallly reflected with the control getting highlighted)
> > * Focused (AKA, Keyboard focus. The user moved focus to the view via pressing tab / arrow keys / whatever else. This is usually reflected with a "focus ring" or the control being highlighted in some way)
> > * Pressed / Selected (AKA, user clicked / touched the control. This ones easy).
> > 
> > I remember you mentioned that win32, in addition to keyboard focus has "mouse focus". Just to be sure, that's _not_ the same thing as hover, correct? Your comment also seems to say there is a "keyboard hover" state. Is that the same thing as keyboard focus?
> > Sorry for all the questions. I'm hoping it'll help me (and others) better understand platform semantics across macOS and win32.
> 
> @ksiler would probably have a better explanation? But mouse focus differs from hover. I forget if mouse focus usually moves with hover or if it usually only moves in response to mouse click events, but the way I think of it is, there's one control that has "focus". "mouse" vs "keyboard" denotes how the control got focus, i.e. whether it was clicked or tabbed to. Depending on the source of focus, we may show different visuals. So clicking on a control doesn't invoke focus borders on controls but it does now have mouse focus. If you then pressed tab, focus would move to the next element and not start at the first element, because focus was moved to the clicked control, and show thick borders, because tab denotes that it now has keyboard focus.

Sorry for the late response. I'm not an expert on this but I think mouse focus does move with mouse hover but I don't know if they're one and the same. In my experience, if the mouse is over a control it has mouse focus and you don't need to click on it for it to receive mouse focus. Adding @PPatBoyd to see if he knows more.

PPatBoyd:
> 



> > > > Unfortunately, React doesn't differentiate between mouse and keyboard hovering
> > > 
> > > 
> > > Sorry I've asked this before, but I'll ask again: What is this extra state in win32? From my understanding, there are 3 states a View can be in:
> > > 
> > > * Hovered (AKA, mouse is on top of the control. This is an addition to React Native Core in the desktop forks. This ice usuallly reflected with the control getting highlighted)
> > > * Focused (AKA, Keyboard focus. The user moved focus to the view via pressing tab / arrow keys / whatever else. This is usually reflected with a "focus ring" or the control being highlighted in some way)
> > > * Pressed / Selected (AKA, user clicked / touched the control. This ones easy).
> > > 
> > > I remember you mentioned that win32, in addition to keyboard focus has "mouse focus". Just to be sure, that's _not_ the same thing as hover, correct? Your comment also seems to say there is a "keyboard hover" state. Is that the same thing as keyboard focus?
> > > Sorry for all the questions. I'm hoping it'll help me (and others) better understand platform semantics across macOS and win32.
> > 
> > 
> > @ksiler would probably have a better explanation? But mouse focus differs from hover. I forget if mouse focus usually moves with hover or if it usually only moves in response to mouse click events, but the way I think of it is, there's one control that has "focus". "mouse" vs "keyboard" denotes how the control got focus, i.e. whether it was clicked or tabbed to. Depending on the source of focus, we may show different visuals. So clicking on a control doesn't invoke focus borders on controls but it does now have mouse focus. If you then pressed tab, focus would move to the next element and not start at the first element, because focus was moved to the clicked control, and show thick borders, because tab denotes that it now has keyboard focus.
> 
> Sorry for the late response. I'm not an expert on this but I think mouse focus does move with mouse hover but I don't know if they're one and the same. In my experience, if the mouse is over a control it has mouse focus and you don't need to click on it for it to receive mouse focus. Adding @PPatBoyd to see if he knows more.

Apologies for the late response lost track of this tab -- in the general routing-target-bubbling event propagation pattern I would call the 'target' element of the mouse/keyboard event the 'focused' element.  'Hovered' is more accurately synonymous with 'Mouse Within', if the element is any element on the routing-bubbling element path from root to 'target' and wants to react to that change in state.  'Key Within' or 'Focus In' would be a keyboard variation of 'hovered'.  The 'Key Within' state is important to help users visually track key focus e.g. a list of comment threads would have individually comment threads highlight when focus is on a text input inside the thread, creating an easier to track visual than just the text input's focus rect and caret.

The important events for 'hovered' (mouse focus) are onMouseOver, onMouseEnter, and onMouseLeave
The important events for 'pressed' (mouse/pointer press) are onPressDown, onPressUp, and onPress (in other contexts possibly 'mouse' or 'click')
The important events for 'focused' (key focus) are onFocus, onBlur, and in some contexts also onFocusIn and onFocusOut

With a menu the two become a little contentious; generally a mouse move should have the uniquely-hovered menu item become the target of any mouse or keyboard event.  On arrow key/tab the key focus should move in the appropriate manner to change the target of a keyboard event but _not_ a mouse event.  A press event may target neither.  I'm saying this from my own thoughts and may not be accurate for existing behavior on all native platforms :)

- If I open a menu and hover item 3, clicking or pressing enter/space should activate item 3 -- mouse focus changing in the menu has implied key focus changing.
- If I then tap 'arrow down' to move key focus to item 4, clicking should activate item 3 but pressing enter/space should activate item 4 -- key focus changing in the menu does not imply mouse focus changing.
- A pressIn could come in for item 1 -- we would go immediately to a 'pressed' state without a 'hovered' state rendered, and I'm not specifically sure but would suppose to receive mouse move/enter/leave events in the same manner as if we had moved the mouse from item 3 to item 1, without any intermediate events for elements in between due to the pointer "teleporting".
- An onPress should fire in kind with onPressOut only with an appropriate on onPressIn target -- some of the exceptions I'm implying don't really affect menus due to their transient nature, but if I pressIn on item 1, then move the pointer to item 2, mouse and key focus moves to item 2 and will activate item 2 on either pressOut or space/enter.
- The main exception to the above that's important for good UX is allowing users to "undo" their press intent e.g. press in on a button, move the pointer off of the button, and press out not activating the button.  Inverted, pressing in away from a button, moving the pointer to the button, and pressing out should likely not activate the button's onClick/onPress.  Further exceptions for hit slop.
- Menus as a group of pressable items take the appropriate twist on this behavior to honor the press out action on the menu button so long as the press in occurred inside the menu.  If our menu had unclickable areas I would expect pressOut on this unclickable area to no-op and not dismiss the menu, but that's more exotic than we've designed for here.  A pressIn on a menu item with a pressOut outside of the menu should dismiss the menu with no other action.


... I said a lot more than I expected to and think I covered the space without a big gap, but that's all with the caveats of defining appropriate press/mouse/key interactions for good user experiences that may not be consistent between platforms today depending on the context.

Saadnajmi:
Thanks @PPatBoyd for your explanation and everyone else for theirs =)

It sounds like there _isn't_ really a concept of separate mouse and keyboard focus, there's mouse hover and keyboard focus. And with the 3 states of "mouse hover, keyboard focus, press", we can achieve all the states we want to for Menus.

The current implementation of `ContexutalMenu` supports your bullet points 1 - 5, but not 6. With this change, the new V1 `Menu` component should also support the same. Bullet point 6 is also the same behavior natively on macOS, but I couldn't quite get that working in React Native.

rurikoaraki:
Okay PM is good with the behavior here for the most part. Separate issues will be tracked for the small bugs I saw.

packages/experimental/Menu/src/MenuItem/useMenuItem.ts
Saadnajmi: Is this still an issue? I wasn't sure if your last refactor meant this is shared now

rurikoaraki: Ah the refactoring only merged the types, I didn't touch any of the render impl since they're somewhat different between the two components.

rurikoaraki: Anyway thanks for catching, added the behavior to MenuItemCheckbox (and Radio) too.

Saadnajmi (approved):

PR #1742, opened 2022-06-13
Added notification test to fluent tester

joannaquu:
### Platforms Impacted
- [x ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added test for FURN notification component

### Verification

Screen Shot 2022-06-14 at 11 24 37 AM

### Pull request checklist

- [ ] Automated Tests
- [x ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,183 @@
+import * as React from 'react';
+// import { Notification } from
+import { View, Text } from 'react-native';
+import { Test, TestSection, PlatformStatus } from '../Test';
+// import { ColorValue } from 'react-native';
```
Saadnajmi, line 5: Nit: remove commented out lines. 

We can work on importing `Notification` from it's own package as a followup PR =)

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+
+const PrimaryToastTest: React.FunctionComponent = () => {
```
Saadnajmi, line 8: Some notes:

1) "Toast" is the Android name for a notification: https://material.io/archive/guidelines/components/snackbars-toasts.html#snackbars-toasts-usage . I think they popularized the concept, so the name gets used in a lot of places (including our own designs). However, the right "FluentUI" name for this component is "Notification". That's also reflected in the name of the FluentUI Apple control. Could we change the name to "Notification"


2) I like that you have separate tests for each variant (Primary, Neutral, etc...)  of Notification! There's a lot of common code between them. Could we separate out the common code into one function component called `Notification`? Then `Notification` could takes one prop called "variant" and we could choose the styles / colors based on that. We could also give it more props for the strings inside each Text component on lines 37-38.

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,183 @@
...
+      fontSize: 16,
+      flexDirection: 'row',
+      justifyContent: 'flex-start',
+    },
+    right: {
+      fontSize: 16,
+      fontWeight: '500',
+      flexDirection: 'row',
+      justifyContent: 'flex-end',
+    },
```
Saadnajmi, line 32: A better name for these would be "start" and "end". This is because the order of the content can flip in a Right-To-Left (RTL) language, like Hebrew or Arabic. Notice that `justify-content` on lines 25 and 31 use `flex-start` and flex-end` for this reason.

Saadnajmi:
Could you add a screenshot of the test page along with a description of your changes to the PR notes so that future people who look at this change know what was happening? :)

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+// const NotificationVariant string: 'primary' | 'neutral' | 'danger' | 'warning';
+enum NotificationVariant { primary = 1, neutral, danger, warning };
+
+const Notification = (props) => {
+  const variant = props.variant;
+  let startText = ''; 
+  let endText = '';
+
+  const styles: {[key: string]: any} = {
```
Saadnajmi, line 13: I think it would be better if we used "StyleSheet.create()", so that we get proper type checking

https://reactnative.dev/docs/stylesheet

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,139 @@
...
+  if (variant === 'primary') {
+    styles.container['backgroundColor'] = 'skyblue';
+    styles.container['justifyContent'] = 'space-between';
+    
+    startText = 'Mail Archived';
+    endText = 'Undo';
+  }
+  else if (variant === 'neutral') {
+    styles.container['backgroundColor'] = 'lightgrey';
+    styles.container['alignItems'] = 'center';
```
Saadnajmi, line 39: I'd rather you only construct the style once, rather than create it once and then modify it. You could do something like:
```ts
let containerBackgroundColor = 'clear';
switch(variant) {
  case 'neutral:
    containerBackgroundColor = 'skyBlue';
    break;
  case 'warning':
    ....
}

let styles = StyleSheet.create({
  ...
  container: {
  ...
  backgroundColor: containerBackgroundColor;
  },
}

```

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,131 @@
+import * as React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+import { Test, TestSection, PlatformStatus } from '../Test';
+
+// const NotificationVariant string: 'primary' | 'neutral' | 'danger' | 'warning';
+enum NotificationVariant { primary = 1, neutral, danger, warning };
```
Saadnajmi, line 6: Can we actually use the union over the enum? Some info on why unions are preferred:

https://blog.bam.tech/developer-news/should-you-use-enums-or-union-types-in-typescript

https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant

Saadnajmi: I think you just need to declare it as a type. 

From `pacakges/experimental/Avatar/src/Avatar.tsx`: 
``` 
export type Size = 'xSmall' | 'small' | 'medium' | 'large' | 'xLarge' | 'xxLarge';
```
But also, now that I understand it better, I like what `packages/experimental/Badge/src/Badge.types` does too:

```
export const BadgeSizes = ['smallest', 'smaller', 'small', 'medium', 'large', 'largest'] as const;
export type BadgeSize = typeof BadgeSizes[number];
```

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,131 @@
...
+      flexDirection: 'row',
+      justifyContent: 'space-between',
+      alignItems: 'center',
+      padding: 16,
+    },
+    start: {
+      fontSize: 16,
+      color: foregroundColor,
+
+      width: 0,
```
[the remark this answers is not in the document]
joannaquu: Not completely sure but react native has a lot of difficulty with wrapping text (there's a flexWrap: 'wrap' but it doesn't work). Having width: 0, flexGrow: 1, flex: 1 was the only workaround I could find that would actually wrap the text.

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,131 @@
...
+      break;
+  };
+
+  let styles = StyleSheet.create({
+    container: {
+      width: null,
+      borderRadius: 12,
+      backgroundColor: containerBackgroundColor,
+  
+      flexDirection: 'row',
```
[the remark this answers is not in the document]
joannaquu: Hmm commenting it out does change the look so I assume it is working but I'll add in flex: 1 for clarity

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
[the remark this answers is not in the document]
joannaquu: never mind just deleted it and it works now!

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <View style={styles.container}>
+      <Text style={styles.start}>{startText}</Text>
+      <Text style={styles.end}>{endText}</Text>
+    </View>
+  );
+};
+
+const PrimaryTest: React.FunctionComponent = () => {
+  return <Notification variant={NotificationVariants[0]} startText="Mail Archived" endText="Undo" />;
```
Saadnajmi, line 68: Why can't you just do `variant={'primary'}` ?

PR #1748, opened 2022-06-14
Expose enableFocusRing as an explicit prop

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Apparently the base ViewProp doesn't include enableFocusRing, it's only on the desktop platforms. So people can't override enableFocusRing if they want to use their own focus visuals.

Explicitly exposing enableFocusRing.

### Verification

Tested with tester, ensured code compiled

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

ValentinaKozlova (approved):

Saadnajmi:
A thought for future components since I've seen errors like this before:

Is there a system where we can just pass '...rest' to the root slot of a component, and expose view props at the top level? Or maybe instruct clients to use customize() if they run into this issue in the future? 

The former suggestion I _think_ lines up with the official React guidance on higher order components? https://reactjs.org/docs/higher-order-components.html#convention-pass-unrelated-props-through-to-the-wrapped-component

rurikoaraki:
> A thought for future components since I've seen errors like this before:
> 
> Is there a system where we can just pass '...rest' to the root slot of a component, and expose view props at the top level? Or maybe instruct clients to use customize() if they run into this issue in the future?
> 
> The former suggestion I _think_ lines up with the official React guidance on higher order components? https://reactjs.org/docs/higher-order-components.html#convention-pass-unrelated-props-through-to-the-wrapped-component

This component does expose View props at the top level, the problem is that the ViewProps don't include enableFocusRing except on certain platforms, so TS doesn't actually think they're part of ViewProps even though they'll work perfectly fine.

Also the mergedProps we do on the inner render is how we pass in all extra props to the root slot.

Saadnajmi:
> > A thought for future components since I've seen errors like this before:
> > Is there a system where we can just pass '...rest' to the root slot of a component, and expose view props at the top level? Or maybe instruct clients to use customize() if they run into this issue in the future?
> > The former suggestion I _think_ lines up with the official React guidance on higher order components? https://reactjs.org/docs/higher-order-components.html#convention-pass-unrelated-props-through-to-the-wrapped-component
> 
> This component does expose View props at the top level, the problem is that the ViewProps don't include enableFocusRing except on certain platforms, so TS doesn't actually think they're part of ViewProps even though they'll work perfectly fine. It works for us since we use adapter/IViewProps
> 
> Also the mergedProps we do on the inner render is how we pass in all extra props to the root slot.

I see, thanks! This relates to the typescript platform checking work that the rnx-kit people have been working on. I think the next (latest?) version of Typescript will recognize platform specific types. Thanks for validating we're already doing all the right things =)

PR #1749, opened 2022-06-15
MenuItem text as child instead of `content`

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

To align with FluentUI and Button, passing text of a MenuItem* as a child of the component instead of through the `content` prop

### Verification

Snapshot tests show that outside of removing the extra prop that used to be passed in, the structure of the components is the same, which should mean that text shows up the same as it did before, and accessibility labels are generated as they were before. CI will verify custom accessibility labels.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.tsx
+++ b/packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.tsx
@@ -12,6 +12,7 @@ import {
 } from './MenuItemCheckbox.types';
 import { useMenuItemCheckbox } from './useMenuItemCheckbox';
 import { stylingSettings } from './MenuItemCheckbox.styling';
+import React from 'react';
 
 export const MenuItemCheckbox = compose<MenuItemCheckboxType>({
   displayName: menuItemCheckboxName,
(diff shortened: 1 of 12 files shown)


packages/experimental/Menu/src/MenuItemCheckbox/MenuItemCheckbox.tsx
```diff
@@ -33,17 +34,27 @@ export const menuItemFinalRender = (
...
+    let childText = '';
+    if (accessibilityLabel === undefined) {
+      React.Children.forEach(children, (child) => {
+        if (typeof child === 'string') {
+          childText = child; // We only automatically support the one child string.
+        }
+      });
+    }
+    const label = accessibilityLabel ?? childText;
+
```
ValentinaKozlova, line 53: As this part of code is repeated, maybe it should be moved to a method?

rurikoaraki: Yeah, good call

PR #1761, opened 2022-06-17
Created new notification component

joannaquu:
### Platforms Impacted
- [x ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Created the FURN notification component and updated the notification test to use it

### Verification

Screen Shot 2022-06-14 at 11 24 37 AM

### Pull request checklist

- [ ] Automated Tests
- [x ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover


package-lock.json
```diff
@@ -0,0 +1,9107 @@
+{
+  "name": "@fluentui-react-native/repo",
```
Saadnajmi, line 2: Delete this file please. Did component-generator create it?

package.json
```diff
@@ -82,5 +82,9 @@
     "react-native-windows": "^0.64.3",
     "react-native-macos": "^0.64.6",
     "simple-plist": "2.0.0-rc.0"
+  },
+  "dependencies": {
+    "gulp": "^4.0.2",
+    "gulp-rename": "^2.0.0"
```
Saadnajmi, line 88: why this change?

packages/components/Notification/jest.config.js
```diff
@@ -0,0 +1,2 @@
+const { configureReactNativeJest } = require('@fluentui-react-native/scripts');
+// module.exports = configureReactNativeJest('android');
```
Saadnajmi, line 2: Tempte generated a commented out line?

packages/components/Notification/src/useNotification.ts
```diff
@@ -0,0 +1,12 @@
...
+export const useNotification = (props: NotificationProps): NotificationProps => {
+  const { variant, startText, endText, ...rest } = props;
+
+  return {
+    variant,
+    startText,
+    endText,
+    ...rest,
+  };
+};
```
Saadnajmi, line 12: Lol, a very not useful hook at the moment :D. I guess we might as well keep it because I think it will change.

yarn.lock
```diff
@@ -1407,6 +1407,24 @@
   resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-windows/-/design-tokens-windows-0.13.0.tgz#1d8610634920bd5decaca669cae8183c0713d2ef"
   integrity sha512-o9ir+tpie+Fm9zO+nCbJLBfzVjUhugRWRsPS8n6wBaujC1qyeUYjhoGg6ejXzKc0nVCkIsFmRRRgC0brfrPIzw==
 
+"@fluentui-react-native/framework@0.7.17":
```
Saadnajmi, line 1410: I think the dependencies in your Notication's package.json might be out of date. Can you make sure it's depending on the latest version of all the @fluentui-* packages? Their versions are just elsewhere in our repo, also in package.json files

package-lock.json
joannaquu: Yep

package.json
joannaquu: This was from one of the commands run to use the component generator

Saadnajmi: I would also delete these from this change

package-lock.json
Saadnajmi: @ValentinaKozlova FYI. It seems like component-generator is using NPM instead of yarn, and therefore generating a package-lock.json? It's also adding dependencies to the root package.json. Otherwise, thanks for the awesome FHL project!

packages/components/Notification/jest.config.js
Saadnajmi: Also @ValentinaKozlova , was this being commented out on purpose?

Saadnajmi (approved):

packages/components/Notification/jest.config.js
Saadnajmi: JK, this is necessary. Jest tests fail without it :(

Saadnajmi (approved):

packages/components/Notification/src/NotificationTokens.ts
```diff
@@ -0,0 +1,22 @@
...
+import { NotificationTokens } from './Notification.types';
+
+export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme> = (t: Theme) =>
+  ({
+    backgroundColor: 'skyblue',
+    color: t.colors.bodyText,
+    borderRadius: 12,
+    padding: 16,
+    neutral: {
+      backgroundColor: 'lightgrey',
```
rurikoaraki, line 12: Out of curiosity is there a plan to use design token driven values in the future?

I suppose this is meant for iOS so the issue is that there aren't any tokens to use yet?

packages/components/Notification/src/useNotification.ts
```diff
@@ -0,0 +1,11 @@
+import { NotificationProps } from './Notification.types';
+
+export const useNotification = (props: NotificationProps): NotificationProps => {
+  const { variant, endText, ...rest } = props;
+
+  return {
```
rurikoaraki, line 6: It doesn't seem like this function is needed, since it seems like the object is just being recreated but none of the properties are changing... Is there a plan to add more logic in here in a future change?

joannaquu: This was created by the component generator and will most likely eventually contain more logic but it can be removed for this initial pr.

packages/components/Notification/src/NotificationTokens.ts
Saadnajmi: The plan is to follow the FluentUI Apple control, which is tokenized, refers to some global/alias tokens, but isn't hooked up to the token-pipeline yet. I'm not familiar with the timeline to hook up FluentUI Apple to the token-pipeline, or stand up the iOS design tokens package. Because standing up the iOS design tokens package is out of scope for this project, we'll probably have to copy / hardcode most of the default tokens. I think there are some cases where we can refer to design tokens right now, since it looks like the FluentUI Apple control does. 

If you're curious, the WIP FluentUI Apple control's token definitions are here: https://github.com/microsoft/fluentui-apple/blob/fluent2-tokens/ios/FluentUI/Vnext/Notification/NotificationTokens.swift
You can see line 65 refers to globalTokens.brandColor.tint40, and line 156-160 refers to aliasTokens.shadow.shadow16

rurikoaraki: Sounds good, thanks!

packages/components/Notification/src/useNotification.ts
rurikoaraki: I think it's fine to check in as is if there's future plans for using the function.

rurikoaraki (approved):

packages/components/Notification/src/NotificationTokens.ts
Saadnajmi: Ha, we both commented at the same time :D

PR #1762, opened 2022-06-18
Picker Replacement

nabil-k:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Replaced all instances of the deprecated Picker control with a MenuPicker component for Win32 and MacOS. Made Android and iOS use a Picker community package. The purpose of this was to help unblock the update to RN 0.66

The MenuPicker is essentially a wrapper of the Menu control. The MenuPicker component is located at: `apps\fluenttester\src\FluentTester\TestComponents\Common\MenuPicker.tsx`.

Note: There is a lot of code duplication. For example for `AlignmentPicker.macos.tsx` and `AlignmentPicker.win32.tsx` their only difference is their file name. I'd be happy to hear any suggestions on the best way to handle this.

### Verification
Manually checked if the functionality of the tester app on each platform was preserved in the sections that had the Picker control swapped.

Screenshot 2022-06-17 162408 | Screenshot 2022-06-17 162322

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Avatar/BasicAvatar.macos.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <Text style={textStyles}>Set square Avatar</Text>
+          <Switch value={isSquare} onValueChange={() => setSquare(!isSquare)} />
+        </View>
+        <View style={commonStyles.switch}>
+          <Text style={textStyles}>Set outOfOffice</Text>
+          <Switch value={outOfOffice} onValueChange={() => setOutOfOffice(!outOfOffice)} />
+        </View>
+
+        <MenuPicker prompt="Size" selected={imageSize.toString()} onChange={onSizeChange} collection={avatarSizesForPicker} />
+        <MenuPicker prompt="Active" selected={active} onChange={onActiveChange} collection={avatarActive} />
```
Saadnajmi, line 79: I really don't think we should be duplicating every test like this thrice on each platform. Instead, I think we should have one file for each test (that references <MenuPicker>), and MenuPicker within itself can render a different control based on platform.  

Then, you would be containing the responsibility of knowing what works on what platform into the one control, and client code (aka, all the tests) can use <MenuPicker> without worry and duplication.

Saadnajmi (changes requested):

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/BasicAvatar.macos.tsx
Saadnajmi: It seems you already put in the effort to make sure <MenuPicker> has the same API as StyledPicker (BasicAvatar.tsx, line 29). So combining the two should be easy?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { View, Picker, StyleSheet } from 'react-native';
```
ksiler, line 2: I noticed in ThemePicker.android.tsx we refer to `PickerPropsAndroid` from 'react-native`. We probably want to remove that reference and replace it with something else.

apps/fluent-tester/src/FluentTester/TestComponents/Common/MenuPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+<svg width="12" height="16" viewBox="0 0 11 6" color="#000">
+  <path fill='currentColor' d='M0.646447 0.646447C0.841709 0.451184 1.15829 0.451184 1.35355 0.646447L5.5 4.79289L9.64645 0.646447C9.84171 0.451185 10.1583 0.451185 10.3536 0.646447C10.5488 0.841709 10.5488 1.15829 10.3536 1.35355L5.85355 5.85355C5.65829 6.04882 5.34171 6.04882 5.14645 5.85355L0.646447 1.35355C0.451184 1.15829 0.451184 0.841709 0.646447 0.646447Z' />
+</svg>`;
+
+export const MenuPicker = (props) => {
+  const { prompt, selected, onChange, collection } = props;
+
+  return (
+    <View style={{ flexDirection: 'row', alignItems: 'center', margin: 5 }}>
+      <Text style={{ marginRight: 5 }}>{prompt}</Text>
```
rurikoaraki, line 17: nit: These styles can be const objects like the chevronXml. Doing that would reduce unnecessary renders, since React will treat the styles as a new object if they are passed in like this.

rurikoaraki: Same with the view styling below

apps/fluent-tester/src/FluentTester/TestComponents/Persona/StandardUsage.tsx
```diff
@@ -43,6 +45,34 @@ export const StandardUsage: React.FunctionComponent = () => {
   const [showTertiary, setShowTertiary] = React.useState(true);
   const [showOptional, setShowOptional] = React.useState(true);
   const [imageSize, setImageSize] = React.useState<PersonaSize | undefined>('size72');
+  const renderMenuPicker = Platform.OS == ('win32' as any) || Platform.OS == 'macos';
+
+  const Pickers = () => {
+    return (
+      <Picker
```
rurikoaraki, line 52: Seems a little odd to me that we write two sets of UI for each test that uses pickers. I think it'd be better to have a component which wraps MenuPicker and Picker and renders one or the other based on the platform. That way we're not duplicating picker code for each test with pickers.

rurikoaraki: Also with the way it is written right now, I think the way it works out is that both sets of code run on all platforms and then only one of the two gets used.

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/BasicAvatar.tsx
```diff
@@ -9,12 +9,14 @@ import {
...
 import { PresenceBadgeStatuses, PresenceBadgeStatus } from '@fluentui-react-native/badge';
-import { Switch, View, Text, Picker, ColorValue, Platform } from 'react-native';
+import { Switch, View, Text, ColorValue, Platform } from 'react-native';
+import { Picker } from '@react-native-picker/picker';
 import { satyaPhotoUrl, undefinedText } from './../PersonaCoin/styles';
 import { commonTestStyles as commonStyles } from '../Common/styles';
 import { useTheme } from '@fluentui-react-native/theme-types';
 import TestSvg from '../../test-data/test.svg';
 import { SvgIconProps } from '@fluentui-react-native/icon';
+import { MenuPicker } from '../Common/MenuPicker';
```
Saadnajmi, line 19: (Feedback is for after MenuPicker / StyledPicker are moved to a separate file / component)

When metro does it's bundling, It will see a dependency on "@react-native-picker/picker" for all platforms, and include that in the final bundle. It's not smart enough to know you did a Platform.OS check for win32 on line 63. That won't cause an error, but it will be unnecessary bloat to the final bundle size (why bundle the JS code for Picker if it's never used?).
What you _can_ do is make platform specific files, and only. have the import for the platforms that need it. So you could have your `Picker.tsx` import react-native-picker/picker and render `StyledPicker`, while `Picker.win32.tsx / Picker.macos.tsx` import MenuButton, and render `MenuPicker`

apps/fluent-tester/src/FluentTester/TestComponents/Avatar/BasicAvatar.macos.tsx
nabil-k: I see how that can be a big issue. I've gone back and made it so only the control is responsible for choosing what to render.

apps/fluent-tester/src/FluentTester/theme/ThemePickers.tsx
nabil-k: I didn't notice that, thanks! I've replaced it with the picker props.

apps/fluent-tester/src/FluentTester/TestComponents/Common/MenuPicker.tsx
```diff
@@ -0,0 +1,69 @@
...
+  collection?: any[];
+  style?: any;
+}
+
+export const MenuPicker: React.FunctionComponent<MenuPickerProps> = (props: MenuPickerProps) => {
+  const { prompt, selected, onChange, collection, style } = props;
+
+  const renderDesktopPicker = Platform.OS == ('win32' as any) || Platform.OS == 'macos';
+
+  const DesktopPicker = () => {
```
Saadnajmi, line 26: Nit: It's a bit of a misnomer because it doesn't include windows, but I guess that's alright

Saadnajmi: Could we maybe add a comment to the top of MenuPicker explaining we do it this way because Picker was removed from RN Core so we use the community picker where possible, and our own in house Menu for macOS/win32?

Saadnajmi (approved):

apps/fluent-tester/src/FluentTester/TestComponents/Common/MenuPicker.windows.tsx
```diff
@@ -0,0 +1 @@
+export * from './MenuPicker.desktop';
```
Saadnajmi: I think this will fail because FURN windows doesn't have `Callout`, which `Menu` and therefore `MenuPicker` depend on. You'll probably have to delete this and make windows use the community picker

apps/fluent-tester/src/FluentTester/theme/ThemePickers.android.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { View, PickerPropsAndroid } from 'react-native';
+import { View, PickerProps } from 'react-native';
```
ksiler, line 2: I think `PickerProps` will also be removed in RN 0.66. Is there a `PickerProps` definition you can use from '@react-native-picker/picker'?

apps/fluent-tester/src/FluentTester/theme/ThemePickers.macos.tsx
```diff
@@ -32,6 +34,37 @@ type PartPickerProps = {
...
 };
 
+const ThemeMenuPicker = ({ selected, onChange, collection, style }) => {
+  const chevronXml = `
+  <svg width="12" height="16" viewBox="0 0 11 6" color="#000">
+    <path fill='currentColor' d='M0.646447 0.646447C0.841709 0.451184 1.15829 0.451184 1.35355 0.646447L5.5 4.79289L9.64645 0.646447C9.84171 0.451185 10.1583 0.451185 10.3536 0.646447C10.5488 0.841709 10.5488 1.15829 10.3536 1.35355L5.85355 5.85355C5.65829 6.04882 5.34171 6.04882 5.14645 5.85355L0.646447 1.35355C0.451184 1.15829 0.451184 0.841709 0.646447 0.646447Z' />
+  </svg>`;
+
+  return (
+    <View style={{ flexDirection: 'row', alignItems: 'center', margin: 5, ...style }}>
```
rurikoaraki, line 44: Out of curiosity why can't we use MenuPicker here?

rurikoaraki: Oh it's probably because it's outside of TestComponents. Okay.

Saadnajmi: We could probably do a relative import `import { MenuPicker } from '../TestComponents/Common' if we wanted to. It looks like `ThemePickers.tsx` is already doing that. It doesn't look like we can delete `ThemePickers.macos.tsx` right now, but the relative import should be fine?

Context:
When I worked on bringing FluentTester to mobile, I found that Picker on iOS didn't work well with FluentTester's layout: It was a giant carousel instead of a button with a menu. That was an issue unique to iOS picker (Android, Windows, and macOS didn't have this problem). We also didn't want the light/dark option overridden in the apple theme because PlatformColr stuff. Sooo.. I made ThemePicker platform specific. 

 but we can just change this to be a real

Saadnajmi:
If I had to guess, the Windows PR CI errors are because auto linking Picker on windows failed. Have you tried locally running the windows test app?

rurikoaraki:
The win32 tester changes should not be checked in.

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "CurrentProjectSetting": "No Configurations"
```
rurikoaraki, line 2: What are the .vs files for?

tido64: Pretty sure the whole `.vs` folder can be ignored.

tido64 (changes requested):

apps/fluent-tester/CHANGELOG.md
```diff
@@ -3017,7 +3017,7 @@ Fri, 25 Jun 2021 06:21:11 GMT
 
 ### Minor changes
 
-- 📦 applying package updates ***NO_CI*** (tamasane@gmail.com)
+- 📦 applying package updates **_NO_CI_** (tamasane@gmail.com)
```
tido64, line 3020: Revert?

apps/fluent-tester/src/FluentTester/TestComponents/Common/MenuPicker.windows.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { prompt, onChange, collection } = props;
+  return (
+    <View style={menuPickerStyles.container}>
+      <Text style={menuPickerStyles.prompt}>{prompt}</Text>
+      {collection.map((value, index) => (
+        <Button onPress={() => onChange(value, index)} key={index} title={value} />
+      ))}
+    </View>
+  );
+};
```
Saadnajmi, line 23: Wouldn't this render all the menu items always?

rurikoaraki: I think React has special cases for arrays of children such that it will make a comparison using the key prop

Saadnajmi: I meant that this will just render a View, with a Text component inside of it and a Button for each menu item. There's no "click to open menu" kinda thing happening here unless I am not understanding it properly

tido64 (commented):
\*unblocking*

tido64 (approved):
\*unblocking for realz*

yarn.lock
```diff
@@ -7211,9 +7198,9 @@ globule@^1.0.0:
     minimatch "~3.0.2"
 
 got@^11.0.2:
-  version "11.8.5"
-  resolved "https://registry.yarnpkg.com/got/-/got-11.8.5.tgz#ce77d045136de56e8f024bebb82ea349bc730046"
-  integrity sha512-o0Je4NvQObAuZPHLFoRSkdG2lTgtcynqymzg2Vupdx6PorhaT5MCbIyXG6d4D94kk8ZG57QeosgdiqfJWhEhlQ==
+  version "11.8.3"
```
rurikoaraki, line 7201: Not sure what happened here but this change would bring back a security issue, so we should keep the got version at 11.8.5

yarn.lock
```diff
@@ -2897,9 +2892,9 @@
     form-data "^3.0.0"
 
 "@types/node@*":
-  version "18.0.0"
-  resolved "https://registry.yarnpkg.com/@types/node/-/node-18.0.0.tgz#67c7b724e1bcdd7a8821ce0d5ee184d3b4dd525a"
-  integrity sha512-cHlGmko4gWLVI27cGJntjs/Sj8th9aYwplmZFwmmgYQQvL5NUsgVJG7OddLvNfLqYS31KFN0s3qlaD9qCaxACA==
+  version "17.0.23"
```
rurikoaraki, line 2895: I'm seeing some downgrades of packages in the yarn.lock file. Based on the current changes I don't think the yarn.lock files need to change, can we revert the changes to this file?

rurikoaraki (changes requested):

apps/fluent-tester/src/FluentTester/TestComponents/Common/MenuPicker.windows.tsx
```diff
@@ -0,0 +1,38 @@
...
+  collection?: any[];
+  style?: any;
+}
+
+export const MenuPicker: React.FunctionComponent<MenuPickerProps> = (props: MenuPickerProps) => {
+  const { prompt, onChange, collection } = props;
+  return (
+    <View style={menuPickerStyles.container}>
+      <Text style={menuPickerStyles.prompt}>{prompt}</Text>
+      {collection.map((value, index) => (
```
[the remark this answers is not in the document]
rurikoaraki: Also curious how it affects the BasicAvatar test page

FalseLobster:
I know this addresses a short term need for the FURN tester, but I'm curious if we talked about productizing this as a FURN picker at all? We do still need that control in FURN itself, right?

nabil-k:
> I know this addresses a short term need for the FURN tester, but I'm curious if we talked about productizing this as a FURN picker at all? We do still need that control in FURN itself, right?

I personally didn't have a conversation about that, but I think it would be nice if we had a FURN picker.

PR #1781, opened 2022-06-23
Remove menu popover from being tab stop

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Bug was that the menu popover was a tab stop when tabbing when the menu was opened via mouse. This was because it was listed as tabble when the Menu is invoked via mouse.

Copied this fix from ContextualMenu. Focusable is set to false on the container as soon as focus is shifted away from it, removing it from the tab loop.

### Verification

Tested via tester.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
+++ b/packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
@@ -48,6 +48,11 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
     setPopoverHoverOutTimer(timer);
   }, [openOnHover, setOpen, setPopoverHoverOutTimer]);
 
+  const [canFocusOnContainer, setCanFocusOnContainer] = React.useState<boolean>(shouldFocusOnContainer);
+  const onBlur = React.useCallback(() => {
+    setCanFocusOnContainer(false);
+  }, [setCanFocusOnContainer]);
+
   return {
     props: {
       accessibilityRole,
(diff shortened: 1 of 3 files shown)


packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -48,6 +48,11 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
     setPopoverHoverOutTimer(timer);
   }, [openOnHover, setOpen, setPopoverHoverOutTimer]);
 
+  const [canFocusOnContainer, setCanFocusOnContainer] = React.useState<boolean>(shouldFocusOnContainer);
+  const onBlur = React.useCallback(() => {
+    setCanFocusOnContainer(false);
```
Saadnajmi, line 53: nit, can we name this Popover instead of container. I was very much not a fan of the "container" name from the get go

rurikoaraki: Yeah, just renamed.

Saadnajmi (approved):

packages/experimental/Menu/src/MenuPopover/useMenuPopover.ts
Saadnajmi: Approved without testing on macOS because... well I can't keep doing that. I think it should be OK though, I'm not even sure we expose that prop.

PR #1785, opened 2022-06-24
Fix Menu focus visual bug

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This bug didn't show until I added hover visuals for focused components. Although components are focused when hovered state is changed because of useEffect, the component doesn't show the UI until the next rerender because the focus state is triggered using useEffect. This means that the UI shows up when a component is, for example, hovered out of.

To fix this, change running focus() to useLayoutEffect so that it runs sooner in the render cycle. The visual now updates as part of hovering the component.

### Verification

Tested via tester.

Before:
focus_visual_bug

After:
focus_visual_fix

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/experimental/Menu/src/MenuItem/useMenuItem.ts
@@ -105,7 +105,7 @@ function getAccessibilityStateWorker(disabled: boolean, accessibilityState?: Acc
 }
 
 export const useHoverFocusEffect = (hovered: boolean, componentRef: React.MutableRefObject<any>) => {
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     if (hovered) {
       componentRef?.current?.focus();
     } else {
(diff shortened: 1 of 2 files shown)


packages/experimental/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -105,7 +105,7 @@ function getAccessibilityStateWorker(disabled: boolean, accessibilityState?: Acc
 }
 
 export const useHoverFocusEffect = (hovered: boolean, componentRef: React.MutableRefObject<any>) => {
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
```
Saadnajmi, line 108: oof on the bug, but glad you caught it!

Saadnajmi (approved):

packages/experimental/Menu/src/MenuItem/useMenuItem.ts
rurikoaraki: lol thank EJ for making me change the colors

Saadnajmi: Hover state on macOS matches rest (I only color the focus state differently), so I would have probably missed it too

PR #1794, opened 2022-06-28
Fix various arrowing behaviors for closing menus

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change fixes three bugs around arrowing left (or right in RTL) to close a submenu:

1. Fixes bug where arrowing on disabled items wouldn't close the submenu (move disable check to just checking executing onClick)
2. Fixes bug where arrowing on checkbox/radio items wouldn't close the submenu (add arrowing behavior to checkbox/radio items
3. Fixes bug where arrowing on submenu triggers in submenus wouldn't close the submenu (handle closing the submenu at one level higher so that the event is outside of the submenu the trigger is responsible for. Comment added to explain case).

Web's solution (which is to handle arrow open on the MenuTrigger and arrow close on the MenuPopover) is more elegant here but 1. takes advantage of being able to query the HTML attribute of the source element of the event to look for disabled state of children (which we don't have access to on native) and 2. relies on arrowing events going up the UI tree, which we don't get because arrow events get handled natively if not handled first by shifting focus in a list. So we get this instead :)

### Verification

Tested with tester.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Menu/MenuTest.tsx
@@ -119,8 +119,8 @@ const Submenu: React.FunctionComponent = () => {
       </MenuTrigger>
       <MenuPopover>
         <MenuList>
+          <MenuItemCheckbox name={'a'}>A nested MenuItemCheckbox</MenuItemCheckbox>
           <MenuItem>A nested MenuItem</MenuItem>
-          <MenuItem>A second nested MenuItem</MenuItem>
           <Submenu />
         </MenuList>
       </MenuPopover>
--- a/change/@fluentui-react-native-menu-1ecf27ef-0190-4927-b0a9-5e6de8ec7806.json
+++ b/change/@fluentui-react-native-menu-1ecf27ef-0190-4927-b0a9-5e6de8ec7806.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix arrow close not working",
+  "packageName": "@fluentui-react-native/menu",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@fluentui-react-native-tester-96e03ad1-b18d-4da5-a9cf-7c53d38490dc.json
+++ b/change/@fluentui-react-native-tester-96e03ad1-b18d-4da5-a9cf-7c53d38490dc.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Modify a test",
+  "packageName": "@fluentui-react-native/tester",
+  "email": "ruaraki@microsoft.com",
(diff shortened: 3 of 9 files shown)


Saadnajmi:
Could you link the code where web queries source?

Saadnajmi:
> Arrow events get handled natively, i.e shift focus before we can get to them.

That should not be the case if we're using 'useKeyDownProps'. That hook passes 'keyDownEvents' to View, who's purpose is to suppress the native handling of the event on windows/ win32. That sounds like a bug if you're seeing that, perhaps in the hook?

rurikoaraki:
> > Arrow events get handled natively, i.e shift focus before we can get to them.
> 
> That should not be the case if we're using 'useKeyDownProps'. That hook passes 'keyDownEvents' to View, who's purpose is to suppress the native handling of the event on windows/ win32. That sounds like a bug if you're seeing that, perhaps in the hook?

I didn't look too deep into it, but I was doing the useKeyDownProps hookup at the MenuPopover level to mimic web, and the event didn't seem to go up the tree. So maybe the issue is that the event didn't propagate up.

Saadnajmi:
> > > Arrow events get handled natively, i.e shift focus before we can get to them.
> > 
> > That should not be the case if we're using 'useKeyDownProps'. That hook passes 'keyDownEvents' to View, who's purpose is to suppress the native handling of the event on windows/ win32. That sounds like a bug if you're seeing that, perhaps in the hook?
> 
> I didn't look too deep into it, but I was doing the useKeyDownProps hookup at the MenuPopover level to mimic web, and the event didn't seem to go up the tree. So maybe the issue is that the event didn't propagate up.

It's probably because MenuPopover is a Callout is a native window, and native events stop when they reach the window. I thought JS events might continue past but I guess i guess I'm wrong.

rurikoaraki:
> > > > Arrow events get handled natively, i.e shift focus before we can get to them.
> > > 
> > > 
> > > That should not be the case if we're using 'useKeyDownProps'. That hook passes 'keyDownEvents' to View, who's purpose is to suppress the native handling of the event on windows/ win32. That sounds like a bug if you're seeing that, perhaps in the hook?
> > 
> > 
> > I didn't look too deep into it, but I was doing the useKeyDownProps hookup at the MenuPopover level to mimic web, and the event didn't seem to go up the tree. So maybe the issue is that the event didn't propagate up.
> 
> It's probably because MenuPopover is a Callout is a native window, and native events stop when they reach the window. I thought JS events might continue past but I guess i guess I'm wrong.

I was having it handled on the inner view of the popover, so that shouldn't have been a problem, I think?

```
--- MenuPopover
<Callout>
    <View onKeyDown>
---
        <MenuList (aka View)>
            <MenuItem>
```

Saadnajmi (approved):

PR #1808, opened 2022-06-30
[Testing] Update how appium windows driver gets installed

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.ado/templates/e2e-testing-uwp.yml
+++ b/.ado/templates/e2e-testing-uwp.yml
@@ -4,8 +4,15 @@ steps:
     workingDirectory: apps/windows
     displayName: 'yarn bundle'
 
+  # The shrinkwrap file forces installation of all the dev dependencies of the package as well, which are unnecessary.
+  # Delete the file so we only install the necessary files
+  - powershell: |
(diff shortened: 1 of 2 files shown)


Saadnajmi:
Is this still necessary? Asking since we're any to have a brown bag about e2e testing

rurikoaraki:
> Is this still necessary? Asking since we're any to have a brown bag about e2e testing

I mean it'd be nice since it means that the windows driver would use the already downloaded copy of appium-windows-driver instead of downloading a new copy with all its dev dependencies (which we don't need). It's not something that's changing what appium does, just how the driver gets pulled/installed.

Saadnajmi:
Thanks for the clarification! I wasn't sure the original intention, and that makes sense :)

rurikoaraki:
> Thanks for the clarification! I wasn't sure the original intention, and that makes sense :). It's probably also valid for macOS then since we are most likely also installing appium twice

Ah yeah I just noticed that I left everything blank. Probably because I was using this PR to test CI.

PR #1810, opened 2022-06-30
Incorporate button/pressable into notification

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update notification to use button for action and pressable for notification body

### Verification

| Colors | Simulator Screen Shot - iPhone 13 Pro - 2022-06-30 at 09 35 47 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


change/@fluentui-react-native-avatar-bde7ae22-cfb0-4643-8057-f7a10b48ccb1.json
```diff
@@ -0,0 +1,7 @@
+{
```
Saadnajmi, line 1: I think your fork of FURN is out of sync with upstream. When that happens, beachball makes a lot of unnecessary change messages. Could you double check that it's in sync and retry?

packages/components/Notification/src/Notification.styling.ts
```diff
@@ -3,7 +3,15 @@ import { Theme, UseStylingOptions, buildProps } from '@fluentui-react-native/fra
 import { borderStyles, layoutStyles } from '@fluentui-react-native/tokens';
 import { defaultNotificationTokens } from './NotificationTokens';
 
-export const notificationStates: (keyof NotificationTokens)[] = ['primary', 'neutral', 'danger', 'warning'];
+export const notificationStates: (keyof NotificationTokens)[] = [
```
Saadnajmi, line 6: I would name this "variants". States will include things like "pressed"

packages/components/Notification/src/Notification.styling.ts
```diff
@@ -28,23 +37,29 @@ export const stylingSettings: UseStylingOptions<NotificationProps, NotificationS
         return {
           style: {
             color: tokens.color,
-            fontSize: 16,
+            fontSize: 15,
+            fontWeight: '600',
+            lineHeight: 20,
+            letterSpacing: -0.24,
```
Saadnajmi, line 43: Why -0.24?

packages/components/Notification/src/Notification.tsx
```diff
@@ -1,10 +1,12 @@
 /** @jsx withSlots */
-import { View } from 'react-native';
 import { notification, NotificationType, NotificationProps } from './Notification.types';
+import { Pressable } from 'react-native';
```
Saadnajmi, line 3: For now, import `Pressable` from '@fluentui-react-native/pressable`. We have a copy of Pressable we use, that we should eventually switch off of.

packages/components/Notification/src/Notification.styling.ts
joannaquu: From the figma

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
```diff
@@ -1,51 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
-
-exports[`Notification component tests Notification default 1`] = `
```
Saadnajmi, line 3: Why is this deleted?

Saadnajmi: You can update the test with 'yarn test -u' at the end. But that makes sense.

joannaquu: Ooh ok sounds good!

harrieshin:
action button colors are incorrect in the after screenshot?

joannaquu:
> action button colors are incorrect in the after screenshot?

Yes, I've been having trouble changing the button text style but Saad recommended checking this in without the change since the best person to contact for more info on this is Ruriko and she's oof

packages/components/Notification/src/useNotification.ts
```diff
@@ -1,11 +1,11 @@
...
-  const { variant, endText, ...rest } = props;
+  const { variant, action, ...rest } = props;
 
   return {
     variant,
-    endText,
+    action,
     ...rest,
   };
 };
```
Saadnajmi, line 11: As mentioned offline, I think we can delete this file till we actually need it.

Saadnajmi (approved):

packages/components/Notification/src/NotificationTokens.ts
```diff
@@ -6,6 +6,8 @@ export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme>
   ({
     backgroundColor: t.colors.background,
     color: t.colors.brandForeground1,
+    borderColor: '#FFFFFF',
```
rurikoaraki, line 9: Hardcoding white here seems like it would show poorly on dark modes, does it look ok?

joannaquu: Looks ok so far

joannaquu: Edit: I think I misread your comment. For toasts, the border is actually supposed to be clear but white was the closest I could get.

rurikoaraki: You can set the color to 'transparent' if it's supposed to be clear.

rurikoaraki: (probably transparentStroke is better since then on HC it'll still draw a border)

joannaquu: hmm transparentStroke adds a black border even without HC?

rurikoaraki: That's odd. @chiuam do you know what's up here?

joannaquu: It seems that black is the default color and transparentStroke either doesn't exist or is not working so it defaults to black

rurikoaraki: Ah yeah it might not be filled out on iOS. Transparent seems ok then, though you might need to change the color when using a high contrast mode if those exist on iOS.

Saadnajmi: It looks like on iOS, we never defined the token transparentStroke, and in appleColors.ios.ts, I just did this:
>     primaryButtonBorder: 'transparent',

So it's probably OK to just set 'transparent' where needed. We have iOS FURN token support planned out for Fy23 and can update this when that happens.

PR #1812, opened 2022-06-30
Shadow: get tokens from theme and use stagedcomponent

lyzhan7:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- Update + move shadowStyleFromTheme method
  - create a copy of shadowStyle.ts in Shadow
  - make shadowStyleFromTheme return key and ambient
  - make helper methods in shadowStyleFromTheme
  - replace deprecated substr method with substring method
  - round shadowOpacity to two decimal places
- Remove hardcoded shadow values, use shadow values from theme 
- Update shadowDepth type to include brand shadows
- Use stagedComponent and memoize

### Verification
(Known issue: some shadows don't show up such a shadow 64 in dark mode, we have a bug filed)

| Light mode | Dark mode |
| --- | --- |
| Screen Shot 2022-07-08 at 10 11 00 AM | Screen Shot 2022-07-08 at 10 15 47 AM|

Testing
- Confirmed that the values we were getting from the tokens matched the hardcoded values I had grabbed from the Figma


packages/components/Button/src/FAB/FABColorTokens.ts
```diff
@@ -1,10 +1,12 @@
 import { Theme } from '@fluentui-react-native/framework';
 import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { ButtonCoreTokens } from '../Button.types';
-import { shadowStyleFromTheme } from '../shadowStyle';
+import { shadowStyleFromTheme } from '@fluentui-react-native/experimental-shadow';
 
 export const defaultFABColorTokens: TokenSettings<ButtonCoreTokens, Theme> = (t: Theme): ButtonCoreTokens => ({
-  ...shadowStyleFromTheme(t, 'shadow8'),
+  // // Grab the second shadow from the global token, as we only support displaying one shadow
```
lyzhan7, line 7: Copied this comment over from shadowStyle.ts

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -1,31 +1,25 @@
...
-    '16': { shadowOffset: { width: 0, height: 8 }, shadowOpacity: 0.14, shadowRadius: 16 },
-    '28': { shadowOffset: { width: 0, height: 14 }, shadowOpacity: 0.24, shadowRadius: 28 },
-    '64': { shadowOffset: { width: 0, height: 32 }, shadowOpacity: 0.24, shadowRadius: 64 },
+const useShadowTokens = buildUseTokens<ShadowTokens>(() => ({}), shadowName);
+
+export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps) => {
+  const { depth } = props;
+  const theme = useFluentTheme();
+
+  const shadowStyle = shadowStyleFromTheme(theme, 'shadow' + depth.toString());
```
lyzhan7, line 13: Not sure about the 'shadow' + depth.toString() part, thinking I can either make ShadowDepth to be "shadow2, shadow4, ..." instead of "2, 4, ..."

packages/experimental/Shadow/src/shadowStyle.ts
```diff
@@ -0,0 +1,46 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { ColorValue } from 'react-native';
+
+export const shadowStyleFromTheme = (t: Theme, shadowToken: string) => {
+  const keyShadow = t.shadows[shadowToken].key;
+  const ambientShadow = t.shadows[shadowToken].ambient;
+
+  return {
+    key: {
+      // iOS Shadow props
```
lyzhan7, line 10: Thinking I should remove this comment (and the one on line 22), and then was thinking maybe we could move the android shadow props somewhere else? Maybe a different method?

packages/components/Button/package.json
```diff
@@ -30,6 +30,7 @@
     "@fluentui-react-native/adapters": ">=0.8.5 <1.0.0",
     "@fluentui-react-native/experimental-activity-indicator": "^0.5.29",
     "@fluentui-react-native/experimental-text": ">=0.9.0 <1.0.0",
+    "@fluentui-react-native/experimental-shadow": "0.2.0",
```
Saadnajmi, line 33: @rurikoaraki We add this dependency here because the `shadowStyleFromTheme` function used to exist in the Button package, but now we're moving it to the Shadow package. Do you see any issue with adding a new dependency here now that we're V1? If so, we can just duplicate the file, and then update Button when the Shadow work is complete.

packages/experimental/Shadow/src/shadowStyle.ts
Saadnajmi: Having platform specific files I think would be a good idea. The iOS props (afaik) are applicable to iOS/macOS/win32, so that could stay `shadowStyle.ts`, while Android has a `shadowStyle.android.ts` that simply returns "elevation".

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -1,31 +1,25 @@
...
+
+  return (extra: ShadowProps, children: React.ReactNode) => {
+    return (
+      <View style={shadowStyle.ambient}>
+        <View style={shadowStyle.key}>{children}</View>
+      </View>
```
Saadnajmi, line 19: I suggest we pass the rest of props down to the inner view. That way, the client can use ShadowComponent like any other View, with ViewProps, but also gets a shadow effect. 

The convention of passing unrelated props is listed / documented here, but I find it not very readable: https://reactjs.org/docs/higher-order-components.html#convention-pass-unrelated-props-through-to-the-wrapped-component

```
const mergedProps = mergeProps(
  {style: {shadowStyle.key} },
  ...rest
 );
 
 
 return (
  <View style={shadowStyle.ambient}>
         <View {...mergedProps}>{children}</View>
  </View>
 )
 ```

File: packages/experimental/Shadow/src/Shadow.tsx [hunk 6]
```diff
@@ -1,31 +1,25 @@
...
-    '64': { shadowOffset: { width: 0, height: 32 }, shadowOpacity: 0.24, shadowRadius: 64 },
+const useShadowTokens = buildUseTokens<ShadowTokens>(() => ({}), shadowName);
+
+export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps) => {
+  const { depth } = props;
+  const theme = useFluentTheme();
+
+  const shadowStyle = shadowStyleFromTheme(theme, 'shadow' + depth.toString());
+
+  return (extra: ShadowProps, children: React.ReactNode) => {
```
lyzhan7, line 15: yarn buildci is going to fail because I don't use 'extra' anywhere, but if I delete that parameter I get red dialog boxes. Honestly not sure what this is supposed to be for (copied from other examples of compressible) - I could just print it out or something to get the build to pass, but not sure if there's a better alternative (ex. is there a way to remove it entirely without getting errors, and add it back in later when I get to this part?)

lyzhan7: When I delete it I get this error:

Screen Shot 2022-07-01 at 3 09 22 PM

Putting an underscore works! Although I didn't need it with the merge props change. Will play around with it.

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: Updated to take in the type ShadowDepth, but ended up also making ShadowDepth "shadow2, shadow4, etc" because I realized that there are also different shadows for brands (ex. shadow2brand). 

@rurikoaraki I noticed that [in the PR you made adding shadow tokens](https://github.com/rurikoaraki/fluentui-design-tokens/blob/e0fb4335a666ed62c08812bb87d1bc33826b2bad/src/shadow.json), the object is divided into 'global' and 'brand' shadows, and then within each group there's 2, 4, 6, 8, etc.  but in FURN at some point the depths and global vs. brand get flattened into BaseShadowAliasTokens (i.e. shadow2, shadow2brand, etc). Just wanted to check with you that having the ShadowDepth type match BaseShadowAliasTokens makes the most sense?

packages/components/Button/src/FAB/FABColorTokens.ts
```diff
@@ -1,10 +1,12 @@
...
 import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { ButtonCoreTokens } from '../Button.types';
-import { shadowStyleFromTheme } from '../shadowStyle';
+import { shadowStyleFromTheme } from '@fluentui-react-native/experimental-shadow';
 
 export const defaultFABColorTokens: TokenSettings<ButtonCoreTokens, Theme> = (t: Theme): ButtonCoreTokens => ({
-  ...shadowStyleFromTheme(t, 'shadow8'),
+  // Grab the second shadow from the global token, as we only support displaying one shadow
+  ...shadowStyleFromTheme(t, '8').key,
+
```
Saadnajmi, line 9: As mentioned offline, I think it would be best to revert this change, duplicate the file, and also delete the change file for experimental-button

packages/components/Button/package.json
rurikoaraki: Adding dependencies is fine if it's necessary for new functionality.

packages/experimental/Shadow/src/Shadow.types.ts
```diff
@@ -1,7 +1,23 @@
 import { ViewProps } from 'react-native';
 
-export type ShadowDepth = '2' | '4' | '8' | '16' | '28' | '64';
+export const shadowName = 'Shadow';
+
+export type ShadowDepth =
```
rurikoaraki, line 5: You could also have two props. one for `ShadowDepth` and one for `isBrand`

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -1,31 +1,27 @@
...
-    '16': { shadowOffset: { width: 0, height: 8 }, shadowOpacity: 0.14, shadowRadius: 16 },
-    '28': { shadowOffset: { width: 0, height: 14 }, shadowOpacity: 0.24, shadowRadius: 28 },
-    '64': { shadowOffset: { width: 0, height: 32 }, shadowOpacity: 0.24, shadowRadius: 64 },
+const useShadowTokens = buildUseTokens<ShadowTokens>(() => ({}), shadowName);
+
+export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps) => {
+  const { depth } = props;
+  const theme = useFluentTheme();
+
+  const shadowStyle = shadowStyleFromTheme(theme, depth);
```
rurikoaraki, line 13: Hm we probably want some sort of memoization here. Otherwise the style object gets recreated on every rerender and that would cause excessive rerenderings

rurikoaraki (changes requested):

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
```diff
@@ -8,7 +8,7 @@ describe('Shadow component tests', () => {
   it('Shadow default', () => {
     const tree = renderer
       .create(
-        <Shadow>
+        <Shadow depth="shadow2">
```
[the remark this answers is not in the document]
lyzhan7: Will do (possibly in another PR if this one gets too big) - currently working on fixing the tests ([posted in teams here](https://teams.microsoft.com/l/message/19:fb48dcfcd2bd4db8bfd439a15c556028@thread.skype/1657043439259?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=d0ed8f47-0461-46ca-8ac3-e9f613623741&parentMessageId=1657043439259&teamName=Office%20CXE&channelName=React%20Native%20Controls%20Crew&createdTime=1657043439259))

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: I'm assuming I can add the memoization in Shadow.tsx, with the cache that I see in other components that implement compressible? Let me know if you were thinking of something different

rurikoaraki: Yeah, that should be able to work. You can use something similar to what other tokens do in other compressible components or you can use memoize

packages/components/Button/src/FAB/FABColorTokens.ts
lyzhan7: Just updated

packages/components/Button/package.json
lyzhan7: Ended up just duplicating the file

packages/experimental/Shadow/src/shadowStyle.ts
lyzhan7: Discussed offline but I tried this and had some issues with running the tests by platform, we also decided Android is out of scope for this component

packages/experimental/Shadow/src/shadowStyle.ts
```diff
@@ -0,0 +1,41 @@
...
+  return rgbaColor.toString().substring(0, 7);
+};
+
+const shadowOpacityFromRGBAColor = (rgbaColor: ColorValue) => {
+  const opacityAsHex = '0x' + rgbaColor.toString().substring(7);
+  const opacityAsDecimal = Number(opacityAsHex) / 255.0;
+
+  // Round to two decimal places
+  return Math.round(opacityAsDecimal * 100) / 100;
+};
```
Saadnajmi, line 41: I like the more exact names here :)

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -1,22 +1,43 @@
...
 
 const useShadowTokens = buildUseTokens<ShadowTokens>(() => ({}), shadowName);
 
-export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps) => {
+export const Shadow = compressible<ShadowProps, ShadowTokens>((props: ShadowProps, useTokens: UseTokens<ShadowTokens>) => {
   const { depth } = props;
   const theme = useFluentTheme();
 
+  const [tokens, cache] = useTokens(theme);
   const shadowStyle = shadowStyleFromTheme(theme, depth);
```
Saadnajmi, line 14: I would move this line to inside the cache() call.

```js
const [keyShadowTokenStyle] = cache(
  () => {
    const shadowStyle = shadowStyleFromTheme(theme, depth);
    return {
      shadowColor: shadowStyle.key.shadowColor,
      shadowOpacity: shadowStyle.key.shadowOpacity,
      shadowRadius: shadowStyle.key.shadowRadius,
      shadowOffset: shadowStyle.key.shadowOffset,
    }
  },
  ['theme', 'depth'],
);
```

lyzhan7: I see, if it's outside the cache() call is it defeating the purpose of the cache?

Saadnajmi: You also _can_ combine the two cache calls if you want, that's more of a stylistic choice. I think they both perform similar.

```js
const [shadowTokenStyleSet] = cache(
  () => (shadowStyleFromTheme(theme, depth), 
  ['theme', 'depth'],
);

....
<View style={shadowTokenStyleSet.ambient}>
  <View style={shadowTokenStyleSet.key}></View>
</View>
```
```

Comment (Saadnajmi) [reply packages/experimental/Shadow/src/Shadow.tsx]: > I see, if it's outside the cache() call is it defeating the purpose of the cache?

A bit. You're doing unnecessary work, but it won't cause an unnecessary render (since you're using the cached value in the end anyway)

File: packages/experimental/Shadow/src/Shadow.tsx [hunk 13]
```diff
@@ -1,31 +1,47 @@
...
+
+  const [keyShadowTokenStyle] = cache(() => {
+    const keyShadowStyle = shadowStyleFromTheme(theme, depth).key;
+    return {
+      shadowColor: keyShadowStyle.shadowColor,
+      shadowOpacity: keyShadowStyle.shadowOpacity,
+      shadowRadius: keyShadowStyle.shadowRadius,
+      shadowOffset: keyShadowStyle.shadowOffset,
+    };
+  }, ['shadowColor', 'shadowOpacity', 'shadowRadius', 'shadowOffset']);
```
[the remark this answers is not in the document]
lyzhan7: Yes sorry I should have commented earlier - talked though this with Saad yesterday and have a lot of local changes (including the fix for this), just working through right now separating them out into commits. I realized there were a lot of issues with the cache (ex. in the screenshot I posted the depth ramp is actually missing one of the shadows). We ended up thinking of using StagedComponent (with memoize) instead of compressible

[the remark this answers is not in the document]
rurikoaraki: okay sounds good. let me know when i should review, i'll hold off til then.

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: I see okay, thanks Saad!

packages/experimental/Shadow/src/Shadow.tsx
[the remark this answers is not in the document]
lyzhan7: Thanks Ruriko, I pushed some fixes to the cache but there are still known issues and I don't think this PR is ready to review until those are fixed. I've decided to pivot to prototype stagedComponent so will mark this PR as a draft for now.

packages/experimental/Shadow/src/Shadow.types.ts
lyzhan7: Going to leave it for now, but thank you for the suggestion!

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowDepthTestSection.tsx
```diff
@@ -1,53 +1,71 @@
...
+          themedStyles.effectBox,
+          themedStyles.vmargin,
+          themedStyles.padding,
+          { backgroundColor: props.backgroundColor },
+        ]}
+        variant="bodySemibold"
+      >
+        {props.depth}
       </Text>
     </Shadow>
```
Saadnajmi, line 48: Would be nice if we can print the key and ambient shadow tokens as well as depth on line 46, but not a blocking requirement. Just a thought :)

packages/experimental/Shadow/src/__tests__/__snapshots__/Shadow.test.tsx.snap
```diff
@@ -1,10 +1,408 @@
...
+>
+  <View
+    style={
+      Object {
+        "shadowColor": "#000000",
+        "shadowOffset": Object {
+          "height": 32,
+          "width": 0,
+        },
+        "shadowOpacity": 0.25,
```
Saadnajmi, line 195: Nit: This is supposed to be 24% but is rounded to 25%. In practice I don't think the 1% difference matters but wanted to call it out.

Saadnajmi (approved):
Looks good apart from my comment about memoize function naming!

packages/experimental/Shadow/src/__tests__/__snapshots__/Shadow.test.tsx.snap
lyzhan7: Ah thanks for catching that, I didn't notice. Yeah I'm not sure what I can do about this, I doubt there's a perfect way to extract the right value. Maybe I can note it down as a known issue in the shadowStyleFromTheme method

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: Just added, will update screenshots

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -1,31 +1,29 @@
...
+    );
   };
+});
+
+Shadow.displayName = shadowName;
+
+const getShadowTokenStyleSet = memoize(getShadowTokenStyleSetWorker);
+function getShadowTokenStyleSetWorker(theme: Theme, depth: ShadowDepth) {
+  return shadowStyleFromTheme(theme, depth);
+}
```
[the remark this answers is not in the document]
lyzhan7: I'm now calling shadowStyleFromTheme from the Test page, so thinking now it actually makes sense to keep it as two functions- let me know if you still think combining them in some way makes sense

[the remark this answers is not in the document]
rurikoaraki: Yeah we should generally be using memoized functions for things that would be expensive to calculate on each render. I know we're not the best at preventing wasted renders in our Tester app, but it's probably better to call the memoized version in the test page too.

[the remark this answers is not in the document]
lyzhan7: Thanks for the suggestion Saad & thanks Ruriko for the explanation, should be fixed now!

packages/experimental/Shadow/package.json
```diff
@@ -25,6 +25,9 @@
     "url": "https://github.com/microsoft/fluentui-react-native.git",
     "directory": "packages/experimental/Shadow"
   },
+  "dependencies": {
+    "@fluentui-react-native/framework": "0.7.30"
```
rurikoaraki, line 29: Needs to be updated to 0.7.31 so that we pull from the workspace and not a yarn.lock entry

lyzhan7: Ooh thank you that must be why the checks are failing, thought it was related to the package.json file but didn't know there was a new version

rurikoaraki (approved):

PR #1837, opened 2022-07-11
Add optional icon to notification

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added optional icon to notification component.

### Verification

| before | 9 - Icon |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
@@ -7,6 +7,7 @@ import PlayButton from './assets/play_button.svg';
 const svgProps: SvgIconProps = {
   src: PlayButton,
 };
+const iconProps = { svgSource: svgProps };
 
 const PrimaryTest: React.FunctionComponent = () => {
   return (
@@ -26,7 +27,7 @@ const PrimaryTestWithTitleAndIcon: React.FunctionComponent = () => {
   return (
     <Notification
       variant={'primary'}
-      icon={{ svgSource: svgProps }}
+      icon={iconProps}
       title="Kat's iPhone X"
       action="X"
       onPress={() => {
--- a/apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
+++ b/apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
@@ -1,6 +1,12 @@
 import * as React from 'react';
 import { Notification } from '@fluentui-react-native/notification';
 import { Test, TestSection, PlatformStatus } from '../Test';
+import { SvgIconProps } from '@fluentui-react-native/icon';
+import PlayButton from './assets/play_button.svg';
+
+const svgProps: SvgIconProps = {
+  src: PlayButton,
+};
 
 const PrimaryTest: React.FunctionComponent = () => {
   return (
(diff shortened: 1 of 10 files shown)


apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -16,10 +22,11 @@ const PrimaryTest: React.FunctionComponent = () => {
   );
 };
 
-const PrimaryTestWithTitle: React.FunctionComponent = () => {
+const PrimaryTestWithTitleAndIcon: React.FunctionComponent = () => {
   return (
     <Notification
       variant={'primary'}
+      icon={{ svgSource: svgProps }}
```
Saadnajmi, line 29: This creates a new object `{ svgSource: svgProps }` every time this is run, causing unnecessary renders. I'd prefer this is created on a separate line within the function then referenced. something like

```
const PrimaryTestWithTitleAndIcon: React.FunctionComponent = () => {
  const iconProps = { svgSource: svgProps };
  return (
    <Notification
       icon={{ svgSource: svgProps }}
       ...
  )

```

packages/components/Notification/src/Notification.styling.ts
```diff
@@ -33,6 +33,12 @@ export const stylingSettings: UseStylingOptions<NotificationProps, NotificationS
       }),
       ['backgroundColor', ...borderStyles.keys, ...layoutStyles.keys],
     ),
+    icon: buildProps(() => ({
+      style: {
+        alignSelf: 'center',
+        marginRight: 16,
```
Saadnajmi, line 39: Note for future: I wonder if we want to specify layout tokens for each slots. 
iconMargin, titleMargin, etc. I would say that depends on whether FluentUI Apple specifies those.

packages/components/Notification/src/Notification.tsx
```diff
@@ -46,11 +48,12 @@ export const Notification = compose<NotificationType>({
     }, ['isBar']);
 
     return (final: NotificationProps, ...children: React.ReactNode[]) => {
-      const { variant, title, action, ...rest } = mergeProps(userProps, final);
+      const { variant, icon, title, action, ...rest } = mergeProps(userProps, final);
       const mergedProps = mergeProps<PressableProps>(rest, rootStyle);
 
       return (
         <Slots.root {...mergedProps}>
+          {icon && <Slots.icon {...icon} width={24} height={24}></Slots.icon>}
```
Saadnajmi, line 56: Move the hardcoded size to icon's buildProps?

Saadnajmi: Also, you didn't need to call `createIconProps` ?

packages/components/Notification/src/Notification.styling.ts
joannaquu: FUA doesn't but good to keep in mind

packages/components/Notification/src/Notification.tsx
joannaquu: Nope, works without it which makes me think this code is buggy

joannaquu: nvm, figured it out

Saadnajmi (approved):

packages/components/Notification/src/Notification.styling.ts
```diff
@@ -33,14 +33,18 @@ export const stylingSettings: UseStylingOptions<NotificationProps, NotificationS
...
-        marginRight: 16,
-      },
-      height: 24,
-      width: 24,
-    })),
+    icon: buildProps(
+      (tokens: NotificationTokens) => ({
+        style: {
+          alignSelf: 'center',
+          marginRight: 16,
```
Saadnajmi, line 40: (As I must ask whenever I see "marginRight" instead of "marginStart"...)
Should this flip in RTL? If so, we probably want to use marginStart?

Saadnajmi (approved):

PR #1838, opened 2022-07-11
Clear timers when unmounting components

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Forgot to clear the timeouts when unmounting components. This causes warnings to come up from React about being unable to change the state of unmounted components.
Clear timeouts when unmounting the component which creates them.

### Verification

Tried reproducing using client provided repro steps and unable to repro.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
+++ b/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
@@ -55,9 +55,7 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
 
   React.useEffect(() => {
     return function cleanup() {
-      clearTimeout(triggerHoverOutTimer);
       clearTimeout(popoverHoverOutTimer);
-      clearTimeout(parentPopoverHoverOutTimer);
     };
   });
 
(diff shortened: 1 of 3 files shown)


packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -53,6 +53,12 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
     setCanFocusOnPopover(false);
   }, [setCanFocusOnPopover]);
 
+  React.useEffect(() => {
+    return function cleanup() {
+      clearTimeout(popoverHoverOutTimer);
+    };
+  });
+
```
Saadnajmi, line 61: Could we move the call to `setTimeOut` for the same timer into the same effect? That way, the code related to the popoverOverOutTimer is all in one place?

Got that idea from this talk about when hooks were introduced, where the creator does the same thing -> https://youtu.be/dpw9EHDh2bM?t=2555 (starts at timestamp where he mentions cleanup)

rurikoaraki: Yeah, we could, but since it's in a callback right now I'd have to add some code which would set some state to say "please set a timer on the next rerender cycle" which would then set the timer. It seems kinda convoluted to me. I could move this code closer to the code  which sets the time if that would help?

Saadnajmi (approved):

ValentinaKozlova (approved):

PR #1839, opened 2022-07-11
[Draft] Add shadows to iOS FAB

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add shadows to FAB (which is an iOS only component)
Initially added shadows to the default macOS button, but still unsure about whether we still want to do this & which shadow should be added. See commit history to see the changes for macOS button (later reverted)

Notes/open questions:
- Unexpectedly, Shadow seem to behave correctly for buttons with a border radius. Might be an iOS/macOS only thing - there's some behavior where props related to Shadow get automatically applied to the children (might not be the case for other platforms)
- Maybe makes more sense to have a section in the Button test page for shadows rather than a section in the Shadow test page for shadows?
- The way shadows is implemented adds two additional views to any component that has a shadow on it. This may have some performance impacts. We could potentially only have add one additional view if we used something like rootWithShadow

Still planned for this PR (will mark as ready for review when these are done or determined to be not in scope):
- Add jest tests for button
- Not sure yet how a button with a custom shadow would work (Button.customize...) - for example a default button that doesn't have a shadow.

Not is scope for this PR:
- Other platforms (win32)
- Add shadows to the macOS default button
  - Need to decide which button variants should have which shadows
    - Currently, the primary and subtle buttons do not have shadows. This initial choice was based on the (likely outdated) designs. Maybe the primary button should be using brand shadows?
    - The default button component is using shadow2. This does not exactly match the (likely outdated) designs, shadow2 was just the closest to the designs
  - Also there are some other types of buttons that might need shadows: compound buttons, toggle buttons

### Verification
(ignore the changes to the icon)

| FAB without shadow | Before (key shadow only) | After (key + ambient) |
| --- |--- | --- |
|  noshadow | before | after |


lyzhan7:
I currently see this error for both FAB on iOS and the default button on macOS (before i reverted the shadow changes to macOS). I have googled this error and read up a bit on keys but don't understand how my changes are related.

Screen Shot 2022-07-11 at 2 20 50 PM

apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/styles.ts
```diff
@@ -31,6 +32,7 @@ export const GridButton = Button.compose({
     root: View,
     icon: Icon,
     content: Text,
+    shadow: Shadow,
```
Saadnajmi, line 35: Why are we adding this to FocusZone?

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -0,0 +1,31 @@
...
+    );
+  } else if (Platform.OS === 'ios') {
+    return (
+      <View style={{ padding: 24, margin: 16 }}>
+        <FAB style={commonTestStyles.vmargin} icon={{ svgSource: svgProps, width: 20, height: 20 }}>
+          FAB
+        </FAB>
+      </View>
+    );
+  } else {
```
Saadnajmi, line 28: To keep it simple, should we just keep this scoped to FAB for now?

packages/components/Button/src/Button.tsx
```diff
@@ -81,6 +83,13 @@ export const Button = compose<ButtonType>({
           {shouldShowIcon && iconPosition === 'after' && <Slots.icon {...iconProps} />}
         </Slots.root>
       );
+
+      const shouldShowShadow = !appearance && Platform.OS === 'macos';
+      if (shouldShowShadow) {
+        return <Slots.shadow>{buttonWithoutShadow}</Slots.shadow>;
+      } else {
+        return buttonWithoutShadow;
+      }
```
Saadnajmi, line 92: See comment above, could we scope the change to just FAB for this PR? Reason I say this is because I'd rather have a more general way to say "if no shadow token, don't render shadow" rather than explicitly checking the platform. If one day designers are like "JK no shadows on macOS", it would be more work than a simple token change to fix Button then.

packages/components/Button/src/Button.types.ts
```diff
@@ -42,6 +43,11 @@ export interface ButtonCoreTokens extends LayoutTokens, FontTokens, IBorderToken
    */
   spacingIconContentAfter?: number;
 
+  /**
+   * The depth of the shadow of the button
+   */
+  shadowDepth?: ShadowDepth;
```
Saadnajmi, line 49: I wonder if we have some public facing documentation describing our shadow ramp, that we could link here.

packages/components/Button/src/ButtonColorTokens.macos.ts
```diff
@@ -8,6 +8,7 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
     color: t.colors.defaultContent,
     borderColor: t.colors.defaultBorder,
     iconColor: t.colors.defaultContent,
+    shadowDepth: 'shadow2',
```
Saadnajmi, line 11: Same comment about excluding this for now.

packages/components/Button/src/ToggleButton/ToggleButton.tsx
```diff
@@ -18,6 +19,7 @@ export const ToggleButton = compose<ToggleButtonType>({
     root: View,
     icon: Icon,
     content: Text,
+    shadow: Shadow,
```
Saadnajmi, line 22: Added slot, but don't render it below? But also we might want to exclude this until we have a more general solution to not render a shadow if no token is specified?

packages/components/README.md
```diff
@@ -46,14 +46,15 @@ Let's look at how to write a complex component using the compose framework. This
 
 Each control dedicates file using the format [Button.types.ts](./Button/src/deprecated/Button.types.ts) to define the component's tokens, props, and slots.
 
-Button is essentially a composition of View which is a layout element, a Text element, and an Icon element. The Button's slots reflect this composition. The root slot is a View which will wrap another View, represented as a stack slot which will contain the image and text content.
+Button is essentially a composition of View which is a layout element, a Text element, an Icon element, and a Shadow element. The Button's slots reflect this composition. The root slot is a View which will wrap another View, represented as a stack slot which will contain the image and text content.
```
Saadnajmi, line 49: Nice catch!

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
lyzhan7: Sorry I forgot to push the commit that reverts the changes to macOS, in my description I confirmed that macOS is out of scope

apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/styles.ts
lyzhan7: I think something was complaining until I added this. I can double check if I need to keep it in

packages/components/Button/src/Button.tsx
lyzhan7: Yep sorry just forgot to push the commit that reverts the macOS stuff

packages/components/Button/src/ToggleButton/ToggleButton.tsx
lyzhan7: Yeah the way I set it up, things wouldn't build unless I added the slot everywhere...I can take another look to see if this can be optional

apps/fluent-tester/src/FluentTester/TestComponents/FocusZone/styles.ts
Saadnajmi: Ah, it was because it was a Button slot. Probably don't need it anymore.

packages/components/Button/src/Button.types.ts
lyzhan7: Nothing that I'm aware of but was planning to write Shadow documentation and think that would make a lot of sense to add there

Saadnajmi:
A note: Modifying FAB means that it will also affect FAB on Android. It might be worth setting up an Android emulator to make sure FBA isn't broken there.

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Platform, View } from 'react-native';
+import { FAB } from '@fluentui/react-native';
+import { commonTestStyles } from '../Common/styles';
+import { svgProps } from '../Common/iconExamples';
+
+export const ShadowButtonTestSection: React.FunctionComponent = () => {
+  if (Platform.OS === 'ios') {
+    return (
+      <View style={{ padding: 24, margin: 16 }}>
+        <FAB style={commonTestStyles.vmargin} icon={{ svgSource: svgProps, width: 20, height: 20 }}>
```
Saadnajmi, line 11: Nit: move `{ svgSource: svgProps, width: 20, height: 20 }` to it's own variable / line so it's not recreated everytime this code runs, causing unnecessary re-renders

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
lyzhan7: Yeah I forgot that in the first commit but should be fixed now

packages/components/Button/src/Button.tsx
```diff
@@ -1,6 +1,6 @@
 /** @jsx withSlots */
 import * as React from 'react';
-import { View } from 'react-native';
+import { Platform, View } from 'react-native';
```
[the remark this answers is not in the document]
lyzhan7: You review very quickly haha have a fix but only just pushed

packages/components/Button/src/Button.types.ts
Saadnajmi: It could be part of the SPEC.md file as a start. More detailed documentation is probably a designer problem

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Platform, View } from 'react-native';
+import { FAB } from '@fluentui/react-native';
+import { commonTestStyles } from '../Common/styles';
+import { svgProps } from '../Common/iconExamples';
+
+export const ShadowButtonTestSection: React.FunctionComponent = () => {
+  if (Platform.OS === 'ios') {
+    const iconProps = { svgSource: svgProps, width: 20, height: 20 };
+    return (
+      <View style={{ padding: 24, margin: 16 }}>
```
Saadnajmi, line 11: Nit: Move style creation inside a Stylesheet, or reuse an existing style.

packages/components/Button/src/FAB/FABColorTokens.ts
```diff
@@ -1,15 +1,14 @@
...
-import { shadowStyleFromTheme } from '../shadowStyle';
 
 export const defaultFABColorTokens: TokenSettings<ButtonCoreTokens, Theme> = (t: Theme): ButtonCoreTokens => ({
-  ...shadowStyleFromTheme(t, 'shadow8'),
   // coloring same as primary
   backgroundColor: t.colors.brandedBackground,
   color: t.colors.brandedContent,
   borderColor: t.colors.brandedBorder,
   iconColor: t.colors.brandedIcon,
+  shadowDepth: 'shadow8',
```
Saadnajmi, line 11: Since this isn't a color token, maybe move this to `FABTokens.ts`?
Also, `shadowDepth: t.shadows.shadow8`?

lyzhan7: For moving it to FABTokens.ts - this is where the shadow tokens from before were initially defined, and also when I briefly did macOS I put the tokens in ButtonColorTokens.macos.ts since I can specify which variants have shadows there. I do think it makes more sense in FABTokens.ts but wasn't sure if there was reasoning behind why the shadows were initially defined there

Saadnajmi: I was probably just lazy and put it there is the answer. Or I told someone else to stick it in there.

packages/components/README.md
rurikoaraki, line 49: This is talking about and giving code examples for the old Button so I'm not sure if it makes sense to add details about the Shadow here

packages/components/Button/src/Button.types.ts
```diff
@@ -148,6 +154,7 @@ export interface ButtonSlotProps {
   root: React.PropsWithRef<IViewProps>;
   icon: IconProps;
   content: TextProps;
+  shadow?: ShadowProps;
```
rurikoaraki, line 157: Is the plan to add a shadow slot to every button?

packages/components/README.md
Saadnajmi: Oh, this isn't for ButtonV1? We should revert this then. I assume the document for ButtonV1 is SPEC.md?

packages/components/Button/src/Button.types.ts
Saadnajmi: At least for macOS, we are meeting later today to discuss which variants should have shadows by default. For the final implementation, I have two ideas on this bit:

1) Create a Shadow slot that we optionally render if we have a shadow token specified. This will always add 2 extra views to the hierarchy. This is more or less what this PR is doing. We scoped it to FAB here since that Button will _always_ have a shadow. 

2) Create a "RootWithShadow" component where the shadow slots' innerView becomes the new root, and the "withShadow" bit is in charge of rendering a key and ambient shadow. This would only add 1 more view to the hierarchy, since it would take the place of the root slot. The issue is we would have to be more careful about making sure props are passed properly. Also.. we might then want a "PressableWithShadow" too if we do that.

rurikoaraki: For option 2 you could do something like what I do with the MenuTrigger's child, cloning the child element and adding the props needed to create a shadow.

packages/components/Button/src/Button.types.ts
lyzhan7: Made a task for myself to write documentation

packages/components/README.md
rurikoaraki: Yeah. This is explaining generally how components are built (it's the README in the components folder), but it's outdated and references the old framework.

packages/components/Button/src/FAB/FABColorTokens.ts
lyzhan7: Planning to refactor stuff so I can pass in t.shadows.shadow8 in a later PR

packages/components/README.md
```diff
@@ -126,6 +126,7 @@ export const settings: IComposeSettings<IButtonType> = [
     } as ViewProps,
     content: {},
     icon: {},
+    shadow: {},
```
lyzhan7, line 129: Will check if I can remove this too

packages/components/Button/src/Button.types.ts
lyzhan7: Added a task for myself to investigate this later, thanks for the suggestions!

apps/fluent-tester/src/FluentTester/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { Platform, View } from 'react-native';
+import { FAB, Text } from '@fluentui/react-native';
+import { commonTestStyles, stackStyle } from '../Common/styles';
+
+export const ShadowButtonTestSection: React.FunctionComponent = () => {
+  if (Platform.OS === 'ios') {
+    const CustomFABNoShadow = FAB.customize({ shadowDepth: undefined });
```
rurikoaraki, line 8: customized components should be created outside of render functions, since the function ends up creating a new object when run.

lyzhan7: Got it, thanks for catching that and explaining why

packages/components/Button/src/FAB/FAB.test.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { FAB } from './FAB';
+import { checkRenderConsistency, checkReRender } from '@fluentui-react-native/test-tools';
+import * as renderer from 'react-test-renderer';
+
+jest.mock('react-native/Libraries/Utilities/Platform', () => ({
```
lyzhan7, line 6: Got this off of stack overflow, by default I think all the jest tests might be running on android. Couldn't figure out how to have both a test for android + iOS in the same file, can look into it more if we want to test Android as well.

lyzhan7: https://stackoverflow.com/questions/43161416/mocking-platform-detection-in-jest-and-react-native

Saadnajmi: I think leaving a link to that above line 6 with a short explanation is fine. I know that's literally linking stackoverflow in our code but 🤷🏾

rurikoaraki: Make sure that the mock gets cleared on shutdown, otherwise it might carry over to tests that run afterward.

rurikoaraki: Also personally this isn't really worth linking something for. Mocking functions is in jest's documentation isn't it?

lyzhan7: From what I could tell, the mock for Platform wasn't carrying over to other tests (just put console.logs in the other button tests), I couldn't find much documentation about the scope of a mock except that "[The jest object is automatically in scope within every test file](https://jestjs.io/docs/jest-object#jestunmockmodulename)" but wasn't sure if that meant it would be out of scope in other files. Either way though, updated so that the mock is getting created before all the FAB tests and it gets unmocked at the end. Confirmed this by console.logging the platform in the beforeAll and afterAll methods, it returns iOS is the beforeAll and android in the afterAll method.

Saadnajmi (approved):

packages/components/Button/src/FAB/FAB.test.tsx
lyzhan7: @rurikoaraki going to resolve this, if the unmock code I added wasn't what you were looking for let me know and will fix in another PR!

PR #1844, opened 2022-07-12
Updated notification button style

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Updated action button style to use the same font color, weight, size as message. If no action is specified, a dismissal button is rendered instead.

### Verification

|--------------------------------------------|--------------------------------------------|
| 9 - Icon | 10 - Button |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Nit: Make title of PR more specific to Notification. Thought this was a PR for the FURN V1 button for a sec :)

packages/components/Notification/src/Notification.types.ts
```diff
@@ -39,6 +40,8 @@ export interface NotificationSlotProps {
   title?: ITextProps;
   message: ITextProps;
   action?: ButtonProps;
+  actionText?: ITextProps;
+  dismissIcon?: SvgProps;
```
Saadnajmi, line 44: I think actionText and action are Redundant now?
Also, could we keep the naming the same? so `actionIcon` maybe?

Saadnajmi: Also, could it be `IconProps` instead of `SvgProps` so that the developer can pass any type of icon we support in there?

Saadnajmi: ^ That suggestion also means some refactoring of ` getDismissIconPath` too

packages/components/Notification/src/Notification.helper.tsx
```diff
@@ -1,6 +1,42 @@
 import React from 'react';
+import { ButtonProps, ButtonTokens, ButtonV1 as Button } from '@fluentui-react-native/button';
+import { mergeProps, stagedComponent } from '@fluentui-react-native/framework';
 import { Path } from 'react-native-svg';
 
+export type ActionButtonColorStates = { disabledColor; hoveredColor; pressedColor; focusedColor };
+type ActionButtonProps = ButtonProps & ButtonTokens & ActionButtonColorStates;
+export const ActionButton = stagedComponent((props: ActionButtonProps) => {
```
Saadnajmi, line 8: nit, newlines between the declarations. I would also leave a simple 1-2 line comment explaining why ActionButton exists. Something like:


```
// We need to customize Notifiations' `action` slots' tokens based on Notification's variant prop. Compose doesn't let us easily do that via styling settings, so we introduce this helper component to customize tokens via props.

File: packages/components/Notification/src/NotificationTokens.ts [hunk 3]
```diff
@@ -27,53 +27,51 @@ export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme>
...
-      backgroundColor: '#EBF3FC',
-      color: '#0F6CBD',
-
-      disabledColor: '#2886DE',
+      backgroundColor: '#EBF3FC', // brandBackground4
+      color: '#0F6CBD', // brandForeground4
+      disabledColor: '#2886DE', // brandForegroundDisabled
+      /**
+       * None of the foreground tokens here have pressed versions so the foreground color with an alpha value was used.
+       * The FUA NotificationView was used to color match.
```
Saadnajmi, line 35: Nit: FUA -> "FluentUI Apple".

Saadnajmi (approved):

Saadnajmi (approved):

PR #1855, opened 2022-07-14
Shadows: add shadow to iOS FAB (attempt 2)

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

PR #1839 got reverted, this PR is to add the changes back. Since the first PR, there have been some package updates that may have fixed the errors that caused the PR to be backed out. It is also possible that the errors that caused the PR to get backed out were unrelated to the original change.

### Verification

Confirmed that yarn buildci and yarn test succeed locally.

Not sure if this will actually check anything new but after this gets checked in, planning to pull the latest master and confirm that yarn buildci still passes.


Saadnajmi:
Curious: does adding back 'test' to Shadows' package.json now pass?

lyzhan7:
> Curious: does adding back 'test' to Shadows' package.json now pass?

Seems to be pass locally, waiting now to see if the pipelines pass

Saadnajmi (approved):

PR #1860, opened 2022-07-15
Add in dark mode for notifications

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added in dark mode support for notification. Fluent 2 colors are not yet in the token pipeline for iOS so DynamicColorIOS is necessary for dark mode. DynamicColorIOS does not work with svgs so all svg icons will default to black in light mode and white in dark mode. This is a known issue that will be resolved once Fluent 2 colors are ready.

### Verification

| 10 - Button | Simulator Screen Shot - iPhone 13 Pro - 2022-07-19 at 14 01 32 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Notification/src/NotificationTokens.ts
```diff
@@ -1,6 +1,7 @@
 import { Theme } from '@fluentui-react-native/framework';
 import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { NotificationTokens } from './Notification.types';
+import { notificationColors } from './NotificationTokens.ios';
```
Saadnajmi, line 4: Yeah this will fail because metro won't include the base file `notificationTokens.ts` in your bundle, I'm pretty sure.

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -0,0 +1,63 @@
+import { DynamicColorIOS } from 'react-native';
+
+/**
+ * Fluent 2 colors are not yet tokenized for iOS so DynamicColorIOS is necessary for dark mode
```
Saadnajmi, line 4: Not: "not yet I'm the token pipeline"

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -1,7 +1,85 @@
...
+      paddingVertical: 12,
+    },
+    isBar: {
+      borderRadius: 0,
+      fontWeight: '400',
+    },
+    primary: {
+      backgroundColor: notificationColors.brandBackground4,
+      color: notificationColors.brandForeground4,
+      // color: '#0F6CBD',
```
Saadnajmi, line 33: delete?

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -1,7 +1,85 @@
...
+      disabledColor: notificationColors.brandForegroundDisabled,
+      /**
+       * None of the foreground tokens here have pressed versions so the foreground color with an alpha value was used.
+       * The FluentUI Apple NotificationView was used to color match.
+       */
+      pressedColor: '#0F6CBD30',
```
Saadnajmi, line 39: Could we either use `rgba(15, 108, 189, 0.19)` instead of the hex value, or leave a comment besides the opacity in decimal? I mostly wanted to highlight that the opacity of 20% is something we guesstimated by comparing with the native button.

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -1,7 +1,85 @@
...
+      disabledColor: notificationColors.foregroundDisabled2,
+      pressedColor: '#61616145',
+    },
+  } as NotificationTokens);
+
 /**
- * Fluent 2 colors are not yet tokenized for iOS so DynamicColorIOS is necessary for dark mode
+ * Fluent 2 colors are not yet in the token pipeline for iOS so DynamicColorIOS is necessary for dark mode
  */
 export const notificationColors = {
```
Saadnajmi, line 84: Nit, these don't need to be exported since they're only referenced in this file. Also, we _could_ move it to the top of the file so we see they're declared before using them, but idk it matters. That's more personal preference.

packages/components/Notification/src/NotificationTokens.ts
```diff
@@ -1,6 +1,7 @@
...
 
 export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme> = (t: Theme) =>
```
Saadnajmi, line 6: Since this file won't be used by iOS anymore, you can remove the iOS only props from line 14 and 20.

Saadnajmi:
Update screenshots?

packages/components/Notification/src/Notification.styling.ts
```diff
@@ -102,19 +100,5 @@ export const stylingSettings: UseStylingOptions<NotificationProps, NotificationS
...
-            alignSelf: 'center',
-            flex: 1,
-            height: 20,
-            width: 20,
-          },
-          color: tokens.color,
-        };
-      },
-      ['color'],
-    ),
```
Saadnajmi, line 118: why did the go away?

joannaquu: Button has the option to be an icon only button so I thought that would be more useful than having a svg inside an empty button. Also the old way meant that the svg wouldn't change colors when pressed.

packages/components/Notification/src/Notification.tsx
```diff
@@ -51,18 +50,46 @@ export const Notification = compose<NotificationType>({
...
-          {action ? <Slots.action>{action}</Slots.action> : !isBar && <Slots.dismissIcon>{getDismissIconPath()}</Slots.dismissIcon>}
+          {onActionPress ? (
+            action ? (
+              <Slots.action onClick={onActionPress}>{action}</Slots.action>
+            ) : (
+              <Slots.action
+                icon={dismissIconProps}
+                onClick={() => {
+                  console.log('Dismiss button tapped');
+                }}
```
Saadnajmi, line 80: This seems like test code that should be deleted.

packages/components/Notification/src/Notification.tsx
```diff
@@ -51,18 +50,46 @@ export const Notification = compose<NotificationType>({
...
+              />
+            )
+          ) : (
+            !isBar && (
+              <Slots.action
+                icon={dismissIconProps}
+                onClick={() => {
+                  console.log('Dismiss button tapped');
+                }}
```
Saadnajmi, line 89: Same here

packages/components/Notification/src/Notification.tsx
```diff
@@ -51,18 +50,46 @@ export const Notification = compose<NotificationType>({
...
+              />
+            )
+          )}
```
Saadnajmi, line 92: In general this section feels it got very complicated. Can we do any simplification here?

Saadnajmi (changes requested):

packages/components/Notification/src/Notification.tsx
joannaquu: Yeah it changed quite a bit after I talked to Jeanie and realized I had been reasoning about the action button wrong. At first I had assumed that if there is no action specified, a dismiss button is added (that's how it was). However, she differentiated between the action's text and the callback function itself. So if there's a callback and a text, it should render the text (line 74) but it there's a callback and no text, it should render a dismiss button (line 76-81). And the finally, toasts should have an action so if there's no callback, it should have a dismiss button (line 84-91)

Saadnajmi: Thanks, this looks cleaner now :) 

Could you perhaps leave a comment in code somewhere describing when to show the dismiss vs action button based on the props passed? Should we simplify to some booleans like `const shouldShowDismissIcon = blahblah`?

packages/components/Notification/src/Notification.tsx
```diff
@@ -62,7 +63,8 @@ export const Notification = compose<NotificationType>({
             {title && <Slots.title>{title}</Slots.title>}
             <Slots.message style={messageStyle}>{children}</Slots.message>
           </Slots.contentContainer>
-          {action ? <Slots.action>{action}</Slots.action> : !isBar && <Slots.dismissIcon>{getDismissIconPath()}</Slots.dismissIcon>}
+          {notificationButtonType === 'action' && <Slots.action onClick={onActionPress}>{action}</Slots.action>}
+          {notificationButtonType === 'dismiss' && <Slots.action icon={dismissIconProps} />}
```
Saadnajmi, line 67: Could we pass the notificationButtonType as a prop into NotificationButton, and let NotificationButton decide what to render inside itself? That way all we have to do is render `<Slots.action {...actionProps} >` here?

packages/components/Notification/src/Notification.helper.tsx
```diff
@@ -27,26 +31,55 @@ export const ActionButton = stagedComponent((props: ActionButtonProps) => {
...
+  };
+  return svgProps;
+}
+
+/**
+ * Depending on what props are passed in, either an action button or dismiss button will be rendered.
+ * If the props contain an action to be performed (an onActionPress callback function) and an action text, the action button will be rendered.
+ * If the props contain an action to be performed but no action text, the dismiss button will be rendered.
+ * If the props contain neither but the variant indicates a toast notification, the dismiss button will be rendered.
+ * In all other cases, no button will be rendered.
```
Saadnajmi, line 70: Could also go in SPEC.md

Saadnajmi (approved):

packages/components/Notification/src/Notification.tsx
```diff
@@ -51,9 +49,10 @@ export const Notification = compose<NotificationType>({
     }, ['isBar']);
 
     return (final: NotificationProps, ...children: React.ReactNode[]) => {
-      const { variant, icon, title, action, ...rest } = mergeProps(userProps, final);
-      const iconProps = createIconProps(icon);
+      const { variant, icon, title, action, onActionPress, ...rest } = mergeProps(userProps, final);
       const mergedProps = mergeProps<PressableProps>(rest, rootStyle);
+      const iconProps = createIconProps(icon);
+      const notificationButtonProps = createNotificationButtonProps(userProps);
```
Saadnajmi, line 55: Starting to sound awfully similar to a "useNotification" :D, we don't have to change that though. I'm still not sure what would count as "state" for us

Issue #743, opened 2021-06-24
Remove our UseAsPressable implementation, use the one from core


Saadnajmi:
https://github.com/microsoft/fluentui-react-native/tree/master/packages/utils/interactive-hooks/src/Pressability

Before RN-Core made Pressable public, we copied the source code of Pressable into FURN and converted it to typescript so we can use it. Now that everything is on RN 0.63, we should delete this.

A caveat. Maybe we should wait till everything _downstream_ is also RN 0.63. 

I bring this up because some of our automated Dep bumps fails due to an error In our Pressable copy-pasta, specifically the @types/react bumps. 

@JasonVMo as an FYI. I talked about it with him when we updated the repo to RN 0.63, but I didn't end up merging it.

Saadnajmi:
Gonna just tag some people on CXE who actually know / have used the "UseAsPressable" hooks to get their thoughts on it 

@samuelfreiberg @ksiler @rurikoaraki

PR #1868, opened 2022-07-18
Add onAccessibilityTap support for win32

lenahong:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes
- For pressable text, onPress is the default onAccessibilityTap value unless set otherwise
- Adding both examples where onPress is invoked by default and where onAccessibilityTap is overridden by setting a separate callback function

### Verification
- Tests added to Experimental Text page in tester app

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [X] Voiceover


packages/experimental/Text/src/Text.tsx
```diff
@@ -50,6 +51,15 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
 
+  const onAccTap =
+    onAccessibilityTap ??
+    React.useCallback(
+      (event?) => {
+        props.onPress(event);
+      },
+      [props.onPress],
+    );
+
```
Saadnajmi, line 62: Thanks for doing this! This feels like something `...pressableProps` should handle. I wonder if that's a valid upstream change in the future 🤔

lenahong:
> I looked into why this failed on macOS originally, and I have a theory.
> 
> macOS mostly inherits the API of RN on iOS. On iOS, `onAccessibilityTap` is a View prop, but not a Text prop. I don't see it in the view config / view manager for Text. However, Text does have the prop `onAccessibilityAction`. What's the difference between the two?

onAccessibilityAction takes in an event parameter to handle an action request, so the action supported needs to be predefined with accessibilityActions. OnAccessibilityTap on the other hand, will always be invoked by the system on tap(Narrator Key + Enter for windows) when accessible. So we do need support for both for pressible text, I believe? Adding @ksiler for input

ksiler:
> > I looked into why this failed on macOS originally, and I have a theory.
> > macOS mostly inherits the API of RN on iOS. On iOS, `onAccessibilityTap` is a View prop, but not a Text prop. I don't see it in the view config / view manager for Text. However, Text does have the prop `onAccessibilityAction`. What's the difference between the two?
> 
> onAccessibilityAction takes in an event parameter to handle an action request, so the action supported needs to be predefined with accessibilityActions. OnAccessibilityTap on the other hand, will always be invoked by the system on tap(Narrator Key + Enter for windows) when accessible. So we do need support for both for pressible text, I believe? Adding @ksiler for input

Yeah, we've been using `onAccessibilityTap` as a proxy to the Invoke Provider on windows which preforms a single, unambiguous action on a control that does not maintain state. Usually, this preforms the same action as a click and is preformed when someone uses Narrator Key + Enter on a selected control (if they're using Narrator).

`onAccessibilityAction`, on the other hand, is used to define additional accessibility actions used by assistive technology to interact with the control. Technically, this could include window's invoke action but since we've been using `onAccessibilityTap` for that purpose, it hasn't been hooked up (at least for rn-win32, I think rn-windows is the same)

packages/experimental/Text/src/Text.tsx
Saadnajmi: We should probably log that somewhere..... so for now I'll link this issue: https://github.com/microsoft/fluentui-react-native/issues/743

packages/experimental/Text/src/Text.tsx
```diff
@@ -51,14 +51,12 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
-      (event?) => {
-        props.onPress(event);
-      },
-      [props.onPress],
-    );
+  const onAccTap = React.useCallback(
+    (event?) => {
+      onAccessibilityTap ? onAccessibilityTap() : props.onPress(event);
+    },
+    [props.onPress],
```
Saadnajmi, line 58: `onAccessibilityTab` should also be in the dependency array? Also I thought we had an enlist rule that told us not to do`foo.bar` inside the array and instead pull out `bar` as a separate var. AKA, pull out onPress between lines 21-36 so that we can just have `onPress` instead of `props.onPress` in the array.

packages/experimental/Text/src/Text.tsx
lenahong: [Task: 6276444](https://office.visualstudio.com/OC/_sprints/taskboard/CXE%20React/OC/CY22Q3%20(Jul-Sep)/2207-Jul/Jul02?workitem=5525048)

Created task under V1 Pressable

Saadnajmi: We have a V1 Pressable task? :o

Saadnajmi (approved):
Not seeing a Redbox on macOS with this change anymore :)

PR #1890, opened 2022-07-21
Add animation to notification and customizable test page

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added initial animations for notifications. Non-bar notifications present in 0.6 seconds, with a damping ratio of 0.5. Bars present in 0.3 seconds, with a damping ratio of 1. Both hide in 0.25 seconds. Currently, the notifications present from the bottom of their parent (each test section), not the bottom the the screen.

### Customizable test page

| Non-bar                                       | Bar                                      |
| Simulator Screen Recording - iPhone 13 Pro - 2022-07-25 at 10 20 15 |Simulator Screen Recording - iPhone 13 Pro - 2022-07-25 at 13 49 49 |

### Animation
ezgif-2-44a46b30d2

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,128 +10,464 @@ const svgProps: SvgIconProps = {
...
+
+  const [hidden, setHidden] = React.useState<boolean>(!visible);
+  const height = React.useRef(new Animated.Value(0)).current;
+
+  const show = () => {
+    setHidden(false);
+    Animated.sequence([
+      Animated.timing(height, {
+        toValue: -50,
+        duration: 600,
```
Saadnajmi, line 25: Can we pull out these constants into const variables so we can easily reference them and change them?

Saadnajmi:
Gonna summon @chiuam for this one bc animations, in case you notice code smells I don't :P

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,128 +10,464 @@ const svgProps: SvgIconProps = {
...
+      }),
+      Animated.delay(3000),
+      Animated.timing(height, {
+        toValue: 0,
+        duration: 250,
+        useNativeDriver: true,
+      }),
+    ]).start(() => {
+      setHidden(true);
+      setVisible(false);
```
Saadnajmi, line 37: Do we need both variables? Can't we just set one / simplify our logic so we only use one?

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,128 +10,464 @@ const svgProps: SvgIconProps = {
...
+        toValue: -50,
+        duration: 600,
+        useNativeDriver: true,
+        easing: Easing.elastic(1.5),
+      }),
+      Animated.delay(3000),
+      Animated.timing(height, {
+        toValue: 0,
+        duration: 250,
+        useNativeDriver: true,
```
Saadnajmi, line 33: Because I happen to know that win32 does not have a native driver, it _might_ be worth pulling this out into a separate variable should we ever decide to bring notification to other platforms and need to do a. platform check. _Might_, I'm still unsure we'll ever bring this to desktop.

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,128 +10,464 @@ const svgProps: SvgIconProps = {
...
+      }).start(() => {
+        setHidden(true);
+      });
+    }
+  }, [visible, height]);
+
+  const animatedViewProps = {
+    transform: [{ translateY: height }],
+  };
```
Saadnajmi, line 173: Any chance we can pull out this logic into something re-usable so we don't need to keep copy-pasting everywhere?

Saadnajmi: An idea could be to combine all the tests into one test with a Picker to set the variant (similar to the pickers for Brand / Theme)

joannaquu: I did attempt to do this but I believe it violates some hook rules (hooks should only be called at the top level). [https://reactjs.org/docs/hooks-rules.html](url)

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
joannaquu: The main reason we need both variables is because one controls the animation and the other controls whether or not to render the notification. If we remove the hidden variable, lines such as 73, 130 will be updated to {visible && ( ... )}. The minute the hide button is clicked, the visible variable will flip and the notification will disappear before you see the hide animation.

packages/components/Notification/src/Notification.tsx
```diff
@@ -49,7 +49,7 @@ export const Notification = compose<NotificationType>({
     }, ['isBar']);
 
     return (final: NotificationProps, ...children: React.ReactNode[]) => {
-      const { variant, icon, title, action, onActionPress, ...rest } = mergeProps(userProps, final);
+      const { variant, visible, icon, title, action, onActionPress, ...rest } = mergeProps(userProps, final);
```
Saadnajmi, line 52: This doesn't seem to be used anywhere. Without the animation, does this prop actually hide the notification?

joannaquu: Oops, this was remaining from when visible was in Notification.tsx. Will remove

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import { Notification } from '@fluentui-react-native/notification';
 import { Test, TestSection, PlatformStatus } from '../Test';
-import { Animated, Button, Easing, View } from 'react-native';
+import { Animated, Button, Easing, Switch, TextInput, View } from 'react-native';
+import { Text } from '@fluentui-react-native/experimental-text';
+import { StyledPicker } from '../Common/StyledPicker';
+import { commonTestStyles as commonStyles } from '../Common/styles';
+import { NotificationVariant, NotificationVariants } from '../../../../../../packages/components/Notification/src/Notification.types';
```
Saadnajmi, line 8: Fix import here

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -14,10 +18,88 @@ const toastShowDuration = 600;
...
+
+  return (
+    <View>
+      <StyledPicker
+        prompt="Variant"
+        selected={variant}
+        onChange={onVariantChange}
+        collection={NotificationVariants.filter((variant) => variant.length < 10)}
+      />
+      <TextInput value={title} onChangeText={setTitle} style={{ borderWidth: 1 }} />
```
Saadnajmi, line 42: Nit: put style inside a Stylesheet so it's (maybe?) not recreated on every render

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -14,10 +18,88 @@ const toastShowDuration = 600;
...
+  const [showIcon, setShowIcon] = React.useState(true);
+  const [showTitle, setShowTitle] = React.useState(true);
+
+  return (
+    <View>
+      <StyledPicker
+        prompt="Variant"
+        selected={variant}
+        onChange={onVariantChange}
+        collection={NotificationVariants.filter((variant) => variant.length < 10)}
```
Saadnajmi, line 40: Why does it have to be less than 10? Is that your way of filtering the bar variants? :P

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -14,10 +18,88 @@ const toastShowDuration = 600;
...
+      <TextInput value={action} onChangeText={setAction} style={{ borderWidth: 1 }} />
+      <View style={commonStyles.switch}>
+        <Text>Show icon </Text>
+        <Switch value={showIcon} onValueChange={setShowIcon} />
+      </View>
+      <View style={commonStyles.switch}>
+        <Text>Show title </Text>
+        <Switch value={showTitle} onValueChange={setShowTitle} />
+      </View>
+      <View style={{ marginTop: 70 }}>
```
Saadnajmi, line 53: Same for this style

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,156 +14,251 @@ const svgProps: SvgIconProps = {
...
-      }}
-    >
-      Mail Archived
-    </Notification>
-  );
-};
+const toastShowDuration = 600;
+const barShowDuration = 300;
+const hideDuration = 250;
+const toastBounciness = 1.5;
```
Saadnajmi, line 20: nit: remake toast* to notification* ? `toastShowDuration -> notificationShowDuration`.
Also I think for constants, we usually use ALL_CAPS? So, `NOTIFICATION_SHOW_DURATION`?

Reason for name change is that "toast" is another name for notification, kind of an "alternate" name.
[Open UI](https://open-ui.org/components/toast.research) mentioned that can be named both. And it's actually the name of a native Android control as well (https://developer.android.com/guide/topics/ui/notifiers/toasts). To better distinguish ours / keep consistency, I think we should avoid the word toast :)

Saadnajmi:
Also update screenshots with new test page? :)

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
joannaquu: Yes, that was my lazy way of filtering for toast variants but I can make it more clear.

apps/fluent-tester/src/FluentTester/TestComponents/Notification/NotificationTest.tsx
joannaquu: Using ~/ doesn't seem to work. Any other ideas?

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,156 +13,263 @@ const svgProps: SvgIconProps = {
...
+const PrimaryBarWithAutoHide: React.FunctionComponent = () => {
+  const [visible, setVisible] = React.useState(false);
+  const onButtonPress = () => setVisible(!visible);
+
+  const [hidden, setHidden] = React.useState<boolean>(!visible);
+  const height = React.useRef(new Animated.Value(0)).current;
+
+  const show = () => {
+    setHidden(false);
+    Animated.sequence([
```
Saadnajmi, line 196: Is there any difference with the bar animation code?

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -1,6 +1,10 @@
 import * as React from 'react';
-import { Notification } from '@fluentui-react-native/notification';
+import { Notification, NotificationVariant, NotificationVariants } from '@fluentui-react-native/notification';
 import { Test, TestSection, PlatformStatus } from '../Test';
+import { Animated, Button, Easing, StyleSheet, Switch, TextInput, View } from 'react-native';
```
Saadnajmi, line 4: I wouuuuld not use "Button" from React Native. That control is pretty old. I would prefer we use our own Button here.

packages/components/Notification/src/Notification.tsx
```diff
@@ -46,9 +46,9 @@ export const Notification = compose<NotificationType>({
       return { marginHorizontal: marginHorizontal };
     }, ['isBar']);
     const messageStyle: ViewStyle = useMemo(() => {
-      const alignSelf = isBar ? 'center' : 'flex-start';
+      const alignSelf = userProps.onActionPress ? 'flex-start' : 'center';
       return { alignSelf: alignSelf };
-    }, ['isBar']);
+    }, ['userProps']);
```
Saadnajmi, line 51: Why this change? I'd also rather pull out `const onActionPress = userProps.onActionPress` and use that here. I think `useMemo` is more efficient when the keys in the dependency array are simpler objects.

packages/components/Notification/src/Notification.types.ts
```diff
@@ -30,10 +30,30 @@ export interface NotificationProps {
    * Notification variants: 'primary' | 'neutral' | 'primaryBar' | 'primaryOutlineBar' | 'neutralBar' | 'danger' | 'warning'
    */
   variant: NotificationVariant;
+
+  /**
+   * Optional icon that appears on the left side of non-bar notifications
+   */
   icon?: IconSourcesType;
```
Saadnajmi, line 37: Thank you!

packages/components/Notification/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export { Notification } from './Notification';
-export type { NotificationTokens, NotificationProps, NotificationSlotProps, NotificationType } from './Notification.types';
+export * from './Notification.types';
```
Saadnajmi, line 2: I think we prefer not Tod o export * because of some Javascript bundler optimization thing. Could we actually go back to explicitly exporting the things?

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
joannaquu: Besides the show duration and lack of easing, no

packages/components/Notification/src/Notification.tsx
joannaquu: After talking to Jeanie, she mentioned that bar notifications can also have action buttons so the alignment of the message no longer depends on variant. Rather, it depends on whether or not there is an action (onActionPress)

packages/components/Notification/src/Notification.types.ts
joannaquu: ofc!

packages/components/Notification/src/Notification.tsx
Saadnajmi: Oh ok, so it's a tiny change that is paired with your bigger change, makes sense. Note about pulling out `onActionPress` still holds too then

joannaquu: Cool, thanks!

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,156 +14,262 @@ const svgProps: SvgIconProps = {
...
+  const height = React.useRef(new Animated.Value(0)).current;
+
+  const show = () => {
+    setHidden(false);
+    Animated.sequence([
+      Animated.timing(height, {
+        toValue: SHOW_HEIGHT,
+        duration: NOTIFICATION_SHOW_DURATION,
+        useNativeDriver: USE_NATIVE_DRIVER_IOS,
+        easing: Easing.elastic(NOTIFICATION_BOUNCINESS),
```
harrieshin, line 127: what are some of the option here? not sure it is the pr captioning issue but I would expect there should be more damping effect after notification is shown?

joannaquu: So `toValue` is the height that the notification will reach after `duration` milliseconds. There is no option for damping effect in Animations but it can be mocked with an elastic easing, which causes the notification to reach the `toValue` height, overshoot, and then return to the `toValue` height.

Saadnajmi: https://reactnative.dev/docs/easing Does this help?

Saadnajmi: Could we use spring anywhere? I see that has a configurable damping
https://reactnative.dev/docs/animated#spring

packages/components/Notification/src/SPEC.md
```diff
@@ -2,36 +2,35 @@
...
+| PrimaryBar        | `#EBF3FC` | `#082338` |
+| PrimaryOutlineBar | `#FFFFFF` | `#000000` |
+| NeutralBar        | `#F0F0F0` | `#3D3D3D` |
+| Danger            | `#FDF6F6` | `#3F1011` |
+| Warning           | `#FFFBD6` | `#4C4400` |
 
 ## Icon
 
-Toast notifications can include an optional `icon` that appears to the left of the `message`.
+Non-bar notifications can include an optional `icon` that appears to the left of the `message`.
```
PPatBoyd, line 17: nit Can we identify which notifications this applies to in the positive form?

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
```diff
@@ -9,156 +14,262 @@ const svgProps: SvgIconProps = {
...
+const PrimaryBarWithAutoHide: React.FunctionComponent = () => {
+  const [visible, setVisible] = React.useState(false);
+  const onButtonPress = () => setVisible(!visible);
+
+  const [hidden, setHidden] = React.useState<boolean>(!visible);
+  const height = React.useRef(new Animated.Value(0)).current;
+
+  const show = () => {
+    setHidden(false);
+    Animated.sequence([
```
PPatBoyd, line 196: Did you happen to consider using LayoutAnimations for this animation?  Apologies for asking so late after you've been talking through this animation w/ Saad for a while, there are a couple things we need to mind with animations cross-platform.

react-native-win32 doesn't have a Native Animation Driver, so it will use the JS Animation Driver and worse busy-lock the app for the duration of the animation.  If we move forward w/ Animated here should the animations be disabled for win32, or the component not offered on win32 until the Native Animation Driver is available?

LayoutAnimation as an alternative does have an implementation for win32 and would change the implementation to only use layout to determine position vs laying out to a pre-animation position and then animating to its intended position.

Actually :( I'm realizing the more pressing issue for win32 is that the `transform` property is poorly implemented and hit testing won't work as expected (the button will activate if you click its laid out position, not its transformed position). I realize this is coming as a macOS/iOS-firstpriority component but the cross-plat impl will need to be kept off of win32 or not animated for win32 until we get the platform gaps filled that I mentioned above :\

PPatBoyd (changes requested):

PPatBoyd (commented):
realized after Saad pointed it out to me that the animations are in the test page -- I'll leave the comments for conversation but nixing the 'request changes' review state

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
harrieshin: i would like to comment notification so far is very mobile concept. i haven't seen anything in desktop design.

PPatBoyd (commented):
.

PPatBoyd (dismissed):

PPatBoyd (commented):
unsetting to just a comment aaaaaahhhhhhh github UI...

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
PPatBoyd: > i would like to comment notification so far is very mobile concept. i haven't seen anything in desktop design.

Would limiting platforms down to iOS/android make sense?

Saadnajmi: I think this will be a mobile only FURN component in the long run. I saw web had a similar looking "Alert", but Ididn't see it in FluentUI React v9. I imagine if/when FURN implements that, it would be a separate component. 

Regarding win32 animation concerns, I guess this is another point towards "Let's keep the animation method separate so that the client can customize it". I asked Joanna to not touch `Notification.tsx` as much as possible and try to keep the Notification _view_ and Notification _animation_ separate as much as possible :)

packages/components/Notification/src/SPEC.md
joannaquu: Originally Saad had requested that we retire the toast name but after looking at FUA, I'll be adding them back.

apps/fluent-tester/src/TestComponents/Notification/NotificationTest.tsx
joannaquu: hm spring has a configurable damping but no duration

PR #1896, opened 2022-07-22
useViewCommandFocus: Check that Focus exists in ViewManager first

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

See: https://github.com/microsoft/fluentui-react-native/pull/1535#pullrequestreview-1047650737

This change should prevent some crashes we see on Platforms where RCTView doesn't have a `focus` command exposed. This also happens to fix the Redbox we kept seeing on iOS :)  

### Verification

E2E tests in CI should be good.

iOS no longer Redboxes on button clicks

https://user-images.githubusercontent.com/6722175/180575666-cdb2b8a3-7399-4768-8cf4-2807032f140e.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/utils/interactive-hooks/src/useViewCommandFocus.ts
+++ b/packages/utils/interactive-hooks/src/useViewCommandFocus.ts
@@ -29,9 +29,12 @@ export function useViewCommandFocus(
        * Add focus() as a callable function to the forwarded reference.
        */
       if (localRef) {
-        localRef.focus = () => {
-          UIManager.dispatchViewManagerCommand(findNodeHandle(localRef), UIManager.getViewManagerConfig('RCTView').Commands.focus, null);
-        };
+        const focusCommand = UIManager.getViewManagerConfig('RCTView').Commands.focus;
+        if (localRef && focusCommand) {
+          localRef.focus = () => {
+            UIManager.dispatchViewManagerCommand(findNodeHandle(localRef), focusCommand, null);
+          };
+        }
       }
     },
   });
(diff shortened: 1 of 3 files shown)


Saadnajmi:
JK, this breaks on iOS where I was hoping it would fix it. Hold...

packages/utils/interactive-hooks/src/useViewCommandFocus.ts
```diff
@@ -30,7 +30,12 @@ export function useViewCommandFocus(
        */
       if (localRef) {
         localRef.focus = () => {
-          UIManager.dispatchViewManagerCommand(findNodeHandle(localRef), UIManager.getViewManagerConfig('RCTView').Commands.focus, null);
+          const commands = UIManager.getViewManagerConfig('RCTView').Commands;
+          if (localRef && 'focus' in commands) {
```
harrieshin, line 34: js newbie: why do we have to check localRef here again ?

Saadnajmi: Not a newbie question at all. A lot of this code is actually copied from RN, and IMO we should get rid of it from FURN or fix RN Core/macOS/Windows so we don't need this.

I think what this hook overall is doing is making sure that any of our controls' refs (aka React's version of a pointer) has the method `focus()` attached to it. How it's doing that with `setAndForwardRef` (which itself is a local copy of the same file inside React Native) is not clear to me. 

What this change does is make sure the native platform's RCTView does indeed have a Command called "focus" before trying to add one. Now that I type this out, I wonder if this would work on Android because I don't think they name their native view RCTView...

Saadnajmi: oh, you mean the double check. Probably don't need that, I'l fix

harrieshin (approved):

PR #1897, opened 2022-07-22
Add shadows to notification

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding shadows to toast notifications. Bar notifications do not have shadows.

### Verification

| Simulator Screen Shot - iPhone 13 Pro - 2022-07-22 at 15 33 04 | 13 - Shadow |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Notification/src/Notification.styling.ts
```diff
@@ -29,6 +29,7 @@ export const stylingSettings: UseStylingOptions<NotificationProps, NotificationS
           justifyContent: 'space-between',
           ...borderStyles.from(tokens, theme),
           ...layoutStyles.from(tokens, theme),
+          ...shadowStyles.from(tokens, theme),
```
Saadnajmi, line 32: I don't think we need this?

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -3,6 +3,11 @@ import { TokenSettings } from '@fluentui-react-native/use-styling';
 import { NotificationTokens } from './Notification.types';
 import { DynamicColorIOS } from 'react-native';
 
+const emptyShadowStyle = {
+  ambient: { x: 0, y: 0, blur: 0, color: '#00000000' },
+  key: { x: 0, y: 0, blur: 0, color: '#00000000' },
+};
```
Saadnajmi, line 9: Heh. More reason to ask design to make this part of the token set. :(

Saadnajmi (approved):

harrieshin:
based on your screenshot how big is your shadow frame? it seems like almost the entire democontroller background is darker not just near the region of notification?

joannaquu:
> based on your screenshot how big is your shadow frame? it seems like almost the entire democontroller background is darker not just near the region of notification?

Nice catch, I had used one of the preset shadows (shadow16 based off of FUA) but after digging into the actual values, they seem to have changed since NotificationView was coded.

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -3,6 +3,16 @@ import { TokenSettings } from '@fluentui-react-native/use-styling';
...
+
+const notificationShadowStyle = {
+  ambient: { x: 0, y: 8, blur: 8, color: '#00000024' },
+  key: { x: 0, y: 0, blur: 1, color: '#0000001f' },
+};
+
```
Saadnajmi, line 15: @TravisSpomer Not sure if you're the correct person to ask for this. Is there a way we can get this "shadow0" style into the token pipeline? Basically a token to say "render no shadows". I know it sounds a little counterintuitive, but would help simplify some code on our end :)

TravisSpomer: @saadnajmi Yeah, I checked and there isn't an "official" `shadow0` token, but it seems entirely reasonable to me. If your non-iOS colleagues are fine with it we can add it to the shadow tokens, or it can just be part of the exported iOS code with no actual token backing it. Start up a group chat or just let me know which y'all prefer and I'll help out wherever I can.

rurikoaraki: My only two cents here is that we should look into whether setting any shadow at all has any render implications. Like if cycles are spent drawing an invisible shadow/it ends up inserting some new element because there's a shadow. It's not great to end up with two extra Views in the UI tree if they don't add value.

TravisSpomer: Agreed 100%. I hereby clarify my response to say that it seems entirely reasonable to me *only if it can be done in a way that does not impact performance.* 🙂 I'm not familiar enough (well, at all) with iOS dev to have any specific advice there.

Saadnajmi: Agree that not rendering a shadow is the preferable option. The issue I saw we ran into is this:

Control A by default renders shadow8, so its at the top level of controlATokens.ts,

```
controlATokens: {
   color: t.colors.brandForeground
   shadow: t.shadows.shadow8
   ...
}
```


Control A's variant "foo" does not render a shadow ( in this case, Notification's variant "bar"). We generally override this by nesting some control token definitions inside the controlATokens.ts file. But, what do we override the shadow as?


```
controlATokens: {
   color: t.colors.neutralForeground
   shadow: t.shadows.shadow8
   ...
   foo: {
     color: t.colors.brandForeground
     shadow: ???
   }
}
```
We can pass `undefined`, and that could work, but it looked kinda hacky. We could alternatively in the `Shadow` component check if the shadow-to-render is shadow) or undefined, and then not render 2 extra views and just return `{children}`. Regardless, this is one of the questions that came up as we implemented shadows both as a slot and into some controls. We can follow up offline :)

Saadnajmi: Note that this issue is specific to how FURN implements tokens and overriding per variant, @TravisSpomer .I guess we should decide what we want to do internally in FURN before making a design ask.

TravisSpomer: Sure, no problem. Just send me a boop once you've settled on what would work best.

PR #1906, opened 2022-07-26
Testing MacOS CI

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR integrates macOS E2E testing within our CI Pipeline. Specifically, it clicks on every test page in the app and validates no crashes.

I had to separate out the macOS and iOS pipeline into separate jobs because the previous integration using Matrices made it nearly impossible to integrate with (compile-time and runtime discrepancies).

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.ado/azure-pipelines.yml
```diff
@@ -70,20 +70,13 @@ jobs:
           JAVA_HOME: $(JAVA_HOME_11_X64)
           PATH: $(JAVA_HOME_11_X64)/bin;$(PATH)
 
-  - job: ApplePR
-    displayName: Apple PR
+  - job: MacPR
+    displayName: MacOS PR
```
Saadnajmi, line 74: Nit: Mac or macOS, but not MacOS. I generally prefer "macOS" to match "iOS"

.ado/templates/e2e-testing-macos.yml
```diff
@@ -0,0 +1,34 @@
+steps:
+  - script: |
+      yarn appium driver install mac2
+    workingDirectory: apps/fluent-tester
+    displayName: 'Installing appium mac2 driver'
+
+  - script: |
+      yarn macos
+    workingDirectory: apps/fluent-tester
+    displayName: 'start the server'
```
Saadnajmi, line 10: Technically this command both builds and runs the test app AND starts the server

apps/fluent-tester/wdio.conf.macos.js
```diff
@@ -113,8 +113,22 @@ exports.config = {
...
+      return fluentTesterWindow != null;
+    },
+    {
+      timeout: 10000,
+      timeoutMsg: 'Could not find the FluentTester window. Cannot maximize app.',
+      interval: 500
+    });
+
+    const maxButton = fluentTesterWindow.$('//*[@identifier="_XCUI:FullScreenWindow" and @elementType=9]');
+    maxButton.click();
```
Saadnajmi, line 130: Could we leave a more detailed comment to the tune of "appium doesn't have an easy way to maximize the test app window, so we look up the maximize window button directly through it's XCUI identifier"?

Saadnajmi (approved):

.ado/templates/e2e-testing-macos.yml
samuelfreiberg: Ugh, I know. I need to add a comment explaining the reasoning behind this (same issue for Windows).

Basically, if I run `yarn start`, the command prompt runs continuously because there's no exit command. It won't process the next step.

This way, it opens up a second command prompt for the react-native server, rather than running it on the main one.

I'm sure there may be a better way to do this, but when I first implemented the Windows testing I couldn't find one. Maybe we can revisit it again

Saadnajmi: At least in bash, you can add "&" to any job to background it. So "yarn start &" would run it in the background. I remember I experimented with that when we set up the Apple CI checks. I ultimately decided not to go with it because I didn't like the idea of starting background jobs and never closing them.

https://stackoverflow.com/questions/44222883/run-a-shell-script-and-immediately-background-it-however-keep-the-ability-to-in

Saadnajmi: Good to know though, that this CI check relies on `yarn macos`  auto-launching the server. I had played around with a change where we don't do that. Glad I kept it in 😅

samuelfreiberg: Hmm, I didn't know about the "&" syntax. Maybe I'll try it. I wonder if we can force quit that terminal after we're done (or if it's even necessary, since E2E testing is the last thing that's run, and then the job is over)

apps/fluent-tester/wdio.conf.macos.js
samuelfreiberg: Improved comment in newest iteration

Saadnajmi:
Another thought: do we need to start the metro server? If we bundle first, then the test app will embed the JS bundle in the final app, so you can run without the need for a server?

.ado/templates/e2e-testing-macos.yml
Saadnajmi: Could probably find the job ID and send a `kill` command. Or just let ADO destroying the VM handle it for us :) 

Or just not worry about it, and just run `yarn macos` :D

samuelfreiberg:
> Another thought: do we need to start the metro server? If we bundle first, then the test app will embed the JS bundle in the final app, so you can run without the need for a server?

I'm actually not sure - I'll remove the task to see the behavior. Run I run `yarn macos` locally, I see another command prompt opening that starts the server, so I assumed we needed it running

.ado/templates/e2e-testing-macos.yml
samuelfreiberg: Looks like I didn't even need to start the server - thanks for the tip. Resolving this now :)

PR #1921, opened 2022-07-29
Switch Control Win32

nabil-k:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added the Switch control for Win32. The control is not fully completed yet. This PR just has the Switch's core props implemented as well as tokens.

I still need to work on accessibility, testing, and some animation issues.

image

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,58 @@
...
+    onChange && onChange(newCheckedState, e);
+    LayoutAnimation.configureNext(LayoutAnimation.create(200, LayoutAnimation.Types.linear, LayoutAnimation.Properties.scaleX));
+    setCheckedState(newCheckedState);
+  };
+
+  const onClickWithFocus = useOnPressWithFocus(focusRef, toggleCallback);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onClickWithFocus });
+  const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
+  const hasTogglePattern = props.accessibilityActions && !!props.accessibilityActions.find((action) => action.name === 'Toggle');
+  console.log(props.labelPosition);
```
ValentinaKozlova, line 28: Do you need cosole.log here?

packages/components/Switch/src/SwitchTokens.ts
```diff
@@ -0,0 +1,65 @@
+import { Theme } from '@fluentui-react-native/framework';
+// import { globalTokens } from '@fluentui-react-native/theme-tokens';
```
ValentinaKozlova, line 2: nit: commented line

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { SWITCH_TESTPAGE } from './consts';
+import { View, StyleSheet } from 'react-native';
+import { Switch } from '@fluentui-react-native/switch';
+
+const styles = StyleSheet.create({
+  container: {
```
ValentinaKozlova, line 8: container style exists in Common/styles.ts - `settingsPicker` object

packages/components/Switch/src/Switch.styling.ts
```diff
@@ -0,0 +1,95 @@
...
+];
+
+export const stylingSettings: UseStylingOptions<SwitchProps, SwitchSlotProps, SwitchTokens> = {
+  tokens: [defaultSwitchTokens, switchName],
+  states: switchStates,
+  slotProps: {
+    root: buildProps(
+      (tokens: SwitchTokens) => ({
+        style: {
+          display: 'flex',
```
ValentinaKozlova, line 25: Is there a reason why we put `display: 'flex` everywhere if it's a default display value? I see this not only in your component

packages/components/Switch/src/Switch.styling.ts
```diff
@@ -0,0 +1,95 @@
...
+    track: buildProps(
+      (tokens: SwitchTokens, theme: Theme) => ({
+        style: {
+          flexDirection: 'row',
+          height: 20,
+          borderRadius: 50,
+          width: 40,
+          backgroundColor: tokens.background,
+          borderColor: tokens.stroke,
+          borderWidth: 1,
```
ValentinaKozlova, line 55: I'm not sure you need to add here borderColor/Width/Radius because you have then `...borderStyles.from(tokens, theme)`

packages/components/Switch/src/Switch.tsx
```diff
@@ -0,0 +1,54 @@
+/** @jsx withSlots */
+import { View, Text } from 'react-native';
+import { switchName, SwitchType, SwitchState, SwitchProps } from './Switch.types';
+import { stylingSettings } from './Switch.styling';
+import { compose, mergeProps, withSlots, UseSlots } from '@fluentui-react-native/framework';
+import { useSwitch } from './useSwitch';
+/**
+ * A function which determines if a set of styles should be applied to the compoent given the current state and props of the button.
```
ValentinaKozlova, line 8: nit: props of the Switch.

packages/components/Switch/src/Switch.tsx
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param layer The name of the state that is being checked for
+ * @param state The current state of the button
+ * @param userProps The props that were passed into the button
+ * @returns Whether the styles that are assigned to the layer should be applied to the button
```
ValentinaKozlova, line 13: Should it be button here or Switch?

samuelfreiberg:
I'd highly recommend adding E2E tests in this PR. That way, as you're iterating on this control, you will have continuous validation that you're not regressing anything.

E2E testing should always been implemented along with the base component for this purpose.

packages/components/Switch/src/useSwitch.ts
nabil-k: Left it behind on accident thanks!

packages/components/Switch/src/Switch.tsx
nabil-k: It should be Switch

packages/components/Switch/src/Switch.styling.ts
nabil-k: Personally I just do it out of habit from web development. Not sure why others do it though.

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Switch defaultChecked={false} label={'Default Checked False'} />
+      <Switch defaultChecked={true} label={'Disabled Default Checked True'} disabled />
+      <Switch defaultChecked={false} label={'Disabled Default Checked False'} disabled />
+    </View>
+  );
+};
+
+const OnChangeUsage: React.FunctionComponent = () => {
+  const defaultDisplaySquare = true;
+  const [displaySquare, setDisplaySquare] = React.useState(defaultDisplaySquare);
```
ksiler, line 33: Instead of using `deafultDisplaySquare` to set the initial value of `displaySquare`, you can just use `React.useState(true);`

ValentinaKozlova:
It would be good to have screenshots. I'm sure your control looks great :)

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const toggleSquare = (checked) => {
+    setDisplaySquare(checked);
+  };
+
+  return (
+    <View style={styles.container}>
+      <Switch defaultChecked={defaultDisplaySquare} onChange={toggleSquare} />
```
ksiler, line 41: Does `toggleSquare` do anything here if the value of checked is never passed?

I would rewrite this as `<Switch defaultChecked={displaySquare} onChange={() => setDisplaySquare(!displaySquare)} />`

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,99 @@
...
+export const SwitchTest: React.FunctionComponent = () => {
+  const status: PlatformStatus = {
+    win32Status: 'Experimental',
+    uwpStatus: 'Backlog',
+    iosStatus: 'Backlog',
+    macosStatus: 'Backlog',
+    androidStatus: 'Backlog',
+  };
+
+  const description = 'Switch is a control that has 2 mutually exclusive states.';
```
ksiler, line 96: nit: I would write out the number 2 as "two"

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import { useAsPressable, useKeyProps, useOnPressWithFocus, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import { SwitchProps, SwitchInfo } from './Switch.types';
+import { memoize } from '@fluentui-react-native/framework';
+import { AccessibilityState, LayoutAnimation } from 'react-native';
+import { InteractionEvent } from '@fluentui-react-native/interactive-hooks';
+
+export const useSwitch = (props: SwitchProps): SwitchInfo => {
+  // attach the pressable state handlers
```
ksiler, line 9: this comment seems out of place

packages/components/Switch/src/Switch.styling.ts
ValentinaKozlova: I had the same habit :)

packages/components/Switch/src/Switch.styling.ts
```diff
@@ -0,0 +1,94 @@
...
+      }),
+      [],
+    ),
+    onOffText: buildProps(
+      () => ({
+        style: {},
+      }),
+      [],
+    ),
+  },
```
joannaquu, line 93: If there are no props I believe you can leave it out?

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
nabil-k: Replaced it with settingPicker

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
nabil-k: I got rid of `defaultDisplaySquare`, using just `true`

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
nabil-k: Yeah `toggleSquare` wouldn't do anything if the checked value was never passed, but the job of onChange is to pass the updated checked value. So we don't need to worry about.

The `defaultChecked` prop is only for setting the inital checked state. Users do not have to worry about managing the checked state. The checked state is managed by the control. If the users would like to manage the checked state themselves they would need to to use the `checked` prop instead.

packages/components/Switch/src/useSwitch.ts
nabil-k: It was, thanks for catching that :)

packages/components/Switch/src/Switch.styling.ts
nabil-k: Yeah I think it's better if I just leave them out until I need them.

packages/components/Switch/src/Switch.styling.ts
nabil-k: If there are different `borderRadius` for more than one slot I'm guessing I wouldnt be able to use `...borderStyles.from(tokens, theme)`? I think I can get borderColor and borderWidth out of the styles and instead be applied through `...borderStyles.from(tokens, theme)`. But I don't think I can do the same for `borderRadius`. Would it be better to just remove `...borderStyles.from(tokens, theme)`?

ValentinaKozlova: Why won't you be able to use `...borderStyles.from(tokens, theme)`? It has the same properties you need

nabil-k: Sorry I assumed I had to remove `borderRadius` on the other slot as well. But I can just leave that slot as it is and update the track slot. I've made the changes now :)

packages/components/Switch/src/Switch.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const Slots = useSlots(userProps, (layer) => switchLookup(layer, switchInfo.state, switchInfo.props));
+
+    // now return the handler for finishing render
+    return (final: SwitchProps) => {
+      const { label, offText, onText, labelPosition, accessibilityLabel, ...mergedProps } = mergeProps(switchInfo.props, final);
+      const onOffText = switchInfo.state.toggleOn ? onText : offText;
+      const displayOnOffText = !!offText || !!onText;
+
+      return (
+        <Slots.root {...mergedProps} accessibilityLabel={label}>
```
ksiler, line 42: Setting `accessibilityLabel` here isn't necessary since you set it in useSwitch.ts

ValentinaKozlova (approved):

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+  };
+};
+
+const getAccessibilityState = memoize(getAccessibilityStateWorker);
+function getAccessibilityStateWorker(disabled: boolean, accessibilityState?: AccessibilityState) {
+  if (accessibilityState) {
+    return { disabled: disabled, ...accessibilityState };
+  }
+  return { disabled: disabled };
+}
```
ksiler, line 94: Can ignore if addressing accessibility in another PR - I would add `checked: boolean` as one of the parameters for `getAccessibilityStateWorker` since it's an expected state of this control and set it equal to `checkedState` on line 70 and remove the `accessibilityState` prop you create on line 45.

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+      LayoutAnimation.configureNext(LayoutAnimation.create(200, LayoutAnimation.Types.linear, LayoutAnimation.Properties.scaleX));
+      setCheckedState(newCheckedState);
+    },
+    [onChange, setCheckedState, checkedState, checked],
+  );
+
+  const onClickWithFocus = useOnPressWithFocus(focusRef, toggleCallback);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onClickWithFocus });
+  const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
+  const currentCheckedState = checked ?? checkedState;
```
ksiler, line 43: What's the purpose of `currentCheckedState`? Can you not just used `checkedState` if `checkedState` is being set to `checked` if the `checked` prop is provided (on line 26 and 27)?

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+import { memoize } from '@fluentui-react-native/framework';
+import { AccessibilityState, AccessibilityActionEvent, LayoutAnimation, AccessibilityActionInfo } from 'react-native';
+import { InteractionEvent } from '@fluentui-react-native/interactive-hooks';
+
+export const useSwitch = (props: SwitchProps): SwitchInfo => {
+  const defaultComponentRef = React.useRef(null);
+  const {
+    onChange,
+    checked,
+    defaultChecked,
```
ksiler, line 13: Are the `defaultChecked` and `checked` props mutually exclusive or can they both be set? What's the expected behavior if they're both set? 

I would think setting `defaultChecked` means you're setting the initial value for Switch and that it would be mutually exclusive to `checked`. Using `checked` would mean you're defining the checked state of Switch (so setting `defaultChecked` wouldn't matter) and you intend to control the value of checked at a higher level instead of letting the Switch manage the state.

If they are mutually exclusive, I would add a check that warns the user to use one or the other and not set both.

apps/fluent-tester/src/TestComponents/Switch/E2ESwitchTest.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [showTextChecked, setShowChecked] = React.useState(false);
+
+  const onToggleChecked = (checked: boolean) => {
+    setShowChecked(!checked);
+  };
+
+  return (
+    <Stack style={stackStyle}>
+      <Switch testID={SWITCH_TEST_COMPONENT} label={'Checked'} checked={showTextChecked} onChange={onToggleChecked} />
+      {showTextChecked && <Text testID={SWITCH_TOGGLE_ON}>Checked Switch toggled to true</Text>}
```
samuelfreiberg, line 18: I'd change a couple things about this logic:

The goal of showing this text is not to test the checked state of the Toggle; the goal is to validate that the `onChange()` callback was fired. Ideally, we'd be able to spy on the onChange() function and determine when it's been called, but this UI Automation doesn't have that capability.

Because of that, we have to use this hacky-ish workaround to show text if the onChange() callback was fired.

Currently, it looks like this logic tests for the state of the `checked` prop for Toggle, which is not the goal (that's very easily done in a different way). The goal is to check if the onChange() was fired.

Please look at how Checkbox does this (same exact logic for onChange()). There should only be one constant `SWITCH_ON_PRESS`.

Relevant files + methods:
CheckboxPageObject.ts - didOnChangeCallbackFire()
Checkbox.spec.win.ts -   it("Click on a Checkbox -> Validate it toggles correctly AND calls the user's onChange")`
CheckboxE2ETest.tsx - `{checkboxPressed ? <Text testID={CHECKBOX_ON_PRESS}>Checkbox Selected</Text> : null}`

If you have any further questions, or if this is confusing, feel free to reach out! It's very simple and might be confusing over text, so a call might help if you're confused

apps/fluent-tester/src/TestComponents/Switch/consts.ts
```diff
@@ -0,0 +1,9 @@
+export const HOMEPAGE_SWITCH_BUTTON = 'Homepage_Switch_Button';
+export const SWITCH_TESTPAGE = 'Switch_TestPage';
+
+/* Text Switch 1 */
+export const SWITCH_TEST_COMPONENT = 'Switch_Test_Component';
+export const SWITCH_ACCESSIBILITY_LABEL = 'E2E testing Switch accessibility label';
+
+export const SWITCH_TOGGLE_ON = 'Switch_Toggle_On';
```
samuelfreiberg, line 8: These two constants should only be one:
`SWITCH_ON_PRESS`

Please see comment above

yarn.lock
```diff
@@ -1439,6 +1439,24 @@
   resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-windows/-/design-tokens-windows-0.13.0.tgz#1d8610634920bd5decaca669cae8183c0713d2ef"
   integrity sha512-o9ir+tpie+Fm9zO+nCbJLBfzVjUhugRWRsPS8n6wBaujC1qyeUYjhoGg6ejXzKc0nVCkIsFmRRRgC0brfrPIzw==
 
+"@fluentui-react-native/framework@0.7.30":
```
rurikoaraki, line 1442: This shouldn't be in our yarn.lock. I think it's likely that your package is referring to an older version of the framework package.

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+  const onClickWithFocus = useOnPressWithFocus(focusRef, toggleCallback);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onClickWithFocus });
+  const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
+  const currentCheckedState = checked ?? checkedState;
+
+  const accessibilityState: AccessibilityState = {
+    checked: props.accessibilityState?.checked ?? currentCheckedState,
+  };
+
+  const accessibilityActionsProp: AccessibilityActionInfo[] = [{ name: 'Toggle' }];
```
ksiler, line 49: I'd rewrite this like this to account for cases where the developer may want to add to the default accessibilityActions property since you call the user defined prop `onAccessibilityActions` on line 58

```

const accessibilityActionsProp = accessibilityActions
    ? [...defaultAccessibilityActions, ...accessibilityActions]
    : defaultAccessibilityActions;

File: apps/fluent-tester/src/E2E/Switch/specs/SwitchSpec.win32.ts [hunk 21]
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Click on component button to navigate to test page */
+    NavigateAppPage.clickAndGoToSwitchPage();
+    SwitchPageObject.waitForPageDisplayed(PAGE_TIMEOUT);
+
+    expect(SwitchPageObject.isPageLoaded()).toBeTruthy(SwitchPageObject.ERRORMESSAGE_PAGELOAD);
+    expect(SwitchPageObject.didAssertPopup()).toBeFalsy(SwitchPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped up
+  });
+});
+
```
samuelfreiberg, line 25: Please add Accessibility Testing.

See Checkbox.win32.spec.ts as an example.

Should test:
Accessibility Label
Accessibility Role

packages/components/Switch/src/index.ts
```diff
@@ -0,0 +1 @@
+export { Switch } from './Switch';
```
rurikoaraki, line 1: You probably want to export the types for the props/tokens/slots and name of the Switch as well. It will also be useful to export the useSwitch function for clients to access.

rurikoaraki:
nit: generally we add in progress components to experimental so that we know they're not ship-ready.

apps/fluent-tester/src/TestComponents/Switch/consts.ts
```diff
@@ -0,0 +1,9 @@
+export const HOMEPAGE_SWITCH_BUTTON = 'Homepage_Switch_Button';
+export const SWITCH_TESTPAGE = 'Switch_TestPage';
+
+/* Text Switch 1 */
+export const SWITCH_TEST_COMPONENT = 'Switch_Test_Component';
+export const SWITCH_ACCESSIBILITY_LABEL = 'E2E testing Switch accessibility label';
```
samuelfreiberg, line 6: Please add tests for accessibility label

apps/fluent-tester/src/E2E/Switch/specs/SwitchSpec.win32.ts
```diff
@@ -0,0 +1,62 @@
...
+describe('Switch Functional Testing', () => {
+  /* Scrolls and waits for the Switch to be visible on the Test Page */
+  beforeEach(() => {
+    SwitchPageObject.scrollToTestElement();
+    SwitchPageObject.waitForPrimaryElementDisplayed(PAGE_TIMEOUT);
+  });
+
+  it('Validate onChange() callback was fired -> Click', () => {
```
samuelfreiberg, line 33: See comment below about onChange().

Saadnajmi:
> nit: generally we add in progress components to experimental so that we know they're not ship-ready.

** nervously looks at Notification 😅 ** (my bad)

packages/components/Switch/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/components/Button"
+  },
+  "dependencies": {
+    "@fluentui-react-native/adapters": ">=0.8.5 <1.0.0",
+    "@fluentui-react-native/experimental-text": ">=0.9.0 <1.0.0",
```
rurikoaraki, line 30: I think lena recently migrated the Text component to the text package.

packages/components/Switch/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "lint": "fluentui-scripts eslint",
+    "test": "fluentui-scripts jest",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "prettier": "fluentui-scripts prettier",
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/components/Button"
```
rurikoaraki, line 26: Should be updated

apps/fluent-tester/src/E2E/Switch/specs/SwitchSpec.win32.ts
```diff
@@ -0,0 +1,62 @@
...
+    /* Click on component button to navigate to test page */
+    NavigateAppPage.clickAndGoToSwitchPage();
+    SwitchPageObject.waitForPageDisplayed(PAGE_TIMEOUT);
+
+    expect(SwitchPageObject.isPageLoaded()).toBeTruthy(SwitchPageObject.ERRORMESSAGE_PAGELOAD);
+    expect(SwitchPageObject.didAssertPopup()).toBeFalsy(SwitchPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped up
+  });
+});
+
+describe('Switch Functional Testing', () => {
```
samuelfreiberg, line 26: Please see the checkbox.spec.win32.ts. Those tests query for the Checkbox toggle state AND check that the onChange() was called in the same test methods. We should follow that pattern here.

samuelfreiberg (changes requested):

rurikoaraki:
> > nit: generally we add in progress components to experimental so that we know they're not ship-ready.
> 
> ** nervously looks at Notification 😅 ** (my bad)

I'm open to revisiting this pattern/only doing it for components that have an old version. I feel like it's easier for people to notice "oh this is in experimental" than to find "this is in beta" in documentation somewhere, but it does also create more work so it's not like a strong opinion

rurikoaraki:
Suggestion: This is a lot of code to review and be at check in quality at once. I'd recommend breaking down changes into smaller chunks in the future. There's several advantages to that:

1. It will take less time to get through review, because there's less code to get to quality and less code to review.
2. Chances of something stopping you from checking in are smaller because the code "surface area" is smaller. For example, you're less likely to run into conflicts.
3. Reviewers will leave better quality feedback, because it takes less brain cycles to understand the change and less code to review in one go. This means more brain cycles go toward thinking more deeply about the implications of a change. Also, long reviews mean people will put them off until they have time to sit down and look through it.
4. If a change needs to be reverted, the revert is less impactful and you have less code to update before trying to commit it again.

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <Switch defaultChecked={true} label={'Disabled Default Checked True'} disabled />
+      <Switch defaultChecked={false} label={'Disabled Default Checked False'} disabled />
+    </View>
+  );
+};
+
+const OnChangeUsage: React.FunctionComponent = () => {
+  const [displaySquare, setDisplaySquare] = React.useState(true);
+  const [evenDisplaySquare, setEvenDisplaySquare] = React.useState(true);
+  const [count, setCount] = React.useState(0);
```
rurikoaraki, line 31: nit: I think the count can be a plain var and not state, since it doesn't impact rendering of any components. It's just used for logic

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const defaultToggleSquare = (checked: boolean) => {
+    setDisplaySquare(checked);
+  };
+
+  const toggleSquareEven = (checked: boolean) => {
+    if (count % 2 == 1) {
```
rurikoaraki, line 38: nit: the even-ness could probably be represented with a boolean instead of a count.

```
let isEven = false;

if (checked) {
    isEven = !isEven;
    setEvenDisplaySquare(isEven);    
} else {
    setEvenDisplaySquare(false);    
}
```

packages/components/Switch/src/Switch.styling.ts
```diff
@@ -0,0 +1,84 @@
...
+  tokens: [defaultSwitchTokens, switchName],
+  states: switchStates,
+  slotProps: {
+    root: buildProps(
+      (tokens: SwitchTokens) => ({
+        style: {
+          alignItems: 'center',
+          flexDirection: tokens.flexDirection,
+          alignSelf: 'flex-start',
+          minHeight: 28,
```
rurikoaraki, line 28: Some of these seem like things that we would want to let clients customize, like minWidth, or the borderRadius of the thumb, or padding. Is that something you were thinking of changing later?

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+      ref: useViewCommandFocus(componentRef),
+      ...onKeyUpProps,
+      checked: currentCheckedState,
+      label: label,
+      labelPosition: labelPosition ?? 'after',
+    },
+    state: {
+      ...pressable.state,
+      toggleOn: currentCheckedState == true,
+      toggleOff: currentCheckedState == false,
```
rurikoaraki, line 83: Why do we need both? Can we have a "toggled" prop and check for its value in the lookup function instead?

packages/components/Switch/src/useSwitch.ts
rurikoaraki, line 49: The `[{ name: 'Toggle' }]` object should be created outside of the component. Otherwise it will be recreated on every render, causing unnecessary rerenders.

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+    },
+    [toggleCallback, onAccessibilityAction],
+  );
+
+  return {
+    props: {
+      ...pressable.props,
+      accessible: true,
+      accessibilityRole: accessibilityRole ?? 'switch',
+      onAccessibilityTap: onAccessibilityTap ?? onChange,
```
rurikoaraki, line 68: Should onAccessibilityTap be hooked up if we're handling accessibilityActions?

packages/components/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,94 @@
...
+  const isDisabled = !!disabled;
+  const initialCheckedState = !!(checked ?? defaultChecked);
+  const [checkedState, setCheckedState] = React.useState(initialCheckedState);
+  const focusRef = isDisabled ? null : componentRef;
+
+  const toggleCallback = React.useCallback(
+    (e: InteractionEvent) => {
+      const newCheckedState = checked !== undefined ? checked : !checkedState;
+      onChange && onChange(newCheckedState, e);
+      LayoutAnimation.configureNext(LayoutAnimation.create(200, LayoutAnimation.Types.linear, LayoutAnimation.Properties.scaleX));
```
rurikoaraki, line 34: Should this still run if the Switch's checked value is controlled?

rurikoaraki: And do animations need to play if the Switch's checked value changes via the checked prop?

packages/components/Switch/src/useSwitch.ts
nabil-k: They are mutually exclusive. I can add a check that warns users.

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
nabil-k: I actually removed this example and replaced it with something else. I was trying to show how you could control the checked state of the Switch. But I realize a much simpler and more understandable example would be to just have 2 buttons that change the Switch's state. So that's what I've added to replace this instead.

apps/fluent-tester/src/TestComponents/Switch/Switch.tsx
nabil-k: Removed this example.

packages/components/Switch/src/useSwitch.ts
nabil-k: Yeah I've removed accessibility changes from this PR and moved it to another branch which I will do a PR for later.

packages/components/Switch/src/useSwitch.ts
nabil-k: So `currentCheckedState` acts as the final truth to what the checked state of the control is. For this control, if the user sets the `checked` prop, we will obey that value no matter what. However if the user does not have that defined, then we will instead use the `checkedState` value. This `checkedState` is the checked state that the control manages for itself (for example when the user decides to use `defaultChecked` because they don't want to manage to Switch's state themselves).

packages/components/Switch/src/useSwitch.ts
nabil-k: I'll make this change in a accessibility PR.

packages/components/Switch/src/useSwitch.ts
ksiler: Good question/callout. `onAccessibilityTap` should be used for buttons that don't maintain state (at least on win32). Since this control does maintain state, we should use the 'Toggle' command and not hook up `onAccessibilityTap`.

apps/fluent-tester/src/TestComponents/Switch/E2ESwitchTest.tsx
nabil-k: I've implemented testing onChange like you've described. Thanks for writing out all this info, it was really helpful! :)

packages/components/Switch/src/useSwitch.ts
nabil-k: Ahh I see, thanks for pointing that and the explanation 👍

packages/components/Switch/src/useSwitch.ts
nabil-k: I think it would be good if it still ran even though the value is controlled because it can allow developers to still detect if the Switch has been attempted to be toggled. This can be useful if they want a warning to pop up saying that the Switch cannot be used at the moment or something like that.

nabil-k: Honestly I'm not 100% confident on whether or not animations need to play in that situation. However, preferably I would like them too. Although at the moment they do not.

packages/components/Switch/src/useSwitch.ts
nabil-k: We don't, I've added a toggled state and made the lookup function handle the toggleOn and toggleOff tokens. Thanks for helping me clean that up.

packages/components/Switch/src/Switch.styling.ts
nabil-k: Yeah at the moment I wasn't focusing on having it customizable. I just wanted the bones of it down.

apps/fluent-tester/src/E2E/Switch/specs/SwitchSpec.win32.ts
nabil-k: After this PR I'm going to make a accessibility PR where I'll add the accessibility tests a long with it in an attempt to help declutter this PR. 

Although, in order to implement the onChange E2E I had to add some accessibility props but those aren't fully implemented. They were just added simply for the onChange E2E tests. I can add a Accessibility Role test though since that is pretty much set. Would that be fine?

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  if (defaultChecked !== undefined && checked !== undefined) {
+    console.warn('The props defaultChecked and checked are mutually exclusive. Use only one of the props, do not use both.');
+  }
+
+  const toggleCallback = React.useCallback(
+    (e: InteractionEvent) => {
+      const newCheckedState = checked !== undefined ? checked : !checkedState;
+      onChange && onChange(newCheckedState, e);
+      LayoutAnimation.configureNext(LayoutAnimation.create(200, LayoutAnimation.Types.linear, LayoutAnimation.Properties.scaleX));
```
PPatBoyd, line 36: per the controlled vs uncontrolled, we should run the animation off of the state change and not the user interaction

apps/fluent-tester/src/E2E/Switch/specs/SwitchSpec.win32.ts
samuelfreiberg: Sounds good to me as long as we make sure the other accessibility testing gets in in a subsequent PR :)

packages/experimental/Switch/src/useSwitch.ts
nabil-k: Ahh yeah, thanks! I've made it animate off of state change instead.

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  const toggleCallback = React.useCallback(
+    (e: InteractionEvent) => {
+      const newCheckedState = checked !== undefined ? checked : !checkedState;
+      onChange && onChange(newCheckedState, e);
+      setCheckedState(newCheckedState);
+    },
+    [onChange, setCheckedState, checkedState, checked],
+  );
```
ksiler, line 40: Instead of using your own callback for `toggleCallback`, could you use `useAsToggleWithEvent` under interactive-hooks? This would help maintain the "real" checked state like you're tracking with `currentCheckedState` but without the need for the extra variable.

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,72 @@
...
+    ...rest
+  } = props;
+
+  const isDisabled = !!disabled;
+  const initialCheckedState = !!(checked ?? defaultChecked);
+  const [checkedState, setCheckedState] = React.useState(initialCheckedState);
+  const [prevCheckedState, setPrevCheckedState] = React.useState(initialCheckedState);
+  const focusRef = isDisabled ? null : componentRef;
+
+  if (defaultChecked !== undefined && checked !== undefined) {
```
rurikoaraki, line 29: We should gate this to only fire while in dev mode. console.logs cause memory leaks in react so we should avoid them in production.

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  const onClickWithFocus = useOnPressWithFocus(focusRef, toggleCallback);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onClickWithFocus });
+  const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
+  const currentCheckedState = checked ?? checkedState;
+
+  // Triggers animation only when the checked state changes
+  if (prevCheckedState !== currentCheckedState) {
+    LayoutAnimation.configureNext(LayoutAnimation.create(200, LayoutAnimation.Types.linear, LayoutAnimation.Properties.scaleX));
```
rurikoaraki, line 49: accessibility: we should make sure the animation respects any "turn animation off" settings from the OS.

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -0,0 +1,72 @@
...
+    [onChange, setCheckedState, checkedState, checked],
+  );
+
+  const onClickWithFocus = useOnPressWithFocus(focusRef, toggleCallback);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onClickWithFocus });
+  const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
+  const currentCheckedState = checked ?? checkedState;
+
+  // Triggers animation only when the checked state changes
+  if (prevCheckedState !== currentCheckedState) {
```
rurikoaraki, line 48: You can use React.useEffect for this:

React.useEffect(() => {
    setPrevCheckedState(currentCheckedState);
}, [currentCheckedState]); // only run when currentCheckedState changes

rurikoaraki: This means you can remove prevCheckedState too

packages/experimental/Switch/src/useSwitch.ts
rurikoaraki, line 49: We should also see if we need to cancel the animation if the component is unmounted.

packages/experimental/Switch/src/useSwitch.ts
nabil-k: I've made it use `useAsToggleWithEvent `, it cleaned up things well thanks!

ksiler: Glad it worked!

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -35,7 +35,7 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
   const onKeyUpProps = useKeyProps(toggleCallback, ' ', 'Enter');
 
   // Triggers animation only when the checked state changes
-  if (prevCheckedState !== checkedState) {
+  if (prevCheckedState !== checkedState && AccessibilityInfo.isReduceMotionEnabled) {
```
ksiler, line 38: Have you tested this? Does it work? Just wondering if there's any native work required for win32 to make `isReduceMotionEnabled` return the right value according to the OS settings.

nabil-k: Yeah I tried it out and it worked. I went toggled this "Show animations in windows" setting off.
image

packages/experimental/Switch/src/useSwitch.ts
nabil-k: using React.useLayoutEffect now

packages/experimental/Switch/src/useSwitch.ts
ksiler: Okay, good to know!

packages/experimental/Switch/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@fluentui-react-native/experimental-switch",
```
rurikoaraki, line 2: We can just use "switch" if that package name isn't used yet. Saves some headache of changing the name later

rurikoaraki:
Code is looking much cleaner now. The Switch looks great in the screenshots!

rurikoaraki (approved):

ksiler (approved):

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
```diff
@@ -0,0 +1,65 @@
...
+
+describe('Switch Accessibility Testing', () => {
+  /* Scrolls and waits for the Switch to be visible on the Test Page */
+  beforeEach(() => {
+    SwitchPageObject.scrollToTestElement();
+    SwitchPageObject.waitForPrimaryElementDisplayed(PAGE_TIMEOUT);
+  });
+
+  it('Switch - Validate accessibilityRole is correct', () => {
+    expect(SwitchPageObject.getAccessibilityRole()).toEqual(BUTTON_A11Y_ROLE);
```
samuelfreiberg, line 34: Is this correct? The switch is a Button? In the spec, it says accessibilityRole = 'switch'

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(SwitchPageObject.didAssertPopup()).toBeFalsy(SwitchPageObject.ERRORMESSAGE_ASSERT);
+  });
+});
+
+describe('Switch Functional Testing', () => {
```
samuelfreiberg, line 39: What about keyboard interaction? Does the switch toggle on space/enter? That's easily added here too (see Checkbox again)

nabil-k: I had that in a accessibility branch but that should have been included in this one. Thanks for catching that!

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
nabil-k: Yeah there isn't a Switch control type identifier. You can see that [here](https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-controltype-ids). It gets resolved as a button in win32 I believe.

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  it('Click the "Enter" on a Switch and verify it toggles', () => {
+    /* Validate the Switch is initially toggled OFF */
+    expect(SwitchPageObject.isSwitchChecked()).toBeFalsy();
+
+    /* Presses the "Enter" to select the Switch */
+    SwitchPageObject.sendKey(SwitchComponentSelector.PrimaryComponent, Keys.Enter);
+    SwitchPageObject.waitForSwitchChecked(PAGE_TIMEOUT);
+
```
samuelfreiberg, line 73: `expect(SwitchPageObject.didOnChangeCallbackFire()).toBeTruthy();`

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  it('Click the "Spacebar" on a Switch and verify it toggles', () => {
+    /* Validate the Switch is initially toggled OFF */
+    expect(SwitchPageObject.isSwitchChecked()).toBeFalsy();
+
+    /* Presses the "space bar" to select the Switch */
+    SwitchPageObject.sendKey(SwitchComponentSelector.PrimaryComponent, Keys.Spacebar);
+    SwitchPageObject.waitForSwitchChecked(PAGE_TIMEOUT);
+
```
samuelfreiberg, line 92: `expect(SwitchPageObject.didOnChangeCallbackFire()).toBeTruthy();`

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
nabil-k: Added that in

apps/fluent-tester/src/E2E/Switch/specs/Switch.spec.win.ts
nabil-k: Added that in

samuelfreiberg (approved):

PR #1931, opened 2022-07-29
[Button] Prefer user key events over those handled by Fluent

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Currently, on Windows, the `Button` component handles keyUp events (keyDown for macOS). What that means is someone can't have a keyUp (or keyDown) handler of their own to do "special" things on keypresses on buttons (e.g. perhaps implement a split button).

Previously (prior to PR https://github.com/microsoft/fluentui-react-native/pull/1373), `Button` always used keyUp, so button users could at least use keyDown to do things, but now depending on the target platform, a button user would have to handle the slot NOT used by Fluent.

I investigated the possibility of creating a composite handler (such that for user requested keys we always pass back to the user, and if not specified, then we pass back the handling to Fluent). It looks like it will need substantial refactoring because the event types for key handlers in useKeyProps does not match that exposed by Win32. Further, the approach to specifying valid keys (the name of the prop, as well the specifics like modifiers) is different for the Mac.

Therefore, in this change, I am instead suggesting the *other* approach I previously discussed with Saad -- if a button user tries to override a particular key handler, the user is also responsible for providing the default behaviors (like Enter\Space to invoke the button event).

For client usage correctness (and convenience), also expose a flag to indicate platform behavior with respect to keyUp or keyDown.

### Verification

Manually, using Fluent tester.

* Regular button use case (no key handler): Enter\Space works as expected
* New button added to handle key events: ONLY handled key events pass through; since Enter\Space isn't one of them, this lets us prevent default action (of course, handling those two keys manually in the custom handler would restore that behavior).

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/utils/interactive-hooks/src/useKeyProps.ts
+++ b/packages/utils/interactive-hooks/src/useKeyProps.ts
@@ -55,3 +55,6 @@ export const useKeyDownProps = noOp2;
  * @returns KeyPressProps: An object containing the correct platform specific props to  handle key press
  */
 export const useKeyProps = noOp2;
+
+/** Exposes the behavior of useKeyProps for the current platform as a boolean */
+export const preferKeyDownForKeyEvents = false;
--- a/apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
@@ -11,14 +11,37 @@ import {
   BUTTON_NO_A11Y_LABEL_COMPONENT,
   BUTTON_ACCESSIBILITY_LABEL,
   BUTTON_TEST_COMPONENT_LABEL,
+  BUTTON_ON_KEY,
+  BUTTON_PRESS_TEST_COMPONENT,
+  BUTTON_PRESS_TEST_COMPONENT_LABEL,
 } from './consts';
+import { IViewWin32Props } from '@office-iss/react-native-win32';
 
 export const E2EButtonExperimentalTest: React.FunctionComponent = () => {
   const [buttonPressed, setButtonPressed] = React.useState(false);
+  const [keyDetected, setKeyDetected] = React.useState('');
 
   const onClick = React.useCallback(() => {
     setButtonPressed(!buttonPressed);
+    setKeyDetected('');
   }, [buttonPressed]);
+
+  const keyPressProps: Omit<IViewWin32Props, 'accessibilityRole' | 'onBlur' | 'onFocus'> = {
+    keyDownEvents: [{ key: 'a' }],
+    onKeyDown: (args) => {
+      if (args.nativeEvent.key === 'a') {
+        setKeyDetected('a (down)');
+        args.stopPropagation();
+      }
+    },
+    keyUpEvents: [{ key: 'b' }],
+    onKeyUp: (args) => {
+      if (args.nativeEvent.key === 'b') {
+        setKeyDetected('b (up)');
+        args.stopPropagation();
+      }
+    },
+  };
 
   return (
     <View>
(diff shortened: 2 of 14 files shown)


rurikoaraki:
For my own understanding, the aim here is to make it so that if a custom key handler is passed in, then it ignores the built in functionality completely?

nakambo:
> For my own understanding, the aim here is to make it so that if a custom key handler is passed in, then it ignores the built in functionality completely?

Indeed, this is what Saad and I previously discussed. We liked this option because it was simpler, since it didn't require reconciling two sets of event listeners (Fluent and the actual component user\client) and possibly notifying both of them\not notifying keys to the listener that didn't request that key.

Saadnajmi:
@Ruriko for more context, a nicer solution would be to have all RN desktop platforms respect `e.preventDefault()` so that client code could choose programmatically to suppress the default behavior. Doing so requires a bunch of native changes to React Native macOS so it's harder. Thanks for making the fix @nakambo !

packages/utils/interactive-hooks/src/useKeyProps.ts
```diff
@@ -55,3 +55,6 @@ export const useKeyDownProps = noOp2;
  * @returns KeyPressProps: An object containing the correct platform specific props to  handle key press
  */
 export const useKeyProps = noOp2;
+
+/** Exposes the behavior of useKeyProps for the current platform as a boolean */
+export const preferKeyDownForKeyEvents = false;
```
Saadnajmi, line 60: Where do you actually read and use this? In client code?

nakambo: That's the idea -- it could be used in the client code. I plan to handle my key + Enter + Space in the handler indicated by this boolean, but in theory, someone could do what I was previously doing, viz. use this flag to use the OTHER slot that isn't used by Fluent.

nakambo:
> For my own understanding, the aim here is to make it so that if a custom key handler is passed in, then it ignores the built in functionality completely?

@rurikoaraki, so this is indeed a breaking change to (already broken) clients -- because if they don't already handle Enter\Space for example, they'll need to now (previously they were broken because their custom keys wouldn't be handled, now they'd be broken for a different reason).

I'm also okay with just exposing the boolean flag (just [this](https://github.com/microsoft/fluentui-react-native/pull/1931/commits/3f1fcb97191d42b1df0fe84ced503bbb723a6875) commit).

rurikoaraki:
@nakambo okay let me think about it a little bit, i'll get back to the PR later today. If we do decide to add the full functionality we should update the documentation to say that this is the behavior of specifying custom key handlers.

Saadnajmi (approved):
Good from my side, but I would wait for final approval from @rurikoaraki.

For what it's worth, my thoughts overall are:

1) If we can get away with just exporting the one flag and not changing behavior, I prefer that. 
2) If we do change behavior, I would document it (probably Buttons' SPEC.md file). 
3) I'd normally push for a more flexible API, but given the priority of this bug / the level of refactoring needed to do that / timeline, I don't think that's feasible at this point.

rurikoaraki:
@nakambo Okay given that preventDefault doesn't work properly this change makes sense to me. Could add a note in documentation about this behavior?

Follow up on our side would be to make this behavior consistent across our components.

nakambo:
> @nakambo Okay given that preventDefault doesn't work properly this change makes sense to me. Could you add a note in documentation about this behavior?
> 
> Follow up on our side would be to make this behavior consistent across our components.

I have added a note -- if that isn't what you were expecting to see, let me know specifically which file\what note you'd like to see.

rurikoaraki:
> > @nakambo Okay given that preventDefault doesn't work properly this change makes sense to me. Could you add a note in documentation about this behavior?
> > Follow up on our side would be to make this behavior consistent across our components.
> 
> I have added a note -- if that isn't what you were expecting to see, let me know specifically which file\what note you'd like to see.

Awesome, thanks!

rurikoaraki (approved):

Ruriko:
@Saadnajmi next time ping the correct person

PR #1939, opened 2022-08-01
RadioGroup Prop Alignment & Update Tester

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
The control is not fully implemented yet - this is an initial PR to get groundwork checked in for upcoming changes. Started work in the experimental folder, copying over existing component implementation. Updated all references of `RadioButton` to `Radio`. Renamed interfaces to follow existing naming convention. Renamed `Slots.content` to `Slots.label`. Updated tester page to use experimental `Radio` and `RadioGroup`. Updated prop names in `Radio` and `RadioGroup` to align with web props. 

Need to update E2E testing and design tokens, which will be in next PR.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/RadioGroup/src/Radio.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { RadioGroupContext } from './RadioGroup';
+
+export const Radio = compose<RadioType>({
+  displayName: radioName,
+
+  usePrepareProps: (userProps: RadioProps, useStyling: IUseComposeStyling<RadioType>) => {
+    const defaultComponentRef = React.useRef(null);
+    const {
+      label,
+      value,
```
ValentinaKozlova, line 23: Why are you using old framework?

packages/experimental/RadioGroup/src/Radio.types.ts
```diff
@@ -0,0 +1,70 @@
...
+import { FontTokens, IForegroundColorTokens, IBackgroundColorTokens, IBorderTokens } from '@fluentui-react-native/tokens';
+import { IPressableProps } from '@fluentui-react-native/pressable';
+
+export const radioName = 'Radio';
+
+export interface RadioProps extends IPressableProps {
+  /**
+   * The text string for the option
+   */
+  label: string;
```
ValentinaKozlova, line 16: Is label a required prop?

ValentinaKozlova:
Currently we use `@fluentui-react-native/framework` for all the new components. You might want to use a component-generator to create a component with a new framework

apps/fluent-tester/src/TestComponents/RadioGroup/RadioGroupTest.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { View } from 'react-native';
-import { RadioButton, RadioGroup, Separator } from '@fluentui/react-native';
+import { Separator } from '@fluentui/react-native';
```
samuelfreiberg, line 3: I believe we should make a new Experimental RadioGroup test page (I think that's been the pattern as we develop updated Experimental controls).

The reason being it's going to take some time to get:

1. The new control fully developed
2. Partners consuming the new control

Also, we want our E2E testing to continue to test our current RadioGroup control. Partners are still using it and we need to validate we don't regress it.

packages/experimental/RadioGroup/macos/FRNRadioButtonViewManager.m
```diff
@@ -0,0 +1,20 @@
+#import <React/RCTComponent.h>
```
samuelfreiberg, line 1: Are you responsible for the macOS side to this as well? I'd assume not - hopefully we can loop Harrie's team into this? I'm wary you should even be the one copying it over

samuelfreiberg (changes requested):

packages/experimental/RadioGroup/src/Radio.types.ts
```diff
@@ -0,0 +1,70 @@
...
+   * radio group.
+   */
+  accessibilityPositionInSet?: number;
+
+  /**
+   * Defines the number of radio buttons in the group for accessibility purposes.It's recommended to set this value if radio
+   * buttons are not direct children of radio group. This value is auto-generated if radio buttons are direct children of
+   * radio group.
+   */
+  accessibilitySetSize?: number;
```
ksiler, line 45: I don't think you need to define `accessibilityLabel`, `accessibilityPositionInSet`, and `accessibilitySetSize` props here. They should be included as part of View props which you can set on the root.

apps/fluent-tester/src/TestComponents/RadioGroup/RadioGroupTest.tsx
gulnazsayed: Makes sense, creating a new RadioGroupExperimental test page.

packages/experimental/RadioGroup/macos/FRNRadioButtonViewManager.m
Saadnajmi: The native module assumes a specific API (Like the text being in the prop "content", see line 6. I wouldn't copy this over at all unless we're really sure this works. 

For other components that had a macOS native module, when win32 went to V1, we kept the native module in the "V0" implementation.

packages/experimental/RadioGroup/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@fluentui-react-native/experimental-radio-group",
+  "version": "0.15.0",
```
rurikoaraki, line 3: You should start with 0.1.0 or 0.0.1

rurikoaraki:
The changelog files are automatically generated so they should not be included in a new folder.

packages/experimental/RadioGroup/src/RadioGroup.tsx
```diff
@@ -30,14 +30,14 @@ export const RadioGroupContext = React.createContext<IRadioGroupContext>({
...
   displayName: radioGroupName,
 
-  usePrepareProps: (userProps: IRadioGroupProps, useStyling: IUseComposeStyling<IRadioGroupType>) => {
-    const { label, ariaLabel, accessibilityLabel, selectedKey, defaultSelectedKey, ...rest } = userProps;
+  usePrepareProps: (userProps: RadioGroupProps, useStyling: IUseComposeStyling<RadioGroupType>) => {
+    const { label, accessibilityLabel, value, defaultValue, ...rest } = userProps;
 
     // This hook updates the Selected Button and calls the customer's onClick function. This gets called after a button is pressed.
-    const data = useSelectedKey(selectedKey || defaultSelectedKey || null, userProps.onChange);
+    const data = useSelectedKey(value || defaultValue || null, userProps.onChange);
```
rurikoaraki, line 40: SHould probably udpate the hook to use the new name, too

packages/experimental/RadioGroup/src/RadioGroup.types.ts
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+import { ITextProps } from '@fluentui-react-native/text';
+import type { IViewWin32Props } from '@office-iss/react-native-win32';
```
rurikoaraki, line 3: We should pull IViewProps from @fluentui-react-native/adapters instead. react-native-win32 is win32 specific.

packages/experimental/RadioGroup/src/RadioGroup.types.ts
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+import { ITextProps } from '@fluentui-react-native/text';
```
rurikoaraki, line 2: Should be updated to use the new TextProps, but can be a separate change.

PR #1946, opened 2022-08-02
Restore ability to create un-focusable buttons

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Previously (with the old deprecated buttons) it was possible to create buttons that aren't disabled but still are unfocusable. (this may sound like a bizarre use case, but is handy for creating things like split buttons where the individual component buttons are not focusable).

We seem to have lost this ability with the new Button -- which will override focusable even if passed by the component user. Let's not -- if that is specified, we respect it (and allow people to create un-focusable enabled buttons, or focusable disabled buttons as before).

### Verification

Manually tested the `focusable={false}` prop in the Win32 tester app. Existing buttons in the E2E section that don't have this property are unaffected.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
@@ -14,6 +14,8 @@ import {
   BUTTON_ON_KEY,
   BUTTON_PRESS_TEST_COMPONENT,
   BUTTON_PRESS_TEST_COMPONENT_LABEL,
+  BUTTON_FOCUSABLE_TEST_COMPONENT,
+  BUTTON_FOCUSABLE_TEST_COMPONENT_LABEL,
 } from './consts';
 import { IViewWin32Props } from '@office-iss/react-native-win32';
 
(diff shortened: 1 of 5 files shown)


rurikoaraki:
@nakambo Thanks for catching this and fixing, appreciate it!

apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
```diff
@@ -60,6 +62,14 @@ export const E2EButtonExperimentalTest: React.FunctionComponent = () => {
         >
           Press &quot;a&quot; or &quot;b&quot; after focusing this button
         </Button>
+        <Button
+          testID={BUTTON_FOCUSABLE_TEST_COMPONENT}
+          onClick={onClick}
+          accessibilityLabel={BUTTON_FOCUSABLE_TEST_COMPONENT_LABEL}
+          focusable={false}
+        >
+          This button isn&apos;t focusable (but isn&apos;t disabled or indeterminate either)
```
rurikoaraki, line 71: Is the "&apos;" required? I don't think I've seen it before

rurikoaraki: I guess it's right up above in a different button so uh, sure

rurikoaraki (approved):

apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
nakambo: you mean `isn't` directly without escaping it? VSCode seemed to complain about it, I didn't actually try building with that

nakambo: that's so strange, so VSCode is picking up some eslint rule from somewhere, but `npm run build` does not complain.

image

image

I'll guess I'll leave it escaped?

rurikoaraki:
Filed an issue to improve the other controls, too. Unfortunately not high on the priority list for me, so it might take a little bit to get to it.

apps/fluent-tester/src/TestComponents/Button/E2EButtonTest.tsx
rurikoaraki: Yeah that's fine, I was just curious.

PR #1953, opened 2022-08-03
Win32 Switch Control - Accessibility

nabil-k:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added accessibility features to the Switch control.

### Verification

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -64,3 +89,11 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
     },
   };
 };
+
+const getAccessibilityState = memoize(getAccessibilityStateWorker);
+function getAccessibilityStateWorker(checked: boolean, disabled: boolean, accessibilityState?: AccessibilityState) {
+  if (accessibilityState) {
+    return { checked, disabled, ...accessibilityState };
```
ksiler, line 96: I think this needs to be specified as `checked: checked, disabled: disabled, ...` (same as line 98).

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -48,9 +70,12 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
     props: {
       ...pressable.props,
       accessible: true,
+      accessibilityLabel: accessibilityLabel ?? label,
       accessibilityRole: accessibilityRole ?? 'switch',
-      accessibilityState: { disabled: disabled, checked: checkedState }, // Needed for E2E Testing to detect toggle state, will provide a better implementation in a Accessibility PR
-      accessibilityActions: defaultAccessibilityActions, // Needed for E2E Testing to detect toggle state, will provide a better implementation in a Accessibility PR
+      accessibilityHint: accessibilityHint,
```
lenahong, line 75: Is there a reason for why you don't pass in rest props here?
For props like accessibilityHint and label that aren't overriden in here, it may be simpler to let them be passed into `...rest` set of props, and in here instead of reassigning them, pass in `...rest` as you did with pressable.props.

packages/experimental/Switch/src/useSwitch.ts
nabil-k: I wrote it like that because that's how I saw other files like `useCheckbox.ts` do it. It looks like it's just a syntatic shortcut for when the variable's name and the key's name are the same. Here's a [article](https://attacomsian.com/blog/javascript-object-property-shorthand#:~:text=The%20object%20literal%20property%20value,make%20the%20object%20initialization%20shorter.&text=The%20property%20value%20shorthand%20syntax,value%20as%20a%20property%20value.) with an example. I can still change it if you'd like though.

packages/experimental/Switch/src/useSwitch.ts
nabil-k: I think I was trying to filter out props that were not part of the control. But I think with what you pointed out there are also other issues with doing that.  Thanks!

packages/experimental/Switch/src/useSwitch.ts
ValentinaKozlova: Yeah, I'm also interested why we can't use shorten version. Is it some convention that we need to write `checked: checked`?

ksiler: Oh interesting, I didn't realize that worked. That's totally fine then. Thanks for teaching me something new 🙂

ksiler (approved):

PR #1969, opened 2022-08-08
Radio New Framework & Design Tokens

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Implemented Radio component with new framework and integrated design tokens. Updated tester page to use new Radio components. Created folder for E2E testing, will have E2E testing code in next PR to make the review more manageable. 

### Verification

| image | 
image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -57,21 +57,25 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
-      },
-    })),
+    label: buildProps(
+      (tokens: RadioTokens) => ({
+        variant: tokens.variant,
+        style: {
+          marginTop: 2,
+          borderStyle: 'solid',
+          borderColor: tokens.textBorderColor,
+          borderWidth: 2,
```
ValentinaKozlova, line 74: Should borderWidth be in Tokens? Because currently borderWidth is not customizable

gulnazsayed: Thanks for pointing that out! Added to tokens. It wasn't highlighted in the web spec, so I didn't realize it was customizable.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,81 @@
...
+import { UseStylingOptions, buildProps } from '@fluentui-react-native/framework';
+import { defaultRadioTokens } from './RadioTokens';
+
+const radioSize = 20;
+const radioRadius = radioSize / 2;
+
+const radioInnerCircleSize = 10;
+const radioInnerCircleRadius = radioInnerCircleSize / 2;
+
+export const radioSelectActionLabel = 'Select a RadioButton';
```
rurikoaraki, line 11: We cannot have any hardcoded text as they would not get localized.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const radioStates: (keyof RadioTokens)[] = ['focused', 'hovered', 'pressed', 'selected', 'disabled'];
+
+export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, RadioTokens> = {
+  tokens: [defaultRadioTokens, radioName],
+  states: radioStates,
+  slotProps: {
+    root: buildProps(() => ({
+      accessible: true,
```
rurikoaraki, line 20: accessible, focusable, and accessibilityRole should be props that can be overridden by props passed in by a parent prop, so these should be configured in useRadio.

Also, we try to keep styling files focused on token integration and styling, and accessible, focusable, and accessibilityRole are not considered styles.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,81 @@
...
+      (tokens: RadioTokens) => ({
+        style: {
+          backgroundColor: 'transparent',
+          width: radioSize,
+          height: radioSize,
+          top: 0,
+          left: 0,
+          borderWidth: 1,
+          borderStyle: 'solid',
+          borderRadius: radioRadius,
```
rurikoaraki, line 42: This seems like something that should be customizable. Are you trying to implement all tokens in this PR or will you be implementing more tokens in the future?

packages/experimental/RadioGroup/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return (final: RadioProps, ...children: React.ReactNode[]) => {
+      const { label, ...mergedProps } = mergeProps(radio.props, final);
+
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.button>
+            <Slots.innerCircle />
+          </Slots.button>
+          <Slots.label>{label}</Slots.label>
+          {children}
```
rurikoaraki, line 51: I wonder what children we should expect to see? Should we allow children to be passed in at all?

rurikoaraki: FluentUI's version doesn't seem to, as a piece of input.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Inner circle color when selected
+   */
+  radioFill?: ColorValue;
+
+  /**
+   * Label text color
+   */
+  labelColor?: ColorValue;
```
rurikoaraki, line 29: Can we use `color` from font tokens?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+  }, [info, value, componentRef]);
+
+  // Ensure focus is placed on button after click
+  const changeSelectionWithFocus = useOnPressWithFocus(componentRef, changeSelection);
+
+  /* RadioButton changes selection when focus is moved between each RadioButton and on a click */
+  const pressable = useAsPressable({
+    ...rest,
+    onPress: changeSelectionWithFocus,
+    onFocus: changeSelection,
```
rurikoaraki, line 48: Does FocusZone make sure that if you press "tab" while in the focus zone, that you leave the focus zone?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+  return {
+    props: {
+      value: value,
+      ...rest,
+      ref: buttonRef,
+      ...pressable.props,
+      accessibilityRole: 'radio',
+      accessibilityLabel,
+      accessibilityState: { disabled: state.disabled, selected: state.selected },
+      accessibilityActions: [{ name: 'Select', label: radioSelectActionLabel }],
```
rurikoaraki, line 78: I don't think you need label here

rurikoaraki, line 78: The array passed to accessibilityActions should be a const so that a rerender doesn't trigger a wasted render if nothing changes

Also, the way the prop is currently set up, accessibilityActions doesn't allow for other types of actions to be added. Is this intended?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  return {
+    props: {
+      value: value,
+      ...rest,
+      ref: buttonRef,
+      ...pressable.props,
+      accessibilityRole: 'radio',
+      accessibilityLabel,
```
rurikoaraki, line 76: Should we fall back to anything if accessibilityLabel is undefined?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+      info.onChange && info.onChange(value);
+      info.updateSelectedButtonRef && componentRef && info.updateSelectedButtonRef(componentRef);
+    }
+  }, [info, value, componentRef]);
+
+  /* We use the componentRef of the currently selected button to maintain the default tabbable
+    element in a RadioGroup. Since the componentRef isn't generated until after initial render,
+    we must update it once here. */
+  React.useEffect(() => {
+    if (value == info.value) {
```
rurikoaraki, line 36: It was hard to realize that info.value tracks the current selected radio. Can we rename it to make it more obvious that it's holding the current selection?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+      info.updateSelectedButtonRef && componentRef && info.updateSelectedButtonRef(componentRef);
+    }
+  }, [info, value, componentRef]);
```
rurikoaraki, line 39: I feel like this would fire more often than it should because it's taking a dependency on the entire info object instead of the parts of the info object that matter, so any time anything on info is changed, info would be recreated and trigger this useEffect.

rurikoaraki: Also this seems redundant except for the first time it runs because of line 28... Is it possible to have it only run after first render by passing [] as the dependency array? Since this would run after the first render the ref should be populated by the time it runs the first time...

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,86 @@
...
+    props: {
+      value: value,
+      ...rest,
+      ref: buttonRef,
+      ...pressable.props,
+      accessibilityRole: 'radio',
+      accessibilityLabel,
+      accessibilityState: { disabled: state.disabled, selected: state.selected },
+      accessibilityActions: [{ name: 'Select', label: radioSelectActionLabel }],
+      accessibilityPositionInSet: accessibilityPositionInSet ?? info.buttonKeys.findIndex((x) => x == value) + 1,
```
rurikoaraki, line 79: This seems a bit fragile... it depends on buttonKeys being in the same order as the Radios passed into the RadioGroup. Do we have to manually populate a11yPosInSet (i.e. Windows won't do the work for us?) If so we might need to figure out a more robust way to keep track of child ordering

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { RadioTokens } from './Radio.types';
+
+export const defaultRadioTokens: TokenSettings<RadioTokens, Theme> = (t: Theme) =>
+  ({
+    borderColor: t.colors.menuItemText,
+    color: t.colors.menuItemText,
+    backgroundColor: t.colors.menuItemText,
+    textBorderColor: t.colors.transparentStroke,
+    radioVisibility: 1,
```
rurikoaraki, line 13: If the default state is unchecked, then I think radioVisibility should start at 0

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
```diff
@@ -0,0 +1,67 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+
+import { RadioTokens } from './Radio.types';
+
+export const defaultRadioTokens: TokenSettings<RadioTokens, Theme> = (t: Theme) =>
```
rurikoaraki, line 7: Just a general note that while I recognize we don't have redlines, generally Fluent 2 controls have rounded corners, so that might be one exception to the "stick with the old visuals" rule

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: That makes sense. These props are now configured in useRadio.

packages/experimental/RadioGroup/src/Radio/Radio.tsx
gulnazsayed: Good catch, thanks! We don't want to pass in children anymore. If any arbitrary views need to be used, they would be placed next to a Radio.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Are you suggesting using `fontStyles` directly in the styling versus having `labelColor` as a token here?

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
gulnazsayed: Thanks for catching that!

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
gulnazsayed: Would you suggest adding rounded corners to the focusable border that appears on the selected Radio?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Thanks for pointing that out! I renamed it to `selectedInfo`, but please let me know if that's still confusing.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: I'm trying to implement all of the tokens in this PR. Would you suggest adding the size of the Radio as a RadioToken?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: We can fallback to `label` if `accessibilityLabel` is undefined. Updated!

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
rurikoaraki: Yeah, that's what I would expect to see.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
rurikoaraki: color would come from IColorTokens but yeah, in RN `color` is recognized as the font color prop since that's the prop name used for font color in the `Text` component.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
rurikoaraki: Yeah, anything that a client might want to change styling wise should be a token. So the size of the radio would be one of those.

ValentinaKozlova: I think border, also, can be customizable. We usually use `...borderStyles.from(tokens, theme)`

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: The way buttonKeys is populated in RadioGroup means that it should theoretically be in the correct order every time. I'm also not sure what scenario might occur for them _not_ to be in the same order? 

Discussed with @ksiler about this and agreed that we will keep this as is for the time being. We do have to manually populate a11yPosInSet and the alternative to this approach would be to have RadioGroup check how many children/position on render which would be more expensive as it would re-count on every re-render.

apps/fluent-tester/src/E2E/RadioGroupExperimental/pages/RadioGroupExperimentalTabsPageObject.win.ts
```diff
@@ -0,0 +1 @@
+// TODO
```
ValentinaKozlova, line 1: Do you plan to do this in the next PR?

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,76 @@
...
+  states: radioStates,
+  slotProps: {
+    root: buildProps(() => ({
+      style: {
+        display: 'flex',
+        alignItems: 'center',
+        flexDirection: 'row',
+        minHeight: 20,
+        marginTop: 0,
+        position: 'relative',
```
ValentinaKozlova, line 24: position relative in RN is by default

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,76 @@
...
+          top: 0,
+          left: 0,
+          borderWidth: 1,
+          borderStyle: 'solid',
+          borderRadius: radioRadius,
+          borderColor: tokens.radioBorder,
+          marginTop: 4,
+          marginRight: 6,
+          marginBottom: 6,
+          marginLeft: 6,
```
ValentinaKozlova, line 42: I think we can use global tokens for that -> globalTokens.spacing

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -0,0 +1,76 @@
...
+        },
+      }),
+      ['radioBorder'],
+    ),
+    innerCircle: buildProps(
+      (tokens: RadioTokens) => ({
+        style: {
+          position: 'relative',
```
ValentinaKozlova, line 50: Here's the same thing, relative is defult position

apps/fluent-tester/src/E2E/RadioGroupExperimental/pages/RadioGroupExperimentalTabsPageObject.win.ts
gulnazsayed: Yes, this will be in the next PR aligning with the new async testing framework Sammy is working on.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Thanks for pointing that out, removed!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Removed here as well

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Fixed both of these!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Makes sense, fixed it! Thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
ksiler: Right, the label field is optional and goes unused AFAIK

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,90 @@
+import { RadioProps, RadioState } from './Radio.types';
+import * as React from 'react';
+import { useRadioGroupContext } from '../RadioGroup/radioGroupContext';
+import { useAsPressable, useOnPressWithFocus, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+
+export const radioSelectActionLabel = 'Select a RadioButton';
```
rurikoaraki, line 6: >'Select a RadioButton' [](http://example.com/codeflow?start=38&length=22)

We shouldn't use hardcoded strings as labels/accessibility labels

packages/experimental/RadioGroup/src/Radio/useRadio.ts
rurikoaraki: Ah I misunderstood how the buttonKeys info was populated. I thought it was a prop to radioGroup for some reason.

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: It wasn't intended, but I wonder what other actions we might want besides select? I fixed the logic here and removed instances of hard-coded strings.

gulnazsayed: Removed!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Removed all instances of hard-coded strings, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Removed!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Inner circle color when selected
+   */
+  radioFill?: ColorValue;
+
+  /**
+   * Label text color
+   */
+  // labelColor?: ColorValue;
```
rurikoaraki, line 29: Is the commented out token going to be used later? If not, commented code should be deleted/not checked in

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
gulnazsayed: Replaced with enableFocusRing and added rounded corners.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Just didn't get around to cleaning up the code yet! Removed now.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Added! Is it alright that I put it directly in the styling file so that I don't have to create new RadioTokens for these?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: That makes sense! Yes, I can pass in an empty dependency array so that it only runs once at the start. Updated!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
rurikoaraki: I think margins are pretty common for clients to want to customize. I would ask PM what they want.

gulnazsayed: Checked with PM and web spec. I don't think we want to make it customizable at this time.

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Yes it does.

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,97 @@
...
+  const changeSelectionWithFocus = useOnPressWithFocus(componentRef, changeSelection);
+
+  /* RadioButton changes selection when focus is moved between each RadioButton and on a click */
+  const pressable = useAsPressable({
+    ...rest,
+    onPress: changeSelectionWithFocus,
+    onFocus: changeSelection,
+  });
+
+  const accessibilityActionsProp = accessibilityActions
```
rurikoaraki, line 55: >const accessibilityActionsProp = accessibilityActions [](http://example.com/codeflow?start=2&length=53)

Creating this array probably needs to be memoized otherwise the array is recreated on every render.

rurikoaraki: Sorry "this array" is on the next line.

gulnazsayed: I used a callback and memoized it by leaving the dependency array empty. Does that sound right?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -0,0 +1,97 @@
...
+  const changeSelectionWithFocus = useOnPressWithFocus(componentRef, changeSelection);
+
+  /* RadioButton changes selection when focus is moved between each RadioButton and on a click */
+  const pressable = useAsPressable({
+    ...rest,
+    onPress: changeSelectionWithFocus,
+    onFocus: changeSelection,
+  });
+
+  const accessibilityActionsProp = React.useCallback(() => {
```
rurikoaraki, line 55: post-merge commit:

This shouldn't be a callback function since it's not a function you're calling as a result of an event. It's also currently set to not change after first render, so if accessibilityActions change, as an example, that won't be recognized by the component.

You should use [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) for this code.

PR #1989, opened 2022-08-11
Win32 Switch Shimmy Bug Fixed

nabil-k:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The Switch control used to "jiggle" when toggling if the onText and offText differed in sizes. Now we render a dummy On Text and Off Text at the same time a height of 0. This helps set the Switch to the correct width while also hiding the text. We then render the On/Off Text like how we normally would on top of the dummy texts. 

### Verification

- Toggled the Switch with different onText/offText sizes.
- Checked cases where the user may decide to change the onText/offText prop values after the initial render.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -47,15 +99,17 @@ export const Switch = compose<SwitchType>({
       const displayOnOffText = !!offText || !!onText;
       const isReduceMotionEnabled = AccessibilityInfo.isReduceMotionEnabled;
       const thumbAnimation = isReduceMotionEnabled ? { animationClass: 'Ribbon_SwitchThumb' } : null;
+      const currentOpacity = onOffTextTest ? 0 : 1; // hides the control during measurements
+      const newMinWidth = onOffTextTest ? null : { minWidth: finalWidth };
 
       return (
-        <Slots.root {...mergedProps}>
+        <Slots.root {...mergedProps} style={[{ opacity: currentOpacity }]}>
```
samuelfreiberg, line 106: Instead of passing in a `style` prop here, shouldn't we instead put this style in the .styling file?

samuelfreiberg:
"On my machine it feels like the control takes 1 second longer to load in."

This seems like a large performance hit for a bug like this. Does it take longer than other controls on the test page? Is it noticeable?

I'm wondering if there's anyway to fix this bug without the perf hit. I'll think about it a bit more today

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: Actually, doing that change (at least the way I did it), prevented the opacity from properly working. I'll look into doing this change in a better way.

nabil-k:
> This seems like a large performance hit for a bug like this. Does it take longer than other controls on the test page? Is it noticeable?
Yeah I agree. On the test page a button and a Switch which didn't have to do any measurements would load sooner than a Switch that had to do measurements. It's pretty noticeable if it's right next to other controls which like i said earlier happen to load sooner.

nabil-k:
> I'm wondering if there's anyway to fix this bug without the perf hit. I'll think about it a bit more today

If there's a way that would be neat. The reason I went w/ this approach was because I couldn't think of a way to predict text size besides actually rendering the text to measure it.

packages/experimental/Switch/src/Switch.types.ts
```diff
@@ -168,17 +173,21 @@ export interface SwitchProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
...
 export type SwitchState = IPressableState & { toggled?: boolean };
 
+export type textBeingTestedStates = 'init' | 'onText' | 'offText' | 'done';
+
 export interface SwitchInfo {
   props: IWithPressableEvents<SwitchProps & React.ComponentPropsWithRef<any>>;
   state: SwitchState;
 }
 
+export type toggleContainerProps = IViewProps & { ref: React.RefObject<View> };
```
rurikoaraki, line 183: >IViewProps [](http://example.com/codeflow?start=35&length=10)

Does `React.PropsWithRef<IViewProps>;` not work here?

packages/experimental/Switch/src/Switch.win32.tsx
rurikoaraki: This will fail a rerender test, so it should be memoized.

packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -36,26 +37,80 @@ export const Switch = compose<SwitchType>({
...
+
     // grab the styled slots
     const Slots = useSlots(userProps, (layer) => switchLookup(layer, switchInfo.state, switchInfo.props));
 
+    /*
+      Controls the rendering of the onText and offText in order to take measurements of what the minWidth should
+      be. Having the correct minWidth prevents the control from "jiggling" when the onText and offText
+      are different sizes.
+    */
+    const measureOnOffTexts = React.useCallback(
```
rurikoaraki, line 55: >const measureOnOffTexts = React.useCallback( [](http://example.com/codeflow?start=4&length=44)

We've generally put all hooks in the use<Component> function. This allows clients to reuse those hooks if they want to use it for a different component or change the render function of the component for some reason.

I think it's fine that the measuring code isn't in the main useSwitch hook but I think it would be good for it to be in its own hook so that if people want to modify the render function, they don't have to rewrite all of this logic.

packages/experimental/Switch/src/Switch.win32.tsx
rurikoaraki: Here's an example of a rerender test: [https://github.com/microsoft/fluentui-react-native/blob/a077890983b969cec18bfc138917a7499f46ef5b/packages/components/Button/src/Button.test.tsx#L66](https://github.com/microsoft/fluentui-react-native/blob/a077890983b969cec18bfc138917a7499f46ef5b/packages/components/Button/src/Button.test.tsx#L66)

packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -36,26 +37,80 @@ export const Switch = compose<SwitchType>({
...
+
     // now return the handler for finishing render
     return (final: SwitchProps) => {
       const { label, offText, onText, labelPosition, ...mergedProps } = mergeProps(switchInfo.props, final);
       const onOffText = switchInfo.state.toggled ? onText : offText;
       const displayOnOffText = !!offText || !!onText;
       const isReduceMotionEnabled = AccessibilityInfo.isReduceMotionEnabled;
       const thumbAnimation = isReduceMotionEnabled ? { animationClass: 'Ribbon_SwitchThumb' } : null;
+      const currentOpacity = onOffTextTest ? 0 : 1;
+      const newMinWidth = onOffTextTest ? null : { minWidth: finalWidth };
```
rurikoaraki, line 104: > { minWidth: finalWidth }; [](http://example.com/codeflow?start=48&length=26)

This will also fail a rerender check

packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -36,26 +37,80 @@ export const Switch = compose<SwitchType>({
...
+      }
+    }, [onText, offText]);
+
     // now return the handler for finishing render
     return (final: SwitchProps) => {
       const { label, offText, onText, labelPosition, ...mergedProps } = mergeProps(switchInfo.props, final);
       const onOffText = switchInfo.state.toggled ? onText : offText;
       const displayOnOffText = !!offText || !!onText;
       const isReduceMotionEnabled = AccessibilityInfo.isReduceMotionEnabled;
       const thumbAnimation = isReduceMotionEnabled ? { animationClass: 'Ribbon_SwitchThumb' } : null;
```
rurikoaraki, line 102: >{ animationClass: 'Ribbon_SwitchThumb' } [](http://example.com/codeflow?start=53&length=40)

This will also fail a rerender check

packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -36,26 +37,80 @@ export const Switch = compose<SwitchType>({
...
+      const currentOpacity = onOffTextTest ? 0 : 1;
+      const newMinWidth = onOffTextTest ? null : { minWidth: finalWidth };
 
       return (
-        <Slots.root {...mergedProps}>
+        <Slots.root {...mergedProps} style={{opacity: currentOpacity}}>
           <Slots.label>{label}</Slots.label>
-          <Slots.toggleContainer>
+          <Slots.toggleContainer ref={toggleContainerRef} style={newMinWidth}>
```
rurikoaraki, line 109: >style={newMinWidth} [](http://example.com/codeflow?start=58&length=19)

Would this stomp any styling that the toggleContainer has on it from buildSlotProps/the styling file?

packages/experimental/Switch/src/Switch.win32.tsx
```diff
@@ -36,26 +37,80 @@ export const Switch = compose<SwitchType>({
...
+    }, [onText, offText]);
+
     // now return the handler for finishing render
     return (final: SwitchProps) => {
       const { label, offText, onText, labelPosition, ...mergedProps } = mergeProps(switchInfo.props, final);
       const onOffText = switchInfo.state.toggled ? onText : offText;
       const displayOnOffText = !!offText || !!onText;
       const isReduceMotionEnabled = AccessibilityInfo.isReduceMotionEnabled;
       const thumbAnimation = isReduceMotionEnabled ? { animationClass: 'Ribbon_SwitchThumb' } : null;
+      const currentOpacity = onOffTextTest ? 0 : 1;
```
rurikoaraki, line 103: >  const currentOpacity = onOffTextTest ? 0 : 1; [](http://example.com/codeflow?start=4&length=47)

This would mean that in the weird case that the Switch's on/off text was programmatically changed midway in a session, that the switch would flicker since it would disappear and then come back once it's done measuring

rurikoaraki: Opacity also doesn't hide the component from screen readers so the switch (and I assume the on/off text) would still be readable via a screen reader while it's switching between the text to figure out what width it should have. It's not likely that that would be an issue, but something to note.

rurikoaraki:
I think this is a case where it makes sense to keep components for both the on and off text around. This saves having to flip between the two texts to measure both before settling on the correct label, though it is more memory since we keep both components around. You'll also have to hide the unused text both visually and from screen readers.

The less smart but also less perf-hit way to do this is to enforce a strict width for the on/off label that seems reasonable and that would scale with text scaling.

packages/experimental/Switch/src/Switch.types.ts
nabil-k: No it doesn't. When I try it I get the error `Type 'IViewProps' does not satisfy the constraint 'ElementType<any>`

nabil-k: I get the error `Type 'MutableRefObject<any>' has no properties in common with type 'IViewProps'` when I add it.

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: I've tested it and it looks like it doesn't.

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: Changed implementation.

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: Changed implementation.

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: Changed implementation.

packages/experimental/Switch/src/Switch.types.ts
nabil-k: Changed implementation.

packages/experimental/Switch/src/Switch.win32.tsx
nabil-k: Changed implementation.

nabil-k:
No longer required, this behavior is expected.

PR #1990, opened 2022-08-12
Shadow: Reduce extra views from 2 to 1

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Removed one of the extra views that was being added for the key shadow, and moved those props to the child view.

### Verification

No visible changes, but verified with the snapshots that the key shadow props are now included in the child props rather than in a separate view.

Note: don't think it matters whether the ambient shadow is on the inside or the key shadow is on the inside, tried both and they look the same visually.


Saadnajmi (approved):

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -7,13 +7,13 @@ import { getShadowTokenStyleSet } from './shadowStyle';
...
-    const mergedProps = mergeProps(final, { style: shadowTokenStyleSet.key });
+    const mergedProps = mergeProps(final, { style: shadowTokenStyleSet.ambient });
 
-    return (
-      <View style={shadowTokenStyleSet.ambient}>
-        <View {...mergedProps}>{children}</View>
-      </View>
-    );
+    const childrenArray = React.Children.toArray(children) as React.ReactElement[];
+    const child = childrenArray[0];
```
Saadnajmi, line 13: Can we add some sort of check to child to enforce that is a <View> ? Also can we put it in the SPEC.md that Shadow _must_ take one child of type View?

rurikoaraki: There is a React.Children.only() function which will check if there is only one child: [https://reactjs.org/docs/react-api.html#reactchildrenonly](https://reactjs.org/docs/react-api.html#reactchildrenonly)

In theory most children extend ViewProps so I don't think we need to necessarily enforce that the child is a View element? But I guess doing that is easier than trying to enforce that it has a style prop which can take shadow styling.

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -7,13 +7,13 @@ import { getShadowTokenStyleSet } from './shadowStyle';
...
+    const childWithKeyShadow = React.cloneElement(child, mergeProps(child.props, { style: shadowTokenStyleSet.key }));
```
rurikoaraki, line 14: >(child.props, { style: shadowTokenStyleSet.key } [](http://example.com/codeflow?start=67&length=48)

This would stomp over any styling that was set on the child, I think?

rurikoaraki: Oh nvm, I forgot that mergeProps automatically handles merging styling smartly, too.

packages/experimental/Shadow/src/Shadow.tsx
rurikoaraki: In terms of checking for a specific type of child:

[https://stackoverflow.com/questions/27366077/only-allow-children-of-a-specific-type-in-a-react-component#:~:text=Use%20the%20React.Children.forEach%20method%20to%20iterate%20over%20the,%28%22Only%20card%20components%20allowed%20as%20children.%22%29%3B%20%7D%20%7D](https://stackoverflow.com/questions/27366077/only-allow-children-of-a-specific-type-in-a-react-component#:~:text=Use%20the%20React.Children.forEach%20method%20to%20iterate%20over%20the,%28%22Only%20card%20components%20allowed%20as%20children.%22%29%3B%20%7D%20%7D)

rurikoaraki: Ideally we would be able to change the type of children to only be a single node or something instead of an Array but I haven't figured out how to do that with the stagedRender.

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: Yep I tested to make sure that styling on the child is merged correctly! Can also see it in the snapshots

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: I looked into React.Children.only, but my code kept triggering this error "React.Children.only expected to receive a single React element child" - I think because children is a ReactNode instead of a ReactElement. Didn't know if there was a good way to fix it, but I saw that in MenuTrigger, where I got some of the code for Shadow, we check the number of elements in the array and throw a console warning if there isn't exactly one. I copied that approach and also added a warning if the child isn't a view

lyzhan7:
Realized my check for the type to be View doesn't always work, ex. child.type will return false for Text components when I expect it to return true. Will continue looking into this

Saadnajmi:
> Realized my check for the type to be View doesn't always work, ex. child.type will return false for Text components when I expect it to return true. Will continue looking into this

Text isn't a view, so that makes sense. We can leave out the check if it's too restrictive.

lyzhan7:
Yeah I think I'm just going to remove it + update the documentation. I was trying to figure out if I could check that the child will handle the style prop but don't know if that's possible

PR #1997, opened 2022-08-13
Initial SPEC.md for shadows

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Initial documentation for shadows, including screenshots of the ramp, sample code, and some notes about the API.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Shadow/SPEC.md
+++ b/packages/experimental/Shadow/SPEC.md
@@ -1,18 +1,19 @@
 # Shadow
 
 ## Background
+
 A `Shadow` is a component that can be used to add Fluent shadows to other components.
 
-Shadows defined in the Fluent design system are actually comprised of two shadows. The **key shadow** is used to create a feeling of distance, and the **ambient shadow** is used to define the edges of the shape. 
+Shadows defined in the Fluent design system are actually comprised of two shadows. The **key shadow** is used to create a feeling of distance, and the **ambient shadow** is used to define the edges of the shape.
 
 ## Shadow Depth Ramp
 
 These are the current `Shadow` variants:
 
-| | Light mode | Dark mode |
-| - | - | - |
+|                 | Light mode                                                                                                                                                                | Dark mode                                                                                                                                                                 |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Regular shadows | <img width="801" alt="Screen Shot 2022-08-12 at 10 46 45 AM" src="https://user-images.githubusercontent.com/78454019/184426272-fe23d23f-3c60-4811-82d8-16ae8dc501d9.png"> | <img width="788" alt="Screen Shot 2022-08-12 at 12 08 56 PM" src="https://user-images.githubusercontent.com/78454019/184427236-0b93d797-fdde-4367-9caa-751bcf335eea.png"> |
-| Brand shadows | <img width="802" alt="Screen Shot 2022-08-12 at 12 09 15 PM" src="https://user-images.githubusercontent.com/78454019/184427300-06dacedc-5f39-4dc1-b07b-536186115d2e.png"> | <img width="789" alt="Screen Shot 2022-08-12 at 12 09 03 PM" src="https://user-images.githubusercontent.com/78454019/184427338-c322d223-4719-4593-a550-8360c45aa2e5.png"> |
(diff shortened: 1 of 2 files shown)


packages/experimental/Shadow/SPEC.md
```diff
@@ -1,5 +1,45 @@
...
+```
+For more examples of using Shadow, please see the [ShadowTest test page](https://github.com/microsoft/fluentui-react-native/tree/main/apps/fluent-tester/src/TestComponents/Shadow) in the [Fluent Tester app](https://github.com/microsoft/fluentui-react-native/blob/main/apps/fluent-tester/README.md).
+
+For an example of adding a Shadow as a slot to a Fluent component, please see the [FAB component](https://github.com/microsoft/fluentui-react-native/tree/main/packages/components/Button/src/FAB) (only on the iOS platform)
+
+## API
+### Props
+
+The `Shadow` component takes in one prop called `ShadowToken`. The variants of ShadowToken are all defined in the theme.
+If `undefined` is passed into the `shadowToken` prop, no shadow will be rendered.
```
Saadnajmi, line 39: We don't do this in code right? We could make it so that if shadowToken is undefined, we only render the child without the ambient shadow? 

Also would we prefer to have a "shadow0" token in the ramp, that does the same effect? That way, instead of specifying `undefined` in Token files, we can specify "t.shadows.shadow0"?


The first we should probably do if possible, the second I'm not sure if it's a good idea or not, so I'm open to more discussion about it.

packages/experimental/Shadow/SPEC.md
```diff
@@ -1,5 +1,45 @@
...
+
+`Shadow` must take exactly one child, which must be of type View.
+
+## Notes
+- Known issue: there may be some slight rounding discrepancies. Ex. opacity 0.24 may end up at 0.25.
```
Saadnajmi, line 44: "discrepancies due to hex to decimal rounding errors"? Once I read that line, I felt like I wanted to know the reason 😅

packages/experimental/Shadow/SPEC.md
```diff
@@ -1,5 +1,45 @@
...
+</Shadow>
+```
+```tsx
+<Shadow shadowToken={theme.shadows.shadow8brand}>
+  <Text>Text box with shadow8brand</Text>
+</Shadow>
+```
+For more examples of using Shadow, please see the [ShadowTest test page](https://github.com/microsoft/fluentui-react-native/tree/main/apps/fluent-tester/src/TestComponents/Shadow) in the [Fluent Tester app](https://github.com/microsoft/fluentui-react-native/blob/main/apps/fluent-tester/README.md).
+
+For an example of adding a Shadow as a slot to a Fluent component, please see the [FAB component](https://github.com/microsoft/fluentui-react-native/tree/main/packages/components/Button/src/FAB) (only on the iOS platform)
```
Saadnajmi, line 33: Nit: iOS and Android. Also Notification is another example we can use but I think Fab night be a simpler example now that I think about it.

packages/experimental/Shadow/SPEC.md
lyzhan7: Yes sorry that's totally my bad, I have another PR in progress that's not up yet that makes it so we don't render any extra shadows if undefined gets passed in, will have that PR up soon. I saw the conversation on one of Joanna's Notification PRs about shadow0, I've spent some time thinking about it, and yeah not sure what to do...if we have a shadow0 we wouldn't actually be using any of the values, so it seems strange to have to access the theme for that? But in terms of what to pass into the token slot, shadow0 is the most intuitive, and it would work. I was thinking maybe we could support both undefined getting passed in & shadow0 getting passed in, neither of which would render extra views.

Saadnajmi: For simplicity's sake, we can leave it as `undefined` and then introduce shadow0 if people really don't like writing undefined or it trips them up.

lyzhan7: That sounds good

harrieshin (approved):

packages/experimental/Shadow/SPEC.md
```diff
@@ -2,4 +2,48 @@
...
+For an example of adding a Shadow as a slot to a Fluent component, please see the [FAB component](https://github.com/microsoft/fluentui-react-native/tree/main/packages/components/Button/src/FAB) - this component exists on both iOS and Android, but currently only the iOS version uses the Shadow component. The [Notification component](https://github.com/microsoft/fluentui-react-native/tree/main/packages/components/Notification) is another example that uses the Shadow component.
+
+## API
+
+### Props
+
+The `Shadow` component takes in one prop called `ShadowToken`. The variants of ShadowToken are all defined in the theme.
+If `undefined` is passed into the `shadowToken` prop, no shadow will be rendered.
+
+`Shadow` must take exactly one child, which should be of type View.
```
rurikoaraki, line 45: The sample code contradicts this statement. It would be more accurate to say that the child must support a style which derives from ViewStyle, or something like that.

lyzhan7: Yes meant to update this but was working on the wording, thanks for the suggestion & just fixed!

PR #2001, opened 2022-08-15
Shadow: don't render the extra view if there isn't a shadow

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Follow up to a conversation in #1897 

Just return the child if there is no shadow (i.e. if ShadowToken is undefined) - don't add an extra view, or add unnecessary shadow props. Also updated Notification to pass in undefined as well.

Passing in undefined to not render is a Shadow isn't ideal, but not sure what a better solution would be. The other idea we had was to add a shadow0 token to the token pipeline, which would work too, it just seems a little strange because I don't think we'd ever actually use any of the values. Still though, if passing in undefined is confusing we could add support so that if either undefined or shadow0 are passed in, we just return the child.

### Verification

No visible changes, just verified with the snapshot tests. Added a test for a FAB that doesn't have a shadow, first commit contains snapshot of what this snapshot used to be before the change, second commit contains the change + what the snapshot looks like after the change - no longer has the extra view or the extra shadow props.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
+++ b/packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
@@ -2,84 +2,64 @@
 
 exports[`Custom FAB with no shadow(iOS) 1`] = `
 <View
+  accessibilityLabel="Custom FAB with no shadow(iOS)"
+  accessibilityRole="button"
+  accessibilityState={
+    Object {
+      "disabled": false,
+    }
+  }
+  accessible={true}
+  disabled={false}
+  enableFocusRing={true}
+  focusable={true}
+  iconPosition="before"
+  onBlur={[Function]}
+  onClick={[Function]}
+  onFocus={[Function]}
+  onPress={[Function]}
+  onResponderGrant={[Function]}
+  onResponderMove={[Function]}
+  onResponderRelease={[Function]}
+  onResponderTerminate={[Function]}
+  onResponderTerminationRequest={[Function]}
+  onStartShouldSetResponder={[Function]}
   style={
     Object {
-      "shadowColor": "000000",
-      "shadowOffset": Object {
-        "height": 0,
-        "width": 0,
-      },
-      "shadowOpacity": 0,
-      "shadowRadius": 0,
+      "alignItems": "center",
+      "alignSelf": "flex-start",
+      "backgroundColor": "#0078d4",
+      "borderColor": "#005a9e",
+      "borderRadius": 9999,
+      "display": "flex",
+      "flexDirection": "row",
+      "justifyContent": "center",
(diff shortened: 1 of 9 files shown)


packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
```diff
@@ -1,5 +1,68 @@
...
+      "backgroundColor": "#0078d4",
+      "borderColor": "#005a9e",
+      "borderRadius": 9999,
+      "display": "flex",
+      "flexDirection": "row",
+      "justifyContent": "center",
+      "minHeight": 56,
+      "minWidth": 56,
+      "padding": 16,
+      "width": undefined,
```
lyzhan7, line 40: Hmm just noticed that this got added & not sure why. Looking into it

rurikoaraki: Are you talking about this particular width style or the test snapshot itself? The test snapshot addition is expected since there's a new test in the file?

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -6,21 +6,25 @@ import { getShadowTokenStyleSet } from './shadowStyle';
 
 export const Shadow = stagedComponent((props: ShadowProps) => {
   return (final: ShadowProps, children: React.ReactNode) => {
-    const shadowTokenStyleSet = getShadowTokenStyleSet(props.shadowToken);
-    const mergedProps = mergeProps(final, { style: shadowTokenStyleSet.ambient });
+    if (!props.shadowToken) {
+      return <>{children}</>;
+    } else {
```
rurikoaraki, line 11: nit: We don't need to wrap the rest of the component in an else block. I generally prefer early returns over wrapping most of the code in an else block as it minimizes the "depth" of the code.

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Specifically just the "width": undefined line, it wasn't in the original snapshot.

Saadnajmi (approved):

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: DIdn't figure out why this was added but did see that this is in a lot of other snapshot tests, think it should be okay to merge

PR #2002, opened 2022-08-15
Switch's labelPosition set to 'after' and On/Off Text props are Exclusive 

nabil-k:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

If the labelPosition prop is set to 'after' the onText and offText will not render and the user will be given a warning. This change is needed to prevent the variable width of the onText and offText from changing the position of the Switch as the control is toggled.

Before this change, when the labelPosition was set to 'after', the Switch's On/Off Text would be placed _before_ the Switch. On/Off Text would then move the Switch around as the Switch's toggle state would change due to the variable length of the onText and offText props.

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Switch/SPEC.md
+++ b/packages/experimental/Switch/SPEC.md
@@ -88,7 +88,7 @@ export interface SwitchProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
   onText?: string;
 
   /**
-   * Sets the position of the Switch's label. The position value 'after' is mutually exclusive with the onText and offText props.
+   * Sets the position of the Switch's label. The position value 'after' is mutually exclusive with the onText and offText props. This is due to variable width of the text props causing the Switch's position to change when it shouldn't.
    */
   labelPosition?: 'before' | 'above' | 'after';
 }
--- a/packages/experimental/Switch/SPEC.md
+++ b/packages/experimental/Switch/SPEC.md
@@ -88,7 +88,7 @@ export interface SwitchProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
   onText?: string;
 
   /**
-   * Sets the position of the Switch's label
+   * Sets the position of the Switch's label. The position value 'after' is mutually exclusive with the onText and offText props.
    */
   labelPosition?: 'before' | 'above' | 'after';
 }
--- a/packages/experimental/Switch/SPEC.md
+++ b/packages/experimental/Switch/SPEC.md
@@ -88,7 +88,9 @@ export interface SwitchProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
   onText?: string;
 
   /**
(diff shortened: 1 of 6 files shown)


packages/experimental/Switch/SPEC.md
```diff
@@ -88,7 +88,7 @@ export interface SwitchProps extends Omit<IWithPressableOptions<ViewProps>, 'onP
   onText?: string;
 
   /**
-   * Sets the position of the Switch's label
+   * Sets the position of the Switch's label. The position value 'after' is mutually exclusive with the onText and offText props.
```
samuelfreiberg, line 91: I'd also do a quick explanation for why they're mutually exclusive.

For example:
" The position value 'after' is mutually exclusive with the onText and offText props. This is due to the variable width of the text props negatively affecting the render of the Switch"

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -31,7 +31,15 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
   const focusRef = isDisabled ? null : componentRef;
 
   if (__DEV__ && defaultChecked !== undefined && checked !== undefined) {
-    console.warn('The props defaultChecked and checked are mutually exclusive. Use only one of the props, do not use both.');
+    console.warn("The props 'defaultChecked' and 'checked' are mutually exclusive. Use only one of the props, do not use both.");
+  }
+
+  if (__DEV__ && labelPosition === 'after' && (!!props.onText || !!props.offText)) {
```
samuelfreiberg, line 37: Alternatively, you can simply just check for `if (__DEV__ && labelPosition === 'after')`. The only condition we need to know to set these values to null is labelPosition

samuelfreiberg:
Please update the description of the PR to include **why** we're making this change. It's always important to include as much information as possible because people might look back to this PR and wonder why a change was made. It's important to include a reasoning :)

packages/experimental/Switch/src/useSwitch.ts
nabil-k: If we had that, it would trigger a warning even if the user didn't set a onText/offText prop. So should we still warn them even if they're using the prop correctly?

packages/experimental/Switch/SPEC.md
nabil-k: Done, thanks!

packages/experimental/Switch/src/useSwitch.ts
rurikoaraki: Setting on/off text to null should happen regardless of whether we're in release/dev. Otherwise we'll have odd behavior in release even though the odd behavior doesn't show up in dev. But the warning should only fire in dev.

nabil-k: Oops, thanks for catching that!

rurikoaraki (approved):

samuelfreiberg (approved):

PR #2010, opened 2022-08-17
Improve Perf of E2E Testing

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

WebDriverIO 7.0 provides some important API improvements. One being the ability to [organize our test suites ](https://webdriver.io/docs/organizingsuites/#grouping-test-specs-to-run-sequentially)in more effective ways.

Specifically, this new change will make it so the app doesn't need to be rebooted for each control. Instead, it will continue running the session for each spec file. This will save time in the CI Pipeline because we won't need to reset the app after each component.

Specific improvement % not yet known - will review the CI to see.

This involved updated WDIO from v6 to v7 (something we should do anyways)

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Changes themselves look fine but the E2E tests seem to be failing now ☹️

samuelfreiberg:
> Changes themselves look fine but the E2E tests seem to be failing now ☹️

Weird, I had it working locally... Will investigate

rurikoaraki:
Actually I was trying to get appium to use the package that we get from declaring it as a dev dependency instead of letting it install it itself, since the package that it gets includes all of the driver's dev dependencies, which we don't need. Not worth trying to get to work in this PR though

PR #2013, opened 2022-08-18
Update Avatar color hashing to match native Avatar for iOS, macOS, Android

joannaquu:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Updated Avatar color hashing to match native Avatar for iOS, macOS, Android

### Verification

| Native                                       | New hashing                                      | Old hashing                                    |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 13 Pro - 2022-08-18 at 11 10 56 | Screen Shot 2022-08-19 at 9 20 10 AM | Screen Shot 2022-08-19 at 9 20 00 AM 
| Simulator Screen Shot - iPhone 13 Pro - 2022-08-18 at 11 10 11 | Screen Shot 2022-08-19 at 9 20 21 AM | Screen Shot 2022-08-19 at 9 19 48 AM |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Avatar/src/Avatar.types.ts
```diff
@@ -57,9 +57,6 @@ export const AvatarColors = [
   'mink',
   'platinum',
   'anchor',
-  'burgundy',
-  'hotPink',
-  'orchid',
```
ValentinaKozlova, line 62: Why are these colors removed?

apps/fluent-tester/src/TestComponents/Avatar/InitialsAvatar.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { FunctionComponent } from 'react';
```
ValentinaKozlova, line 1: Will it be more examples in this file? Because this color logic could be added to CustomizedAvatar section

rurikoaraki:
Please make these changes only affect macOS - win32 is aligning with the web's logic for now, so the expectation is that the hashing algorithm is different on each platform.

packages/components/Avatar/src/Avatar.types.ts
mischreiber: These three shouldn't be part of the set of colors used in avatar color calculation. The full set of colors used by Fluent UI Apple ([link here](https://github.com/microsoft/fluentui-apple/blob/dfa5ede5894c3a3db5367b7a1c7965eaa99cfb8e/ios/FluentUI/Avatar/AvatarTokens.swift#L145-L176)):
```
darkRed
cranberry
red
pumpkin
peach
marigold
gold
brass
brown
forest
seafoam
darkGreen
lightTeal
teal
steel
blue
royalBlue
cornflower
navy
lavender
purple
grape
lilac
pink
magenta
plum
beige
mink
platinum
anchor
```

rurikoaraki: win32 does use these colors though, and removing the colors from here affects that. Can you instead change the hash to only look at the first N-3 colors in the list?

packages/components/Avatar/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './Avatar';
 export * from './Avatar.types';
 export { useAvatar } from './useAvatar';
+export { getHashCodeWeb } from './getHashCode';
```
mischreiber, line 4: Does this mean we always export `getHashCodeWeb` to all customers? Is that something we want to do?

mischreiber, line 4: Does this mean we always export `getHashCodeWeb` to all customers? Is that something we want to do?

ValentinaKozlova: We used it for win32 because we were following the Web implementation. I'm not sure if each platform should have its own getHashCode algorithm. @rurikoaraki , what do you think?

ValentinaKozlova: Oh, I see that algorithm is changed back :)

joannaquu: So do we not want to export this?

ValentinaKozlova: Sorry for misleading comment. How you implemented it, it's good. Each platform will have it's own getHashCode algorithm. @rurikoaraki , correct me if I'm wrong

joannaquu: Ohh ok awesome, thank you!

packages/components/Avatar/src/getHashCode.ios.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * A function which returns hash code for the avatar color. iOS implementation
+ *
+ * @param str is idForColor or name
+ * @returns hash code of the color
+ */
+export const getHashCodeIOS = (str: string): number => {
```
harrieshin, line 7: scope creep but can we do the same thing for macOS too please?

harrieshin (approved):

mischreiber (approved):

apps/fluent-tester/src/TestComponents/Avatar/CustomizedAvatar.tsx
```diff
@@ -65,6 +65,9 @@ export const CustomizeUsage: React.FunctionComponent = () => {
     viewBox: '0 0 500 500',
   };
 
+  const useJavaHashCode = ['ios', 'macos', 'android'].includes(Platform.OS);
+  const hashedColor = AvatarColors[getJavaHashCode(name) % (AvatarColors.length - 3)];
+
```
Saadnajmi, line 70: Instead of doing this in the test, could we just make this the default in Avatar for these platforms?

Saadnajmi: (Sorry, I couldn't comment before this merged)

packages/components/Avatar/src/Avatar.types.ts
Saadnajmi: Why is there a design difference between web/win32 and mobile here? The list of colors feels like it should be a globally consistent thing. Should we take that up with design to update one or the other?

apps/fluent-tester/src/TestComponents/Avatar/CustomizedAvatar.tsx
rurikoaraki: Yeah, otherwise users of the Avatar in these platforms will have to do the same thing to get the expected behavior.

packages/components/Avatar/src/Avatar.types.ts
harrieshin: afaik web and ios were using the same colors. so i am confused as well.

Saadnajmi: Following up conversation offline

ValentinaKozlova: @Saadnajmi , @harrieshin  These three colors are used in PersonaCoin component

PR #2020, opened 2022-08-22
Switch High Contrast

nabil-k:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

No longer depending on the values of the tokens in the redlines to provide the colors for high contrast. Instead now we are checking to see if we're in high contrast mode and map tokens we picked ourselves to provide the colors for high contrast.

This makes the Switch's high contrast look closer to what we expected a high contrast Switch to look like. It also resolves the issue of the thumb not being visible when using high contrast on windows 11.

### Verification

Windows 10 High Contrast
| image | image |

Windows 11 Aquatic High Contract
Note: The first switch in these before/after screen shots has a mouse hovered over them.
| image | image |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Switch/src/SwitchTokens.win32.ts
+++ b/packages/experimental/Switch/src/SwitchTokens.win32.ts
@@ -1,0 +1,140 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { SwitchTokens } from './Switch.types';
+
+// toggleOn: {
+//   trackColor: t.colors.compoundBrandBackground1,
+//   thumbColor: t.colors.neutralForegroundOnBrandHover,
+//   borderColor: t.colors.compoundBrandBackground1,
+//   hovered: {
+//     trackColor: t.colors.neutralForegroundOnBrandHover,
+//     thumbColor: t.colors.compoundBrandBackground1Hover,
+//     borderColor: t.colors.compoundBrandBackground1Hover,
+//   },
+//   pressed: {
+//     trackColor: t.colors.compoundBrandBackground1Pressed,
+//     thumbColor: t.colors.neutralForegroundOnBrandHoverPressed,
+//     borderColor: t.colors.compoundBrandBackground1Pressed,
+//   },
+//   disabled: {
+//     trackColor: t.colors.neutralForegroundDisabled,
+//     thumbColor: t.colors.brandBackground,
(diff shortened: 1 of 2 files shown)


packages/experimental/Switch/src/SwitchTokens.win32.ts
```diff
@@ -0,0 +1,99 @@
...
+      thumbColor: isHighContrast(t) ? t.colors.compoundBrandBackground1Hover : t.colors.neutralForegroundInvertedLink,
+      borderColor: t.colors.compoundBrandBackground1Hover,
+    },
+    pressed: {
+      trackColor: t.colors.compoundBrandBackground1Pressed,
+      thumbColor: isHighContrast(t) ? t.colors.neutralForegroundOnBrandHoverPressed : t.colors.neutralForegroundInvertedLink,
+      borderColor: t.colors.compoundBrandBackground1Pressed,
+    },
+    disabled: {
+      trackColor: isHighContrast ? t.colors.neutralForegroundDisabled : t.colors.neutralBackgroundDisabled,
```
joannaquu, line 64: Should this be isHighContrast(t) as well?

nabil-k: thanks for catching that!

packages/experimental/Switch/src/SwitchTokens.win32.ts
```diff
@@ -0,0 +1,99 @@
...
+    toggleContainerFlexDirection: 'row',
+  },
+
+  after: {
+    flexDirection: 'row-reverse',
+    toggleContainerFlexDirection: 'row-reverse',
+  },
+
+  toggleOn: {
+    trackColor: t.colors.compoundBrandBackground1,
```
ksiler, line 49: for tokens that don't check for `isHighContrast(t)`, are they already resolving to the right colors for high contrast?

rurikoaraki (approved):

packages/experimental/Switch/src/SwitchTokens.win32.ts
nabil-k: Yeah they are. When testing the control in high contrast all the colors seem to be appropriate.

ksiler (approved):

PR #2027, opened 2022-08-24
Change Platform Status to be collapsible

joannaquu:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Changed Platform Status to be collapsible on macOS/win32. Also added Platform Status to mobile and fixed issue where Status Definitions don't wrap when resized horizontally on macOS/win32 test app. Previously, resizing the FluentTester window smaller caused the status definitions to extend beyond the page and be cut off. This change moves the status definitions below the platform status, allowing for correct wrapping.

### Verification

Before

https://user-images.githubusercontent.com/55368679/186288207-8c2d92c3-6f15-416c-86d0-871c8ffdcca0.mov

After

https://user-images.githubusercontent.com/55368679/186747691-ef79bda8-62d5-4724-a4ab-a0bff42dc164.mov

| Simulator Screen Shot - iPhone 13 Pro - 2022-08-25 at 10 20 03 | Simulator Screen Shot - iPhone 13 Pro - 2022-08-25 at 12 01 42 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Test.tsx
+++ b/apps/fluent-tester/src/TestComponents/Test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { StyleSheet, View, Platform } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 import { Text, ToggleButton, Separator } from '@fluentui/react-native';
 import { Stack } from '@fluentui-react-native/stack';
 import { stackStyle } from './Common/styles';
@@ -27,6 +27,14 @@ export interface TestProps {
   sections: TestSection[];
 }
 
+const definitions = {
+  Production: 'Control is ready for broad partner use and to be used in production-ready scenarios.',
+  Beta: 'Control is ready for partner consumption, but not ready for production release (e.g. fixing bugs).',
+  Experimental: 'Control code checked into repo, but not ready for partner use.',
+  Backlog: 'Control is in plan and on our backlog to deliver.',
+  'N/A': 'Control is not in current plan.',
+};
+
 const styles = StyleSheet.create({
   name: {
     marginTop: 4,
@@ -44,11 +52,8 @@ const styles = StyleSheet.create({
   },
   statusView: {
     flexDirection: 'row',
-    alignItems: 'flex-start',
+    alignItems: 'center',
     justifyContent: 'space-between',
-  },
-  statusHeader: {
-    marginBottom: 6,
   },
   statusLabel: {
     marginTop: 2,
@@ -58,9 +63,6 @@ const styles = StyleSheet.create({
     fontWeight: 'normal',
   },
 });
-
-// mobile platform check to not render status components.
-const isMobile = Platform.OS == 'android' || (Platform.OS == 'ios' && !Platform.isPad);
 
 export const Test = (props: TestProps): React.ReactElement<Record<string, never>> => {
   const [showStatus, setShowStatus] = React.useState(false);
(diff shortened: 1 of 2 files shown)


Saadnajmi:
Instead of a box inside a box, could all the text just be in one box?

Saadnajmi:
Talked with @rurikoaraki a bit more about this, and one suggestion was: can we make the definitions collapsible? 

To quote: "Should just be a ToggleButton with a chevron or plus sign which will toggle a bool that's stored as state. and depending on the bool's value we show the text or not"

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -58,6 +64,24 @@ const styles = StyleSheet.create({
...
+        <G>
+          <Path d={path} fill="black" />
+        </G>
+      </Svg>
+    );
+  };
+  const svgProps: SvgIconProps = {
+    src: toggleSvg,
+  };
+  const toggleIconProps = createIconProps({ svgSource: svgProps, width: 12, height: 12 });
```
Saadnajmi, line 83: This won't work on windows, where SVGs aren't supported yet. Could we use a font icon instead? Like literally the "+" and "-" characters maybe? 😅

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -67,55 +73,52 @@ export const Test = (props: TestProps): React.ReactElement<Record<string, never>
       <Stack style={stackStyle}>
         <Text style={styles.description}>{props.description}</Text>
       </Stack>
-      {!isMobile && (
+      <ToggleButton onClick={() => setShowStatus(!showStatus)} style={[styles.statusLabel]}>
+        Platform Status {showStatus ? '-' : '+'}
```
Saadnajmi, line 77: An FYI that our icon component supports font icons, which may be better for scaling here

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -63,6 +64,11 @@ const isMobile = Platform.OS == 'android' || (Platform.OS == 'ios' && !Platform.
 
 export const Test = (props: TestProps): React.ReactElement<Record<string, never>> => {
   const [showStatus, setShowStatus] = React.useState(false);
+  const fontBuiltInProps = {
+    fontFamily: 'Arial',
+    codepoint: showStatus ? 0x2796 : 0x2795,
```
Saadnajmi, line 69: Maybe leave a comment what these code points are? or do something ike:

```
const plusCodepoint = 0x2796;
const minusCodepoint = 0x2795;
```

Also what happens if we don't specify the font family? Or get the better the family from the theme?

```
const theme = useTheme()
const fontFamily = theme.typography.families.primary

```

Saadnajmi:
More nitpicks:

As is, it's not obvious what the ToggleButton does. I'd rather we either:
1) Always have the Platform Status be visible, and make the explainers collapsible (which I mentioned offline), orrr
2) Just move subheader "Platform Status" out of the text box and put it beside the button so it's always visible.

I'm not sure why, but I wasn't so much a fan of putting "Platform Status" inside the button. It felt.. odd? I think (2) is my preferred option.

Saadnajmi (approved):

Saadnajmi:
Now that it's collapsible, could we add it back to mobile perhaps?

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -58,6 +63,18 @@ const styles = StyleSheet.create({
 const isMobile = Platform.OS == 'android' || (Platform.OS == 'ios' && !Platform.isPad);
 
 export const Test = (props: TestProps): React.ReactElement<Record<string, never>> => {
+  const [showStatus, setShowStatus] = React.useState(false);
+  const theme = useTheme();
+  const fontFamily = theme.typography.families.primary;
+
+  const plusCodepoint = 0x2795;
```
harrieshin, line 70: can we use prettier fluent icons directly?

Saadnajmi: W can't use SVGs because they won't work properly on windows.. so maybe if pngs are available?

joannaquu: Pngs aren't available but pdfs are so I will try that.

rurikoaraki: Yeah pngs generally aren't great since they don't scale with DPI, so to handle that gracefully we would need to add assets for each DPI level.

It /is/ a test app so probably not as important though.

Personally I think this is fine since it adds a test case for font icons

joannaquu: Offline Saad had mentioned font icons for windows and the fluent svg icon for other platforms. The latest commit added that so any opinions?

harrieshin (approved):

PR #2065, opened 2022-08-31
Shadows: adjust blur value to match designs

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In Fluent UI Apple the blur had to be divided by 2 in order to match the [Figma](https://www.figma.com/file/KB9oUjMKen2cKnyPG7RgdS/Design-tokens-superset?node-id=4553%3A38) ([see note here](https://www.figma.com/file/KB9oUjMKen2cKnyPG7RgdS/Design-tokens-superset?node-id=5887%3A0)), and after checking in FURN dividing the blur by 2 also matches the Figma better. Screenshots from Figma:

| Light | Light Brand |
| - | - |
| Screen Shot 2022-08-30 at 5 26 56 PM | Screen Shot 2022-08-30 at 5 27 07 PM |

### Verification

| | Before                                       | After                                      |
| - |----------------------------------------------|--------------------------------------------|
| Light | before-light |  after-light |
| Light Brand |  before-light-brand | after-light-brand |
| Dark | before-dark | after-dark |
| Dark Brand | before-dark-brand | after-dark-brand

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Shadow/src/shadowStyle.ts
+++ b/packages/experimental/Shadow/src/shadowStyle.ts
@@ -2,6 +2,10 @@ import { memoize } from '@fluentui-react-native/framework';
 import { ShadowToken } from '@fluentui-react-native/theme-types';
 import { ColorValue } from 'react-native';
 
+// The blur property from the token is not the same as the shadow radius value,
+// it needs to be divided by 2 to achieve the same effect.
+const shadowBlurAdjustment = 0.5;
+
 export const getShadowTokenStyleSet = memoize(getShadowTokenStyleSetWorker);
 
 function getShadowTokenStyleSetWorker(shadowToken: ShadowToken) {
@@ -12,7 +16,7 @@ function getShadowTokenStyleSetWorker(shadowToken: ShadowToken) {
     key: {
       shadowColor: shadowColorFromRGBAColor(keyShadow.color),
       shadowOpacity: shadowOpacityFromRGBAColor(keyShadow.color),
-      shadowRadius: keyShadow.blur,
+      shadowRadius: keyShadow.blur * shadowBlurAdjustment,
       shadowOffset: {
         width: keyShadow.x,
         height: keyShadow.y,
@@ -21,7 +25,7 @@ function getShadowTokenStyleSetWorker(shadowToken: ShadowToken) {
     ambient: {
       shadowColor: shadowColorFromRGBAColor(ambientShadow.color),
       shadowOpacity: shadowOpacityFromRGBAColor(ambientShadow.color),
-      shadowRadius: ambientShadow.blur,
+      shadowRadius: ambientShadow.blur * shadowBlurAdjustment,
       shadowOffset: {
         width: ambientShadow.x,
         height: ambientShadow.y,
--- a/packages/experimental/Shadow/src/shadowStyle.ts
+++ b/packages/experimental/Shadow/src/shadowStyle.ts
@@ -3,7 +3,7 @@ import { ShadowToken } from '@fluentui-react-native/theme-types';
 import { ColorValue, Platform } from 'react-native';
 
 // For iOS/macOS, the blur property from the token is not the same as the shadow radius value,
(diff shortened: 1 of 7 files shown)


packages/experimental/Shadow/src/shadowStyle.ts
```diff
@@ -2,6 +2,10 @@ import { memoize } from '@fluentui-react-native/framework';
 import { ShadowToken } from '@fluentui-react-native/theme-types';
 import { ColorValue } from 'react-native';
 
+// The blur property from the token is not the same as the shadow radius value,
+// it needs to be divided by 2 to achieve the same effect.
+const shadowBlurAdjustment = 0.5;
+
```
Saadnajmi, line 8: I wonder 2 things:

1) Should we do this "transform" at the token pipeline level before it ever gets to FURN? My initial feeling is no, since the `shadowToken` type is not RN Specific at the moment and more like a designer value. `shadowStyle.ts` is doing the transform to RN types for us, so it feels appropriate here. But then.. should we move the whole `shadowStyle` function to the token pipeline then?


2) Should we restrict this per platform?

It seems this is an iOS /macOS specific issue.. If we can get a repro on win32 that would be nice. Till then could we only do the adjustment if we're macOS or iOS?

lyzhan7: 1. I briefly talked to Hunter about this, I think the value in the FUA token pipeline is the same as the Figma value now (at some point it was the adjusted value, but got switched back). Not really sure what you mean with moving the whole shadowStyle function to the token pipeline?
2. Yes possibly, I wasn't sure if win32 would require the same fix or not. I actually thought it was more likely that it would require the same fix, otherwise right now adding a shadow to RNmacOS vs. RNwin32 would look different? I noted it down as something that should be checked on win32

lyzhan7: Update after a lot of investigation/discussion offline- left shadowStyle function in FURN, and restricted to macOS/iOS.

packages/experimental/Shadow/src/shadowStyle.ts
```diff
@@ -1,16 +1,18 @@
 import { memoize } from '@fluentui-react-native/framework';
 import { ShadowToken } from '@fluentui-react-native/theme-types';
-import { ColorValue } from 'react-native';
+import { ColorValue, Platform } from 'react-native';
 
-// The blur property from the token is not the same as the shadow radius value,
+// For iOS/macOS, the blur property from the token is not the same as the shadow radius value,
 // it needs to be divided by 2 to achieve the same effect.
```
Saadnajmi, line 6: Once https://github.com/microsoft/apple-ux-guide/pull/13 lands, we can link it here

mischreiber (approved):

chiuam (approved):

packages/experimental/Shadow/src/shadowStyle.ts
```diff
@@ -3,7 +3,7 @@ import { ShadowToken } from '@fluentui-react-native/theme-types';
 import { ColorValue, Platform } from 'react-native';
 
 // For iOS/macOS, the blur property from the token is not the same as the shadow radius value,
-// it needs to be divided by 2 to achieve the same effect.
+// it needs to be divided by 2 to achieve the same effect. See https://github.com/microsoft/apple-ux-guide/blob/gh-pages/Shadows.md
```
Saadnajmi, line 6: nit:

Saadnajmi (approved):

packages/experimental/Shadow/src/shadowStyle.ts
lyzhan7: Sorry didn't see this before it got merged, will fix later

PR #2068, opened 2022-09-01
(macOS) Fix bug when FocusZone wraps a ScrollView

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There is a bug on macOS when a FocusZone contains a ScrollView inside of it.

FocusZone compares the rects of every view inside itself to figure out which one is closest in the cardinal direction corresponding to the users' arrow key press. However, If the firstResponder is in a ScrollView, we don't actually want to focus on the closest View in X direction to us. We instead want to focus on the closet View _also in the ScrollView_. Or if we're at the edge of the ScrollView, the closest view outside the edge. 

We didn't take into account whether or not we were in a ScollView, so we would sometimes move focus in and out of the scrollview with the same keypress as we kept jumping focus between a View outside the ScrollView and the next View inside the ScrollView. See the video here:

https://user-images.githubusercontent.com/6722175/187857010-6cf36ae4-fb94-44fc-8da7-77b00003f8f0.mov

Let's fix this by adding a special case in FocusZone logic for when the firstResponder is inside a ScrollView

I also took the time to rewrite and cleanup the FocusZone test page...

### Verification

Same test now has the expected behavior.

https://user-images.githubusercontent.com/6722175/187857229-f7555e76-5711-419c-b1d2-f07132ba3581.mov

Also a cleaner test page

image

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/FocusZone/FocusZoneTest.tsx
```diff
@@ -32,6 +32,17 @@ const ListOfDisabledCheckboxes: React.FunctionComponent = () => {
 const EdgeCasesFocusZone: React.FunctionComponent = () => {
   return (
     <Stack style={stackStyleFocusZone}>
+        <SubheaderText>FocusZone wrapping a ScrollView</SubheaderText>
+        <Button>Outside Focus Zone</Button>
+        <FocusZone focusZoneDirection="vertical">
+          <Button>Inside Focus Zone</Button>
+          <ScrollView style={{height: 100}} contentContainerStyle={{backgroundColor: 'grey'}} showsVerticalScrollIndicator={true}>
+              <Button style={{margin: 100}}>1</Button>
```
Saadnajmi, line 40: I promise to get rid of the inline styles before check in!

chiuam:
What's the expected behavior for tabbing into a scrollview from a non-FocusZone element?

Saadnajmi:
> What's the expected behavior for tabbing into a scrollview from a non-FocusZone element? 

Assuming the scrollview is inside the focus zone, I would assume it would focus on the first focusable view in the scroll view. Aka, view that's closest to top left .

chiuam:
> > What's the expected behavior for tabbing into a scrollview from a non-FocusZone element?
> 
> Assuming the scrollview is inside the focus zone, I would assume it would focus on the first focusable view in the scroll view. Aka, view that's closest to top left .

Would direction matter? Like if I wanna tab into the scrollview from the bottom, see screenshot below, pressing tab + shift should take me to button 3?
image

Saadnajmi:
> > > What's the expected behavior for tabbing into a scrollview from a non-FocusZone element?
> > 
> > Assuming the scrollview is inside the focus zone, I would assume it would focus on the first focusable view in the scroll view. Aka, view that's closest to top left .
> 
> Would direction matter? Like if I wanna tab into the scrollview from the bottom, see screenshot below, pressing tab + shift should take me to button 3?
> image
> 

I think that's an existing bug in FocusZone? It doesn't know if you're tabbing into it or shift+tabbing into it, so it'll always focus on the first focusable view or defaultTabbabbleElememt.

Saadnajmi:
Talked offline. For some reason `NSScroller` is focusable and messing up the key view loop / focus zone. Soo I need to special case that out..

Saadnajmi:
Did some more investigating about the NSScroller issue. 

I can't really intercept a `Shift+Tab` into a FocusZone, since that doesn't seem to call `acceptsFirstResponder` :( . As such, I can't stop NSScroller shows up in the key view loop. An alternate fix is to update React Native macOS's `RCTScrollView` class to not have it's `NSScroller` focusable. That would be a followup fix. 

I removed the custom handling for `NSScroller` and marking this ready for review now.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -17,6 +17,11 @@
 // enumerated in the same row (or column) as the current focused view
 static const CGFloat FocusZoneBuffer = 3;
 
+// Classes FocusZone should never focus on
+static NSArray<Class> *ViewClassesToSkip() {
```
amgleitman, line 21: Is this used anywhere?

amgleitman (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: oof, thought I deleted that. Thanks!

chiuam (approved):

mischreiber (approved):

PR #2082, opened 2022-09-01
Text empty props fix

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The Text component throws an error when it doesn't receive any props because it expects `props` to be defined when destructuring it. Unfortunately it's pretty common to do something like `<Text>test</Text>` so the component should be able to handle the case where props is undefined.

Adding a fallback empty object when props is undefined. This way the destructuring works because all the props are returned as undefined.

### Verification

The Option component which doesn't have any props for its Text slot now works, whereas before it would run into a redbox error.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/text/src/Text.tsx
+++ b/packages/components/text/src/Text.tsx
@@ -15,10 +15,15 @@ import { textName, TextProps, TextTokens } from './Text.types';
 import { useTextTokens } from './TextTokens';
 import React from 'react';
 
+const emptyProps = {};
 export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTokens: UseTokens<TextTokens>) => {
+  if (props === undefined) {
+    props = emptyProps;
+  }
+
   // split out color and variant from props
   const {
-    align,
+    align = undefined,
     block,
     color,
     font,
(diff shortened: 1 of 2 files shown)


packages/components/text/src/Text.tsx
```diff
@@ -15,10 +15,15 @@ import { textName, TextProps, TextTokens } from './Text.types';
...
 export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTokens: UseTokens<TextTokens>) => {
+  if (props === undefined) {
+    props = emptyProps;
+  }
+
   // split out color and variant from props
   const {
-    align,
+    align = undefined,
     block,
```
ValentinaKozlova, line 27: It's not part of this PR, but looks like `block` prop is not used in Text.tsx

rurikoaraki: Hmm I think that was something that was not fully resolved when Text was implemented? I'll let @lenahong decide what to do about that.

PR #2083, opened 2022-09-02
Transforming E2E Framework from Sync to Async and Updating WebDriverIO to V7

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

WebDriverIO's Synchronous framework is[ now deprecated](https://webdriver.io/docs/sync-vs-async/) due to some breaking changes in Node. We need to move to async. WDIO v7 has some crucial improvements to async that we need as well.

I also made a task to update the E2E Testing docs to include async examples.


samuelfreiberg:
Porting questions from previous PR here:

> Dumb, out of scope question: Does using an async framework mean we don't need waiters anymore?

I believe the answer here is yes. Let's take this example:

`browser.waitUntil(() => element.isDisplayed(), timeout: 25000, interval: 1000);`

vs.

`await element.isDisplayed()`

The async `await` won't continuously wait for that statement to be true, it will just wait until the first result of that is returned. The first example will wait a certain amount of time for a condition to be true - meaning it will check the condition more than once.

Does that make sense?

samuelfreiberg:
Porting questions from previous PR here:

> I think for some of the await code (like `await (await ...))`) the syntax I would've expected to see would've been something like
> 
> `const result = doSomething().then((result) => { return expect(result).toBeTruthy();});`
> 
> or something like that. But I understand that using await everywhere is probably easier to port and reason about.

I definitely understand the thought here, and I thought about it as well. I think for readability purposes, it's a lot easier and simpler to understand using the double `await`. From what I can, there aren't extra benefits to using one of the other. Maybe we can leave it like this for now and revisit the idea? I can make a task to do some light investigation

apps/fluent-tester/src/E2E/MenuButton/pages/MenuButtonPageObject.win.ts
```diff
@@ -1,80 +1,80 @@
-import {
-  MENU_BUTTON_TESTPAGE,
-  MENU_BUTTON_TEST_COMPONENT,
-  HOMEPAGE_MENUBUTTON_BUTTON,
-  MENU_BUTTON_NO_A11Y_LABEL_COMPONENT,
-  MENU_ITEM_1_COMPONENT,
-} from '../../../TestComponents/MenuButton/consts';
-import { BasePage, By } from '../../common/BasePage.win';
+// import {
```
ksiler, line 1: Not sure if you've already answered this but why are the MenuButton E2E test pages commented out? Should they be?

apps/fluent-tester/src/E2E/common/BasePage.win.ts
```diff
@@ -2,7 +2,7 @@ const DUMMY_CHAR = '';
 export const COMPONENT_SCROLL_COORDINATES = { x: -0, y: -100 }; // These are the offsets. Y is negative because we want the touch to move up (and thus it scrolls down)
 
 /* Win32/UWP-Specific Selector. We use this to get elements on the test page */
-export function By(identifier: string): WebdriverIO.Element {
+export function By(identifier: string) {
```
ksiler, line 5: This works? I remember having trouble converting this one and ended up doing this:

```
export async function By(identifier: string): Promise<WebdriverIO.Element> {
  const element = await $('~' + identifier);
  return element;
}
```
but what you've done looks cleaner.

apps/fluent-tester/src/E2E/common/NativeTesting/pages/NativeTestingPageObject.win.ts
```diff
@@ -1,56 +1,56 @@
...
-    return foundValidButton;
-  }
-
-  get _scrollViewParent() {
-    return By(ScrollViewTestId);
-  }
-}
-
-export default new NativeTestingPageObject();
+// import { BasePage, By } from '../../../common/BasePage.win';
```
ksiler, line 1: Similar question to the one above, is this meant to be commented out (along with the associated spec file)?

samuelfreiberg: Had a local issue with this one, but just figured it out. Adding back now!

apps/fluent-tester/src/E2E/MenuButton/pages/MenuButtonPageObject.win.ts
samuelfreiberg: Same as below, ran into local issue but I figured it out, adding it back now.

ksiler (approved):

rurikoaraki (approved):

apps/fluent-tester/src/E2E/common/BasePage.win.ts
samuelfreiberg: Yes, this works :)

Do you remember the issue you had with it?

PR #2094, opened 2022-09-07
[Link] Initial Implementation

lenahong:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [X] android

### Description of changes
Adding a new component package experimental-link using the new compose framework and integrated design tokens.
Created an Experimental Link test page. 

### Verification

link-test

### Pull request checklist

- [X] Automated Tests
- [X] Keyboard Accessibility
- [X] Voiceover
- [X] Internationalization and Right-to-left Layouts


packages/experimental/Link/src/Link.tsx
```diff
@@ -0,0 +1,53 @@
...
+import * as React from 'react';
+import { View } from 'react-native';
+import { linkName, LinkType, LinkProps, ILinkState } from './Link.types';
+import { compose, mergeProps, withSlots, UseSlots } from '@fluentui-react-native/framework';
+import { useLink } from './useLink';
+import { TextV1 as Text } from '@fluentui-react-native/text';
+import { stylingSettings } from './Link.styling';
+
+/**
+ * A function which determines if a set of styles should be applied to the compoent given the current state and props of the link.
```
samuelfreiberg, line 11: Nit: Spelling - "compoent" -> "component"

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,86 @@
...
+    enableFocusRing,
+    focusable,
+    ...rest
+  } = props;
+  const isDisabled = !!disabled;
+
+  const [visitedState, setVisitedState] = React.useState({ visited: false });
+  const linkOnPress = React.useCallback(
+    (e) => {
+      if (!disabled) {
```
samuelfreiberg, line 28: Just curious - does the onPress() callback get fired on a disabled component?

samuelfreiberg:
Can we add E2E tests along with this PR? Or a subsequent one? Just want to make sure we have baseline testing in to catch regressions as it's iterated on

packages/experimental/Link/src/useLink.ts
lenahong: it shouldn't. I can remove this check since it's not really necessary

apps/fluent-tester/src/TestComponents/Link/LinkTest.tsx
```diff
@@ -15,7 +15,7 @@ const Links: React.FunctionComponent = () => {
   return (
     <Stack style={stackStyle}>
       <Link url="https://www.bing.com/" content="Click to navigate." />
-      <Link onPress={doPress} content="Click to alert." />
+      <Link disabled onPress={doPress} content="Click to alert." />
```
rurikoaraki, line 18: Hmm maybe do a separate entry/test for the disabled case? It feels like having a proper test case for onPress would be useful separate from testing that disabled doesn't fire the onPress callback.

yarn.lock
```diff
@@ -328,7 +328,7 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.19.3.tgz#707b939793f867f5a73b2666e6d9a3396eb03151"
   integrity sha512-prBHMK4JYYK+wDjJF1q99KK4JLL+egWS4nmNqdlMUgCExMZ+iZW0hGhyC3VEbsPjvaN0TBhW//VIFwBrk8sEiw==
 
-"@babel/core@^7.1.0", "@babel/core@^7.12.3", "@babel/core@^7.13.16", "@babel/core@^7.14.0", "@babel/core@^7.18.5", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
```
rurikoaraki, line 331: I wouldn't expect any yarn.lock changes with your changes. I'd make sure that you're up to date on main and that the dependency versions you're referencing are up to date (i.e. the framework package is on 0.8.8)

lenahong:
> Can we add E2E tests along with this PR? Or a subsequent one? Just want to make sure we have baseline testing in to catch regressions as it's iterated on.
> 
> It should be fairly quick to implement. Basically copy/paste from the previous Link's E2E tests, just updated constants/testIDs

Done!

apps/fluent-tester/src/TestComponents/Link/LinkTest.tsx
lenahong: oh this isn't testing the new component. I think I was trying web debugging a scenario. I'll delete this bit. All the testing for the new Link should be under Experimental Link

apps/fluent-tester/src/E2E/LinkExperimental/specs/Link.spec.win.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+});
+
+describe('Link Accessibility Testing', () => {
+  it('Link - Validate accessibilityRole is correct', async () => {
+    await ExperimentalLinkPageObject.scrollToTestElement();
+    await ExperimentalLinkPageObject.waitForPrimaryElementDisplayed(PAGE_TIMEOUT);
+
+    await expect(await ExperimentalLinkPageObject.getAccessibilityRole()).toEqual(LINK_A11Y_ROLE);
+    await expect(await ExperimentalLinkPageObject.didAssertPopup()).toBeFalsy(ExperimentalLinkPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped u
```
rurikoaraki, line 34: super nit: typo on "up" in the comment

Same in the next test

rurikoaraki:
Suggestion for future PRs - this is a lot to review at once, it'd be easier as a reviewer if this were split into smaller changes.

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+      focusable: focusable ?? !isDisabled,
+      ...{ cursor: isDisabled ? 'not-allowed' : 'pointer' },
+      ref: useViewCommandFocus(componentRef),
+      tooltip: linkTooltip,
+    },
+    state: newState,
+  };
+};
+
+/* Copied from useButton. Move to sharable location */
```
rurikoaraki, line 77: +1 on this

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+  );
+
+  const linkTooltip = tooltip ?? url ?? undefined;
+
+  return {
+    props: {
+      ...rest,
+      ...onKeyUpProps,
+      ...pressable.props, // allow user key events to override those set by us
+      accessible: true,
```
rurikoaraki, line 63: Might want to make it so that the client can change this if they want to - not that it'd happen frequently

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+      accessibilityRole: 'link',
+      ...(Platform.OS === (('win32' as any) || 'windows') && { onAccessibilityTap: onAccTap }),
```
rurikoaraki, line 65: I didn't realize onA11yTap should be windows only, I think with other components we usually pass it in regardless of OS. Is it causing problems on a different component?

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+      accessibilityState: getAccessibilityState(isDisabled, accessibilityState),
+      enableFocusRing: enableFocusRing ?? true,
+      focusable: focusable ?? !isDisabled,
+      ...{ cursor: isDisabled ? 'not-allowed' : 'pointer' },
```
rurikoaraki, line 69: Does this need to be wrapped in its own object?

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+    accessibilityState,
+    componentRef = defaultComponentRef,
+    disabled,
+    enableFocusRing,
+    focusable,
+    ...rest
+  } = props;
+  const isDisabled = !!disabled;
+
+  const [visitedState, setVisitedState] = React.useState({ visited: false });
```
rurikoaraki, line 25: I think this can be a bool instead of an object that has a bool as a property

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+      e.stopPropagation();
+    },
+    [setVisitedState, url, onPress],
+  );
+
+  // GH #1336: Set focusRef to null if link is disabled to prevent getting keyboard focus.
+  const focusRef = isDisabled && !focusable ? null : componentRef;
+  const onPressWithFocus = useOnPressWithFocus(focusRef, linkOnPress);
+  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onPressWithFocus });
```
rurikoaraki, line 42: Saad's been working on switching things to use Pressable/usePressableState. Is that possible with Link or not really because of the Text aspect?

rurikoaraki: As an example: https://github.com/microsoft/fluentui-react-native/pull/2156/files#diff-209b8ee779e86a23bc406c0d0dff9ab5275c85ff6a9197095fafdc89e6ea24a3

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,84 @@
+import * as React from 'react';
+import { useAsPressable, useKeyProps, useOnPressWithFocus, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import { LinkProps, LinkState } from './Link.types';
+import { memoize } from '@fluentui-react-native/framework';
+import { AccessibilityState, Linking, Platform } from 'react-native';
+
+export const useLink = (props: LinkProps): LinkState => {
+  // attach the pressable state handlers
```
rurikoaraki, line 8: I don't think this comment makes sense here, maybe where we call useAsPressable?

packages/experimental/Link/src/__tests__/Link.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+it('Link all props', () => {
+  const tree = renderer.create(<Link url="https://www.bing.com">Link to Bing</Link>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it('Link with child', () => {
+  const tree = renderer.create(<Link url="https://www.bing.com">Link to bing</Link>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
rurikoaraki, line 13: Is adding rerender tests in a future PR?

packages/experimental/Link/src/__tests__/Link.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Link } from '..';
+import * as renderer from 'react-test-renderer';
+
+it('Link all props', () => {
+  const tree = renderer.create(<Link url="https://www.bing.com">Link to Bing</Link>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it('Link with child', () => {
+  const tree = renderer.create(<Link url="https://www.bing.com">Link to bing</Link>).toJSON();
```
rurikoaraki, line 11: These tests don't differ in content, was the intent to put something different here?

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,101 @@
...
+  url?: string;
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type IWithLinkOptions<T extends object> = LinkOptions & IWithPressableOptions<T>;
+
+/**
+ * Link props, extending Text props with Pressable options
+ */
+export interface LinkCoreProps extends IWithLinkOptions<TextProps> {
```
rurikoaraki, line 59: I don't see LinkCoreProps used in a different file, is there a reason to keep it separate from LinkProps?

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,101 @@
...
+   * link appearance
+   */
+  appearance?: LinkAppearance;
+  /*
+   * link inline
+   */
+  inline?: boolean;
+}
+
+export type LinkHooks<T extends object> = {
```
rurikoaraki, line 86: I feel like this could be collapsed into LinkState = {
props: IWithPressableEvents<LinkProps & React.ComponentPropsWithRef<any>>
state: ILinkState
}

The template doesn't seem necessary if we'll only have one isntance of this interface

packages/experimental/Link/src/Link.tsx
```diff
@@ -0,0 +1,53 @@
...
+    root: View,
+    content: Text,
+  },
+  useRender: (userProps: LinkProps, useSlots: UseSlots<LinkType>) => {
+    const link = useLink(userProps);
+    // grab the styled slots
+    const Slots = useSlots(userProps, (layer) => linkLookup(layer, link.state, userProps));
+    // now return the handler for finishing render
+    return (final: LinkProps, ...children: React.ReactNode[]) => {
+      // pull onPress out to let Link's Pressable deal with all press events
```
rurikoaraki, line 41: I don't fully understand what this is doing, do you mind explaining in more detail how this works when this Link doesn't have a Pressable?

packages/experimental/Link/src/Link.styling.ts
```diff
@@ -0,0 +1,36 @@
...
+  tokens: [defaultLinkTokens, linkName],
+  states: linkStates,
+  slotProps: {
+    root: buildProps(
+      (tokens: LinkTokens) => ({
+        style: {
+          alignSelf: tokens.alignSelf,
+        },
+      }),
+      [],
```
rurikoaraki, line 18: the cache key needs to have all tokens that affect the function in it. In this case, alignSelf should be part of the cache key

packages/experimental/Link/src/useLink.ts
lenahong: onA11y tap should be fine on other platforms! I must have been considering onKeyDown for leaving out other platforms.

packages/experimental/Link/src/Link.types.ts
lenahong: Consolidating them makes sense. Originally I was thinking about separating out inline vs block Links, but I think I ended up not doing that

packages/experimental/Link/src/useLink.ts
lenahong: Can I make this change in a separate PR along with the pressable. I might have to change how Link handles pressable  events with the changes in Text.

packages/experimental/Link/src/Link.tsx
lenahong: I had to prevent passing in onPress onto Text slot because it was getting called twice as an inline link.

lenahong: This may also change with my upcoming pr with pressable text.

packages/experimental/Link/src/__tests__/Link.test.tsx
lenahong: yes, adding another example her

packages/experimental/Link/src/useLink.ts
lenahong: [https://office.visualstudio.com/OC/_sprints/taskboard/CXE%20React/OC/CY22Q3%20(Jul-Sep)/2208-Aug/Aug02?workitem=6894889](work item for switching Text and Link to Pressable)

rurikoaraki: Yeah for sure

rurikoaraki: @saadnajmi FYI

packages/experimental/Link/src/Link.tsx
rurikoaraki: Oh okay so pressable here is native, not the RN Pressable

apps/fluent-tester/src/TestComponents/LinkExperimental/LinkTest.tsx
```diff
@@ -0,0 +1,225 @@
...
+import { Alert, View, StyleSheet, Text, TextInput } from 'react-native';
+import { Link, LinkTokens } from '@fluentui-react-native/experimental-link';
+import { Stack } from '@fluentui-react-native/stack';
+import { stackStyle, commonTestStyles as commonStyles } from '../Common/styles';
+import { EXPERIMENTAL_LINK_TESTPAGE } from './consts';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { LinkE2ETest } from './E2ELinkTest';
+
+const DefaultLinks: React.FunctionComponent = () => {
+  const doPress = (): void => {
```
rurikoaraki, line 11: These functions should be wrapped in React.useCallback

packages/utils/interactive-hooks/src/getAccessibilityState.ts
```diff
@@ -0,0 +1,10 @@
+import { memoize } from '@fluentui-react-native/framework';
+import { AccessibilityState } from 'react-native';
+
+export const getAccessibilityState = memoize(getAccessibilityStateWorker);
```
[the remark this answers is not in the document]
rurikoaraki: Also follow up PR to generalize this a bit more so we can apply it to Menu/Checkbox/etc.

packages/experimental/Link/src/useLink.ts
```diff
@@ -0,0 +1,79 @@
...
+      } else if (onPress) {
+        onPress(e);
+      }
+      e.stopPropagation();
+    },
+    [setVisitedState, url, onPress],
+  );
+
+  // GH #1336: Set focusRef to null if link is disabled to prevent getting keyboard focus.
+  const focusRef = isDisabled && !focusable ? null : componentRef;
```
rurikoaraki, line 45: Probably worth separating this out into its own hook too, since the logic just keeps being reused. Not blocking.

packages/utils/interactive-hooks/src/index.ts
```diff
@@ -61,3 +61,4 @@ export { preferKeyDownForKeyEvents, useKeyCallback, useKeyDownProps, useKeyProps
 export type { KeyCallback, KeyPressProps } from './useKeyProps';
 export { useOnPressWithFocus } from './useOnPressWithFocus';
 export type { OnPressCallback, OnPressWithFocusCallback } from './useOnPressWithFocus';
+export { getAccessibilityState, getAccessibilityStateWorker } from './getAccessibilityState';
```
rurikoaraki, line 64: You shouldn't need to export the worker, getAccessibilityState should be all that needs to be referenced.

rurikoaraki: Unless you're exporting it for client extensibility?

packages/experimental/Link/src/__tests__/Link.test.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { Alert } from 'react-native';
+import { Link } from '..';
+import * as renderer from 'react-test-renderer';
+
+it('Default Link', () => {
```
rurikoaraki, line 6: Rerender tests should also be part of the test suite, in case adding all the tests was part of the aim of the PR

packages/experimental/Link/src/LinkTokens.ts
```diff
@@ -0,0 +1,66 @@
...
+export const defaultLinkTokens: TokenSettings<LinkTokens, Theme> = (t: Theme) =>
+  ({
+    color: t.colors.link,
+    alignSelf: 'flex-start',
+    variant: 'body1',
+    inline: {
+      textDecorationLine: 'underline',
+    },
+    disabled: {
+      color: t.colors.disabledBodyText,
```
rurikoaraki, line 14: I think this is an old token. We should be using only Fluent2 alias tokens for colors for the new components.

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,97 @@
+import { TextTokens, TextProps } from '@fluentui-react-native/text';
+import { IFocusable, IPressableState, IWithPressableEvents, IWithPressableOptions } from '@fluentui-react-native/interactive-hooks';
+import { ViewProps } from 'react-native';
+
+export const linkName = 'Link';
+
+export type ILinkState = IPressableState & {
```
rurikoaraki, line 7: naming nit: We're moving away from using "I" in front of interfaces

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface LinkOptions {
+  /**
+   * The URL that is opened when the link is clicked.  This value supersedes the 'onPress' callback when both are present.
+   * @default undefined
+   */
+  url?: string;
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type IWithLinkOptions<T extends object> = LinkOptions & IWithPressableOptions<T>;
```
rurikoaraki, line 54: What's the reason for having this as a separate type? It doesn't seem to be used anywhere else

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export interface LinkOptions {
+  /**
+   * The URL that is opened when the link is clicked.  This value supersedes the 'onPress' callback when both are present.
+   * @default undefined
+   */
+  url?: string;
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
```
rurikoaraki, line 53: What's the "extends object" needed for? I think just "T" might be ok here

packages/experimental/Link/src/Link.types.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @default false
+   */
+  inline?: boolean;
+  /**
+   * Specifies whether the link is subtle.
+   * @default false
+   */
+  subtle?: boolean;
+};
+/**
```
rurikoaraki, line 29: nit: for readability can we have a line break between interfaces/blocks?

packages/experimental/Link/src/Link.types.ts
rurikoaraki: Ah, I see that the problem is the name clash below.

I think we can name this LinkState and the interface below something like "LinkInfo" since the object below has both props and state

packages/experimental/Link/src/Link.tsx
rurikoaraki: Also does that mean that onPress should be passed along in the non-inline case?

packages/experimental/Link/src/Link.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * @param state The current state of the link
+ * @param userProps The props that were passed into the link
+ * @returns Whether the styles that are assigned to the layer should be applied to the link
+ */
+export const linkLookup = (layer: string, state: ILinkState, userProps: LinkProps): boolean => {
+  return (
+    state[layer] ||
+    userProps[layer] ||
+    layer === userProps['appearance'] ||
+    (layer === 'inline' && userProps.inline) ||
```
rurikoaraki, line 23: nit: the `userProps[layer]` should take care of the inline and disabled lines in this function.

packages/experimental/Link/src/LinkTokens.ts
rurikoaraki: This is the main blocker for this PR

packages/experimental/Link/src/__tests__/Link.test.tsx
rurikoaraki: Rerender tests being part of the test suite is a component v1 requirement. You can see other examples of the tests in other components, like [here](https://github.com/microsoft/fluentui-react-native/blob/3096f1d203c987c89a09fbb37642aa4ad85d3c5a/packages/components/Button/src/Button.test.tsx#L62).

rurikoaraki (approved):
Offline agreed to handle larger comments in separate PRs, rest of comments are nits/not blocking

packages/experimental/Link/src/LinkTokens.ts
```diff
@@ -0,0 +1,66 @@
...
+      textDecorationLine: 'underline',
+    },
+    disabled: {
+      color: t.colors.neutralForegroundDisabled,
+      textDecorationLine: 'none',
+      focused: {
+        borderColor: t.colors.focusBorder,
+      },
+    },
+    hovered: {
```
rurikoaraki, line 20: I think linkHovered/linkPressed/link/link might all be old color names too? everything else looks like a new alias token though.

rurikoaraki: I want to say it's likely the token you're looking for is brandForegroundLink/Hover/Pressed? But I'd check redlines

packages/experimental/Link/src/LinkTokens.ts
lenahong: Updated to use the new colors

packages/experimental/Link/src/Link.tsx
lenahong: Even for non-inline cases, onPress gets handled on the JS side so everything seems to work fine with the current implementation.

lenahong: I rephrased the comment.

packages/utils/interactive-hooks/src/getAccessibilityState.ts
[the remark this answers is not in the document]
lenahong: Changed to hook.

packages/experimental/Link/src/Link.types.ts
lenahong: removed!

PR #2096, opened 2022-09-07
Shadow: Restrict child to only take type View

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is a follow up to investigation that happened here: https://github.com/microsoft/react-native-macos/pull/1402

After fixing a bug related to shadow consistency in react-native-macOS, one of the tests in react-native-macOS started failing. The issue seemed to be with the test itself (and its still unclear why the test only started failing after the bug fix) - the test was setting shadow props (ex. shadowColor, shadowOpacity, shadowOffset, shadowRadius) on a Text element, when [Text actually has its own specific shadow props](https://reactnative.dev/docs/text-style-props#textshadowcolor) - i.e. textShadowColor, textShadowOffset, textShadowRadius (interestingly textShadowOpacity is not a thing)

I considered updating the Shadow component to also work for Text, but decided against it because Shadows on Text doesn't seem to be an actual use case - usually it's a Shadow on a View that has text on it.

This PR
- updates the jest tests that were doing the same thing as the test in react-native-macOS, the Text should be wrapped in a view and then the Shadow is wrapped around that
- updates the documentation to reflect this change

### Verification

I tried the following code that was in the Shadow jest tests in FluentTester:

```javascript
<Shadow shadowToken={theme.shadows.shadow64}>
    <Text>
        test
    </Text>
</Shadow>
```
and this actually caused a red box to appear (same error as in the react-native-macOS test that started failing)

Screen Shot 2022-09-07 at 12 02 33 PM

Changing the code to this stopped the red box from showing up.
```javascript
<Shadow shadowToken={theme.shadows.shadow64}>
    <View>
        <Text>
            test
        </Text>
    </View>
</Shadow>
```

I would have expected the jest tests to fail on any code that causes a red box, if anyone has any insight on why the jest tests did not catch this please let me know (best guess is because the red box error is coming from native code, and jest tests don't catch those).

No visible changes in the tester.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
```diff
@@ -14,7 +14,9 @@ const TestShadow: React.FunctionComponent<ShadowTestProps> = (props: ShadowTestP
   const theme = useFluentTheme();
   return (
     <Shadow shadowToken={theme.shadows[props.depth]}>
-      <Text>{props.displayText}</Text>
+      <View>
+        <Text>{props.displayText}</Text>
+      </View>
```
Saadnajmi, line 19: For my own sake, could you add a test case where a Shadow wraps a `Pressable` from React Native? That _should_ be a View but I wanna make sure

rurikoaraki (approved):

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -20,6 +20,10 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
       if (childrenArray.length !== 1) {
         console.warn('Shadow must only have one child');
       }
+
+      if (child.type != View) {
+        console.warn('Shadow child must be a view');
+      }
```
Saadnajmi, line 26: Talked offline: As it turns out this won't catch things derived from View, like Pressable. So I think if you put a `Button` here, it would also fail :(

lyzhan7: Yep will remove the warning altogether. Will add jest tests too, we should have tests for Shadow on different types of components.

Saadnajmi: Yeah I think having Shadows wrap a few different components is a good test / example we can add here

lyzhan7: Actually Button doesn't trigger this warning (at least FAB doesn't). Notification does though, so will add tests for both Button and Notification

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
-import { Text } from 'react-native';
+import { Text, View } from 'react-native';
 import { Shadow } from '../Shadow';
 import { useFluentTheme } from '@fluentui-react-native/framework';
 import * as renderer from 'react-test-renderer';
 import { checkRenderConsistency, checkReRender } from '@fluentui-react-native/test-tools';
+import { Notification } from '@fluentui-react-native/notification';
+import { FAB } from '@fluentui-react-native/button';
+import { Pressable } from '@fluentui-react-native/pressable';
```
mischreiber, line 9: Nit: do we care about keeping these imports alphabetized or in any other order?

mischreiber (approved):

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
lyzhan7: I haven't really been alphabetizing imports, the files I've seen don't have the imports alphabetized. Could be something we start doing though, or probably some kind of tool to do this for us

Saadnajmi: Theoretically we should be, but unless prettier imports it I don't think we need to.

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: Removed the tests for Button and Notification because I think they add circular dependencies...should be getting tested in those respective component jest tests though.

PR #2097, opened 2022-09-07
iOS E2E Testing + Improving E2E Code-Reuse

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

This PR does a couple things:

1. Adds iOS E2E Testing and integrates within out CI
2. Improves code-reuse + cross-platformability of our E2E Testing infrastructure. Now, we only need **one page object per component**. This could change in the future when the scope of testing is built out further. We can do this because we can use the same selector cross-platform (I previously didn't think we could), thus, we don't need separate page objects to import separate selectors.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/win32/wdio.conf.js
```diff
@@ -6,7 +6,7 @@ const appArgs = 'basePath ' + path.resolve('dist') + ' plugin defaultplugin bund
 const appDir = path.dirname(require.resolve('@office-iss/rex-win32/rex-win32.js'));
 
 const defaultWaitForTimeout = 20000;
-const defaultConnectionRetryTimeout = 20000;
+const defaultConnectionRetryTimeout = 60000;
```
ksiler, line 9: Is there a reason for the time increase?

ksiler:
I noticed the pipeline has a couple of tasks called "Copy non-Win32 Appium Reports/tree dump screenshots" when copying reports/screenshots for platforms that aren't win32 (in [e2e-publish-artifacts.yml](https://github.com/microsoft/fluentui-react-native/blob/main/.ado/templates/e2e-publish-artifacts.yml#L14)). Could we make this "Copy ${{ parameters.applicationType }} Appium Reports/tree dump screenshots" instead?

Awesome work by the way! Glad to see the added improvements.

samuelfreiberg:
> I noticed the pipeline has a couple of tasks called "Copy non-Win32 Appium Reports/tree dump screenshots" when copying reports/screenshots for platforms that aren't win32 (in [e2e-publish-artifacts.yml](https://github.com/microsoft/fluentui-react-native/blob/main/.ado/templates/e2e-publish-artifacts.yml#L14)). Could we make this "Copy ${{ parameters.applicationType }} Appium Reports/tree dump screenshots" instead?
> 
> Awesome work by the way! Glad to see the added improvements.

Yes, definitely! Just pushed the changes, thanks for the improvement!

apps/win32/wdio.conf.js
samuelfreiberg: I've noticed the first test had to re-run a couple times over the past week due to a failure, so I think increasing this timer will give WDIO more time to connect to the app

ksiler (approved):

Saadnajmi:
:o iOS E2E tests work now!

I guess the ".apple" naming scheme didn't work?

samuelfreiberg:
> :o iOS E2E tests work now!
> 
> I guess the ".apple" naming scheme didn't work?

Yes, it's working!

The reason I couldn't do the .apple convention like I mentioned is because iOS has a different way of scrolling than macOS (because it's mobile). So they couldn't have the same .spec file (which is why I used .ios and .macos again).

Saadnajmi (approved):

chiuam (approved):

Saadnajmi:
Would it be possible to not need these two "copy win32" checks?

image

samuelfreiberg:
> Would it be possible to not need these two "copy win32" checks?
> 
> image

Fixed :)

PR #2103, opened 2022-09-09
Button android ripple

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

**Added ripple effect for button on Android -**
Ripple effect is taken care by React Native's Pressable component. Using it however caused Fluent RN's Fork of Pressable to not be hit anymore, this caused the state of button ('pressed', 'hovered', 'focused') to not change anymore and the associated styles were not being applied.

**Changes made -**
Added a 'Ripple' Pressable layer that provides the animation 
(Did not replace the 'Root' view since Pressable needed to be wrapped with a View to handle [curved borders](https://stackoverflow.com/questions/63048178/ripple-effect-leaking-at-corners-as-if-pressable-button-has-a-borderradius))
Split styles and props as required between Root and Pressable Layer
Added a function to separate out margins from style to be applied onto the wrapper view
Added a hook ripplePressed to trigger styling change onPress
 
**Resolutions tried -**
Tried to apply "useAsButton" props directly on the Pressable component / call functions associated with state directly, had no effect.
Tried to change the state of button explicitly by calling the setPressed function defined in useAsPressable in the onPress of the new Pressable layer. But it did not cause the Button to re-render even after changing the state value.
Attempted to add the ripple animation using the [Animated](https://reactnative.dev/docs/animated) library provided by RN, but abandoned the approach due to excessive complexity 
  

### Verification

| ezgif com-gif-maker | after-ezgif com-gif-maker |

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/components/Button/src/Button.styling.android.ts
```diff
@@ -0,0 +1,116 @@
...
+  'hasIconBefore',
+  'rounded',
+  'circular',
+  'square',
+  'focused',
+  'pressed',
+  'disabled',
+];
+
+export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, ButtonTokens> = {
```
rurikoaraki, line 31: I would just fork the functions that are different. You don't need to copy things that don't change - if they don't exist in the forked file they'll get picked up from the main file correctly.

packages/components/Button/src/Button.tsx
```diff
@@ -70,17 +97,46 @@ export const Button = compose<ButtonType>({
...
+      const hasRipple = Platform.OS === 'android';
+      if (hasRipple) {
+        const [extractedMargin, styleWithoutMargin] = extractMarginAndroid(mergedProps.style);
+        return (
+          <Slots.root style={extractedMargin}>
+            <Slots.ripple
+              accessibilityLabel={label}
+              {...mergedProps}
+              style={styleWithoutMargin}
+              onPressIn={() => {
```
rurikoaraki, line 120: This should be a memoized function. Same with the function that's passed into onPressOut

However, don't use hooks, use memoize directly. We can't use hooks in the inner render as the inner render is conditionally called.

rurikoaraki: I'm guessing this will cause our rerender tests to fail, so I would run our unit tests using yarn test

packages/components/Button/src/FAB/FABCore.tsx
```diff
@@ -65,20 +66,54 @@ export const FAB = compose<FABType>({
...
+            <Slots.root {...mergedProps} accessibilityLabel={label}>
+              {buttonContent}
+            </Slots.root>
+          </Slots.shadow>
+        );
+      } else if (hasRipple) {
+        const [extractedMargin, styleWithoutMargin] = extractMarginAndroid(mergedProps.style);
+        return (
+          <Slots.root style={extractedMargin}>
+            <Slots.ripple
```
Saadnajmi, line 94: There is some duplicated logic between extracting the margin in the shadow slot and here in the FAB. I spoke offline about this briefly. My proposal is this: Could we do something where we extract the "RootWithRipple" into it's own slot. I think as is right now, this logic gets very complicated, especially if we need to duplicate it for any components that need either a shadow or ripple

packages/components/Button/src/FAB/FAB.types.ts
```diff
@@ -1,13 +1,20 @@
...
 export const fabName = 'FAB';
 
 export interface FABSlotProps extends ButtonSlotProps {
   shadow?: ShadowProps;
+  ripple?: PressableProps;
 }
+
+export interface FABProps extends ButtonCoreProps {
+  showChildren?: boolean;
+}
```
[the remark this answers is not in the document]
ayush547: Added this to provide the functionality of collapsible FAB where content can be hidden and shown as required, have added a sample on the tester app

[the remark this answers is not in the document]
rurikoaraki: Could that be a separate PR from adding the ripple? The ripple addition is a fairly big change and it would be easier to review new functionality if one PR is focused on one addition at a time instead of having them be in the same PR.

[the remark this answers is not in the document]
ayush547: Alright, will separate the addition of showChildren on FAB and the ripple changes on FAB and Button

packages/components/Button/src/FAB/FABCore.tsx
```diff
@@ -65,20 +66,54 @@ export const FAB = compose<FABType>({
...
-      const fabWithoutShadow = (
-        <Slots.root {...mergedProps} accessibilityLabel={label}>
+      const buttonContent = (
+        <React.Fragment>
           {icon && <Slots.icon {...iconProps} />}
-          {React.Children.map(children, (child) =>
-            typeof child === 'string' ? <Slots.content key="content">{child}</Slots.content> : child,
-          )}
-        </Slots.root>
+          {showChildren &&
```
rurikoaraki, line 72: Is it possible to take out the showChildren logic out of this PR now that it's in another one?

packages/components/Button/src/Button.styling.ts
```diff
@@ -32,22 +32,59 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
+              display: 'flex',
+              alignItems: 'center',
+              flexDirection: 'row',
+              alignSelf: 'flex-start',
+              justifyContent: 'center',
+              width: tokens.width,
+              ...layoutStyles.from(tokens, theme),
+            },
+            android_ripple: {
+              color: 'buttonbackgroundpressed',
```
rurikoaraki, line 63: This references an old semantic color. Could we use one of the new tokens instead?

packages/components/Button/src/Button.styling.ts
```diff
@@ -32,22 +32,59 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
+            },
+          }),
+          ['backgroundColor', 'width', ...layoutStyles.keys],
```
rurikoaraki, line 66: backgroundColor shouldn't be needed in this cache key array

packages/components/Button/src/Button.tsx
```diff
@@ -34,11 +34,35 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
+    'marginEnd',
+    'marginVertical',
+    'marginHorizontal',
+  ];
+  const extractedMargin = {},
+    mask = {};
+  marginKeys.forEach((key) => {
+    if (style && style[key]) {
+      extractedMargin[key] = style[key];
+      mask[key] = 0;
```
rurikoaraki, line 54: I feel like undefined would be better instead, since that's more equivalent to "this property isn't set"

packages/components/Button/src/Button.tsx
```diff
@@ -70,17 +94,46 @@ export const Button = compose<ButtonType>({
...
           {shouldShowIcon && iconPosition === 'after' && <Slots.icon {...iconProps} />}
-        </Slots.root>
+        </React.Fragment>
       );
+
+      const hasRipple = Platform.OS === 'android';
+      if (hasRipple) {
+        const [extractedMargin, styleWithoutMargin] = extractMarginAndroid(mergedProps.style);
+        return (
+          <Slots.root style={extractedMargin}>
```
rurikoaraki, line 112: Can you add a comment here saying Pressable needs to be wrapped in a View to handle rounded corners?

packages/components/Button/src/Button.tsx
```diff
@@ -70,17 +94,46 @@ export const Button = compose<ButtonType>({
...
             typeof child === 'string' ? <Slots.content key="content">{child}</Slots.content> : child,
           )}
           {shouldShowIcon && iconPosition === 'after' && <Slots.icon {...iconProps} />}
-        </Slots.root>
+        </React.Fragment>
       );
+
+      const hasRipple = Platform.OS === 'android';
+      if (hasRipple) {
+        const [extractedMargin, styleWithoutMargin] = extractMarginAndroid(mergedProps.style);
```
rurikoaraki, line 110: Probably should be a memoized function

packages/components/Button/src/Button.styling.ts
```diff
@@ -32,22 +32,60 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
+              justifyContent: 'center',
+              width: tokens.width,
+              ...borderStyles.from(tokens, theme),
+              ...layoutStyles.from(tokens, theme),
+            },
+            android_ripple: {
+              color: tokens.pressed.backgroundColor,
+            },
+          }),
+          ['pressed', 'width', ...layoutStyles.keys, ...borderStyles.keys],
```
ValentinaKozlova, line 67: Not sure if caching will work for `pressed.backgroundColor` as we have here only 'pressed' passed to the array. @rurikoaraki , do you know if this syntax will work?

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
+    'marginHorizontal',
+    'start',
+    'end',
+    'left',
+    'right',
+    'top',
+    'bottom',
+  ];
+  const outerStyleProps = {},
+    resetProps = {};
```
ValentinaKozlova, line 56: I think for better readability it's good to have `const resetProps = {};` on a new line

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
+  marginKeys.forEach((key) => {
+    if (style && style[key]) {
+      outerStyleProps[key] = style[key];
+      resetProps[key] = 0;
+    }
+  });
+  return [
+    { ...outerStyleProps, display: style ? style.display : 'flex' },
```
ValentinaKozlova, line 64: Do you need to set `display` explicitly? Shouldn't it be 'flex' by default?

ayush547: Have added this to handle the case where `display: 'none'` is used to hide a component, handling nothing sent with 'flex'

packages/components/Button/src/Button.tsx
rurikoaraki: Yeah we generally don't instantiate multiple variables in one statement.

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
+    'start',
+    'end',
+    'left',
+    'right',
+    'top',
+    'bottom',
+  ];
+  const outerStyleProps = {},
+    resetProps = {};
+  marginKeys.forEach((key) => {
```
rurikoaraki, line 57: Please insert newlines around blocks of code like this for readability. Generally I add newlines around function calls, before return statements, and after a block of declaring and instantiating variables

packages/components/Button/src/Button.tsx
rurikoaraki: For native it's usually better to hide a component by not rendering it rather than hiding it with display: 'none', since I don't think that guarantees that the component will be hidden from screen readers.

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
+    if (style && style[key]) {
+      outerStyleProps[key] = style[key];
+      resetProps[key] = 0;
```
rurikoaraki, line 60: I think "undefined" is a better "reset".

Also it seems like it'd be cleaner to do something like

```
const {<all outer props>, ...restStyles} = style;

return [
   {<all outer props>},
   restStyles
]
```

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
   );
 };
 
+export const extractOuterStylePropsAndroid = memoize((style) => {
```
rurikoaraki, line 37: nit: for better typesafety it'd be nice to define the return type of this function.

packages/components/Button/src/Button.tsx
```diff
@@ -34,19 +34,56 @@ export const buttonLookup = (layer: string, state: IPressableState, userProps: B
...
   ...stylingSettings,
   slots: {
     root: View,
+    ripple: Pressable,
     icon: Icon,
     content: Text,
   },
   useRender: (userProps: ButtonProps, useSlots: UseSlots<ButtonType>) => {
     const button = useButton(userProps);
+    const [ripplePressedAndroid, setRippleStateAndroid] = React.useState(false);
```
rurikoaraki, line 80: I think we can put this in useButton instead of putting it in this file so that all the hooks are in one place.

packages/components/Button/src/Button.tsx
rurikoaraki, line 37: I think it'd be nice for this to be in its own file which has a forked impl for Android vs. other platforms, so that this Android-only logic only gets bundled into bundles that run Android.

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
```diff
@@ -26,18 +26,11 @@ exports[`Custom FAB with no shadow(iOS) 1`] = `
   onStartShouldSetResponder={[Function]}
   style={
     Object {
-      "alignItems": "center",
-      "alignSelf": "flex-start",
+      "alignSelf": "baseline",
```
rurikoaraki, line 29: The align-self change seems like an unintended change?

rurikoaraki: Also all of these tests are meant to test iOS - you should check with Lynn if this is okay or if you need to be standing up separate tests for Android.

packages/components/Button/src/Button.tsx
ayush547: Added this to make sure that if used, `display: 'none'` works as intended, without this the Outer View would still be rendered taking up space while the Inner Views would not be displayed

packages/components/Button/src/Button.tsx
ayush547: Had done that initially, but then useButton is shared by Compound Button and Toggle Button as well and would unnecessarily impact them. 
Also, based on discussions with @Saadnajmi on his work on migrating to [usePressableState](https://github.com/microsoft/fluentui-react-native/pull/2153), we feel that this hook work around should be a temporary workaround, especially since usePressableState directly allows us to use androidRipple and works better by changing the state (and colors) of button even on a short tap (only on long press color change occurs), while useAsPressable does not do that.

packages/components/Button/src/Button.tsx
ayush547: Hoping for this function to be used for Shadows as well, do you think it can be placed somewhere more appropriate

lyzhan7: Shadow has its own version of this (i.e. logic splitting inner/outer props) - it's not exactly a function the way you have it, it's kind of just built into the component: https://github.com/microsoft/fluentui-react-native/blob/main/packages/experimental/Shadow/src/Shadow.tsx.

Saad has brought up that we should find a way to consolidate this logic (I think he might have mentioned this somewhere else in this PR) but his suggestion was to make a rootWithShadow component, or as I understand it just a general component that has an inner and outer view that can be customized, and have Shadow and ripple both use that. I can take on this work but not sure when I'll be able to get to it.

packages/components/Button/src/FAB/FABCore.tsx
```diff
@@ -64,21 +69,54 @@ export const FAB = compose<FABType>({
...
       const hasShadow = Platform.OS === 'ios';
+      const hasRipple = Platform.OS === 'android';
+
       if (hasShadow) {
-        return <Slots.shadow>{fabWithoutShadow}</Slots.shadow>;
+        return (
+          <Slots.shadow>
+            <Slots.root {...mergedProps} accessibilityLabel={label}>
+              {buttonContent}
+            </Slots.root>
```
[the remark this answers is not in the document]
lyzhan7: Yes the FAB shadow should always be shown on iOS

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Yeah the FAB tests specifically currently have the Platform check mocked out to always return 'ios', so they wouldn't be testing the ripple. I think you will need to add some of your own tests where the Platform check isn't returning 'ios' (either it should be mocked out to return 'android' or I think if the default if nothing is specified is actually to test Android). If we can keep the tests in the same file that seems better to me than splitting it into a FAB ios and FAB android test file, since I think we would want Shadow on Android at some point (not sure if we'd want ripple on iOS) but I don't have a strong opinion about this.

About the changes to the snapshot: I'm not sure if the changes are okay or not, but I can help you test that these changes haven't impacted the iOS fab in the tester - feel free to message me on teams.

packages/components/Button/src/Button.tsx
ayush547: Based on how things are currently, the structure for the Button on Android is currently -
`<View><Pressable>{Inner Stuff}</Pressable></View>` 
While for Button on iOS for the shadow, it is -
`<View><View>{Inner Stuff}</View></View>` so it would not be possible to do this rn.

But! This can be done once @Saadnajmi 's PR of moving to [usePressableState](https://github.com/microsoft/fluentui-react-native/pull/2153) goes through (It will enable us to use `<View>` directly instead of `<Pressable>` to provide the ripple effect)

packages/components/Button/src/Button.tsx
ayush547: I was wondering if it would be fine to repeat the keys twice but turns out it's actually much much cleaner, going for this!

packages/components/Button/src/FAB/FABCore.tsx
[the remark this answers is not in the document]
ayush547: Moving `<Slots.root {...mergedProps} accessibilityLabel={label}>
              {buttonContent}
            </Slots.root>` into another const called `buttonContentWithRoot`

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
ayush547: Hey @lyzhan7,
I don't think 
```
beforeAll(() => {
  jest.mock('react-native/Libraries/Utilities/Platform', () => ({
    OS: 'ios',
    select: () => null,
  }));
});
```
 is working as intended, I set the FAB.ios.tsx file to return null but the test still ran successfully! 
I think it is running with Android as platform, which also explains why the ripple addition caused this snapshot to change

packages/components/Button/src/Button.styling.ts
rurikoaraki: Hm I'm worried about the caching being invalidated more frequently because it would be comparing whether `pressed` has been assigned a new object instead of just a primitive value. Unfortunately the cache won't allow pressed.backgroundColor since that's not part of the ButtonTokens type.

I think for the purposes of making sure the token system works properly, the ripple color should be its own token.

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Hmm okay just to check, is `Platform.OS === 'android'` on line 83 in FABCore.tsx returning true? If it's returning true that's unexpected & I'll look into that. I see also in your code you've introduced `Platform.select({android:...}`. I'm not familiar with Platform.select, but I think the tests might need to mock out that method as well to keep all of the Android ripple stuff out of iOS (would guess that you add another jest.mock like under the current one, but it mocks out `select` instead of `ios` - something like that)

packages/components/Button/src/Button.styling.ts
```diff
@@ -32,22 +32,60 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
-          justifyContent: 'center',
-          width: tokens.width,
-          backgroundColor: tokens.backgroundColor,
-          ...borderStyles.from(tokens, theme),
-          ...layoutStyles.from(tokens, theme),
-        },
-      }),
-      ['backgroundColor', 'width', ...borderStyles.keys, ...layoutStyles.keys],
-    ),
+    ...Platform.select({
```
Saadnajmi, line 35: I think we should have a `Button.styling.android.ts` file. There's a precedent for that. It'll also get picked up by jest tests, where `Platform.OS / .select` won't without mocking.

packages/components/Button/src/utils.android.ts
```diff
@@ -0,0 +1,48 @@
+import { memoize } from '@fluentui-react-native/framework';
+import { ViewStyle } from '@office-iss/react-native-win32';
```
Saadnajmi, line 2: Don't import this from react-native-win32, do it from `react-native`

packages/components/Button/src/utils.android.ts
```diff
@@ -0,0 +1,48 @@
+import { memoize } from '@fluentui-react-native/framework';
```
Saadnajmi, line 1: Could we rename this file to `extractStyle.ts` or the like, and then instead just have a comment explaining this is used for Android's ripple container?

packages/components/Button/src/FAB/FABCore.tsx
```diff
@@ -64,21 +69,49 @@ export const FAB = compose<FABType>({
...
+              {...mergedProps}
+              style={innerStyleProps}
+              onPressIn={memoize(() => {
+                setRippleStateAndroid(true);
+                mergedProps.onPressIn?.();
+              })}
+              onPressOut={memoize(() => {
+                setRippleStateAndroid(false);
+                mergedProps.onPressOut?.();
+              })}
```
Saadnajmi, line 107: Can we extract this logic to a hook, similar to what we do with `useOnPressWithFocus()` ?

Saadnajmi: Actually, could this setting of `onPressIn/Out` be put in `useButton.ts` ?

packages/components/Button/src/Button.styling.ts
Saadnajmi: I take it back. For styling files it's probably fine. For cleaner code I think I'd rather we do a spread:

```
...(Platform.OS === 'android' && {android_ripple: ...})
```

ayush547: I think Platform.Select might be cleaner here since there is also some code being added for "not Android", this would have to be placed in a separate if statement otherwise

packages/components/Button/src/FAB/FABCore.tsx
ayush547: Had initially placed it in useButton.ts, but it was unnecessarily being also used for ToggleButton and CompoundButton.
Have removed this, shouldn't be required now that we have migrated to RN Pressable.

packages/components/Button/src/Button.tsx
```diff
@@ -5,11 +5,11 @@ import { ActivityIndicator } from '@fluentui-react-native/experimental-activity-
 import { buttonName, ButtonType, ButtonProps } from './Button.types';
 import { TextV1 as Text } from '@fluentui-react-native/text';
 import { stylingSettings, getDefaultSize } from './Button.styling';
-import { compose, mergeProps, withSlots, UseSlots, memoize } from '@fluentui-react-native/framework';
+import { compose, mergeProps, withSlots, UseSlots } from '@fluentui-react-native/framework';
 import { useButton } from './useButton';
 import { Icon } from '@fluentui-react-native/icon';
 import { createIconProps, IPressableState } from '@fluentui-react-native/interactive-hooks';
-import { extractOuterStylePropsAndroid } from './utils.android';
+import { extractOuterStylePropsAndroid } from './ExtractStyle.android';
```
Saadnajmi, line 12: Ah so doing it this way (importing the .android file directly) means it will end up getting included in all platform bundles anyway. At this point I think it's worth just renaming the file

packages/components/Button/src/ExtractStyle.android.ts
```diff
@@ -1,6 +1,7 @@
 import { memoize } from '@fluentui-react-native/framework';
-import { ViewStyle } from '@office-iss/react-native-win32';
+import { ViewStyle } from 'react-native';
 
+// This function is used to seperate out inner and outer styles for the RippleContainer
```
Saadnajmi, line 4: Could you explain _why_ we need to do that? If I was a new dev looking at this file, I may not have known about the whole ripple corner-radius bug and been confused why there's an inner and outer to begin with

packages/components/Button/src/Button.tsx
```diff
@@ -69,18 +72,37 @@ export const Button = compose<ButtonType>({
...
           )}
           {shouldShowIcon && iconPosition === 'after' && <Slots.icon {...iconProps} />}
-        </Slots.root>
+        </React.Fragment>
       );
+
+      const hasRipple = Platform.OS === 'android';
+      if (hasRipple) {
+        const [outerStyleProps, innerStyleProps] = extractOuterStylePropsAndroid(mergedProps.style);
+        return (
```
Saadnajmi, line 91: I still feel like this whole "hasRipple" logic can be extracted to a simple function component / stagedComponent so that it can be re-used across multiple components, rather than copied for every new component that has a ripple. I'm sure that will be most controls.

packages/components/Button/src/Button.styling.ts
Saadnajmi: What "not Android code" is there? As far as I can tell, the `root` slot of both Android and default are the same, minus an extra `ripple_config`. It should be safe to add the ripple config to other Platforms because that's a prop from RN Core. Rather than duplicating the Root slot twice (which could lead to bugs where engineers update the root on win32/macOS/iOS but not Android), I'd rather we just have one root slot and _maybe_ conditionally create the `rippleContainer` slot per platform. Even if we just _always_ created the slotProps for ripple container on all platforms, I don't think that would be too bad.

packages/components/Button/src/FAB/FAB.styling.ts
```diff
@@ -1,34 +1,75 @@
...
-          ...layoutStyles.from(tokens, theme),
-          ...shadowStyles.from(tokens, theme),
-        },
-        elevation: tokens.elevation,
-      }),
-      ['backgroundColor', 'width', 'elevation', ...borderStyles.keys, ...layoutStyles.keys, ...shadowStyles.keys],
-    ),
+    ...Platform.select({
+      android: {
+        rippleContainer: buildProps(
```
Saadnajmi, line 17: Similar comment about platform.select here.

packages/components/Button/src/Button.types.ts
```diff
@@ -154,6 +155,7 @@ export interface ButtonInfo {
 
 export interface ButtonSlotProps {
   root: React.PropsWithRef<PressablePropsExtended>;
+  rippleContainer?: IViewProps;
```
Saadnajmi, line 158: nit but not necessary, I'd leave a simple ` // Android only` comment around rippleContainer.

packages/components/Button/src/Button.styling.ts
ayush547: Alright, will make the change to have only one root

packages/components/Button/src/Button.tsx
ayush547: Based on offline discussions, would like to continue with the current approach for now. When the requirement for the ripple on other components arises, will pick up work on completing the ripple as a component (currently in [draft](https://github.com/microsoft/fluentui-react-native/pull/2273)

packages/components/Button/src/Button.types.ts
```diff
@@ -17,6 +18,11 @@ export interface ButtonCoreTokens extends LayoutTokens, FontTokens, IBorderToken
    */
   iconColor?: ColorValue;
 
+  /**
+   * Ripple color for Android.
+   */
+  rippleColor?: ColorValue;
+
```
Saadnajmi, line 25: Any chance this can move to `IColorTokens` ? Looks like that comes from the file `color-tokens.ts` and you could probably copy what's being done for `IForegroundColorTokens/foregroundColorTokens` and make a `RippleColorTokens`. Or add it to `IBackgroundColorTokens if that's more appropriate. 

Not blocking though, and could be a followup

packages/components/Button/src/ButtonColorTokens.ts
```diff
@@ -8,6 +8,7 @@ export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t:
     color: t.colors.buttonText,
     borderColor: t.colors.buttonBorder,
     iconColor: t.colors.buttonIcon,
+    rippleColor: t.colors.defaultPressedBackground, // Android only
```
Saadnajmi, line 11: Given that there's already a `ButtonColorTokens.<macOS | win32 | windows>.ts, can we make a `ButtonColorTokens.android.ts` ? I know that's more duplicated code, but I personally feel we should either have completely platform separate files, or one file, not a mix.

packages/components/Button/src/ExtractStyle.android.ts
```diff
@@ -0,0 +1,57 @@
...
+      start,
+      end,
+      left,
+      right,
+      top,
+      bottom,
+      display,
+      opacity,
+    },
+    { borderWidth: 0, ...restOfProps },
```
Saadnajmi, line 55: Why is borderWidth: 0? I'm guessing to account for the ripple still?

packages/components/Button/src/FAB/FABColorTokens.ts
```diff
@@ -8,6 +8,7 @@ export const defaultFABColorTokens: TokenSettings<ButtonCoreTokens, Theme> = (t:
   color: t.colors.brandedContent,
   borderColor: t.colors.brandedBorder,
   iconColor: t.colors.brandedIcon,
+  rippleColor: t.colors.defaultPressedBackground, // Android only
```
Saadnajmi, line 11: Similar comment here.

packages/experimental/MenuButton/jest.config.js
```diff
@@ -1,2 +1,2 @@
 const { configureReactNativeJest } = require('@fluentui-react-native/scripts');
-module.exports = configureReactNativeJest('android');
+module.exports = configureReactNativeJest('ios');
```
Saadnajmi, line 2: Why are we changing this back? 
If anything, this should be either win32 or macOS because those are the only platforms where MenuButton is available. But can we do that outside of this review?

Saadnajmi (approved):

packages/components/Button/src/ButtonColorTokens.ts
ayush547: I will be doing this as a part of a separate PR!

packages/components/Button/src/ExtractStyle.android.ts
ayush547: To ensure inside view doesn't have a border width, had initially thought of doing this in the styling, but have placed all the code related to ripple here in one place

packages/experimental/MenuButton/jest.config.js
ayush547: Have changed it to ios to maintain parity across all Button tests, we can change it to win32 when we change all the other tests as well

packages/components/Button/src/Button.types.ts
ayush547: Taking this offline

PR #2105, opened 2022-09-09
Option interactive

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding interactivity state & tokens, and default styling for hovered, pressed, and disabled states.

The changes are:

- Add componentRef (which in retrospect, wasn't necessary, but it will be in the future)
- Add useAsPressable to useOption, and then return pressable props and state.
- Add interaction states and tokens, and add default tokens for those states.
- Include the new states in the lookup function for useSlots, and then merge the output of useOption into the final props

### Verification

option_interaction

### DIFF
--- a/packages/experimental/Dropdown/src/Option/Option.styling.ts
+++ b/packages/experimental/Dropdown/src/Option/Option.styling.ts
@@ -2,6 +2,8 @@ import { borderStyles, buildProps, fontStyles, layoutStyles, Theme, UseStylingOp
 import { optionName, OptionProps, OptionSlotProps, OptionTokens } from './Option.types';
 import { defaultOptionTokens } from './OptionTokens';
 
+export const menuItemStates: (keyof OptionTokens)[] = ['hovered', 'focused', 'pressed', 'disabled'];
+
 export const stylingSettings: UseStylingOptions<OptionProps, OptionSlotProps, OptionTokens> = {
   tokens: [defaultOptionTokens, optionName],
   slotProps: {
--- a/packages/experimental/Dropdown/src/Option/Option.styling.ts
+++ b/packages/experimental/Dropdown/src/Option/Option.styling.ts
@@ -2,23 +2,25 @@ import { borderStyles, buildProps, fontStyles, layoutStyles, Theme, UseStylingOp
 import { optionName, OptionProps, OptionSlotProps, OptionTokens } from './Option.types';
 import { defaultOptionTokens } from './OptionTokens';
 
-export const menuItemStates: (keyof OptionTokens)[] = ['hovered', 'focused', 'pressed', 'disabled'];
+export const optionStates: (keyof OptionTokens)[] = ['hovered', 'focused', 'pressed', 'disabled'];
 
 export const stylingSettings: UseStylingOptions<OptionProps, OptionSlotProps, OptionTokens> = {
   tokens: [defaultOptionTokens, optionName],
+  states: optionStates,
   slotProps: {
     root: buildProps(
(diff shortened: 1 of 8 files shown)


Saadnajmi:
Overall question: any reason this is a new component rather than reusing menu item?

rurikoaraki:
> Overall question: any reason this is a new component rather than reusing menu item?

Dropdown is semantically different from Menu, so that's the reason why there's new components instead of reusing Menu to build Dropdown. In the future Option will have different accessibility settings, etc.

There's probably opportunity to refactor code to share behavior logic between the two, though

packages/experimental/Dropdown/src/Option/useOption.ts
```diff
@@ -1,8 +1,8 @@
...
 
-export const useOption = (_props: OptionProps): OptionState => {
-  return {
-    props: {},
-    state: {},
-  };
+export const useOption = (props: OptionProps): OptionState => {
+  const pressable = useAsPressable(props);
+
+  return pressable;
```
Saadnajmi, line 7: (Feel free to reject this suggestion)
Could we try using Pressable from `react-native` along with `usePressableState` for new components? Super Ok if you'd rather leave that for later.

packages/experimental/Dropdown/src/Option/Option.styling.ts
```diff
@@ -2,21 +2,25 @@ import { borderStyles, buildProps, fontStyles, layoutStyles, Theme, UseStylingOp
 import { optionName, OptionProps, OptionSlotProps, OptionTokens } from './Option.types';
 import { defaultOptionTokens } from './OptionTokens';
 
+export const optionStates: (keyof OptionTokens)[] = ['hovered', 'focused', 'pressed', 'disabled'];
```
Saadnajmi, line 5: It feels like we might or should have a type for these 4 states in particular we can spread? Future fodder if not.. I don't know that would make code cleaner..

Saadnajmi (approved):

samuelfreiberg (approved):

packages/experimental/Dropdown/src/Option/Option.styling.ts
rurikoaraki: I don't know how well that would play with typing. We can try it, I guess a template might make it feasible

packages/experimental/Dropdown/src/Option/useOption.ts
rurikoaraki: Is Pressable ready for use?

Saadnajmi: It should be. The only think lacking from when I last worked on it was types in the DefinitelyTyped repo.

packages/experimental/Dropdown/src/Option/Option.styling.ts
Saadnajmi: I think I was confusing old and new framework, and different types. `useAsPressable.types.ts` has "I<Focus | Press | Hover>State` types. Nevermind..

gulnazsayed (approved):

packages/experimental/Dropdown/src/Option/useOption.ts
rurikoaraki: I think I'll commit this for now so it doesn't get too out of sync from main, and work on the Pressable separately.

PR #2111, opened 2022-09-12
Update shadow jest test

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Updating shadow jest test to unblock pipeline

### DIFF
--- a/change/@fluentui-react-native-experimental-shadow-3b16c1fa-ab69-4770-81be-865706cfafb3.json
+++ b/change/@fluentui-react-native-experimental-shadow-3b16c1fa-ab69-4770-81be-865706cfafb3.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Change the test",
+  "packageName": "@fluentui-react-native/experimental-shadow",
+  "email": "ruaraki@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):

lyzhan7 (approved):

lyzhan7:
Thanks for fixing this - do you have any ideas how this started failing? Seems like it was related to https://github.com/microsoft/fluentui-react-native/pull/2065 but the jest tests were passing then, is there anything I could have done to prevent this?

rurikoaraki:
> Thanks for fixing this - do you have any ideas how this started failing? Seems like it was related to #2065 but the jest tests were passing then, is there anything I could have done to prevent this?

It was a result of adding tests to https://github.com/microsoft/fluentui-react-native/pull/2096 and not updating them in https://github.com/microsoft/fluentui-react-native/pull/2065. Tests passed for both since, at the time of running the code in https://github.com/microsoft/fluentui-react-native/pull/2065, the new test wasn't in its set of tests. I think this case is mostly about being mindful about how your changes might conflict, but stuff like this happens occasionally without people realizing it will happen.

lyzhan7:
^ Never mind sorry just saw your post on Teams. Idk why my Teams was very slow to update in the morning, just started seeing new notifications from this morning now

lyzhan7:
> > Thanks for fixing this - do you have any ideas how this started failing? Seems like it was related to #2065 but the jest tests were passing then, is there anything I could have done to prevent this?
> 
> It was a result of adding tests to #2096 and not updating them in #2065. Tests passed for both since, at the time of running the code in #2065, the new test wasn't in its set of tests. I think this case is mostly about being mindful about how your changes might conflict, but stuff like this happens occasionally without people realizing it will happen.

That makes sense, thanks for explaining! Didn't realize that could happen, will try to be more mindful in the future.

PR #2115, opened 2022-09-12
Changing Location of Appium Driver Dependency Installs

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

Appium v2, requires an extra dependency for each platform's specific driver. When the Appium server starts, it looks in the current folders' _node_modules_ for the driver package. However, in FURN, all packages are downloaded in the roots' _node_modules_ rather than each individual folder. This is causing issues when trying to run E2E tests locally.

The fix is to add the `nohoist` property in the roots' package.json for these specific packages. This makes it so the package gets installed in the _node_modules_ folder where that package.json lives, thereby fixing this issue.

### Verification

Local _ CI

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,12 @@
       "apps/*",
       "packages/**",
       "scripts"
+    ],
+    "nohoist": [
+      "@fluentui-react-native/tester-win32/appium-windows-driver",
+      "@fluentui-react-native/tester/appium-mac2-driver",
+      "@fluentui-react-native/tester/appium-xcuitest-driver",
+      "@fluentui-react-native/tester/appium-windows-driver"
     ]
   },
   "resolutions": {
--- a/.ado/templates/e2e-testing-ios.yml
+++ b/.ado/templates/e2e-testing-ios.yml
@@ -1,9 +1,4 @@
 steps:
-  - script: |
-      yarn appium driver install xcuitest
-    workingDirectory: apps/fluent-tester
-    displayName: 'Installing appium xcuitest driver'
-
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -51,6 +51,12 @@
       "apps/*",
       "packages/**",
       "scripts"
+    ],
+    "nohoist": [
+      "@fluentui-react-native/tester-win32/appium-windows-driver",
+      "@fluentui-react-native/tester/appium-mac2-driver",
+      "@fluentui-react-native/tester/appium-xcuitest-driver",
+      "@fluentui-react-native/tester/appium-windows-driver"
     ]
```
Saadnajmi, line 60: dumb question, could we specify this in `apps/fluent-tester/package.json` instead? Not sure that's actually better but was curious.

Saadnajmi (approved):
Such a simple fix, I love it :)

rurikoaraki (approved):

package.json
rurikoaraki: If nohoist is a property of workspaces I'm guessing not, but quick glance at documentation doesn't say one way or the other

samuelfreiberg: From what I've seen, this is common practice. Rather than having nohoise in multiple different package.json, we have one at the root with all the workspaces

PR #2117, opened 2022-09-13
Getting Shadow working with backgroundColor - margins, padding, borders

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Background: The current Shadow component only works on Apple platforms (iOS/macOS) due to some buggy behavior related to the background color (if no background color is set, then shadow props get set on the child instead of the parent). This behavior happened to work well for the Shadow component, but should not be relied on since Shadows don't work on win32 (and also might have other unwanted impacts on Apple platforms). On apple platforms we can suppress the unwanted behavior by setting a background color. With the behavior suppressed we need to consider all the props that can be set on the child, and determine which view they should be set on.

This PR:
1. Updates the Shadow component to check that the child has a backgroundColor (which by default gets set on both inner/outer views, which then suppresses the buggy Apple behavior) and then splits out any child props that should be set on the inner shadow view only, or outer shadow view only, and does so accordingly
2. Adds a new test page to test Shadows on children that have specific props. Added tests (both on FluentTester and jest) for margin/padding, borderRadius, borderWidth
3. The styling in ShadowDepthTestSection was updated, at some point we stopped needing the theme and also the styles weren't being properly flattened...not really sure how this worked before
4. Had to make a small fix with the Notification component:

| Before the change to the Notification component (but with the other changes in this PR | After the change to the Notification component |
| - | - |
| Screen Shot 2022-09-12 at 7 29 16 PM | Screen Shot 2022-09-15 at 5 43 00 PM |

Future work:
- this PR deals with only some props that needed consideration - margins, padding, and borders, but there are other props that will need to be considered as well. 

Open questions:
- Right now by default, the child style props are getting set on both the inner and outer views. Other options are to set it on the inner view only by default, or outer view only by default. I currently don't see an obvious right answer here (all three of them will have their own exceptions that we need to specify), if anyone has strong opinions here let me know, otherwise thinking of going with this for now until there's a reason to switch to another default (ex. if choosing another default means there are less exceptions we need to specify)
- Based on 3 and 4 above, it kind of seems like margin and padding props can be set anywhere in the props object (doesn't have to be in 'styles') and it will still work. This doesn't seem like expected behavior to me, my understanding is that margin and padding should be getting set in the style object. If we want to support Shadows that work with margin/padding getting set elsewhere this will require additional work.

### Verification

See new section added to the bottom

| Light Mode | Dark Mode |
| - | - |
| Screen Shot 2022-09-12 at 6 50 22 PM | Screen Shot 2022-09-12 at 6 50 30 PM |

Manual testing
- checked the Shadow test page on iOS and macOS
- checked that FAB and Notification look the same as before

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


lyzhan7:
Known issue: yarn test is failing on Shadow - specifically the "Shadow simple rendering does not invalidate styling" and "Shadow re-renders correctly" tests. The last time I saw these two tests fail was because I needed to memoize some stuff, but I've tried various ways of memoizing the new code (ex. memoizing the all the innerShadowProps/outerShadowProps, vs. just the style objects for the innerShadowView/outerShadowView), but haven't managed to get these tests to pass...so thinking it might not be related to memoize - if anyone has any ideas let me know.

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -22,12 +21,95 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
       }
     }
 
-    const childWithKeyShadow = React.cloneElement(child, mergeProps(child.props, { style: shadowTokenStyleSet.key }));
+    const { style: childStyle = {}, ...restOfChildProps } = child.props;
```
rurikoaraki, line 24: the `= {}` creates a new style object, probably invalidating the memoized function below if the child doesn't have a style on it. You should define the empty style as a const outside of the component so that it's constant.

rurikoaraki: You could also leave it `undefined` and change the memoized function to check for undefined before trying to access properties off of it

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -22,12 +21,95 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
...
+
+    const childWithInnerShadow = React.cloneElement(child, innerShadowViewProps);
 
-    return <View {...mergedProps}>{childWithKeyShadow}</View>;
+    return <View {...outerShadowViewProps}>{childWithInnerShadow}</View>;
   };
 });
 
+const getStylePropsForShadowViews = memoize(getStylePropsForShadowViewsWorker);
+function getStylePropsForShadowViewsWorker(childStyleProps: any, shadowTokenStyleSet: any) {
```
rurikoaraki, line 37: `shadowTokenStyleSet` should at least have a well-defined type.

It would be nice if we could not have `any` for childStyleProps, too, but that's probably much harder to accomplish. You could try restricting to `StyleProp<ViewStyle>` but TS might complain

Saadnajmi: There's also the type `ShadowStyleIOS` in the RN-types repo:

```
export interface ShadowStyleIOS {
    shadowColor?: ColorValue | undefined;
    shadowOffset?: { width: number; height: number } | undefined;
    shadowOpacity?: number | undefined;
    shadowRadius?: number | undefined;
}
```

rurikoaraki: I'm pretty sure we have a type defined for shadow styles in our theming package? I just don't remember what it is

rurikoaraki:
The screenshots look like the red is bleeding through a little bit on the rounded corners. Maybe we should align backgroundColor with the child view?

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -22,12 +21,95 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
...
+    paddingVertical,
+
+    ...restOfChildStyleProps
+  } = childStyleProps;
+
+  return {
+    inner: {
+      style: {
+        backgroundColor: 'red', // will not be shown, just needed in macOS/iOS to ensure buggy behaviour is suppressed
+        ...(padding && { padding }),
```
rurikoaraki, line 74: I think you could just pass padding (and the other properties) directly through like

padding,
paddingBottom,
etc.

Is there a reason to need the spread operator + creating an object around every defined style property?

packages/experimental/Shadow/src/Shadow.tsx
Saadnajmi: Not for the actual RN type, we have the "design" type. In the future we can perhaps just have the pipeline convert the "design type" to the RN props so we don't have to know to do stuff like divide blur by 2 on iOS, but 🤷🏾

```
/**
 * Base structure of a shadow style, which has x and y offset, blur amount, and color specified
 */
export interface ShadowValue {
  x: number;
  y: number;
  blur: number;
  color: ColorValue;
}

```

Saadnajmi: But yeah, let's not use `any`

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
```diff
@@ -32,21 +24,12 @@ interface ShadowTestBoxProps {
...
-      <View
-        style={[
-          commonTestStyles.view,
-          themedStyles.effectBox,
-          themedStyles.vmargin,
-          themedStyles.padding,
-          { backgroundColor: backgroundColor },
-        ]}
-      >
+      <View style={mergeStyles(styles.shadowTestBox, { backgroundColor: backgroundColor })}>
```
Saadnajmi, line 32: Nit: Could we move the style creates / merging to its' own line? I think in general, we should strive to not have inline styles in FluentTester (even If this one is probably fine because `mergeStyles` will memoize)

apps/fluent-tester/src/TestComponents/Shadow/ShadowWithDifferentPropsTestSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+});
+
+export const ShadowWithDifferentPropsTestSection: React.FunctionComponent = () => {
+  const theme = useFluentTheme();
+  const themedStyles = getThemedStyles(theme);
+
+  return (
+    <View>
+      <Text style={{ padding: 16 }}>The following tests have a Shadow16 set on a child view that has some specific props.</Text>
```
Saadnajmi, line 28: Nit: inline style

apps/fluent-tester/src/TestComponents/Shadow/ShadowWithDifferentPropsTestSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+      minHeight: 64,
+      padding: 20,
+      marginVertical: 16,
+      marginHorizontal: 32,
+      backgroundColor: t.colors.background,
+    },
+  };
+});
+
+export const ShadowWithDifferentPropsTestSection: React.FunctionComponent = () => {
```
Saadnajmi, line 22: So that this isn't a "catch-all" test section, could we actually have different sections with different sets of props? Not sure if that makes sense / is over-engineering it.

packages/components/Notification/src/Notification.tsx
```diff
@@ -81,7 +81,7 @@ export const Notification = compose<NotificationType>({
 
     return (final: NotificationProps, ...children: React.ReactNode[]) => {
       const { variant, icon, title, action, onActionPress, ...rest } = mergeProps(userProps, final);
-      const mergedProps = mergeProps<PressableProps>(rest, rootStyle);
+      const mergedProps = mergeProps<PressableProps>(rest, { style: rootStyle });
```
Saadnajmi, line 84: Nice catch 😅

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -22,12 +21,95 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
...
+    paddingTop,
+    paddingVertical,
+
+    ...restOfChildStyleProps
+  } = childStyleProps;
+
+  return {
+    inner: {
+      style: {
+        backgroundColor: 'red', // will not be shown, just needed in macOS/iOS to ensure buggy behaviour is suppressed
```
Saadnajmi, line 73: I think this should be a copy of the outer view's background color, for "bleeding through" reasons discussed already.

packages/experimental/Shadow/src/Shadow.tsx
Saadnajmi: +1. I feel like there should be a more generic way to do this.

lyzhan7: I had it originally like that (or something similar, had "padding: padding" - didn't know it could be just "padding") but I think either way will add a bunch of undefined style props - ex. if padding wasn't set on the child originally, now it'll have "padding: undefined, paddingBottom: undefined" etc. Changed it to this so that props are only added if they aren't undefined.

lyzhan7:
> The screenshots look like the red is bleeding through a little bit on the rounded corners. Maybe we should align backgroundColor with the child view?

Oh yeah you're right, can't really tell on the tester but I see it on Notification, thanks for pointing that out. Yeah I can make it so both outer and inner views use the background color, also maybe should be enforced somewhere that there needs to be a background color with opacity = 1

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: That was it, thank you Ruriko

apps/fluent-tester/src/TestComponents/Shadow/ShadowWithDifferentPropsTestSection.tsx
lyzhan7: I thought about that, but I think each section would probably just have one test - feel like it makes more sense if different sections would have more than one test. But don't have a strong preference, I see the tester sometimes has sections that only have a test

packages/experimental/Shadow/src/Shadow.tsx
rurikoaraki: imo that's fine since undefined is the same as it not being set at all. to RN it's the same.

lyzhan7: Okay, changed it back. I think it makes the output of the jest tests harder to read but maybe that's fine

lyzhan7:
Update after making the background colors match:
Screen Shot 2022-09-15 at 5 43 00 PM

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: There wasn't really one that fit what I needed exactly but just moved the getShadowStyle call into the getStyleProps method so that I can just pass in the shadowToken

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: What's the reasoning between not having inline styles? Is it related to the re-render test failing? I.e. if a style object is created inline, re-rendering won't work as expected

lyzhan7: Tried to fix but still a little unclear with what exactly I'm trying to avoid. Is the issue that mergeStyles is not on its own line, or the issue that we're creating a new backgroundColor object? If the reason why we don't like inline styles is related to re-rendering consistency I'm not sure my changes actually fix the issue

apps/fluent-tester/src/TestComponents/Shadow/ShadowWithDifferentPropsTestSection.tsx
lyzhan7: Moved to the themed stylesheet. Also moved some stuff that was in the style objects on the other tests below into the style sheet

lyzhan7:
Overall just having issues with render consistency tests in general, after I made the one line fix in Notification the tests there are failing too and I've tried multiple things (ex. moved creating the style object to a new line above, and making it memoized) but can't get the tests to pass. Are there any kind of guidelines/documentation about this?

packages/components/Notification/src/Notification.tsx
rurikoaraki, line 84: Change rootStyle to return {style: /*whatever the internals are*/ instead of creating the new style object here.

rurikoaraki: The test should tell you why it failed, but if it's not you can change compareTrees in enzymeTests.ts to print out log statements instead of throwing. Beyond that it's figuring out why something changed when it shouldn't have. General guidelines for preventing wasted renders can be found on sites like https://www.freecodecamp.org/news/how-to-identify-and-resolve-wasted-renders-in-react-cc4b1e910d10/

Common culprits are objects (which include arrays and functions) that are recreated on every render - React only does a shallow compare to figure out if props have changed, so new objects get counted as a changed prop even if their internals are the same.

lyzhan7: Tried that, it complained about not having any properties in common with ViewStyle
Screen Shot 2022-09-16 at 12 04 32 PM

So I removed the type of rootStyle, and then the mergeProps line started complaining:
Screen Shot 2022-09-16 at 12 05 58 PM

And then this is something else I tried, skipping mergeProps entirely but the render tests are still failing with this:
Screen Shot 2022-09-16 at 12 02 29 PM

Also another thing I tried with rootStyleObject
Screen Shot 2022-09-16 at 12 03 49 PM

rurikoaraki: The return type changes once you change it to return {style: {styleInternals}} as opposed to {styleInternals}

rurikoaraki: Specifically you could try setting the return type to ViewProps['style'] or NotificationProps

lyzhan7: Okay updating the type to just ViewProps got rid of all the red errors, but the render tests still fail with this

Screen Shot 2022-09-16 at 12 43 37 PM

rurikoaraki: Where does the test say it's failing?

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: ViewStyle seems to work for childStyleProps

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
```diff
@@ -32,21 +24,13 @@ interface ShadowTestBoxProps {
 
 const ShadowTestBox: React.FunctionComponent<ShadowTestBoxProps> = (props: ShadowTestBoxProps) => {
   const theme = useTheme();
-  const themedStyles = getThemedStyles(theme);
   const backgroundColor = props.isBrand ? theme.colors.brandedBackground : theme.colors.background;
```
lyzhan7, line 27: I think I need to memoize backgroundColor here if we want to not cause extra re-renders?

packages/components/Notification/src/Notification.tsx
lyzhan7: Screen Shot 2022-09-16 at 1 12 07 PM

Also committed/pushed the code above so the pipelines should show the error when they're done running

packages/components/Notification/src/Notification.tsx
```diff
@@ -66,17 +66,18 @@ export const Notification = compose<NotificationType>({
...
-        return { alignSelf: 'center', marginHorizontal: marginHorizontal, width: width };
+        return { style: { alignSelf: 'center', marginHorizontal: marginHorizontal, width: width } };
       } else {
-        return { marginHorizontal: marginHorizontal };
+        return { style: { marginHorizontal: marginHorizontal } };
       }
     }, [isBar, width]);
-    const messageStyle: ViewStyle = useMemo(() => {
+
+    const messageStyle: ViewProps = useMemo(() => {
```
rurikoaraki, line 78: messageStyle should stay as {alignSelf} since the result is being passed into the style prop of the message slot.

packages/components/Notification/src/Notification.tsx
rurikoaraki: Okay I misunderstood which test was failing, I thought it was the rerender test (i.e. force an update and make sure the component stays the same) and not the render consistency test (if you render two copies of the same component, the props are all the same since we memoize stuff).

It might be the case that in this component's case, this test is expected to fail. Notification creates a new style. I'll have to think about it a bit

rurikoaraki: Okay so there is a way to get it to work. You need to store the rootStyle on the larger style cache so that the style can be reused across instances of Notification. The way you do that is to create a function which takes in the inputs for creating the rootStyle and returns the root style, and then use memoize from our framework to wrap it (which uses a framework-wide cache) instead of useMemo, which is per-component.

I don't understand why it's not failing already though.. I would kind of expect it to fail as it is the way the component is set up.

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
Saadnajmi: You could move it (and the viewStyle below) into getThemedStyles

packages/components/Notification/src/Notification.tsx
lyzhan7: No sorry my fault for the confusion, I didn't really distinguish between the rerender test/render consistency test in my head because they've usually failed together, I think I wrote render consistency when it should have been rerender

Will try your suggestion. That kind of seems like how getShadowTokenStyleSet works?

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: I need to props.isBrand though...haven't seen examples of this but I could make getThemedStyles take in an extra parameter?

packages/components/Notification/src/Notification.tsx
lyzhan7: Yep, sorry

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
Saadnajmi: You could make two styles inside the stylesheet: one brand and one not

packages/components/Notification/src/Notification.tsx
lyzhan7: That worked locally! Although now I don't really understand only rootStyle needed memoize and not messageStyle, seems like the messageStyle has the same issue rootStyle used to have and should cause the test to fail

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: Oooh

packages/components/Notification/src/Notification.tsx
rurikoaraki: We limit how far the test goes into the render tree. I imagine that if we increase the depth to 5 or something, it would start failing? For completeness it would make sense to fix it for the text styling, too.

lyzhan7: Oh right yeah I think I noticed that at some point with this PR, when the Shadow tests were failing it was only complaining about the outer view when the inner view had the same issue. Will update messageStyle

packages/components/Notification/src/Notification.tsx
```diff
@@ -100,3 +94,13 @@ export const Notification = compose<NotificationType>({
...
+
+const getRootStyle = memoize(getRootStyleWorker);
+function getRootStyleWorker(isBar: boolean, width: number, sizeClass: SizeClassIOS): ViewProps {
+  const marginHorizontal = isBar ? 0 : 16;
+  if (sizeClass === 'regular' && !isBar) {
+    return { style: { alignSelf: 'center', marginHorizontal: marginHorizontal, width: width } };
+  } else {
+    return { style: { marginHorizontal: marginHorizontal } };
+  }
+}
```
Saadnajmi, line 106: Nit, would move this method below line 48, before `Notification`

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
```diff
@@ -4,6 +4,22 @@ exports[`Notification component tests Notification default 1`] = `
...
     Object {
+      "backgroundColor": "#EBF3FC",
+      "borderColor": "transparent",
+      "borderRadius": 12,
+      "flex": 1,
+      "flexDirection": "row",
+      "justifyContent": "space-between",
+      "margin": undefined,
+      "marginBottom": undefined,
+      "marginEnd": undefined,
```
Saadnajmi, line 15: Well this is harder to read... it's not _technically_ any different and we avoid a bunch of spread operators.. so I guess this is the way to go :')

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
```diff
@@ -104,6 +120,21 @@ describe('Shadow component tests', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('Shadow on a child with margin and padding', () => {
+    const tree = renderer.create(<TestShadowOnChildViewWithProps childViewStyleProps={{ margin: 2, padding: 2 }} />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
Saadnajmi, line 126: Also add tests for just margin and just padding?

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -22,12 +21,101 @@ export const Shadow = stagedComponent((props: ShadowProps) => {
...
+
+    const childWithInnerShadow = React.cloneElement(child, innerShadowViewProps);
 
-    return <View {...mergedProps}>{childWithKeyShadow}</View>;
+    return <View {...outerShadowViewProps}>{childWithInnerShadow}</View>;
   };
 });
 
+const getStylePropsForShadowViews = memoize(getStylePropsForShadowViewsWorker);
+function getStylePropsForShadowViewsWorker(childStyleProps: ViewStyle = {}, shadowToken: ShadowToken) {
```
Saadnajmi, line 37: Nit again: Would move declaration above usage. I guess it works though..

Saadnajmi (approved):
Assuming tests pass now, LGTM :)

packages/experimental/Shadow/src/Shadow.tsx
Saadnajmi: I guess with the whole worker/memoize we usually keep it at bottom...

lyzhan7: Yeah from what I've seen the memoize stuff goes at the bottom. Will leave it as is

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: Fixed, let me know if what I did was different from what you had in mind

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
lyzhan7: Yeah and I think there will only be more added. Are spread operators bad?

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
lyzhan7: Personally I think it's more obvious when something is wrong when there's both margin/padding set on the same view. Are there any specific things you think having specific tests will catch?

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
lyzhan7: Think this has been resolved in some other comments below

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
Saadnajmi: No, but it's probably less efficient to have had so many of them. I have no idea tbh

packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
Saadnajmi: I always get the two confused... So I kinda wanted to see a test with one or the other 😅

PR #2141, opened 2022-09-20
Add iOS typography tokens

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

* Add iOS typography token support based on values that come from the newly published `@fluentui-react-native/design-tokens-ios`.
* Add types for text styles that iOS needs (specifically, `Caption1Strong`, `Caption2`, `Title2`, and `Title3`).
* Add a V2 usage section to the Text test page.

### Verification

* Validated via debugging that the appropriate style props are being forwarded to the resulting RN `Text` element.
* Validated that the sizes look correct. Also validated that changing the values in appleTypography.ios.ts updates the test page accordingly (after rerunning `yarn build`).
* Validated that the other test pages in FluentTester don't cause any typography-related errors.

#### Light Mode Screenshot
image

#### Dark Mode Screenshot
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


amgleitman:
One outstanding question that I have is, how do we want to handle types that exist on one platform but not another? For example, `Title3` exists on iOS but not Win32, and `Subtitle1` exists on Win32 but not iOS.

We could try to map unsupported styles to "close-enough" supported ones on a platform-by-platform basis (optionally throwing a warning), or we could do something more abrupt, depending on what we want the development experience to be like.

packages/utils/tokens/src/text-tokens.ts
```diff
@@ -19,19 +19,21 @@ export interface FontStyleTokens {
 export type FontTokens = FontStyleTokens & FontVariantTokens;
 
 export const fontStyles: TokenBuilder<FontTokens> = {
-  from: ({ fontFamily, fontSize, fontWeight, variant }: FontTokens, { typography }: Theme) => {
+  from: ({ fontFamily, fontLetterSpacing, fontLineHeight, fontSize, fontWeight, variant }: FontTokens, { typography }: Theme) => {
     const { families, sizes, weights, variants } = typography;
-    if (fontFamily || fontSize || fontWeight || variant) {
+    if (fontFamily || fontLetterSpacing !== undefined || fontLineHeight !== undefined || fontSize || fontWeight || variant) {
```
lyzhan7, line 24: Seems like all five of these should be consistent? Or is there something about the types of fontLetterSpacing/fontLetterHeight that require "!== undefined". Same thing below, why use ?? rather than ||

lyzhan7:
Do you have screenshots for dark mode?

apps/fluent-tester/src/TestComponents/Text/TextTest.tsx
```diff
@@ -3,13 +3,18 @@ import { StandardUsage } from './StandardUsage';
...
 import { Test, TestSection, PlatformStatus } from '../Test';
 
 const textSections: TestSection[] = [
   {
     name: 'Standard Usage',
     component: StandardUsage,
   },
+  {
+    name: 'V2 Usage',
+    component: V2Usage,
```
Saadnajmi, line 16: We're on V2 now? Not V1?

apps/fluent-tester/src/TestComponents/Text/V2Usage.tsx
```diff
@@ -0,0 +1,43 @@
...
+          <Title2>Title2</Title2>
+          <Title1>Title1</Title1>
+          <LargeTitle>LargeTitle</LargeTitle>
+          <Display>Display</Display>
+        </Stack>
+      </View>
+    );
+  } else {
+    return <Text>Only available on iOS for now.</Text>;
+  }
```
Saadnajmi, line 42: Would prefer we try and support the V1 API, rather than make a V2 that's only on iOS.

packages/theming/apple-theme/src/appleTypography.ios.ts
```diff
@@ -45,20 +49,91 @@ export function appleTypography(): Typography {
       heroSemibold: { face: 'primary', size: 'hero', weight: '600' },
       heroLargeStandard: { face: 'primary', size: 'heroLarge', weight: '400' },
       heroLargeSemibold: { face: 'primary', size: 'heroLarge', weight: '600' },
```
Saadnajmi, line 51: Can we remove these older ones? Or does stuff still reference it?

packages/utils/tokens/src/text-tokens.ts
lyzhan7: Is it just because fontLetterSpacing/fontLetterHeight are optional

apps/fluent-tester/src/TestComponents/Text/TextTest.tsx
rurikoaraki: I think we already have a test for this in the Experimental Text Test...

Sorry for the confusion, but the experimental text test tests the newer Text version. We haven't combined the Text test pages yet.

packages/components/text/src/Variants.win32.ts
```diff
@@ -3,6 +3,8 @@ import { Text } from './Text';
 export const Caption1 = Text.customize({
   variant: 'caption1',
 });
+export const Caption1Strong = Text; // Not supported on Win32
```
rurikoaraki, line 6: If they're not supported, would it make more sense to return null?

rurikoaraki:
> One outstanding question that I have is, how do we want to handle types that exist on one platform but not another? For example, `Title3` exists on iOS but not Win32, and `Subtitle1` exists on Win32 but not iOS.
> 
> We could try to map unsupported styles to "close-enough" supported ones on a platform-by-platform basis (optionally throwing a warning), or we could do something more abrupt, depending on what we want the development experience to be like.

I would personally err on being more abrupt so that clients are more aware of when things aren't supported.

packages/utils/tokens/src/text-tokens.ts
amgleitman: I went with `!== undefined` because otherwise we would ignore defined-but-falsy values such as `{fontLetterSpacing: 0}`.

Honestly, we probably should be testing for being defined in these cases, as well as using `??` instead of `||` for this same reason.

packages/theming/apple-theme/src/appleTypography.ios.ts
amgleitman: I'll have to defer to @rurikoaraki on this one.

rurikoaraki: These are not optional properties on the Typography type. We're not removing the underlying types from the Typography yet because we'll probably break people if we do that right now. I think we need to get all the platforms onto the new tokens first, then we can set about trying to deprecate the old stuff.

packages/components/text/src/Variants.win32.ts
amgleitman: Good idea, and it's consistent with [this comment](https://github.com/microsoft/fluentui-react-native/pull/2141#issuecomment-1252889912).

Resolved with fa48f6d.

apps/fluent-tester/src/TestComponents/Text/V2Usage.tsx
amgleitman: 913342d introduces a better approach for this.

apps/fluent-tester/src/TestComponents/Text/TextTest.tsx
amgleitman: Done.

packages/utils/tokens/src/text-tokens.ts
amgleitman: Resolved with 9f835ea and b96a19f.

rurikoaraki (approved):

yarn.lock
```diff
@@ -1520,6 +1520,11 @@
   resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-android/-/design-tokens-android-0.13.0.tgz#f473d1f3eefdf720715c627b7816a2b41ba35627"
   integrity sha512-bkWBgViYMpJ712+wd82+RIvHcU1ox5TGxV4O46FnJLI9LW7sd+HVP70yW7fd1Kuhkx/MTmI9K3YEG/Aw4JnyOA==
 
+"@fluentui-react-native/design-tokens-ios@^0.19.0":
+  version "0.19.0"
+  resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-ios/-/design-tokens-ios-0.19.0.tgz#8c084af9080e93705239bc8b353da55bbbb507aa"
+  integrity sha512-Kvf3plp5eXN9S6mHK8SeDNhg2XukciVBm7rU954wgIYCePSif9/zixhH/OSSOE9HcYGyXzthRY6KDuu1KJDVCA==
+
 "@fluentui-react-native/design-tokens-macos@^0.13.0":
```
rurikoaraki, line 1528: nit: Might be nice to align these package versions across the platforms, but that can be done in its own PR

amgleitman: I'd say create an issue for it and we can discuss it later.

lyzhan7 (approved):

PR #2142, opened 2022-09-20
Bumping FluentUI Apple version (0.8.3)

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Bumping FluentUI Apple version to 0.8.3

### Verification

Ran test app, no crashes and things are looking good

| Avatar Native                                       | Datepicker Native                                      |
| image | image |

Mac app:
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Avatar/FRNAvatar.podspec
+++ b/packages/experimental/Avatar/FRNAvatar.podspec
@@ -16,11 +16,11 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = "14.0"
   s.ios.source_files      = "ios/*.{swift,h,m}"
-  s.ios.dependency 'MicrosoftFluentUI/Avatar_ios', '0.6.1'
+  s.ios.dependency 'MicrosoftFluentUI', '0.8.3'
 
   s.osx.deployment_target = "10.15"
   s.osx.source_files      = "macos/*.{swift,h,m}"
-  s.osx.dependency 'MicrosoftFluentUI', '0.6.1'
+  s.osx.dependency 'MicrosoftFluentUI', '0.8.3'
 
   s.dependency 'React'
 end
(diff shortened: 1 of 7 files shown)


Saadnajmi:
Should also update the rest of the macOS podspecs

packages/experimental/Avatar/FRNAvatar.podspec
```diff
@@ -16,11 +16,11 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = "14.0"
   s.ios.source_files      = "ios/*.{swift,h,m}"
-  s.ios.dependency 'MicrosoftFluentUI/Avatar_ios', '0.6.1'
```
lyzhan7, line 19: Is there a reason the '/Avatar_ios/ part got removed

lyzhan7 (commented):
Is there a reason why this got removed

packages/experimental/Avatar/FRNAvatar.podspec
sophialee0416: @Saadnajmi had mentioned that it was added because we can use pod subspaces to manage specific components (please correct me if I'm wrong), but updating the pod version alone should be sufficient so we decided to remove it (also one less line to update in the future)

Saadnajmi: Reason was it made it slightly simpler to help Sophia update FUA in FURN. We don't _need_ the subspecs, but it Is technically more correct to have them. In the end, I don't think it actually makes any difference to the binary size in the Nuget for these libraries anyway.

Saadnajmi:
> Should also update the rest of the macOS podspecs

Guesss we don't.. just need to update the lock. In the future we should keep them in sync I suppose.

Saadnajmi (approved):

PR #2145, opened 2022-09-21
Switch Tabs components to use React Native Pressable

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Following up from #1070, let's move over components off of `useAsPressable` / our fork of Pressable and onto Pressable from React Native, which contains the newer support for pointer events / android ripple / codegen commands.

### Verification

Hover, Focus, and Press work as expected on macOS. E2E testing on CI should confirm as much for windows/win32, but I'll try and test it out myself too.

https://user-images.githubusercontent.com/6722175/191470443-2362ba5c-305d-43f0-8ed5-071848617f54.mov

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Tabs/src/TabsItem.types.ts
+++ b/packages/experimental/Tabs/src/TabsItem.types.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ViewStyle, ColorValue } from 'react-native';
+import { PressableProps, ViewStyle, ColorValue } from 'react-native';
 import { FontTokens, IBorderTokens } from '@fluentui-react-native/tokens';
 import { IFocusable, IWithPressableEvents, IPressableState, IWithPressableOptions } from '@fluentui-react-native/interactive-hooks';
 import { IconProps, IconSourcesType } from '@fluentui-react-native/icon';
@@ -146,7 +146,7 @@ export interface TabsItemInfo {
 }
 
 export interface TabsItemSlotProps {
-  root: React.PropsWithRef<IViewProps>;
+  root: React.PropsWithRef<PressableProps>;
   icon: IconProps;
   stack: IViewProps;
   indicator: IViewProps;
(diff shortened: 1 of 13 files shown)


packages/experimental/Tabs/src/TabsItem.types.ts
```diff
@@ -146,7 +146,7 @@ export interface TabsItemInfo {
 }
 
 export interface TabsItemSlotProps {
-  root: React.PropsWithRef<IViewProps>;
+  root: React.PropsWithRef<PressableProps>;
```
Saadnajmi, line 149: Technically we lose the adapter props here. Will have to think about that. Could probably easily make an "IPressableProps" if I _really_ wanted to.

rurikoaraki: Yeah in practice as long as the props we expose for the component stay constant I think we're ok... Even if props that don't match the type are passed in, they're still on the object so they should get passed through, they just will make TS angry if someone tries to modify the root slot directly and add props that don't exist on PressableProps

rurikoaraki: So really it'll be more of a problem for us than for clients.

rurikoaraki:
I can test it for win32/windows for you.

packages/experimental/Tabs/src/TabsItem.types.ts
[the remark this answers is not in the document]
Saadnajmi: Oh... That's probably what I'm missing :D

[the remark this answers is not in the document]
Saadnajmi: Fixed

Saadnajmi: I think type safety is still a but of an open problem in FURN. Adapters is a step towards it but if we enable platforms specific type safety checking via rex-kit, there's still issues. I'll try to spend a day thinking about how best to not regress the experience for ourselves.

rurikoaraki (approved):

PR #2147, opened 2022-09-21
Switch Menu components to Pressable

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Helps to resolve #1035

Back in RN 0.62, we forked RN Core's Pressable, with the intention of eventually removing that fork. We've gone through the effort of updating React Native Core, Windows, Win32, macOS, and the DefinitelyTyped types to make sure that the upstream pressable supports all the hover/focus/press states and types we need. Let's complete this work by replacing our components one by one with Pressable. 

Specifically, we should replace references of "useAsPressable" (which uses our fork) with "usePressableState" (which is designed to work with React Native's Pressable. 

### Verification

Tested that hover/focus/press still work on macOS ContextualMenu. 
Tested that Menu items and disabled menu items are read correctly by VoiceOver
Tested that keyboarding accessibility still behaves as expected

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #2157, opened 2022-09-22
Refactor PressablePropsExtended

Saadnajmi:
EDIT: Refactored this PR to a different approach. Will update description shortly.

### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Related to #1035 

One downside of switching a bunch of components' root slots' to Pressable is the type goes from `IViewProps` to `PressableProps`. The former includes the extra platform specific props. Let's try to fix that with `PressablePropsExtended`!

1) Pulls out `usePressableState` and its types into their own files, so they are no longer sharing types / implementation with `useAsPressable`. This should make it easier to eventually deprecate the latter and make sure that components aren't accidentally using a type from `useAsPressable` (I had a lot of type issues from this mismatch)

2) Refactors `PressablePropsExtended` to be kinda like what IViewProps is for ViewProps: ViewProps + the extra props from the desktop forks. I'll admit it's not the most elegant implementation, but I think it'll work for our purposes. 

3) Refactor Tabs / Experimental Tabs to use the new types. This is where I spent a lot of work to make sure none of the old I-prefixed types were still being used. I also went ahead and fixed some eslint hook warnings while I was at it. 

### Verification

CI should pass. Also can try locally applying change to others like #2147 to make sure the types work.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Tabs/src/useTabsItem.ts
+++ b/packages/experimental/Tabs/src/useTabsItem.ts
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { usePressableState, useKeyProps, useOnPressWithFocus, useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
+import {
+  usePressableState,
+  useKeyProps,
+  useOnPressWithFocus,
+  useViewCommandFocus,
+  IFocusable,
+} from '@fluentui-react-native/interactive-hooks';
 import { TabsItemProps, TabsItemInfo } from './TabsItem.types';
 import { TabsContext } from './Tabs';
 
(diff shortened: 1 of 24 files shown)


rurikoaraki:
> Given that there's also been two "ButtonProps / CheckboxProps / RadiogroupProps".. I'm not too worried about it.

Actually I'm fairly certain it's IButtonProps -> ButtonProps, ICheckboxProps -> CheckboxProps, etc... So you might want to make sure the name doesn't clash

rurikoaraki (approved):

Saadnajmi:
@rurikoaraki I think I've gotten through the rest of the type errors. The main changes since your last review:

1) I'm going with the name "PressablePropsExtended" because that type already existed and I'm reusing it.
 - If we like that name, in the future we could rename IViewProps -> ViewPropsExtended. We'll see though 😅
2) The type exists in `@fluentui-react-native/interactive-hooks` rather than adapters to reduce the amount of copying I had to do in the short term while DefinitelyTyped updates. Mostly just didn't wanna copy `PressableHoverProps`.

If CI passes, I'll probably merge this to unblock other PRs ( when I have downtime...)

packages/experimental/Tabs/src/useTabsItem.ts
```diff
@@ -53,7 +59,7 @@ export const useTabsItem = (props: TabsItemProps): TabsItemInfo => {
     if (itemKey == info.selectedKey) {
       info.updateSelectedTabsItemRef && componentRef && info.updateSelectedTabsItemRef(componentRef);
     }
-  }, []);
+  }, [componentRef, info, itemKey]);
```
rurikoaraki, line 62: post commit review: Based on the comment I'm pretty sure the dependency array is meant to be empty, since the intent is to run it once post-render.

PR #2159, opened 2022-09-22
Android tokens integration - set1

ankraj12:
This PR proposes to add design tokens integrations for android in FURN.  The folders with the key changes are:
**1.  fluentui-react-native\packages\theming\android-theme\src**
Update for default theme to pick colors from alias tokens as part of theme creation ( `colorTokens.ts`) baaed on mode.
Adding code path for picking alias tokens from token pipeline values and mapping into `AliasColorTokens`.

New api's are added in `@fluentui-react-native/theme-tokens`' and '`@fluentui-react-native/theming-utils`, to allow platform specific code flows when filling tokens. The designs are different in endpoints, and the pipeline will not populate all values for android. The token paths for android depend on the new android implementations, keeping the default theme implementation unchanged.

**2.fluentui-react-native\packages\theming\default-theme**
Alias color tokens updated to depend on platform specific API's added in "'@fluentui-react-native/theme-tokens".  
Package is updated to use new version for android design tokens.

**3 packages/theming/theme-tokens/src/**
Exported `getAliasTokens` and `getShadowTokens` from this package, which are implemented in platform implementation. This package currently exported globalTokens, so seemed the right place for pipeline -> tokens mapping. 

```
packages/theming/theme-tokens/src/highContrast/tokens-alias.win32.ts
packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
packages/theming/theme-tokens/src/getShadowTokens.ts
packages/theming/theme-tokens/src/getColorTokens.ts
```
Above are not any code changes, but code refactoring to move this logic out from default theme package to theme tokens package.  I changed this currently only for default themes, but this can be extended for other themes.

4. **packages/theming/theme-types/src/Color.types.ts**
`AliasColorTokens` properties are marked optional for the properties which are unavailable for android currently via pipeline. This allows rest of endpoint to keep reusing the complete set, while android can fill it partially. None of the existing tokens are removed. There would be certain changes as more tokens are integrated via pipeline, but even at the end - the token set for mobile will not be 1-1 with other endpoints until the designs across platforms converge to 1 figma.

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Verification
Tested in fluent tester. Theme values are correctly loaded for android. 
I still need to see coverage for automation tests. 

Loaded and checked the tester app for both Android and win32 with this change - it loads.

This is not full set of tokens for android, but remaining will follow in same lines.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.vscode/settings.json
```diff
@@ -51,7 +51,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
Saadnajmi, line 54: Is one better? 😅

packages/theming/android-theme/src/androidTheme.ts
```diff
@@ -4,6 +4,7 @@ import { paletteFromAndroidColors } from './colorsTokens';
 import { androidTypography } from './androidTypography';
 import { getAndroidPalette } from './colorsBase';
 
+// VSO# - Remaning tokens for android.
```
Saadnajmi, line 7: If we go have an issue to file, we try to use public GitHub issues with GH#

packages/theming/android-theme/src/colorsTokens.ts
```diff
@@ -1,11 +1,13 @@
 import { ThemeColorDefinition } from '@fluentui-react-native/theme-types';
 import { FluentUIAndroidPalette } from './colorsSemantic';
+import { createColorAliasTokens } from './createAliasTokens';
 
-/** creates a palette of colors for the android theme, given the FluentUI Android Palette */
+/** creates a palette of colors for the android theme, given the FluentUI Android Palette - ??.
+ * What is the rest if the palette color?
```
Saadnajmi, line 6: Intentional?

packages/theming/theme-tokens/src/getColorTokens.ts
```diff
@@ -0,0 +1,21 @@
...
+import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+import { assertNever } from 'assert-never';
+
+export function getAliasTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return lightAliasTokens;
+  } else if (mode === 'dark') {
+    return darkAliasTokens;
+  } else if (mode === 'highContrast') {
+    return hcAliasTokens;
```
Saadnajmi, line 13: Does Android even have a high contrast mode? I only see the option "High contrast text" on my phone.

packages/theming/theme-tokens/src/getShadowTokens.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function getShadowTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return lightShadowTokens.shadow;
+  } else if (mode === 'dark') {
+    return darkShadowTokens.shadow;
+  }
+
+  // HC mode.
+  return hcShadowTokens.shadow;
```
Saadnajmi, line 14: Shadows are fundamentally different on Android, so I'd expect this file to change in the future. I guess it's ok as is.

packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
```diff
@@ -0,0 +1,18 @@
...
+import aliasTokens from '@fluentui-react-native/design-tokens-win32/hc/tokens-aliases.json';
+
+export const hcAliasTokens = processAliasTokens(aliasTokens);
+
+function processAliasTokens(aliasTokens: any) {
+  for (const key in aliasTokens) {
+    for (const innerKey in aliasTokens[key]) {
+      const entry = aliasTokens[key][innerKey];
+      if (typeof entry === 'string' && entry.includes('PlatformColor')) {
+        const color = 'SystemColor' + entry.substring(14, entry.length - 1) + 'Color';
```
Saadnajmi, line 11: This looks like possibly Android specific logic?

packages/theming/theming-utils/src/mapPipelineToTheme.android.ts
```diff
@@ -0,0 +1,94 @@
+import { AliasColorTokens, Variants } from '@fluentui-react-native/theme-types';
+
+// API that translates tokens coming for android to Theme color values.
+// This is implemented in a per-plaform fashion, for each endpoint that maps to similar token sets in design - i.e. map to similar
+// pipeline output.
```
Saadnajmi, line 5: What part of this is Android specific?

.vscode/settings.json
rurikoaraki: We shouldn't be updating vscode settings for everyone without discussing it first, please revert changes to this file

ankraj12: This change is not intentional, something in my VS code updated this. I will restore this.

packages/theming/android-theme/src/androidTheme.ts
ankraj12: Sounds good, will log a GH issue.

packages/theming/android-theme/src/colorsTokens.ts
ankraj12: Kind of - it was something I wanted to ask on the PR - but not as a check-in comment inline.

packages/theming/theme-tokens/src/getColorTokens.ts
ankraj12: No, there are no high contrast tokens in android. Android will not hit this code path as the bundle will pick `export function getAliasTokens(mode: AppearanceOptions)` from `getColorTokens.android.ts` added in PR. This api is an in current implementation (packages/theming/default-theme/src/createAliasTokens.ts) - which is moved to "theme-tokens" package, and both android theme package and default theme now use `getAliasTokens` from  here -
`import { getAliasTokens, getShadowTokens } from '@fluentui-react-native/theme-tokens';
`

packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
ankraj12: No, this is existing code from "default-themes" where getTokens.ts uses this,  getTokens.ts is moved in this package, this folder is also refactored here. 

So I need to have platform implementations for android for the API's used in "[createAliasTokes.ts](https://github.com/microsoft/fluentui-react-native/pull/2159/files#diff-0066b5262771ef012f09b7c921851dc60560a7f5a23234fd2d7b0b73817f1682)" for android -
 1)  aliasTokens = `getAliasTokens`(mode);

This api returns the tokens based on light and dark mode for android based on appearance ( no high contrast mode). This is a subset of theme tokens for other endpoints.
The alternative was to keep a file 'getTokens.ts' in every folder, but I moved getTokens to be exported from theme-tokens packages as:
1.1 ) The flow in fluent tester triggers default theme flow - IDK why yet, but not having a android implementation for this mixes flow to pick "./getTokens" api for default themes and then calls 
`mapPipelineToTheme`(aliasTokens) implemented for android - crashes as token set is not 1-1. So there needs a common package export for this api which is theme-tokens.

1.2) I felt theme-tokens is fitting right, this exports globals - and having "./getTokens" in same folder keeps all the code referring to design-tokens of fluentui-desing-tokens directly in same package.

 2) `mapPipelineToTheme`(aliasTokens); 
This needed a fork for android, as there are tokens not present in android. For eg, 
    transparentStroke: **pipelineOutput.transparentStroke**.**strokeColorRest**,
    transparentStrokeInteractive: **pipelineOutput.transparentStroke**.**strokeColorHover**,
    transparentStrokeDisabled: **pipelineOutput.transparentStroke**.**strokeColorDisabled**,

The **bold** ones are not by design in android, so passing through default `packages/theming/theming-utils/src/mapPipelineToTheme.ts` crashes for multiple such tokens. The android corresponding file added PR only has the tokens which are meaningful for android.  I considered using optional chaining in the file, but this looked more readable.

The need is to allow decoupling of tokens code between existing desktop and android, if there are alternate options on how to achieve this, keen for suggestions.

packages/theming/android-theme/package.json
```diff
@@ -32,7 +32,8 @@
   "license": "MIT",
   "dependencies": {
     "@fluentui-react-native/theme-types": ">=0.18.0 <1.0.0",
-    "@fluentui-react-native/theme": ">=0.7.0 <1.0.0"
+    "@fluentui-react-native/theme": ">=0.7.0 <1.0.0",
+    "@fluentui-react-native/theme-tokens": "^0.19.0"
```
rurikoaraki, line 36: FYI these packages are outdated now

packages/theming/default-theme/src/createAliasTokens.ts
```diff
@@ -1,4 +1,4 @@
-import { getAliasTokens, getShadowTokens } from './getTokens';
+import { getAliasTokens, getShadowTokens } from '@fluentui-react-native/theme-tokens';
```
rurikoaraki, line 1: Please don't add all the platform tokens into the theme-tokens package. We put alias tokens for each platform in their own <platform>-theme packages to minimize the cost of importing themes, per feedback we received. Clients don't want to import theme information for platforms they don't support, but if we do this then they'll be forced to add theme tokens as a dependency for platforms they don't care about. Please put the getAliasTokens and getShadowTokens to what they had before.

packages/theming/theme-tokens/src/getColorTokens.android.ts
```diff
@@ -0,0 +1,12 @@
+import lightAliasTokens from '@fluentui-react-native/design-tokens-android/light/tokens-aliases.json';
```
rurikoaraki, line 1: This package should be global-tokens only - alias tokens should be dealt with in their own platform-theme packages.

rurikoaraki: theme-tokens is a bit of a misnomer since it used to hold all the tokens, but we moved alias tokens out of this package based on feedback from clients.

rurikoaraki (changes requested):
Please keep alias tokens within platform-theme packages and not central packages like theme-tokens

packages/theming/default-theme/src/createAliasTokens.ts
rurikoaraki: For reference: https://github.com/microsoft/fluentui-react-native/pull/1562

rurikoaraki: Chatted offline, with what @ankraj12 was seeing this seems like a good approach. Looking through it again and making sure that the approach makes sense and won't cause the bundle size issues we had before.

packages/theming/theme-tokens/src/getTokens.ts
```diff
@@ -0,0 +1,21 @@
+import lightAliasTokens from '@fluentui-react-native/design-tokens-windows/light/tokens-aliases.json';
+import darkAliasTokens from '@fluentui-react-native/design-tokens-windows/dark/tokens-aliases.json';
+import { hcAliasTokens } from './highContrast/tokens-alias';
+import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+import { assertNever } from 'assert-never';
+
+export function getAliasTokens(mode: AppearanceOptions) {
```
rurikoaraki, line 7: Is there a reason we can't move this file & the highContrast files over? I see adds, not a file move

packages/components/Avatar/src/AvatarTokens.ts
```diff
@@ -198,4 +198,4 @@ function getColorProps(color: string, theme: Theme) {
         ringColor: theme.colors.transparentStroke,
       };
   }
-}
+}
```
rurikoaraki, line 201: nit: this should be reverted. I think you'll want to run prettier across the files.

packages/theming/default-theme/src/createAliasTokens.ts
ankraj12: Based on conversation with @rurikoaraki , I would be updating this review resolving remaining feedbacks. This approach on forking alias tokens to be picked from theme-tokens package seems safe, and should not cause size bloat with unneeded tokens for an endpoint. More on this comment - https://github.com/microsoft/fluentui-react-native/pull/2159#discussion_r978318779

As next steps:
- I would forward merge with main, and cleanup and resolve other feedbacks.
- The changes for win32 and macOS to follow this design would be not in this PR, I would log GH issues to track this.

packages/theming/theme-tokens/src/getTokens.ts
ankraj12: Actually, they should be moves as I didn't change the most of the files, let me try correcting this in next commit.

ankraj12: I updated the change to mark this as Move, please review and let know any questions!

packages/theming/android-theme/package.json
ankraj12: Also, running yarn test with this code fails some snapshot tests - would it safe to add all the updated snap files to fix those errors? Or there are additional recommended validations I should trigger?

rurikoaraki (approved):

packages/theming/android-theme/package.json
ankraj12: Updated the package versions and new snapshots

PR #2160, opened 2022-09-22
Moved Switch from experimental to components folder, publishing as V1

ejlayne:
### Platforms Impacted
- [ ] iOS
- [ x ] macOS
- [ x ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Moved Switch from experimental to components folder, publishing as V1

### Verification

Did manual testing of Switch component in tester on Win32, and also ran yarn test on full repo to verify all tests still pass.

### Pull request checklist

- [ x ] Automated Tests
- [ x ] Documentation and examples
- [ x ] Keyboard Accessibility
- [ x ] Voiceover
- [ x ] Internationalization and Right-to-left Layouts


packages/components/Switch/jest.config.js
```diff
@@ -0,0 +1,2 @@
+const { configureReactNativeJest } = require('@fluentui-react-native/scripts');
+module.exports = configureReactNativeJest('android');
```
lyzhan7, line 2: I'm a little unsure why a couple files got added instead of just renamed -I think there is a packages/experimental/Switch/jest.config.js file, and a packages/experimental/Switch/src/switch.types.ts file, so these should be registered as "file renamed without changes" instead of as a new file.

packages/experimental/RadioGroup/src/__tests__/__snapshots__/RadioGroupExperimental.test.tsx.snap
```diff
@@ -0,0 +1,21 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
lyzhan7, line 1: This doesn't seem related to your change, maybe it got added when you ran yarn test?

packages/libraries/core/package.json
```diff
@@ -42,6 +42,7 @@
     "@fluentui-react-native/pressable": "0.9.12",
     "@fluentui-react-native/radio-group": "0.15.15",
     "@fluentui-react-native/separator": "0.12.4",
+    "@fluentui-react-native/switch": "0.6.8",
```
rurikoaraki, line 45: We've been erring towards not bloating the core library because of bundle size concerns.

I think we should move towards not having the core bundle and just have people import components separately. Maybe have a min-bundle that includes the very basic things, like Text and Button. But it's not useful for people to have all the components.

packages/experimental/RadioGroup/src/__tests__/__snapshots__/RadioGroupExperimental.test.tsx.snap
ejlayne: Agreed, I thought I removed this when I saw it was part of my CL, thanks for catching this.

Saadnajmi:
Anything blocking this?

Saadnajmi (approved):

PR #2173, opened 2022-09-28
Implemented label subtexts in RadioGroups

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Implemented a new prop for label subtexts in Radio. 

### Verification

| image | image |

### Pull request checklist

- [] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
+++ b/apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
@@ -11,10 +11,10 @@ export const SubtextRadioGroup: React.FunctionComponent = () => {
   return (
     <View>
       <RadioGroup label="Uncontrolled RadioGroup" defaultValue="X" onChange={onChange}>
-        <Radio label="Option W" value="W" accessibilityLabel="Test Accessibility Label" />
-        <Radio label="Option X" value="X" />
-        <Radio label="Option Y" value="C" />
-        <Radio label="Option Z" value="Z" />
+        <Radio label="Apple" subtext="This is a type of fruit" value="W" />
+        <Radio label="Pear" value="X" />
+        <Radio label="Banana" value="C" />
+        <Radio label="Orange" value="Z" />
       </RadioGroup>
     </View>
   );
--- a/apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
+++ b/apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
@@ -1,0 +1,21 @@
+import { RadioGroup, Radio } from '@fluentui-react-native/experimental-radio-group';
+import * as React from 'react';
+import { View } from 'react-native';
+
+export const SubtextRadioGroup: React.FunctionComponent = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
+    console.log(key);
+  };
+
+  return (
+    <View>
+      <RadioGroup label="Uncontrolled RadioGroup" defaultValue="X" onChange={onChange}>
+        <Radio label="Option W" value="W" accessibilityLabel="Test Accessibility Label" />
+        <Radio label="Option X" value="X" />
+        <Radio label="Option Y" value="C" />
+        <Radio label="Option Z" value="Z" />
+      </RadioGroup>
+    </View>
+  );
+};
--- a/packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
+++ b/packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
@@ -1,7 +1,8 @@
 import { radioName, RadioTokens, RadioSlotProps, RadioProps } from './Radio.types';
-import { UseStylingOptions, buildProps } from '@fluentui-react-native/framework';
+import { Theme, UseStylingOptions, buildProps } from '@fluentui-react-native/framework';
 import { globalTokens } from '@fluentui-react-native/theme-tokens';
 import { defaultRadioTokens } from './RadioTokens';
+import { fontStyles } from '@fluentui-react-native/tokens';
 
 export const radioStates: (keyof RadioTokens)[] = ['focused', 'hovered', 'pressed', 'selected', 'disabled'];
 
@@ -13,7 +14,7 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
       (tokens: RadioTokens) => ({
         style: {
           display: 'flex',
-          alignItems: 'center',
+          alignItems: 'flex-start',
           flexDirection: 'row',
           minHeight: 20,
           marginTop: 0,
@@ -56,12 +57,35 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
       }),
       ['radioInnerCircleSize', 'radioVisibility', 'radioFill'],
     ),
+    content: buildProps(
+      () => ({
+        style: {
+          display: 'flex',
+          alignItems: 'flex-start',
+          flexDirection: 'column',
+        },
+      }),
+      [],
+    ),
     label: buildProps(
-      (tokens: RadioTokens) => ({
+      (tokens: RadioTokens, theme: Theme) => ({
         variant: tokens.variant,
         style: {
           marginTop: 2,
           color: tokens.color,
+          ...fontStyles.from(tokens, theme),
+        },
+      }),
+      ['color', ...fontStyles.keys],
+    ),
+    subtext: buildProps(
+      (tokens: RadioTokens) => ({
+        variant: tokens.variant,
+        style: {
+          marginTop: globalTokens.spacing.xxs,
+          marginRight: globalTokens.spacing.xxs,
+          color: tokens.color,
+          fontSize: tokens.subtextSize,
         },
       }),
       ['color'],
--- a/packages/experimental/RadioGroup/src/Radio/Radio.tsx
+++ b/packages/experimental/RadioGroup/src/Radio/Radio.tsx
@@ -27,7 +27,9 @@ export const Radio = compose<RadioType>({
     root: View,
     button: View,
     innerCircle: View,
+    content: View,
     label: Text,
+    subtext: Text,
   },
   filters: {
     button: filterViewProps,
@@ -39,14 +41,21 @@ export const Radio = compose<RadioType>({
 
     // now return the handler for finishing render
     return (final: RadioProps) => {
-      const { label, ...mergedProps } = mergeProps(radio.props, final);
+      const { label, subtext, ...mergedProps } = mergeProps(radio.props, final);
+
+      const labelComponent = (
+        <Slots.content>
+          <Slots.label>{label}</Slots.label>
+          {!!subtext && <Slots.subtext>{subtext}</Slots.subtext>}
+        </Slots.content>
+      );
 
       return (
         <Slots.root {...mergedProps}>
           <Slots.button>
             <Slots.innerCircle />
           </Slots.button>
-          <Slots.label>{label}</Slots.label>
+          {labelComponent}
         </Slots.root>
       );
     };
--- a/packages/experimental/RadioGroup/src/Radio/Radio.types.ts
+++ b/packages/experimental/RadioGroup/src/Radio/Radio.types.ts
@@ -39,6 +39,11 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
   radioBorderWidth?: number;
 
   /**
+   * Border width of Radio
+   */
+  subtextSize?: number;
+
+  /**
    * States that can be applied to a Radio
    */
   selected?: RadioTokens;
@@ -53,6 +58,11 @@ export interface RadioProps extends IPressableProps {
    * The text string for the option
    */
   label: string;
+
+  /**
+   * Label subtext for the option
+   */
+  subtext?: string;
 
   /**
    * A unique key-identifier for each option
@@ -82,7 +92,9 @@ export interface RadioSlotProps {
   root: IViewProps;
   button: IViewProps;
   innerCircle: IViewProps;
+  content: IViewProps;
   label: TextProps;
+  subtext?: TextProps;
 }
 
 export interface RadioType {
--- a/packages/experimental/RadioGroup/src/Radio/useRadio.ts
+++ b/packages/experimental/RadioGroup/src/Radio/useRadio.ts
@@ -11,6 +11,7 @@ export const useRadio = (props: RadioProps): RadioState => {
   const defaultComponentRef = React.useRef(null);
   const {
     label,
+    subtext,
     value,
     disabled,
     accessibilityActions,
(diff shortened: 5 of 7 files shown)


packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -38,6 +38,11 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
    */
   radioBorderWidth?: number;
 
+  /**
+   * Border width of Radio
```
lyzhan7, line 42: Update this comment?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -82,12 +83,13 @@ export const useRadio = (props: RadioProps): RadioState => {
...
       value,
       label,
+      subtext,
       ...rest,
       ref: buttonRef,
       ...pressable.props,
       accessible: true,
       accessibilityRole: 'radio',
-      accessibilityLabel: accessibilityLabel ?? label,
+      accessibilityLabel: accessibilityLabel ?? (subtext ? label + subtext : label),
```
lyzhan7, line 92: Do you maybe need to add a space or something here between label and subtext

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,12 +57,35 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
       }),
       ['radioInnerCircleSize', 'radioVisibility', 'radioFill'],
     ),
+    content: buildProps(
+      () => ({
+        style: {
+          display: 'flex',
```
lyzhan7, line 63: Was just curious if you need to set display: 'flex' for things to work? I think by default display is already set to 'flex'

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Removing this line doesn't stylistically change anything, so I suppose it is set by default. Removed, thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Thanks for catching this! Consulted @ksiler and we decided to put subtext in accessibilityHint.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,12 +57,34 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
+      (tokens: RadioTokens) => ({
+        variant: tokens.variant,
+        style: {
+          marginTop: globalTokens.spacing.xxs,
+          marginRight: globalTokens.spacing.xxs,
+          color: tokens.color,
+          fontSize: tokens.subtextSize,
         },
       }),
       ['color'],
```
rurikoaraki, line 90: Cache key needs to include all tokens that would invalidate the style. In this case that would be variant, color, and subtextSize

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,12 +57,34 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
+          ...fontStyles.from(tokens, theme),
+        },
+      }),
+      ['color', ...fontStyles.keys],
+    ),
+    subtext: buildProps(
+      (tokens: RadioTokens) => ({
+        variant: tokens.variant,
+        style: {
+          marginTop: globalTokens.spacing.xxs,
```
rurikoaraki, line 84: Feels like these could be tokens, a client may want to vary how much spacing is between the label and the sublabel

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -38,6 +38,11 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
    */
   radioBorderWidth?: number;
 
+  /**
+   * Font size of label subtext
+   */
+  subtextSize?: number;
```
rurikoaraki, line 44: Should this be a variant instead of a direct size number? I recognize that the example just used size, but in general we should be steering people towards using font alias tokens/variants over global tokens/size constants

rurikoaraki (commented):
Some questions

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: That's a good point, thanks for the suggestion. Updated!

apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
```diff
@@ -0,0 +1,21 @@
+import { RadioGroup, Radio } from '@fluentui-react-native/experimental-radio-group';
+import * as React from 'react';
+import { View } from 'react-native';
+
+export const SubtextRadioGroup: React.FunctionComponent = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
```
rurikoaraki, line 7: Should use React.useCallback

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,15 +57,38 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
+        variant: tokens.subtextVariant,
+        style: {
+          marginTop: tokens.marginTop,
+          marginRight: tokens.marginRight,
+          marginBottom: tokens.marginBottom,
           color: tokens.color,
         },
       }),
-      ['color'],
+      ['variant', 'marginTop', 'marginRight', 'marginBottom', 'color', 'subtextVariant'],
```
rurikoaraki, line 91: variant doesn't need to be part of the cache key since it's not being used here.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,15 +57,38 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
+      (tokens: RadioTokens, theme: Theme) => ({
         variant: tokens.variant,
         style: {
           marginTop: 2,
+          marginRight: tokens.marginRight,
+          color: tokens.color,
+          ...fontStyles.from(tokens, theme),
+        },
+      }),
+      ['color', 'marginRight', ...fontStyles.keys],
```
rurikoaraki, line 79: variant should be added here as a cache key

apps/fluent-tester/src/TestComponents/RadioGroupExperimental/SubtextRadioGroup.tsx
gulnazsayed: Thanks for calling this out. Updated this in all of the experimental radiogroup test components.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -56,15 +57,38 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
       }),
       ['radioInnerCircleSize', 'radioVisibility', 'radioFill'],
     ),
+    content: buildProps(
```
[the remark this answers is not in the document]
gulnazsayed: Removed!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Removed!

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Added, thanks!

apps/fluent-tester/src/TestComponents/RadioGroupExperimental/DisabledRadioGroup.tsx
```diff
@@ -1,22 +1,25 @@
...
   return (
     <View>
+      <Button></Button>
       <RadioGroup disabled label="Disabled RadioGroup" defaultValue="X" onChange={onChange}>
         <Radio label="Option W" value="W" accessibilityLabel="Test Accessibility Label" />
         <Radio label="Option X" value="X" />
         <Radio label="Option Y" value="C" />
         <Radio label="Option Z" value="Z" />
       </RadioGroup>
+      <Button></Button>
```
rurikoaraki, line 22: Is this button meant to be here?

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
```diff
@@ -31,13 +31,12 @@ export const useRadioGroup = (props: RadioGroupProps): RadioGroupInfo => {
...
     props: {
       ...props,
-      disabled,
       required,
-      accessible: true,
       accessibilityRole: 'radiogroup',
       accessibilityLabel: accessibilityLabel ?? label,
       accessibilityState: getAccessibilityState(state.disabled, state.required, accessibilityState),
       defaultTabbableElement: selectedButtonRef,
+      focusable: !state.disabled,
```
rurikoaraki, line 39: Is the radiogroup supposed to be focusable separately from radios?

Also this change doesn't seem to be related to the PR, it's easy to miss this change in a large PR like this, so in the future I'd recommend making separate PRs for changes so that they're reviewed properly

samuelfreiberg: +1 I don't think the RadioGroup should be focusable itself, just the Radios

apps/fluent-tester/src/TestComponents/RadioGroupExperimental/DisabledRadioGroup.tsx
gulnazsayed: Nope, sorry, forgot to remove this after some testing. Removed!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -38,6 +39,26 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
...
 
+  /**
+   * Variant of label subtext
+   */
+  subtextVariant?: keyof Variant;
+
+  /**
+   * Padding between label and label subtext
+   */
+  marginTop?: ViewStyle['marginTop'];
```
samuelfreiberg, line 50: Should the names of these signify they're specific towards label subtext?

Ie. `subtextMarginTop`

Right now, I would assume that these values are for the label itself

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -82,7 +108,9 @@ export interface RadioSlotProps {
   root: IViewProps;
   button: IViewProps;
   innerCircle: IViewProps;
+  content: IViewProps;
```
samuelfreiberg, line 111: Can we get more specific with the `content` name? Looking at this, I'd assume this slot wraps the whole component. But looking at Radio.tsx, it looks like it's specific for the label + subtext.

Maybe we rename it `labelContent`?

samuelfreiberg: It saves the developer the extra step of having to figure out what content this slot is referring to

packages/experimental/RadioGroup/src/Radio/Radio.tsx
```diff
@@ -39,14 +41,21 @@ export const Radio = compose<RadioType>({
 
     // now return the handler for finishing render
     return (final: RadioProps) => {
-      const { label, ...mergedProps } = mergeProps(radio.props, final);
+      const { label, subtext, disabled, ...mergedProps } = mergeProps(radio.props, final);
```
rurikoaraki, line 44: Curious why disabled is pulled out here?

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Thanks for the suggestion! I agree that these names should be more descriptive.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Thanks for the suggestion, updated!

packages/experimental/RadioGroup/src/Radio/Radio.tsx
gulnazsayed: Pulled this out to test if it might be affecting the focus zone of disabled RadioGroups, but it didn't end up mattering if it's applied or pulled out here. Reverted this change since it doesn't apply to this PR / affect it.

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
```diff
@@ -45,7 +45,7 @@ export const RadioGroup = compose<RadioGroupType>({
         // @ts-ignore - TODO, fix typing error
         contextValue.buttonKeys = React.Children.map(children, (child: React.ReactChild) => {
```
rurikoaraki, line 46: nit: should the name of this property change to reflect that it's not using buttonKeys anymore?

Also what happens if the value is the same across multiple radios? (though I guess the key didn't have any safeguards either. But a key is expected to be unique, at least)

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
```diff
@@ -45,7 +45,7 @@ export const RadioGroup = compose<RadioGroupType>({
...
           if (React.isValidElement(child)) {
-            return child.props.buttonKey;
+            return child.props.value;
```
rurikoaraki, line 48: I think this means that buttonKey is not used as a prop anymore on the Radio.

gulnazsayed: We renamed buttonKey prop to value in V1 Radio

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
gulnazsayed: Updated to values to reflect V1 naming convention. 

We don't have safeguards to ensure that the values are unique, nor did old RadioGroup with the buttonKeys. We have a comment that says the value must be a unique key-identifier to prompt the user to pass in unique values. If there are multiple Radios with the same value, they would all show as selected when one is clicked.

samuelfreiberg (approved):

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
rurikoaraki: I think that's a bug. In theory we should be able to handle cases where values between multiple options are the same (for example a joke radio group having "yes, yes, yes" as options.

PR #2175, opened 2022-09-28
Improve E2E Scrolling Logic for Win32/Windows

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

I've seen a couple flaky tests + test retries in Win32 due to the scrolling behavior. Previously, we were using an API called `.touchScroll()` which uses a touch to scroll. This seems very flaky. Now, we are using "PageDown" to scroll, which seems a lot more stable than touch.

In order to PageDown on a test page, we need to put focus on the page first. That's why I added a button at the top of each test page so we can click it to put focus on the page whenever we want. I made the opacity 0 so it's not visible to anyone using the app.

Also, this PR adds all Keys to the constants and changes all references to the old Keys (which is why so many files have changed).

Temporarily removing Menu test as it's causing a failure. Created a task to follow up

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


PPatBoyd:
the only thing I dislike about `PageDown` is chunk-scroll could lead to partial element visibility depending on layout; is mouse wheel" scrolling available?  Also any insight into why the touch scrolling is flaky, what might be affecting it?

samuelfreiberg:
> the only thing I dislike about `PageDown` is chunk-scroll could lead to partial element visibility depending on layout; is mouse wheel" scrolling available? Also any insight into why the touch scrolling is flaky, what might be affecting it?

I definitely understand that concern with PageDown. I add an extra PageDown at the end to ensure the component as a whole is showing.

There is no mouse wheel scrolling available. As for touchScroll, we aren't exactly sure why it's flaky. Sometimes we get 500 internal server error from WinAppDriver and we don't know why. I've investigated and can't find the reason. WinAppDriver is close-sourced and hasn't been iterated on in a long time, which is concerning. I've reached out to some Microsoft employees who worked on it previously to hopefully learn a bit more.

But for now, removing the touchScroll API until we figure out the reason for the flakiness seems necessary.

ksiler (approved):

PPatBoyd (approved):

PR #2176, opened 2022-09-28
RadioGroup Horizontal and Horizontal-Stacked

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added layout prop to RadioGroup and labelPosition prop to Radio. Added section to tester page for other layouts of RadioGroups.

### Verification

| n/a | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
The layout for the Horizontal-Stacked RadioGroup looks a bit funky... The spacing between the Radio's aren't consistent, leading to a weird looking UI. I think that needs to be fixed. Probably need extra styling logic

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
```diff
@@ -12,6 +12,11 @@ export interface RadioGroupState extends RadioGroupProps {
    */
   value: string | null;
 
+  /**
+   * Layout of the Radios. True if horizontal or horizontal-stacked.
+   */
+  isHorizontal?: boolean;
```
rurikoaraki, line 18: Is this needed if RadioGroupState extends RadioGroupProps and it already has `layout`?

rurikoaraki: I guess it helps with token setting.

rurikoaraki: Though the logic could be moved to the lookup function without needing to add a property to the state I think

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -38,9 +38,20 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
...
+
+  /**
+   * The flex direction of the root
+   */
+  marginTop?: ViewStyle['marginTop'];
+
   /**
    * States that can be applied to a Radio
    */
+  isBelow?: RadioTokens;
```
rurikoaraki, line 54: Can we rename this to be more descriptive of what is below what? like labelPositionBelow?

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -38,9 +38,20 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
    */
   radioBorderWidth?: number;
 
+  /**
+   * The flex direction of the root
+   */
+  flexDirection?: ViewStyle['flexDirection'];
+
+  /**
+   * The flex direction of the root
```
rurikoaraki, line 47: nit: comment copied

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -23,8 +28,7 @@ export const useRadio = (props: RadioProps): RadioState => {
     ...rest
   } = props;
 
-  // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
-  const selectedInfo = useRadioGroupContext();
+  console.log(labelPosition);
```
lyzhan7, line 31: console.log added accidentally?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -76,12 +80,14 @@ export const useRadio = (props: RadioProps): RadioState => {
     ...pressable.state,
     selected: selectedInfo.value === props.value,
     disabled: selectedInfo.disabled || disabled || false,
+    isBelow: labelPosition === 'below' ? true : false,
```
lyzhan7, line 83: I think you could just do `isBelow: labelPosition === 'below',`

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
```diff
@@ -24,6 +25,8 @@ export const useRadioGroup = (props: RadioGroupProps): RadioGroupInfo => {
     value: data.selectedKey,
     required: required || false,
     disabled: disabled || false,
+    layout: layout || 'vertical',
+    isHorizontal: layout === 'horizontal' || layout === 'horizontal-stacked' ? true : false,
```
lyzhan7, line 29: Also here, could probably just do `isHorizontal: layout === 'horizontal' || layout === 'horizontal-stacked',`

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.styling.ts
```diff
@@ -44,5 +44,16 @@ export const stylingSettings: UseStylingOptions<RadioGroupProps, RadioGroupSlotP
       }),
       ['requiredColor', 'requiredPadding'],
     ),
+    options: buildProps(
```
lyzhan7, line 47: I was thinking the name 'layout' might make more sense here (the style props below are all grouped together as 'layout' in [react native docs](https://reactnative.dev/docs/flexbox)), and maybe the 'layout' prop you added could be 'labelPosition' or 'labelDirection' or something

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
```diff
@@ -55,7 +56,7 @@ export const RadioGroup = compose<RadioGroupType>({
           <Slots.root {...mergedProps}>
             {label && labelComponent}
             <Slots.container isCircularNavigation defaultTabbableElement={defaultTabbableElement}>
-              {children}
+              <Slots.options>{children}</Slots.options>
```
lyzhan7, line 59: It seems like we're adding an extra View just for the layout stuff - is it possible for the layout stuff in Slots.options to be included in Slots.container? Not really familiar with FocusZone though so not sure if that's appropriate/possible to do or not

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Thanks for catching this!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: That's a good idea. Replaced!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -9,10 +9,15 @@ const defaultAccessibilityActions = [{ name: 'Select' }];
 
 export const useRadio = (props: RadioProps): RadioState => {
   const defaultComponentRef = React.useRef(null);
+
+  // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
+  const selectedInfo = useRadioGroupContext();
```
rurikoaraki, line 14: nit: Can the name of this be something like "context" or "radioGroupContext" instead? "selectedInfo" doesn't seem like a good name for this since it has more than information about which radio is selected.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -76,7 +97,12 @@ export interface RadioProps extends IPressableProps {
   enableFocusRing?: boolean;
 }
 
-export interface RadioState extends IPressableHooks<RadioProps & React.ComponentPropsWithRef<any>> {}
+export interface RadioState extends IPressableHooks<RadioProps & React.ComponentPropsWithRef<any>> {
+  /**
+   * If horizontal-stacked, label should be below indicator.
+   */
+  labelPositionBelow?: boolean;
```
rurikoaraki, line 104: Nit this isn't necessarily required - you could change the lookup function to check for the labelPositionBelow state by looking at the props

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.styling.ts
```diff
@@ -44,5 +44,16 @@ export const stylingSettings: UseStylingOptions<RadioGroupProps, RadioGroupSlotP
       }),
       ['requiredColor', 'requiredPadding'],
```
rurikoaraki, line 45: Note to self: I got here

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.styling.ts
rurikoaraki: As a slot I think options or optionsContainer makes sense, since it tells us where in the UI the slot is being used. I think layout is pretty vague and clients can always add non-layout props to any slot, so we shouldn't rely on the props to decide what to call a slot.

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
rurikoaraki, line 18: Nit: Doesn't necessarily need to be a state, can be derived from props

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.styling.ts
gulnazsayed: Leaving this named as options for now. Please let me know if I should change.

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: I agree, it's not descriptive. Changed, thanks!

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
```diff
@@ -24,6 +25,8 @@ export const useRadioGroup = (props: RadioGroupProps): RadioGroupInfo => {
     value: data.selectedKey,
     required: required || false,
     disabled: disabled || false,
+    layout: layout || 'vertical',
```
rurikoaraki, line 28: Why are both of these needed on the state?

rurikoaraki:
To give more clarity on props vs. state in RN, props are information about the component you get from the parent, whereas state is information that is kept track of within the component which affects its render behavior. So things like hover, pressed, are state, because they are kept track of internally by the component. Both are available to the component for use in determining rendering.

So in general, if things are able to be derived from props, we shouldn't need to add them to the state. If you need the information on to build a context, then that's a case where there can be an expection since we've been using the pattern of using state to fill out the context (although I guess in theory in doesn't need to be an exception).

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
gulnazsayed: Moved the logic to lookup function, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Moved logic to lookup function

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
gulnazsayed: Moved logic to lookup function

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
gulnazsayed: Moved the logic to lookup function, but keeping layout in the state to maintain pattern of using state to fill out context.

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -23,25 +28,22 @@ export const useRadio = (props: RadioProps): RadioState => {
...
-  }, [selectedInfo, value, componentRef]);
+  }, [radioGroupContext, value, componentRef]);
 
   /* We use the componentRef of the currently selected button to maintain the default tabbable
     element in a RadioGroup. Since the componentRef isn't generated until after initial render,
     we must update it once here. */
   React.useEffect(() => {
-    if (value == selectedInfo.value) {
-      selectedInfo.updateSelectedButtonRef && componentRef && selectedInfo.updateSelectedButtonRef(componentRef);
+    if (value == radioGroupContext.value) {
```
ValentinaKozlova, line 45: Should there be strict equality `===`?

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -23,25 +28,22 @@ export const useRadio = (props: RadioProps): RadioState => {
...
   } = props;
 
-  // Grabs the context information from RadioGroup (currently selected button and client's onChange callback)
-  const selectedInfo = useRadioGroupContext();
-
   const buttonRef = useViewCommandFocus(componentRef);
 
   /* We don't want to call the user's onChange multiple times on the same selection. */
   const changeSelection = React.useCallback(() => {
-    if (value != selectedInfo.value) {
```
ValentinaKozlova, line 33: Strict equality `!==`?

apps/fluent-tester/src/TestComponents/RadioGroupExperimental/HorizontalRadioGroup.tsx
```diff
@@ -0,0 +1,29 @@
+import { RadioGroup, Radio } from '@fluentui-react-native/experimental-radio-group';
+import { Separator } from '@fluentui/react-native';
+import * as React from 'react';
+import { View } from 'react-native';
+
+export const HorizontalRadioGroup: React.FunctionComponent = () => {
+  // Client's example onChange function
+  const onChange = (key: string) => {
```
samuelfreiberg, line 8: React.usecallback()

packages/experimental/RadioGroup/src/Radio/useRadio.ts
gulnazsayed: Updated, thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -93,8 +97,8 @@ export const useRadio = (props: RadioProps): RadioState => {
       accessibilityHint: accessibilityHint ?? subtext,
       accessibilityState: getAccessibilityState(state.disabled, state.selected, accessibilityState),
       accessibilityActions: accessibilityActionsProp,
-      accessibilityPositionInSet: accessibilityPositionInSet ?? selectedInfo.values.findIndex((x) => x == value) + 1,
-      accessibilitySetSize: accessibilitySetSize ?? selectedInfo.values.length,
+      accessibilityPositionInSet: accessibilityPositionInSet ?? radioGroupContext.buttonKeys.findIndex((x) => x == value) + 1,
```
rurikoaraki: I thought this was changed to values?

rurikoaraki (approved):

samuelfreiberg (approved):

packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -25,25 +30,22 @@ export const useRadio = (props: RadioProps): RadioState => {
...
   /* We use the componentRef of the currently selected button to maintain the default tabbable
     element in a RadioGroup. Since the componentRef isn't generated until after initial render,
     we must update it once here. */
   React.useEffect(() => {
-    if (value == selectedInfo.value) {
-      selectedInfo.updateSelectedButtonRef && componentRef && selectedInfo.updateSelectedButtonRef(componentRef);
+    if (value === radioGroupContext.value) {
+      radioGroupContext.updateSelectedButtonRef && componentRef && radioGroupContext.updateSelectedButtonRef(componentRef);
     }
   }, []);
```
Saadnajmi, line 50: Post review comment, but I see an eslint error about the empty dependency array on line 48. Should it be empty?

gulnazsayed: Yes, it should be empty here since we only want to update only once. I believe this is an error that occurs for old RadioGroup too.

PR #2178, opened 2022-09-29
Fix component cache

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

I noticed that the component cache doesn't actually cache the result from the first time we hit the code, so it's not actually caching anything.

Fixing the cache to save the result of requiring the native component.

Added unit test
Also added new config to run jest debugger for the current folder's tests

### Verification

Test written to ensure that the cache is working and that the requireNativeComponent code isn't hit more than once per component

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.vscode/launch.json
```diff
@@ -31,6 +31,21 @@
       "sourceMaps": true,
       "outFiles": []
     },
+    {
+      "name": "Debug current folder test",
```
rurikoaraki, line 35: This adds a way to attach a debugger to the jest instance that will run the tests in your current folder.

packages/framework/component-cache/jest.config.js
```diff
@@ -0,0 +1,2 @@
+const { configureReactNativeJest } = require('@fluentui-react-native/scripts');
+module.exports = configureReactNativeJest('android', { preset: 'react-native' });
```
rurikoaraki, line 2: Jest recommends that you add the react-native preset for running in RN scenarios. I'm not actually sure how much this helps so I'm open to removing the preset.

PR #2180, opened 2022-09-29
Set theme to 'Office' as default for win32 tester

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Since this causes a lot of confusion and misalignment of what's seen in the tester vs. what will show up in products, setting the default of the theme picker to "Office" for win32 so that the tester boots straight into "What will partners see".

### Verification

Booted tester in win32, it starts on "Office".

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/theme/CustomThemes.ts
+++ b/apps/fluent-tester/src/theme/CustomThemes.ts
@@ -32,7 +32,7 @@ export const lightnessOptions = [
 ];
 
 export class TesterThemeReference extends ThemeReference {
-  private _themeName: ThemeNames = 'Default';
+  private _themeName: ThemeNames = Platform.OS === ('win32' as any) ? 'Office' : 'Default';
   private _brand: OfficeBrand = 'Default';
 
   private options: ThemeOptions;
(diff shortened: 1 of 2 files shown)


apps/fluent-tester/src/theme/CustomThemes.ts
```diff
@@ -32,7 +32,7 @@ export const lightnessOptions = [
 ];
 
 export class TesterThemeReference extends ThemeReference {
-  private _themeName: ThemeNames = 'Default';
+  private _themeName: ThemeNames = Platform.OS === ('win32' as any) ? 'Office' : 'Default';
```
ValentinaKozlova, line 35: Is there a type for platforms (not to use any)?

rurikoaraki: This is expected as win32 isn't a public platform for RN. So this conversion is necessary for the code to accept win32

ValentinaKozlova (approved):

PR #2187, opened 2022-09-30
Deprecate old theming

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Deprecating all exports from @uifabricshared/theming-react-native and @uifabricshared/theme-registry, both of which support the old theming system we don't use anymore.

### Verification

Code only touches comments except for one change which changes what is probably a C-style cast into TS cast.


packages/deprecated/theming-react-native/src/Theme.types.ts
```diff
@@ -1,6 +1,15 @@
 import { IThemeRegistry, IProcessTheme } from '@uifabricshared/theme-registry';
 import { ITheme, IPartialTheme } from '@uifabricshared/theming-ramp';
 
+/**
+ * @deprecated
+ */
 export { ITheme, IPartialTheme } from '@uifabricshared/theming-ramp';
+/**
```
ankraj12, line 8: supernite's on line break above and few other places

ankraj12 (approved):
Curious to know - what is the impact on partners using the now deprecated API's..? Does it build fail for consuming packages..?

rurikoaraki:
> Curious to know - what is the impact on partners using the now deprecated API's..? Does it build fail for consuming packages..?

Not really unless you have a linter set up for that scenario. Partners will see the APIs struck out in their IDEs, but in terms of development it won't impact their build.

PR #2197, opened 2022-10-05
Add iOS spacing tokens

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

* Add iOS spacing tokens based on values that come from the latest version of `@fluentui-react-native/design-tokens-ios`.
* Add relevant spacing types based on latest token specifications.
* Add a test page showing off basic usage of these tokens.

### Verification

* Validated via the newly created test page.
Simulator Screen Shot - iPad (9th generation) - 2022-10-04 at 17 07 30

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -6,12 +6,26 @@ import { PartialShadowDefinition, ThemeShadowDefinition } from './Shadow.types';
...
+
 export interface Spacing {
   s2: string;
   s1: string;
-  m: string;
+  // m is now defined as a v2 type
   l1: string;
   l2: string;
+  // v2 types
+  none?: SpacingValue;
```
rurikoaraki, line 18: Thanks for this, I should probably get the win32 spacing to come through here too instead of pulling directly from global tokens, which is what we do today

rurikoaraki: It does mean that clients are able to change spacing which we generally don't allow for global tokens though.. hm.

rurikoaraki: Do you know if design has settled on spacing having alias tokens?

packages/theming/apple-theme/src/appleTheme.ios.ts
```diff
@@ -1,9 +1,18 @@
+import globalTokens from '@fluentui-react-native/design-tokens-ios/light/tokens-global.json';
+
 import { Theme, Spacing } from '@fluentui-react-native/theme-types';
 import { paletteFromAppleColors } from './appleColors.ios';
 import { appleTypography } from './appleTypography.ios';
 
 function appleSpacing(): Spacing {
```
lyzhan7, line 7: Would it make sense for this method to be renamed to iOSSpacing?

apps/fluent-tester/src/TestComponents/Spacing/SpacingTest.tsx
```diff
@@ -0,0 +1,63 @@
...
+        {spacingExample('XXXS', appleTheme.spacing.xxxs)}
+        {spacingExample('XXS', appleTheme.spacing.xxs)}
+        {spacingExample('XS', appleTheme.spacing.xs)}
+        {spacingExample('S', appleTheme.spacing.s)}
+        {spacingExample('M', appleTheme.spacing.m)}
+        {spacingExample('L', appleTheme.spacing.l)}
+        {spacingExample('XL', appleTheme.spacing.xl)}
+        {spacingExample('XXL', appleTheme.spacing.xxl)}
+        {spacingExample('XXXL', appleTheme.spacing.xxxl)}
+        {spacingExample('XXXXL', appleTheme.spacing.xxxxl)}
```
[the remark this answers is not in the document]
amgleitman: Since the s2/s1/l1/l2 values aren't part of the latest design spec, I think it would be better to deprecate them.

packages/theming/apple-theme/src/appleTheme.ios.ts
amgleitman: Yeah, although it would probably also make sense to apply this for all .ios.ts and .macos.ts files in this folder, so I'm thinking this can be done separately.

apps/fluent-tester/src/TestComponents/Spacing/SpacingTest.tsx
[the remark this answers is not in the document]
rurikoaraki: Yeah, they should be depreacted over time. Spacing is a lesser-know property so it might be easier to deprecate more quickly.

packages/theming/theme-types/src/Theme.types.ts
amgleitman: As far as I can tell, spacing doesn't have alias tokens.

lyzhan7 (approved):

lyzhan7:
Since these are iOS tokens - do you also have screenshots for iOS? 
Also I realized there already is a Token page in FluentTester. I'm not sure if the plan was to put all tokens there, or maybe we should rename that page to Color Tokens or something

packages/theming/theme-types/src/Theme.types.ts
ankraj12: I would be interested in this conversation, I am putting up a similar change for Android and if spacing needs to be put elsewhere outside Theme to be non modifiable by clients - what would be a good way to handle this..?

PR #2201, opened 2022-10-05
Align Avatar color hash algorithm to web

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

This reverts commit https://github.com/microsoft/fluentui-react-native/commit/3e10c16fa33ba27bd97861e18525ec140593140f (PR https://github.com/microsoft/fluentui-react-native/pull/2013) that aligned macOS, iOS, and Android to native behavior. We decided we want to match web instead (see the web hash algorithm [here](https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-avatar/src/components/Avatar/useAvatar.tsx#:~:text=const%20getHashCode%20%3D%20(str%3A%20string)%3A%20number%20%3D%3E%20%7B](https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-avatar/src/components/Avatar/useAvatar.tsx#:~:text=const%20getHashCode%20%3D%20%28str:%20string%29:%20number%20%3D%3e%20%7b))), which was actually what we were doing before.

Changes in addition to what we were doing before: 
- used length-3 as the number of Avatar colors, since there are extra colors for the PersonaCoin in FURN
- added automated tests

### Verification

In the Avatar test page, I used the "Customize Usage" section with some specific inputs to test the color algorithm.
To confirm that the color was the same as FluentUI web, I used this [codesandbox](https://codesandbox.io/s/avatarcolor-2b1qjm?file=/example.tsx).

| Before                                       | After                                      | Web |
|----------------------------------------------|--------------------------------------------|-|
| before-bill | after-bill | web-bill |
| before-satya | after-satya | web-satya
| before-noname | after-noname | web-noname

Notes
-  There's probably a better way to do the length-3 adjustment besides subtracting a hardcoded number - this will need to be updated if we add any more colors, this also breaks if we change the order of the colors - I was thinking of maybe splitting up the colors into something like 'Avatar only colors' and 'Avatar + PersonaCoin' colors. If there are other suggestions let me know
- I personally found the Avatar test page a little confusing to navigate. When changing the inputs in the "standard usage" section it's not clear which inputs will affect which of the avatars below. Maybe each of the avatars could be labelled more with what specifically they're trying to test.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ValentinaKozlova:
I think it's time to have only one changing Avatar and put other avatars to a separate section. It was added on the beginning for easier testing, but now it's not needed

PR #2207, opened 2022-10-07
Remove extra buttons android

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Based on design guidelines Toggle Button and Compound are not required for Android and have been removed.

FAB uses the following pattern to return null for the component -
https://github.com/microsoft/fluentui-react-native/blob/3963a0726c20a60634e8d3e23e3b38524f3f15fc/packages/components/Button/src/FAB/FAB.tsx#L9-L21

While Checkbox on MacOS has gone for the following approach -
https://github.com/microsoft/fluentui-react-native/blob/3963a0726c20a60634e8d3e23e3b38524f3f15fc/packages/components/Checkbox/src/Checkbox.macos.tsx#L7-L27

In my testing, I found the following to work as well -
https://github.com/microsoft/fluentui-react-native/blob/96e8fdc7316e392d847b3b447b571fbd6433053c/packages/components/Button/src/ToggleButton/ToggleButton.android.tsx#L8-L21

Is any one of the approaches better than the other?

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/Button/ButtonTest.tsx
```diff
@@ -26,10 +27,13 @@ const buttonSections: TestSection[] = [
     name: 'Icon Button',
     component: ButtonIconTest,
   },
-  {
-    name: 'Toggle Button',
-    component: ToggleButtonTest,
-  },
+  Platform.select({
+    android: null, //Toggle Button is not implemented for Android
```
ankraj12, line 31: Do we have a similar tester update needed for Compound button as well?

ankraj12 (approved):

apps/fluent-tester/src/TestComponents/Button/ButtonTest.tsx
ayush547: No, it is not required since Compound button does not have a dedicated section, it just appears in other test sections

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -155,6 +155,7 @@ export const Test = (props: TestProps): React.ReactElement<Record<string, never>
         )}
       </Stack>
       {props.sections.map((section, index) => {
+        if (section == null) return <></>;
```
rurikoaraki, line 158: style nit: For readability, it'd be nice to:
1. Have a newline after this conditional statement
2. Have the return statement be on its own line.

Having it on one line like this without any breaks makes this easy to miss while reading code.

packages/components/Button/src/CompoundButton/__snapshots__/CompoundButton.test.tsx.snap
```diff
@@ -1,91 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`CompoundButton default 1`] = `
```
rurikoaraki, line 3: This isn't good, we'll lose coverage for render tree changes in these buttons. Can we change the platform that jest runs on so we can keep these snapshot tests? The jest.config file is currently configured to use Android as its platform.

rurikoaraki (changes requested):
I'd like to keep our snapshot test coverage

packages/components/Button/src/CompoundButton/__snapshots__/CompoundButton.test.tsx.snap
ayush547: Have modified the jest testing platform to iOS for Button related components

rurikoaraki (approved):

PR #2210, opened 2022-10-07
Update shadow test page background color in FluentTester

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The current dark mode background color in the iOS tester is black (0x000000). Shadows don't show up against a black background, so this PR is to change the background color to something that shadows are visible against.

Notes:
- the new background color is based off the background color macOS uses. Open to other suggestions (was considering white since that's what's shown in Figma)
- I would have preferred to change the entire background of the test page for Shadow, instead of each individual section, but I think I would have needed to make changes in the general tester code and I wanted to keep this change scoped to the Shadow test section. 
- I removed the stackStyle styling from the shadow button test section, this was on purpose I think the section does not need it.

### Verification

- checked that macOS light mode/dark mode have not been impacted
- checked that iOS light mode has not been impacted

Before 

https://user-images.githubusercontent.com/78454019/194469006-e7582ae8-e309-4dab-b174-dab4fcd3d7ee.mov

After

https://user-images.githubusercontent.com/78454019/194647115-f8539c73-d77e-4940-a4b9-bad20c59f646.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


lyzhan7:
Just merged PR that added dark mode tokens, and updated the 'after' screen recording. Before the 'after' screen recording was actually using light mode tokens, after the merge it is now using dark mode tokens as expected. Note there is still a bug (work tracked elsewhere) where switching between light/dark mode doesn't always update the shadows.

Before merge commit
https://user-images.githubusercontent.com/78454019/194647433-ab4cc9db-de51-4ef2-b549-4ab4d0bf7248.mov

After merge commit
https://user-images.githubusercontent.com/78454019/194647477-b783d1e3-59e8-48fb-b56f-af97680fe313.mov

ValentinaKozlova (approved):

apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -11,7 +12,7 @@ const CustomFABShadow64 = FAB.customize({
 export const ShadowButtonTestSection: React.FunctionComponent = () => {
   if (Platform.OS === 'ios') {
     return (
-      <View style={stackStyle}>
+      <View style={{ backgroundColor: shadowTestPageBackgroundColor() }}>
```
Saadnajmi, line 15: Nit: Inline style would force unnecessary re-renders

apps/fluent-tester/src/TestComponents/Shadow/ShadowTestBackgroundColor.tsx
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * For dark mode iOS, returns a dark grey that shadows are visible on. Otherwise, will return either undefined or the default
+ * background color from the theme based on whether or not a theme has been provided.
+ */
+export const shadowTestPageBackgroundColor = (t?: Theme) => {
+  if (Platform.OS === 'ios' && Appearance.getColorScheme() === 'dark') {
+    return '#313131';
+  }
+  return t?.colors.background;
+};
```
Saadnajmi, line 16: This is pretty close to [building your own hook](https://reactjs.org/docs/hooks-custom.html), it would just need to switch from `getColorScheme` to `useColorScheme`. If we did that change, then this method should begin with `use`, e.g: `useShadowBackgroundColor`. However, I think something like this would be better in a themed style sheet, where you can just read the theme's apprearance rather than using `Appearance.getColorScheme()`

Saadnajmi (changes requested):

apps/fluent-tester/src/TestComponents/Shadow/ShadowTestBackgroundColor.tsx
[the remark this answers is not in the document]
lyzhan7: No I haven't brought this up with design yet, but I have an upcoming meeting on to talk about Shadows and can try to ask about it then. My assumption was that sometimes you wouldn't see the shadows in dark mode and other times you would (if a different background is specified), so for the tester let's just create a scenario where the shadows are visible.

apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
lyzhan7: Oh missed that, thanks for catching it

lyzhan7: Should be fixed now after changing to using themedStyleSheet

apps/fluent-tester/src/TestComponents/Shadow/ShadowTestBackgroundColor.tsx
lyzhan7: Switched to themedStyleSheet

apps/fluent-tester/src/TestComponents/Shadow/ShadowDepthTestSection.tsx
```diff
@@ -54,7 +53,7 @@ export const ShadowDepthTestSection: React.FunctionComponent = () => {
   const theme = useFluentTheme();
 
   return (
-    <View>
+    <View style={shadowTestPageStyles(theme).backgroundColor}>
```
Saadnajmi, line 56: I'd still prefer to move the function call to it's own line

lyzhan7:
> @lyzhan7 Anything blocking this from being merged?

No sorry I'll do that now

apps/fluent-tester/src/TestComponents/Shadow/ShadowTestPageStyles.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * For dark mode iOS, returns a dark grey that shadows are visible on. Otherwise, will return either undefined or the default
+ * background color from the theme based on whether or not a theme has been provided.
+ */
+
+export const shadowTestPageStyles = themedStyleSheet((t: Theme) => {
+  const useLighterBackgroundColor = Platform.OS === 'ios' && t.host.appearance === 'dark';
+
+  return {
+    backgroundColor: {
+      backgroundColor: useLighterBackgroundColor ? '#313131' : t.colors.background,
```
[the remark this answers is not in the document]
lyzhan7: Hmm I'm not sure if we have good candidates for iOS. I'll take another look when I start adding color tokens to FURN iOS

PR #2212, opened 2022-10-07
[Dropdown] Basic Dropdown implementation, add slots for Dropdown

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change provides a basic implementation of Dropdown. Dropdown was switched to use compressible due to needing more flexibility to implement the Button slot (specifically its onClick callback). This change adds basic slots for Dropdown (but not the callout).

Chevron SVG used is ChevronDown 16 regular from Fluent-System-Icons, in line with the small size of the Dropdown spec.

Styling, tokens, more props, and integration with Listbox and Options will be in separate PRs. 

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Dropdown/src/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,49 @@
...
+const Dropdown = compressible<DropdownProps, DropdownTokens>((props: DropdownProps, _useTokens: UseTokens<DropdownTokens>) => {
+  const onButtonClick = React.useCallback(() => {}, []); //eslint-disable-line
+  const buttonProps: ButtonProps = React.useMemo(
+    () => ({
+      onClick: onButtonClick,
+    }),
+    [onButtonClick],
+  );
+
+  const expandIconProps: SvgProps = React.useMemo(
```
rurikoaraki, line 19: This could be a const object outside of the component, but it will probably take tokens in the future so I think it makes sense to keep it as a memoized function which returns a constant

Dewsk (approved):

PR #2214, opened 2022-10-08
Use alias token definitions for iOS typography

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Our token pipeline generates alias tokens already, so we should be using those. There's a bit of conversion logic that we need to write, but it's not too bad.

One other thing we needed to do is to add `normal` and `bold` as potential `FontWeights` properties since [these are legal font weight values in React Native](https://reactnative.dev/docs/text-style-props#fontweight). These might cause type-checking errors if anyone wants to define their own font weights, but these should be easily resolved.

### Verification

The text elements still look right.

Screen Shot 2022-10-07 at 5 22 08 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ankraj12 (commented):
Please review if the font pipeline to tokens can be moved to theming-utils

rurikoaraki:
I'm not quite understanding this:

> One other thing we needed to do is to add normal and bold as potential FontWeights properties since [these are legal font weight values in React Native](https://reactnative.dev/docs/text-style-props#fontweight).

Wouldn't that mean that if someone passes in normal or bold into a Text component RN should support it? I'm not understanding why that means FURN theming should support it

packages/theming/apple-theme/src/appleTypography.ios.ts
```diff
@@ -50,90 +60,18 @@ export function appleTypography(): Typography {
...
-        weight: fontTokens.weight.bold,
-        lineHeight: fontTokens.lineHeight[900],
-        letterSpacing: 0.26,
-      },
+      caption2: convertAliasFont(aliasTokens.caption2 as FontStyleTokens),
+      caption1: convertAliasFont(aliasTokens.caption1 as FontStyleTokens),
+      caption1Strong: convertAliasFont(aliasTokens.caption1Strong as FontStyleTokens),
+      body2: convertAliasFont(aliasTokens.body2 as FontStyleTokens),
+      body2Strong: convertAliasFont(aliasTokens.body2Strong as FontStyleTokens),
+      body1: convertAliasFont(aliasTokens.body1 as FontStyleTokens),
```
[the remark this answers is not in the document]
amgleitman: Ah, I didn't know this existed. I'll see if we can share any of this logic.

> I'm not quite understanding this:
> 
> > One other thing we needed to do is to add normal and bold as potential FontWeights properties since [these are legal font weight values in React Native](https://reactnative.dev/docs/text-style-props#fontweight).
> 
> Wouldn't that mean that if someone passes in normal or bold into a Text component core-RN should support it? I'm not understanding why that means FURN theming, which is to support Fluent Design, should support it

Our alias font tokens come in the form of [`FontStyleTokens`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/utils/tokens/src/text-tokens.ts#L11-L17). Its `fontWeight` property is allowed to be a key of [`TextStyle.fontWeight`](https://reactnative.dev/docs/text-style-props#fontweight), which includes values of `'normal'` and `'bold'`.

In order to properly convert this to a [`VariantValue`](https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/theme-types/src/Typography.types.ts#L107-L113)'s [`FontWeight`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L92), we must do one of the following:
* Add it as a key of [`FontWeights`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L84-L87). This is what I opted for because it seemed to fit well with other other existing names of `regular` and `semibold`.
* Add it to the [`FontWeightValue`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L73) enum. This didn't make as much sense since this enum is otherwise full of numbers.
* Add a function that converts these new keywords to their numerical counterparts (that is, `'400'` for `'normal'` and `'700'` for `'bold'`). This seems reasonable as these numerical mappings appear to be standard, but it may not be future-proof if React Native chooses to support new names, which is why I shied away from this option (although the option I went with would require updating things anyway, so maybe it wouldn't make a difference).

amgleitman:
> I'm not quite understanding this:
> 
> > One other thing we needed to do is to add normal and bold as potential FontWeights properties since [these are legal font weight values in React Native](https://reactnative.dev/docs/text-style-props#fontweight).
> 
> Wouldn't that mean that if someone passes in normal or bold into a Text component core-RN should support it? I'm not understanding why that means FURN theming, which is to support Fluent Design, should support it

Our alias font tokens come in the form of [`FontStyleTokens`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/utils/tokens/src/text-tokens.ts#L11-L17). Its `fontWeight` property is allowed to be a key of [`TextStyle.fontWeight`](https://reactnative.dev/docs/text-style-props#fontweight), which includes values of `'normal'` and `'bold'`.

In order to properly convert this to a [`VariantValue`](https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/theme-types/src/Typography.types.ts#L107-L113)'s [`FontWeight`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L92), we must do one of the following:
* Add it as a key of [`FontWeights`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L84-L87). This is what I opted for because it seemed to fit well with other other existing names of `regular` and `semibold`.
* Add it to the [`FontWeightValue`](https://github.com/microsoft/fluentui-react-native/blob/e07c6548be5d6e18613e46e309ab667040455369/packages/theming/theme-types/src/Typography.types.ts#L73) enum. This didn't make as much sense since this enum is otherwise full of numbers.
* Add a function that converts these new keywords to their numerical counterparts (that is, `'400'` for `'normal'` and `'700'` for `'bold'`). This seems reasonable as these numerical mappings appear to be standard, but it may not be future-proof if React Native chooses to support new names, which is why I shied away from this option (although the option I went with would require updating things anyway, so maybe it wouldn't make a difference).

amgleitman:
After some offline discussion, we determined that the second bullet point above was the best thing to go with since it makes things consistent with what React Native gives us.

rurikoaraki (approved):

packages/theming/apple-theme/src/appleTypography.ios.ts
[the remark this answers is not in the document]
amgleitman: It doesn't look like there's an elegant way to combine the logic, but I can at least move the relevant parts of code to this package.

ankraj12 (approved):

rurikoaraki (approved):

Saadnajmi:
@amgleitman still relevant?

amgleitman:
> @amgleitman still relevant?

I think we should still do this, although we'd have to rework some parts of this. Specifically, we'd need to adjust our token pipeline to output Dynamic Type information, but we can hard-code them here temporarily.

packages/theming/apple-theme/src/appleTypography.ios.ts
```diff
@@ -1,8 +1,5 @@
-import globalTokens from '@fluentui-react-native/design-tokens-ios/light/tokens-global.json';
-
 import { FontSize, FontSizes, FontWeightValue, Typography, Variants } from '@fluentui-react-native/theme-types';
-
-const fontTokens = globalTokens.font;
+import { createFontAliasTokens } from './createFontAliasTokens.ios';
```
Saadnajmi, line 2: I don't think we need the .ios here

yarn.lock
```diff
@@ -1465,6 +1465,15 @@
...
 
+"@fluentui-react-native/tokens@^0.17.2":
+  version "0.17.8"
+  resolved "https://registry.yarnpkg.com/@fluentui-react-native/tokens/-/tokens-0.17.8.tgz#1b5adebaa03c1d7cd0dd184c838f9c4b20152969"
+  integrity sha512-hOVk4ZXd1D/kP+IowLwPLcGTtmLPu8KgaPBq3j3HEOb58kuUoITe05P09FMvf0IwC4lFPhdR2ZXQBKx3BG/xHg==
+  dependencies:
+    "@fluentui-react-native/adapters" ">=0.10.0 <1.0.0"
+    "@fluentui-react-native/theme-types" ">=0.23.1 <1.0.0"
+    tslib "^2.3.1"
+
```
Saadnajmi, line 1476: This showing up in your lock file generally means that you're using an older version of the package as a dependency somewhere, which causes problems

packages/theming/apple-theme/src/appleTypography.ios.ts
amgleitman: For now we don't, but this does fit the pattern of [other imports](https://github.com/microsoft/fluentui-react-native/blob/8466c50ff1204515570e59456d6428bdd6a933ff/packages/theming/apple-theme/src/appleTheme.ios.ts#L2-L4) in this package, so I say let's keep it this way.

yarn.lock
amgleitman: This should be resolved now.

PR #2222, opened 2022-10-10
Re-enable and rewrite parts of menu e2etest

lawrencewin:
E2E testing for the Menu component was disabled until now due to the spec failing in the pipeline. 

This PR re-enables the Menu tests and fixes the logic errors within the test spec which caused failures in CI. 

### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- Removed the spec file from the list of excluded specs in the wdio config.
- Use getAttribute('ExpandCollapse.ExpandCollapseState') on primary component to check if menu is open rather than the value of a text component controlled by the menu

### Verification

2 / 3 functional tests were failing before. Now they all pass.

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
+++ b/apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
@@ -9,6 +9,7 @@ import {
   ExpandCollapseState,
 } from '../../../TestComponents/Menu/consts';
 import { BasePage, By } from '../../common/BasePage';
+import { Keys } from '../../common/consts';
 
 /* This enum gives the spec file an EASY way to interact with SPECIFIC UI elements on the page.
  * The spec file should import this enum and use it when wanting to interact with different elements on the page. */
@@ -41,16 +42,7 @@ class MenuPageObject extends BasePage {
   }
 
   async getMenuItemAccessibilityLabel(componentSelector: MenuComponentSelector): Promise<string> {
-    switch (componentSelector) {
-      case MenuComponentSelector.PrimaryComponent:
-        return await this._primaryComponent.getAttribute('Name');
-
-      case MenuComponentSelector.SecondaryComponent:
-        return await this._secondaryComponent.getAttribute('Name');
-
-      case MenuComponentSelector.TertiaryComponent:
-        return await this._tertiaryComponent.getAttribute('Name');
-    }
+    return await (await this.getMenuComponentSelector(componentSelector)).getAttribute('Name');
   }
 
   async getMenuAccessibilityRole(): Promise<string> {
@@ -68,14 +60,24 @@ class MenuPageObject extends BasePage {
 
   /* Returns the correct WebDriverIO element from the Button Selector */
   async getMenuComponentSelector(menuComponentSelector?: MenuComponentSelector): Promise<WebdriverIO.Element> {
-    if (menuComponentSelector == MenuComponentSelector.PrimaryComponent) {
-      return await this._primaryComponent;
+    switch (menuComponentSelector) {
+      case MenuComponentSelector.PrimaryComponent:
+        return await this._primaryComponent;
+      case MenuComponentSelector.SecondaryComponent:
+        return await this._secondaryComponent;
+      case MenuComponentSelector.TertiaryComponent:
+        return await this._tertiaryComponent;
+      default:
+        return await this._primaryComponent;
     }
-    return await this._primaryComponent;
   }
 
   async resetTest() {
-    await (await this._defocusButton).click();
+    // Both escape on the menu trigger to hard dismiss menu and click defocus to reset focus
+    if (await this.menuIsExpanded()) {
+      await this.sendKey(MenuComponentSelector.PrimaryComponent, Keys.ESCAPE);
+      await (await this._defocusButton).click();
+    }
   }
 
   /*****************************************/
--- a/apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
+++ b/apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
@@ -6,10 +6,9 @@ import {
   MENUITEM_TEST_COMPONENT,
   MENUPOPOVER_TEST_COMPONENT,
   MENU_DEFOCUS_BUTTON,
-  ExpandCollapseState,
 } from '../../../TestComponents/Menu/consts';
 import { BasePage, By } from '../../common/BasePage';
-import { Keys } from '../../common/consts';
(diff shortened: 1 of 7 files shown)


apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
```diff
@@ -23,25 +22,19 @@ class MenuPageObject extends BasePage {
...
-    await browser.waitUntil(async () => await callbackText.isDisplayed(), {
-      timeout: this.waitForUiEvent,
-      timeoutMsg: 'The Menu did not close.',
-      interval: 1000,
-    });
-
-    return await callbackText.isDisplayed();
+  async menuIsExpanded(): Promise<boolean> {
+    const expandState = await (await this._primaryComponent).getAttribute('ExpandCollapse.ExpandCollapseState');
+    console.log(expandState);
```
PPatBoyd, line 36: would it be fair to say that we're testing that the ExpandCollapse state changes as-expected when we invoke the button, but we haven't actually verified that a menu appeared on-screen? Asking in part relative to the commented out code in Menu.spec.win.ts re: `TestID for Callouts aren't propagating, so this test is failing`

lawrencewin: This is a fair assessment. I've added a line of code that also tests if the primary menu item is also displayed in the same method.

samuelfreiberg: The accessibility tests also open the menu and query for a menu item, so we verify the menu appears there. And the line of code Larry added does the same.

@lawrencewin Perhaps we create a new accessibility test to check for the ExpandCollapse state (since it is the accessibility state). And then in the functional testing block, we only check if the menu is open by determining if the menu item is visible or not.

This way, we don't have 1 test checking for two separate conditions (accessibility state AND menu item visibility).

We want these tests to be as specific as possible. Right now, if this test fails, the developer has no idea if it failed because the ExpandCollapse state was wrong or if the menu item wasn't displayed.

Does that make sense?

apps/fluent-tester/src/TestComponents/Menu/consts.ts
```diff
@@ -16,3 +16,8 @@ export const MENU_ON_OPEN = 'Menu_On_Open'; // For testing the open functionalit
 export const MENU_ON_CLOSE = 'Menu_On_Close'; // For testing the open functionality of Menu
 
 export const MENU_DEFOCUS_BUTTON = 'Menu_Defocus_Button'; // Used to move focus away from menu button
+
+export const enum ExpandCollapseState {
```
PPatBoyd, line 20: Would it make sense to share the ExpandCollapseState values between all test components?

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
```diff
@@ -78,6 +74,10 @@ class MenuPageObject extends BasePage {
     return await this._primaryComponent;
```
PPatBoyd, line 74: driveby comment -- if this function always returns `await this._primaryComponent`, can we simplify or correct its purpose/add a use case?

PPatBoyd (approved):

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
PPatBoyd: @samuelfreiberg ahh yeah my comment was mostly directed at clarifying this function vs the test as a whole, I articulated that poorly.  I'm not totally sure of all interpretations I made while reading this PR, can you help clarifying for me for me how the `'Validate OnOpenChange() callback was fired -> Type "SPACE"'` test queries the menu item?

apps/fluent-tester/src/TestComponents/Menu/consts.ts
samuelfreiberg: Agreed with Pat. This should probably go in common/consts

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
```diff
@@ -78,6 +74,10 @@ class MenuPageObject extends BasePage {
...
   }
 
+  async resetTest() {
+    await (await this._defocusButton).click();
```
samuelfreiberg, line 78: Does this actually close the Menu? I just tried on the tester app and it didn't. Can you confirm?

samuelfreiberg (changes requested):

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
PPatBoyd: it should light dismiss off a click (testable :D) but also `escape` is a surefire menu dismissal

samuelfreiberg: What is a "light" dismiss?

PPatBoyd: @lawrencewin fyi for context; `light dismiss UI` is transient UI such as callouts and menus that have less screen presence/permanence than dialogs (transient UI that requires explicit user dismissal).  The most common action is a tap/click away from the light-dismiss UI, but ranges in behavior between features and apps. [Our Callout on Win32 currently offers](https://github.com/microsoft/fluentui-react-native/blob/6dea077f3b2fc93fc533128ecbc58d00cc9644ed/packages/components/Callout/src/Callout.types.ts#L29) `preventDismissOnKeyDown` and `preventDismissOnClickOutside` to disable these common light-dismiss interactions, which could be used in UI such as an @ mention picker or a teaching UI bubble (respectively) to allow them to stay on-screen longer for intentional behavior design or to avoid unintended user dismissals.

lawrencewin: > Does this actually close the Menu? I just tried on the tester app and it didn't. Can you confirm?

Works on my machine, though I did add an escape press on the menu trigger button. I also keep the click on the defocus button where only pressing escape somehow breaks the e2etests and resetting focus somehow fixes that.

apps/fluent-tester/src/TestComponents/Menu/consts.ts
ksiler: +1

ksiler (approved):

PR #2224, opened 2022-10-11
Update android token set

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Updating a new package version.

### Verification

Build successful.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ValentinaKozlova (approved):

rurikoaraki (approved):

packages/theming/win32-theme/jest.config.js
```diff
@@ -1,2 +1,2 @@
 const { configureReactNativeJest } = require('@fluentui-react-native/scripts');
-module.exports = configureReactNativeJest('android');
+module.exports = configureReactNativeJest('ios');
```
ankraj12, line 2: The pipeline still would error on other packages needing similar mitigation, but want to get an early opinion if there are concerns over updating the config to run on iOS. The root cause is that the global tokens for android is not a superset anymore, so the tests banking upon it would fail. 

The right way would be to enable tests for each platform, but adding a `win32` here errors in rex code - and I couldn't check deeper currently, but is there a better alternative while this fix happens?

ankraj12: @ayush547 is looking into something similar - could be a good candidate to fix together as needed.

PR #2226, opened 2022-10-11
Use selector chaining to improve element query performance

lawrencewin:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, querying elements starts its search at the top of the whole UIA tree. This works for the current tester app because the tree is small, but will fail in an environment where the tree is larger. 

This PR creates and designates a new root to start from in the search. It also modifies the `By` function in BasePage.ts to cache and save this root to chain queries to. 

### Verification

Ran existing E2E tests locally to ensure changes don't break anything. 

| image | image Added a new node (Fluent_Tester_Root_View) in tree to start element searches from.|

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


PPatBoyd:
for context with the PR, what's the # tests and speed up gain with this change?

samuelfreiberg:
> for context with the PR, what's the # tests and speed up gain with this change?

This won't speed anything up on the JS side. The speed gain will come from the Native side once the FluentTester is hosted in Lime. We don't have exact data on how fast yet, but it will be substantial (from my local testing)

apps/fluent-tester/src/E2E/Avatar/pages/AvatarPageObject.ts
```diff
@@ -12,29 +12,29 @@ export const enum AvatarComponentSelector {
 }
 class AvatarPageObject extends BasePage {
   async getPrimaryComponentAttribute(attribute: string): Promise<string> {
-    return await this._primaryComponent.getAttribute(attribute);
+    return await (await this._primaryComponent).getAttribute(attribute);
   }
 
   async getSecondaryComponentAttribute(attribute: string): Promise<string> {
-    return await this._secondaryComponent.getAttribute(attribute);
+    return await (await this._secondaryComponent).getAttribute(attribute);
```
ksiler, line 19: Did not having this extra `await` cause issues? How was it okay before?

apps/fluent-tester/src/E2E/common/BasePage.ts
```diff
@@ -1,13 +1,41 @@
...
 
 const DUMMY_CHAR = '';
+// The E2ETEST_PLATFORM environment variable should be set in the beforeSession hook in the wdio.conf file for the respective platform
+const PLATFORM = process.env['E2ETEST_PLATFORM'];
+if (PLATFORM) {
+  console.log(`Using platform: ${PLATFORM}`);
+}
 export const COMPONENT_SCROLL_COORDINATES = { x: -0, y: -100 }; // These are the offsets. Y is negative because we want the touch to move up (and thus it scrolls down)
 
+let _rootView: WebdriverIO.Element | null = null;
```
ksiler, line 13: I've never seen `null =  null` as a type declaration. Did you just mean `null`?

ksiler: Oh wait, you're defining the type and then immediately assigning it to `null`?

apps/fluent-tester/src/E2E/Avatar/pages/AvatarPageObject.ts
samuelfreiberg: We don't need to chain `await` commands, I don't believe this is necessary:

WDIO supports chaining commands without multiple `await`'s. See [this](https://blog.kiprosh.com/webdriverio-elements-chaining/)

samuelfreiberg: So I think we should remove all the chained awaits

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -224,7 +225,7 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
   };
 
   return (
-    <View style={commonTestStyles.flex}>
+    <View accessible={!enableSinglePaneView} testID={ROOT_VIEW} style={commonTestStyles.flex}>
```
ksiler, line 228: Shouldn't this view always be accessible if we're setting a testID on it and we want the testID to be discoverable on both mobile and non-mobile platforms?

ksiler: From what I recall (at least on win32), testID is only set if the component is accessible.

lawrencewin: When making this view accessible, this apparently breaks querying for elements on the iOS side, which is why the CI was continuously failing for this PR. Not entirely sure why.

samuelfreiberg: @Saadnajmi Any input here? Any idea why setting this View as accessible would break iOS testing?

@lawrencewin What is the exact error in the iOS testing when this is accessible?

lawrencewin: For whatever reason, having that view be accessible in iOS makes finding the header with testID='Base_TestPage' always fail. 

The problem is that the view also has to be accessible because it'll get optimized out in win32 if it isn't. I'm pretty sure this is a bug because the [react native docs](https://reactnative.dev/docs/view#testid) say that including a testID prop disables layout optimization.

apps/fluent-tester/src/E2E/Avatar/pages/AvatarPageObject.ts
lawrencewin:
```
async function QueryWithChaining(identifier) {
  if (_rootView === null) {
    // Most of the elements we're searching for will be children of this _rootView node.
    _rootView = await $('~' + ROOT_VIEW);
  }
  const selector = '~' + identifier;
  let queryResult: WebdriverIO.Element;
  queryResult = await _rootView.$(selector);
  if (queryResult.error) {
    // In some cases, such as opened ContextualMenu items, the element nodes are not children of the _rootView node, meaning we need to start our search from the top of the tree.
    queryResult = await $(selector);
  }
  return queryResult;
  ```
  The problem here is that elements such as the ContextualMenu appear outside of the root, meaning that the chained query can fail. I have this check that returns a query from the top of the tree if the chained query fails, but this means that i have to return a promise rather than webdriver's chainablepromise.
  
  A workaround could be to have another query method specifically for elements using ContextualMenu.

File: apps/fluent-tester/src/E2E/common/BasePage.ts (hunk 2 above)
Comment (lawrencewin) [reply apps/fluent-tester/src/E2E/common/BasePage.ts:13]: Yeah, just a type definition saying this is either an Element or null (on startup).

File: apps/fluent-tester/src/FluentTester.tsx (hunk 3 above)
Comment (ksiler) [reply apps/fluent-tester/src/FluentTester.tsx:228]: If we do have to make this distinction for mobile, I would create another boolean value to determine whether or not the view should be accessible rather than reuse `enableSinglePaneView` which is used for another purpose.

Comment (lawrencewin) [reply apps/fluent-tester/src/FluentTester.tsx]: Changed it to explicitly test whether the current platform is iOS.

Comment (Saadnajmi) [reply apps/fluent-tester/src/FluentTester.tsx]: I'm not sure... I guess our E2E tests (which use xc / xcii tests under the hood) are very tied to the accessibility implementation of React Native Core, so it.. kinda? makes sense that there could be a weird bug like this? Can we file a bug / issue about this so that it's easier to route someone to investigate?

File: apps/fluent-tester/src/E2E/Pressable/specs/Pressable.spec.win.ts [hunk 4]
```diff
@@ -10,6 +10,10 @@ describe('Pressable Testing Initialization', function () {
   });
 
   it('Click and navigate to Pressable test page', async () => {
+    /* Scroll to component test page button in scrollview if not already visible*/
+
+    await PressablePageObject.waitForButtonDisplayed(PAGE_TIMEOUT);
```
ksiler, line 15: The win32 E2E tests are failing from Pressable onward. I think this is because the buttons for the test pages in the test list are not in view. Do we need this call for `waitForButtonDisplayed`? Or can we just call `clickAndGoToPressablePage()` so that it automatically scrolls and clicks on the button for the test page?

apps/fluent-tester/src/E2E/common/BasePage.ts
```diff
@@ -1,13 +1,41 @@
...
+    case MobilePlatform.iOS:
+    case NativePlatform.Win32:
+      return await QueryWithChaining(identifier);
+    default:
+      return await $('~' + identifier);
+  }
+}
+
+async function QueryWithChaining(identifier) {
+  if (_rootView === null) {
```
ksiler, line 27: Any reason why this variable is prefixed with an underscore instead of just `rootView`?

apps/fluent-tester/src/E2E/Pressable/specs/Pressable.spec.win.ts
samuelfreiberg: +1 `await PressablePageObject.waitForButtonDisplayed(PAGE_TIMEOUT);` is not needed here (or anywhere else in win32/macos specs). It will always fail

apps/fluent-tester/src/E2E/common/BasePage.ts
lawrencewin: I initially underscored this variable since it's not exported out of the file, though I do see why it's not needed.

lawrencewin: Renamed.

apps/fluent-tester/src/E2E/common/BasePage.ts
```diff
@@ -1,13 +1,41 @@
...
+let rootView: WebdriverIO.Element | null = null;
+
 /* Win32/UWP-Specific Selector. We use this to get elements on the test page */
-export function By(identifier: string) {
-  return $('~' + identifier);
+export async function By(identifier: string) {
+  switch (PLATFORM) {
+    case MobilePlatform.iOS:
+    case NativePlatform.Win32:
+      return await QueryWithChaining(identifier);
```
samuelfreiberg, line 20: Does this QueryWithChaining only work for Win32? What about windows/macos?

apps/fluent-tester/src/E2E/common/BasePage.ts
```diff
@@ -61,34 +96,33 @@ export class BasePage {
...
 
   /* Returns true if the test page's button is displayed (the button that navigates to each test page) */
   async isButtonInView(): Promise<boolean> {
-    return await this._pageButton.isDisplayed();
+    return await (await this._pageButton).isDisplayed();
   }
 
   async clickComponent(): Promise<void> {
-    await this._primaryComponent.click();
+    await (await this._primaryComponent).click();
```
samuelfreiberg, line 108: Are we sure we need these extra `await`'s? https://blog.kiprosh.com/webdriverio-elements-chaining/

samuelfreiberg:
Why is ios/Podfile.lock being changed? Is this correct?

apps/fluent-tester/src/E2E/common/BasePage.ts
lawrencewin: For UWP apps, `QueryWithChaining` always falls back to searching from the top of the tree because `Fluent_Tester_Root_View` can't be found within the tree. I've tried to test this, but I can't seem to build the UWP tester app. MacOS seems fine however.

lawrencewin:
> Why is ios/Podfile.lock being changed? Is this correct?

Pretty sure this is leftover from testing on my own mac. Will remove.

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
```diff
@@ -93,11 +93,11 @@ class MenuPageObject extends BasePage {
...
 
   get _secondaryComponent() {
-    return By(MENUITEM_TEST_COMPONENT);
+    return By(MENUITEM_TEST_COMPONENT, true);
   }
 
   get _tertiaryComponent() {
-    return By(MENUITEM_NO_A11Y_LABEL_COMPONENT);
+    return By(MENUITEM_NO_A11Y_LABEL_COMPONENT, true);
   }
```
lawrencewin, line 101: @samuelfreiberg Calls to `By` shouldn't need an extra `await` anymore, but in return, users have to specify if they want to start from the root of the element tree. Here, we have to explicitly say to start from the root because we're working with menu items. Is this approach better than the previous?

apps/fluent-tester/src/E2E/common/BasePage.ts
lawrencewin: Addressed here: https://github.com/microsoft/fluentui-react-native/pull/2226#discussion_r1001000310

apps/fluent-tester/src/E2E/Menu/pages/MenuPageObject.ts
samuelfreiberg: I don't think this is better than previous. I'd rather have two awaits. But still confused - why can't we use: 
`await this._primaryComponent.click();`
instead of 
`await (await this._primaryComponent).click();`

Doesn't WDIO handle this correctly?

lawrencewin: When you call `$` alone, you don't need that extra await since `$` returns a `ChainablePromiseElement` which allows for `await this._primaryComponent.click()`.

With our queries, we start from an element called `Fluent_Tester_Root_View` which encompasses **most** of the elements in the test app we're searching for. Keyword is **most**. In the case of MenuItems, these spawn outside of `Fluent_Tester_Root_View`. For chained queries, searching for MenuItems returns null, causing an error.

Thus, in our query function, we have to check for this edge case. ```ts
let queryResult: WebdriverIO.Element;
  queryResult = await _rootView.$(selector);
  if (queryResult.error) {
    // In some cases, such as opened ContextualMenu items, the element nodes are not children of the _rootView node, meaning we need to start our search from the top of the tree.
    queryResult = await $(selector);
  }
  return queryResult;
```
Because we have to check if the first query returns null, we have to let the first query resolve into an element. This unfortunately means that we have to work in an async function, and thus can only return a `Promise<Element>` rather than a `ChainablePromiseElement`.

File: apps/fluent-tester/src/E2E/common/BasePage.ts [hunk 9]
```diff
@@ -1,13 +1,39 @@
-import { Keys } from './consts';
+import { Keys, ROOT_VIEW } from './consts';
 import { TESTPAGE_BUTTONS_SCROLLVIEWER } from '../../TestComponents/Common/consts';
 import { Attribute } from './consts';
 
 const DUMMY_CHAR = '';
+// The E2ETEST_PLATFORM environment variable should be set in the beforeSession hook in the wdio.conf file for the respective platform
+const PLATFORM = process.env['E2ETEST_PLATFORM'];
+if (PLATFORM) {
```
samuelfreiberg, line 8: Is this still needed? Leftover from testing? When running a test, you need to specify a platform-specific wdio file, so I I don't think this adds value.

lawrencewin: This is for getting the current running platform within code and the page objects. Is there an existing way of doing this with wdio?

apps/fluent-tester/src/E2E/common/BasePage.ts
```diff
@@ -1,13 +1,39 @@
...
+async function QueryWithChaining(identifier) {
+  if (rootView === null) {
+    // Most of the elements we're searching for will be children of this rootView node.
+    rootView = await $('~' + ROOT_VIEW);
+  }
+  const selector = '~' + identifier;
+  let queryResult: WebdriverIO.Element;
+  queryResult = await rootView.$(selector);
+  if (queryResult.error) {
+    // In some cases, such as opened ContextualMenu items, the element nodes are not children of the rootView node, meaning we need to start our search from the top of the tree.
```
samuelfreiberg, line 33: Can we find the parent of these ContextualMenu's? They must have some parent. I'm assuming this is specific to Callouts.

@PPatBoyd Are Callout's not apart of the same root window tree? It seems as though it has a different root completely.

@lawrencewin We should be able to find the parent. Once we do, we should try to set that as the ROO_VIEW for that specific `By()` call.

If we have to search for each callout from the root, the performance will be very bad on larger UI trees

samuelfreiberg (changes requested):

apps/fluent-tester/src/E2E/common/BasePage.ts
samuelfreiberg: Sorry, I meant the if statement + console.log on line 8-10.

apps/fluent-tester/src/E2E/common/BasePage.ts
lawrencewin: image
ContextualMenuItems and the rest of the app share a parent in the Tester Window, but there isn't an automation ID I can use to query parent items.

samuelfreiberg: Perhaps we can use a [different selector strategy ](https://webdriver.io/docs/selectors/)to find this "FluentUI Tester" window component?

If we can, that would be the most ideal solution, and it should work for all platforms.

apps/fluent-tester/src/E2E/ContextualMenu/specs/ContextualMenu.spec.win.ts
```diff
@@ -1,7 +1,6 @@
 import NavigateAppPage from '../../common/NavigateAppPage';
 import ContextualMenuPageObjectObject, { ContextualMenuSelector } from '../pages/ContextualMenuPageObject.win';
 import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT, Keys } from '../../common/consts';
-
```
ksiler, line 4: nit: revert this extra line deletion

apps/fluent-tester/src/E2E/common/BasePage.ts
lawrencewin: @samuelfreiberg @ksiler Okay so after some experimentation with xpath, class name, and name, it looks like there isn't a single selector strategy that can work across all platforms.

With that said, it is definitely possible to find the root Fluent Tester window element if we write platform specific webdriver code. I've had success locating the window on win32 and uwp. Have not tested on iOS, mac, or android. It's janky, but it allows us to remove chained awaits and removes the conditional for ContextualMenu.

apps/fluent-tester/src/E2E/common/BasePage.ts
samuelfreiberg: Friendly bump on this

lawrencewin: Fixed :)

samuelfreiberg:
Looking at the Win32 E2E CI run, ctrl+f for "Fluent_Tester_Root_View" and you'll see a ton of queries for it. Is there anyway we can "save" the result from the first time we find it rather than re-searching everything we call `By()`?

lawrencewin:
> Looking at the Win32 E2E CI run, ctrl+f for "Fluent_Tester_Root_View" and you'll see a ton of queries for it. Is there anyway we can "save" the result from the first time we find it rather than re-searching everything we call `By()`?

So my immediate thought was to save the reference to the "Fluent_Tester_Root_View" between tests, but the issue is that each time a new spec is run, a new session is created. Element References for a given element differ between sessions so it looks like it isn't realistically possible to save this across different tests.

samuelfreiberg:
> > Looking at the Win32 E2E CI run, ctrl+f for "Fluent_Tester_Root_View" and you'll see a ton of queries for it. Is there anyway we can "save" the result from the first time we find it rather than re-searching everything we call `By()`?
> 
> So my immediate thought was to save the Element reference to the "Fluent_Tester_Root_View" between tests.
> 
> The issue is that each time a new spec is run, a new session is created. Element References for the same element differ between sessions so it looks like it isn't realistically possible to save this across different tests.

Sounds good. Are you sure it's saved within each spec? Just want to make sure, there are a ton of hits when ctrl+f

lawrencewin:
> > > Looking at the Win32 E2E CI run, ctrl+f for "Fluent_Tester_Root_View" and you'll see a ton of queries for it. Is there anyway we can "save" the result from the first time we find it rather than re-searching everything we call `By()`?
> > 
> > 
> > So my immediate thought was to save the Element reference to the "Fluent_Tester_Root_View" between tests.
> > The issue is that each time a new spec is run, a new session is created. Element References for the same element differ between sessions so it looks like it isn't realistically possible to save this across different tests.
> 
> Sounds good. Are you sure it's saved within each spec? Just want to make sure, there are a ton of hits when ctrl+f

Yeah. From looking at the logs, there's a lot of hits because each time the test starts and the driver queries for the root, the app hasn't fully loaded yet. Once it finds the root, the test seems to run fine.

samuelfreiberg (approved):

apps/fluent-tester/src/FluentTester.tsx
lawrencewin: Filed a issue, will resolve

PR #2228, opened 2022-10-12
Add @fluentui-react-native/dependency-profiles package

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add @fluentui-react-native/dependency-profiles package to store and update dependency profiles for FURN for users of [@rnx-kit/align-deps](https://github.com/microsoft/rnx-kit/tree/main/packages/align-deps). 

A dependency profile is a mapping of npm packages, typically at a react-native version. This dependency profile provides a mapping of all the latest packages FURN publishes for react-native `0.68`. Other packages or repositories who depend on FURN can use this dependency profile to update their FURN dependencies to latest.

When we bump the react-native version for this repo in the future, we should "freeze" the old dependency profile for the last version to a file (like `furn-dependencies-0.68.js` and include that in our profile definition in `update-profile.js`. A comment was added there with instructions.

### Verification

Ran `yarn update-profile` under @fluentui-react-native/dependency-profiles and verified it updates the dependency profile with all the latest package versions for FURN packages. This can be run manually but is also run whenever we bump packages to new versions with beachball so the dependency profile stays up to date.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/dependency-profiles/src/index.js
```diff
@@ -0,0 +1,314 @@
...
+      "version": "^0.68.0"
+    },
+    "core-macos": {
+      "name": "react-native-macos",
+      "version": "^0.68.0"
+    },
+    "core-windows": {
+      "name": "react-native-windows",
+      "version": "^0.68.0"
+    },
```
tido64, line 20: I don't think you want to add these. They overwrite consumers' react-native version. Instead, if you want people to use 0.68, you should rename `latest` to `0.68`. This will tell dep-check to append your list of dependencies to the profile for react-native 0.68.

packages/dependency-profiles/src/index.js
```diff
@@ -0,0 +1,314 @@
+// This file was generated by 'build.js'
+/* eslint-disable */
+module.exports = {
+  "latest": {
```
tido64, line 4: Do we still support users on other versions of react-native?

packages/dependency-profiles/src/index.js
ksiler: Thanks! I was confused on whether or not to add these. That makes sense. Do you see any issue with just maintaining a `latest` profile instead of a profile for specific react-native versions?

tido64: Currently, consumers select a profile to align against, e.g. to ensure all your dependencies are compatible with react-native 0.68, one would typically pick the profile named "0.68". The way you extend/override the profile is to have your own custom profile whose name matches the profile you want to extend. This means that if you want to add FURN packages to the 0.68 profile, you need to also name it "0.68". `latest` doesn't exist so it would just become a standalone profile.

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,99 @@
+{
+  "name": "@fluentui-react-native/dependency-profiles",
+  "version": "0.0.0",
+  "description": "@rnx-kit/dep-check profiles covering packages published from FluentUI-React-Native",
+  "license": "UNLICENSED",
```
rurikoaraki, line 5: We should have a license defined. I think most packages use MIT?

packages/dependency-profiles/src/index.js
ksiler: I see. I've updated this to define profiles at a specific react-native version. Since we don't maintain older releases, I didn't add any logic to maintain older RN profiles. Instead, I added a comment in the `update-profile` script instructing maintainers to "freeze" the old profile to a new file when updating to a newer RN version and append any older profiles to the list.

packages/dependency-profiles/src/index.js
ksiler: We'll start with RN `0.68` and "freeze" older profiles as we update RN versions since we don't currently maintain older releases.

packages/dependency-profiles/package.json
ksiler: Fixed. Thanks!

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,100 @@
+{
+  "name": "@fluentui-react-native/dependency-profiles",
+  "version": "0.0.0",
```
samuelfreiberg, line 3: Should this start at a non-zero number?

ksiler: Good question. @rurikoaraki do you know? I think I've seen new packages start at `0.0.0` before but also `0.1.0`.

rurikoaraki: Either is fine. Beachball will bump it to a reasonable number as soon as it gets a chance to run

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "src/*"
+  ],
+  "main": "src/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/dependency-profiles"
+  },
+  "scripts": {
+    "build": "fluentui-scripts build",
```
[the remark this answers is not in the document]
ksiler: Yes, I was planning to add a `postBump` hook to beachball.config.js to run `yarn update-profile` when new versions of FURN packages are published but I still need to test it.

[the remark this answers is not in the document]
ksiler: On second thought, that may not work. Still thinking about it.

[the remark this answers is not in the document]
tido64: I think that's what @acoates-ms did internally. Might wanna check in with him.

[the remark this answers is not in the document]
ksiler: Thanks, I noted that but didn't fully understand how it worked. Now that I understand it better, I added a `postBump` hook to the config to update the profile.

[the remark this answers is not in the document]
ksiler: Yes, that's my understanding.

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,100 @@
+{
+  "name": "@fluentui-react-native/dependency-profiles",
+  "version": "0.0.0",
+  "description": "@rnx-kit/dep-check profiles covering packages published from FluentUI-React-Native",
```
tido64, line 4: Just a heads up: `@rnx-kit/dep-check` is being renamed to `@rnx-kit/align-deps`. This package will still work as is.

tido64 (approved):

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "@uifabricshared/foundation-tokens": "*",
+    "@uifabricshared/theme-registry": "*",
+    "@uifabricshared/themed-settings": "*",
+    "@uifabricshared/theming-ramp": "*",
+    "@uifabricshared/theming-react-native": "*",
+    "@office-iss/react-native-win32": ">=0.68.0 <0.68.6",
+    "react-native": "^0.68.0",
+    "react-native-macos": "^0.68.0",
+    "react-native-svg": "12.3.0",
+    "react-native-windows": "^0.68.0",
```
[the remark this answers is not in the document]
ksiler: Thanks for the catch. I forgot to remove these when I removed the react-native-* capabilities from the profile.

packages/dependency-profiles/package.json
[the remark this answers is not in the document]
Saadnajmi: @lyzhan7 FYI, I think the bug to update our podfile.locks is assigned to you. You could do a similar strategy here.

packages/dependency-profiles/src/index.js
```diff
@@ -0,0 +1,302 @@
+// This file was generated by 'update-profile.js'
+/* eslint-disable */
+module.exports = {
+  "0.68": {
+    "svg": {
+      "name": "react-native-svg",
+      "version": "12.3.0"
+    },
```
ksiler, line 8: Unsure if we need to specify this dependency as part of our profile. I added it in because we are pinned to a specific version of react-native-svg in our repo but that won't be the case forever. I'm considering removing it for now.

Saadnajmi: I think eventually we might want/need it since a lot of our icons are SVGs. But yeah makes sense to not add it just yet!

packages/dependency-profiles/src/index.js
```diff
@@ -0,0 +1,298 @@
...
+    "@fluentui-react-native/text": {
+      "name": "@fluentui-react-native/text",
+      "version": "0.15.8"
+    },
+    "@uifabricshared/foundation-composable": {
+      "name": "@uifabricshared/foundation-composable",
+      "version": "0.11.1"
+    },
+    "@uifabricshared/foundation-compose": {
```
PPatBoyd, line 101: note for future versions, are we able to / should we mark the @uifabricshared packages as deprecated, or not include them entirely?

packages/dependency-profiles/src/index.js
```diff
@@ -0,0 +1,298 @@
...
+  "0.68": {
+    "@fluentui-react-native/tester": {
+      "name": "@fluentui-react-native/tester",
+      "version": "0.103.0"
+    },
+    "@fluentui-react-native/tester-win32": {
+      "name": "@fluentui-react-native/tester-win32",
+      "version": "0.27.18"
+    },
+    "@fluentui-react-native/codemods": {
```
PPatBoyd, line 13: nit should codemods be dev only?

PPatBoyd (approved):

packages/dependency-profiles/src/index.js
PPatBoyd: it's actually a part of the base dependency profile for 0.68 already, even https://github.com/microsoft/rnx-kit/blob/49412e1cc1677275cd3ecfafd960314a685b0693/packages/dep-check/src/presets/microsoft/profile-0.68.ts#L105 which to me is a bit odd but mostly because the mainline rn-svg is up to version 13.4.0

Saadnajmi: I think these are mostly manually updated by people who have used it and know what works (The "magic" part of rep-check is just manual verification..). I guess nobody we know has tried to use rn-svg 13.4.0 yet so it hasn't been updated? FURN breaks with 12.40+ so we personally haven't upgraded. That probably informs the dependencies of our downstream partners.

packages/dependency-profiles/src/index.js
Saadnajmi: I want to say this is the goal of having all the V1 components. I believe Ruriko mentioned to me she wants to write a codemod to mark all the `@uifabricshared` stuff as deprecated in the future. Main holdout still using the old frameworks in my head is Callout..

packages/dependency-profiles/src/index.js
tido64: > I guess nobody we know has tried to use rn-svg 13.4.0 yet so it hasn't been updated?

This is mostly the case. Just from looking at the changelog, it also looks like they broke backwards compatibility in 13.x. 13.4 is supposed to restore support for react-native older than 0.68. If FURN does upgrade to 13.x, let us know and we can update the recommendations.

packages/dependency-profiles/package.json
```diff
@@ -0,0 +1,96 @@
+{
+  "name": "@fluentui-react-native/dependency-profiles",
+  "version": "0.0.0",
```
gulnazsayed, line 3: Is this supposed to be version 0?

ksiler: Yeah, I've seen new packages start at `0.0.0` or `0.1.0`. I don't think it matters because Beachball will bump it to a reasonable number as soon as it gets a chance to run (after this PR is merged).

PR #2229, opened 2022-10-12
Make `Theme.spacing` read-only

amgleitman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Make `Theme.spacing` read-only. Since spacing tokens are global, we really shouldn't be overriding them.

### Verification

Validated that `yarn build` still works, which means that `tsc` isn't complaining about any attempts to edit any `Theme`'s `spacing` prop.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/theme-types/src/Theme.types.ts
+++ b/packages/theming/theme-types/src/Theme.types.ts
@@ -25,7 +25,7 @@ export interface Theme {
     [key: string]: object; // eslint-disable-line @typescript-eslint/ban-types
   };
   shadows: ThemeShadowDefinition;
-  spacing: Spacing;
+  readonly spacing: Spacing;
   host: {
     // appearance of the theme, this corresponds to the react-native Appearance library values, though can be overwritten
     // dynamic refers to a theme that handles it's own appearance switching, such as one that uses the PlatformColor API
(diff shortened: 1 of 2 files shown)


lyzhan7 (approved):

rurikoaraki (approved):

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -25,7 +25,7 @@ export interface Theme {
     [key: string]: object; // eslint-disable-line @typescript-eslint/ban-types
   };
   shadows: ThemeShadowDefinition;
-  spacing: Spacing;
+  readonly spacing: Spacing;
```
ankraj12, line 28: Apologies for putting this up in a merge code, but I want to understand if this can break an existing partner overriding Spacing till now...? Do we have any that we are aware.?

rurikoaraki: Post-merge comments are fine @ankraj12

While yes, this change is technically breaking, I don't think anyone really uses the spacing values from the Theme or overrides them so I think this is fine. We can always limit the readonly to just the new entries if we get feedback.

PR #2231, opened 2022-10-13
Add stroke width tokens test page

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Add test page to show the global stroke width tokens

Note - is it expected that stroke width 0.5 doesn't show up on macOS?

### Verification

https://user-images.githubusercontent.com/78454019/201436142-639b81e9-ca8f-4f4d-8dfd-c6e6d6448eec.mp4

https://user-images.githubusercontent.com/78454019/201436146-795c4e01-4edb-483d-bf73-f85dd9599e0e.mp4

https://user-images.githubusercontent.com/78454019/201436161-56fc41b9-1a03-47e2-b73f-39303e2a7c56.mov

https://user-images.githubusercontent.com/78454019/201436253-c345b88c-3738-4c36-8e91-1824e9a8b8c6.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <View style={styles.root}>
+      <Text variant="bodySemibold">{props.name + ': ' + props.strokeWidth + 'pt'}</Text>
+      <View style={styles.exampleStrokeBackground}>
+        <View style={exampleStrokeStyle} />
+      </View>
+    </View>
+  );
+};
```
lyzhan7, line 38: Still trying to understand how to make sure components aren't getting rendered too many times. Is this component okay? Does exampleStrokeStyle need to be memoized?

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -40,6 +50,9 @@ export interface Theme {
   };
   shadows: ThemeShadowDefinition;
   readonly spacing: Spacing;
+  readonly stroke: {
```
ankraj12, line 53: I am trying to understand this - why do we need to add the strokes to theme object ( instead of accessing using `globalTokens.stroke.thin `as other endpoints currently). My understanding of theme was it's supposed to be the set of customisations for an end user ( aliases primarily as globals are immutable in design), so why do we want to add this?

ankraj12 (commented):
Please help review the scenario of adding strokes in theme object.

packages/theming/theme-types/src/Theme.types.ts
lyzhan7: My understanding is that going forward we don't want to reference global tokens directly and instead go through the theme object, and for tokens we don't want the user to be able to customize, we'll mark them as readonly. This PR is based off the recent work to add iOS spacing tokens. I think @amgleitman or @rurikoaraki have more background behind this decision

rurikoaraki: Spacing is different because it already exists on the Theming object, so it would be weird for it to be haphazardly missing some spacing tokens. I asked @amgleitman to mark them readonly to make clear that they are unable to be messed with, to align with how FluentUI libraries view global tokens.

However, in general I would rather we use global tokens directly to avoid bloating the Theme object.

apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
rurikoaraki: If exampleStrokeStyle  is taken outside of the component, it's fine. 

The way React components figure out if it needs to rerender is by shallow comparing all of its props and if anything is different, it will rerender the component. For objects (objects, arrays, functions), that means that if a new instance of that object is created and passed into the compare, then the shallow compare will fail because it's a different instance, even if the contents of the object are the same and so in reality no work needed to be done. You can think of it as a pointer comparison if you're used to native.

Anything inside of a (function) component is rerun when something asks components to rerender. So that means any instantiation of a const object will be redone, and new instances of those objects will be created. In this case `exampleStrokeStyle` has a new instance on every render and so the inner most view will be invalidated on every render.

Moving the const object outside the components works because that code is only run once.

rurikoaraki (commented):
Posting a comment

lyzhan7:
Reverted the work that was getting the stroke width tokens through the theme
Updated to get the tokens through the globalTokens object. 
Updated the description

Saadnajmi:
Are the videos updated? If so, could we have the dark mode examples be a light separator on a dark background?

lyzhan7:
> Are the videos updated? If so, could we have the dark mode examples be a light separator on a dark background?

Added dark mode specific colors + updated the videos

apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
```diff
@@ -6,17 +6,24 @@ import { Stack } from '@fluentui-react-native/stack';
...
-  },
+const getThemedStyles = themedStyleSheet((t: Theme) => {
+  const isLightMode = getCurrentAppearance(t.host.appearance, 'light') === 'light';
+  return {
+    root: {
+      paddingBottom: 16,
+    },
+    exampleStrokeBackground: {
+      height: 100,
+      backgroundColor: isLightMode ? globalTokens.color.grey80 : globalTokens.color.grey10,
```
rurikoaraki, line 21: nit: alias token?

lyzhan7: I couldn't find an alias token that worked like I wanted to for iOS/macOS and dark/light mode...the next thing on my plate is iOS color tokens, so maybe once that's done there will be a color I can use that works for all platforms. Is there overall a preference to refer to alias tokens over global tokens?

rurikoaraki: For color, yes. It should be pretty rare to reference global colors esp if you're using neutrals. At the moment this test won't show well on HC or win32 themes, where alias tokens would take care of that.

rurikoaraki: It's probably not a blocker though since it's only a test app and for showing tokens.

rurikoaraki:
Same suggestion to add E2E test page

apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
lyzhan7: Found one that works okay. Will update to corner radius tokens PR as well. Updated the screenshots in the description

apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
```diff
@@ -7,18 +7,16 @@ import { stackStyle } from '../Common/styles';
...
 const getThemedStyles = themedStyleSheet((t: Theme) => {
-  const isLightMode = getCurrentAppearance(t.host.appearance, 'light') === 'light';
   return {
     root: {
       paddingBottom: 16,
     },
     exampleStrokeBackground: {
       height: 100,
-      backgroundColor: isLightMode ? globalTokens.color.grey80 : globalTokens.color.grey10,
+      backgroundColor: t.colors.bodyStandoutBackground,
```
rurikoaraki, line 19: Sorry @lyzhan7 but this isn't an alias token 😓 this is an older color that we'll want to remove at some point. You should take a color from [fluentui-react-native/Color.types.ts at f14e559f197dc2c1b7631de0100aa30742648544 · microsoft/fluentui-react-native (github.com)](https://github.com/microsoft/fluentui-react-native/blob/f14e559f197dc2c1b7631de0100aa30742648544/packages/theming/theme-types/src/Color.types.ts#L453)

rurikoaraki: Same with bodyText

lyzhan7:
The win32 e2e tests are failing the test "Validate the children of ScrollView are the buttons that navigate to each test page with proper automationId" - ([run here](https://dev.azure.com/ms/ui-fabric-react-native/_build/results?buildId=397574&view=logs&j=b330ab76-88a4-5f6b-0b60-6040ed7b6063&t=2d8b0817-fa8e-5ef8-a24f-ee0fcf5b4ca5)) 

It's not obvious to me what the issue is, currently looking into it

rurikoaraki:
> The win32 e2e tests are failing the test "Validate the children of ScrollView are the buttons that navigate to each test page with proper automationId" - ([run here](https://dev.azure.com/ms/ui-fabric-react-native/_build/results?buildId=397574&view=logs&j=b330ab76-88a4-5f6b-0b60-6040ed7b6063&t=2d8b0817-fa8e-5ef8-a24f-ee0fcf5b4ca5))
> 
> It's not obvious to me what the issue is, currently looking into it

@samuelfreiberg and @lawrencewin FYI if you could help @lyzhan7 out that would be appreciated

apps/fluent-tester/src/TestComponents/StrokeWidth/StrokeWidthTest.tsx
lyzhan7: Ah okay, I don't think any of the aliascolortokens exist on iOS (yet). I can either just update this right now to use one of those tokens (and have a temporary special case for iOS using the globalToken values I had), or just use globalTokens for now and come back to this after I finish the color token work for ios

rurikoaraki: Yeah you can use globals for now that's fine. Thanks for working through my comments!

lyzhan7: Okay, I'll go back to global tokens for now. Made an ADO item to come back to these tester pages as part of the iOS color token work

rurikoaraki:
@lyzhan7 in case you haven't seen this yet: https://github.com/microsoft/fluentui-react-native/tree/main/apps/fluent-tester/src/E2E#authoring-e2e-test

lyzhan7:
> 

Yep I've been following that documentation (did the corner radius e2e tests with that). This PR doesn't have any e2e tests yet though so I'm not sure how much of it is relevant. Maybe new components that have a new test page need e2e tests? But not sure about that because I don't think Shadow has an e2e test yet

rurikoaraki:
> > 
> 
> Yep I've been following that documentation (did the corner radius e2e tests with that). This PR doesn't have any e2e tests yet though so I'm not sure how much of it is relevant. Maybe new components that have a new test page need e2e tests? But not sure about that because I don't think Shadow has an e2e test yet

@lawrencewin is pushing for all pages to have an E2E test for automation coverage, so all pages will eventually be covered by an E2E of some sort.

samuelfreiberg:
@rurikoaraki  @lyzhan7  The E2E tests are failing because the wrong constant was set in apps/fluent-tester/src/testPages.ts.

The `testPage` attribute should be set to the test page button, which in this case, is `HOMEPAGE_STROKEWIDTH_BUTTON`, not `STROKEWIDTH_TESTPAGE`

lyzhan7:
> @rurikoaraki @lyzhan7 The E2E tests are failing because the wrong constant was set in apps/fluent-tester/src/testPages.ts.
> 
> The `testPage` attribute should be set to the test page button, which in this case, is `HOMEPAGE_STROKEWIDTH_BUTTON`, not `STROKEWIDTH_TESTPAGE`

Thanks Sammy - totally missed that, just fixed.

Saadnajmi:
Re: stroke width 0.5 doesn't show up on macOS. I guess that depends on your monitors dpi. Could we test it at a higher dpi (by switching to a lower resolution on the same machine)? Happy to also help offline. 

I also don't consider this blocking since 0.5 is really only there for iOS.

apps/fluent-tester/src/testPages.ts
```diff
@@ -27,6 +27,7 @@ import { RadioGroupExperimentalTest, HOMEPAGE_RADIO_GROUP_EXPERIMENTAL_BUTTON }
 import { SeparatorTest, HOMEPAGE_SEPARATOR_BUTTON } from './TestComponents/Separator';
 import { SpacingTokensTest, HOMEPAGE_SPACING_BUTTON } from './TestComponents/Spacing';
 import { ShadowTest, HOMEPAGE_SHADOW_BUTTON } from './TestComponents/Shadow';
+import { StrokeWidthTest, HOMEPAGE_STROKEWIDTH_BUTTON } from './TestComponents/StrokeWidth';
```
rurikoaraki, line 30: nit: alphabetically stroke comes after shimmer

rurikoaraki (approved):

lyzhan7:
> Re: stroke width 0.5 doesn't show up on macOS. I guess that depends on your monitors dpi. Could we test it at a higher dpi (by switching to a lower resolution on the same machine)? Happy to also help offline.
> 
> I also don't consider this blocking since 0.5 is really only there for iOS.

I tried changing the resolution of my display & also trying different displays but haven't gotten it to show up. I'll log an issue to look into it

PR #2234, opened 2022-10-13
Fix for iOS bug where switching light/dark mode was not updating the theme

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

While adding in iOS Shadow Tokens, I noticed that the theme wasn't always getting updated when switching between light and dark mode. It would get updated only when refreshing the app. As a result I would see the light mode shadows in dark mode or vice versa.

After putting in some console.log statements I noticed that:
- the theme was getting invalidated at the right times
- createAppleTheme.iOS.ts/baseAppleTheme() was also getting called at the right times
- however, paletteFromAppleColors(), appleTypography(), etc. was not getting called when they should be

Replacing the BaseAppleLightThemeIOS/BaseAppleDarkThemeIOS with method calls (ex. with getBaseAppleLightThemeIOS/getBaseAppleDarkThemeIOS) fixed the problem. This is more similar to what macOS does as well. I'm actually not exactly sure why the previous code did not work (maybe the BaseAppleLightThemeIOS object was getting cached somehow?) - if anyone has any insight into why, I'd love to know.

I also went ahead and simplified the logic for light mode/dark mode to use the same method (i.e. getBaseAppleThemeIOS), which also simplified some of the logic in createAppleTheme.ios.ts. I changed the parameter in paletteFromAppleColors to be isLight instead of isDark for consistency.

### Verification

Checked that after making the change, console.log statements in paletteFromAppleColors/appleTypography were getting called as expected.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/apple-theme/src/createAppleTheme.ios.ts
+++ b/packages/theming/apple-theme/src/createAppleTheme.ios.ts
@@ -1,15 +1,12 @@
 import { ThemeReference } from '@fluentui-react-native/theme';
 import { Theme } from '@fluentui-react-native/theme-types';
 import { Appearance } from 'react-native';
-import { BaseAppleDarkThemeIOS, BaseAppleLightThemeIOS } from './appleTheme.ios';
+import { getBaseAppleThemeIOS } from './appleTheme.ios';
 
 export function createAppleTheme(): ThemeReference {
-  const baseAppleTheme = () => {
-    const current = Appearance.getColorScheme();
-    return current === 'light' ? BaseAppleLightThemeIOS : BaseAppleDarkThemeIOS;
-  };
-
-  const appleThemeReference = new ThemeReference({} as Theme, baseAppleTheme);
+  const appleThemeReference = new ThemeReference({} as Theme, () => {
+    return getBaseAppleThemeIOS(Appearance.getColorScheme() === 'light');
+  });
 
   Appearance.addChangeListener(() => {
     appleThemeReference.invalidate();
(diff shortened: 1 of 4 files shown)


chiuam (approved):

lyzhan7:
Update - so I realized something in an unrelated conversation and now my current understanding on why the previous code wasn't working is that the BaseAppleLightThemeIOS and BaseAppleDarkThemeIOS objects were only getting created once (sometime at the start but I don't really know exactly when - is it just the first time they get used?). So the values in BaseAppleLightThemeIOS and BaseAppleDarkThemeIOS are based off of whatever appearance mode we were in when those objects got created (because there's a call checking for dark mode in appleTypography and the other methods). I originally thought that the baseAppleTheme() method was returning the wrong object, but now I realize it was returning the right object, but the object has the wrong values (ex. both the BaseAppleLightThemeIOS and BaseAppleDarkThemeIOS had dark mode shadows, or both had light mode shadows, depending on if we were in light mode or dark mode when those objects got created)

packages/theming/apple-theme/src/createAppleTheme.ios.ts
```diff
@@ -1,15 +1,12 @@
...
 export function createAppleTheme(): ThemeReference {
-  const baseAppleTheme = () => {
-    const current = Appearance.getColorScheme();
-    return current === 'light' ? BaseAppleLightThemeIOS : BaseAppleDarkThemeIOS;
-  };
-
-  const appleThemeReference = new ThemeReference({} as Theme, baseAppleTheme);
+  const appleThemeReference = new ThemeReference({} as Theme, () => {
+    return getBaseAppleThemeIOS(Appearance.getColorScheme() === 'light');
+  });
```
Saadnajmi, line 9: Could we keep the behavior above of having isLight be its own variable? Helps with future debugging :)

Saadnajmi (approved):

PR #2241, opened 2022-10-14
Initial Android Shadow Implementation

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

This PR adds the initial implementation of Fluent shadows for Android. This is done by just mapping the key shadow's blur to the Android elevation prop. 

I also intended this PR to do the "margin extraction / ripple constraining" logic that #2103 needs, but will probably factor that out. 

### Verification

Shadow test page on Android works, and notification component has shadows

Screenshot 2022-10-14 at 11 59 37 AM
Screenshot 2022-10-14 at 11 41 50 AM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Shadow/src/shadowStyle.android.ts
+++ b/packages/experimental/Shadow/src/shadowStyle.android.ts
@@ -1,0 +1,25 @@
+import { memoize } from '@fluentui-react-native/framework';
+import { ShadowToken } from '@fluentui-react-native/theme-types';
+
+/**
+ * For Android, we do not set two independent shadows on two views.
+ * Instead, the `elevation` prop takes care of synthesizing a key and ambient
+ * shadow together with it's own depth ramp. The `blur` token of our key shadow roughly
+ * matches the elevation value. See the following:
+ * https://material.io/design/environment/light-shadows.html#light
+ * https://github.com/microsoft/apple-ux-guide/blob/gh-pages/Shadows.md
+ *
+ * We still however, want to have two views in our heirarchy so that the "ambient" shadow
+ * constrains the Android Ripple's borders due to the following bug:
+ * https://stackoverflow.com/questions/63048178/ripple-effect-leaking-at-corners-as-if-pressable-button-has-a-borderradius
+ */
+export const getShadowTokenStyleSet = memoize(getShadowTokenStyleSetWorker);
+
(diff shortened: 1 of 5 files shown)


packages/experimental/Shadow/src/shadowStyle.android.ts
```diff
@@ -0,0 +1,25 @@
+import { memoize } from '@fluentui-react-native/framework';
+import { ShadowToken } from '@fluentui-react-native/theme-types';
+
+/**
+ * For Android, we do not set two independent shadows on two views.
+ * Instead, the `elevation` prop takes care of synthesizing a key and ambient
+ * shadow together with it's own depth ramp. The `blur` token of our key shadow roughly
```
lyzhan7, line 7: 'its' not 'it's'

lyzhan7:
Could also test Android FAB

packages/experimental/Shadow/src/shadowStyle.android.ts
```diff
@@ -0,0 +1,25 @@
...
+ * matches the elevation value. See the following:
+ * https://material.io/design/environment/light-shadows.html#light
+ * https://github.com/microsoft/apple-ux-guide/blob/gh-pages/Shadows.md
```
lyzhan7, line 10: Just fyi the screenshots I have for Android in https://github.com/microsoft/apple-ux-guide/blob/gh-pages/Shadows.md look pretty different from the screenshots for iOS, so referencing that doc might be confusing

Saadnajmi:
I'm actually going to close this for a few reasons:

1) My original intent was to re-use `Shadow` as the extra view needed to clip an Android Ripple's corner radius. I spoke with @rurikoaraki and it's probably best not to conflate Shadow and Ripple together. 
2) This change would change all the jest files to match the Android version of Shadow... which is much simpler and I feel a "worse" test. We should first update our jest config to be more platform specific, like here: https://github.com/microsoft/rnx-kit/tree/main/packages/jest-preset . I believe @ayush547 / @ankraj12 are working on that?
1) While mapping blur to elevation is the design answer and seems to work... I still think it's wrong? Feels like we should revisit with design.. I'll let Android people do that.

PR #2242, opened 2022-10-14
Use iOS global tokens from pipeline

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Main change here is the new global-tokens.ios.ts file. The addition of this file means that the iOS bundle will start grabbing global tokens from the design-tokens-ios package (ex. globalTokens.... will reference the iOS package. Tokens accessed through the theme are not impacted by this). Before this change, by default iOS was grabbing these global tokens from the design-tokens-windows package.

This caused two breaking issues that needed to be fixed along with this change:
1. on iOS we don't have globalTokens.spacing.sNudge and we don't plan to add this value. This value will be undefined everywhere, which is okay (tester will not error) except in ButtonTokens.ts, where we're attempting to do an arithmetic operation with this token value
2. Some win32 test suites were actually running on iOS, which was working because iOS was grabbing global tokens from windows (and win32/windows global token schema are more similar). After adding the iOS file, some tests that were testing win32/windows specific tokens started failing, but this was expected because iOS has different token definitions. Ended up updating the platform for these test suites to run on win32, which we actually don't do anywhere else yet. The updated @rnx-kit/jest-preset package was needed to fix some errors running a jest test suite on win32.

Notes:
- just want to confirm that both win32-theme.test.ts and default-theme.test.ts should be running on win32?
- There may be some visual changes on iOS, since globalTokens values have potentially changed or are now undefined. I think checking for issues here and fixing them should be done per component later.
- The snapshots for the tests have changed - the snapshots used to contain windows token values because the tests ran on iOS, which were using windows tokens. They should be using win32 tokens now.

### Verification

Made sure that the global-tokens.ios.ts was getting used (checked with console.log statements), opened every tester page to make sure none of them errored out. If there is anything specific I can test please let me know.

https://user-images.githubusercontent.com/78454019/197848590-6b6ebbe4-34ca-4b7f-84c0-286626dc7144.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/ButtonTokens.ts
```diff
@@ -9,7 +10,10 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
       width: '100%',
     },
     medium: {
-      padding: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
+      padding: Platform.select({
+        ios: 6,
+        default: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
+      }),
```
Saadnajmi, line 16: Can we move this to a ButtonTokens.ios.ts file?

Saadnajmi: Or is this just here for testing and I was too quick 😅

lyzhan7: This is a temporary patch needed just to get the tester stood up without errors, I copied what Android is doing from https://github.com/microsoft/fluentui-react-native/pull/2257. I'm not sure that we would still need a .ios.ts or .android.ts file once a better fix is added (maybe we're removing sNudge altogether?), so a little hesitant to add it now if we're just going to delete it later

Saadnajmi: When we were deciding how to put platform specific control tokes, we decided to use platform specific files for one main reasons: The overall bundle is smaller, since the other platform files are filtered out. This isn't true if you use platform select. 

There might have been other reasons, but I can't remember them. So as a rule, I try to have control tokens actually be separated into different platform files.

scripts/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "devDependencies": {
     "@react-native-community/cli": "^7.0.3",
-    "@rnx-kit/jest-preset": "^0.1.9",
+    "@rnx-kit/jest-preset": "^0.1.12",
```
lyzhan7, line 26: Wasn't sure if this was the right way to update the package? Usually I see dependabot update just the yarn.lock file. Should I wait for the dependabot PR to update merge it?

Saadnajmi: I think this is right. This is setting the minimum, which we need because that's the minimum version that has the fix to support win32 specific tests.

packages/components/Button/src/ButtonTokens.ts
```diff
@@ -12,6 +13,7 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
     medium: {
       padding: Platform.select({
         android: globalTokens.spacing.xs,
+        ios: globalTokens.spacing.xs,
         default: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
       }),
```
Saadnajmi, line 18: Revisiting: Can we move this to platform specific files? Or if those don't exist and that causes issues, at least leave a comment why w need this at the moment?

packages/theming/win32-theme/package.json
```diff
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": ">=0.68.0 <0.68.6",
```
Saadnajmi, line 46: Could we make the server match the other ones we have for react-native-win32? this format looks different to me

Saadnajmi: Also can we add it to peer dependencies below.. and maybe add the "core-win32" capability to "rex-kit" if it exists?

packages/components/Button/src/ButtonTokens.ts
lyzhan7: Yes I was planning to fix this, this was just when I merged in their changes

packages/theming/win32-theme/package.json
lyzhan7: ? Where are you looking at where this is different

Saadnajmi: Talked offline, I was wrong, nvm.

Saadnajmi (approved):

packages/components/Button/src/ToggleButton/__snapshots__/ToggleButton.test.tsx.snap
```diff
@@ -37,13 +37,13 @@ exports[`ToggleButton default 1`] = `
       "backgroundColor": "#f3f2f1",
       "borderColor": "#8a8886",
       "borderRadius": 4,
-      "borderWidth": 1,
+      "borderWidth": 1.5,
       "display": "flex",
       "flexDirection": "row",
       "justifyContent": "center",
```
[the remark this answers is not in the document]
Saadnajmi: I don't think the tests were ever actually running on win32? I think it's an open question still which platform each control should run it's test on.. ideally each component would just have tests for all platforms, which I think is possible but would slow us down 5x 😅 

Probably worth a follow-up. I think the answer changes for each component.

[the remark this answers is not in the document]
lyzhan7: This is the first time a jest test is running on win32. Most jest tests have been running on Android (with the exception of some tests recently getting switched over to iOS). It might look like the tests were running on win32/windows because up until recently iOS was using windows global tokens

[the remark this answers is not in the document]
ankraj12: yup, I do think that the tests were still having more of a win32/windows coverage because of the tokens being same in both android and iOS till recently. Maybe actually having tests running for each endpoint that validates snapshots and endpoint specific unit tests would be a good coverage, but if its possible and slows down inner loop then it's a weighted choice. Taking it more in offline chat.

PR #2245, opened 2022-10-17
Menu (macOS) Fix redbox related to onHoverOut

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We use MenuV1 on macOS inside FluentTester, but don't call it production ready yet. We were getting this Redbox every time someone moved a mouse out of a MenuV1 component:

Screen Shot 2022-10-17 at 9 17 39 AM

Upon investigation, it's because in `usePressableState`, we call `onHoverOut?.(e)`, but MenuV1 was passing in `onHoverOut: isCloseOnHoverOutEnabled && _onHoverOut` as the prop. This was evaluated as a bool instead of a function, since `isCloseOnHoverOutEnabled` is false on macOS. 

I see no reason to keep it false on macOS, so let's just remove that boolean altogether and simplify our logic. 

### Verification

Mouse hovering in Fluent Tester on macOS no longer causes Redboxes. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -71,7 +70,7 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
     },
     innerView: {
       onMouseEnter,
-      onMouseLeave: isCloseOnHoverOutEnabled && onMouseLeave,
+      onMouseLeave: onMouseLeave,
```
rurikoaraki, line 73: nit: could just be onMouseLeave

rurikoaraki (approved):

lyzhan7 (approved):

PR #2246, opened 2022-10-17
MenuV1 (macOS): Add FocusZone

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change adds a FocusZone on MenuV1 for macOS only, which allows better keyboarding. support. Keeping this change simple so that I don't inadvertently try to pair a lot of bug fixes in here as well. 

### Verification

Keyboarding up and down the MenuList is working...

https://user-images.githubusercontent.com/6722175/196241329-886987be-9294-40b5-a473-47b4237c0d22.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +34,39 @@ export const MenuList = compose<MenuListType>({
   ...stylingSettings,
   slots: {
     root: MenuStack,
+    focusZone: FocusZone,
```
chiuam, line 37: Do we need a platform check here?

Saadnajmi: I haven't done a Platform check in this particular place in the past, so I elected to not. I could try doing so and see if it makes any difference.

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +34,39 @@ export const MenuList = compose<MenuListType>({
...
+      focusZoneRef?.current?.focus();
+    }, []);
+
     return (_final: MenuListProps, children: React.ReactNode) => {
-      return (
-        <MenuListProvider value={contextValue}>
+      const content =
+        Platform.OS === 'macos' ? (
+          <Slots.root>
+            <FocusZone
```
rurikoaraki, line 54: Should this be Slots.focusZone?

Also nit: I wonder if you could define a wrapper component that's FocusZone on one platform and React.fragment on a different one (or swap the slot per platform?). Then the render code would look the same on all plats:

```
<Slots.root>
    <WrapperComponent>
        {children}
    </WrapperComponent>
</Slots.root>
```

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -33,18 +34,39 @@ export const MenuList = compose<MenuListType>({
...
+              defaultTabbableElement={focusZoneRef}
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore FocusZone takes ViewProps, but that isn't defined on it's type.
+              enableFocusRing={false}
+            >
+              {children}
+            </FocusZone>
+          </Slots.root>
+        ) : (
           <Slots.root>{children}</Slots.root>
```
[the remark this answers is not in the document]
Saadnajmi: What do you mean? Should what be inside what?

packages/components/Menu/src/MenuList/MenuList.tsx
Saadnajmi: Right, good catch. 
A wrapper component would probably work. iirc, we've done this elsewhere. I'll look into it

packages/components/Menu/src/MenuList/MenuList.tsx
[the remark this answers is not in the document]
chiuam: sorry it was a typo. I meant to say "Should this be inside <MenuListProvider>"

packages/components/Menu/src/MenuList/MenuList.tsx
Saadnajmi: Actually, I could make a follow-up change to let componentRef also be of type 'React.ForwardedRef' in all our components I suppose.

rurikoaraki: The wrapper thing was more of a suggestion, it seems like it's not worth the work, which is fine.

Saadnajmi: Filed https://github.com/microsoft/fluentui-react-native/issues/2283 about the componentRef forwarded ref thing. For this PR, I think I'll keep it as is. I also worry that not using `stagedComponent` for FocusZoneWrapper might be bad.. even though I'm pretty sure I can't since I need the ref.

packages/components/Menu/src/MenuList/MenuList.tsx
Saadnajmi: I think my take on this is I'd rather not have slots only be defined on one platform. I'd rather have slots be defined on all platforms (and have their slotProps calculated) for the sake of consistency. I gave similar feedback in the PR #2103 about the new android only slot. So I'll leave it as is.

packages/components/Menu/src/MenuList/MenuList.tsx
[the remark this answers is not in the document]
Saadnajmi: @chiuam good now? `<MenuListProvider>` is still there on line 70

chiuam (approved):

PR #2250, opened 2022-10-19
[Shadow] Update spec to explain why the shadow component was created

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update shadow spec to explain why the component was created

### Verification

n/a

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Shadow/SPEC.md
+++ b/packages/experimental/Shadow/SPEC.md
@@ -6,6 +6,8 @@ A `Shadow` is a component that can be used to add Fluent shadows to other compon
 
 Shadows defined in the Fluent design system are actually comprised of two shadows. The **key shadow** is used to create a feeling of distance, and the **ambient shadow** is used to define the edges of the shape.
 
+The `Shadow` component is different from other components in that it was created to address limitations with React Native - Views in React Native can only have one Shadow set at a time. This component provides an easy way to add a Fluent shadow to another component.
+
 ## Shadow Depth Ramp
 
 These are the current `Shadow` variants:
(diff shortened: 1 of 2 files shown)


packages/experimental/Shadow/SPEC.md
```diff
@@ -21,13 +23,17 @@ Examples adding some Shadows to some text:
 
 ```tsx
 <Shadow shadowToken={theme.shadows.shadow8}>
-  <Text>Text box with shadow8</Text>
+  <View>
+    <Text>Text box with shadow8</Text>
+  </View>
```
Saadnajmi, line 28: Nit: A `<Button>` might be simpler to convey the point :)

packages/experimental/Shadow/SPEC.md
```diff
@@ -47,3 +53,4 @@ If `undefined` is passed into the `shadowToken` prop, no shadow will be rendered
 ## Notes
```
Saadnajmi, line 53: Could you add a `## Details / Implementation` (not sure which name I prefer  section where we briefly explain how Shadow adds an extra View around your view for the ambient shadow, and smartly extracts your style props so that the extra view and the child are exactly the same size? Maybe we can also add a bullet point about how in our investigation, Apple platforms needed their `blur radius` divided by 2 in comparison with web.

rurikoaraki (approved):

PR #2251, opened 2022-10-19
Add an example of how spacing tokens can be used in FURN

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This adds an example of how spacing tokens can be used in FURN. The tokens were chosen such that their iOS values line up with the original hard-coded values.

### Verification

Notification looks the same on the test page.

| Before | After |
|-|-|
| Screen Shot 2022-10-19 at 2 37 16 PM | Screen Shot 2022-10-19 at 2 37 43 PM |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Notification/src/NotificationTokens.ios.ts
+++ b/packages/components/Notification/src/NotificationTokens.ios.ts
@@ -104,8 +104,8 @@ export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme>
     fontSize: 15,
     fontWeight: '400',
     minHeight: 52,
-    padding: 16,
-    paddingVertical: 12,
+    padding: t.spacing.m,
+    paddingVertical: t.spacing.s,
     shadowToken: notificationShadowStyle,
     isBar: {
       borderRadius: 0,
--- a/packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
+++ b/packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
@@ -78,7 +78,7 @@ exports[`Notification component tests Notification default 1`] = `
         "flexWrap": undefined,
         "justifyContent": "space-between",
         "minHeight": 52,
-        "padding": 16,
+        "padding": "16px",
         "paddingBottom": undefined,
         "paddingEnd": undefined,
         "paddingHorizontal": undefined,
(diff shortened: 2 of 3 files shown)


harrieshin:
can you add the screenshots?

packages/components/Notification/src/NotificationTokens.ios.ts
```diff
@@ -104,8 +104,8 @@ export const defaultNotificationTokens: TokenSettings<NotificationTokens, Theme>
     fontSize: 15,
     fontWeight: '400',
     minHeight: 52,
-    padding: 16,
-    paddingVertical: 12,
+    padding: t.spacing.m,
```
harrieshin, line 107: i must have missed the train on this one. but ".m" or "s" is it too abbreviate it? is it meant to be medium or small?

amgleitman: This abbreviation is by design.

amgleitman:
> can you add the screenshots?

Added. They're not all that exciting.

Saadnajmi:
I think you can fix the CI error by running `yarn test -u` and then committing the updated snapshot

rurikoaraki (approved):

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
```diff
@@ -78,15 +78,15 @@ exports[`Notification component tests Notification default 1`] = `
...
+        "padding": "16px",
         "paddingBottom": undefined,
         "paddingEnd": undefined,
         "paddingHorizontal": undefined,
         "paddingLeft": undefined,
         "paddingRight": undefined,
         "paddingStart": undefined,
         "paddingTop": undefined,
-        "paddingVertical": 12,
+        "paddingVertical": undefined,
```
rurikoaraki, line 89: Not blocking, just a comment that this is unfortunate. Maybe this is a case where we should force jest to run on ios.

rurikoaraki:
I think that having an example of pulling spacing from the Theme object is good to have, but I think that my overall preference would still be to lean towards pulling global tokens as a whole from the tokens package directly instead of from the Theme object. My thought is that spacing was updated on the Theming object because it was already there from the old Theme, but I would minimize usage of pulling global tokens from the Theme for the sake of consistency and reducing confusion.

packages/components/Notification/src/__tests__/__snapshots__/Notification.test.tsx.snap
amgleitman: This is due to the fact that our jest tests use the default theme instead of the iOS theme, so this is a separate issue.

PR #2254, opened 2022-10-20
[Menu] Add ability to pass own callbacks to child of MenuTrigger

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The child of the MenuTrigger had several callbacks overwritten but did not expose a way to have an additional callback provided by the client. Any callbacks passed into the child of the MenuTrigger would not get called. This change fixes that by calling any callbacks that are passed in for onClick and onHoverFoo, which are directly overwritten by useMenuTrigger. Also fixes a bug where the functions were recursively calling themselves forever.

### Verification

Test cases were added to test that any passed in callbacks were properly called.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
+++ b/packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
@@ -1,0 +1,48 @@
+import { AccessibilityActionEvent } from 'react-native';
+import { memoize } from '@fluentui-react-native/framework';
+import { InteractionEvent, isMouseEvent } from '@fluentui-react-native/interactive-hooks';
+import { MenuTriggerChildProps, MenuTriggerState } from './MenuTrigger.types';
+
+export const getRevisedProps = memoize(getRevisedPropsWorker);
+function getRevisedPropsWorker(state: MenuTriggerState, props: any): MenuTriggerChildProps {
+  const revisedProps = { ...state.props };
+  if (props.accessibilityState) {
+    revisedProps.accessibilityState = { ...state.props.accessibilityState, ...props.accessibilityState };
+  }
+
+  if (props.accessibilityActions) {
+    revisedProps.accessibilityActions = [...state.props.accessibilityActions, ...props.accessibilityActions];
+  }
+
+  if (props.onAccessibilityAction) {
+    revisedProps.onAccessibilityAction = (e: AccessibilityActionEvent) => {
+      props.onAccessibilityAction(e);
+      state.props.onAccessibilityAction(e);
+    };
+  }
+
+  if (props.onClick) {
+    revisedProps.onClick = (e: InteractionEvent) => {
+      props.onClick(e);
+      state.props.onClick(e);
+    };
+  }
+
+  let onHoverIn = undefined;
+  if (props.onHoverIn) {
+    onHoverIn = (e: InteractionEvent) => {
+      props.onHoverIn(e);
+      state.props.onHoverIn(isMouseEvent(e) && e);
+    };
+  }
+
+  let onHoverOut = undefined;
+  if (props.onHoverOut) {
+    onHoverOut = (e: InteractionEvent) => {
+      props.onHoverOut(e);
+      state.props.onHoverOut(isMouseEvent(e) && e);
+    };
+  }
+
+  return { ...revisedProps, onHoverIn, onHoverOut };
+}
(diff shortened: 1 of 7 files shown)


packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
```diff
@@ -0,0 +1,48 @@
+import { AccessibilityActionEvent } from 'react-native';
+import { memoize } from '@fluentui-react-native/framework';
+import { InteractionEvent, isMouseEvent } from '@fluentui-react-native/interactive-hooks';
+import { MenuTriggerChildProps, MenuTriggerState } from './MenuTrigger.types';
+
+export const getRevisedProps = memoize(getRevisedPropsWorker);
+function getRevisedPropsWorker(state: MenuTriggerState, props: any): MenuTriggerChildProps {
+  const revisedProps = { ...state.props };
```
rurikoaraki, line 8: This is done so that revisedProps.Foo and state.props.Foo are different references, so that calling state.props.Foo inside revisedProps.Foo doesn't re-call revisedProps.Foo. Before I had revisedProps.Foo re-call revisedProps.Foo, which resulted in an infinite recursion loop

packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  if (props.onClick) {
+    revisedProps.onClick = (e: InteractionEvent) => {
+      props.onClick(e);
+      state.props.onClick(e);
+    };
+  }
+
+  let onHoverIn = undefined;
```
rurikoaraki, line 31: This is done because onHoverIn and onHoverOut are Readonly properties, so I have to explicitly add them through creating a new object below instead of assigning them to revisedProps directly.

Saadnajmi (approved):

packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
```diff
@@ -0,0 +1,48 @@
+import { AccessibilityActionEvent } from 'react-native';
+import { memoize } from '@fluentui-react-native/framework';
+import { InteractionEvent, isMouseEvent } from '@fluentui-react-native/interactive-hooks';
+import { MenuTriggerChildProps, MenuTriggerState } from './MenuTrigger.types';
+
+export const getRevisedProps = memoize(getRevisedPropsWorker);
```
Saadnajmi, line 6: Small comment what revised props is doing?

packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
```diff
@@ -0,0 +1,48 @@
...
+    revisedProps.onAccessibilityAction = (e: AccessibilityActionEvent) => {
+      props.onAccessibilityAction(e);
+      state.props.onAccessibilityAction(e);
+    };
+  }
+
+  if (props.onClick) {
+    revisedProps.onClick = (e: InteractionEvent) => {
+      props.onClick(e);
+      state.props.onClick(e);
```
Saadnajmi, line 27: Not sure it does, but does order matter? I've normally seen our "injected logic' fire first before the client passed callback. I think this is the opposite?

Saadnajmi (approved):

packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
rurikoaraki: I don't think it really matters, I picked this order because that's what web has. I guess if there's a way for the client to tell us to not run our version of the event by telling us to not propagate the event or something, that would be a reason to run the client version first

rurikoaraki: But I can switch it for consistency for now

packages/components/Menu/src/MenuTrigger/getRevisedProps.ts
Saadnajmi: Post merge question: What's making them readonly? Should we fix a type somewhere to make it not? Was that because of my last merge with all the Pressable stuff?

Saadnajmi: Not anymore because MenuV1 doesn't use our fork anymore?

Saadnajmi: I think we (APEX) will be looking at MenuV1 again relatively soon, so I'll take a second look myself when we do.

rurikoaraki: It's because the MenuTriggerChildProps extends from IWithPressableOptions which uses the type I linked above. If it should derive from something else, then yeah that should be fixed.

Saadnajmi: I guess I missed that. That should be `PressablePropsExtended` No worries, I'll take care of that when I touch these myself.

PR #2257, opened 2022-10-20
Integrate typography tokens for android

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
Consumed 

### Verification

Integrated font variant values from pipeline. 
Stubbed out button token that causes crashes due to token unavailable in android. The right token set will be picked up as part of control tokenization.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/ButtonTokens.ts
```diff
@@ -2,14 +2,18 @@ import { Theme } from '@fluentui-react-native/framework';
...
   ({
     block: {
       width: '100%',
     },
     medium: {
-      padding: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
+       padding: Platform.select({
+        android: 6,
+        default: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
+       }),
```
Saadnajmi, line 16: Same comment in the iOS review: can we move this to a platform specific file?

packages/components/Button/src/ButtonTokens.ts
```diff
@@ -2,14 +2,18 @@ import { Theme } from '@fluentui-react-native/framework';
...
 
 export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
   ({
     block: {
       width: '100%',
     },
     medium: {
-      padding: globalTokens.spacing.sNudge - globalTokens.stroke.width.thin,
+       padding: Platform.select({
+        android: 6,
```
rurikoaraki, line 14: nit: ideally spacing is defined by spacing tokens and not through hardcoded numbers

rurikoaraki (approved):

packages/components/Button/src/ButtonTokens.ts
ankraj12: This is just a unblocking fix for the tester to load, and we have a issue to track the style updates so that the iOS/ Android styling comes from a different file. At this point, I have not yet fully tracked the set of token changes needed for Button, so leaving this to fix up in the next changes. Reviewed your inputs on Lynn's draft, agree that this is not a pattern for platform styling we would like to be in place.

packages/components/Button/src/ButtonTokens.ts
ankraj12: I updated this to pick from another token instead of hardcoded value - the actual token work is still an open item.

PR #2260, opened 2022-10-20
Implement Pressable states for Text

lenahong:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes
Support pressable states for Text as compressible

### Verification

TODO: Add tests to reflect state changes.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Link/src/useLink.ts
```diff
@@ -44,7 +44,7 @@ export const useLink = (props: LinkProps): LinkInfo => {
   // GH #1336: Set focusRef to null if link is disabled to prevent getting keyboard focus.
   const focusRef = isDisabled && !focusable ? null : componentRef;
   const onPressWithFocus = useOnPressWithFocus(focusRef, linkOnPress);
-  const pressable = useAsPressable({ ...rest, disabled: isDisabled, onPress: onPressWithFocus });
+  const pressable = usePressableState({ ...rest, disabled: isDisabled, onPress: onPressWithFocus });
```
Saadnajmi, line 47: I saw elsewhere in Text.tsx that you added a call to `useAsPressable`.I don't think the two should be mixed. Is the intention of this PR to use `useAsPressable` (our forked implementation) or `usePressableState` (reuses the implementation in React Native) ?  The answer informs future PR comments :)

lenahong: I believe I'm intending to use our forked implementation of useAsPressable. I'll change Link to use this as well.

packages/components/text/package.json
```diff
@@ -43,7 +43,8 @@
   },
   "peerDependencies": {
     "react": "17.0.2",
-    "react-native": "^0.68.0"
+    "react-native": "^0.68.0",
+    "@fluentui-react-native/interactive-hooks": ">=0.19.0 <1.0.0"
```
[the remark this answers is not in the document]
rurikoaraki: I don't think this is the proper way to handle this, because this isn't what peer dependencies are meant for. Peer dependencies are for us to declare compatibility with a different library, and generally they don't get automatically installed, so it's not a good way to declare an actual dependency. This is why we have RN as both a dev and a peer dependency.

I think the proper solution would be to move createIconProps into a different package so that interactive-hooks doesn't depend on the icon package. And I think that change makes sense anyway because createIconProps isn't a hook or something that's for interactivity.

packages/components/text/src/Text.types.ts
```diff
@@ -18,80 +25,79 @@ export type TextWeight = 'regular' | 'medium' | 'semibold' | 'bold';
 /**
  * Text props, based off of the standard react-native TextProps with some new extensions
  */
-export type TextProps<TBase = ITextProps> = TBase &
-  FontVariantTokens & {
-    /**
-     * foreground text color
-     */
-    color?: ColorValue;
+export interface TextProps extends IWithPressableOptions<ITextProps & FontVariantTokens> {
```
rurikoaraki, line 28: Could you make sure the SPEC is updated to have the correct values?

packages/components/text/package.json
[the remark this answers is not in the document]
lenahong: I moved it to the icon package for now since it's a hook that will always be used when IconProps are defined or an Icon slot is created.

packages/libraries/core/src/index.ts
```diff
@@ -200,7 +200,6 @@ export type {
 } from '@fluentui-react-native/text';
 
 export {
-  createIconProps,
```
Saadnajmi, line 203: Can we also do a quick check downstream to make sure nobody is referring to this, in case this inadvertently breaks someone?

lenahong: That's a good point. Where are all the places I have to check to see?

Saadnajmi: Responded offline!

lenahong: It looks like it was only used internally in FURN by other components in this repo

packages/components/text/src/Text.types.ts
lenahong: Done!

apps/fluent-tester/src/TestComponents/TextExperimental/PressableUsage.tsx
```diff
@@ -33,13 +33,7 @@ export const PressableUsage: React.FunctionComponent = () => {
...
         <Text variant={'bodyStandard'}>
           To learn more about microsoft, visit this{' '}
-          <Text
-            variant={'bodyStandard'}
-            color="blue"
-            keyDownEvents={handledNativeKeyboardEvents}
-            onPress={_onPress}
-            onKeyDown={_onKeyDown}
-          >
+          <Text variant={'bodyStandard'} color="blue" keyDownEvents={handledNativeKeyboardEvents} onPress={_onPress} onKeyDown={_onKeyDown}>
```
Saadnajmi, line 36: Nit: use `useKeyProps` instead of passing in `onKeyDown / keyDownEvents` manually hook so that onKeyDown would (eventually) also work on macOS

packages/components/text/src/Text.tsx
```diff
@@ -107,6 +104,7 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     ...pressable?.props,
     numberOfLines: truncate || !wrap ? 1 : 0,
     onKeyDown: Platform.OS === (('win32' as any) || 'windows') ? onKeyDown : undefined,
+    onMouseEnter: Platform.OS === (('win32' as any) || 'windows') ? pressable.props.onMouseEnter : undefined,
```
Saadnajmi, line 107: Could we combine these into an object we spread and then have a comment like "These callbacks are not implemented on iOS/macOS, and cause Redboxes if passed in. Limit to only windows/win32 for now".

Also not sure if they cause issues on Android but I think it's fine to filter to only windows.

Saadnajmi: Also, `onMouseLeave`

packages/experimental/Link/src/useLink.ts
```diff
@@ -65,6 +66,8 @@ export const useLink = (props: LinkProps): LinkInfo => {
       ...rest,
       ...onKeyUpProps,
       ...pressable.props, // allow user key events to override those set by us
+      onKeyDown: Platform.OS === (('win32' as any) || 'windows') ? onKeyDown : undefined,
+      onMouseEnter: Platform.OS === (('win32' as any) || 'windows') ? pressable.props.onMouseEnter : undefined,
```
Saadnajmi, line 70: Similar comment here

Saadnajmi: Also `onKeyUp` here, since the spread of `onKeyUpProps` on line 67 is probably passing that in.

packages/components/text/src/Text.tsx
```diff
@@ -52,6 +53,16 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
+  const textOnPress = React.useCallback(
+    (e) => {
+      if (onPress) {
+        onPress(e);
+      }
+      e.stopPropagation();
+    },
+    [onPress],
+  );
+  const onKeyUpProps = useKeyProps(textOnPress, ' ', 'Enter');
```
Saadnajmi, line 65: nit: This hook returns `keyDown` for macOS, and `keyUp` for windows, since I found that to be a platform difference which was the default. So a better name would exclude the "up" to avoid confusion

Saadnajmi:
I'm noticing some strange `onPress` behavior. It seems when I click the inner text of a <Button>, the button's onClick isn't fired but instead, the inner Text's onClick, and I get this weird press effect. Seems to also only be on this PR. Any idea what's going on?

https://user-images.githubusercontent.com/6722175/198715315-b980060c-11aa-43b5-b7f5-d744fcd37446.mov

lenahong:
Making Text useAsPressable by default is causing issues for other components that consume Text. I'm adding a boolean check on Text isPressable to prevent this behavior

Saadnajmi:
> Making Text useAsPressable by default is causing issues for other components that consume Text. I'm adding a boolean check on Text isPressable to prevent this behavior

I wonder if just wrapping Text with `usePressableState` would be better, since doing so won't "inject" pressability into Text, but re-use the logic that's already there?

PR #2267, opened 2022-10-24
FocusZone (macOS): New prop to ignore OS keyboard shortcut preference

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change allows FocusZone to make Menu always focusable on macOS, regardless of the user's system preference. This matches what native macOS Context menus do.

### Verification

Tested turning the system setting on and off, and verified that you could still keyboard through a menu.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -18,53 +18,26 @@
 // enumerated in the same row (or column) as the current focused view
```
Saadnajmi, line 18: This file looks like it has a lot of changes because I moved a lot of static methods outside of `@implementation RCTFocusZone`. Didn't see a need to keep them inside the class.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -18,53 +18,26 @@
...
+static NSView *GetFirstKeyViewWithin(NSView *parentView, BOOL forceFocus)
 {
-    if ([[parentView subviews] count] < 1)
-    {
-        return nil;
-    }
+		if ([[parentView subviews] count] < 1)
+		{
+				return nil;
+		}
```
lyzhan7, line 33: These whitespace changes look a little odd in Github, just wanted to check that they're actually fine on your side

packages/components/FocusZone/src/FocusZone.types.ts
```diff
@@ -18,32 +18,55 @@ export interface FocusZoneProps {
...
 
   /**
-   ** Defines which arrows to react to
+   ** Defines which arrow keys to react to
    */
   focusZoneDirection?: FocusZoneDirection;
 
   /**
-   ** If set, the FocusZone will not be tabbable and keyboard navigation will be disabled
+   * Disabled the FocusZone, with slightly different behavior on macOS and win32.
```
lyzhan7, line 26: Disables?

packages/components/FocusZone/src/FocusZone.types.ts
```diff
@@ -18,32 +18,55 @@ export interface FocusZoneProps {
...
+   * On win32, the FocusZone will not be tabbable and keyboard navigation will be disabled.
+   * on macOS, the FocusZone will "pass through" events, and the children will respond as if there was no FocusZone.
    */
   disabled?: boolean;
 
-  /* Circular Navigation prop */
+  /**
+   * Circular Navigation prop
+   * @platform win32
+   */
```
lyzhan7, line 35: Following the naming for forceFocusMacOS, if this prop is win32 only should we name it isCircularNavigationWin32? Same thing below

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Github defaults to 8 spaces for a tab, instead of 4. I think there's a file we can check in to the repo to fix this. We should also just have... clang format in our repo because I literally get this comment every time I edit native code...

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -18,53 +18,26 @@
...
-	CGFloat rect2Offset = isRTL ? rect2.size.width : 0;
-	
-	NSPoint rect1Corner = NSMakePoint(rect1.origin.x + rect1Offset , rect1.origin.y);
-	NSPoint rect2Corner = NSMakePoint(rect2.origin.x + rect2Offset , rect2.origin.y);
-
-	return GetDistanceBetweenPoints(rect1Corner, rect2Corner);
-}
-
-static inline CGFloat GetMinDistanceBetweenRectVerticesAndPoint(NSRect rect, NSPoint point)
+BOOL ShouldFocusOnView(NSView *view, BOOL forceFocus)
```
chiuam, line 21: Can we access `forceFocus` with [self forceFocus] instead? This is to avoid adding unnecessary function parameters.

packages/components/FocusZone/src/FocusZone.types.ts
Saadnajmi: That would be a change to the public API of the control, which we don't want to do. 

Additions are fine, changes/removals to existing props are not.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -18,53 +18,26 @@
...
 {
-	return fmin(
-		fmin(GetDistanceBetweenPoints(point, NSMakePoint(NSMinX(rect), NSMinY(rect))),
-			GetDistanceBetweenPoints(point, NSMakePoint(NSMaxX(rect), NSMaxY(rect)))),
-		fmin(GetDistanceBetweenPoints(point, NSMakePoint(NSMaxX(rect), NSMinY(rect))),
-			GetDistanceBetweenPoints(point, NSMakePoint(NSMinX(rect), NSMaxY(rect))))
-	);
+	return forceFocus ? [view acceptsFirstResponder] : [view canBecomeKeyView];
```
chiuam, line 23: Worth adding documentation explaining this logic? I thought it was quite confusing at first :)

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Doing so would make this no longer a static function, and would lead to a bunch of code that looks like this:
`[self firstFocusableViewWithin:self]` since I'd have to make other previously static C methods into instance methods. 

After talking with @toddha , I went with keeping this static to match the rest of this file.

packages/components/FocusZone/src/FocusZone.types.ts
Saadnajmi: Fixed!

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Sure. I think I'll also rename the argument to "forceFocusonAllResponder" or "onlyFocusOnKeyViews". Since this is an argument inside the internal implementation, I'm more OK with the exact name.

amgleitman (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Added docs here and a few other methods with my last commit. Let me know if there are things that could be worded better :)

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Looking locally, it looks fine to me. RCTFocusZone uses Tabs everywhere. I think React Native Core / macOS actually use spaces, so maybe when we add a clang format, we could copy that over and bulk edit all our native code.

PR #2269, opened 2022-10-24
Use FRNFontMetrics module in `Text`

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Since it may be a while before proper Dynamic Type support makes it into RN Core ([PR here](https://github.com/facebook/react-native/pull/35017)), we're adding it as a temporary native module. Basically it's a reimplementation of [the work we did in react-native-macos](https://github.com/microsoft/react-native-macos/pull/1445).

This works by keeping track of the current Dynamic Type scale factors in a single location, with `RCTEventEmitter` events being used to send any new updates to the JS side. A hook tells each of the FURN `Text` elements to rerender.

One issue with this is that this can cause two rerenders when we change the preferred font size: once to rerender any native text elements, and again to rerender FURN text elements. From my experiments, it takes about 89 ms to complete both rerenders. Unfortunately, this is unavoidable, since most of this time is spent waiting for the event propagate across the bridge (about 60-70 ms).

Thankfully, this shouldn't be something we have to worry about too much. There are two main ways to change the preferred text size, and they each have a way to "hide" this second rerender.
* If the user changes the text size through the Accessibility menu in the Settings app, we should already be in the clear, since there's no way we can view the target app and the Settings app at the same time. (Even on iPad, you can't have the Settings app as part of a split view or a "floating" window.)
* If the user changes the text size through the Control Center, the rest of the screen is blurred, so the effects of any changes aren't as noticeable. This is especially true on iPhone, where the control center takes up most of the entire screen. (On iPad, the Control Center mainly lives in the corner, but the blurred background still applies.)

### Verification

| Default size                                       | Largest accessibility size (AX5)                                      |
| Screen Shot 2022-10-24 at 4 07 08 PM | Screen Shot 2022-10-24 at 4 08 02 PM |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


amgleitman:
This currently uses a synchronous approach to fetching the appropriate `UIFontMetrics` information, so it's not ideal in its current state, but I figure that putting a proposed change out there is still useful enough to warrant a draft PR.

Saadnajmi (commented):
Update the nuspec too please

yarn.lock
```diff
@@ -1495,6 +1495,9 @@
   resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-windows/-/design-tokens-windows-0.14.0.tgz#1483caa1dd4e7097ac48a5872e93d7f896e8402c"
   integrity sha512-WNNVTMpRir6a2Voa0yu3BZJmy2w7fofrokL+nyFQM40bACtX68KDRDE96HA5W5+4q8rEaNSXV3DzSzLXm6ekfA==
 
+"@fluentui-react-native/experimental-native-font-metrics@file:packages/experimental/NativeFontMetrics":
+  version "0.1.0"
+
```
Saadnajmi, line 1500: I think you actually want 0.2.0

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.types.ts
```diff
@@ -1,3 +1,9 @@
+export interface IFontMetrics {
```
Saadnajmi, line 1: We try to stop using the "I" prefix in FURN now.

packages/experimental/NativeFontMetrics/src/fontMetrics.ts
```diff
@@ -0,0 +1,23 @@
...
+      this._scaleFactors = newScaleFactors;
+    });
+  }
+
+  get scaleFactors(): ScaleFactors {
+    return this._scaleFactors;
+  }
+}
+
+const fontMetricsImpl = Platform.OS === 'ios' ? new FontMetricsImpl() : { scaleFactors: {} };
```
Saadnajmi, line 22: Any chance we could separate this into a separate `.ios` file? I think that will result in a smaller bundle size and make it more clear it's iOS only.

packages/experimental/NativeFontMetrics/src/useFontMetrics.ts
```diff
@@ -1,20 +1,21 @@
...
 import { ScaleFactors } from './NativeFontMetrics.types';
 
 const eventEmitter = new NativeEventEmitter(NativeFontMetrics as any);
 
-export function useFontMetrics(): ScaleFactors {
+export function useFontMetricsScaleFactors(): ScaleFactors {
   if (Platform.OS !== 'ios') {
     console.warn('NativeFontMetrics is only available on iOS');
     return {};
   }
```
Saadnajmi, line 14: I've seen that these warnings to show up a lot on boot for all platforms. Can we verify this doesn't show up on say, macOS? Or better yet, refactor to make sure this is only used on iOS?

packages/components/text/src/Text.tsx
```diff
@@ -49,6 +50,8 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
   // get the tokens from the theme
   let [tokens, cache] = useTokens(theme);
 
+  const fontMetricsScaleFactors = useFontMetricsScaleFactors();
+
```
Saadnajmi, line 54: This calls it for every platform, meaning every platform will get the warning from the other file, right?

packages/components/text/src/Text.tsx
```diff
@@ -119,14 +125,30 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
     const mergedProps = {
       ...rest,
       ...keyProps,
       ...filteredProps,
       ...extra,
       onPress,
       numberOfLines: truncate || !wrap ? 1 : 0,
-      style: mergeStyles(tokenStyle, props.style, extra?.style),
+      style: mergeStyles(tokenStyle, props.style, extra?.style, scaleStyleAdjustmentProps),
+      allowFontScaling: dynamicTypeVariant === undefined, // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
```
Saadnajmi, line 151: You could also do:

`....(Platform.OS === 'ios') && {allowFontScaling: true}`

packages/components/text/src/Text.tsx
```diff
@@ -119,14 +125,30 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
 
   // return a continuation function that allows this text to be compressed
   return (extra: TextProps, children: React.ReactNode) => {
+    // [TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
+    interface ScaleStyleAdjustmentProps {
+      fontSize?: number;
+      lineHeight?: number;
+    }
```
Saadnajmi, line 132: I'd move the type definition out of here into either the top or a types file if possible. Or better yet, re-use an existing type?

packages/components/text/src/Text.tsx
```diff
@@ -119,14 +125,30 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
+    let scaleStyleAdjustmentProps: ScaleStyleAdjustmentProps = {};
+    if (dynamicTypeVariant !== undefined && typeof tokenStyle.fontSize === 'number' && typeof tokenStyle.lineHeight === 'number') {
+      const scaleFactor = fontMetricsScaleFactors[dynamicTypeVariant] ?? 1;
+      scaleStyleAdjustmentProps = {
+        fontSize: tokenStyle.fontSize * scaleFactor,
+        lineHeight: tokenStyle.lineHeight * scaleFactor,
+      };
+    }
```
Saadnajmi, line 140: Is there any way we can simplify this if () calculation? or move it to a separate function / into the existing hook?

packages/components/text/src/Text.tsx
amgleitman: Honestly, probably better to have a `Platform.OS === 'ios'` guard here too.

packages/components/text/src/Text.tsx
amgleitman: Good idea. It also allows us to leave the test snapshot alone.

packages/components/text/src/Text.tsx
amgleitman: This was done as an initial thing because token typing isn't necessarily obvious upon first glance. After digging a bit deeper, it looks like `TextTokens` works just fine here.

packages/components/text/src/Text.tsx
amgleitman: The problem is that `tokenStyle.fontSize` can also be a `string` (e.g., `"14px"`). Ditto for `tokenStyle.lineHeight`. We could try to parse out the numerical value, but that would make things more complicated.

It works out because all of our font size tokens are pure numbers, but I don't know if this is an assumption we absolutely want to stick with. Maybe we should clarify this?

packages/experimental/NativeFontMetrics/src/useFontMetrics.ts
amgleitman: Our test code doesn't run macOS code, so as this stands it would be on the client to ensure that this is only used on iOS. That being said, it would also be nice to make sure we don't try to access `NativeFontMetrics` on any other platform.

For some strange reason, I find that if we try splitting this into its own useFontMetrics.ios.ts file, the auto-linking from `pod install` fails to pull FRNFontMetrics into the iOS test app. Apparently the macOS test app has auto-linking problems too, so maybe it's just a weird quirk.

packages/components/text/src/Text.tsx
```diff
@@ -49,6 +50,8 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
   // get the tokens from the theme
   let [tokens, cache] = useTokens(theme);
 
+  const fontMetricsScaleFactors = Platform.OS === 'ios' ? useFontMetricsScaleFactors() : undefined;
```
rurikoaraki, line 53: hooks must run on every render, so you need to call useFontMetricsScaleFactors unconditionally

Saadnajmi: Right, thanks for catching. A `Platform` check is effectively unconditional but eslint doesn't recognize it as such. How we've solved this elsewhere is to move `if` checks to inside the custom hook: https://github.com/microsoft/fluentui-react-native/blob/14465a397887dc9cbe29bedd35461a3356be2cbf/packages/components/Notification/src/Notification.tsx#L24

Also worth linking again: https://reactjs.org/docs/hooks-rules.html

packages/components/text/src/Text.tsx
Saadnajmi: Talked offline about this. I feel like this makes a good case that we should enforce a number as much as possible. If we can do that in our tokens, great. If not, let's at least add a comment on top of these checks?

packages/experimental/NativeFontMetrics/src/useFontMetrics.ts
Saadnajmi: I want to follow up on the "auto linking doesn't work" part. That feels strange to me because auto linking should only be dependent on ` pod spec` file being available at the root of your package. 

That being said, I did run this change on macOS and didn't see extra warnings pop up so I think so yay?

packages/components/text/src/Text.tsx
amgleitman: Resolved with latest changes.

packages/experimental/NativeFontMetrics/src/useFontMetrics.ts
amgleitman: Either way, I think we can get around this by just having this module always return a multiplier of 1 on non-iOS platforms.

packages/components/text/src/Text.tsx
amgleitman: Done.

packages/experimental/NativeFontMetrics/src/fontMetrics.ts
amgleitman: Done.

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.types.ts
amgleitman: Fixed.

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.types.ts
```diff
@@ -1,3 +1,9 @@
+export interface FontMetrics {
+  readonly scaleFactors: ScaleFactors;
+}
+
+export type ScaleFactors = { [K in TextStyle]?: number };
```
Saadnajmi, line 5: Don't we have to import `TextSyle` somewhere in this file?

packages/components/text/src/Variants.ios.ts
```diff
@@ -1,43 +1,57 @@
 import { Text } from './Text';
 
+// TODO(#2268): Remove "as any" designations once RN Core properly supports Dynamic Type scaling
+
 export const Caption1 = Text.customize({
   variant: 'caption1',
-});
+  dynamicTypeRamp: 'footnote',
```
Saadnajmi, line 7: We should probably have a way to reconcile this in the future (i.e: `variant` sets dynamicTypeRamp`). I don't know if that's possible in the current PR

packages/components/text/src/Text.types.ts
```diff
@@ -8,7 +8,11 @@ export const textName = 'Text';
  * Text tokens, these are the internally configurable values for Text elements. In particular these
  * drive decisions on how to build the styles
  */
-export type TextTokens = Omit<FontTokens, 'fontFamily'> & IForegroundColorTokens & Omit<TextStyle, 'fontSize' | 'fontWeight' | 'color'>;
+export type TextTokens = Omit<FontTokens, 'fontFamily'> &
+  IForegroundColorTokens &
+  Omit<TextStyle, 'fontSize' | 'fontWeight' | 'color'> & {
+    dynamicTypeRamp?: string; // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
```
Saadnajmi, line 14: Speaking of... we should update the types in DefinitelyTyped/React Native Core if we haven't already. Feel free to ping me about that.

packages/components/text/src/Text.tsx
```diff
@@ -126,7 +146,8 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
       ...extra,
       onPress,
       numberOfLines: truncate || !wrap ? 1 : 0,
-      style: mergeStyles(tokenStyle, props.style, extra?.style),
+      style: mergeStyles(tokenStyle, props.style, extra?.style, scaleStyleAdjustments),
+      ...(dynamicTypeVariant !== undefined && { allowFontScaling: false }), // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
```
Saadnajmi, line 150: For consistency sake, could we put the spread next to the other spreads? 
I also would have expected `...rest` to be the last line so that everything is user-overridable, but I guess that's outside this PR's scope.

packages/components/text/src/Text.tsx
```diff
@@ -106,6 +113,19 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
...
+  // tokenStyle.fontSize and tokenStyle.lineHeight can also be strings (e.g., "14px").
+  // Therefore, we only support scaling for number-based size values in order to avoid any messy calculations.
+  if (dynamicTypeVariant !== undefined && typeof tokenStyle.fontSize === 'number' && typeof tokenStyle.lineHeight === 'number') {
+    const scaleFactor = fontMetricsScaleFactors[dynamicTypeVariant] ?? 1;
+    scaleStyleAdjustments = {
+      fontSize: tokenStyle.fontSize * scaleFactor,
+      lineHeight: tokenStyle.lineHeight * scaleFactor,
+    };
+  }
+  // ]TODO(#2268)
```
Saadnajmi, line 127: To make it more clear this is an iOS only calculation that should be done, can we put this inside a `if (Platform.OS === 'ios')` check? I try to avoid platform specific checks, but this feels like a valid case. I know that the scale adjustments would just return "1", but I don't think that's obvious if someone is skimming this file. In a similar vein, could we add a comment to this block of code to the vein of `"Make iOS specific adjustments to the font scaling to better match Apple's Dynamic Type ramp"`?

packages/components/text/src/Text.tsx
```diff
@@ -79,6 +83,9 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     [onPress, onAccessibilityTap],
   );
 
+  // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
+  const dynamicTypeVariant = Platform.OS === 'ios' ? tokens.dynamicTypeRamp : undefined;
+
```
Saadnajmi, line 88: Any reason this has to be here and not right before line 117? If not, could we add this to be inside the `Platform.OS === 'ios'` check I suggested earlier?

Saadnajmi (approved):
Left some minor comments and approved :) It would be nice if someone on the windows side can quickly confirm this doesn't affect their shipped V1 (V2?) control, but I think in our testing we found it shouldn't!

rurikoaraki:
> Left some minor comments and approved :) It would be nice if someone on the windows side can quickly confirm this doesn't affect their shipped V1 (V2?) control, but I think in our testing we found it shouldn't!

Working on it...

rurikoaraki:
> > Left some minor comments and approved :) It would be nice if someone on the windows side can quickly confirm this doesn't affect their shipped V1 (V2?) control, but I think in our testing we found it shouldn't!
> 
> Working on it...

@amgleitman @Saadnajmi looks good to me

packages/components/text/src/Text.types.ts
amgleitman: Will take offline.

packages/components/text/src/Text.tsx
amgleitman: I think this is cleaner because we also check `dynamicTypeVariant` in the continuation where we define `mergedProps`. This also ensures that `dynamicTypeVariant` is only used when it's actually relevant.

packages/components/text/src/Variants.ios.ts
amgleitman: If we wanted to do this in appleTypography.ios.ts, we would need to wait until `dynamicTypeRamp` becomes part of React Native's `TextStyle` (which should happen in 0.71), so this isn't possible right now.

packages/components/text/src/Text.tsx
amgleitman: I made the `Platform.OS == 'ios'` check as part of defining `dynamicTypeVariant` because we also check the value of `dynamicTypeVariant` in the spread that determines whether or not we should add `{allowFontScaling: false}`.

PR #2274, opened 2022-10-25
Issue with useButton when using an old haul polyfill

acoates-ms:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Packages built using older versions of Haul include a polyfill for Object.assign which does not handle values of false, only objects and null.  This change fixes a case where false was getting passed into Object.assign.

This was a recent regression, so only got hit when bringing in a new version of FURN into fabric-internal.

### Verification

Made this change manually in node_modules of fabric-internal to be able to boot an SDX that wasn't booting without this change.


Saadnajmi (approved):
Thanks for the catch!

PR #2291, opened 2022-10-31
Use global spacing tokens instead of those from `Theme`

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We now use global tokens to get iOS-specific spacing parameters for Notification as an improvement to #2251. As a nice side effect, this undoes the changes previously made to Notification.test.tsx.snap to gives us more accurate values that we would see in the wild, so yay!

### Verification

Things look the same. Not really much to write home about.

| Screen Shot 2022-10-31 at 4 35 19 PM | Screen Shot 2022-10-31 at 4 34 55 PM |

### DIFF
--- a/change/@fluentui-react-native-notification-4e2bf526-3c88-4619-a8a9-3cf10764f7f4.json
+++ b/change/@fluentui-react-native-notification-4e2bf526-3c88-4619-a8a9-3cf10764f7f4.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Use global spacing tokens instead of those from `Theme`",
+  "packageName": "@fluentui-react-native/notification",
+  "email": "adgleitm@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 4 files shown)


Saadnajmi (approved):

lyzhan7:
Do we also want to update the Tester page to grab values from globalTokens instead of from the theme? And also is the eventual plan to remove Spacing altogether from the theme?

amgleitman:
> Do we also want to update the Tester page to grab values from globalTokens instead of from the theme? And also is the eventual plan to remove Spacing altogether from the theme?

Yeah, good catch. I'll do this in a separate PR.

rurikoaraki:
> > Do we also want to update the Tester page to grab values from globalTokens instead of from the theme? And also is the eventual plan to remove Spacing altogether from the theme?
> 
> Yeah, good catch. I'll do this in a separate PR.
> 
> As far as removing spacing tokens from the theme altogether, I'd like to do this just to eliminate any confusion, but that may require some further discussion.

If we want to be aggressive we can spot check the clients we know about and see if any of them are using the spacing entries and remove the entries if nobody is using them. But ideally we'd wait until we're done with our initial V1 components and give clients the chance to move to them before removing the older theming stuff. We could mark it as deprecated though.

PR #2294, opened 2022-11-01
Undo workaround needed for old haul polyfill

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

With #2274 we needed to workaround a JS feature not available in older haul polyfills. We no longer use that haul polypill downstream, so I can bring them back. 

### Verification

CI should be enough

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-button-6399fee5-cc70-4032-9e1f-c1a2aad9cc6e.json
+++ b/change/@fluentui-react-native-button-6399fee5-cc70-4032-9e1f-c1a2aad9cc6e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Undo workaround needed for old haul polyfill",
+  "packageName": "@fluentui-react-native/button",
+  "email": "sanajmi@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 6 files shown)


lyzhan7:
Is there any functional difference between using the ?? conditional rendering operator vs. using the "<...> ? <...> : null" ternary operator? Is the former considered better code style over the latter?

Saadnajmi:
> Is there any functional difference between using the ?? conditional rendering operator vs. using the "<...> ? <...> : null" ternary operator? Is the former considered better code style over the latter?

No difference as far as I know. Apparently the way I was doing didn't work in some old versions of JavaScript. I don't know that one is preferred style, but I like mine better.

lyzhan7 (approved):

lyzhan7:
> > Is there any functional difference between using the ?? conditional rendering operator vs. using the "<...> ? <...> : null" ternary operator? Is the former considered better code style over the latter?
> 
> No difference as far as I know. Apparently the way I was doing didn't work in some old versions of JavaScript. I don't know that one is preferred style, but I like mine better.

Gotcha okay. I don't really have a preference except for the fact that "??" can cause issues if the condition might resolve to an empty string.

PR #2297, opened 2022-11-02
Fix infinite loop in ContextualMenu

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR fixes a hang on tabbing inside a contextual menu. The hang happens in this while loop: `while([nextViewToFocus isDescendantOf:focusZoneAncestor])` in FocusZone where it tries to get to the next key view that's outside of the current FocusZone view. In the context of ContextualMenu, the while loop tries to search for the next valid key view (outside the menu view) but it becomes infinite due to the fact that `isDescendantOf: self` returns true. The expected outcome for tabbing inside a contextual menu should be no op. Let's catch this edge case for the tab and shift-tab scenarios. 

### Verification

Before:

https://user-images.githubusercontent.com/67026167/200396582-2df8e0e2-85e6-4bb9-b834-062d203078d7.mov

After:

https://user-images.githubusercontent.com/67026167/200396618-94e9e8ab-361a-4a82-acef-7fac7410d363.mov

1. Keyboard navigation works as usual in ContextualMenu
2. Ensured the behavior remains unchanged in the FocusZone test page 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -432,6 +432,12 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 		nextViewToFocus = [self previousValidKeyView];
 		while([nextViewToFocus isDescendantOf:focusZoneAncestor])
 		{
+			// there are no views left in the key view loop
+			if ([nextViewToFocus isEqual:focusZoneAncestor])
+			{
+				nextViewToFocus = nil;
+				break;
+			}
 			nextViewToFocus = [nextViewToFocus previousValidKeyView];
 		}
 
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -492,11 +492,8 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
-	else if (viewToFocus == nil)
-	{
-		// No view to focus, do nothing
-	}
-	else
+	// Call super only if there are views to focus
+	else if (viewToFocus != nil)
 	{
 		[super keyDown:event];
 	}
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -418,6 +418,7 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 		nextViewToFocus = [self nextValidKeyView];
 		while([nextViewToFocus isDescendantOf:focusZoneAncestor])
 		{
+			// there are no views left in the key view loop
 			if ([nextViewToFocus isEqual:focusZoneAncestor])
 			{
 				nextViewToFocus = nil;
@@ -458,8 +459,7 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 
 	BOOL passthrough = NO;
 	NSView *viewToFocus = nil;
-	// A flag for handling no op cases
-	BOOL noOp = NO;
+
 	if ([self disabled] || action == FocusZoneActionNone)
 	{
 		passthrough = YES;
@@ -467,10 +467,6 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 	else if (action == FocusZoneActionTab || action == FocusZoneActionShiftTab)
 	{
 		viewToFocus = [self nextViewToFocusOutsideZone:action];
-		if (viewToFocus == nil)
-		{
-			noOp = YES;
-		}
 	}
 	else if ((focusZoneDirection == FocusZoneDirectionVertical
 			&& (action == FocusZoneActionRightArrow || action == FocusZoneActionLeftArrow))
@@ -490,7 +486,7 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
-	else if (noOp)
+	else if (viewToFocus == nil)
 	{
 		// No view to focus, do nothing
 	}
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -418,6 +418,11 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 		nextViewToFocus = [self nextValidKeyView];
 		while([nextViewToFocus isDescendantOf:focusZoneAncestor])
 		{
+			if ([nextViewToFocus isEqual:focusZoneAncestor])
+			{
+				nextViewToFocus = nil;
+				break;
+			}
 			nextViewToFocus = [nextViewToFocus nextValidKeyView];
 		}
 	}
@@ -453,7 +458,7 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 
 	BOOL passthrough = NO;
 	NSView *viewToFocus = nil;
-
+	BOOL noOp = NO;
 	if ([self disabled] || action == FocusZoneActionNone)
 	{
 		passthrough = YES;
@@ -461,6 +466,10 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 	else if (action == FocusZoneActionTab || action == FocusZoneActionShiftTab)
 	{
 		viewToFocus = [self nextViewToFocusOutsideZone:action];
+		if (viewToFocus == nil)
+		{
+			noOp = YES;
+		}
 	}
 	else if ((focusZoneDirection == FocusZoneDirectionVertical
 			&& (action == FocusZoneActionRightArrow || action == FocusZoneActionLeftArrow))
@@ -479,6 +488,10 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 	{
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
+	}
+	if (noOp == YES)
+	{
+		// No view to focus, do nothing
 	}
 	else
 	{
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -458,6 +458,7 @@ static RCTFocusZone *GetFocusZoneAncestor(NSView *view)
 
 	BOOL passthrough = NO;
 	NSView *viewToFocus = nil;
+	// A flag for handling no op cases
 	BOOL noOp = NO;
 	if ([self disabled] || action == FocusZoneActionNone)
 	{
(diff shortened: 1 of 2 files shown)


harrieshin:
Isnt it still valid in some cases the tab should infact take u outside of focuszone?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -480,6 +489,10 @@ - (void)keyDown:(NSEvent *)event
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
+	if (noOp == YES)
```
harrieshin, line 492: Isnt it more simplified if !noop

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -480,6 +489,10 @@ - (void)keyDown:(NSEvent *)event
...
+	{
+		// No view to focus, do nothing
+	}
```
lyzhan7, line 495: Just wanted to double check that it was intentional for the else block below to be part of a different if statement? Also if so, could optionally do `else if (noOp == NO) ...`

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -453,14 +458,18 @@ - (void)keyDown:(NSEvent *)event
 
 	BOOL passthrough = NO;
 	NSView *viewToFocus = nil;
-
+	BOOL noOp = NO;
 	if ([self disabled] || action == FocusZoneActionNone)
```
harrieshin, line 462: The variable doesnt seem very descriptive

harrieshin (changes requested):
I think tab should work outside focuszone

chiuam:
> Isnt it still valid in some cases the tab should infact take u outside of focuszone?

Yes, I made sure tabbing through FocusZones still works as expected for all test cases in the test page. (tabbing into/out of disabled FZ is buggy but not a regression from this change)

chiuam:
> I think tab should work outside focuszone

It does work?

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: The newly added if clause should be inside of that block, so I'll update it

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: Hmm, adding documentation should clarify things up?

chiuam: Let me know if there's any suggestions on renaming this variable

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -418,6 +418,11 @@ - (NSView *)nextViewToFocusOutsideZone:(FocusZoneAction)action
 		nextViewToFocus = [self nextValidKeyView];
 		while([nextViewToFocus isDescendantOf:focusZoneAncestor])
```
mischreiber, line 419: Ahhh I didn't realize that `isDescendantOf:` returns `YES` for `self`. Good to know!

Saadnajmi: we could add a check here `&& nextViewToFocus != self` ?

chiuam: @Saadnajmi That would make the code cleaner but I'd need to set `nextViewToFocus` to nil when `nextViewToFocus == self` returns true? Unless there's a workaround for that 😅

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -480,6 +490,10 @@ - (void)keyDown:(NSEvent *)event
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
+	else if (noOp)
+	{
+		// No view to focus, do nothing
+	}
```
Saadnajmi, line 496: I don't like the idea of `noOp`big here. It prevents the keyboard event from being handled by `super` which means something else in the responder chain (superviews, the window, open menus) won't be able to handle the event. That seems like bad practice to me.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -418,6 +418,11 @@ - (NSView *)nextViewToFocusOutsideZone:(FocusZoneAction)action
 		nextViewToFocus = [self nextValidKeyView];
 		while([nextViewToFocus isDescendantOf:focusZoneAncestor])
 		{
+			if ([nextViewToFocus isEqual:focusZoneAncestor])
+			{
+				nextViewToFocus = nil;
+				break;
+			}
```
mischreiber, line 425: Shouldn't this similar logic exist for shift-tab (the block below)? Presuming we'll crash/hang the same way there.

packages/components/FocusZone/macos/RCTFocusZone.m
mischreiber: Is there a callstack showing why we beachball? What causes the infinite loop when invoking `super`?

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: I will say that I also relied on `isDescendantOf` returning self when I first added that logic... but I can't remember why. Simplified something

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: ^ that's my question. What's the actual infinite loop that is happening? Is `nextValidKeyView == self` ?

chiuam: @Saadnajmi  I feel like FocusZone as a feature overrides the native keyboarding behavior to custom fit our business requirements, so it's okay to deviate from it in this case?

@mischreiber Calling `super` doesn't cause the infinite loop. I intentionally avoid calling `super` because then it'd move focus up/down which we don't want for tab

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Heh... I prefer the opposite so that the default value of a bool is false. I'm willing to go whatever way makes like easier though :D

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: Good point, I didn't check the shift-tab scenario. Will report back...

packages/components/FocusZone/macos/RCTFocusZone.m
Saadnajmi: Even though FocusZone overrides native keyboarding behavior, I still think it has to respect the responder chain. In my view, FocusZone is effectively an NSView that does some extra handling in `keyDown` on arrow keys/tab to ensure focus goes where we like. If we determine that FocusZone doesn't have an action to do, it should call `super` so that the next responder handles the command. 

I think I don't understand why FocusZone needs to handle a keyboard event by explicitly doing nothing. I'll follow up with you offline.

chiuam: > ^ that's my question. What's the actual infinite loop that is happening? Is `nextValidKeyView == self` ?

Yes, `nextValidKeyView == self` is causing the infinite loop

Saadnajmi (approved):
Talked offline, approval with 2 comments:

1) Let's document with comments `while([nextViewToFocus isDescendantOf:focusZoneAncestor])` means "there are no views left in the key view loop

2) Let's not introduce a new `noOp` boolean, and just leave the check as "nextViewToFocus === nil` with the same comment / logic

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -480,6 +490,10 @@ - (void)keyDown:(NSEvent *)event
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
+	else if (noOp)
+	{
+		// No view to focus, do nothing
```
Saadnajmi, line 495: Out of curiosity, if in the noOp case (`nextViewToFocus == nil`) we called `[super keyDown], what happens?

chiuam: In the case of FocusZoneActionTab, calling `[super keyDown]` would move focus to the next menu item (the next valid key view?)

Saadnajmi: I see, thanks!

harrieshin:
before and after video please

harrieshin:
update the rest of the pr description as well please?

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: plot twist: no more boolean 🙃

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -480,6 +492,10 @@ - (void)keyDown:(NSEvent *)event
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
+	else if (viewToFocus == nil)
```
harrieshin, line 495: why does this empty block exisit? it can't be else if (viewToFocus != nil) { [super keydown]} ?

chiuam: It's intentional to not call `[super keydown]` for when `vieweToFocus == nil` returns true because then it'd move focus to the next menu item (in the context menu scenario).

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: Applied the same logic to `FocusZoneActionShiftTab` (duping the same code), there might be a way to make it cleaner though... And thanks for catching this early :)

chiuam:
> update the rest of the pr description as well please?

I'll be updating the PR description with before and after videos. Anything else?

packages/components/FocusZone/macos/RCTFocusZone.m
harrieshin: isn't that the exact same logic i proposed?

chiuam: Followed through offline, addressed in 74c28db9c230ef16aa8b9d24dd5c8662d0399795

harrieshin:
hmm from the video of after, it didn't seem like focus changed based on your tab key? your mouse hover changed the context menu item selection?

harrieshin (approved):

chiuam:
> hmm from the video of after, it didn't seem like focus changed based on your tab key? your mouse hover changed the context menu item selection?

Tab shouldn't move focus, so that's expected?

chiuam:
> might be worth testing arrow key as well after tab?


https://user-images.githubusercontent.com/67026167/200404227-487b3005-c890-4281-8286-15d79a42a488.mov

Up/down works after tab :)

PR #2298, opened 2022-11-03
Add font metrics package to FURN

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is a temporary native module that gives JavaScript access to a subset of `UIFontMetrics`.

Dynamic Type support is on its way into RNCore (as per [this PR](https://github.com/facebook/react-native/pull/35017)), but we want to be able to take advantage of it now, so we're adding this module accordingly. Once this publishes to NPM, we'll use it when rendering a `Text`.

The idea of using a hook to access font metrics data is inspired by [`RCTAppearance`](https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/CoreModules/RCTAppearance.mm) and [`useColorScheme`](https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Utilities/useColorScheme.js), which uses a similar subscription hook to fetch an initial "expensive" (i.e., "synchronously obtained") bit of information.

### Verification

Validated using the draft PR #2269.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/NativeFontMetrics/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "directory": "packages/experimental/NativeFontMetrics"
+  },
+  "devDependencies": {
+    "@fluentui-react-native/eslint-config-rules": "^0.1.1",
+    "@fluentui-react-native/scripts": "^0.1.1",
+    "@types/react-native": "^0.68.0",
+    "@types/use-subscription": "1.0.0",
+    "react": "17.0.2",
+    "react-native": "^0.68.0",
+    "use-subscription": "1.8.0"
```
Saadnajmi, line 38: I think you actually want >= 1.6.0. I think 1.8 is only React 18+. React-native-macos uses the lower version.

rurikoaraki: Wouldn't that mean you want the version to be < 1.8.0?

FWIW via yarn.lock it looks like our current version of RN prefers >=1.0.0 <1.6.0 (RNW is the outlier).

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.tsx
```diff
@@ -0,0 +1,24 @@
...
+  | 'body'
+  | 'headline'
+  | 'title3'
+  | 'title2'
+  | 'title1'
+  | 'largeTitle';
+
+export type ScaleFactors = { [K in TextStyle]: number | undefined };
+
+interface NativeFontMetricsInterface {
```
rurikoaraki, line 20: Out of curiosity which platforms is this meant to run on, all of them? Just iOS? Should the package be stubbed for platforms that won't provide useful values?

amgleitman: Yeah, it's only meant to be run on iOS. We can stub this out for other platforms.

amgleitman: Resolved with a4f879f.

packages/experimental/NativeFontMetrics/package.json
amgleitman: Based on this, I'm using `>=1.0.0 <1.6.0`.

Saadnajmi: My bad, I had the wrong sign

rurikoaraki (approved):

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.h
```diff
@@ -0,0 +1,31 @@
+#import <React/RCTBridgeModule.h>
+#import <React/RCTEventEmitter.h>
+
+@interface FRNFontMetrics : RCTEventEmitter <RCTBridgeModule>
+@end
+
+#import <React/RCTConvert.h>
+
+typedef NS_ENUM(NSInteger, FRNTextStyle) {
```
Saadnajmi, line 9: If useful, there is a `RCT_ENUM_CONVERTER` that sets a default (I see you. do that in the .m file).

I use it in `FRNAVatarCiewManager.m`

change/@fluentui-react-native-experimental-native-font-metrics-53961c7c-6784-4f8c-b197-79ec8ba74332.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add NativeFontMetrics package",
+  "packageName": "@fluentui-react-native/experimental-native-font-metrics",
```
mischreiber, line 4: Why is this listed under `experimental`?

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.h
```diff
@@ -0,0 +1,31 @@
...
+};
+
+@interface RCTConvert (FRNFontMetrics)
+
++ (FRNTextStyle)FRNTextStyle:(nullable id)json;
+
+@end
+
+UIFontMetrics *_Nonnull FRNUIFontMetricsForTextStyle(FRNTextStyle textStyle);
+CGFloat FRNBaseSizeForTextStyle(FRNTextStyle textStyle);
```
mischreiber, line 31: Any reason these two can't/shouldn't be class methods on `RCTConvert`? Why C functions? Sorry if there's a good reason that I don't have familiarity with.

packages/experimental/NativeFontMetrics/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "peerDependencies": {
+    "react": "17.0.2",
+    "react-native": "^0.68.0"
+  },
+  "rnx-kit": {
+    "reactNativeVersion": "^0.68",
+    "reactNativeDevVersion": "^0.68",
+    "kitType": "library",
+    "capabilities": [
```
Saadnajmi, line 48: Not blocking: It would be nice if there was a capability that aligned the `use-subscription` dependency, since I think that one's tricky. I didn't find one from a quick search. We can also probably remove `core-android` since this is an iOS only package.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.m
```diff
@@ -0,0 +1,140 @@
...
+            @(FRNTextStyleBody): UIFontTextStyleBody,
+            @(FRNTextStyleHeadline): UIFontTextStyleHeadline,
+            @(FRNTextStyleTitle3): UIFontTextStyleTitle3,
+            @(FRNTextStyleTitle2): UIFontTextStyleTitle2,
+            @(FRNTextStyleTitle1): UIFontTextStyleTitle1,
+            @(FRNTextStyleLargeTitle): UIFontTextStyleLargeTitle,
+        };
+    });
+
+    id uiFontTextStyle = mapping[@(textStyle)] ?: UIFontTextStyleBody; // Default to body if we don't recognize the specified ramp
```
mischreiber, line 51: Should try to avoid explicit `id` types in code -- this can just be `UIFontTextStyle`.

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.ts
```diff
@@ -0,0 +1,19 @@
+import { ScaleFactors, TextStyle } from './NativeFontMetrics.types';
+
+interface NativeFontMetricsInterface {
+  allScaleFactors(): ScaleFactors;
+  scaleFactorForStyle(style: TextStyle): number;
+}
+
+const NativeFontMetrics: NativeFontMetricsInterface = {
+  allScaleFactors: () => {
+    console.warn('NativeFontMetrics is only available on iOS');
```
Saadnajmi, line 10: What I've seen happen with warnings like these is that stye end up firing on all platforms. Can we quickly check that this doesn't show up on say, macOS on boot?

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.ios.ts
```diff
@@ -0,0 +1,10 @@
+import { NativeModules } from 'react-native';
+import { ScaleFactors, TextStyle } from './NativeFontMetrics.types';
+
+export const NativeFontMetrics = NativeModules.FRNFontMetrics;
+
+interface NativeFontMetricsInterface {
```
Saadnajmi, line 6: I like this is an interface. Should make it easier to convert this to a Turbo Module in the future (if we need to)

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.m
```diff
@@ -0,0 +1,140 @@
...
+            @(FRNTextStyleHeadline): @17,
+            @(FRNTextStyleTitle3): @20,
+            @(FRNTextStyleTitle2): @22,
+            @(FRNTextStyleTitle1): @28,
+            @(FRNTextStyleLargeTitle): @34,
+        };
+    });
+
+    NSNumber *baseSize = mapping[@(textStyle)] ?: @17; // Default to body size if we don't recognize the specified ramp
+    return CGFLOAT_IS_DOUBLE ? [baseSize doubleValue] : [baseSize floatValue];
```
mischreiber, line 76: Do we still need to do `CGFLOAT_IS_DOUBLE` checks? I wasn't aware that we still supported any 32-bit architectures.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.m
```diff
@@ -0,0 +1,140 @@
...
+}
+
+@implementation FRNFontMetrics {
+    BOOL hasListeners;
```
mischreiber, line 80: Instance variables should prefix with `_`.

change/@fluentui-react-native-experimental-native-font-metrics-53961c7c-6784-4f8c-b197-79ec8ba74332.json
Saadnajmi: We tend to prefix packages with `experimental` until they're V1 ready. I personally don't like changing package names so much but 🤷🏾.

For this package in particular, I think it's fair to keep it since we want to eventually use core anyway.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.h
Saadnajmi: Looking at the `.m` file, I think it's so that we can do this?

```
  FRNTextStyle style = [RCTConvert FRNTextStyle:styleString];
  UIFontMetrics *fontMetrics = FRNUIFontMetricsForTextStyle(style);
```

change/@fluentui-react-native-experimental-native-font-metrics-53961c7c-6784-4f8c-b197-79ec8ba74332.json
amgleitman: As Saad said, we don't plan on this being a permanent thing anyway, so experimental is fine.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.h
amgleitman: These enum values have to be defined somewhere first. The values in FRNAvatarViewManager.m are already defined in fluentui-apple.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.h
amgleitman: `RCTConvert` is specifically for deserialization of an `id` to a native value. It's not really meant to do any other logic that these C functions are doing.

packages/experimental/NativeFontMetrics/ios/FRNFontMetrics.m
amgleitman: It doesn't hurt to have it.

packages/experimental/NativeFontMetrics/package.json
amgleitman: The main reason we have to specify this explicitly is because otherwise `depcheck` complains.

Saadnajmi: Oh really? What does it complain? Resolving comment so it's clear this isn't blocking.

PPatBoyd:
Question for future dev guidance: from [Apple's HIG](https://developer.apple.com/design/human-interface-guidelines/foundations/accessibility/) meaningful interface icons should scale with text scaling; if my icon is an SVG, what would be the right way to have it scale with DynamicType / this metrics package such that it scales the same as the automatically-scaling [SF Symbols](https://developer.apple.com/design/human-interface-guidelines/foundations/sf-symbols/)?

packages/experimental/NativeFontMetrics/src/NativeFontMetrics.ts
amgleitman: It doesn't, so we're good.

amgleitman:
> Question for future dev guidance: from [Apple's HIG](https://developer.apple.com/design/human-interface-guidelines/foundations/accessibility/) meaningful interface icons should scale with text scaling; if my icon is an SVG, what would be the right way to have it scale with DynamicType / this metrics package such that it scales the same as the automatically-scaling [SF Symbols](https://developer.apple.com/design/human-interface-guidelines/foundations/sf-symbols/)?

We intend to scale V2 text elements based on this scale, so as long as you match the text styles, scaling your SVGs by the appropriate factor from here should do you well.

As an aside, this won't give the EXACT same font sizes as listed on [Apple's typography page](https://developer.apple.com/design/human-interface-guidelines/foundations/typography#specifications) because `UIFontMetrics` appears to operate based on ratios of the *leading* values, not the *size* values. For example, `.body`-style text on `.xxxLarge` would give you a value of 17×29/22 ≈ 22.41, which is close to, but not exactly the same as, the listed "size" of 23, but this shouldn't matter as much as long as all our text elements are using the same (leading ratio-based) scaling mechanism.

PR #2302, opened 2022-11-04
Add missing color tokens

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Added tokens from pipeline.
### Verification

Build and tests.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theming-utils/src/mapPipelineToTheme.android.ts
```diff
@@ -24,6 +24,9 @@ export function mapPipelineToTheme(pipelineOutput: any): AliasColorTokens {
     neutralForeground3BrandPressed: pipelineOutput.neutralForeground3.fillColorBrandPressed,
     neutralForeground3BrandSelected: pipelineOutput.neutralForeground3.fillColorBrandSelected,
     neutralForegroundInverted: pipelineOutput.neutralForegroundInverted.fillColorRest,
+    neutralForegroundDisabled: pipelineOutput.neutralForegroundDisabled1.fillColorRest,
+    neutralForegroundDisabled1: pipelineOutput.neutralForegroundDisabled2.fillColorRest,
```
ayush547, line 28: I feel these tokens should be either called ('neutralForegroundDisabled' and 'neutralForegroundDisabled2') or ('neutralForegroundDisabled1' and 'neutralForegroundDisabled2'), current naming is misleading for 'neutralForegroundDisabled1'

ankraj12: Valid, updated.

ayush547 (approved):

packages/theming/theme-types/src/Color.types.ts
```diff
@@ -469,6 +469,9 @@ export interface AliasColorTokens {
   neutralForeground3BrandHover: ColorValue;
   neutralForeground3BrandPressed: ColorValue;
   neutralForeground3BrandSelected: ColorValue;
+  neutralForegroundDisabled1?: ColorValue;
+  neutralForegroundDisabled2?: ColorValue;
+  neutralForegroundOnColor?: ColorValue;
 
   neutralForeground4?: ColorValue;
   neutralForegroundDisabled?: ColorValue;
```
rurikoaraki, line 477: post merge comment: I think having both neutralForegroundDisabled and neutralForegroundDisabled1 is redundant, since I'd guess that they're supposed to represent similar values across endpoints... but I guess the naming discrepancy exists in the design tokens themselves, too, so it might be something to bring up to design. That said I'm guessing the discrepancy will stay, at least in the short term, since it would be a breaking change for web and our own theming object.

ankraj12: Just saw this.
Yes - I do agree that `neutralForegroundDisabled` and `neutralForegroundDisabled1` are mostly the same thing - just naming differences between endpoints for the same token. I initially just used neutralForegroundDisabled existing `AliasColorToken`, but it was slightly confusing while usage ( as the designs depict them as `neutralForegroundDisabled1`) - so changed to add another property `neutralForegroundDisabled1`. I would take up a follow up to check on this.

PR #2304, opened 2022-11-07
chore: address Android build issues

tido64:
A recent publishing mistake is breaking all Android build regardless of React Native version. This patch release works around the issue.

For more details, see https://github.com/facebook/react-native/issues/35210.

### DIFF
--- a/apps/fluent-tester/package.json
+++ b/apps/fluent-tester/package.json
@@ -109,7 +109,7 @@
     "metro-config": "^0.67.0",
     "metro-react-native-babel-preset": "^0.67.0",
     "react-native-svg-transformer": "^1.0.0",
-    "react-native-test-app": "^1.6.2",
+    "react-native-test-app": "^2.0.2",
     "react-test-renderer": "17.0.2",
     "ts-node": "^8.10.1",
     "tsconfig-paths": "^3.9.0",
@@ -169,12 +169,11 @@
       "core-ios",
       "core-macos",
       "core-windows",
-      "react",
-      "svg",
-      "metro-config",
       "babel-preset-react-native",
-      "test-app",
-      "react-dom"
+      "metro-config",
+      "react",
+      "react-dom",
+      "svg"
     ]
   },
   "depcheck": {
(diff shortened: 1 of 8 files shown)


kelset (approved):

apps/fluent-tester/package.json
```diff
@@ -109,7 +109,7 @@
     "metro-config": "^0.67.0",
     "metro-react-native-babel-preset": "^0.67.0",
     "react-native-svg-transformer": "^1.0.0",
-    "react-native-test-app": "^1.6.2",
+    "react-native-test-app": "^2.0.2",
```
ankraj12, line 112: I have a basic question to understand the build tooling. Based on the details in issue - https://github.com/facebook/react-native/issues/35210, I thought the suggested fix was to update the react native patch version. So, something like:
`"react-native": "^0.68.5",`

in dependencies above.. which is very different from how the fix here is... how does this work :\``

ankraj12 (commented):
Query on how does the fix fixes

apps/fluent-tester/package.json
ankraj12: So the yarn.lock now shows - version "0.68.5", I just don't understand the react-native-test-app update vs updating the react-native version part ( that's me not knowing things here fully :)). The fix definitely works as the pipelines passed, so will just drop a ping offline to get my head over this part.

ankraj12 (approved):

Saadnajmi (approved):

apps/fluent-tester/package.json
tido64: react-native-test-app adds the workaround for locking the react-native version to the one that is installed locally. Today, practically everyone are consuming react-native via `implementation("com.facebook.react:react-native:+")`, and relying on Gradle to resolve the correct version from `node_modules`. However, this has the nasty side effect that if someone uploads a more recent version on Maven, it will resolve to that version instead. So the incident that started on Friday was caused by 0.71.0-rc.0 being published to Maven. Gradle saw a newer version and downloaded it instead of using the one that was installed under `node_modules`.

You can read more about the issue here: https://github.com/facebook/react-native/issues/35210

PR #2308, opened 2022-11-07
Fix double press behavior for Link

lenahong:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes
- update LinkTokens: focusBorder was an outdated token
- Text and Link: filter out props that are currently not supported on macOS and iOS
-  remove createIconProps hook from interactive-hooks to icon to avoid dep cycle error

### Verification
Same tests on FURN tester

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/text/src/Text.tsx
```diff
@@ -91,15 +106,25 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     ['color', 'fontStyle', 'textAlign', 'textDecorationLine', ...fontStyles.keys],
   );
 
+  const filteredProps = {
+    onKeyUp: Platform.OS === (('win32' as any) || 'windows') ? onKeyUp : undefined,
```
ValentinaKozlova, line 110: Maybe it would be good to move this check to a variable e.g. `isWinPlatform`?

rurikoaraki:
Uhhh I thought some of this was checked in already?

lenahong:
> Uhhh I thought some of this was checked in already? Maybe I'm just imagining it...

I ended up abandoning that pr because adding pressable states to Text before supporting it on RN Text was not going to work. Adding all the necessary changes excluding the native changes in this pr.

packages/experimental/Link/src/useLink.ts
```diff
@@ -60,14 +64,27 @@ export const useLink = (props: LinkProps): LinkInfo => {
 
   const linkTooltip = tooltip ?? url ?? undefined;
 
+  /*These callbacks are not implemented on iOS/macOS, and cause Redboxes if passed in. Limit to only windows/win32 for now*/
+  const filteredProps = {
```
ValentinaKozlova, line 68: Are these filteredProps the same as in Text.tsx? Should it be moved to some helper file to avoid duplication?

lenahong: Currently, filteredProps in Text doesn't include the MouseOver events. I'm not sure how long these would be filtered out for, but if Text.tsx ends up supporting mouseOver events, I'll move them out to a shared location!

Saadnajmi: I don't think we on the macOS side have any plans to implement the missing callbacks for a while. And even if we do, Android and iOS wouldn't get them. It makes sense to keep these filtered props to just the Text component to me though, since it's Text specific. 

Previously, adapters did this for us but 🤷🏾

rurikoaraki (approved):

PR #2309, opened 2022-11-07
Enable accessibilityLevel as a prop for Text on Win32

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The `accessibilityLevel` prop was added to TextWin32 in order to set custom heading levels for Text components defined as headers (via `accessibilityRole="header"`). In order to make this prop available on FURN Text for win32, we must bump versions of @office-iss/react-native-win32 and add the prop to the adapters lists.
### Verification

Manually tested w/ fluent tester.

(coming in a future PR) Add an E2E test case for Text which tests setting the heading level of on a Text component for two scenarios:

- Heading level is set to "1" and accessibilityRole="header" is set (heading level should be 1)
- Heading level is not set but accessibilityRole="header" is set (default level should be 2)
- Heading level is set but accessibilityRole="header" is not set (heading level should be none)

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


apps/fluent-tester/package.json
```diff
@@ -117,7 +117,7 @@
     "webdriverio": "7.23.0"
   },
   "peerDependencies": {
-    "@office-iss/react-native-win32": ">=0.68.0 <0.68.6"
+    "@office-iss/react-native-win32": ">=0.68.0 <=0.68.8"
```
rurikoaraki, line 120: Should we remove the range now that we can move past 0.68.6? That was to keep us from getting an update with a bug in it right?

rurikoaraki (approved):

apps/fluent-tester/package.json
ksiler: Oh sure, I wasn't sure so I just kept the format.

PR #2311, opened 2022-11-07
Fix Callout testID bug and re-enable Callout E2E tests

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Previously, the Callout component had a bug where an assigned `testID` prop would not be correctly passed onto its corresponding native element as an `automationID`. This bug meant that the functional + accessibility E2E tests written for Callout were disabled for the time being.

This PR fixes this by:
1. ~~Upgrading `@office-iss/rex-win32` to the latest devmain version with the Callout `testID` bug fixed.~~ (thanks Krystal)
2. Re-enabling and slightly re-writing the commented out Callout E2E tests.
3. Removing the Callout page + test on the UWP app. Callout does not look like it is properly implemented on the platform, which causes a ViewManager error when running E2E testing in the pipeline. 

This PR also fixes a separate minor issue with the Persona page. The previous image URLs return a 404 error, causing an assert we don't want. I've inserted better links to use to stop the asserts. 

### Verification

Callout functional + accessibility E2E tests pass locally. 
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
+++ b/apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
@@ -1,6 +1,8 @@
 import NavigateAppPage from '../../common/NavigateAppPage';
 import CalloutPageObject from '../pages/CalloutPageObject.win';
-import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT } from '../../common/consts';
+import { CALLOUT_ACCESSIBILITY_LABEL } from '../../../TestComponents/Callout/consts';
+import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT, CALLOUT_A11Y_ROLE } from '../../common/consts';
+import { ComponentSelector } from '../../common/BasePage';
 
 // Before testing begins, allow up to 60 seconds for app to open
 describe('Callout Testing Initialization', function () {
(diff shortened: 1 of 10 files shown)


ksiler:
FYI I just merged a change that bumped the rex-win32 version so you may not need to

apps/fluent-tester/wdio.conf.windows.js
```diff
@@ -11,7 +11,6 @@ exports.config = {
   specs: [
     'src/E2E/ActivityIndicator/specs/*.win.ts',
     'src/E2E/Button/specs/*.win.ts',
-    'src/E2E/Callout/specs/*.win.ts',
```
samuelfreiberg, line 14: Please leave a comment in the description for why we are removing this... "Callout not supported by UWP. Removing test" will do :)

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
```diff
@@ -19,26 +21,40 @@ describe('Callout Testing Initialization', function () {
...
+    await CalloutPageObject.openCallout();
+    await CalloutPageObject.waitForCalloutComponentInView(PAGE_TIMEOUT);
+  });
+
+  it('Validate accessibilityRole is correct', async () => {
+    await expect(await CalloutPageObject.getAccessibilityRole()).toEqual(CALLOUT_A11Y_ROLE);
+    await expect(await CalloutPageObject.didAssertPopup()).toBeFalsy(CalloutPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped up
+  });
+
+  it('Set accessibilityLabel', async () => {
```
samuelfreiberg, line 36: Does a Callout actually have an accessibility label on it? Or is it on the items inside the callout?

What's normal usage? @PPatBoyd Is it necessary to have this test?

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
```diff
@@ -19,26 +21,40 @@ describe('Callout Testing Initialization', function () {
...
+describe('Callout Functional Testing', () => {
+  beforeEach(async () => {
+    await CalloutPageObject.scrollToTestElement();
+    await CalloutPageObject.openCallout();
+    await CalloutPageObject.waitForCalloutComponentInView(PAGE_TIMEOUT);
+  });
+
+  it('Open the callout and validate it loaded correctly (visible)', async () => {
+    await expect(await CalloutPageObject.didCalloutLoad()).toBeTruthy();
+  });
```
samuelfreiberg, line 55: While we're at it, maybe let's add some more tests? We could test the `DismissBehaviors` prop

rurikoaraki: I'd also be ok with more tests being its own PR

apps/fluent-tester/wdio.conf.windows.js
lawrencewin: edited

apps/fluent-tester/src/TestComponents/Persona/styles.ts
```diff
@@ -1,6 +1,6 @@
 export * from '../PersonaCoin/styles';
 
 export const rajeshImageUrl =
-  'https://3er1viui9wo30pkxh1v2nh4w-wpengine.netdna-ssl.com/wp-content/uploads/prod/2016/09/Rajesh-BW-Square.jpg';
+  'https://media-exp1.licdn.com/dms/image/C4E03AQFH5EV54YFN9g/profile-displayphoto-shrink_800_800/0/1517670612089?e=1673481600&v=beta&t=XaxIkit14fANqZy2jQPlPA03FXhl4iQbF24lq3V2XzM';
```
PPatBoyd, line 4: is it possible for us to find safer links, or at least links we explicitly control instead of hard linking to CDNs or sources more liable to changing underneath us?  Perhaps Wikipedia?

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
PPatBoyd: TBH I'm a little unsure from memory, @ksiler do you recall better the usage of accessibilityLabel relative to a Callout? I believe we need it because the window does get added to the accessibility tree as uhhh unsure what control type but can find out, but I'm less clear on when the label would be used (what level of verbosity etc when a callout appears)

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
PPatBoyd: Sadly @rurikoaraki this issue attributes back to our fix for Menu in the rex-win32 tester (vs other host apps); I identified it w/ Larry and we'll take an issue down for later follow up

PPatBoyd (changes requested):

apps/fluent-tester/src/TestComponents/Persona/styles.ts
lawrencewin: I found a rajesh link from the seattle times and a mj link from wikipedia. Should those suffice?

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
samuelfreiberg: Bump

samuelfreiberg (approved):

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
ksiler: I think the control type is "Group" on win32 but I don't recall if an a11y label is needed or not.

ksiler: I would think just the items in the callout would need to be accessible but I'm unsure how someone would place keyboard focus within the callout if there's focusable items in it. Does focus automatically get moved there? And you escape it via "Esc"?

PPatBoyd (approved):

apps/fluent-tester/src/E2E/Callout/specs/Callout.spec.win.ts
PPatBoyd: Generally `Esc` will close any callout, yes (barring host app interactions).  With regards to moving focus there that may not currently be testable with rex-win32, and generally indicates we're in special territory that also relies on host app interactions to place focus inside the callout (this is relatively win32-specific as far as I know) e.g. some applications use `f6` as a navigation key

PR #2316, opened 2022-11-08
Update useViewCommandFocus

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR maps the JavaScript ref.blur to the native ref, which enables the blurring effect to work properly on macOS.

### Verification

Tested hover states on Menu, menu item loses focus on mouse leave.

Before:

https://user-images.githubusercontent.com/67026167/200672625-f4643c84-583d-4f22-be3a-ed8f9629eea4.mov

After:

https://user-images.githubusercontent.com/67026167/200672818-c8e38392-d30f-45cc-bcde-d912a4cdfe7e.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-interactive-hooks-3e4c94c4-418f-487a-9669-47e9a4b76d95.json
+++ b/change/@fluentui-react-native-interactive-hooks-3e4c94c4-418f-487a-9669-47e9a4b76d95.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update useViewCommandFocus",
+  "packageName": "@fluentui-react-native/interactive-hooks",
+  "email": "amchiu@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/utils/interactive-hooks/src/useViewCommandFocus.ts
+++ b/packages/utils/interactive-hooks/src/useViewCommandFocus.ts
@@ -30,14 +30,14 @@ export function useViewCommandFocus(
        */
       if (localRef) {
         localRef.focus = () => {
-          const commands = UIManager.getViewManagerConfig('RCTView').Commands;
-          if ('focus' in commands) {
+          const commands = UIManager.getViewManagerConfig('RCTView')?.Commands;
+          if (commands != null && 'focus' in commands) {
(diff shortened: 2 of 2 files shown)


Saadnajmi (approved):
I would be interested in still removing the need for this Hook altogether, but I think that's a "New Architecture" problem.

Saadnajmi:
Should we also update the win32 version of the hook? Does `ref.current.blur()` work there? @rurikoaraki FYI

rurikoaraki:
> Should we also update the win32 version of the hook? Does `ref.current.blur()` work there? @rurikoaraki FYI

Sounds reasonable. I think blur actually no-ops on win32 but the API is still exposed for consistency so, /shrug

chiuam:
> > Should we also update the win32 version of the hook? Does `ref.current.blur()` work there? @rurikoaraki FYI
> 
> Sounds reasonable. I think blur actually no-ops on win32 but the API is still exposed for consistency so, /shrug

That's interesting. How does win32 handle blur if it too no-ops there? IMO, it might be better to not update win32 to avoid more confusion in the code?

rurikoaraki:
> > > Should we also update the win32 version of the hook? Does `ref.current.blur()` work there? @rurikoaraki FYI
> > 
> > 
> > Sounds reasonable. I think blur actually no-ops on win32 but the API is still exposed for consistency so, /shrug
> 
> That's interesting. How does win32 handle blur if it too no-ops there? IMO, it might be better to not update win32 to avoid more confusion in the code?

I mean "blur" isn't really possible in Windows anyway since /something/ always needs to be focused. RNW has custom focus handling for blur, but we didn't reimpl for win32 so it just does nothing.

I'm fine with either, we can keep it as is in win32.

Saadnajmi:
> I mean "blur" isn't really possible in Windows anyway since /something/ always needs to be focus

Does this mean that the call to `ref.current.blur()` in `useMenuItems` didn't work on _any_ platform? :D

PR #2317, opened 2022-11-08
Less inline function components in Fluent Tester

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

I need to debug some stuff with Flatlist + FocusZone... so I want an example of that in Fluent Tester. I added one in #2028 but I couldn't merge that because it touched a lot of stuff.

Let's break out "part" of the refactor so I can push it in a few smaller changes. This one specifically:

- Moves a lot of Function components out of "FluentTester", so that they aren't re-rendered all the time
- Removes MobileHeader, moving all the logic to Header.
- Adds a padding of 8 to TestComponentView so that on macOS, focus rings aren't clipped by the enclosing scroll view. 

The resulting app should look the same, but (maybe?) be more performant. I'll save the refactoring of "TestList" for a future PR...

### Verification

Tested on macOS, iOS (iPhone and iPad), and Android. All the testers seem to behave fine. There seems to be an extra re-render on Android I'm not sure about, but otherwise things behave well.

image

https://user-images.githubusercontent.com/6722175/207170457-0c4c0566-61e9-423a-9cd7-2285b6d895da.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ValentinaKozlova:
I checked on win32. I used Button component for testing (as it failed in E2E tests). After sometime there's an assertion failure.

ankraj12 (approved):
Would appreciate if you can test the android tester as well!

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -43,6 +45,10 @@ const getThemedStyles = themedStyleSheet((t: Theme) => {
   };
 });
 
+const EmptyComponent: React.FunctionComponent = () => {
+  return <RNText style={fluentTesterStyles.noTest}>Select a component from the left.</RNText>;
```
rurikoaraki, line 49: Does this need to say something other than "left" in mobile?

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -53,31 +59,62 @@ const TestListSeparator = Separator.customize((t) => ({
...
+      <View style={fluentTesterStyles.header}>
+        {/* On iPhone, We need a back button. Android has an OS back button, while desktop platforms have a two-pane view */}
+        {Platform.OS === 'ios' && !Platform.isPad && (
+          <Button
+            appearance="subtle"
+            style={fluentTesterStyles.backButton}
+            onClick={onBackButtonPressedIOS}
+            disabled={!enableBackButtonIOS}
+          >
+            ‹ Back
```
rurikoaraki, line 87: super nit: RTL?

rurikoaraki (approved):

apps/fluent-tester/src/FluentTester.tsx
Saadnajmi: I think this is never rendered on mobile. You just see the test page. Good catch though.

Idea I've had: Replace empty test page with a short description of how to use tester. I.E:
- Look at platform status to see what's available where
- Play around with Theme, etc.

I'm happy to change the wording here to "Select a component from the test list" for now though, even if it doesn't affect mobile.

apps/fluent-tester/src/FluentTester.tsx
Saadnajmi: Hmm.. Let me try. This might be one of those RTL friendly characters. Let's see. 
But also, none of our strings are localized... so LTR is kinda implied...

Saadnajmi: I shall go fix! 
image

rurikoaraki: you might be ok if you remove the chevron and have it just say "back"? Or lose the text and only have an icon?

apps/fluent-tester/src/FluentTester.tsx
Saadnajmi: Fixed from "left" -> "test list".

apps/fluent-tester/src/FluentTester.tsx
Saadnajmi: Fixed! 

Screenshot 2022-12-12 at 3 00 58 PM

Saadnajmi: Ah, noticed your comment about losing the chevron. Hmm.. I think I'll keep it as is as an example of caring about RTL.  I think I would prefer Icon only overall, but would want to make it bigger / bolder. I kinda wanna revisit the header overall but that's out of scope of this PR.

PR #2318, opened 2022-11-09
Bump token packages to 0.31.0

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Update all design-tokens packages in FURN to ~~0.29.0~~ 0.31.0 + update FURN accordingly

I tried to organize my commits - might be easier to review commit by commit. I know this is a pretty big PR - thinking next time I can try to be more intentional with what's included in each package update so that I'm not bringing in too much change at once. If anyone has any other suggestions for how to keep PRs like this small please let me know.

Changes include:
- Stroke width token ramp schema changes from t shirt sizes to numbers according to the old win32/macOS stroke width definitions
  - stroke.width.thin -> stroke.width10
  - stroke.width.thick -> stroke.width20
  - stroke.width.thicker -> stroke.width30
  - stroke.width.thickest -> stroke.width40
- corner radius token ramp schema changes from t shirt sizes to numbers according to the old win32/macOS stroke width definitions
  - corner.radius.none -> corner.radiusNone
  - corner.radius.small -> corner.radius20
  - corner.radius.medium -> corner.radius40
  - corner.radius.large -> corner.radius60 (*not sure about - could also be corner.radius80)
  - corner.radius.extraLarge -> corner.radius80 (*not sure about - could also be corner.radius120)
  - corner.radius.circle -> corner.radiusCircular
- We decided to add a naming rule to the pipeline where font.size[100] becomes font.size100. This was to improve type safety, because something like font.size[99] would not cause any compiler errors, whereas font.size99 would
  - Most of the naming changes in this PR are trivial ex. color.grey[10] -> color.grey10
  - There was a less trivial change I had to make in Text.tsx - i.e. globalTokens.font['size' + size]. I'm not sure if this is the best solution
  - The changes related to Shadow were also not trivial since there was code specific to the old schema
- I ran into some issues with defaultColors due to differences in the color brand ramp token definition schema- ended up splitting up this file by platform

Screenshot 2022-11-09 at 1 55 08 PM

  - defaultColors.win32.ts and defaultColors.macOS.ts are copies of what defaultColors.ts used to be; the references to any brand tokens follow this format: globalTokens.color.brand.shade10
  - on other platforms (defaultColors.ts) the references to any brand tokens follow this format: globalTokens.color.brand10
 
~~Out of scope:~~ now included in this PR
~~- The corner radius updated naming and a small stroke width update are in a newer package version (0.30.0) - planning to do that in another PR to keep this one a little smaller~~
~~- The current iOS color brand ramp in the tokens pipeline is outdated, which is why some of the updated snapshots after the defaultColors.ts change now have some colors as undefined. This will get updated in another PR~~

### Verification

No visual changes, although I think some stroke width thickness went from 1.5 -> 1, but these were actually supposed to be 1 to begin with. Made sure the tester starts up without issues on iOS and that none of the pages cause unexpected errors

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):
Components with updated snapshots should probably be double checked, but otherwise LGTM

Saadnajmi:
Jk actually more files loaded in my view, I'll keep looking at this lol

lyzhan7:
> Components with updated snapshots should probably be double checked, but otherwise LGTM

The snapshots in the [Update snapshots affected by spacing rename](https://github.com/microsoft/fluentui-react-native/pull/2318/commits/93188485434f67446fc593eb5478225e5298a70a) commit are expected, when I added globalTokens.ios.ts some snapshots got changed because at that point the ios stroke width definitions were different from other platforms, I think the changes in this PR are just returning those snapshots to what they were before.

The snapshots in [Update snapshots after previous commit](https://github.com/microsoft/fluentui-react-native/pull/2318/commits/b55d95119b006914d3a99d63fe017af906a57950) (after the defaultColor.ts changes) I'm less sure about - currently a bunch of values got replaced with "undefined" because the iOS brand color ramp is not updated yet (there's a new package, 0.31.0, that I will update to after this that has the brand color update, and those undefined values will get populated with actual values then. There are a couple other changes (ex. color: link) that I'm not sure about.

I do think at some point it would be a good idea to go through each component and check that it has all the right tokens with the right values, there's been a lot of changes to tokens and it's possible that some regressions have slipped in

packages/components/text/src/Text.tsx
```diff
@@ -69,7 +69,7 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     color,
     variant,
     fontFamily: font == 'base' ? 'primary' : font,
-    fontSize: globalTokens.font.size[size],
+    fontSize: globalTokens.font['size' + size],
     fontWeight: globalTokens.font.weight[weight] as FontWeightValue,
```
Saadnajmi, line 73: Considering this is one of our active V1 controls we're about to ship, could we update this so it doesn't use the non-type-safe method of using `[]` to access properties? @rurikoaraki @lenahong fyi

Saadnajmi: To be clear, I don't think this is blocking this change.

packages/theming/default-theme/src/__tests__/__snapshots__/default-theme.test.ts.snap
```diff
@@ -798,8 +798,8 @@ Object {
     "brandForegroundLinkSelected": "#2899f5",
     "brandStroke1": "#2899f5",
     "brandStroke2": "#004c87",
-    "brandedBackground": "#2b88d8",
-    "brandedBorder": "#deecf9",
+    "brandedBackground": undefined,
+    "brandedBorder": undefined,
     "brandedCheckedBackground": "#484644",
     "brandedCheckedContent": "#faf9f8",
```
Saadnajmi, line 804: There's a few of these `undefined` added. Do we know what changed? That seems bad.

packages/theming/default-theme/src/defaultColors.ts
```diff
@@ -263,14 +273,21 @@ export function getStockWebDarkPalette(): ThemeColorDefinition {
...
         themeTertiary: '#235a85',
-        themeLight: globalTokens.color.brand.shade30,
-        themeLighter: globalTokens.color.brand.shade50,
-        themeLighterAlt: globalTokens.color.brand.shade60,
+        // @ts-ignore
+        themeLight: globalTokens.color.brand40,
+        // @ts-ignore
+        themeLighter: globalTokens.color.brand40,
+        // @ts-ignore
+        themeLighterAlt: globalTokens.color.brand40,
```
Saadnajmi, line 290: Why do we need all these @ts-ignore comments?

Saadnajmi: I would also rather we group all these together to one chunk so its more obvious "these tokens are non-standard"

packages/theming/default-theme/src/defaultColors.win32.ts
```diff
@@ -0,0 +1,510 @@
+import { Palette, FabricWebPalette, ThemeColorDefinition } from '@fluentui-react-native/theme-types';
```
Saadnajmi, line 1: Could you explain the difference in actual implementation between `defaultColors` and `defaultColors.macos/.win32` ?

packages/theming/theming-utils/src/mapPipelineToShadow.ts
```diff
@@ -2,22 +2,22 @@ import { ThemeShadowDefinition } from '@fluentui-react-native/theme-types/src/Sh
...
+    shadow16: { ambient: pipelineOutputShadow.global.shadow16[0], key: pipelineOutputShadow.global.shadow16[1] },
+    shadow28: { ambient: pipelineOutputShadow.global.shadow28[0], key: pipelineOutputShadow.global.shadow28[1] },
+    shadow64: { ambient: pipelineOutputShadow.global.shadow64[0], key: pipelineOutputShadow.global.shadow64[1] },
+    shadow2brand: { ambient: pipelineOutputShadow.global.shadowBrand2[0], key: pipelineOutputShadow.global.shadowBrand2[1] },
+    shadow4brand: { ambient: pipelineOutputShadow.global.shadowBrand4[0], key: pipelineOutputShadow.global.shadowBrand4[1] },
+    shadow8brand: { ambient: pipelineOutputShadow.global.shadowBrand8[0], key: pipelineOutputShadow.global.shadowBrand8[1] },
+    shadow16brand: { ambient: pipelineOutputShadow.global.shadowBrand16[0], key: pipelineOutputShadow.global.shadowBrand16[1] },
+    shadow28brand: { ambient: pipelineOutputShadow.global.shadowBrand28[0], key: pipelineOutputShadow.global.shadowBrand28[1] },
+    shadow64brand: { ambient: pipelineOutputShadow.global.shadowBrand64[0], key: pipelineOutputShadow.global.shadowBrand64[1] },
```
Saadnajmi, line 21: @TravisSpomer is this change intentional? Seems weird to see `.global` for the shadow alias tokens.

packages/theming/default-theme/src/__tests__/__snapshots__/default-theme.test.ts.snap
lyzhan7: Yeah I noted this in the PR description and also in my other comment https://github.com/microsoft/fluentui-react-native/pull/2318#issuecomment-1309390185. It's because this package has the outdated iOS color brand ramp. I could get this PR up to 0.31.0, but that would require making more changes that were part of the 0.30.0 update. After this PR is merged I'm planning to immediately start bumping the packages to 0.31.0, at that point there shouldn't ne any undefined values anymore

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Because ESLint is referring the design-tokens-windows global tokens file, which has a different schema for color brand tokens (i.e. globalTokens.color.brand.shade30 is defined in win32/macOS, but globalTokens.color.brand30 is defined in mobile)

Also - thanks for making this comment because I just realized I put brand40 in a bunch of places as a placeholder since I didn't know what the equivalent was (ex. for globalTokens.color.brand.primary) - this should be fixed before I merge. I'll start a conversation about this offline

packages/theming/default-theme/src/defaultColors.win32.ts
lyzhan7: Yes - I should have done that in the PR description. Will update that

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Also just to add on - if we're okay with some global tokens having different schema across platforms, I don't know if there's a way around 1) the suggested/autocomplete values for the token values is sometimes wrong, and 2) having to add $ts-ignore comments, because ESLint is always assuming we're using the windows design-tokens package

lyzhan7: Ex. if you go to any reference to globalTokens.<something> and click go to reference, it always takes you to the windows design-tokens package in node modules

packages/theming/default-theme/src/defaultColors.win32.ts
lyzhan7: Updated. Let me know if it makes sense now

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
```diff
@@ -116,9 +114,8 @@ exports[`Default FAB (iOS) 1`] = `
       Object {
         "alignItems": "center",
         "alignSelf": "flex-start",
-        "backgroundColor": "#0078d4",
+        "backgroundColor": undefined,
         "borderBottomWidth": undefined,
-        "borderColor": "#005a9e",
```
rurikoaraki, line 121: Hm is the borderColor disappearing on buttons an issue?

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Sure I can start making that change in the tokens repo, and once that change makes it into a new package I can do another update in FURN. I think we'll still need @ts-ignore comments, but they would be in the .win32.ts/.macOS.ts file (although it sounds like we want to update macOS to match web as well)

rurikoaraki:
Btw thanks for breaking down your changes into individual commits. Made it a lot easier to get through this review.

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Hmmm I'm not sure, looking into it

packages/components/text/src/Text.tsx
TravisSpomer: Side note that may or may not be helpful at all: TypeScript recently added a weird little thing where you can [dynamically build types using a `${string}_interpolation_${pattern}`](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).

packages/theming/theming-utils/src/mapPipelineToShadow.ts
Saadnajmi: I see. So Shadows are in a weird "global but not global state". Can we...snap it to one or the other? 😅 In the short term, I guess it's OK to have the namespace here be weird since people who use FURN won't need to touch the names

TravisSpomer: Conceptually they're global tokens because they're building blocks with no semantic meaning. Technically they're alias tokens because they refer to other tokens (colors). So they're... global alias tokens. I think we're kind of stuck with them being a weird case.

packages/theming/default-theme/src/__tests__/__snapshots__/default-theme.test.ts.snap
Saadnajmi: I'm not sure how I feel about our repo being in an intermediate broken state, especially because we publish in every change. Do you have an idea how much extra work the bump to 0.30 and 031 are?

lyzhan7: The main thing that 0.30.0 has is the corner radius updated ramp, there's about 70 references I would need to update. All 0.31.0 has is the updated color ramp, overall it's not that much extra work. Do you consider this change as introducing a regression?

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: I'm not sure why the borderColor disappeared, but if I tried updating the iOS design-tokens package locally to match android, and the borderColor comes back, so I think it would be fixed after updating to 0.31.0

packages/theming/default-theme/src/__tests__/__snapshots__/default-theme.test.ts.snap
Saadnajmi: Yes, I would. It would be pretty bad if partners snapped to this version unknowingly, and had a bunch of broken controls in the interim.

lyzhan7: Talked offline - the newly "undefined" tokens only affect iOS, so this change might not be considered a regression. I am still considering updating this PR to 0.31.0 though

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
```diff
@@ -116,9 +114,8 @@ exports[`Default FAB (iOS) 1`] = `
       Object {
         "alignItems": "center",
         "alignSelf": "flex-start",
-        "backgroundColor": "#0078d4",
+        "backgroundColor": undefined,
```
ankraj12, line 117: Is this an expected change - would it impact any visuals in non iOS components..?

packages/theming/default-theme/src/defaultColors.ts
ankraj12: Just thinking around using desing-tokens-windows acting as the type files for all endpoints - can we instead creating a `globalToken` type file in FURN - that would be a superset of types instead of  desing-tokens-windows acting as it? Unless we really use the design-tokens-windows for more reasons. 
Not a review on this PR  - but something for a thought ahead.

ankraj12 (approved):
The changes in the PR look good, I have dropped a couple of discussions that can be taken separately.

packages/theming/default-theme/src/defaultColors.ts
```diff
@@ -31,7 +32,7 @@ export function paletteFromFabricColors(p: FabricWebPalette, isInverted?: boolea
     defaultStateBackground: p.neutralLighterAlt,
```
ankraj12, line 32: Is there a possibility to code share not platform specific code between defaultColors and implementations - most of the code is same across the 2 files..

ankraj12 (approved):

rurikoaraki:
> Just thinking around using desing-tokens-windows acting as the type files for all endpoints - can we instead creating a globalToken type file in FURN - that would be a superset of types instead of desing-tokens-windows acting as it? Unless we really use the design-tokens-windows for more reasons.
Not a review on this PR - but something for a thought ahead.

We want types to come from the pipeline as much as possible - we should not be defining types ourselves inside the repo. That leads to issues with us having to fix up type definitions as we import new versions and lagging behind.

I don't know how easy it is to have the schema be different from what's actually in the JSON file, that would be something that needs to be looked into. Moving the tokens to TS would also help, but changing the schema would probably be less work than moving all tokens to TS.

packages/components/text/src/Text.tsx
lyzhan7: Made a task for this on ADO, going to resolve this conversation. If anyone has any thoughts on using 'size' + size in this PR feel free to reopen

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Made an ADO item to update the windows schema file

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: I agree there is a lot of duplication. I'm not sure if there is a good way to do this though, from what I've seen other files that got split up by platform usually have a lot of duplication

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Yes it's expected, no should not impact any of the non-iOS platforms. Might get updated if I can bump to 0.31.0 instead of 0.29.0

rurikoaraki (approved):

yarn.lock
```diff
@@ -1457,40 +1457,35 @@
...
-"@fluentui-react-native/design-tokens-windows@^0.14.0":
-  version "0.14.0"
-  resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-windows/-/design-tokens-windows-0.14.0.tgz#1483caa1dd4e7097ac48a5872e93d7f896e8402c"
-  integrity sha512-WNNVTMpRir6a2Voa0yu3BZJmy2w7fofrokL+nyFQM40bACtX68KDRDE96HA5W5+4q8rEaNSXV3DzSzLXm6ekfA==
+"@fluentui-react-native/design-tokens-android@^0.31.0":
+  version "0.31.0"
+  resolved "https://registry.yarnpkg.com/@fluentui-react-native/design-tokens-android/-/design-tokens-android-0.31.0.tgz#52a1209522f7d8eb8530b3b31cd1a63267f5483f"
+  integrity sha512-WjJBHW+Sd8CFQl6QnaoD/QMtMXJOCH2iPhCvosil7Z9zlDsjw1BBwU86FcqU1TK8W+oBZf+HYuYMIXu8bogUzg==
+
+"@fluentui-react-native/design-tokens-ios@^0.29.0":
```
rurikoaraki, line 1465: nit: looks like there's still a package referencing the older version of iOS and macOS tokens

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
Saadnajmi: It's back!

packages/theming/default-theme/src/__tests__/__snapshots__/default-theme.test.ts.snap
lyzhan7: Resolved - updated to 0.31.0 (although there is still an open question I have about 0.31.0 for design)

packages/theming/default-theme/src/defaultColors.ts
rurikoaraki: Generally it would be handled by generating part of the object in a function defined in a forked file while the common parts stay in a non-forked file.

packages/components/Button/src/FAB/__snapshots__/FAB.test.tsx.snap
lyzhan7: Updated to 0.31.0, there should be no additional undefined values now

Saadnajmi (approved):
Approved with suggestions:

1) Let's make sure that our `yarn.lock` changes are clean and only reference 0.31
2) For components that have updated snapshots, it would be good to follow up with component owners that they still look right on platforms they're planning to ship on. EX: @ValentinaKozlova for Avatar and Badge on win32. 
3) Would like to follow up on the defaultColors platform specific files. I understand we need them but I'd still prefer we minimize the unnecessary platform differences as soon as we can. I agree it's out of scope for this PR.

yarn.lock
lyzhan7: Ooh thank you I forgot to save my changes to the apple-theme package.json

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Saad do you think still we should "group all these together to one chunk so its more obvious "these tokens are non-standard"? Otherwise will resolve this thread

lyzhan7:
Blocking issues for this PR
- Waiting to hear back about what the mapping from the old brand ramp to the new brand ramp should be. In the updated defaultColors.ts file, I put brand40 in a bunch of places but that should be updated
- (maybe not blocking) waiting to hear back about how to handle the the old corner radius token ramp having different values for mobile vs. desktop. For this PR I went ahead with making sure the tokens kept the same values for win32, but I'm not sure if there are some places we actually want to use the mobile values

rurikoaraki:
> Blocking issues for this PR
> 
> * Waiting to hear back about what the mapping from the old brand ramp to the new brand ramp should be. In the updated defaultColors.ts file, I put brand40 in a bunch of places but that should be updated
> * (maybe not blocking) waiting to hear back about how to handle the the old corner radius token ramp having different values for mobile vs. desktop. For this PR I went ahead with making sure the tokens kept the same values for win32, but I'm not sure if there are some places we actually want to use the mobile values

For the second one, this is where forking styling for different platforms would help. We definitely shouldn't change the style of Avatar on win32 since it's shipped.

lyzhan7:
> > Blocking issues for this PR
> > 
> > * Waiting to hear back about what the mapping from the old brand ramp to the new brand ramp should be. In the updated defaultColors.ts file, I put brand40 in a bunch of places but that should be updated
> > * (maybe not blocking) waiting to hear back about how to handle the the old corner radius token ramp having different values for mobile vs. desktop. For this PR I went ahead with making sure the tokens kept the same values for win32, but I'm not sure if there are some places we actually want to use the mobile values
> 
> For the second one, this is where forking styling for different platforms would help. We definitely shouldn't change the style of Avatar on win32 since it's shipped.

Yeah there's an AvatarTokens.ts and AvatarTokens.win32.ts, I was thinking maybe we would want to modify AvatarTokens.ts or make mobile specific files. I don't think this is blocking though, up until recently we were referencing windows tokens on iOS so the actual values we were using are still the same as before

rurikoaraki:
> > > Blocking issues for this PR
> > > 
> > > * Waiting to hear back about what the mapping from the old brand ramp to the new brand ramp should be. In the updated defaultColors.ts file, I put brand40 in a bunch of places but that should be updated
> > > * (maybe not blocking) waiting to hear back about how to handle the the old corner radius token ramp having different values for mobile vs. desktop. For this PR I went ahead with making sure the tokens kept the same values for win32, but I'm not sure if there are some places we actually want to use the mobile values
> > 
> > 
> > For the second one, this is where forking styling for different platforms would help. We definitely shouldn't change the style of Avatar on win32 since it's shipped.
> 
> Yeah there's an AvatarTokens.ts and AvatarTokens.win32.ts, I was thinking maybe we would want to modify AvatarTokens.ts or make mobile specific files. I don't think this is blocking though, up until recently we were referencing windows tokens on iOS so the actual values we were using are still the same as before

Ah yeah as long as win32 styling isn't changing this is fine. I think you'd end up revisiting the tokens when iOS/Android implements Avatar. I misread and thought you meant that the win32 tokens were being aligned to mobile ones 😅

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Think I'm going to leave this file for now, maybe in the future we can try to consolidate our component token files to share more code. For this file specifically it sounds like we want to merge some theming stuff anyway

packages/theming/default-theme/src/defaultColors.ts
lyzhan7: Should be grouped together now

lyzhan7:
> Blocking issues for this PR
> 
> * Waiting to hear back about what the mapping from the old brand ramp to the new brand ramp should be. In the updated defaultColors.ts file, I put brand40 in a bunch of places but that should be updated
> * (maybe not blocking) waiting to hear back about how to handle the the old corner radius token ramp having different values for mobile vs. desktop. For this PR I went ahead with making sure the tokens kept the same values for win32, but I'm not sure if there are some places we actually want to use the mobile values

I found a [mapping](https://master--628d031b55e942004ac95df1.chromatic.com/?path=/docs/concepts-migration-from-v8-color-mapping--page) for the brand ramp tokens from Fluent UI web and just updated this PR with those values, but still waiting to get confirmation that this is the ramp we should be using. Once I get confirmation I'll merge this PR

lyzhan7:
@gulnazsayed just FYI this PR touches radio tokens. I changed border.radius.thin is refer to border.radius10, which has value 1px. If there are any issues with that let me know

rurikoaraki:
> @gulnazsayed Gulnaz Sayed FTE just FYI this PR touches radio tokens. I changed border.radius.thin is refer to border.radius10, which has value 1px. If there are any issues with that let me know

FYI that some token changes were just merged for radio tokens

lyzhan7:
Did not get confirmation from design that the mapping I used is correct, but after some conversation offline it seems like what mapping gets used here isn't a blocking issue. 

Users that 1) are on iOS/android and 2) don't specify a theme, so the fallback default theme is used, may be affected by what the colors map to, but iOS SDXes not using the apple theme are not in a supported state right now.

PR #2328, opened 2022-11-10
[MenuV1] Add ability to pass ref into child of MenuTrigger

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Per client request, we want to allow a ref to be passed into the child of the MenuTrigger. The complication is that Menu also passes in a ref to the child of the MenuTrigger in order to position the MenuPopover and anchor it to the correct component.

This can be handled by merging refs, which is a concept that is used by React Native and FluentUI React. This PR copies the implementation of useMergedRefs from FluentUI React, and I'm reaching out to the team to see if sharing their package is possible.

While implementing this functionality, I ran up against some limits with reusing code because the inner render doesn't allow hooks, but we also need to modify the child's props which is only available in the inner render. Through snapshot tests it seems that converting the MenuTrigger into a regular FunctionComponent doesn't add any additional entires to the DOM, and since it was a stagedComponent changing the component to an FC doesn't remove any API from it. So I went ahead and changed the MenuTrigger to be a regular FC, which allows us to use hooks to set up the child's props, resulting in cleaner code and removing getRevisedProps.ts.

### Verification

Added a test for the ref case.
Tested Menus in the tester through interaction and going through the Menu test page.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Change is broken down in commits

packages/components/Menu/src/MenuTrigger/useMergeRefs.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copied from @fluentui/react-utilities
+ *
+ * Keeping the file local to Menu for now while looking into whether
+ * we can pull in @fluentui/react-utilities package
```
rurikoaraki, line 5: Adding to interactive-hooks would add to public API and mean that if we change the source of the functionality, it would be taken back.

apps/fluent-tester/src/TestComponents/Menu/MenuRefs.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const MenuTriggerChildRef: React.FunctionComponent = () => {
+  const ref = React.useRef(null);
+  const onClick = React.useCallback(() => {
+    ref?.current?.focus();
+  }, [ref]);
+
+  return (
+    <Stack style={stackStyle}>
+      <Button onClick={onClick}>Click to focus</Button>
```
Saadnajmi, line 15: nit: "Click to focus on Menu Trigger" I've seen this in other tests and have been confused if clicking on the button focuses on itself or some other control.

apps/fluent-tester/src/TestComponents/Menu/MenuRefs.tsx
```diff
@@ -0,0 +1,29 @@
...
+    ref?.current?.focus();
+  }, [ref]);
+
+  return (
+    <Stack style={stackStyle}>
+      <Button onClick={onClick}>Click to focus Menu trigger</Button>
+
+      <Menu>
+        <MenuTrigger>
+          <Button componentRef={ref} />
```
chiuam, line 19: Is it intentional to not have a button label?

apps/fluent-tester/src/TestComponents/Menu/MenuRefs.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const MenuTriggerChildRef: React.FunctionComponent = () => {
+  const ref = React.useRef(null);
+  const onClick = React.useCallback(() => {
+    ref?.current?.focus();
+  }, [ref]);
+
+  return (
+    <Stack style={stackStyle}>
+      <Button onClick={onClick}>Click to focus Menu trigger</Button>
```
chiuam, line 15: When "Click to focus Menu trigger" button is clicked, should the menu item or the other button get focus? This is what I'm seeing on Mac:

https://user-images.githubusercontent.com/67026167/201215603-fa4cbe68-7d97-4e2c-bcc6-eb73a11cd0b8.mov

Saadnajmi: Is this because we have the `onClickWithFocus` hook and the timing of that on macOS and win32 is different? AKA focus is going from MenuTrigger -> Button on macOS, but Button -> MenuTrigger on macOS ?

chiuam: @Saadnajmi yup, I opted out the focus ref in the `useOnPressWithFocus` hook and that fixed it

chiuam: Now my question is why this is a macOS specific issue

rurikoaraki: It's not - it just flickers on win32 enough that I know that the ref.current.focus() is working. I couldn't really think of a better way to test it.

apps/fluent-tester/src/TestComponents/Menu/MenuRefs.tsx
rurikoaraki: Nope, I'll add one.

apps/fluent-tester/src/TestComponents/Menu/MenuRefs.tsx
Saadnajmi: We have a backlog item to actually _not_ use that hook because I think native macOS controls should actually not be "stealing" focus on click most of the time. I interpret this as "The test was buggy to begin with" so it's kind of a moot point / out of scope. I'm ok with leaving it in the PR since I don't know how to test it either...

packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -23,7 +24,33 @@ export const useMenuTrigger = (): MenuTriggerState => {
...
+
+    return baseAccessibilityActions;
+  }, [childAccessibilityActions]);
+
+  const accessibilityState = React.useMemo(() => {
+    const baseState = open ? expandedState : collapsedState;
+
+    if (childAccessibilityState) {
+      return { ...baseState, ...childAccessibilityState };
+    }
```
Saadnajmi, line 50: Dumb question. I noticed that  React Native core doesn't memoize intermediate objects created in a component, but rather memoizes the component itself. Is that something we might want to follow? I wouldn't use them as the example of perfect React code (a lot of it is still legacy..) but thought I'd bring itup.

packages/components/Menu/src/MenuTrigger/MenuTrigger.types.ts
```diff
@@ -1,7 +1,12 @@
+import * as React from 'react';
 import { InteractionEvent, PressablePropsExtended } from '@fluentui-react-native/interactive-hooks';
 
 export const menuTriggerName = 'MenuTrigger';
 
+export interface MenuTriggerProps {
+  children?: React.ReactElement | null;
+}
```
Saadnajmi, line 8: We need to add this so that we can access the inner render? Is there some other type we can use that `MenuTriggerProps` can extend so we don't have to explicitly do this?

Saadnajmi (approved):

packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
rurikoaraki: I feel like we should lean more towards general React practices rather than copying what their codebase does... not a strong opinion though.

Saadnajmi: That's fine. This is more me not knowing what the general practice for how much to actually memoize is. Resolving this comment anyway.

packages/components/Menu/src/MenuTrigger/MenuTrigger.types.ts
rurikoaraki: MenuTrigger doesn't have an inner render. An explicit prop type was added because it was needed for the render function, since it's now using a regular FC.

To be clear our MenuTrigger doesn't have any props outside of children. All of its stuff comes from the context or the child.

Saadnajmi: So could we just set it's type to `React.FunctionComponent` or `React.FC`?

rurikoaraki: We can, although it's less limiting. Specifying the children prop like this restricts the child to always be one element (instead of say accepting multiple children or an array or something).

I can change it to use FunctionComponent<{}> though

Saadnajmi: If it's better than `FunctionComponent<{}>` because it only supports one child, then that makes sense. I'd maybe ask to make that it's own type to make that clear and then have `MenuTriggerProps` extend that. But.. not a strong opinion. Do what feels best.

rurikoaraki: Yeah keeping it as it is would also keep the code cleaner since it doesn't have to check for or deal with string or number children.

I'm trying to extract the type out into its own type but having a hard time getting TS to cooperate. I'll try a bit more but if I can't get it to work I'll probably merge as is, sorry.

chiuam (approved):

packages/components/Menu/src/MenuTrigger/MenuTrigger.types.ts
rurikoaraki: And I got it to work. let me know if you were expecting something different.

PR #2329, opened 2022-11-10
FocusZone (macOS): move focus between first responders instead of key views

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is a simpler version of #2267 . Instead of adding a new prop to differentiate whether we focus on key views vs first responders, this one will just... focus on first responders unconditionally.

The reasoning is as follows:

1) FocusZone itself doesn't override `canBecomeKeyView. So FocusZone itself respects the system keyboard navigation preference.
2) FocusZone itself is pretending to be "one big focusable view" to the eyes of Appkit. Its job is to move focus inside itself between focusable views. Once it gains focus (via the key view loop or programmatically), it probably doesn't matter whether its children were key views. If you're using a FocusZone for your custom control, you probably still want the keyboard events to work regardless.

How this affects components:

- For the case of ContextualMenu, we call `focusZoneRef.current.focus()` once the menu is opened. This will force focus on the Menu regardless of the OS preference (which we want), and then with this change keyboard navigation will work.
- For the case of Radiogroup: Native Appkit RadioGroups are not key views if the keyboard navigation preference is disabled, and you can't move focus or arrow key through them at all. With our JS Radiogroup (and Native RadioGroup), this behavior should be the same, since the enclosing FocusZone will not become a key view.

### Verification

Locally side loaded the built `libRCTFocusZone.a` to test a partner scenario, and it worked fine with no regressions. Tested ContextualMenu and RadioGroup and didn't find any regressions. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
An issue with this PR is it makes scroll views inside a FocusZone always focusable. I found a simple fix upstream in React Native macOS, so I'll go do that before merging this.

harrieshin (approved):

PR #2330, opened 2022-11-10
FocusZone (macOS): Ensure we passthrough unhandled keyboard events

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There was a regression from #2297 , where FocusZone no longer correctly called `[super keyDown:event]` for events it didn't handle. This broke a few cases:

1) Escape to close Callouts / Menus
  - Since FocusZone never called `super` the "escape" key event never made it's way up the responder chain to the CalloutWindow, where it could be handled to close itself
  
2) Nested FocusZone:
  - Although it's officially unsupported, we do have a test case for two horizontal FocusZones nested inside a vertical FocusZone. The inner horizontal FocusZone would not properly pass the "ArrowUp/Down" events to the next responder (its superview: the vertical FocusZone), and thus the test case broke. 
  
  
Followup: Enabling more of the macOS E2E tests would have caught this. So.. we should totally do that. 

### Verification

Tested that "ArrowUp/Down" in a Menu work as expected. Tested that "Escape" still closes a menu, and that "Tab" still no-ops instead of hanging. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -465,13 +465,19 @@ static BOOL ShouldSkipFocusZone(NSView *view)
 		viewToFocus = [self nextViewToFocusWithFallback:action];
 	}
 
-	if (passthrough) {
+	if (passthrough)
+	{
 		[super keyDown:event];
 	}
 	else if (viewToFocus != nil)
 	{
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
+	}
+	else
+	{
+		// There are no views to pass focus to and we don't want to call `[super keyDown:event]`.
+		// Explicitly no-op.
 	}
 }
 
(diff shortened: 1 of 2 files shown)


chiuam (approved):
+1 on "Followup: Enabling more of the macOS E2E tests would have caught this. So.. we should totally do that."

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -465,15 +465,19 @@ - (void)keyDown:(NSEvent *)event
...
 	{
 		[[self window] makeFirstResponder:viewToFocus];
 		[viewToFocus scrollRectToVisible:[viewToFocus bounds]];
 	}
-	// Call super only if there are views to focus
-	else if (viewToFocus != nil)
+	else
 	{
-		[super keyDown:event];
+		// There are no views to pass focus to and we don't want to call `[super keyDown:event]`.
```
harrieshin, line 479: to me empty else block doesn't make things look clean and concise.

Saadnajmi: I can just delete the else altogether. I agree that looks better but I want it to be explicit that there are _some_ cases where we neither handle the event or call super. I'll take a better idea of refactoring too if anyone has one.

Saadnajmi: Found a way to keep the comment, remove the empty block, and still fix regression.

harrieshin (approved):

PR #2332, opened 2022-11-11
Add corner radius token test page

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Add test page showcasing the global corner radius tokens.
Added e2e test that opens this test page.

### Verification

https://user-images.githubusercontent.com/78454019/201428061-aa31e3cb-c890-461f-b72d-af1aa8e0d4f7.mp4

https://user-images.githubusercontent.com/78454019/201428066-16cb413e-c7f1-42e1-bcd9-e6eda0ec166b.mp4

https://user-images.githubusercontent.com/78454019/201428069-27cdb9dc-8730-4d7b-895f-88bc12d021f4.mov

https://user-images.githubusercontent.com/78454019/201428070-c424f897-4e6e-4bd5-bdd7-ee90cd132ce5.mov

Ran e2e test on macOS and it looks good for the corner radius token page.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/CornerRadius/CornerRadiusTest.tsx
```diff
@@ -0,0 +1,82 @@
...
+  name: string;
+  cornerRadius: number;
+}
+
+const CornerRadiusTestComponent: React.FunctionComponent<CornerRadiusTestComponentProps> = (props: CornerRadiusTestComponentProps) => {
+  const outerBoxProps = React.useMemo(
+    () => ({
+      height: 60,
+      margin: 16,
+      backgroundColor: 'lightgrey',
```
rurikoaraki, line 20: super nit: Maybe use a fluent color?

rurikoaraki:
We should probably add a simple E2E test which opens the page and makes sure it doesn't crash. @lawrencewin FYI

apps/fluent-tester/src/TestComponents/CornerRadius/CornerRadiusTest.tsx
lyzhan7: Used colors from globalTokens.color.grey and also updated to swap colors in dark mode

samuelfreiberg:
The assert you're seeing is due to the last test case  (where you're setting the corder radius to 9999px), is causing the assert. Our underlying native texture code doesn't support it, ie. why you're seeing this assert.

We can either:

1. Remove this test case (is it necessary?). If that's valid scenario, then:
2. Add a platform-specific condition to that test scenario (ie. if platform==win32 -> don't show)

samuelfreiberg:
> > The assert you're seeing is due to the last test case (where you're setting the corder radius to 9999px), is causing the assert. Our underlying native texture code doesn't support it, ie. why you're seeing this assert.
> > We can either:
> > 
> > 1. Remove this test case (is it necessary?). If that's valid scenario, then:
> > 2. Add a platform-specific condition to that test scenario (ie. if platform==win32 -> don't show)
> 
> @samuelfreiberg any idea what the right corner radius to set would be if you want a circle (like a PersonaCoin / Avatar)? Does it have to be `width / 2` which is what I see PersonaCoin do? Avatar (which is shipped on win32) uses the circular global token already, does that mean it's currently broken on win32?

Looks like this is specific to only this test page. Avatar looks fine. We'd have to schedule someone time to look into this just a FYI @rurikoaraki. If we want to push this PR through, we can make a Win32-specific check in the test page.

apps/fluent-tester/src/TestComponents/CornerRadius/CornerRadiusTest.tsx
```diff
@@ -0,0 +1,87 @@
...
+      margin: 16,
+      backgroundColor: isLightMode ? globalTokens.color.grey80 : globalTokens.color.grey10,
+      borderRadius: props.cornerRadius,
+    }),
+    [props.cornerRadius, isLightMode],
+  );
+
+  const innerCircleIndicatorProps = React.useMemo(
+    () => ({
+      height: props.cornerRadius * 2,
```
rurikoaraki, line 33: This seems like it's not going to be handled well in the circular case. Probably should cap it to something reasonable?

Saadnajmi:
> > > The assert you're seeing is due to the last test case (where you're setting the corder radius to 9999px), is causing the assert. Our underlying native texture code doesn't support it, ie. why you're seeing this assert.
> > > We can either:
> > > 
> > > 1. Remove this test case (is it necessary?). If that's valid scenario, then:
> > > 2. Add a platform-specific condition to that test scenario (ie. if platform==win32 -> don't show)
> > 
> > 
> > @samuelfreiberg any idea what the right corner radius to set would be if you want a circle (like a PersonaCoin / Avatar)? Does it have to be `width / 2` which is what I see PersonaCoin do? Avatar (which is shipped on win32) uses the circular global token already, does that mean it's currently broken on win32?
> 
> Looks like this is specific to only this test page. Avatar looks fine. We'd have to schedule someone time to look into this just a FYI @rurikoaraki. If we want to push this PR through, we can make a Win32-specific check in the test page.

Looks like I misunderstood. Setting a <View>'s corner radius to 9999 _is_ supported on win32, but that specific test had issues because it was doing other unsupported stuff (setting height to 9999 * 2). Thanks!

apps/fluent-tester/src/TestComponents/CornerRadius/CornerRadiusTest.tsx
```diff
@@ -30,8 +30,8 @@ const CornerRadiusTestComponent: React.FunctionComponent<CornerRadiusTestCompone
 
   const innerCircleIndicatorProps = React.useMemo(
     () => ({
-      height: props.cornerRadius * 2,
-      width: props.cornerRadius * 2,
+      height: props.cornerRadius === globalTokens.corner.radiusCircular ? 60 : props.cornerRadius * 2,
+      width: props.cornerRadius === globalTokens.corner.radiusCircular ? 60 : props.cornerRadius * 2,
```
Saadnajmi, line 34: Could we do like a `Math.min()` function?

lyzhan7: Sure, is there a reason you prefer that

Saadnajmi: IMO looks cleaner, and is more extensible for future token updates. Also is more general logic that works, rather than checking for a specific value.

amgleitman:
The rectangles on the dark mode iOS preview are a bit tough to see relative to the background. I don't know if 04ae705 changed this, but we should confirm that the rectangles are easy to see.

apps/fluent-tester/src/testPages.ts
```diff
@@ -93,6 +94,12 @@ export const tests: TestDescription[] = [
     testPage: HOMEPAGE_CONTEXTUALMENU_BUTTON,
     platforms: ['android', 'macos', 'win32'],
   },
+  {
+    name: 'Corner Radius Tokens',
+    component: CornerRadiusTokensTest,
+    testPage: HOMEPAGE_CORNERRADIUS_BUTTON,
+    platforms: ['android', 'ios', 'macos', 'win32'],
```
Saadnajmi, line 101: Might as well add 'windows' too?

Saadnajmi (approved):

apps/fluent-tester/src/testPages.ts
lyzhan7: Sure. Do we also want 'web'? I saw that some components have 'web' as well, not really sure what that means

lyzhan7:
> The rectangles on the dark mode iOS preview are a bit tough to see relative to the background. I don't know if [04ae705](https://github.com/microsoft/fluentui-react-native/commit/04ae705b92050e8ff89c74244b54bfa5e42f1583) changed this, but we should confirm that the rectangles are easy to see.

I'm planning to revisit the colors here when I do the color token work - have an ADO item already for it and I'll tag you in it. I'll update it to reconsider the iOS contrast. One of the issues is that the iOS background color in dark mode is different from the macOS background color in dark mode...I can find a color that will work with both but I wonder if they should actually match.

PR #2336, opened 2022-11-16
Add shadow crash + assert testing

lawrencewin:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Shadow is one of the developed experimental/v1 components which currently lacks base E2E crash + assert testing on our supported testing platforms.

This PR adds the barebones specs for E2E testing on iOS, macOS, and win32. 

### Verification

Testing via pipeline + local verification. All checks seem to pass for the above platforms. 

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/E2E/Shadow/pages/ShadowPageObject.ts
+++ b/apps/fluent-tester/src/E2E/Shadow/pages/ShadowPageObject.ts
@@ -1,0 +1,17 @@
+import { SHADOW_TESTPAGE, HOMEPAGE_SHADOW_BUTTON } from '../../../TestComponents/Shadow/consts';
+import { BasePage, By } from '../../common/BasePage';
+
+class ShadowTestPage extends BasePage {
+  /*****************************************/
+  /**************** Getters ****************/
+  /*****************************************/
+  get _testPage() {
+    return By(SHADOW_TESTPAGE);
+  }
+
+  get _pageButton() {
+    return By(HOMEPAGE_SHADOW_BUTTON);
+  }
+}
+
+export default new ShadowTestPage();
--- a/apps/fluent-tester/src/E2E/Shadow/specs/Shadow.spec.macos.ts
+++ b/apps/fluent-tester/src/E2E/Shadow/specs/Shadow.spec.macos.ts
@@ -17,6 +17,5 @@ describe('Shadow Testing Initialization', function () {
     await ShadowPageObject.waitForPageDisplayed(PAGE_TIMEOUT);
 
     await expect(await ShadowPageObject.isPageLoaded()).toBeTruthy(ShadowPageObject.ERRORMESSAGE_PAGELOAD);
-    await expect(await ShadowPageObject.didAssertPopup()).toBeFalsy(ShadowPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped up
   });
 });
--- a/apps/fluent-tester/src/E2E/Shadow/specs/Shadow.spec.macos.ts
+++ b/apps/fluent-tester/src/E2E/Shadow/specs/Shadow.spec.macos.ts
@@ -1,0 +1,22 @@
+import NavigateAppPage from '../../common/NavigateAppPage';
+import ShadowPageObject from '../pages/ShadowPageObject';
+import { PAGE_TIMEOUT, BOOT_APP_TIMEOUT } from '../../common/consts';
+
+// Before testing begins, allow up to 60 seconds for app to open
+describe('Shadow Testing Initialization', function () {
+  it('Wait for app load', async () => {
+    await NavigateAppPage.waitForPageDisplayed(BOOT_APP_TIMEOUT);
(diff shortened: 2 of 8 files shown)


lyzhan7:
Thanks for adding this! Let me know if you need any help with this on the apple side

apps/fluent-tester/src/E2E/Shadow/pages/ShadowPageObject.ts
```diff
@@ -0,0 +1,17 @@
+import { SHADOW_TESTPAGE, HOMEPAGE_SHADOW_BUTTON } from '../../../TestComponents/Shadow/consts';
+import { BasePage, By } from '../../common/BasePage';
+
+class ShadowTestPage extends BasePage {
+  /*****************************************/
+  /**************** Getters ****************/
+  /*****************************************/
+  get _testPage() {
```
samuelfreiberg, line 8: I believe there should be another constant here? `_pageName`.

This will cause error's in the error messages

lawrencewin:
> Thanks for adding this! Let me know if you need any help with this on the apple side

@lyzhan7 could you run this on iOS / macOS to see why the shadow E2E test is failing? I don't think I've seen this error before.

samuelfreiberg:
> > Thanks for adding this! Let me know if you need any help with this on the apple side
> 
> @lyzhan7 could you run this on iOS / macOS to see why the shadow E2E test is failing? I don't think I've seen this error before.

macOS and iOS don't have support for `browser.getWindowHandles()`. The Pipeline information says it's an unsupported endpoint. No other macOS or iOS specs have the assert for `didAssertPopup()`. 

I would remove it from both. Also worth filing a task for the macOS/iOS team to do an investigation on other strategies to detect an assert when it happens. Can you create that Larry? And then maybe assign to @Saadnajmi so he can put it on their backlog and triage?

samuelfreiberg:
It's also worth noting that if an assert does pop up, the tests will eventually fail on teardown because of the extra window (at least, from my testing a while ago, they did). But they won't be able to determine which specific test failed

apps/fluent-tester/src/E2E/Shadow/pages/ShadowPageObject.ts
lawrencewin: Turns out the _pageName wasn't the issue, but rather that I was calling the wrong page object to go to the shadow test page.

PPatBoyd (approved):

apps/fluent-tester/src/TestComponents/Shadow/ShadowTest.tsx
```diff
@@ -13,12 +13,10 @@ const shadowSections: TestSection[] = [
   },
   {
     name: 'Shadows on Button Examples',
-    testID: SHADOW_TESTPAGE,
```
lyzhan7, line 16: Was curious why we're removing this?

lyzhan7 (approved):

apps/fluent-tester/src/TestComponents/Shadow/ShadowTest.tsx
lawrencewin: There should only be one instance of this testID per page, which is usually assigned to the first test section. Having multiple testIDs assigned to multiple components is a no-no I believe.

samuelfreiberg: This testID is just a way to determine if we're on the right page or not. It's only necessary to have it one the first one (top of the page), but it wouldn't cause any ill-effect if it's on multiple. Just as long as it's on the first one.

apps/fluent-tester/src/E2E/Shadow/specs/Shadow.spec.macos.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Before testing begins, allow up to 60 seconds for app to open
+describe('Shadow Testing Initialization', function () {
+  it('Wait for app load', async () => {
+    await NavigateAppPage.waitForPageDisplayed(BOOT_APP_TIMEOUT);
+    await expect(await NavigateAppPage.isPageLoaded()).toBeTruthy(NavigateAppPage.ERRORMESSAGE_APPLOAD);
+  });
+
+  it('Click and navigate to Shadow test page', async () => {
+    await ShadowPageObject.waitForButtonDisplayed();
```
samuelfreiberg, line 13: Do we need this check? We're not performing an UI beforehand, so is it necessary>

lawrencewin: I think this is a leftover from other mac specs. Maybe a separate PR to remove these?

samuelfreiberg: Yes, I'd recommend making a new PR and cleaning this up

PR #2337, opened 2022-11-16
Tokenize button Android/iOS

ayush547:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

- Added paddingStart and paddingEnd to LayoutTokens.
- Completed FAB component support for text styling and appearances.
- Expanded Button appearance options to support 'Accent' and 'Outline' as required by mobile endpoints.
   Button existed in FURN with 3 possible “appearance” values -
   > Default (No appearance value passed as prop) – 
   > image
   > Primary – 
   > image
   > Subtle –
   > image

   >While the designer guidelines for Button appearance on mobile is as follows –
   > Accent –
   > image
   > Outline –
   > image
   > Subtle –
   > image

  Since props are shared across platform, consumers could pass appearances that are not implemented for the specific platforms, to get around this the following mapping is done based on similarity between the options -
   > Accent -> Primary on All platforms
   > Default (No appearance passed) -> Primary on Mobile
   > Outline -> Default on Non-Mobile platforms
(Colors in the snapshots for the buttons were updated because of default mapping to primary now)

- Updated FAB 'icon' prop to a required prop. Jest snapshot tests were accordingly updated.
- 'iconOnly' prop was added to all required samples to apply intended styling.
- Added Android token values for Button and FAB.
- Updated spec files for Button and FAB.

#### iOS Changes -

- Appearance Mapping & FAB completion carries over to iOS as well
- Tester App changes and Spec updates is left for when token value are updated for iOS

### Verification
| Dark | Light |
| ------ | ------ |
| MicrosoftTeams-image (5) | MicrosoftTeams-image (8)|
| MicrosoftTeams-image (6) | MicrosoftTeams-image (7) |

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
This is a large PR. In the future, to make this easier to review and to minimize disruption if a change needs to be backed out, I suggest keeping your changes smaller. For example, you could split the Button and FAB changes into separate PRs, or the support for inner borders as its own PR.

packages/components/Button/src/Button.types.ts
```diff
@@ -9,9 +9,10 @@ import { IViewProps } from '@fluentui-react-native/adapters';
 
 export const buttonName = 'Button';
 export type ButtonSize = 'small' | 'medium' | 'large';
-export type ButtonAppearance = 'primary' | 'subtle';
+export type ButtonAppearance = 'primary' | 'subtle' | 'ascent' | 'outline';
```
rurikoaraki, line 12: >'ascent' [](http://example.com/codeflow?start=54&length=8)

nit: typo? Is this supposed to be 'accent'?

apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -16,9 +23,15 @@ export const ShadowButtonTestSection: React.FunctionComponent = () => {
   if (Platform.OS === 'ios') {
     return (
       <View style={shadowTestPageStyles(t).backgroundColor}>
-        <FAB style={commonTestStyles.vmargin}>FAB with default shadow</FAB>
-        <CustomFABShadow64 style={commonTestStyles.vmargin}>Custom FAB with shadow64</CustomFABShadow64>
-        <CustomFABNoShadow style={commonTestStyles.vmargin}>Custom FAB with no shadow</CustomFABNoShadow>
+        <FAB icon={{ svgSource: svgProps }} style={commonTestStyles.vmargin}>
```
rurikoaraki, line 26: >{ svgSource: svgProps } [](http://example.com/codeflow?start=19&length=23)

Please make this into a const object outside of the function component so it doesn't generate a new object on each render

rurikoaraki: Also why is this change necessary?

packages/components/Button/src/Button.types.ts
rurikoaraki, line 12: SPEC should be updated for this

rurikoaraki: Also the addition of outline to ButtonCoreTokens and the updated comment to appearance

packages/components/Button/src/FAB/FAB.types.ts
```diff
@@ -3,20 +3,30 @@ import { ShadowProps } from '@fluentui-react-native/experimental-shadow';
...
 
 export interface FABProps extends ButtonCoreProps {
+  /**
+   * A button can have its content and borders styled for greater emphasis or to be subtle.
+   * - 'primary' or 'ascent': Emphasizes the button as a primary action.
+   *   'ascent' is mobile naming convention, 'primary' included here to maintain parity with Button.
+   * - 'subtle': Minimizes emphasis to blend into the background until hovered or focused.
+   * @default 'primary' (or 'ascent')
+   */
+  appearance?: FABAppearance;
```
rurikoaraki, line 21: Please update the SPEC with this comment

packages/components/Button/src/Button.styling.ts
```diff
@@ -95,6 +96,22 @@ export const getDefaultSize = (): ButtonSize => {
   return 'medium';
 };
 
+export const getPlatformSpecificAppearance = (appearance: ButtonAppearance): ButtonAppearance => {
+  switch (appearance) {
+    case 'ascent': // Included to cover Mobile platform naming guidelines, maps to 'primary'.
+      return 'primary';
+    case 'outline': // Exists only for Mobile platforms, maps to default on other platforms.
+      if (Platform.OS == 'android' || Platform.OS == 'ios') return 'outline';
```
rurikoaraki, line 106: > return 'outline'; [](http://example.com/codeflow?start=59&length=18)

styling nit: For consistency and readability, all conditional bodies should be wrapped in curly braces and bodies should be on their own line.

packages/components/Button/src/ButtonTokens.android.ts
```diff
@@ -3,76 +3,86 @@ import { globalTokens } from '@fluentui-react-native/theme-tokens';
...
-        hasIconBefore: {
-          spacingIconContentBefore: globalTokens.spacing.sNudge,
-        },
-        focused: {
-          paddingHorizontal: globalTokens.spacing.l,
-        },
+      iconSize: 20,
+      fontSize: t.typography.variants.body2Strong.size,
+      fontFamily: t.typography.variants.body2Strong.face,
+      fontWeight: t.typography.variants.body2Strong.weight,
```
rurikoaraki, line 61: Button has a ButtonFontTokens file which these would belong to

packages/components/Button/src/FAB/FAB.test.tsx
```diff
@@ -11,32 +18,48 @@ beforeAll(() => {
 });
 
 it('Default FAB (iOS)', () => {
-  const tree = renderer.create(<FAB>Default FAB (iOS)</FAB>).toJSON();
+  const tree = renderer.create(<FAB icon={{ svgSource: svgProps }}>Default FAB (iOS)</FAB>).toJSON();
```
rurikoaraki: >icon={{ svgSource: svgProps }} [](http://example.com/codeflow?start=36&length=30)

Why are we adding icons to all the test cases? Is it needed for something?

packages/components/Button/src/FAB/FAB.test.tsx
```diff
@@ -2,6 +2,13 @@ import * as React from 'react';
 import { FAB } from './FAB';
 import { checkRenderConsistency, checkReRender } from '@fluentui-react-native/test-tools';
 import * as renderer from 'react-test-renderer';
+import { SvgIconProps } from '@fluentui-react-native/icon';
+
+const TestSvg = require('../../../../../apps/fluent-tester/src/FluentTester/test-data/test.svg');
```
rurikoaraki: Mm doesn't seem great to depend on a file in a package that depends on this one.

packages/components/Button/src/FAB/FAB.types.ts
```diff
@@ -1,22 +1,38 @@
...
+   * A FAB can have its content and borders styled for greater emphasis or to be subtle.
+   * - 'primary' or 'ascent': Emphasizes the button as a primary action.
+   *   'ascent' is mobile naming convention, 'primary' included here to maintain parity with Button.
+   * - 'subtle': Minimizes emphasis to blend into the background until hovered or focused.
+   * @default 'primary' (or 'ascent')
+   */
+  appearance?: FABAppearance;
+
+  /*
+   * Source URL or name of the icon to show on the Button. Overrides optionality.
```
rurikoaraki, line 24: >Overrides optionality. [](http://example.com/codeflow?start=59&length=22)

Can you clarify what this means? What optionality is being overridden?

rurikoaraki:
Curious if this change is meant to span both Android and iOS?

packages/components/Button/src/Button.types.ts
Saadnajmi: Did we finalize whether we want an accent button at all? Also, "accent" not "ascent"?

apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
ayush547: FAB must have an icon, have updated the type to make it a required prop and hence had to add it here as well.

packages/components/Button/src/ButtonTokens.android.ts
ayush547: Added

packages/components/Button/src/FAB/FAB.types.ts
ayush547: 'icon' exists in ButtonCoreProps as an optional prop. Redeclaring it in FABProps to override its optionality type, didn't want to just have it removed from ButtonCoreProps (into ButtonProps) since it would lead to the question of why 'icon' is it not in the common props. This felt cleaner. Would you prefer we do this differently?

packages/components/Button/src/FAB/FAB.test.tsx
ayush547: FAB cannot exist without an icon, have made the prop a required one

ayush547:
> Curious if this change is meant to span both Android and iOS?
> 
> @lyzhan7 and @Saadnajmi FYI

Button and FAB for iOS and Android have similar requirements. Have made all the changes with the intent that it carries over to iOS as well

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.tsx
```diff
@@ -16,18 +16,28 @@ export const ButtonVariantTest: React.FunctionComponent = () => {
...
+      {isMobile && (
+        <>
+          <Button appearance="outline" style={commonTestStyles.vmargin}>
+            Outline
+          </Button>
+        </>
+      )}
       <Button appearance="subtle" style={commonTestStyles.vmargin}>
         Subtle
       </Button>
```
ankraj12, line 43: The components not there is fluent guidelines for android should not be rendered atleast for tester. IMO this is confusing for usage - my opinion on such code would be to not show in tester for an endpoint, as well as mentioned in SPEC. My thought is it's not an approved Fluent component in android.

ankraj12 (commented):
I would suggest following Fluent android guidelines in tester for android. We can find a way to best handle what's not supported, but from what works it should be as what is recommended for android.

packages/components/Button/src/CompoundButton/__snapshots__/CompoundButton.test.tsx.snap
```diff
@@ -48,7 +48,7 @@ exports[`CompoundButton default 1`] = `
       numberOfLines={0}
       style={
         Object {
-          "color": "#323130",
+          "color": "#ffffff",
```
rurikoaraki, line 51: Hm, this change doesn't seem correct

packages/components/Button/src/__snapshots__/Button.test.tsx.snap
```diff
@@ -41,7 +41,7 @@ exports[`Button component tests Button circular 1`] = `
     numberOfLines={0}
     style={
       Object {
-        "color": "#323130",
+        "color": "#ffffff",
```
rurikoaraki, line 44: Same comment here, this doesn't seem like a good change

packages/components/Button/src/Button.styling.ts
```diff
@@ -98,17 +100,26 @@ export const getDefaultSize = (): ButtonSize => {
...
 export const getPlatformSpecificAppearance = (appearance: ButtonAppearance): ButtonAppearance => {
   switch (appearance) {
-    case 'ascent': // Included to cover Mobile platform naming guidelines, maps to 'primary'.
+    case 'accent': // Included to cover Mobile platform naming guidelines, maps to 'primary'.
       return 'primary';
+
     case 'outline': // Exists only for Mobile platforms, maps to default on other platforms.
-      if (Platform.OS == 'android' || Platform.OS == 'ios') return 'outline';
-      else return null;
+      if (Platform.OS == 'android' || Platform.OS == 'ios') {
```
rurikoaraki, line 107: nit: Could you pull out this platform check into a meaningful variable like "doesSupportOutline" or something like that? It makes it easier to understand the intent of the platform check, esp if something should be changed in the future like adding another platform to the check.

Same comment for the default case below

packages/components/Button/src/CompoundButton/__snapshots__/CompoundButton.test.tsx.snap
ayush547: This is intended, button does not have a default view on Android and iOS and this PR maps default to 'primary', this change is coming as a result of that.

rurikoaraki: Ah okay makes sense. Thanks.

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
```diff
@@ -66,35 +59,39 @@ export const ButtonSizeTest: React.FunctionComponent = () => {
...
-        Compound Button
-      </CompoundButton>
-      <CompoundButton secondaryContent="Medium compound button" size="medium" style={commonTestStyles.vmargin}>
-        Compound Button
-      </CompoundButton>
-      <CompoundButton secondaryContent="Large compound button" size="large" style={commonTestStyles.vmargin}>
-        Compound Button
-      </CompoundButton>
-      {svgIconsEnabled && (
+      {Platform.OS != 'android' && (
```
rurikoaraki, line 62: nit: !==

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.android.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { View } from 'react-native';
+import { commonTestStyles, testContentRootViewStyle } from '../Common/styles';
+import { iconProps } from '../Common/iconExamples';
+
+export const ButtonVariantTest: React.FunctionComponent = () => {
+  const [showFABText, setShowFABText] = React.useState(true);
+  const flipFABcontent = React.useCallback(() => setShowFABText(!showFABText), [showFABText]);
+
+  return (
+    <View style={testContentRootViewStyle}>
```
rurikoaraki, line 12: This will mean we lose coverage of FAB on iOS. I would prefer to have a ButtonVariantTestSection.mobile.tsx which android and iOS then re-export.

@saadnajmi @lyzhan7 FYI

packages/components/Button/SPEC.md
```diff
@@ -70,14 +73,26 @@ The `Button` component can include an `icon` that appears before or after its `c
 
 The `Button` component supports different sizing with at least three different sizes: `small`, `medium`, and `large`. `Small` is the default on `win32`, `medium` is the default on other platforms.
 
+#### Size modification is not explicitly supported on Android and is provided as an 'as-is' functionality.
```
rurikoaraki, line 76: I think instead of this being a header it would make more sense to bold this text. In markdown I think that's wrapping it in two asterisks (**)

See https://www.markdownguide.org/cheat-sheet

You could also then remove the `</br>`s

rurikoaraki: Same comment for the other headers you've added

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.android.tsx
Saadnajmi: Generally agree with that sentiment. Mobile specific button variants (FAB) are one of the few cases we _do_ want to not render something that otherwise works on desktop, so moving to platform specific file makes sense. And following `.mobile` and re-export is what I personally like most.

lyzhan7 (commented):
Can we update the PR description "Platforms Impacted" to indicate that this affects iOS as well? And the title as well. Also looks like the PR still refers to "ascent" instead of "accent"

I can help test this on iOS and can send screenshots of what your PR looks like on iOS. Some of the changes seem android specific (i.e. a lot of new .android files) - it might be helpful to have a list of the expected changes for iOS, since it seems like iOS had a different starting point from Android

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
ayush547: Updated.

packages/components/Button/SPEC.md
ayush547: Alright

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.android.tsx
ayush547: While all the changes made (FAB working, mapping of appearances for button/FAB) were made for iOS as well, I have _not_ touched the TesterApp or the Readme for iOS, I feel the final validation for the component (with the appropriate token values) as well as the tester app's look would be better handled by whoever picks up this component for iOS. 
The change from .android to .mobile would be better after that exercise.

ayush547:
> Can we update the PR description "Platforms Impacted" to indicate that this affects iOS as well? And the title as well. Also looks like the PR still refers to "ascent" instead of "accent"
> 
> I can help test this on iOS and can send screenshots of what your PR looks like on iOS. Some of the changes seem android specific (i.e. a lot of new .android files) - it might be helpful to have a list of the expected changes for iOS, since it seems like iOS had a different starting point from Android

Have updated the PR description.
Have not touched the tester app/specs for iOS, though all the code present in the .android files can be used directly for tester app on iOS without any changes.

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.android.tsx
rurikoaraki: Generally I agree with that, I think in this case I'd try to leave iOS coverage of FAB since we do have it already. But if the iOS folks are ok with that then that's fine with me.

ayush547: FAB is still being rendered as a part of the ButtonSizeTestSection for iOS, so if _something_ breaks FAB for iOS, we are still in a good position to catch it

ankraj12 (changes requested):
Feedback over mappings.

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
```diff
@@ -42,6 +42,12 @@ export const ButtonSizeTest: React.FunctionComponent = () => {
           <Button size="large" icon={iconProps} style={commonTestStyles.vmargin}>
             Large Button with icon
           </Button>
```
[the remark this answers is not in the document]
ayush547: Yes they are!

packages/components/Button/src/Button.styling.ts
```diff
@@ -95,6 +98,37 @@ export const getDefaultSize = (): ButtonSize => {
...
+  // Mobile platforms do not have seperate styling when no appearance is passed.
+  const hasDifferentDefaultAppearance = !(Platform.OS === 'android' || Platform.OS === 'ios');
+
+  switch (appearance) {
+    case 'accent': // Included to cover Mobile platform naming guidelines, maps to 'primary'.
+      return 'primary';
+
+    case 'outline':
+      if (supportsOutlineAppearance) {
+        return 'outline';
```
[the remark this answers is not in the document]
ayush547: Even if outline is passed in appearance w/o this code segment, it would still be rendered as default. Had added this line for uniformity in covering all the appearance options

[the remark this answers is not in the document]
ayush547: Replaced explicit handling to remove the notion of mapping. Code handling responsibility is on platform. Passing outline will still lead to default styles being picked due to lack of outline specific Tokens for unsupported platforms

ankraj12 (approved):

PR #2344, opened 2022-11-17
Update token documentation: shadow, corner radius, stroke width, size

lyzhan7:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The previous documentation seemed to be written when the only tokens available were color tokens, updated the documentation to be more general. Added links to where the tokens are defined, also added examples of using all the different types of global tokens.

Note: the documentation about "size" is currently not relevant since as of publishing this PR FURN is still using the old spacing ramp. After the new size ramp is added to FURN the documentation should match.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/docs/pages/Theming/Tokens/Basics.md
+++ b/docs/pages/Theming/Tokens/Basics.md
@@ -37,27 +37,7 @@ const colorTableFluent: { [P in PersonaCoinFluentColor]: string } = {
 
 ### Font
 
-You can find tokens related to fonts such as weight and size in the `globalTokens.font` property.
-
-An example of usage is in our Badge component, where we specify different font sizes for different Badge sizes.
-
-```ts
-import { globalTokens } from '@fluentui-react-native/theme-tokens';
-
-export const badgeFontTokens: TokenSettings<BadgeTokens, Theme> = (t: Theme) =>
-  ({
-    ...
-    fontSize: globalTokens.font.size100,
-    fontWeight: globalTokens.font.weight.regular,
-    large: {
-      fontSize: globalTokens.font.size200,
-    },
-    extraLarge: {
-      fontSize: globalTokens.font.size200,
-    },
-  } as BadgeTokens);
-
-```
+TO DO
 
 ### Stroke
 
--- a/docs/pages/Theming/Tokens/Basics.md
+++ b/docs/pages/Theming/Tokens/Basics.md
@@ -16,7 +16,22 @@ Global tokens can be imported directly:
 
 The full set of global tokens is here: [android](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.android.json) / [ios](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.ios.json) / [macOS](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.macos.json) / [win32](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.win32.json) / [windows](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.windows.json)
 
-### Color
+## Alias
+
+Alias tokens are used to refer to the global tokens that are appropriate based on the current platform or theme. For example for `neutralForeground1`, in a light mode it points to a global token whose value is a light grey, but in a dark mode will change to point to a global token whose value is a dark grey. This allows component builders to point to the same color name and have it change to the correct value without having to know about the underlying theme.
+
+Alias tokens can be accessed from the `Theme` object. So if you have a `Theme` object and want to access the `neutralForeground1` alias token:
+
+```ts
+import { useFluentTheme } from '@fluentui-react-native/framework';
+
+const theme = useFluentTheme();
+const foreground = theme.colors.neutralForeground1;
+```
+
+## Types of tokens
+
+### Color [in progress]
 
 If accessing a specific color, you can find it in the `globalTokens.color` property.
 
@@ -32,26 +47,6 @@ const colorTableFluent: { [P in PersonaCoinFluentColor]: string } = {
   teal: globalTokens.color.teal.primary,
   forest: globalTokens.color.forest.primary,
   ...
-};
-```
-
-### Font
-
-TO DO
-
-### Stroke
-
-You can find tokens related to stroke in the `globalTokens.stroke` property. Currently the only type of token related to stroke is stroke width.
-
-An example of usage is in our Avatar component, where we specify different ring thicknesses for different avatar sizes.
-
-```ts
-import { globalTokens } from '@fluentui-react-native/theme-tokens';
-
-const strokeSize = {
-  small: globalTokens.stroke.width20,
-  medium: globalTokens.stroke.width30,
-  large: globalTokens.stroke.width40,
 };
 ```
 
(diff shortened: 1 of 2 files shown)


docs/pages/Theming/Tokens/Basics.md
```diff
@@ -1,19 +1,23 @@
...
+
+These names and values do not change no matter the platform or theme; however, there are some exceptions. The only current exception is that brand color definitions change depending on platform.
 
 In most cases you will not be referring to global tokens directly. The case where you may be referring to global tokens is when you're defining your own alias token, or need to access brand colors.
 
 Global tokens can be imported directly:
 
 `import { globalTokens } from '@fluentui-react-native\theme-tokens'`
 
+The full set of global tokens is here: [android](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.android.json) / [ios](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.ios.json) / [macOS](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.macos.json) / [win32](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.win32.json) / [windows](https://github.com/microsoft/fluentui-design-tokens/blob/main/src/global.windows.json)
```
ankraj12, line 17: Is it okay to have links to design-tokens repo in docs ( given it's a private repo)?

docs/pages/Theming/Tokens/Basics.md
```diff
@@ -31,13 +35,96 @@ const colorTableFluent: { [P in PersonaCoinFluentColor]: string } = {
 };
 ```
 
-\* An exception is that brand color definitions change depending on platform.
+### Font
+
+You can find tokens related to fonts such as weight and size in the `globalTokens.font` property.
+
+An example of usage is in our Badge component, where we specify different font sizes for different Badge sizes.
+
```
ankraj12, line 43: I am not sure if this is how we want the font tokens to be accessed by partners, or would the recommendation be to use the typography tokens like caption1, caption2 (using theme ) to style controls if needed..? This way of accessing globals would not reflect any changes done by theme customisation, whereas the theme tokens access would - so I was recommending the Android crew to use the tokens from theme typography tokens like     `fontSize: t.typography.sizes.body,`. Most designs also do not directly use the font tokens from global tokens, instead use from alias tokens for styling controls.
I am wondering if there is a single recommendation on this? Badge usage for tokens this way is an outlier, or recommended? If it's an outlier, this section for tokens could better fit under section of alias tokens below?

ankraj12: This query is also valid for any other global tokens which have alias layer - do we want to recommend usage directly from global tokens, or alias? I think it as later to support custom themes, but still learning :)

docs/pages/Theming/Tokens/Basics.md
lyzhan7: I'll check with PM on this - linking directly to the tokens repo was a suggestion as something temporary to link to while more official token documentation is being written

docs/pages/Theming/Tokens/Basics.md
lyzhan7: I added this just based off how I know stroke width/corner radius tokens get used, but sounds like font/typography is a little more complicated. Thinking I'll remove this section altogether and create a work item for @amgleitman to add documentation for font/typography

lyzhan7: Removed the documentation I had added on font; made an issue on ADO assigned to @amgleitman

lyzhan7: Also to address whether we want to generally recommend usage directly from global tokens vs. alias, there was this line in the documentation already:

"In most cases you will not be referring to global tokens directly. The case where you may be referring to global tokens is when you're defining your own alias token, or need to access brand colors."

But I'll also bring this up offline to see what others think

docs/pages/Theming/Tokens/Basics.md
lyzhan7: Asked PM, sounds like we still want to keep this. I also had to add the design-tokens URL to the list of ignored links in the markdown-link-check-config to get the PR checks to pass with this

docs/pages/Theming/Tokens/Basics.md
lyzhan7: Was resolved offline - I've reorganized the page so that the different types of tokens are no longer grouped in either global or alias tokens, and for each type of token we can provide whatever recommended usage we want.

docs/pages/Theming/Tokens/Basics.md
```diff
@@ -1,19 +1,38 @@
...
-This page describes how to access these colors.
+With the Fluent Design System, we have a set of named design values, that we have agreed to use for our UI. These values include colors, shadows, sizing, typography, stroke width, corner radius, etc. The [fluentui-design-tokens](https://github.com/microsoft/fluentui-design-tokens) repo contains all the token definitions. This page describes how to access these values in the fluentui-react-native repo.
 
 ## Global
 
-Our set of global tokens is the complete set of the colors that we agree to use for our UI. Each color has a specific name that is assigned to a hex value. For example, `grey74` is `#bdbdbd`.
-These names and values do not change no matter the platform or theme.\*
+Our set of global tokens is the complete set of the design values that we agree to use for our UI. Each token has a specific name that is assigned to a certain value. For example, `grey74` is `#bdbdbd`.
+
+These names and values do not change no matter the platform or theme; however, there are some exceptions. The only current exception is that brand color definitions change depending on platform.
```
rurikoaraki, line 9: "however, there are some exceptions"

nit: I think you could either remove this or reword it to be "with one exception" since we only have one exception right now.

docs/pages/Theming/Tokens/Basics.md
```diff
@@ -1,19 +1,38 @@
...
+
+const theme = useFluentTheme();
+const foreground = theme.colors.neutralForeground1;
+```
+
+## Types of tokens
+
+### Color [in progress]
+
 If accessing a specific color, you can find it in the `globalTokens.color` property.
```
rurikoaraki, line 36: I think we should change this to err towards alias tokens, since that'll be the primary usage of colors. We can push this down into "special case" or something

docs/pages/Theming/Tokens/Basics.md
```diff
@@ -31,17 +50,83 @@ const colorTableFluent: { [P in PersonaCoinFluentColor]: string } = {
...
+```
+
+### Font [in progress]
 
-Alias tokens can be accessed from the `Theme` object. Alias tokens are defined under the `colors` property on a theme. So if you have a `Theme` object and want to access the `neutralForeground1` alias token:
+TO DO
+
+### Size
+
+You can find tokens related to size in the `globalTokens.size` property.
```
rurikoaraki, line 85: nit: "related to size and spacing" instead of just "related to size", maybe?

rurikoaraki:
Thanks for this write up! Loving the examples and the detail.

docs/pages/Theming/Tokens/Basics.md
lyzhan7: We have a separate work item for updating color documentation, it's currently assigned to Ankita and I just tagged you in it. I also put a note in the documentation about the global token usage being a special case

rurikoaraki (approved):

PR #2346, opened 2022-11-18
[Checkbox] Android platform specification changes

rohanprasadofficial:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [X] android

### Description of changes

CheckboxToken.android.tsx - Removal of Label as it's not part of the Android Specification. 
Checkbox.tsx - Conditional rendering and removal support for Android. 

### Verification

| image | MicrosoftTeams-image (6) | 
| image | MicrosoftTeams-image (5)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rohanprasadofficial:
@microsoft/furn-team  , Addition and removal for tester app will be once @Saadnajmi changes/proposal are in for tackling how we can separate platform specific test examples. 

Spec.md will be once our discussion regarding Platform specific overrides is completed. We need to structure the spec.md file in better way , as we can use it generate the doc page directly from md files.

Saadnajmi (dismissed):
Without these props, the Checkbox looks very unusable? It almost feels like it'd be better to just not have the Checkbox render on Android at all, similar to FAB on Desktop? I feel like we'd need to take this back to design...

rohanprasadofficial:
@ankraj12  , Added back label till we discuss the right approach for the platform differences. 
MicrosoftTeams-image (8)
MicrosoftTeams-image (7)

packages/components/Checkbox/src/Checkbox.tsx
```diff
@@ -19,6 +21,14 @@ export const Checkbox = compose<CheckboxType>({
     required: Text,
   },
   useRender: (userProps: CheckboxProps, useSlots: UseSlots<CheckboxType>) => {
+    // Removal of tokens which are not supported for Android
+    {
+      Platform.OS === 'android' &&
```
ankraj12, line 26: Adding and label and removal of rest of props. which are makes deprecated for android seem okay for unblocking the PR while the right engineering handling for unsupported properties come to an agreement state ( which I think could incline towards not removed). Could you log a tracking item to re-look into this ( both for checkbox from design, as well as engineering code for unimplemented platforms?).

ankraj12:
> @ankraj12 , Added back label till we discuss the right approach for the platform differences.
> MicrosoftTeams-image (8)
> MicrosoftTeams-image (7)

Sounds good for unblocking

> @microsoft/furn-team , Addition and removal for tester app will be once @Saadnajmi changes/proposal are in for tackling how we can separate platform specific test examples.
> 
> Spec.md will be once our discussion regarding Platform specific overrides is completed. We need to structure the spec.md file in better way , as we can use it generate the doc page directly from md files.



> Without these props, the Checkbox looks very unusable? It almost feels like it'd be better to just not have the Checkbox render on Android at all, similar to FAB on Desktop? I feel like we'd need to take this back to design...

I am not fully aware what the mobile designs created the use case for a UI of checkbox with label, so yes this is something we can circle back with designs as it does not seem intuitive . But if the designs require a use case if this flow, removing label should be the right implementation I think.

packages/components/Checkbox/src/Checkbox.tsx
rohanprasadofficial: Yes , sounds good.

ankraj12:
> @microsoft/furn-team , Addition and removal for tester app will be once @Saadnajmi changes/proposal are in for tackling how we can separate platform specific test examples.
> 
> Spec.md will be once our discussion regarding Platform specific overrides is completed. We need to structure the spec.md file in better way , as we can use it generate the doc page directly from md files.

Based on @ejlayne inputs, I believe we are okay with below:
- Removed from test page for android for UI not in kit.
- Marking props in SPEC deprecated.

@rohanprasadofficial - Can you check on above and confirm if you can make the changes?

ankraj12 (commented):
Tester and SPEC query.

rohanprasadofficial:
Tester App view for Android @ankraj12 
Tester App

packages/components/Checkbox/SPEC.md
```diff
@@ -50,6 +50,68 @@ Win32:
...
+##### Light Mode
+
+---
+
+<br/>
+
+![Checkbox with text on Android example](./assets/Checkbox_example_android.png)
+
+```jsx
+<Checkbox label="Unchecked checkbox (undefined)" />
```
ankraj12, line 64: why do we have (undefined) here..?

ankraj12: similar below for "uncontrolled"

ankraj12 (approved):
minor commented, rest look good :)

packages/components/Checkbox/SPEC.md
rohanprasadofficial: Undefined means default here , Win32 tester app has demonstrated in the same.  
Seems the spec example and tester example doesn't match for win32. 
I tried to match it for Android , but i can change it to default if required ?

packages/components/Checkbox/SPEC.md
```diff
@@ -255,14 +324,17 @@ A disabled `Checkbox` is non-interactive, disallowing the user to click/tap on i
 #### Hovered state
 
 A hovered `Checkbox` changes styling to communicate that the user has placed a cursor above it. On win32, this will show a checkmark in a checkbox.
+On Android , This is unsupported
```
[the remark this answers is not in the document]
rohanprasadofficial: Changed everywhere to 'Not Supported'

Saadnajmi:
In the future, I would appreciate if a "request changes" was not dismissed to merge a PR. Generally I use that option rarely to signal that I think a PR needs some significant changes and shouldn't be merged as is. 

In this case, it felt like the initial Android design spec was not aware this is a cross platform control that had a different spec on another platform. Something simple like a label being part of a Checkbox in one platform but not another is not IMO a platform difference (especially because I've never seen a checkbox without a label on Android), but a _design_ difference that can be easily remedied by talking with design. 

A few other notes:

1) It would be good if you could mention in the **PR notes** what the agreement with @ejlayne was to unblock this change. When combing through git commit history, it's much easier to look at a changes' notes instead of the comments of a specific PR. In effect, the change notes become the "official ledger" for engineering choices in a repository overtime.  Only a handful of times have I ever needed to go past that into specific comments to learn more about a change :)

2) Similarly, the screenshots updated in the PR notes would help. The current PR notes still show the old Checkbox screenshots without a label. I appreciate that the test app has been updated with assets, but I'd still have preferred the PR notes updated.

3) I have several other comments for the PR that I would have added for newer iterations or the PR, that I'll just add as post PR comments 😅

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -147,15 +160,23 @@ const TokenCheckbox: React.FunctionComponent = () => {
...
   {
     name: 'Basic Checkboxes',
     testID: EXPERIMENTAL_CHECKBOX_TESTPAGE,
-    component: BasicCheckbox,
+    component: Platform.OS === 'android' ? AndroidBasicCheckbox : BasicCheckbox,
   },
   {
     name: 'Size Checkboxes',
-    component: SizeCheckbox,
+    component: Platform.OS === 'android' ? UnSupportedOnAndroid : SizeCheckbox,
```
Saadnajmi, line 179: I'm not a fan of this approach for a few reasons:

1) Adding conditional logic for a specific platform IMO should be done sparingly. The check becomes more difficult as more platforms are added and it's harder to keep track of what's happening where

2) This isn't exclusive to Android. Other platforms like iOS and macOS may or do not support different sizes and shapes for their checkboxes and I'd like a better way to convey that than "unsupported"

packages/components/Checkbox/SPEC.md
```diff
@@ -50,6 +50,68 @@ Win32:
...
+
+##### Light Mode
+
+---
+
+<br/>
+
+![Checkbox with text on Android example](./assets/Checkbox_example_android.png)
+
+```jsx
```
Saadnajmi, line 63: While I'm happy we're adding more assets to show what this looks like, I do feel like it's not scalable for the 5 platforms we support. However, we've been pretty slow on the "get a website with all the Fluent documentation train" up and running, so I guess it's worth taking anything that improves the current state of the world :)

packages/components/Checkbox/SPEC.md
```diff
@@ -256,14 +325,20 @@ A disabled `Checkbox` is non-interactive, disallowing the user to click/tap on i
 
 A hovered `Checkbox` changes styling to communicate that the user has placed a cursor above it. On win32, this will show a checkmark in a checkbox.
 
+Note: Not Supported on Android
```
Saadnajmi, line 328: And iOS, would have preferred `unsupported on mobile`

packages/components/Checkbox/SPEC.md
```diff
@@ -256,14 +325,20 @@ A disabled `Checkbox` is non-interactive, disallowing the user to click/tap on i
...
 
 A focused `Checkbox` changes styling to communicate that the user has placed keyboard focus on it. This styling is usually the same to the one in the hovered state plus extra styling on the outline to indicate keyboard focus has been placed on the component.
 
+Note: Not Supported on Android
+
 #### Pressed state
 
 A pressed `Checkbox` changes styling to communicate that the user is currently pressing it.
 
+Note: Not Supported on Android
```
Saadnajmi, line 340: This isn't supported an Android? Why not? Wouldn't that be conveyed via the ripple?

packages/components/Checkbox/src/Checkbox.tsx
```diff
@@ -19,6 +21,14 @@ export const Checkbox = compose<CheckboxType>({
...
+        unsupportedAndroidProps.map((e) => {
+          Reflect.deleteProperty(userProps, e);
+        });
+    }
```
Saadnajmi, line 30: What is `Reflect`? Also, this differs from the approach Text took to filter out unsupported props on other platforms, which might have been re-usable

https://github.com/microsoft/fluentui-react-native/blob/037c4c54ac62d19633e03eb15d72084a568ab0f4/packages/components/text/src/Text.tsx#L109-L118

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -147,15 +160,23 @@ const TokenCheckbox: React.FunctionComponent = () => {
...
+    </Text>
+  );
+};
+
 const checkboxSections: TestSection[] = [
   {
     name: 'Basic Checkboxes',
     testID: EXPERIMENTAL_CHECKBOX_TESTPAGE,
-    component: BasicCheckbox,
+    component: Platform.OS === 'android' ? AndroidBasicCheckbox : BasicCheckbox,
```
rurikoaraki, line 175: This would be better as a forked file for this test case.

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -3,16 +3,29 @@ import { EXPERIMENTAL_CHECKBOX_TESTPAGE } from './consts';
...
 import { E2ECheckboxExperimentalTest } from './E2ECheckboxExperimentalTest';
 import { InteractionEvent } from '@fluentui-react-native/interactive-hooks';
 import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
+import { Text } from '@fluentui-react-native/text';
 
 function onChangeUncontrolled(_e: InteractionEvent, isChecked: boolean) {
   console.log(isChecked);
 }
 
+const AndroidBasicCheckbox = () => {
```
ayush547, line 17: nit: type can be added here similar to 'BasicCheckbox'

Also, 
Only difference between BasicCheckbox and AndroidBasicCheckbox seems to be that the last 4 checkboxes are not present on Android, this could be achieved with an inline platform check to reduce redundancy

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
```diff
@@ -147,15 +160,23 @@ const TokenCheckbox: React.FunctionComponent = () => {
   );
 };
 
+const UnSupportedOnAndroid = () => {
```
ayush547, line 163: We should finalize on a uniform way of handling the cases of unsupported props, button current is just removing off those segments.
If we decide to go with the approach used here, we should move it to someplace common between all components

packages/components/Checkbox/src/Checkbox.tsx
ayush547: I feel there is no advantage of removing the props this way,
We aren't stopping the user in any way from passing the props and even if these props were not removed, since the token files do not contain any specific tokens for the states associated with these props they will still not be applied anyway so it is already equivalent to props not being passed in the first place. ( Eg. Say size is not supported, then 'medium', 'large' would not have any token values specified for android, this would mean it would pick the default values)

packages/components/Checkbox/src/Checkbox.tsx
ayush547: Filtering of props would be useful only if they are being applied to component through code (like labelBefore/labelAfter would control where the Text component is being rendered), but if the prop is associated with a state, then there would be no advantage of removing it right? 
(Since the token files do not contain any specific tokens for the states associated with these props they will still not be applied anyway so it is already equivalent to props not being passed in the first place. ( Eg. Say size is not supported, then 'medium', 'large' would not have any token values specified for android, this would mean it would pick the default token values))

packages/components/Checkbox/src/Checkbox.tsx
ankraj12: There are multiple ways of achieving the same implementation in this case, and there seems no clear winner. I would personally incline towards an explicit code which can be read to understand what's being removed, then an inferred code - which could or could not behave as expected in future. It's easier to change a token file in future by someone else, w/o understanding that a specific use case relied on it.

ankraj12 (commented):
Post review checks on open comment, no new feedback.

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
rohanprasadofficial: yes currently we're going with Platform.select() and not rendering the unsupported sections at all.

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
rohanprasadofficial: Address testapp changes on https://github.com/microsoft/fluentui-react-native/pull/2416

apps/fluent-tester/src/TestComponents/CheckboxExperimental/CheckboxTest.tsx
rohanprasadofficial: Correct. 
Addressed on https://github.com/microsoft/fluentui-react-native/pull/2416

packages/components/Checkbox/SPEC.md
rohanprasadofficial: Correct , looks unscalable , we have come up some solution for screenshot images , generally if web platform renders there component and support editors like Codesandbox/codepen for editor there and see how the component looks like. 
But not sure if we can have that for mobile platforms, mostly i have seen screenshots for RN component libraries if they're not supported on Web otherwise, we would chose it and render component from spec.md documentation.

packages/components/Checkbox/SPEC.md
rohanprasadofficial: Addressed on https://github.com/microsoft/fluentui-react-native/pull/2416

packages/components/Checkbox/SPEC.md
rohanprasadofficial: Addressed on https://github.com/microsoft/fluentui-react-native/pull/2416

packages/components/Checkbox/src/Checkbox.tsx
rohanprasadofficial: Address on https://github.com/microsoft/fluentui-react-native/pull/2416 
As internal discussion , we will not remove props but don't show on tester app but still can be render on certain platforms.

Issue #2324, opened 2022-11-09
FocusTrapZone in win32 tester app does not trap focus going forward


lawrencewin:
### Willing to submit a PR to fix?

- [x] I am willing to submit a PR to fix

### Package version(s)

`npm: 8.0.0`, `yarn: 1.22.19`, `react-native-win32: 0.68.0`

### OS version(s)

Windows 11

### Platform

- [ ] iOS
- [ ] macOS
- [x] win32
- [ ] windows
- [ ] android

### Please provide a reproduction of the bug

Not sure if this is an issue with the config of the FocusTrapZone on the tester page or a bug with the native module.

On the main branch as of this bug report: 
1. `yarn` + `yarn build` in the root directory.
2. `cd apps/win32` + `yarn bundle` to build the tester.
3. `yarn run-win32` to start the app.
4. Navigate to the FocusTrapZone page of the app.
5. Switch focus on the view that says "Press space to render or enter to trap, f to focus" and press `Enter` to go into the trapzone.
6. Keep navigating forwards with the `Down Arrow` or `Tab` key. You should exit out of the trapzone without having to do much of anything. (Trying the same thing with `Shift + Tab` or `Up Arrow` will trap focus going backwards.

### Actual behavior

The FocusTrapZone fails to trap user focus when moving forward with `Down Arrow` or `Tab`. The linked video will show this:

https://user-images.githubusercontent.com/15683103/200965200-5674c2c7-0f32-4981-8784-c1659027bfb3.mp4

### Expected behavior

When looking at the [equivalent web component](https://developer.microsoft.com/en-us/fluentui?fabricVer=8#/controls/web/focustrapzone), forwards and backwards focus movement using `Tab` / `Shift + Tab` is both trapped.

lawrencewin:
@rurikoaraki have you encountered this? Is this a test page config issue or an issue with the module?

rurikoaraki:
@lawrencewin I think I did notice that one of the Focus Zone test pages wasn't working properly, I think it's been broken for a while. In terms of whether it's a test page issue of an issue with the component, I would see if the component works properly in a base case, and if not, debug to see what's going on. It's likely an issue with the component itself, test pages are usually pretty plain.

PR #2352, opened 2022-11-22
FURN Tester: Fix FocusTrapZone test component not trapping focus

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, the FocusTrapZone page in the win32 FURN Tester app is not working correctly as the test component fails to trap focus going forward (#2324). This bug is caused by nested FocusTrapZone behavior: functionality for children trapzones in some cases does not take precedence over the parent. The immediate question this brings is why there is a nested FocusTrapZone in the win32 app is because of issue #674 (focus not initially setting in the win32 app) and its fix #722 (adds a FocusTrapZone to the root of the win32 app). 

This commit removes the FocusTrapZone at the root of the win32 app, fixing the behavior of the FocusTrapZone page. To set focus on the app on render, I use a React ref and call `.focus()` on the first button of the TestPage list in a `useEffect` hook. 

### Verification

Manual testing was done to see if:
1. The FocusTrapZone test component traps focus correctly.
2. The win32 app initially sets focus on something.

| [Video of FocusTrapZone not doing its job](https://user-images.githubusercontent.com/15683103/200965200-5674c2c7-0f32-4981-8784-c1659027bfb3.mp4) | [Video of FocusTrapZone doing its job 👍](https://user-images.githubusercontent.com/15683103/203435376-9af86e7a-dc42-4073-b99e-5fa4deb90cd8.mp4) |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester.tsx
@@ -1,5 +1,5 @@
 import { Theme } from '@fluentui-react-native/framework';
-import { FocusTrapZone, Separator, Text } from '@fluentui/react-native';
+import { Separator, Text } from '@fluentui/react-native';
 import { ButtonV1 as Button } from '@fluentui-react-native/button';
 import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
 import * as React from 'react';
@@ -135,6 +135,14 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
   const isTestSectionVisible = !enableSinglePaneView || (enableSinglePaneView && !onTestListView);
 
   const TestList: React.FunctionComponent = () => {
+    const ref = React.useRef<View>();
+
+    React.useEffect(() => {
+      if (Platform.OS === ('win32' as any)) {
+        ref.current.focus();
+      }
+    }, []);
+
     return (
       <View style={fluentTesterStyles.testList}>
         <ScrollView contentContainerStyle={fluentTesterStyles.testListContainerStyle} testID={TESTPAGE_BUTTONS_SCROLLVIEWER}>
@@ -147,6 +155,7 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
                 onClick={() => setSelectedTestIndex(index)}
                 style={fluentTesterStyles.testListItem}
                 testID={description.testPage}
+                componentRef={index === 0 ? ref : undefined}
               >
                 {description.name}
               </Button>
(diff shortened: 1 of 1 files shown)


Saadnajmi (approved):

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -135,6 +135,14 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
...
 
   const TestList: React.FunctionComponent = () => {
+    const ref = React.useRef<View>();
+
+    React.useEffect(() => {
+      if (Platform.OS === ('win32' as any)) {
+        ref.current.focus();
+      }
+    }, []);
+
```
Saadnajmi, line 145: This would run every time TestList re-renders, rather than the whole app re-rendering. I'm not _sure_ but that may re-render more than we like. Just to confirm, clicking around different test pages doesn't cause focus to jump back to the test list?

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -147,6 +155,7 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
                 onClick={() => setSelectedTestIndex(index)}
                 style={fluentTesterStyles.testListItem}
                 testID={description.testPage}
+                componentRef={index === 0 ? ref : undefined}
```
Saadnajmi, line 158: If you prefer avoiding the word `undefined` :

` ...(index === 0 && {componentRef: ref})`

apps/fluent-tester/src/FluentTester.tsx
lawrencewin: You're correct. Moving the block out of the testlist and into the base component fixes this.

apps/fluent-tester/src/FluentTester.tsx
lawrencewin: Thanks for the syntax suggestion!

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -155,7 +156,8 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
                 onClick={() => setSelectedTestIndex(index)}
                 style={fluentTesterStyles.testListItem}
                 testID={description.testPage}
-                componentRef={index === 0 ? ref : undefined}
+                {...(index === 0 && { componentRef: win32FocusOnMount })}
+                // This ref so focus can be set on it when the app mounts in win32. Without this, focus won't be set anywhere.
```
Saadnajmi, line 160: 3 small nits:
1) Comment is usually above the line of code
2) I don't think you need the outer `{}` 
3) Prefer the name `focusOnMountRef` to `win32 focusOnMountRef` :)

lawrencewin: I'll fix nits 1 & 3. For nit 2, my linter goes crazy when removing the outer `{}` so that will stay.

samuelfreiberg (approved):

PR #2353, opened 2022-11-23
Add ScrollView to Menu 

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Let's make Menu scrollable. Currently not working on win32

### Verification
Light mode:
https://user-images.githubusercontent.com/67026167/203460193-bc1bded2-c151-44e7-ad6d-54a4b5a58a2a.mov
Dark mode:

https://user-images.githubusercontent.com/67026167/203641203-0407be33-fdcf-4325-b31e-fbc0b34daf93.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
@@ -68,6 +68,20 @@ export const MenuScrollView: React.FunctionComponent = () => {
             <MenuList>{data}</MenuList>
           </ScrollViewMenuPopover>
         </Menu>
+        <Menu>
+          <MenuTrigger>
+            <Button>Custom height</Button>
+          </MenuTrigger>
+          <MenuPopover maxHeight={100}>
+            <MenuList>
+              <MenuItem>MenuItem 1</MenuItem>
+              <MenuItem>MenuItem 2</MenuItem>
+              <MenuItem>MenuItem 3</MenuItem>
+              <MenuItem>MenuItem 4</MenuItem>
+              <MenuItem>MenuItem 5</MenuItem>
+            </MenuList>
+          </MenuPopover>
+        </Menu>
       </View>
     </Stack>
   );
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
@@ -1,0 +1,40 @@
+import React, { useState } from 'react';
+import { ButtonV1 as Button } from '@fluentui/react-native';
+import { Menu, MenuItem, MenuTrigger, MenuPopover, MenuList } from '@fluentui-react-native/menu';
+import { Stack } from '@fluentui-react-native/stack';
+import { stackStyle } from '../Common/styles';
+import { TextInput, View } from 'react-native';
+import { commonTestStyles as commonStyles } from '../Common/styles';
+
+export const MenuScrollView: React.FunctionComponent = () => {
+  const [maxHeight, setMaxHeight] = useState<number>(800);
+  const ScrollViewMenuPopover = MenuPopover.customize({ maxHeight: maxHeight });
+  const menuItems = [];
+  for (let i = 0; i < 400; i++) {
+    menuItems.push(<MenuItem key={i}>MenuItem</MenuItem>);
+  }
+
+  return (
+    <Stack style={stackStyle}>
+      <View>
+        <TextInput
+          accessibilityLabel="Max height for menu"
+          style={commonStyles.textBox}
+          placeholder="Max height for menu"
+          blurOnSubmit={true}
+          onSubmitEditing={(e) => {
+            setMaxHeight(parseInt(e.nativeEvent.text));
+          }}
+        />
+      </View>
+      <Menu>
+        <MenuTrigger>
+          <Button>Test</Button>
+        </MenuTrigger>
+        <ScrollViewMenuPopover>
+          <MenuList>{menuItems}</MenuList>
+        </ScrollViewMenuPopover>
+      </Menu>
+    </Stack>
+  );
+};
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
@@ -12,9 +12,6 @@ const styles = StyleSheet.create({
 });
 
 export const MenuScrollView: React.FunctionComponent = () => {
-  const [maxHeight, setMaxHeight] = useState<number>(800);
-  const ScrollViewMenuPopover = MenuPopover.customize({ maxHeight: maxHeight });
-  const menuItems = [];
   const [data, setData] = useState([
     <MenuItem key={1}>MenuItem 1</MenuItem>,
     <MenuItem key={2}>MenuItem 2</MenuItem>,
@@ -23,9 +20,6 @@ export const MenuScrollView: React.FunctionComponent = () => {
     <MenuItem key={5}>MenuItem 5</MenuItem>,
   ]);
 
-  for (let i = 0; i < 400; i++) {
-    menuItems.push(<MenuItem key={i}>MenuItem {i}</MenuItem>);
-  }
   const insertOnClick = React.useCallback(() => {
     data.push(<MenuItem key={Math.random()}>MenuItem</MenuItem>);
     const newList = [...data];
@@ -39,6 +33,13 @@ export const MenuScrollView: React.FunctionComponent = () => {
       setData(newList);
     }
   }, [data]);
+
+  const [maxHeight, setMaxHeight] = useState<number>(800);
+  const menuItems = [];
+
+  for (let i = 0; i < 40; i++) {
+    menuItems.push(<MenuItem key={i}>MenuItem {i}</MenuItem>);
+  }
 
   return (
     <Stack style={stackStyle}>
@@ -54,11 +55,11 @@ export const MenuScrollView: React.FunctionComponent = () => {
         />
         <Menu>
           <MenuTrigger>
-            <Button>Test</Button>
+            <Button>Custom height</Button>
           </MenuTrigger>
-          <ScrollViewMenuPopover>
+          <MenuPopover maxHeight={maxHeight}>
             <MenuList>{menuItems}</MenuList>
-          </ScrollViewMenuPopover>
+          </MenuPopover>
         </Menu>
       </View>
       <View style={styles.menuContainerRow}>
@@ -72,14 +73,14 @@ export const MenuScrollView: React.FunctionComponent = () => {
           <MenuTrigger>
             <Button>Mutable Menu</Button>
           </MenuTrigger>
-          <ScrollViewMenuPopover>
+          <MenuPopover>
             <MenuList>{data}</MenuList>
-          </ScrollViewMenuPopover>
+          </MenuPopover>
         </Menu>
       </View>
       <Menu>
         <MenuTrigger>
-          <Button>Custom height</Button>
+          <Button>Height: 100</Button>
         </MenuTrigger>
         <MenuPopover maxHeight={100}>
           <MenuList>
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
@@ -57,8 +57,12 @@ export const MenuScrollView: React.FunctionComponent = () => {
         </ScrollViewMenuPopover>
       </Menu>
       <Text>Add/Remove menu item</Text>
-      <Button onClick={insertOnClick}>Add new menu item</Button>
-      <Button onClick={popOnClick}>Remove last menu item</Button>
+      <Button appearance="subtle" onClick={insertOnClick}>
+        Add new menu item
+      </Button>
+      <Button appearance="subtle" onClick={popOnClick}>
+        Remove last menu item
+      </Button>
       <View>
         <Menu>
           <MenuTrigger>
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
@@ -1,10 +1,16 @@
 import React, { useState } from 'react';
-import { ButtonV1 as Button, Text } from '@fluentui/react-native';
+import { ButtonV1 as Button } from '@fluentui/react-native';
 import { Menu, MenuItem, MenuTrigger, MenuPopover, MenuList } from '@fluentui-react-native/menu';
 import { Stack } from '@fluentui-react-native/stack';
 import { stackStyle } from '../Common/styles';
-import { TextInput, View } from 'react-native';
+import { Text, TextInput, StyleSheet, View } from 'react-native';
 import { commonTestStyles as commonStyles } from '../Common/styles';
+
+const styles = StyleSheet.create({
+  menuContainerSubtitle: { fontWeight: 'bold' },
+  menuContainerRow: { flexDirection: 'row' },
+  menuMargin: { marginRight: 20 },
+});
 
 export const MenuScrollView: React.FunctionComponent = () => {
   const [maxHeight, setMaxHeight] = useState<number>(800);
@@ -37,33 +43,33 @@ export const MenuScrollView: React.FunctionComponent = () => {
 
   return (
     <Stack style={stackStyle}>
-      <View>
+      <View style={styles.menuContainerRow}>
         <TextInput
           accessibilityLabel="Max height for menu"
-          style={commonStyles.textBox}
+          style={[commonStyles.textBox, styles.menuMargin]}
           placeholder="Max height for menu"
           blurOnSubmit={true}
           onSubmitEditing={(e) => {
             setMaxHeight(parseInt(e.nativeEvent.text));
           }}
         />
+        <Menu>
+          <MenuTrigger>
+            <Button>Test</Button>
+          </MenuTrigger>
+          <ScrollViewMenuPopover>
+            <MenuList>{menuItems}</MenuList>
+          </ScrollViewMenuPopover>
+        </Menu>
       </View>
-      <Menu>
-        <MenuTrigger>
-          <Button>Test</Button>
-        </MenuTrigger>
-        <ScrollViewMenuPopover>
-          <MenuList>{menuItems}</MenuList>
-        </ScrollViewMenuPopover>
-      </Menu>
-      <Text>Add/Remove menu item</Text>
-      <Button appearance="subtle" onClick={insertOnClick}>
-        Add new menu item
-      </Button>
-      <Button appearance="subtle" onClick={popOnClick}>
-        Remove last menu item
-      </Button>
-      <View>
+      <Text style={styles.menuContainerSubtitle}>Add/Remove menu item</Text>
+      <View style={styles.menuContainerRow}>
+        <Button appearance="subtle" onClick={insertOnClick}>
+          Add new menu item
+        </Button>
+        <Button appearance="subtle" onClick={popOnClick}>
+          Remove last menu item
+        </Button>
         <Menu>
           <MenuTrigger>
             <Button>Mutable Menu</Button>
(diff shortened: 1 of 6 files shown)


packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -51,20 +52,24 @@ export const MenuList = compose<MenuListType>({
...
-            </Slots.focusZone>
+            <Slots.scrollView>
+              <Slots.focusZone
+                componentRef={focusZoneRef}
+                focusZoneDirection={'vertical'}
+                defaultTabbableElement={focusZoneRef}
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore FocusZone takes ViewProps, but that isn't defined on it's type.
+                enableFocusRing={false}
+                forceFocusMacOS={true}
```
Saadnajmi, line 63: I think this no longer exists so we can get rid of it? 😅

chiuam: Removed

chiuam:
> Do you have dark mode screenshots?

Updated PR description!

rurikoaraki:
Hmmm I'm testing this on win32 and stuff is getting clipped horizontally:

image

Is this not happening for you?

chiuam:
> Hmmm I'm testing this on win32 and stuff is getting clipped horizontally:
> 
> image
> 
> Is this not happening for you?

Hmm, I don't seem to be facing any text truncation issues on Mac:
image

chiuam:
@rurikoaraki Thanks for testing on win32, seems like it's causing issues so I'll make the change macOS specific.

Saadnajmi:
@PPatboyd Do you know if on win32 Callout, you can pass something like `style: { maxHeight: 800 }` and that gets respected in the size of the Callout. That works for macOS so we didn't need a maxHeight prop for ContextualMenu on macOS, but I wasn't sure for win32

Saadnajmi:
Two comments that we can save till next week:

1) The way <Menu> is set up with an inner <MenuList>, I wonder if we could actually leave it up to the client to add their own ScrollView. Another way to ask: Should all Menu's have ScrollViews? My gut feeling is: No we should do it ourselves to make sure all the behaviors / interactions / a11y are handled properly.

2) We can make ScrollView macOS only, but our downstream clients will want it for win32 as well. So we should make sure that we're set up so that win32 can also have this feature?

chiuam:
> Two comments that we can save till next week:
> 
> 1. The way  is set up with an inner , I wonder if we could actually leave it up to the client to add their own ScrollView. Another way to ask: Should all Menu's have ScrollViews? My gut feeling is: No we should do it ourselves to make sure all the behaviors / interactions / a11y are handled properly.
> 2. We can make ScrollView macOS only, but our downstream clients will want it for win32 as well. So we should make sure that we're set up so that win32 can also have this feature?

For #1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.

For #2: I think @rurikoaraki is taking charge on that, are we okay with the disparity for now or we should rollout scroll capability on macOS and win32 at the time?

rurikoaraki:
> > Two comments that we can save till next week:
> > 
> > 1. The way  is set up with an inner , I wonder if we could actually leave it up to the client to add their own ScrollView. Another way to ask: Should all Menu's have ScrollViews? My gut feeling is: No we should do it ourselves to make sure all the behaviors / interactions / a11y are handled properly.
> > 2. We can make ScrollView macOS only, but our downstream clients will want it for win32 as well. So we should make sure that we're set up so that win32 can also have this feature?
> 
> 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> 
> 2: I think @rurikoaraki is taking charge on that, are we okay with the disparity for now or we should rollout scroll capability on macOS and win32 at the time?

Let's take #2 to PM

PPatBoyd:
> @PPatBoyd Do you know if on win32 Callout, you can pass something like `style: { maxHeight: 800 }` and that gets respected in the size of the Callout. That works for macOS so we didn't need a maxHeight prop for ContextualMenu on macOS, but I wasn't sure for win32

Unsure current status, but it should be supported to fit API and user expectations

Saadnajmi:
> > @PPatBoyd Do you know if on win32 Callout, you can pass something like `style: { maxHeight: 800 }` and that gets respected in the size of the Callout. That works for macOS so we didn't need a maxHeight prop for ContextualMenu on macOS, but I wasn't sure for win32
> 
> Unsure current status, but it should be supported to fit API and user expectations -- I assume the `we didn't need a maxHeight for ContextualMenu` is `we inherit it and therefore don't need an additional prop` and not `it isn't a prop available to the component` ?

For ContextualMenu on win32, Max height is an additional prop. That didn't seem necessary to me, and I wasn't sure if that was a native difference or just an API choice.

mischreiber:
> 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.

Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?

Saadnajmi:
> > 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> 
> Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?

It _should_ be dependent on your system preferences whether there's space for the scrollbar. One of my bugs in RN-macos was making sure that there's only extra space for the scrollbar if the OS preference is set as such. I think @chiuam has her machine set to always show scrollbars, so our menu will always have extra space there. That's one way we differ from NSMenu

chiuam:
> > > 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> > 
> > 
> > Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?
> 
> It _should_ be dependent on your system preferences whether there's space for the scrollbar. One of my bugs in RN-macos was making sure that there's only extra space for the scrollbar if the OS preference is set as such. I think @chiuam has her machine set to always show scrollbars, so our menu will always have extra space there. That's one way we differ from NSMenu

+1 to what Saad said and thank you for bringing up the OS preference. I had it set to "Automatically based on mouse or trackpad" which I think means show scrollbar only when the content overflows?

harrieshin:
> > > > 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> > > 
> > > 
> > > Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?
> > 
> > 
> > It _should_ be dependent on your system preferences whether there's space for the scrollbar. One of my bugs in RN-macos was making sure that there's only extra space for the scrollbar if the OS preference is set as such. I think @chiuam has her machine set to always show scrollbars, so our menu will always have extra space there. That's one way we differ from NSMenu
> 
> +1 to what Saad said and thank you for bringing up the OS preference. I had it set to "Automatically based on mouse or trackpad" which I think means show scrollbar only when the content overflows?
> 
> Hmm, the scrollbar overlaps the menu items when the preference is set to "When scrolling" image

feels dejavu. didn't we fix this bug in contextual menu as well long time ago by making sure we add the framesize of the scrollbar?

chiuam:
> > > > > 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> > > > 
> > > > 
> > > > Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?
> > > 
> > > 
> > > It _should_ be dependent on your system preferences whether there's space for the scrollbar. One of my bugs in RN-macos was making sure that there's only extra space for the scrollbar if the OS preference is set as such. I think @chiuam has her machine set to always show scrollbars, so our menu will always have extra space there. That's one way we differ from NSMenu
> > 
> > 
> > +1 to what Saad said and thank you for bringing up the OS preference. I had it set to "Automatically based on mouse or trackpad" which I think means show scrollbar only when the content overflows?
> > Hmm, the scrollbar overlaps the menu items when the preference is set to "When scrolling" image
> 
> feels dejavu. didn't we fix this bug in contextual menu as well long time ago by making sure we add the framesize of the scrollbar?

I wonder if the fix was in Callout because that would mean the bug should be fixed in Menu too?

Saadnajmi:
The 

> > > > > > 1: We talked a bit offline, I think we settled on Menu should have a scrollview by default, due to the reasons you mentioned like a11y.
> > > > > 
> > > > > 
> > > > > Can you show what that would look like when menu items don't overflow? e.g. is the scrollbar always visible in that scenario?
> > > > 
> > > > 
> > > > It _should_ be dependent on your system preferences whether there's space for the scrollbar. One of my bugs in RN-macos was making sure that there's only extra space for the scrollbar if the OS preference is set as such. I think @chiuam has her machine set to always show scrollbars, so our menu will always have extra space there. That's one way we differ from NSMenu
> > > 
> > > 
> > > +1 to what Saad said and thank you for bringing up the OS preference. I had it set to "Automatically based on mouse or trackpad" which I think means show scrollbar only when the content overflows?
> > > Hmm, the scrollbar overlaps the menu items when the preference is set to "When scrolling" image
> > 
> > 
> > feels dejavu. didn't we fix this bug in contextual menu as well long time ago by making sure we add the framesize of the scrollbar?
> 
> I wonder if the fix was in Callout because that would mean the bug should be fixed in Menu too?

When the scrollbar is _overlay_ , it's supposed to overlap, as it's only there while scrolling. When it's set to always be there, then there is dedicated space allocated to it. I'm happy to walk through the different states of Scrollbar and expectations offline if that helps.

Saadnajmi:
A summary for each option:

1) Show Scrollbars **when scrolling**

- We get iOS style overlay scrollbars
- No extra space is allocated for the Scrolbar in React Native macOS

2) Show Scrollbars **Always**

- We get legacy style scrollbars that  take space if content overflows
- If there's no overflow content, React Native macOS doesn't allocate extra space for it
- When content overflows, React Native macOS will re-layout and allocate extra space on the scrollbar. Internally, I think this is done by adding a margin to RCTCustomScrollView's content view (AKA, done in RN-macOS, not Callout)

3) Automatically based on mouse or trackpad

- We either get option (1) or (2), based on what devices are connected to your Mac. I never took the time to see what that was since handling (1) and (2) meant you handled (3) properly.

apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
```diff
@@ -152,6 +152,29 @@ const MenuSubMenu: React.FunctionComponent = () => {
   );
 };
 
+const ScrollViewMenuPopover = MenuPopover.customize({ maxHeight: 100 });
```
Saadnajmi, line 155: Why not just pass as a prop?

packages/components/Menu/src/MenuList/MenuList.types.ts
```diff
@@ -51,6 +52,7 @@ export interface MenuListState extends Omit<MenuListProps, 'checked' | 'onChecke
 export interface MenuListSlotProps {
   root: React.PropsWithRef<IViewProps> & { gap?: number };
   focusZone?: FocusZoneProps; // macOS only
+  scrollView: ScrollViewProps; // macOS only
```
Saadnajmi, line 55: Maybe also make optional since macOS only?

chiuam:
> Test page recommendations: Make the max height an input on the test page Test dynamic menu item addition and removal; this has come up as bugs for us in various ways Test very-large menu (greater than screen dimensions) and behaviors of menus when near the display borders

1. 6037e0a356667c32b3b9039c6556238881efd04f: Added an input for maxHeight, set default to 800. 
2. c4c19ecaf89cea829d7df3acacd9930a4e71ab35: Add/Remove menu items

Thank you for the suggestions!

apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
chiuam: Leaving this one as is for the mutable test case. Added another one for custom height where I pass in `maxHeight` as a prop. see cbc482ce01464538b2ca4bd6a83e0a892dc09858

chiuam:
> A summary for each option:
> 
> 1. Show Scrollbars **when scrolling**
> 
> * We get iOS style overlay scrollbars
> * No extra space is allocated for the Scrolbar in React Native macOS
> 
> 2. Show Scrollbars **Always**
> 
> * We get legacy style scrollbars that  take space if content overflows
> * If there's no overflow content, React Native macOS doesn't allocate extra space for it
> * When content overflows, React Native macOS will re-layout and allocate extra space on the scrollbar. Internally, I think this is done by adding a margin to RCTCustomScrollView's content view (AKA, done in RN-macOS, not Callout)
> 
> 3. Automatically based on mouse or trackpad
> 
> * We either get option (1) or (2), based on what devices are connected to your Mac. I never took the time to see what that was since handling (1) and (2) meant you handled (3) properly.
> 
> @chiuam could you maybe add a video showing each of these cases? It can be one video where you just switch between the 4 options: (overlay style, legacy style) x (no overflow, overflow)

Thanks for the writeup! I updated the light mode video

apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
```diff
@@ -0,0 +1,96 @@
...
+    <MenuItem key={1}>MenuItem 1</MenuItem>,
+    <MenuItem key={2}>MenuItem 2</MenuItem>,
+    <MenuItem key={3}>MenuItem 3</MenuItem>,
+    <MenuItem key={4}>MenuItem 4</MenuItem>,
+    <MenuItem key={5}>MenuItem 5</MenuItem>,
+  ]);
+
+  for (let i = 0; i < 400; i++) {
+    menuItems.push(<MenuItem key={i}>MenuItem {i}</MenuItem>);
+  }
```
Saadnajmi, line 28: Nit: My first read through, I thought we were pushing 400 objects to `data` not `menuItem` because of the ordering of the variable declarations. I think I would prefer to keep all hook calls at the top of the function, and to have `const menuItems = []` defined right before the for loop.

apps/fluent-tester/src/TestComponents/Menu/MenuScrollView.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { commonTestStyles as commonStyles } from '../Common/styles';
+
+const styles = StyleSheet.create({
+  menuContainerRow: { flexDirection: 'row' },
+  menuMargin: { marginRight: 20 },
+});
+
+export const MenuScrollView: React.FunctionComponent = () => {
+  const [maxHeight, setMaxHeight] = useState<number>(800);
+  const ScrollViewMenuPopover = MenuPopover.customize({ maxHeight: maxHeight });
```
Saadnajmi, line 16: Defining this inside a function component means it is recreated on every render. Furthermore, the name feels misleading to me. `ScrollViewMenuPopover` sounds like a special menu popover with a scrollview. But really, the scrollview is in the menu list, and this is just setting the max height. Could we just set the Max height via a style prop on line 59?

Saadnajmi (commented):
Some nits about the new test pages:

1) the 400 menu item menu is really slow. Maybe that's just what a 400 item menu is like with RN.. but perhaps we could limit it to something smaller, like.. 40? Or perhaps we have some innefficiency where all 400 menu items are getting re-rendered every time the mouse moves instead of just 2 menu items..

2) It felt redundant to me that there was a test where you could set the Max height, and a custom height menu right next to each other. At first, I wasn't sure which menu the text box would set the height for: "Test" or "Custom height". I think I'd prefer one or the other. And if choosing one, the latter, since the first one was slow, re-rendered the popover extraneously with the bug I mentioned elsewhere, and had a slightly confusing variable name..

3) The mutable menu dismisses on every click to the Add/remove buttons. I guess that's by design, but it felt a bit weird. The only other option I thought is to make a custom menu item with a + / - button in the menu itself but that may be too much extra work for this PR. I think it may be fine to leave as is, but wanted to mention it felt a little weird while I was at it.

https://user-images.githubusercontent.com/6722175/208965744-f2b21926-4fce-4ebb-84b7-259faa29652a.mov

Saadnajmi:
Another thing I noticed: When I set scrollbars to show as always, I see that MenuV1 has extra padding for both horizontal and vertical scrollbars:

Screenshot 2022-12-21 at 11 12 32 AM

I didn't notice that in ContextualMenu (only vertical scrollbar had extra space). I think I fixed that by setting some extra props as seen here: https://github.com/microsoft/fluentui-react-native/pull/1611/

chiuam:
> Some nits about the new test pages:
> 
> 1. the 400 menu item menu is really slow. Maybe that's just what a 400 item menu is like with RN.. but perhaps we could limit it to something smaller, like.. 40? Or perhaps we have some innefficiency where all 400 menu items are getting re-rendered every time the mouse moves instead of just 2 menu items..
> 2. It felt redundant to me that there was a test where you could set the Max height, and a custom height menu right next to each other. At first, I wasn't sure which menu the text box would set the height for: "Test" or "Custom height". I think I'd prefer one or the other. And if choosing one, the latter, since the first one was slow, re-rendered the popover extraneously with the bug I mentioned elsewhere, and had a slightly confusing variable name..
> 3. The mutable menu dismisses on every click to the Add/remove buttons. I guess that's by design, but it felt a bit weird. The only other option I thought is to make a custom menu item with a + / - button in the menu itself but that may be too much extra work for this PR. I think it may be fine to leave as is, but wanted to mention it felt a little weird while I was at it.
> 
>  Screen.Recording.2022-12-21.at.11.10.02.AM.mov

1. Yeah I noticed the slowness as well. 40 sounds good, long enough to make the menu overflow. Though not sure if we want to keep performance in mind since there's demand for larger lists like > 100 menu items?
2. The test for custom height was to showcase the maxHeight prop is being passed in and respected properly. I prefer keeping both of them since they both serve different purposes. I can update the customizable max height menu title to "Custom height" and the other one to "Height: 100" to better signify each menu test case.
3. Interesting idea, can we have Button as a menu item? Though I have to agree that it be a bit far-fetched for this PR.

Saadnajmi:
> > Some nits about the new test pages:
> > 
> > 1. the 400 menu item menu is really slow. Maybe that's just what a 400 item menu is like with RN.. but perhaps we could limit it to something smaller, like.. 40? Or perhaps we have some innefficiency where all 400 menu items are getting re-rendered every time the mouse moves instead of just 2 menu items..
> > 2. It felt redundant to me that there was a test where you could set the Max height, and a custom height menu right next to each other. At first, I wasn't sure which menu the text box would set the height for: "Test" or "Custom height". I think I'd prefer one or the other. And if choosing one, the latter, since the first one was slow, re-rendered the popover extraneously with the bug I mentioned elsewhere, and had a slightly confusing variable name..
> > 3. The mutable menu dismisses on every click to the Add/remove buttons. I guess that's by design, but it felt a bit weird. The only other option I thought is to make a custom menu item with a + / - button in the menu itself but that may be too much extra work for this PR. I think it may be fine to leave as is, but wanted to mention it felt a little weird while I was at it.
> > 
> >  Screen.Recording.2022-12-21.at.11.10.02.AM.mov
> 
> 1. Yeah I noticed the slowness as well. 40 sounds good, long enough to make the menu overflow. Though not sure if we want to keep performance in mind since there's demand for larger lists like > 100 menu items?
> 2. The test for custom height was to showcase the maxHeight prop is being passed in and respected properly. I prefer keeping both of them since they both serve different purposes. I can update the customizable max height menu title to "Custom height" and the other one to "Height: 100" to better signify each menu test case.
> 3. Interesting idea, can we have Button as a menu item? Though I have to agree that it be a bit far-fetched for this PR. 

MenuV1 is written such that anything inside MenuList will have it's onClick overridden to act like a menu item. While it's possible to write a custom menu item, I agree that we shouldn't be too for this PR. Your other suggestions sound fine.

For performance, it may be worth double checking that every menu item isn't rerendering on every event. You can do that with a console.log() inside menu items' render and seeing how many fire. But also, out of scope of this PR.

Saadnajmi:
> > Some nits about the new test pages:
> > 
> > 1. the 400 menu item menu is really slow. Maybe that's just what a 400 item menu is like with RN.. but perhaps we could limit it to something smaller, like.. 40? Or perhaps we have some innefficiency where all 400 menu items are getting re-rendered every time the mouse moves instead of just 2 menu items..
> > 2. It felt redundant to me that there was a test where you could set the Max height, and a custom height menu right next to each other. At first, I wasn't sure which menu the text box would set the height for: "Test" or "Custom height". I think I'd prefer one or the other. And if choosing one, the latter, since the first one was slow, re-rendered the popover extraneously with the bug I mentioned elsewhere, and had a slightly confusing variable name..
> > 3. The mutable menu dismisses on every click to the Add/remove buttons. I guess that's by design, but it felt a bit weird. The only other option I thought is to make a custom menu item with a + / - button in the menu itself but that may be too much extra work for this PR. I think it may be fine to leave as is, but wanted to mention it felt a little weird while I was at it.
> > 
> >  Screen.Recording.2022-12-21.at.11.10.02.AM.mov
> 
> 1. Yeah I noticed the slowness as well. 40 sounds good, long enough to make the menu overflow. Though not sure if we want to keep performance in mind since there's demand for larger lists like > 100 menu items?
> 2. The test for custom height was to showcase the maxHeight prop is being passed in and respected properly. I prefer keeping both of them since they both serve different purposes. I can update the customizable max height menu title to "Custom height" and the other one to "Height: 100" to better signify each menu test case.
> 3. Interesting idea, can we have Button as a menu item? Though I have to agree that it be a bit far-fetched for this PR. 

MenuV1 is written such that anything inside MenuList will have it's onClick overridden to act like a menu item. While it's possible to write a custom menu item, I agree that we shouldn't be too for this PR. Your other suggestions sound fine.

For performance, it may be worth double checking that every menu item isn't rerendering on every event. You can do that with a console.log() inside menu items' render and seeing how many fire. But also, out of scope of this PR.

Saadnajmi (approved):

PR #2354, opened 2022-11-23
Design token package and new spacing tokens variants bump

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rohanprasadofficial:
@lyzhan7 , Not sure if you have started on this or not 🤔 , but it was blocking on my switch new tokens consumption , so i had to update it. 

We can talk on this if you have planned or work on this updates.  I will continue on my Switch for new token consumption that needed to bump design-token package from 0.31 to 0.36 that broken spacing and few other things.

apps/fluent-tester/src/TestComponents/Spacing/SpacingTest.tsx
```diff
@@ -21,21 +21,29 @@ const BasicUsage: React.FunctionComponent = () => {
...
+        {spacingExample('size280', appleTheme.spacing.size280)}
+        {spacingExample('size320', appleTheme.spacing.size320)}
+        {spacingExample('size360', appleTheme.spacing.size360)}
+        {spacingExample('size400', appleTheme.spacing.size400)}
+        {spacingExample('size480', appleTheme.spacing.size480)}
+        {spacingExample('size560', appleTheme.spacing.size560)}
+        {spacingExample('size640', appleTheme.spacing.size640)}
+        {spacingExample('size720', appleTheme.spacing.size720)}
+        {spacingExample('size800', appleTheme.spacing.size800)}
+        {spacingExample('size1200', appleTheme.spacing.size1200)}
```
lyzhan7, line 46: Let's use tokens from globalTokens instead of from the theme - i.e. globalTokens.spacing.sizeNone. We're trying to deprecate spacing on the theme and access the spacing tokens through globalTokens instead

packages/components/Badge/src/BadgeTokens.ts
```diff
@@ -7,9 +7,9 @@ export const defaultBadgeTokens: TokenSettings<BadgeTokens, Theme> = () =>
   ({
     iconSize: 12,
     borderWidth: globalTokens.stroke.width10,
-    bottom: globalTokens.spacing.none,
-    right: globalTokens.spacing.none,
-    textMargin: globalTokens.spacing.xxs,
+    bottom: globalTokens.sizeNone,
+    right: globalTokens.sizeNone,
+    textMargin: globalTokens.size40,
```
rurikoaraki, line 12: xxs is meant to map to size20.

rurikoaraki: All values should default to web/desktop values except for android-specific files.

packages/components/Menu/src/__tests__/__snapshots__/Menu.test.tsx.snap
```diff
@@ -194,6 +194,7 @@ Array [
               "maxWidth": 300,
               "minHeight": 32,
               "minWidth": 160,
+              "padding": 6,
```
lyzhan7, line 197: Hmm any ideas why this got added?

rurikoaraki: This test is running on Android where sNudge was undefined before, so it makes sense that it's showing up now

packages/theming/theme-types/src/Theme.types.ts
```diff
@@ -15,17 +15,26 @@ export interface Spacing {
...
-  xxs?: SpacingValue;
-  xs?: SpacingValue;
-  s?: SpacingValue;
-  m?: SpacingValue;
-  l?: SpacingValue;
-  xl?: SpacingValue;
-  xxl?: SpacingValue;
-  xxxl?: SpacingValue;
-  xxxxl?: SpacingValue;
+  sizeNone?: SpacingValue;
```
rurikoaraki, line 18: We are not adding global tokens to the theme, these should be removed.

packages/theming/apple-theme/src/appleTheme.macos.ts
```diff
@@ -7,7 +7,7 @@ export function appleSpacing(): Spacing {
   return {
     s2: '8px',
     s1: '12px',
-    m: '16px',
```
lyzhan7, line 10: This seems like a different change than the other ones - I think we should probably keep this named as "m" for now. Eventually we want to remove spacing altogether from the theme and all these will be removed.

rurikoaraki:
Please add a description in your PR about your change.

packages/theming/theming-utils/src/mapPipelineToShadow.ts
```diff
@@ -7,17 +7,17 @@ import { ThemeShadowDefinition } from '@fluentui-react-native/theme-types/src/Sh
...
+    shadow16: { ambient: pipelineOutputShadow.shadow16[0], key: pipelineOutputShadow.shadow16[1] },
+    shadow28: { ambient: pipelineOutputShadow.shadow28[0], key: pipelineOutputShadow.shadow28[1] },
+    shadow64: { ambient: pipelineOutputShadow.shadow64[0], key: pipelineOutputShadow.shadow64[1] },
+    shadow2brand: { ambient: pipelineOutputShadow.shadowBrand2[0], key: pipelineOutputShadow.shadowBrand2[1] },
+    shadow4brand: { ambient: pipelineOutputShadow.shadowBrand4[0], key: pipelineOutputShadow.shadowBrand4[1] },
+    shadow8brand: { ambient: pipelineOutputShadow.shadowBrand8[0], key: pipelineOutputShadow.shadowBrand8[1] },
+    shadow16brand: { ambient: pipelineOutputShadow.shadowBrand16[0], key: pipelineOutputShadow.shadowBrand16[1] },
+    shadow28brand: { ambient: pipelineOutputShadow.shadowBrand28[0], key: pipelineOutputShadow.shadowBrand28[1] },
+    shadow64brand: { ambient: pipelineOutputShadow.shadowBrand64[0], key: pipelineOutputShadow.shadowBrand64[1] },
   };
```
lyzhan7, line 22: Thanks for updating this as well

packages/components/Badge/src/BadgeTokens.ts
lyzhan7: It looks like all the components got updated with the old iOS/Android ramp rather than the old win32/desktop ramp. You can also see the conversion here: https://github.com/microsoft/fluentui-design-tokens/pull/36/files

For corner radius and stroke width we ran into a similar problem where there were two different ramps (ex. on mobile, corner.radius.extraLarge was defined as 120 but on win32 it was 80) so it wasn't obvious what to update a reference to corner.radius.extraLarge to. For corner radius/stroke width we ended up using the desktop mappings for everything, including mobile components. The reasoning was that up until the recent iOS/Android token work, iOS was actually referencing windows tokens, and so most components were built with the assumption that the windows values were being used - using the desktop conversion would cause less unwanted change than the using the mobile conversion.

I would prefer if we stayed consistent here and just updated everything with the desktop conversion, but also fine if we wanted to use the mobile conversion table for ios/android specific files. Eventually I think we should be going through each component to check that with all the tokens changes they still align with the designs.

apps/fluent-tester/src/TestComponents/Spacing/SpacingTest.tsx
rohanprasadofficial: @lyzhan7 , you mean something like globalTokens.sizeNone (globalTokens.spacing.sizeNone we're removing  globalTokens.spacing right ? ). 

yes , let's take it from globalTokens directly.

packages/components/Badge/src/BadgeTokens.ts
rohanprasadofficial: Hmm correct, missed that. 
I mapped everything to Android/iOS tokens, just found out that the mappings are different in Desktop platforms.   

I think let's be consistent and map the values to desktop just like corner radius/stroke width and eventually we should be going through each component to check that with all the tokens changes they still align with the designs for Mobile Platforms.

packages/experimental/RadioGroup/src/__tests__/__snapshots__/RadioExperimental.test.tsx.snap
```diff
@@ -59,10 +59,10 @@ exports[`Radio component tests Radio default 1`] = `
         "borderWidth": 1,
         "height": 20,
         "justifyContent": "center",
-        "marginBottom": undefined,
-        "marginLeft": 8,
-        "marginRight": undefined,
-        "marginTop": undefined,
+        "marginBottom": 6,
+        "marginLeft": 4,
```
ankraj12, line 63: Take a look to confirm if this is an expected change with snapshots, or it's a mapping issue.

packages/theming/default-theme/src/defaultTheme.ts
```diff
@@ -74,7 +74,7 @@ function _defaultTypography(): Typography {
 }
 
 export function defaultSpacing(): Spacing {
-  return { s2: '4px', s1: '8px', m: '16px', l1: '20px', l2: '32px' };
+  return { s2: '4px', s1: '8px', size160: '16px', l1: '20px', l2: '32px' };
```
ankraj12, line 77: This can be reverted, this is not the same "m" as the current global set.

ankraj12 (changes requested):
Please add validations completed with this change.

rohanprasadofficial:
@microsoft/furn-team , i will be raising a new PR to this with few theming related changes removed as we will tackle that in separate PR as a part of removal of spacing from theme. 

New PR will be here as of now it is draft : https://github.com/microsoft/fluentui-react-native/pull/2360

PR #2358, opened 2022-11-23
Make Button respect accessibilityRole passed in props

nakambo:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

e.g. I'd like to be able to use things like menuitem or similar, which indicates a button-like control in a list of similar controls.

### Verification

* I see the expected role now with tools, e.g. Accessibility Insights.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


nakambo:
@microsoft-github-policy-service agree company="Microsoft"

packages/components/Button/src/useButton.ts
```diff
@@ -24,7 +24,7 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
        */
       ...(isDisabled && { disabled: isDisabled }),
       accessible: true,
-      accessibilityRole: 'button',
+      accessibilityRole: accessibilityRole || 'button',
```
Saadnajmi, line 27: Couldn't this also be achieved by doing a `...rest` spread at the bottom of this hook (after line 34)?

nakambo: `rest` is used by `usePressableState`, and the result is mixed in _before_ the overrides here, so it looks like these could be intentional. I do see that certain other properties (`enableFocusRing`, `iconPosition`) do apply the same pattern that I am proposing to apply here.

rurikoaraki:
We do have a Menu control now, FWIW. Change seems reasonable to me though

rurikoaraki (approved):

Saadnajmi (approved):

PR #2359, opened 2022-11-23
[Draft] Fix typescript type of compose()

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Followup to #2356 , attempt to fix the typescript types so we don't need the `@ts-ignore`. 

Needs additional validation to ensure the fix is correct

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/framework/composition/src/composeFactory.ts
+++ b/packages/framework/composition/src/composeFactory.ts
@@ -29,11 +29,13 @@ export type ComposeFactoryOptions<TProps, TSlotProps, TTokens, TTheme, TStatics 
     statics?: TStatics;
   };
 
+type TwoLevelPartial<T> = { [K in keyof T]?: Partial<T[K]> };
+
 export type ComposeFactoryComponent<TProps, TSlotProps, TTokens, TTheme, TStatics extends object = object> = ComposableFunction<TProps> & {
   __options: ComposeFactoryOptions<TProps, TSlotProps, TTokens, TTheme, TStatics>;
   customize: (...tokens: TokenSettings<TTokens, TTheme>[]) => ComposeFactoryComponent<TProps, TSlotProps, TTokens, TTheme, TStatics>;
   compose: (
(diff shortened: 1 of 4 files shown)


packages/framework/composition/src/composeFactory.ts
```diff
@@ -75,7 +77,7 @@ export function composeFactory<TProps, TSlotProps, TTokens, TTheme, TStatics ext
       themeHelper,
     );
 
-  component.compose = (customOptions: Partial<LocalOptions>) =>
+  component.compose = (customOptions: TwoLevelPartial<LocalOptions>) =>
     composeFactory<TProps, TSlotProps, TTokens, TTheme, TStatics>(
       immutableMergeCore(mergeOptions, options, customOptions) as LocalOptions,
```
Saadnajmi, line 82: I wonder if this actually should be a DeepPartial, aka, every type all the way down is optional. The reason being, the call to `immutableMergeCore` will just merge anything we provide on top of the existing composed component. So in effect, every option is optional. Thoughts?

Saadnajmi: For reference, how to implement a DeepPartial: https://stackoverflow.com/a/61132308

Saadnajmi:
I'm gonna close this PR. I had thought that either a TwoLevelPartial or a DeepPartial would be the right type that is passed into `compose` (which eventually calls `immutableMergeCore`.

But actually, the number of levels it recurses is... an argument:

Screenshot 2022-12-21 at 11 34 45 AM

Furthermore, the defined `mergeOptions` for compose specify that everything except tokens gets deep merged:
```ts
/**
 * options get deep merged except the tokens array gets appended
 */
const mergeOptions: MergeOptions = {
  tokens: 'appendArray',
  object: true,
};
```


Between the two of those, I doubt I have enough knowledge of Typescript to define a proper type soooo I'm just gonna close it.

PR #2365, opened 2022-11-29
Change jest endpoint to win32 for Callout, Menu, MenuButton, Separator, Tab

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Changing the jest endpoint to win32 for the aforementioned components, as they are mainly used in win32.
This consists of changing the config file to win32 and adding RN Win32 as a dev dependency to these packages.

### Verification

yarn test passes locally

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/package.json
+++ b/packages/components/Callout/package.json
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
     "@types/react-native": "^0.68.0",
     "react": "17.0.2",
     "react-native": "^0.68.0"
--- a/packages/components/Menu/package.json
+++ b/packages/components/Menu/package.json
@@ -43,6 +43,7 @@
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
     "@fluentui-react-native/test-tools": ">=0.1.1 <1.0.0",
+    "@office-iss/react-native-win32": "^0.68.0",
     "@types/react-native": "^0.68.0",
     "react": "17.0.2",
     "react-native": "^0.68.0"
--- a/packages/components/MenuButton/package.json
+++ b/packages/components/MenuButton/package.json
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
     "@types/react-native": "^0.68.0",
     "react": "17.0.2",
     "react-native": "^0.68.0"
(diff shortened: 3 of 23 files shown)


lyzhan7 (commented):
Thanks for making these changes - wanted to check if all the snapshot changes make sense to you? Ex. there are a couple places now where we're specifying a negative margin, that doesn't seem right but maybe that is actually what win32 is specifying

packages/components/Callout/package.json
```diff
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
```
Saadnajmi, line 43: We should probably add React Native macOS too. And maybe to peer dependencies.

packages/components/Menu/package.json
```diff
@@ -43,6 +43,7 @@
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
     "@fluentui-react-native/test-tools": ">=0.1.1 <1.0.0",
+    "@office-iss/react-native-win32": "^0.68.0",
```
Saadnajmi, line 46: Similar comment about React Native macOS

packages/components/MenuButton/package.json
```diff
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
```
Saadnajmi, line 45: And here...

packages/experimental/MenuButton/package.json
```diff
@@ -35,6 +35,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
```
Saadnajmi, line 38: And here

Saadnajmi (approved):

packages/components/Callout/package.json
rurikoaraki: The RN Win32 is specifically needed for the jest test, not for any runtime code. Is adding RN macOS and adding peer dependencies necessary for that?

rurikoaraki:
> Thanks for making these changes - wanted to check if all the snapshot changes make sense to you? Ex. there are a couple places now where we're specifying a negative margin, that doesn't seem right but maybe that is actually what win32 is specifying

Yeah the negative margins are expected. They come from https://github.com/microsoft/fluentui-react-native/blob/main/packages/utils/styling/src/getMarginAdjustment.win32.ts

lyzhan7 (approved):

packages/components/Callout/package.json
Saadnajmi: Fair enough. I'll resolve the comments.

PR #2369, opened 2022-11-29
Add E2E Win32 Spec Bundle to NPM Package

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

In order to include the JS Bundle in the NPM package, we need to modify the `files` parameter in the package.json.

Additionally, in the Publish pipeline, we need to bundle the E2E folder to create the bundle.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.ado/azure-pipelines.publish.yml
+++ b/.ado/azure-pipelines.publish.yml
@@ -50,6 +50,11 @@ jobs:
         displayName: 'yarn test [test]'
 
       - script: |
+          yarn bundle
+        workingDirectory: apps/E2E
+        displayName: 'bundle e2e spec files'
+
+      - script: |
           echo ##vso[task.setvariable variable=SkipNpmPublishArgs]--no-publish
         displayName: Enable No-Publish (npm)
         condition: ${{ parameters.skipNpmPublish }}
--- a/.ado/azure-pipelines.publish.yml
+++ b/.ado/azure-pipelines.publish.yml
@@ -46,13 +46,8 @@ jobs:
       - template: templates/yarn-install.yml
 
       - script: |
-          yarn test
-        displayName: 'yarn test [test]'
-
-      - script: |
-          yarn bundle
-        workingDirectory: apps/E2E
-        displayName: 'bundle e2e spec files'
+          yarn buildci
+        displayName: 'yarn buildci'
 
       - script: |
           echo ##vso[task.setvariable variable=SkipNpmPublishArgs]--no-publish
--- a/apps/E2E/package.json
+++ b/apps/E2E/package.json
@@ -15,6 +15,10 @@
     "url": "https://github.com/microsoft/fluentui-react-native.git",
     "directory": "apps/E2E"
   },
+  "files": [
+    "src/*",
+    "dist/*"
+  ],
   "devDependencies": {
     "@rnx-kit/cli": "^0.14.8",
     "@rnx-kit/metro-config": "^1.3.1",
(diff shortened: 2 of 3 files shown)


apps/E2E/package.json
```diff
@@ -15,6 +15,10 @@
     "url": "https://github.com/microsoft/fluentui-react-native.git",
     "directory": "apps/E2E"
   },
+  "files": [
+    "src/**/*",
+    "dist/*"
```
Saadnajmi, line 20: We want both src and dist?

Saadnajmi (approved):

apps/E2E/package.json
samuelfreiberg: Good question. I want to do some testing with both. If we only use the Bundle file (that includes all E2E win32 spec files), then WebDriverIO runs that in **one** session (because it's one file). Meaning the app doesn't reset after each spec file - it's sequential.

This may be ideal, but I've also run across flaky failures doing this locally. So I also want to make sure it works by just passing in file paths to each spec file within /src/

.ado/azure-pipelines.publish.yml
```diff
@@ -49,6 +49,11 @@ jobs:
           yarn test
         displayName: 'yarn test [test]'
 
+      - script: |
```
dannyvv, line 52: I'm curious why the step above doesn't use `yarn bundleci` which is supposed to run tests, bundles, depcheck etc....

dannyvv: it would only do work for 2 other projects that have bundle too: apps\win32 and apps\fluent-tester

samuelfreiberg: I'm not sure the history behind that. @Saadnajmi It looks like you added that two years ago. I know that's a long time ago, but do you see any reason we can't use `yarn buildci` instead of `yarn test`?

According to lage.config.js - `buildci: ['build', 'test', 'depcheck', 'bundle'],` and `test: ['build-tools', 'lint', 'build'],`.

I don't think there should be any issue?

Saadnajmi: It's a bit before my time, but:

yarn buildci was basically "Run most of the CI as one command, so that devs could run it locally to debug issues". That's effectively the "JS PR" job. There was a deliberate attempt to keep the building and bundling separate because those are time intensive tasks. 

yarn test is a subset of yarn buildci, and I guess what the platform jobs use. Each platform Job (Apple PR, Windows PR, etc) still end up bundling their platform, as a pre-step to build the test app.

For the publish pipeline, I kinda feel like we want neither. Publish IMO means we already passed PR checks so why are we testing stuff again? But it was there already and was maybe some good extra validation, so I left it there

samuelfreiberg: Did some digging - it looks like other similar GitHub repos (FI + sdx-platform) run their version of `yarn buildci`. All of them build and bundle.

And if we ever have the need to publish bundles like in this PR, it will be done automatically without the need to add a one-off script like I have above. I'm going to change `yarn test` to `yarn buildci` unless anyone has **strong** opinions otherwise.

samuelfreiberg: @rurikoaraki Do you see any issue with what I mentioned above? Any string opinion?

rurikoaraki: Yeah that seems fine to me.

PR #2376, opened 2022-12-01
Allow V2 variants to be used in V1-style Text elements and still respect Dynamic Type

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

As we were developing documentation, we found that a V1-style `Text` element with a V2 variant doesn't respect Dynamic Type. This is because we originally specified the Dynamic Type ramp independently of the variant type (see the old version of Variants.ios.ts).

The fix here is to make the relevant Dynamic Type ramp dependent on the variant itself. As a result, these ramps are now specified in appleTypography.ios.ts.

Also, to show off the fact that we can do this, we've modified the V2 usage test component.

### Verification

Here's the new V2 test component at UIContentSizeCategoryExtraExtraExtraLarge. The first line uses the V2 version; the second uses the V1 counterpart. The expected behavior is that they look the same.

As an aside, I'm aware that Caption2 looks larger than Caption1 here, despite the fact that it's the other way around at the default content size category. They actually render at the same font size, but Caption1 looks smaller because it has a letter spacing of -0.08. Regardless, in different content size categories, Caption2 is not guaranteed to be smaller than Caption1 because `UIFontMetrics` resizes text based on the ratio of the distances between adjacent baselines (also known as "leading"), not the sizes themselves. If we don't like this, we can address it separately, but it's out of scope for this PR.

Simulator Screen Shot - iPad Air (5th generation) - 2022-12-01 at 15 01 21

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/text/src/Text.tsx
+++ b/packages/components/text/src/Text.tsx
@@ -83,9 +83,6 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     [onPress, onAccessibilityTap],
   );
 
-  // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
-  const dynamicTypeVariant = Platform.OS === 'ios' ? tokens.dynamicTypeRamp : undefined;
-
   // override tokens from props
   [tokens, cache] = patchTokens(tokens, cache, {
     color,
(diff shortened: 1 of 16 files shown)


packages/components/text/src/Text.tsx
```diff
@@ -149,6 +149,7 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
       numberOfLines: truncate || !wrap ? 1 : 0,
       style: mergeStyles(tokenStyle, props.style, extra?.style, scaleStyleAdjustments),
     };
+    delete mergedProps.style.dynamicTypeRamp; // TODO(#2268): RN Text doesn't recognize dynamicTypeRamp yet, so don't let it leak through or RN will be sad
```
Saadnajmi, line 152: Can you clarify on the "RN will be sad" bit? Do you get a Redbox?

amgleitman: It gives a warning about an invalid prop. Updated to be more clear.

Saadnajmi (approved):

PR #2379, opened 2022-12-02
Android Avatar variations

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
PR proposes to add support for Avatar component for android.
- Updated tester to demonstrate usage for variations.
- Introduced 2 new tokens to provide support for - ascent and standard inverted variations. All other variations are supported re-using current tokens for supporting common types. The full variations in design guidelines are:

**1. Standard
2. Standard inverted
3. Anonymous
4. Anonymous ascent
5. Initials
6. Image
7. Overflow
8. Group**

The tester file for android demonstrates usage for each of above from consumer point, with variations.

Note that this PR does not aim to add tokens support for size, activity ring and badge. 

### Verification
Manual validation on test app.

dark1
dark2
light1
light2

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
+++ b/apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
@@ -112,13 +112,13 @@ export const StandardUsage: FunctionComponent = () => {
         flex: 1,
         justifyContent: 'space-evenly',
       }}>
-        {/* Variation - Anonymous ascent.*/}
+        {/* Variation - Anonymous accent.*/}
 
         {/* With badge.*/}
         <Avatar
           size={56}
           badge={{ status: 'available' }}
-          avatarColor={'ascent'}
+          avatarColor={'accent'}
         />
 
         {/* No ring and badge.*/}
@@ -126,7 +126,7 @@ export const StandardUsage: FunctionComponent = () => {
           accessibilityLabel="Fall-back Icon"
           accessibilityHint="A picture representing a user"
           size={56}
-          avatarColor={'ascent'}
+          avatarColor={'accent'}
         />
 
         {/* With ring. */}
@@ -136,7 +136,7 @@ export const StandardUsage: FunctionComponent = () => {
           size={56}
           activeAppearance="ring"
           active="active"
-          avatarColor={'ascent'}
+          avatarColor={'accent'}
         />
       </View>
 
--- a/apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
+++ b/apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
@@ -1,0 +1,243 @@
+import React, { FunctionComponent } from 'react';
+import { Avatar } from '@fluentui-react-native/avatar';
+import { View } from 'react-native';
+import { steveBallmerPhotoUrl } from './../PersonaCoin/styles';
+
+export const StandardUsage: FunctionComponent = () => {
+  return (
+    <View style={{ paddingBottom: 40 }} >
+      <View style={{
+        flexDirection: 'row',
+        marginHorizontal: 10,
+        paddingTop: 10,
+        paddingBottom: 5,
+        flex: 1,
+        justifyContent: 'space-evenly',
+      }}>
+
+        {/* Variation - Standard */}
+
+        {/* With badge. */}
+        <Avatar
+          badge={{ status: 'available' }}
+          avatarColor={'brand'}
+          size={40}
+        />
+
+        {/* Variation - No Ring and badge. */}
+        <Avatar
+          size={56}
+          avatarColor={'brand'}
+        />
+
+        {/* Variation - With Ring.*/}
+        <Avatar
+          active="active"
+          activeAppearance="ring"
+          size={56}
+          avatarColor={'brand'}
+        />
+      </View>
+
+      <View style={{
+        flexDirection: 'row',
+        marginHorizontal: 10,
+        paddingTop: 10,
+        paddingBottom: 5,
+        flex: 1,
+        justifyContent: 'space-evenly'
+      }}>
+
+        {/* Variation - Standard Inverted */}
+
+        {/* With badge only.*/}
+        <Avatar
+          size={56}
+          badge={{ status: 'available' }}
+          avatarColor={'brandInverted'}
+        />
+
+        {/* No ring and badge.*/}
+        <Avatar
+          size={56}
+          avatarColor={'brandInverted'}
+        />
+
+        {/* With ring.*/}
+        <Avatar
+          active="active"
+          activeAppearance="ring"
+          size={56}
+          avatarColor={'brandInverted'} />
+      </View>
+
+      <View style={{
+        flexDirection: 'row',
+        marginHorizontal: 10,
+        paddingTop: 10,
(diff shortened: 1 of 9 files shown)


packages/components/Avatar/src/Avatar.types.ts
```diff
@@ -154,6 +157,7 @@ export interface AvatarProps extends IViewProps, AvatarConfigurableProps {
    * using the `getInitials` function.
    *
```
ankraj12, line 158: I had a query on "`getInitials`()" function. Is it intended to keep it (and other functions in useAvatar) exported and mentioned in documentation?

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
```diff
@@ -0,0 +1,243 @@
+import React, { FunctionComponent } from 'react';
+import { Avatar } from '@fluentui-react-native/avatar';
+import { View } from 'react-native';
+import { steveBallmerPhotoUrl } from './../PersonaCoin/styles';
+
+export const StandardUsage: FunctionComponent = () => {
+  return (
+    <View style={{ paddingBottom: 40 }} >
+      <View style={{
```
lyzhan7, line 9: For all the view style properties in this file, we should try to avoid creating the style object inline since that causes unnecessary rerenders. If we move the style object creation outside of the function component then there won't be unnecessary rerenders. 

Here's an [explanation](https://github.com/microsoft/fluentui-react-native/pull/2231#discussion_r995097257) Ruriko posted in one of my PRs:

> The way React components figure out if it needs to rerender is by shallow comparing all of its props and if anything is different, it will rerender the component. For objects (objects, arrays, functions), that means that if a new instance of that object is created and passed into the compare, then the shallow compare will fail because it's a different instance, even if the contents of the object are the same and so in reality no work needed to be done. You can think of it as a pointer comparison if you're used to native.
> 
> Anything inside of a (function) component is rerun when something asks components to rerender. So that means any instantiation of a const object will be redone, and new instances of those objects will be created. In this case exampleStrokeStyle has a new instance on every render and so the inner most view will be invalidated on every render.
>
> Moving the const object outside the components works because that code is only run once.

The other benefit is that there are a couple places where you have the same style declared several times, so that will fix the duplicated code

packages/components/Avatar/src/AvatarTokens.android.ts
```diff
@@ -0,0 +1,203 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { AvatarTokens } from '.';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+
+export const defaultAvatarTokens: TokenSettings<AvatarTokens, Theme> = (t: Theme) =>
+  ({
```
lyzhan7, line 7: I'm realizing that platform specific token files are difficult to review - ex. it's hard for reviewers to know what the diff is between this file and the other AvatarTokens.ts file. Thinking in the future for my PRs if I make a new platform specific token file, I'll try to commit an exact duplicate first and then make the platform-specific changes after so there's a diff people can reference. 

Doing a brief manual diff on this - seems like mostly some colors changed, and we also added brandinverted/accent colors. Are there any other changes?

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
```diff
@@ -0,0 +1,243 @@
...
+        paddingBottom: 5,
+        flex: 1,
+        justifyContent: 'space-evenly',
+      }}>
+
+        {/* Variation - Standard */}
+
+        {/* With badge. */}
+        <Avatar
+          badge={{ status: 'available' }}
```
rurikoaraki, line 22: Please make this a const object or memoize it.

packages/components/Avatar/src/Avatar.types.ts
rurikoaraki: It needs to be exported in order to be tested in a jest file.

rurikoaraki:
Will defer to @ValentinaKozlova to sign off

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
ValentinaKozlova: Yeah, we have styles.ts in Common folder where shared styles can be added

packages/components/Avatar/src/AvatarTokens.android.ts
```diff
@@ -0,0 +1,203 @@
...
+    magenta: getColorProps('magenta', t),
+    plum: getColorProps('plum', t),
+    beige: getColorProps('beige', t),
+    mink: getColorProps('mink', t),
+    platinum: getColorProps('platinum', t),
+    anchor: getColorProps('anchor', t),
+    burgundy: getColorProps('burgundy', t),
+    hotPink: getColorProps('hotPink', t),
+    orchid: getColorProps('orchid', t),
+  } as AvatarTokens);
```
ValentinaKozlova, line 176: Are there only those colors which are used in android? Are there new colors?

packages/components/Avatar/src/AvatarTokens.android.ts
ankraj12: That sounds about right in terms of changes to this file. And it's a good suggestion if we can try to commit unchanged version first, thank you!

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
ankraj12: Updated to const styles.

packages/components/Avatar/src/AvatarTokens.android.ts
ankraj12: There are 3 colors I removed as they are not in Android designs, but no additional colors.

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
```diff
@@ -2,97 +2,45 @@ import React, { FunctionComponent } from 'react';
...
-      }}>
+      <View style={mobileStyles.testVariants}>
         {/*Variation - Anonymous*/}
 
         {/* With Badge.*/}
-        <Avatar
-          size={56}
-          badge={{ status: 'available' }}
-        />
+        <Avatar size={56} badge={{ status: 'available' }} />
```
ValentinaKozlova, line 40: There's one comment left from Ruriko about useMemo for objects. It should be an object for the badge e.g.
 `const badgeAvailable = useMemo(
    () => ({
      status: 'available' as PresenceBadgeStatus
    }),
    [],
  );`
P.S. I'll need to update the same for my examples :)

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
ankraj12: I see, this was  not for styles but for the component props for badge as per @valentynas comment. I would take a note to resolve this comment in next PR's.

PR #2383, opened 2022-12-03
Use V2 font tokens for Notification text elements

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Add some of our new font tokens to Notification so we can see how they're used.

Although not tested on Android, this will almost certainly mess with the appearance of Notification on Android because V2 variants have not been defined yet there. However, we've gotten permission from FURN Android developers to go ahead with  this anyway.

### Verification

The Notification element looks exactly the same on iOS.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Notification/src/Notification.tsx
+++ b/packages/components/Notification/src/Notification.tsx
@@ -3,7 +3,7 @@ import { notification, NotificationType, NotificationProps } from './Notificatio
 import { Pressable } from '@fluentui-react-native/pressable';
 import { Platform, PressableProps, useWindowDimensions, View, ViewStyle, ViewProps } from 'react-native';
 import { Icon, createIconProps } from '@fluentui-react-native/icon';
-import { TextV1 as Text } from '@fluentui-react-native/text';
+import { Body2, Body2Strong } from '@fluentui-react-native/text';
 import { stylingSettings } from './Notification.styling';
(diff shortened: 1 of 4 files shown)


packages/components/Notification/src/Notification.tsx
```diff
@@ -53,8 +53,8 @@ export const Notification = compose<NotificationType>({
     root: Pressable,
     icon: Icon,
     contentContainer: View,
-    title: Text,
-    message: Text,
+    title: Body2Strong,
+    message: Body2,
```
Saadnajmi, line 57: I think I prefer the old styling better, AKA: Text as the slot and the variant set in slot props. Was there a reason for this change I'm perhaps missing?

amgleitman: The reason is to get a full end-to-end test of the new V2 text components.

Saadnajmi: Talked offline: Idea is to use the new component aliases E2E in a component. While it's not my personal preference, it does makes sense and there's nothing technically wrong with it so approving.

Saadnajmi (approved):

PR #2384, opened 2022-12-05
Integrate Shadows in Android Theme

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Enabled code path for android theme to read from pipeline.

As a longer conversation, I would like to reopen a design discussion not long ahead on why do we have a shadow alias token for android (the only usage is to map a fixed numerical value, with nothing that varies on theme modes); and if we can change the token structure to depict more meaningfully the information for elevation ( instead of inferring from blur). But at this point, it is a scoped out work hence fixing the current implementation.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki (approved):

packages/theming/theme-tokens/src/getShadowTokens.android.ts
```diff
@@ -0,0 +1,14 @@
...
+import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+
+export function getShadowTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return lightShadowTokens;
+  } else if (mode === 'dark') {
+    return darkShadowTokens;
+  }
+
+return lightShadowTokens;
```
lyzhan7, line 13: Could maybe add an assert that we don't reach this line? Are there other appearance modes on Android

ankraj12: There are no other appearance options in Android - only light and dark.

PR #2389, opened 2022-12-06
Add maximumFontSize

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This adds a `maximumFontSize` token that can cap the size of Dynamic Type text.

### Verification

Added an example to the TextV2 test page. The text takes up (roughly) the entirety of the width of an iPhone 14 Pro at its maximum size, as seen below.

| `UIContentSizeCategoryLarge` | `UIContentSizeCategoryAccessibilityLarge` (AX2)                                      |
| Simulator Screen Shot - iPhone 14 Pro - 2022-12-05 at 17 52 33 | Simulator Screen Shot - iPhone 14 Pro - 2022-12-05 at 17 50 06 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
+++ b/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
@@ -19,7 +19,6 @@ import {
 } from '@fluentui-react-native/text';
 
 const LimitedDisplay = Display.customize({
-  dynamicTypeRamp: 'display',
   maximumFontSize: 70,
 });
 
--- a/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
+++ b/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
@@ -18,6 +18,11 @@ import {
   Title3,
 } from '@fluentui-react-native/text';
 
+const LimitedDisplay = Display.customize({
+  dynamicTypeRamp: 'display',
+  maximumFontSize: 70,
+});
+
 export const V2Usage: React.FunctionComponent = () => {
   return (
     <View>
@@ -44,6 +49,7 @@ export const V2Usage: React.FunctionComponent = () => {
         <TextV1 variant="title1">Title1</TextV1>
         <LargeTitle>LargeTitle</LargeTitle>
         <TextV1 variant="largeTitle">LargeTitle</TextV1>
+        <LimitedDisplay>Disp(&le;70)</LimitedDisplay>
         <Display>Display</Display>
         <TextV1 variant="display">Display</TextV1>
       </Stack>
--- a/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
+++ b/apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
@@ -18,10 +18,6 @@ import {
   Title3,
 } from '@fluentui-react-native/text';
 
-const LimitedDisplay = Display.customize({
-  maximumFontSize: 70,
-});
-
 export const V2Usage: React.FunctionComponent = () => {
   return (
     <View>
@@ -48,7 +44,6 @@ export const V2Usage: React.FunctionComponent = () => {
         <TextV1 variant="title1">Title1</TextV1>
         <LargeTitle>LargeTitle</LargeTitle>
         <TextV1 variant="largeTitle">LargeTitle</TextV1>
-        <LimitedDisplay>Disp(&le;70)</LimitedDisplay>
         <Display>Display</Display>
         <TextV1 variant="display">Display</TextV1>
       </Stack>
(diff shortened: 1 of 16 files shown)


packages/components/text/src/Text.tsx
```diff
@@ -123,6 +124,15 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
       fontSize: tokenStyle.fontSize * scaleFactor,
       lineHeight: tokenStyle.lineHeight * scaleFactor,
     };
+    if (scaleStyleAdjustments.fontSize > maximumFontSize) {
+      scaleStyleAdjustments = {
+        fontSize: maximumFontSize,
+        // Scale line height accordingly as well
+        // We use "as number" below because TypeScript's type checker thinks scaleStyleAdjustments.fontSize
+        // could be a named font size, even though we just wrote it as a number a few lines up.
+        lineHeight: scaleStyleAdjustments.lineHeight * (maximumFontSize / (scaleStyleAdjustments.fontSize as number)),
```
lyzhan7, line 133: Just wanted to note, I guess we're making an assumption here that the scale factor should be the same for the maximum font size?

amgleitman: Yes. The maximum font size is specified in "real" points, so it's not affected by the scale factor.

(It's actually the other way around; the scale factor is effectively capped at `maximumFontSize / tokenStyle.fontSize`. Maybe it would be clearer to write it this way?)

packages/components/text/src/Text.tsx
```diff
@@ -88,6 +88,7 @@ export const Text = compressible<TextProps, TextTokens>((props: TextProps, useTo
     color,
     variant,
     fontFamily: font == 'base' ? 'primary' : font,
+    fontMaximumSize: tokens.maximumFontSize,
```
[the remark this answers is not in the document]
amgleitman: I hadn't considered this, although I imagine that a specific font size would be easier to fine-tune for appearance purposes than a content size category.

We could try allowing for both if we try to bring this into RNCore, although that's a discussion for another time.

packages/components/text/src/Text.tsx
lyzhan7: Just needed to think about this a bit more and makes more sense now - but if you think there's a clearer way to write it go for it

lyzhan7:
Talked offline - seems like it might be a little unclear to someone who has no context what maximumFontSize is on what the Disp (<= 70) is testing. Maybe making a new section called "maximum font size" and giving a short description on what it is would be more clear.

lyzhan7 (approved):
Besides updating the test page to be more clear this looks good to me. I am wondering though if we should have a discussion about Mike's suggestion before merging? If it's another possible approach maybe it would replace this PR.

packages/utils/tokens/src/text-tokens.ts
```diff
@@ -14,14 +14,16 @@ export interface FontStyleTokens {
   fontWeight?: keyof Typography['weights'] | TextStyle['fontWeight'];
   fontLineHeight?: TextStyle['lineHeight'];
   fontLetterSpacing?: TextStyle['letterSpacing'];
-  fontDynamicTypeRamp?: string; // TODO(#2268): Import type from RN directly
+  // TODO(#2268): Import these from RN directly
+  fontDynamicTypeRamp?: string;
+  fontMaximumSize?: number;
```
PPatBoyd, line 19: Can we add comments indicating these are Apple-specific component tokens?

apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
```diff
@@ -44,6 +48,7 @@ export const V2Usage: React.FunctionComponent = () => {
         <TextV1 variant="title1">Title1</TextV1>
         <LargeTitle>LargeTitle</LargeTitle>
         <TextV1 variant="largeTitle">LargeTitle</TextV1>
+        <LimitedDisplay>Disp(&le;70)</LimitedDisplay>
```
Saadnajmi, line 51: What's with `Disp(&le;70)` ?

packages/components/text/src/Text.types.ts
```diff
@@ -11,7 +11,9 @@ export const textName = 'Text';
 export type TextTokens = Omit<FontTokens, 'fontFamily'> &
   IForegroundColorTokens &
   Omit<TextStyle, 'fontSize' | 'fontWeight' | 'color'> & {
-    dynamicTypeRamp?: string; // TODO(#2268): Remove once RN Core properly supports Dynamic Type scaling
+    // TODO(#2268): Remove these once RN Core properly supports Dynamic Type scaling
+    dynamicTypeRamp?: string;
+    maximumFontSize?: number;
```
Saadnajmi, line 16: nit" Remove these `iOS specific properties`

I think err on over documenting things to be platform specific

Saadnajmi: Also,I thought the format we followed was `GH #999: Do blah`?

Saadnajmi (approved):
Approved with minor comments

packages/utils/tokens/src/text-tokens.ts
amgleitman: Can do.

packages/components/text/src/Text.types.ts
amgleitman: Adding documentation should be easy.

Also, I don't see any particular guide for marking TODOs associated with GitHub issues. This should probably be written down somewhere.

apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.ios.tsx
amgleitman: This is getting replaced with a more obvious usage of `maximumFontSize`.

PR #2394, opened 2022-12-06
Add additional a11y tests for Text V1 component

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Previously, we were only testing the deprecated Text component (exported as Text from @fluentui-react-native/text). This adds tests for TextV1 from @fluentui-react-native/text (exported as TextV1) and adds additional a11y tests for Text from @fluentui-react-native/experimental-text.

Even though Text (not TextV1) is deprecated, there are still people using it so I think it's valuable to have tests for it until it is fully removed.

Also, I added some helpful error logs to some of the base methods from BasePage.ts to help remind others that these methods need to be implemented in order for E2E tests to work.

Text
image
image

Experimental Text
image
image

### Verification

Locally ran tests with `yarn e2etest`

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/Text/consts.ts
```diff
@@ -2,9 +2,17 @@ export const HOMEPAGE_TEXT_BUTTON = 'Homepage_Text_Button';
...
+export const FIRST_TEXT_COMPONENT = 'Text_Test_Component';
 export const TEXT_ACCESSIBILITY_LABEL = 'E2E testing Text_Test_Component accessibility label';
 
 /* E2E Testing Text 2 */
-export const SECOND_TEXT_COMPONENT = 'Text_No_A11y_label_Component';
-export const TEXT_COMPONENT_CONTENT = 'Test Text2 - No Accessibility Label'; // A component on each specific test page
+export const SECOND_TEXT_COMPONENT = 'Text_No_A11y_Label_Component';
+export const TEXT_COMPONENT_CONTENT = 'Text - No Accessibility Label Set';
+
+/* E2E Testing Deprecated Text 1 */
```
samuelfreiberg, line 12: Why does the deprecated text have more constants that V1 test?

apps/fluent-tester/src/TestComponents/Text/consts.ts
```diff
@@ -2,9 +2,17 @@ export const HOMEPAGE_TEXT_BUTTON = 'Homepage_Text_Button';
...
 
 /* E2E Testing Text 1 */
-export const FIRST_TEXT_COMPONENT = 'Text_Test_Component'; // A component on each specific test page
+export const FIRST_TEXT_COMPONENT = 'Text_Test_Component';
 export const TEXT_ACCESSIBILITY_LABEL = 'E2E testing Text_Test_Component accessibility label';
 
 /* E2E Testing Text 2 */
-export const SECOND_TEXT_COMPONENT = 'Text_No_A11y_label_Component';
-export const TEXT_COMPONENT_CONTENT = 'Test Text2 - No Accessibility Label'; // A component on each specific test page
+export const SECOND_TEXT_COMPONENT = 'Text_No_A11y_Label_Component';
```
samuelfreiberg, line 9: Why does deprecated Text have more constants than V1 text?

apps/E2E/src/common/BasePage.ts
```diff
@@ -237,31 +237,36 @@ export class BasePage {
   // Returns: UI Element
   // The Text component on each test page containing the title of that page. We can use this to determine if a test page has loaded correctly.
   get _testPage() {
+    console.error('You have to implement the method _testPage.');
```
samuelfreiberg, line 240: To provide a bit more context, let's add "Each class extending BasePage must implement it's own ___"

apps/E2E/src/Text/specs/Text.spec.win.ts
```diff
@@ -28,17 +32,39 @@ describe('Text Accessibility Testing', () => {
...
 
-  it('Text - Do not accessibilityLabel -> Default to content', async () => {
-    await expect(await TextPageObject.getAccessibilityLabel(ComponentSelector.Secondary)).toEqual(TEXT_COMPONENT_CONTENT);
+  it('Text - Do not set accessibilityLabel -> Default to content', async () => {
+    await expect(await TextPageObject.getTextAccessibilityLabel(TextComponentSelector.Second)).toEqual(TEXT_COMPONENT_CONTENT);
+    await expect(await TextPageObject.didAssertPopup()).toBeFalsy(TextPageObject.ERRORMESSAGE_ASSERT);
+  });
+});
+
+describe('Deprecated Text Accessibility Testing', () => {
```
samuelfreiberg, line 50: How is this `describe()` block different than the `describe()` block above? And how is it different than the ExperimentalText `describe()` block?

I'm a bit confused because I'd assume we'd only have two `describe('accessibility testing')` blocks spanning across the two files - one for deprecated and one for V1. But it seems that we have 3

apps/fluent-tester/src/TestComponents/Text/consts.ts
ksiler: The @fluentui-react-native/text package exports two versions of Text: [TextV1](https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/text/src/index.ts#L1) and the [deprecated Text](https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/text/src/index.ts#L26) component.

apps/E2E/src/Text/pages/TextPageObject.ts
```diff
@@ -22,10 +66,22 @@ class TextPageObject extends BasePage {
     return By(FIRST_TEXT_COMPONENT);
   }
 
-  get _secondaryComponent() {
+  get _firstComponent() {
```
samuelfreiberg, line 69: I think we should get more specific with these names, since they are two different underlying components. The reader should know from reading this file that the `_firstComponent` has a different underlying component than `_thirdComponent`.

This helps in long-term sustainability. We have to assume that these tests are going to fail eventually, and an engineer who knows nothing about who wrote these tests will be tasked with debugging. We want to make the experience as easy as possible for that engineer (this really is lacking on our native side, which is why I think we should be extra diligent about improving it in FURN).

So perhaps these get renamed to: `_v1FirstComponent`, `_v1SecondComponent`, `_deprecatedFirstComponent`,  `_deprecatedSecondComponent`.

And perhaps add a comment about the deprecated components briefly explaining that it's in the process of deprecated, so we still need to write tests to support it

apps/E2E/src/Text/pages/TextPageObject.ts
```diff
@@ -22,10 +66,22 @@ class TextPageObject extends BasePage {
     return By(FIRST_TEXT_COMPONENT);
   }
 
-  get _secondaryComponent() {
```
samuelfreiberg, line 25: It's seems a bit strange to have two components return the same element (`_primaryComponent` and `_firstComponent`).

Rather than forcing a consumer of BasePage to use `_primaryComponent`, maybe we can make an overload that takes a different component as well.

If you look at `scrollToTestElement(component?: WebdriverIO.Element)` in basepage, we give the user the option to pass in a different component rather than using _primaryComponent.

We can also change `waitForPrimaryElementDisplayed()` to take in the same parameter and perform the same logic.

Then, in Text.spec.win.ts, instead of calling ` await TextPageObject.scrollToTestElement();` and `await TextPageObject.waitForPrimaryElementDisplayed(PAGE_TIMEOUT);`, we can call `await TextPageObject.scrollToTestElement(TextComponentSelector.First);` and `await TextPageObject.waitForPrimaryElementDisplayed(TextComponentSelector.First, PAGE_TIMEOUT);`

Then we can remove _primaryComponent completely from this file (or, if there are other functions like above, we can change them in BasePage as well)

apps/E2E/src/Text/pages/TextPageObject.ts
```diff
@@ -1,12 +1,56 @@
...
+export const enum TextComponentSelector {
+  First = 0, // this._firstComponent
+  Second, // this._secondComponent
+  Third, // this._thirdComponent
+  Fourth, // this._fourthComponent
+}
+
 class TextPageObject extends BasePage {
+  /* Gets the accessibility role of an UI element given the selector */
+  async getTextAccessibilityRole(componentSelector: TextComponentSelector): Promise<string> {
```
samuelfreiberg, line 21: To enable code-reuse, let's create a function called `getTextComponent(componentSelector: TextComponentSelector) : Promise<WebdriverIO.Element>` and put the switch statement in there.

That way, in `getTextAccessibilityRole()` and `getTextAccessibilityLabel()`, we can just call 
`return await (this.getElementAttribute(await this.getTextComponent(componentSelector), Attribute.AccessibilityRole)`

apps/E2E/src/Text/pages/TextPageObject.ts
```diff
@@ -1,12 +1,56 @@
...
   FIRST_TEXT_COMPONENT,
   HOMEPAGE_TEXT_BUTTON,
   SECOND_TEXT_COMPONENT,
+  THIRD_TEXT_COMPONENT,
+  FOURTH_TEXT_COMPONENT,
 } from '../../../../fluent-tester/src/TestComponents/Text/consts';
 import { BasePage, By } from '../../common/BasePage';
 
+export const enum TextComponentSelector {
+  First = 0, // this._firstComponent
```
samuelfreiberg, line 13: Let's change these two `V1_First`, `V1_Second`, `Deprecating_First`, `Deprecating_Second`. Something along those lines so a reader knows that these are two separate underlying components.

It might be confusing to the user if they saw us getting the AccessibilityRole of multiple of these components and not know that they are two underlying components

samuelfreiberg (changes requested):

apps/E2E/src/Text/pages/TextPageObject.ts
samuelfreiberg: Or, on second thought, we'd want to scroll to the component lowest on the test page, so probably TextComponentSelector.Fourth

ksiler: >It's seems a bit strange to have two components return the same element (_primaryComponent and _firstComponent).

I agree. I don't think the _primary/secondaryComponent pattern works for everyone. While it provides a good example on how to manage more than one test element, I don't think we should make defining a _primary/secondaryComponent necessary. But requiring an implementation for _testPage/_pageButton/_pageName makes sense.

I created waitForElementDisplayed() in BasePage.ts to allow the user to specify a UI element instead and made `waitForPrimaryElementDisplayed()` call it with _primaryComponent so current users of `waitForPrimaryElementDisplayed` go unaffected.

apps/E2E/src/Text/specs/Text.spec.win.ts
```diff
@@ -21,24 +25,50 @@ describe('Text Testing Initialization', function () {
...
 });
 
-describe('Text Accessibility Testing', () => {
+describe('V1 Text Accessibility Testing', () => {
   beforeEach(async () => {
-    await TextPageObject.scrollToTestElement();
-    await TextPageObject.waitForPrimaryElementDisplayed(PAGE_TIMEOUT);
+    await TextPageObject.scrollToTestElement(await TextPageObject._v1FirstComponent);
+    await TextPageObject.waitForElementDisplayed(await TextPageObject._v1FirstComponent, PAGE_TIMEOUT);
   });
```
ksiler, line 32: @samuelfreiberg in another comment you said:

>Or, on second thought, we'd want to scroll to the component lowest on the test page, so probably TextComponentSelector.Fourth

Do we need to scroll to the lowest item on the test page before each test? The tests have been working fine w/o it. Is this just to ensure we're displaying all the test elements before we run a test? If we're really trying to see if the UI element we're testing is displayed before testing it, shouldn't we be searching for that specific UI element instead of the first/last element to be displayed? Or are we just trying to scroll all the test elements into view for screenshot/errorshots purposes?

Also, shouldn't calling `scrollToTestElement()` be enough here since it checks if the specified Element is displayed?

apps/E2E/src/common/BasePage.ts
```diff
@@ -159,15 +159,20 @@ export class BasePage {
 
   /* Waits for the primary UI test element to be displayed. If the element doesn't load before the timeout, it causes the test to fail. */
   async waitForPrimaryElementDisplayed(timeout?: number): Promise<void> {
-    await browser.waitUntil(async () => await (await this._primaryComponent).isDisplayed(), {
+    await this.waitForElementDisplayed(await this._primaryComponent, timeout);
+  }
+
+  /* Waits for the the specified UI test element to be displayed. If the element doesn't load before the timeout, it causes the test to fail. */
+  async waitForElementDisplayed(component: WebdriverIO.Element, timeout?: number): Promise<void> {
+    await browser.waitUntil(async () => await (await component).isDisplayed(), {
```
samuelfreiberg, line 167: I don't believe we need 2 awaits here, it should be `await component.isDisplayed()`. Because component is a `WebDriverIO.Element`, not a Promise. `await` is only needed with a Promise

change/@fluentui-react-native-e2e-testing-e9079191-951f-43ef-8361-8be50a28b7da.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add additional tests for Text",
+  "packageName": "@fluentui-react-native/e2e-testing",
+  "email": "krsiler@microsoft.com",
+  "dependentChangeType": "patch"
```
samuelfreiberg, line 6: Is this supposed to be "patch" while the "type" is "minor"?

change/@fluentui-react-native-tester-4b0c6e11-d0e5-4e3f-8feb-8ca69747c55f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add additional tests for Text",
+  "packageName": "@fluentui-react-native/tester",
+  "email": "krsiler@microsoft.com",
+  "dependentChangeType": "patch"
```
samuelfreiberg, line 6: Same as above

change/@fluentui-react-native-e2e-testing-e9079191-951f-43ef-8361-8be50a28b7da.json
ksiler: hmmm I would think not but that's what `yarn change` resolved it to when I selected 'Minor` as the change type. I'm not sure how it makes that decision. Though I think this change should be a patch instead (since I'm not changing anything) so I'll update the change files.

samuelfreiberg (approved):

PR #2402, opened 2022-12-08
Rename test pages

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Renaming test pages that are for two versions of the same component from plain vs. experimental to pre-v1 vs. v1. Hopefully this clears up some confusion around "why is it named experimental when the code isn't in experimental?" This also helps with components that are in different stages on different platforms since pre-v1 doesn't necessarily equate to deprecated for all platforms even if a v1 component exists.

### Verification

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/testPages.ts
+++ b/apps/fluent-tester/src/testPages.ts
@@ -78,7 +78,7 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'macos', 'win32'],
   },
   {
-    name: 'Checkbox Pre-V1',
+    name: 'Checkbox Legacy',
     component: CheckboxTest,
     testPage: HOMEPAGE_CHECKBOX_BUTTON,
     platforms: ['android', 'ios', 'macos', 'win32', 'windows'],
@@ -145,7 +145,7 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'ios', 'macos', 'win32'],
   },
   {
-    name: 'Link Pre-V1',
+    name: 'Link Legacy',
     component: LinkTest,
     testPage: HOMEPAGE_LINK_BUTTON,
     platforms: ['android', 'ios', 'macos', 'win32', 'windows'],
@@ -163,7 +163,7 @@ export const tests: TestDescription[] = [
     platforms: ['macos', 'win32'],
   },
   {
-    name: 'MenuButton Pre-V1',
+    name: 'MenuButton Legacy',
     component: MenuButtonTest,
     testPage: HOMEPAGE_MENUBUTTON_BUTTON,
     platforms: ['macos', 'win32'],
@@ -199,7 +199,7 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'ios', 'macos', 'win32', 'windows'],
   },
   {
-    name: 'RadioGroup Pre-V1',
+    name: 'RadioGroup Legacy',
     component: RadioGroupTest,
     testPage: HOMEPAGE_RADIOGROUP_BUTTON,
     platforms: ['macos', 'win32'],
@@ -253,7 +253,7 @@ export const tests: TestDescription[] = [
     platforms: ['win32'],
   },
   {
-    name: 'Tabs Pre-V1',
+    name: 'Tabs Legacy',
     component: TabsTest,
     testPage: HOMEPAGE_TABS_BUTTON,
     platforms: ['macos', 'win32', 'windows'],
@@ -265,7 +265,7 @@ export const tests: TestDescription[] = [
     platforms: ['macos', 'win32', 'windows'],
   },
   {
-    name: 'Text Pre-V1',
+    name: 'Text Legacy',
     component: TextTest,
     testPage: HOMEPAGE_TEXT_BUTTON,
     platforms: ['android', 'ios', 'macos', 'win32', 'windows'],
--- a/apps/fluent-tester/src/testPages.ts
+++ b/apps/fluent-tester/src/testPages.ts
@@ -78,13 +78,13 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'macos', 'win32'],
   },
   {
-    name: 'Checkbox',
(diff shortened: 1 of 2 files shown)


rurikoaraki:
Would it be worth also renaming all the files and stuff? I'm fine with doing that, just didn't know if that would be helpful or not.

apps/fluent-tester/src/testPages.ts
```diff
@@ -78,13 +78,13 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'macos', 'win32'],
   },
   {
-    name: 'Checkbox',
```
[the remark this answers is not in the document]
rurikoaraki: Yeah I'm in the process of trying to combine Checkbox,

For Link and Radiogroup, two pages is expected as they're under construction still.
The rest of the components we would need to clean up.

The main difficulty with combining the test pages is getting E2E testing to cooperate, since there's two controls on one page. That was an issue I ran into when trying to combine checkbox last time.

apps/fluent-tester/src/testPages.ts
```diff
@@ -78,13 +78,13 @@ export const tests: TestDescription[] = [
...
     component: CheckboxTest,
     testPage: HOMEPAGE_CHECKBOX_BUTTON,
     platforms: ['android', 'ios', 'macos', 'win32', 'windows'],
   },
   {
-    name: 'Checkbox (Experimental)',
+    name: 'Checkbox V1',
     component: ExperimentalCheckboxTest,
     testPage: HOMEPAGE_CHECKBOX_EXPERIMENTAL_BUTTON,
     platforms: ['android', 'ios', 'macos', 'win32'], // 'windows': GH#935: Temporarily disabling while SVGs don't work in windows
```
Saadnajmi, line 90: Isn't the first one the V1 Checkbox page, while the second one is the experimental one (native module on macOS, and.. redirect to V1 elsewhere) ?

apps/fluent-tester/src/testPages.ts
```diff
@@ -78,13 +78,13 @@ export const tests: TestDescription[] = [
     platforms: ['android', 'macos', 'win32'],
   },
   {
-    name: 'Checkbox',
+    name: 'Checkbox Pre-V1',
```
Saadnajmi, line 81: I think I like "Legacy" or "v0" over "pre-v1" but naming is hard so I wouldn't call it blocking.

apps/fluent-tester/src/testPages.ts
```diff
@@ -163,13 +163,13 @@ export const tests: TestDescription[] = [
...
     component: MenuButtonTest,
     testPage: HOMEPAGE_MENUBUTTON_BUTTON,
     platforms: ['macos', 'win32'],
   },
   {
-    name: 'MenuButton (Experimental)',
+    name: 'MenuButton V1',
     component: ExperimentalMenuButtonTest,
     testPage: HOMEPAGE_EXPERIMENTAL_MENU_BUTTON,
     platforms: ['macos', 'win32'],
```
Saadnajmi, line 175: Is MenuButton V1? Or is it going away with MenuV1? I wasn't actually sure. @chiuam fyi

apps/fluent-tester/src/testPages.ts
rurikoaraki: No. The first checkbox page uses the deprecated checkbox. It does look like the second one tests the checkbox from experimental, which would use the native module on macOS and V1 on win32.

apps/fluent-tester/src/testPages.ts
rurikoaraki: MenuButton is its own component, but we should probably redo it because it uses the old ContextualMenu.

rurikoaraki: https://react.fluentui.dev/?path=/docs/components-button-menubutton--default

apps/fluent-tester/src/testPages.ts
rurikoaraki: Sure, I can use legacy instead.

apps/fluent-tester/src/testPages.ts
Saadnajmi: Is it worth calling V1 right now then?

rurikoaraki: I guess not. Idk what to name it instead though, suggestions?

Saadnajmi: .... experimental 😅

One of them is a macOS native module, so there's opportunity to move that to a package and maybe just.. not have the old one? Idk

apps/fluent-tester/src/testPages.ts
Saadnajmi: Oh, thanks for the clarification. That's simpler than I thought :)

apps/fluent-tester/src/testPages.ts
rurikoaraki: I'm not a huge fan of experimental since it seems to be causing confusion. I think V1 is ok since it's meant to be the test page for the v1 version of the MenuButton, and it will be changed in the future to use Menu. But if you think that will be confusing too I can leave it at experimental

Saadnajmi: I guess "V1" with proper platform status is fine. That's why those are separate =)

apps/fluent-tester/src/testPages.ts
Saadnajmi: @ejlayne for the future, remind me to ask you about moving the macOS native wrapper components to their own packages / a separate component in the same package. I think this is something we agreed to do in the past but didn't follow through with and is left as a hanging fruit.

Saadnajmi (approved):

PR #2403, opened 2022-12-08
Revert usePressableState for SubmenuItem

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The usage of usePressableState in SubmenuItem causes a bug on macOS where SubmenuItem doesn't get keyboard focus on mouse hover. I couldn't figure out why onHoverIn gets overridden/not getting called in SubmenuItem, but we can live with this for now until MenuV1 transition is ready.

### Verification

Submenu item works as expected on macOS and win32

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):
Approving simple revert, will also try and test the difference between the two myself later

PR #2405, opened 2022-12-09
Update BasePage + ExperimentalButton, FocusZone, NativeTesting page and spec with more intuitive errors

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, FURN E2E tests provide few readable + intuitive errors which makes it hard to understand why tests have failed. Also developing tests for FURN can be hard because of how our PageObjects are currently set up. 

This commit does the following:
- Updates BasePage with a new method for testing Attribute equality which includes a nice error message when the test fails. 
  - This is also implemented in ExperimentalButton and FocusZone win32 tests.
- Turns BasePage into an abstract class so we can ensure that new tests and their page objects will have everything necessary implemented. These include getters for `_pageName`, `_testPage`, and `_pageButton`. (`_primaryComponent` and `_secondaryComponent` aren't included as required implementations because they seem like they may be removed [here](https://github.com/microsoft/fluentui-react-native/pull/2394))
- Add generic `click` and `sendKey` methods in BasePage
- Updated error messages for functional tests on ExperimentalButton and FocusZone's spec as well as for NativeTesting's spec
- Removed `ButtonSelector` enum in ExperimentalButton page object and `GridButtonSelector` enum in FocusZone page object.
- Renamed a prop `testPage` -> `testPageButton` for the navigation button to be clearer in testPage definitions
 
### Verification

Tests pass in CI.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/ButtonExperimental/pages/ButtonExperimentalPageObject.ts
```diff
@@ -34,10 +36,20 @@ class ButtonExperimentalPageObject extends BasePage {
...
+    switch (buttonSelector) {
+      case ButtonSelector.PrimaryButton:
+        return this._primaryComponent;
+      case ButtonSelector.SecondaryButton:
+        return this._secondaryComponent;
     }
-    return await this._primaryComponent;
+  }
+
+  async accessibilityLabelIsEqualTo(selector: ButtonSelector, expectedValue: any): Promise<boolean> {
```
samuelfreiberg, line 47: These two functions will be re-used for every single spec. How can we make this re-usable without creating new definitions for each page object?

samuelfreiberg: I'm thinking we should move this function to BasePage, and change the first parameter to `element: WebDriverIO.Element`. And then from the spec, we can call:

`ButtonExperimentalPageObject.accessibilityLabelIsEqualTo(ButtonExperimentalPageObject._primaryComponent, BUTTON_ACCESSIBILITY_LABEL)`

or

`ButtonExperimentalPageObject.accessibilityLabelIsEqualTo(ButtonExperimentalPageObject.getButtonSelector(ButtonSelector.PrimaryButton), BUTTON_ACCESSIBILITY_LABEL)`

apps/E2E/src/ButtonExperimental/specs/ButtonExperimental.spec.win.ts
```diff
@@ -54,23 +56,29 @@ describe('Experimental Button Functional Testing', async () => {
 
   it('Validate OnClick() callback was fired -> Click', async () => {
     await ButtonExperimentalPageObject.clickComponent();
-    await expect(await ButtonExperimentalPageObject.didOnClickCallbackFire()).toBeTruthy();
+    await expect(await ButtonExperimentalPageObject.didOnClickCallbackFire()).toBeTruthy(
+      `For ButtonExperimental E2E test, the primary button failed to fire an onClick callback with a mouse click.`,
```
samuelfreiberg, line 60: I don't think we need to populate the error message with "For ButtonExperimental E2E test", since that information should be provided already with the test suite

apps/E2E/src/ButtonExperimental/pages/ButtonExperimentalPageObject.ts
lawrencewin: We could probably directly pass in a WebdriverIO element and ditch the selector.

Is there a reason we need to use a Selector enum for things like `sendKey()` or `clickComponent()`?

samuelfreiberg (changes requested):

apps/E2E/src/ButtonExperimental/pages/ButtonExperimentalPageObject.ts
samuelfreiberg: No, we could change those too. All these functions should probably take in a WebDriverIO.Element rather than a page-specific selector

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +72,31 @@ export class BasePage {
...
+    if (expectedValue !== actualValue) {
+      throw new Error(
+        `On ${this._pageName}, a test component should have attribute, ${attributeToEnumName[attribute]} (${attribute}), equal to ${expectedValue}.
+        Instead, ${attributeToEnumName[attribute]} is equal to ${actualValue}.`,
+      );
+    }
+    return true;
+  }
+
+  async accessibilityLabelIsEqualTo(element: Promise<WebdriverIO.Element>, expectedValue: any): Promise<boolean> {
```
samuelfreiberg, line 92: I think we should consolidate these two functions even further. For "x" amount of properties, we shouldn't make "x" amount of methods...

For example, following this pattern, we'd have a `accessibilityLabelIsEqualTo()`, `accessibilityRoleIsEqualTo()`, `expandCollapsestateEqualTo()`, `IsTogglePatternEqualTo()`, etc...

Instead, let's have one method that takes in a Property Name to query for, and the element to query it from.

Such as,

`CompareAttribute(element: Promise<WebDriverIO.Element>, attribute: Attribute, expectedValue: any)`.

Then, in the specs, we can use:
`ButtonExperimentalPageObject.CompareAttribute(ButtonExperimentalPageObject._primaryComponent, Attribute.AccessibilityRole, BUTTON_A11Y_ROLE)`

samuelfreiberg: And `CompareAttribute()` body would be:
`return this._attributeIsEqualTo(await element, attribute, exectedValue)`

samuelfreiberg: Or, we can even re-name `_attributeIsEqualTo` to the `CompareAttribute` and simply use that in spec files

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +72,23 @@ export class BasePage {
   /******************************************************************/
   /**************** UI Element Interaction Methods ******************/
   /******************************************************************/
+
+  /**
+   * Checks to see if an element attribute is strictly equal to an expected value the user passes in.
+   * The advantage to this over testing using .isEqual in a spec is that this throws a detailed error if
+   * the expected and actual values don't match.
+   * This should never be called in a spec. Other PageObjects should use these to make accessibility / attribute testing easier for them. */
```
samuelfreiberg, line 80: Update comment. This is now called in specs

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +72,22 @@ export class BasePage {
...
+
+  /**
+   * Checks to see if an element attribute is strictly equal to an expected value the user passes in.
+   * The advantage to this over testing using .isEqual in a spec is that this throws a detailed error if
+   * the expected and actual values don't match. This should be called for attribute tests in specs. */
+  async compareAttribute(element: Promise<WebdriverIO.Element>, attribute: Attribute, expectedValue: any): Promise<boolean> {
+    const actualValue = await (await element).getAttribute(attribute);
+    if (expectedValue !== actualValue) {
+      throw new Error(
+        `On ${this._pageName}, a test component should have attribute, ${attributeToEnumName[attribute]} (${attribute}), equal to ${expectedValue}.
```
samuelfreiberg, line 84: Nit: One more improvement - I think the `${attributeToEnumName[attribute]} (${attribute})` could be a little confusing to a user. This expands to (for example), `AccessibilityLabel (Name)`. For someone who's not familiar, I would probably be confused at the multiple property names.

Let's instead clarify `${attributeToEnumName[attribute]} (which maps to the Windows '${attribute}' property on the Element)`

samuelfreiberg: Also, we should specify which element is in error.

Is it possible to get the testID/automationID from a WebDriverIO.Element? If so, we can also put that in this error message

apps/E2E/src/common/NavigateAppPage.ts
```diff
@@ -315,6 +315,15 @@ class NavigateAppPage extends BasePage {
   private get experimentalTabsPage() {
     return By(HOMEPAGE_EXPERIMENTAL_TABS_BUTTON);
   }
+
+  get _pageName() {
+    // This is an odd name, but it works well with the error message it is used for.
```
samuelfreiberg, line 320: This comment is a bit confusing. What error message is it used for? Why does it work well?

lyzhan7:
The first time I tried to add an e2e test, this is one of the errors I made: https://github.com/microsoft/fluentui-react-native/pull/2231/commits/ace4497c93c23cee3c1e2a2debd8fb27b2d97ff4

In testPages.ts, for the testPage property of the stroke width description, I accidentally used the wrong constant (STROKEWIDTH_TESTPAGE instead of HOMEPAGE_STROKEWIDTH_BUTTON). Doesn't seem like your PR checks for this, and might not be in scope for this PR but I wonder if there's some error checking we can add to ensure that the right constants are used in the right places. Also thinking maybe the testPage property could be renamed to testPageButton

apps/E2E/src/common/BasePage.ts
```diff
@@ -104,6 +120,15 @@ export class BasePage {
     await (await this._primaryComponent).click();
   }
 
+  /* The goal of click() and keys() is to be generally used across all pageobjects to reduce code repetition in similar methods. */
+  async click(element: Promise<WebdriverIO.Element>): Promise<void> {
+    await (await element).click();
+  }
+
+  async keys(element: Promise<WebdriverIO.Element>, keys: Keys[]): Promise<void> {
```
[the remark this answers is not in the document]
samuelfreiberg: I agree - I think `sendKeys()` makes sense

apps/E2E/src/common/NavigateAppPage.ts
lawrencewin: Each spec calls `NavigateAppPage.waitForPageDisplayed()` at the beginning of its file. If this fails, then this error message gets thrown: 
```ts
timeoutMsg: this._pageName + ' did not render correctly. Please see /errorShots for more information.',
```
which reads "The Fluent Tester did not render correctly. Please see /errorShots for more information.".

apps/E2E/src/common/NativeTesting/pages/NativeTestingPageObject.win.ts
```diff
@@ -41,6 +41,18 @@ class NativeTestingPageObject extends BasePage {
 
     return foundValidButton;
   }
+
```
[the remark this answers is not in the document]
lawrencewin: `win.ts` means the test can be run either on windows or win32, whereas `windows.ts` is only run on windows.

apps/E2E/src/common/BasePage.ts
lawrencewin: Doing a `getAttribute('AutomationId')`  should do the trick.

lawrencewin:
> The first time I tried to add an e2e test, this is one of the errors I made: [ace4497](https://github.com/microsoft/fluentui-react-native/commit/ace4497c93c23cee3c1e2a2debd8fb27b2d97ff4)
> 
> In testPages.ts, for the testPage property of the stroke width description, I accidentally used the wrong constant (STROKEWIDTH_TESTPAGE instead of HOMEPAGE_STROKEWIDTH_BUTTON). I got a pretty generic error `Validate the children of ScrollView are the buttons that navigate to each test page with proper automationId` and spent a long time trying to figure out what was wrong.
> 
> Doesn't seem like your PR checks for this, and might not be in scope for this PR but I wonder if there's some error checking we can add to ensure that the right constants are used in the right places. Also thinking now that maybe the testPage property on the testDescription type should be renamed to testPageButton

This i think is within the scope of PR. Renaming shouldn't be too much, and we can definitely have a better error for the above.

apps/E2E/src/common/BasePage.ts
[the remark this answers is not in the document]
lawrencewin: Did change the name to `sendKeys`. The FocusZonePageObject ended up having a similar method so I ended up refactoring that test as well. Sorry for the change in scope!

apps/E2E/src/common/NavigateAppPage.ts
```diff
@@ -315,6 +315,15 @@ class NavigateAppPage extends BasePage {
...
+    return 'The Fluent Tester';
+  }
```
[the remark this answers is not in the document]
lawrencewin: Mentioned above.

apps/E2E/src/common/NavigateAppPage.ts
lawrencewin: I did make a comment about this in the latest commit.

apps/E2E/src/ButtonExperimental/specs/ButtonExperimental.spec.win.ts
```diff
@@ -22,55 +21,78 @@ describe('Experimental Button Testing Initialization', function () {
 });
 
 describe('Experimental Button Accessibility Testing', async () => {
+  beforeEach(async () => {
+    await ButtonExperimentalPageObject.scrollToTestElement(await ButtonExperimentalPageObject._primaryComponent);
```
ksiler, line 25: I don't think you need to specify `await ButtonExperimentalPageObject._primaryComponent` since `scrollToTestElement` will scroll to the _primaryComponent if no component is specified.

apps/E2E/src/common/BasePage.ts
```diff
@@ -104,6 +124,15 @@ export class BasePage {
     await (await this._primaryComponent).click();
   }
 
+  /* The goal of click() and keys() is to be generally used across all pageobjects to reduce code repetition in similar methods. */
```
ksiler, line 127: nit change name of method in comment from `keys()` to `sendKeys()`

apps/E2E/src/common/NavigateAppPage.ts
samuelfreiberg: Let's rename it to "FluentTester Base Page", since that's what we're technically checking for with `get _testPage()`.

Then, the comment can simply read "This is the first (main) page displayed after launching FluentTester". This makes more sense and translates better

apps/E2E/src/common/BasePage.ts
```diff
@@ -104,6 +124,15 @@ export class BasePage {
     await (await this._primaryComponent).click();
   }
 
+  /* The goal of click() and sendKeys() is to be generally used across all pageobjects to reduce code repetition in similar methods. */
+  async click(element: Promise<WebdriverIO.Element>): Promise<void> {
+    await (await element).click();
+  }
+
+  async sendKeys(element: Promise<WebdriverIO.Element>, keys: Keys[]): Promise<void> {
```
samuelfreiberg, line 132: Do we need `: Promise<void>`?

samuelfreiberg:
> The first time I tried to add an e2e test, this is one of the errors I made: [ace4497](https://github.com/microsoft/fluentui-react-native/commit/ace4497c93c23cee3c1e2a2debd8fb27b2d97ff4)
> 
> In testPages.ts, for the testPage property of the stroke width description, I accidentally used the wrong constant (STROKEWIDTH_TESTPAGE instead of HOMEPAGE_STROKEWIDTH_BUTTON). I got a pretty generic error `Validate the children of ScrollView are the buttons that navigate to each test page with proper automationId` and spent a long time trying to figure out what was wrong.
> 
> Doesn't seem like your PR checks for this, and might not be in scope for this PR but I wonder if there's some error checking we can add to ensure that the right constants are used in the right places. Also thinking now that maybe the testPage property on the testDescription type should be renamed to testPageButton

This is a great point - thank you. This is the first of a couple PRs aiming to improve the debugging experience of an E2E failure. We'll definitely improve on that as more PRs come out

apps/E2E/src/common/BasePage.ts
lawrencewin: It isn't necessary, but is there anything wrong with explicitly stating the return of both functions in this case?

ksiler (approved):

apps/E2E/src/FocusZone/pages/FocusZonePageObject.ts
```diff
@@ -24,20 +23,6 @@ export const enum GridFocusZoneOption {
...
-  Three,
-  Four,
-  Five,
-  Six,
-  Seven,
-  Eight,
-  Nine,
-  After,
-}
-
```
lyzhan7, line 40: What was the reasoning behind removing this enum and replacing it with a number?

lawrencewin: I think that moving away from selector enums and using WebdriverIO in tests would simplify test code, which is why I removed this enum. 

Because of the number of test components for FocusZone, having an enum might be a better idea in this case. @samuelfreiberg?

lyzhan7: My concern with this change is that the enum seemed better for type safety - ex. with number I think you could pass in a decimal or a negative number and everything will still compile. Also maybe there are only 10(?) grid buttons, so passing in 11 or 12 would cause some error

lawrencewin: Re-added an enum for typesafety.

samuelfreiberg (approved):

PR #2413, opened 2022-12-12
[Separator - Android] Tokenization , InsetSpacing and TesterApp and Platform utils addition

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [x] win32 (Office) (Just the tester app : migrated components used from Stock RN to FURN)
- [ ] windows
- [x] android

### Description of changes

This PR targets to tokenize and adding of insetSpacing to the Separator component for Android.
packages/components/Separator/src/* - General separation/cleaning of files and Android Specific changes
apps/fluent-tester/* - tester app changes.

### Verification

Visual Verified for Win32. 

| Light                                       | Dark                                      |
| Screenshot_1670843312 | Screenshot_1670843317 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
FYI that we haven't done any V1 work for this component yet. It's already on the new compose framework probably because Jason ported it as a proof of concept for it, but we haven't touched it otherwise.

rohanprasadofficial:
> FYI that we haven't done any V1 work for this component yet. It's already on the new compose framework probably because it was ported as a proof of concept for it, but we haven't touched it otherwise.

@rurikoaraki , Thanks for the info.

Saadnajmi:
I think the design is shared with iOS 100% for many of the mobile controls. Given that, could we make these changes on iOS too? My reasoning being: I'm always hesitant when I see changes to a component for one specific platform that probably apply to other ones: In this case. mobile overall. It might be more natural in code to have checks that look like "isMobile", rather than "isAndroid". I think that would also be more clear to future engineers reading this code that the difference in design / spec is mobile vs desktop, not Android vs the other 4 platforms. Something like an engineering plan would me help to understand the scope of work planned to be done for various components to better understand the right decision.

rohanprasadofficial:
> I think the design is shared with iOS 100% for many of the mobile controls. Given that, could we make these changes on iOS too? My reasoning being: I'm always hesitant when I see changes to a component for one specific platform that probably apply to other ones: In this case. mobile overall. It might be more natural in code to have checks that look like "isMobile", rather than "isAndroid". I think that would also be more clear to future engineers reading this code that the difference in design / spec is mobile vs desktop, not Android vs the other 4 platforms. Something like an engineering plan would me help to understand the scope of work planned to be done for various components to better understand the right decision.

That definitely make sense @Saadnajmi , Looking into design it almost same for Android and iOS (mobile platforms basically) , I will convert it to 'isMobile' flags and test them in iOS as well to make sure atleast it renders , maybe tokenization might be done by specific platform engineers after that.

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.tsx
```diff
@@ -13,26 +15,66 @@ const SeparatorMainTest: React.FunctionComponent = () => {
...
-    name: 'Basic Button',
-    testID: SEPARATOR_TESTPAGE,
-    component: SeparatorMainTest,
-  },
+  ...Platform.select({
+    android: [
+      {
+        name: 'Basic Mobile Seperator',
+        testID: SEPARATOR_TESTPAGE,
+        component: SeparatorAndroid,
```
rurikoaraki, line 67: My personal preference would be to split this test component into its own file and fork the file between Android and non-Android platforms. It reduces the platform checks within the tests and separating the tests out from the main page would be a cleaner organization.

packages/components/Separator/src/Separator.types.ts
```diff
@@ -1,11 +1,26 @@
...
    * @defaultValue 'false'
    */
   vertical?: boolean;
+
+  /**
+   * Inset Spacing in pixels.
+   *
+   * Inset Spacing is restricted to a limited set of supported values recommended for most uses (see `SeparatorInsetSpacing`) and
+   * based on design guidelines for the Separator control.
+   * @defaultvalue 0
```
rurikoaraki, line 21: Please mark this prop as mobile only

packages/components/Separator/src/Separator.types.ts
```diff
@@ -1,11 +1,26 @@
 import { ViewProps, ColorValue } from 'react-native';
 
-interface SeparatorPropTokens {
+export const separatorName = 'Separator';
+
+export const SeparatorInsetSpacings = [0, 16, 56, 68, 72, 108] as const;
```
rurikoaraki, line 5: Where are these values coming from? Are they supposed to be based on some sort of design token?

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.tsx
rohanprasadofficial: Separated to Platform files.  
Just thinking on the platform checks via Platform.select() API , this could be useful some redundant code but also separating helps on the cleaning and making changes and less testing effort. 

But we can have a discussion on this once Ankita is back , As we're also planning to come of common way to handle this.

packages/components/Separator/src/Separator.types.ts
rohanprasadofficial: @rurikoaraki , This is directly coming from Figma design from Separator Design for Mobile. 
There will be no tokens for this as :
1. There is no ramp that supports all of this values.
2. It is values of the props that user able to pass to 'insetSpacing' - For eg : In case of the Avatar as well :  
export const AvatarSizes = [20, 24, 28, 32, 36, 40, 48, 56, 64, 72, 96, 120] as const; 
https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/packages/components/Avatar/src/Avatar.types.ts#L24 

There're tokens as well but for props and show typings , we have use it like this. 
One more example would be https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/packages/components/Checkbox/src/Checkbox.types.ts#L129

packages/components/Separator/src/Separator.types.ts
rohanprasadofficial: Added.

rohanprasadofficial:
@rurikoaraki 
Currently the tester app is modified for Android/iOS and look like this. 

| Dark                                       | Light                                      |
| Screenshot_1671084750 | Screenshot_1671084740 |

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
-import { Text } from 'react-native';
-import { Button, Separator } from '@fluentui/react-native';
+import { Separator } from '@fluentui-react-native/separator';
+import { Button } from '@fluentui-react-native/experimental-button';
```
rurikoaraki, line 3: Please use ButtonV1 from @fluentui-react-native/button instead, experimental-button is deprecated.

packages/utils/platform-tools/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "lint": "fluentui-scripts eslint",
+    "test": "fluentui-scripts jest",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "prettier": "fluentui-scripts prettier",
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/utils/styling"
```
rurikoaraki, line 26: Please update

packages/utils/platform-tools/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: ['@fluentui-react-native/eslint-config-rules'],
+  rules: {
+    '@typescript-eslint/no-var-requires': 0,
```
rurikoaraki, line 4: Why is turning this rule off required for the whole package? If needed I would rather this be localized to the instance it's needed for instead of for everything

packages/components/Separator/src/Separator.types.ts
rurikoaraki: nit: I've seen `@platform` tags for these sorts of things before.

packages/components/Separator/src/SeparatorTokens.android.tsx
```diff
@@ -0,0 +1,11 @@
+import { Theme } from '@fluentui-react-native/framework';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import { TokenSettings } from '@fluentui-react-native/use-styling';
+import { SeparatorTokens } from './Separator.types';
+
+export const defaultSeparatorTokens: TokenSettings<SeparatorTokens, Theme> = (t: Theme) =>
```
rurikoaraki, line 6: Is this change needed for iOS as well?

rurikoaraki:
Since this PR sets up the test for both Android and iOS, it would be good to make sure that an iOS engineer takes a look at this PR before merging.

packages/components/Separator/src/SeparatorTokens.android.tsx
rurikoaraki: Okay it seems like tokenization is up to individual platforms to complete, so I'm guessing not

packages/components/Separator/src/Separator.types.ts
rohanprasadofficial: Found  @platform tags , thanks I didn't knew it :)

rohanprasadofficial: Added.

packages/utils/platform-tools/.eslintrc.js
rohanprasadofficial: Removed the rule.

packages/utils/platform-tools/package.json
rohanprasadofficial: Updated.

rohanprasadofficial:
@lyzhan7 , can you have a look on this from iOS side?

packages/utils/platform-tools/src/Platform.ts
```diff
@@ -0,0 +1,9 @@
+import { Platform } from 'react-native';
+
+export const isMobile = Platform.OS === 'android' || Platform.OS === 'ios';
+export const isDesktop = Platform.OS === 'macos' || Platform.OS === 'windows' || Platform.OS === ('win32' as any);
+export const isAndroid = Platform.OS === 'android';
+export const isIOS = Platform.OS === 'ios';
+export const isMac = Platform.OS === 'macos';
+export const isWin32 = Platform.OS === ('win32' as any);
+export const isWindows = Platform.OS === 'windows';
```
Saadnajmi, line 9: I was taught early in my career to avoid `utils` objects / classes / etc as they become a catch all. I can see the usefulness of a platform-tools style package. In its current state, I'm not sure these are all that useful. For example: `isMobile` includes iPad where we might want the desktop experience. So I've actually written that check with an extra `Platform.isPad` in it as well. But then iPad has different size classes if you take an app and slide it over another app. In that case, you want the mobile experience again. So then we had to write an extra (hacky) hook to guess that. I could also see this being useful to say, have a helper function that returns components as null if on an unsupported platform (I'm still thinking about that one, I haven't thought how that may best look). In its current form of just doing a bunch of single line if checks, I'm not a fan of it. 

I think I'd prefer the creation of a package like this in a separate PR that _just_ creates the package, and refactors platform checks across the repo to accommodate.

packages/components/Separator/src/Separator.types.ts
Saadnajmi: I _think_ it might be better if the platforms are space delineated rather than comma delineated? It would be easy to check by just hovering over this prop in VS Code. Normally the platform tag only has one platform, so I haven't figured out the best way to list multiple platforms. I went with spaces earlier, would like to know if something else is better.

packages/components/Separator/src/Separator.types.ts
```diff
@@ -1,11 +1,27 @@
...
+
+export interface SeparatorPropTokens {
   /**
    * Specifies whether the separator is vertical or horizontal. If true, then it's vertical.
    * @defaultValue 'false'
    */
   vertical?: boolean;
+
+  /**
+   * Inset Spacing in pixels.
```
Saadnajmi, line 17: I don't think. this is in pixels in iOS. Instead it would be on "points", (1 pixel at 1x DPI, 3 pixels in 3X DPI...). I think that might be true on Android as well. Worth confirming?

packages/utils/platform-tools/src/Platform.ts
Saadnajmi: Publishing a new package also tends to break our somewhat fragile nom pipeline for a bit. So extra reason to put it in its own PR.

packages/components/Separator/src/Separator.types.ts
Saadnajmi: On Android, it's probably DP https://developer.android.com/training/multiscreen/screendensities

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.tsx
rohanprasadofficial: updated.

lyzhan7:
Simulator Screen Shot - iPhone 14 Pro - 2022-12-19 at 20 57 19

Just checked out your PR locally and tried to run it but this error pops up. Usually when I've seen this error, running yarn usually fixes it. I tried completely resetting my repo - i.e. git clean -xfd && yarn && yarn build, but still see the error. It might be an issue locally, will take another look tomorrow

packages/utils/platform-tools/src/Platform.ts
rohanprasadofficial: I have copied the changes locally for different branch and will raise a different PR for it.  
For now, I will have a local check in the file it itself.

packages/components/Separator/src/Separator.types.ts
rohanprasadofficial: Added space delineated for now.

packages/components/Separator/src/Separator.types.ts
rohanprasadofficial: Changed. 
Android - dp 
iOS - points

rohanprasadofficial:
> Just checked out your PR locally and tried to run it but this error pops up. Usually when I've seen this error, running yarn usually fixes it. I tried completely resetting my repo - i.e. git clean -xfd && yarn && yarn build, but still see the error. It might be an issue locally, will take another look tomorrow
> 
> Screenshot 2022-12-19 at 9 41 41 PM

@lyzhan7 , I guess this some issue with yarn and yarn build. maybe the new package didn't come properly. 
For now, based on @Saadnajmi's suggestion, will the new package as a separate PR, for now I have removed it, you can try now it should work, you can just check the Separator figma file, it resembles with Android, so just check that on iOS if it works fine.

rurikoaraki (approved):

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.mobile.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
```
Saadnajmi, line 1: What's the reason we want to separate out the test page to be mobile specific? If possible, could we move the test page to the same for and have a "mobile only" section? 

Reasoning: Less fracturing might mean it's easier to spot mobile vs desktop differences. Happy to see if I'm wrong and there's another way though.

rurikoaraki: I've personally been pushing for test sections to be separated per platform in their implementations differ (or one platform doesn't support that feature, or something) to avoid a bunch of if (Platform) checks inside the test pages itself. I feel like it makes the test pages messy to have those checks and it also means a bunch of code that isn't used is included in the bundle, although that's less important in the tester.

Probably worth getting on the same page about how to handle these differences as part of @ankraj12 's plans

ankraj12: For readability purpose - I am inclined towards separate files if there are a messy amount of "if platform " checks inline. If it's just a couple of clean forks, it could live in same file I guess. I would take a note of this in documentation.

rohanprasadofficial: So for now , we're going it platform level forks for test app ?

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.tsx
rohanprasadofficial: Continuing on another comment, hence resolving this.

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.mobile.tsx
Saadnajmi: I'll go with what the rest choose. So please don't block on me and choose what feels best.

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/Separator/SeparatorTest.mobile.tsx
rohanprasadofficial: Let's have a internal discussion on the doc Ankita is preparing and have a common decision and document it for everyone to refer.

Issue #2264, opened 2022-10-21
`accessible=true` prop on Fluent Tester root view breaks element querying on iOS E2E tests


lawrencewin:
### Willing to submit a PR to fix?

- [ ] I am willing to submit a PR to fix

### Requested priority

Normal

### Products/applications affected

CI fails without disabling the accessible prop on ios for this PR: https://github.com/microsoft/fluentui-react-native/pull/2226 

### Package version(s)
```
- yarn 1.22.19
- npm 8.0.0
- react-native ^0.68.0
- react-native-macos ^0.68.0
- appium 2.0.0-beta.41
- appium-xcuitest-driver 4.11.1
```

### OS version(s)

iOS 15.5

### Platform

- [X] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Xcode version

xCode 14.0.1

### Please provide a reproduction of the bug

1. Clone this PR: https://github.com/microsoft/fluentui-react-native/pull/2226
2. In `apps/fluent-tester/src/FluentTester.tsx`, change the `accessible` prop on the `View` with `testID={ROOT_VIEW}` from `Platform.OS === 'ios'` to `true`.
```tsx
  return (
    // TODO: Figure out why making this view accessible breaks element querying on iOS.
    <View accessible={Platform.OS !== 'ios'} testID={ROOT_VIEW} style={commonTestStyles.flex}>
      {Platform.OS === ('win32' as any) ? (
        <FocusTrapZone style={themedStyles.root}>
          <TesterContent />
        </FocusTrapZone>
      ) : (
        <RootView style={themedStyles.root}>
          <TesterContent />
        </RootView>
      )}
    </View>
  );

```
3. From `apps/fluent-tester`, run `yarn ios` to build the iOS app. Then, once the app is running, run `yarn e2etest:ios` to run the E2E test suite. Watch the tests fail. (if you don't have a mac, just run this in the CI)

### Actual behavior

The compiled native app returns this view hierarchy:
![ca73c893-602a-4fe6-9094-b335db00b42d](https://user-images.githubusercontent.com/15683103/197255516-e28b9624-0161-4d3f-a889-8ea4c1ff58c7.jpg)

[...]

PR #2418, opened 2022-12-13
[E2E tests] Update comment to explain why iOS FluentTester root view cannot have accessible prop set to true

lyzhan7:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Issue https://github.com/microsoft/fluentui-react-native/issues/2264 was opened because setting the accessible prop to true on Fluent Tester root view broke element querying on iOS E2E tests.

After some [investigation](https://github.com/microsoft/fluentui-react-native/issues/2264#issuecomment-1343723430), it seems like the accessible prop on the root view should actually be set to false since iOS accessibility works differently from other platforms - there are no nested accessibility elements on iOS, if a parent view is an accessibility element than its child views cannot be accessibility elements. If the root view is an accessibility element, the whole screen is treated as a single accessibility element

### Verification

The CI will pass if all the e2e tests pass.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester.tsx
@@ -205,7 +205,7 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
     // TODO: Setting the testID should disable React Native's layout-only view removal optimization.
     // However, for win32, the view removal still happens even though the testID is being set on this view.
     // As a temporary patch, set the view to be accessible on win32 to ensure that the view is not removed.
-    <RootView style={themedStyles.root} accessible={Platform.OS === 'macos'} testID={ROOT_VIEW}>
+    <RootView style={themedStyles.root} collapsable={false} testID={ROOT_VIEW}>
       <Header enableSinglePaneView={enableSinglePaneView} enableBackButtonIOS={!onTestListView} onBackButtonPressedIOS={onBackPress} />
       <HeaderSeparator />
       <View style={fluentTesterStyles.testRoot}>
--- a/apps/fluent-tester/src/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester.tsx
@@ -205,7 +205,7 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
     // TODO: Setting the testID should disable React Native's layout-only view removal optimization.
     // However, for win32, the view removal still happens even though the testID is being set on this view.
     // As a temporary patch, set the view to be accessible on win32 to ensure that the view is not removed.
-    <RootView style={themedStyles.root} accessible={Platform.OS === ('win32' as any)} testID={ROOT_VIEW}>
+    <RootView style={themedStyles.root} accessible={Platform.OS === 'macos'} testID={ROOT_VIEW}>
       <Header enableSinglePaneView={enableSinglePaneView} enableBackButtonIOS={!onTestListView} onBackButtonPressedIOS={onBackPress} />
       <HeaderSeparator />
       <View style={fluentTesterStyles.testRoot}>
--- a/apps/fluent-tester/src/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester.tsx
@@ -202,8 +202,10 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
   };
 
   return (
-    // TODO: Figure out why making this view accessible breaks element querying on iOS.
-    <RootView style={themedStyles.root} accessible={Platform.OS !== 'ios'} testID={ROOT_VIEW}>
+    // TODO: Setting the testID should disable React Native's layout-only view removal optimization.
+    // However, for win32, the view removal still happens even though the testID is being set on this view.
+    // As a temporary patch, set the view to be accessible on win32 to ensure that the view is not removed.
+    <RootView style={themedStyles.root} accessible={Platform.OS === ('win32' as any)} testID={ROOT_VIEW}>
       <Header enableSinglePaneView={enableSinglePaneView} enableBackButtonIOS={!onTestListView} onBackButtonPressedIOS={onBackPress} />
       <HeaderSeparator />
       <View style={fluentTesterStyles.testRoot}>
--- a/apps/fluent-tester/src/FluentTester.tsx
+++ b/apps/fluent-tester/src/FluentTester.tsx
@@ -202,11 +202,8 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props: 
   };
 
   return (
-    // Setting the testID should disable React Native's layout-only view removal optimization, and keep the view in the native hierarchy.
-    // However, for win32 and macOS, the view removal still seems to happen even though the testID is being set on this view.
-    // As a temporary patch, set the view to be accessible on win32 and macOS to ensure that the view is not removed.
-    // TODO: remove the need for this temporary patch
-    <RootView style={themedStyles.root} accessible={Platform.OS === ('win32' as any) || Platform.OS === 'macos'} testID={ROOT_VIEW}>
+    // On iOS, the accessible prop must be set to false because iOS does not support nested accessibility elements
+    <RootView style={themedStyles.root} accessible={Platform.OS !== 'ios'} testID={ROOT_VIEW}>
       <Header enableSinglePaneView={enableSinglePaneView} enableBackButtonIOS={!onTestListView} onBackButtonPressedIOS={onBackPress} />
       <HeaderSeparator />
       <View style={fluentTesterStyles.testRoot}>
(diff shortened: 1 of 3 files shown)


apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -202,8 +202,10 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
   };
 
   return (
-    // TODO: Figure out why making this view accessible breaks element querying on iOS.
-    <RootView style={themedStyles.root} accessible={Platform.OS !== 'ios'} testID={ROOT_VIEW}>
+    // TODO: Setting the testID should disable React Native's layout-only view removal optimization.
+    // However, for win32, the view removal still happens even though the testID is being set on this view.
+    // As a temporary patch, set the view to be accessible on win32 to ensure that the view is not removed.
+    <RootView style={themedStyles.root} accessible={Platform.OS === ('win32' as any)} testID={ROOT_VIEW}>
```
Saadnajmi, line 208: I believe that whole "layout-only view removal" is actually controlled by the prop `collapsable`, at least In React Native iOS/Android. Maybe we can just set that to be true? @lawrencewin FYI

Saadnajmi: There was a conference talk / blog about it somewhere. I can try and dig it up if it helps.

Saadnajmi: https://reactnative.dev/architecture/view-flattening
https://reactnative.dev/docs/view#collapsable-android

This was an optimization initially made in Fabric for Android, that they then brought to iOS. My guess is this is what the win32 code is referring to, and that the website docs are slightly out of date (I don't think it's Android only).

lyzhan7: From a very brief investigation into "layout-only view removal"
- couldn't find anything in the react native docs
- found this on stack overflow, take with a grain of salt

> When RN creates the native views from the shadow nodes tree - it performs some optimizations. Views that do not actually show on screen (don't draw anything or just used in JSX to contain and layout their children) can be removed when constructing the native hierarchy. This is why they're called "layout-only" views.

- there are other IDs that apparently also disable the 'layout-only view removal' optimization for this view - [id](https://reactnative.dev/docs/next/view#id) and [nativeid](https://reactnative.dev/docs/next/view#nativeid)

lyzhan7: @Saadnajmi sorry I sent my comment before I saw your second and third ones. Thanks for finding that! I can try setting collapsable to be false on this view

apps/fluent-tester/src/FluentTester.tsx
```diff
@@ -202,8 +202,10 @@ export const FluentTester: React.FunctionComponent<FluentTesterProps> = (props:
   };
 
   return (
-    // TODO: Figure out why making this view accessible breaks element querying on iOS.
-    <RootView style={themedStyles.root} accessible={Platform.OS !== 'ios'} testID={ROOT_VIEW}>
+    // TODO: Setting the testID should disable React Native's layout-only view removal optimization.
+    // However, for win32, the view removal still happens even though the testID is being set on this view.
+    // As a temporary patch, set the view to be accessible on win32 to ensure that the view is not removed.
+    <RootView style={themedStyles.root} accessible={Platform.OS === 'macos'} testID={ROOT_VIEW}>
```
Saadnajmi, line 208: why macOS?

lyzhan7:
Based on the last few commits it seems like both macOS and win32 (but not windows) require accessibility=true for the e2e tests to pass. Error messages below for reference:

| macOS error message | win32 error message |
| - | - |
| Screenshot 2022-12-13 at 4 55 08 PM | Screenshot 2022-12-13 at 4 52 06 PM |

Basically complaining about the same thing, I'm guessing they look a little different because the native macOS testing framework is different from the native win32 testing framework

lyzhan7:
> > 
> 
> It looks like some part of our tests is trying to look up the root view, which can only be done if it's accessible?

So at first it seemed like the root issue of this problem was that win32 (and I think macOS too) were optimizing away the root view when it weren't supposed to be. I think that's why the accessible=true prop was added to the root view, it was another way to ensure the view was there. However, I actually think that setting the accessible prop on the root view to be true is the wrong thing to do for all platforms, but only caused issues on iOS. 

Just checking now to see if setting collasable=false would also work as a temporary solution around what seems to be a rn-windows/rn-macos issue, but without having issues on iOS.

And another thought, I wonder if we need to be looking up the root view in the first place? Maybe rn-macOS/rn-win32 are doing the right thing and the e2e tests just shouldn't be trying to access the root view. Looking at the logs, it seemed like the root view was mainly used just to find other elements, but I wonder if we actually need to go through the root view or if we can just query for the elements directly

lawrencewin:
> > > 
> > 
> > 
> > It looks like some part of our tests is trying to look up the root view, which can only be done if it's accessible?
> 
> So at first it seemed like the root issue of this problem was that win32 (and I think macOS too) were optimizing away the root view when it wasn't supposed to be. I think that's why the accessible=true prop was added to the root view, it was another way to ensure the view was there. However, I actually think that setting the accessible prop on the root view to be true is the wrong thing to do for all platforms, but only caused issues on iOS.
> 
> Just checking now to see if setting collasable=false would also work as a temporary solution around what seems to be a rn-windows/rn-macos issue, but without having issues on iOS.
> 
> And another thought, I wonder if we need to be looking up the root view in the first place? Maybe rn-macOS/rn-win32 are doing the right thing and the e2e tests just shouldn't be trying to access the root view. Looking at the logs, it seemed like the root view was mainly used just to find other elements, but I wonder if we actually need to go through the root view or if we can just query for the elements directly

For the standalone FURN tester, looking up the root-view isn't too necessary because the UI tree ends up being super shallow on windows. However, this strategy of finding the root first was built for testing in an internal testing environment where the UI tree ends up being significantly larger - meaning these chained queries save time. 

I think the FURN tests in the internal testing environment is on the backburner for now so you might be able to get away with removing the chained query, but maybe consult @samuelfreiberg first.

samuelfreiberg:
A couple comments:

1. Is the current implementation of setting `accessible=true` on the root view for every platform other than iOS causing any trouble or noise? 
2. There are scenarios that exist outside of this repo where we use these same exact FluentTester bundle + E2E tests, but host the FluentTester in Office Applications (such as Word, PowerPoint, etc). If you can imagine, the UI tree on these apps is HUGE. The cost it takes to query a specific element in the FluentTester app within Word is humongous (starting at Word's root and going all the way down the tree until we find the element we want). The performance is very bad. This is why we need to find a root parent in the FluentTester, save this value, and then query elements using that root (so rather than starting at Word's root everytime, we start at the root of the FluentTester)

Is there any con to leaving it as-is?

lyzhan7:
@lawrencewin Thanks for explaining why we want to find the root view in the first place Larry - that makes a lot of sense. 

@samuelfreiberg 
1) No, it doesn't seem to be causing issues on the platforms besides iOS
2) That makes a lot of sense, now that I understand the benefit of finding the root view first I agree we should continue doing that. However, why is setting accessible=true necessary for that to happen? If you look at one of the recent iOS e2e logs on this PR (ex. [this one](https://dev.azure.com/ms/ui-fabric-react-native/_build/results?buildId=407873&view=logs&j=d9bb5e1e-64f3-5345-4184-6fa41163601d&t=25dbb2ee-2046-5e2a-3835-66f0c8b7964b)), the accessible prop on the root view was not set to true, but we're still able to find the root view/get the performance benefits of first finding the root view. I'm new to reading the logs though, so if that doesn't seem to be the case to you let me know. 

The root issue seems to be that rn-macos/rn-win32 aren't respecting the testID; even though the testID for the root view is set, there's still some optimization happening that hides the view from the test. Setting accessibility=true is a workaround for this issue, but ideally we shouldn't have to need this workaround

lyzhan7:
Update: talked to @Saadnajmi about the issue I opened on macOS and realized that the react native documentation is going to be biased towards iOS/android, so what the documentation says might not be true for macOS/win32. Setting accessible=true might be the right thing to do for those platforms - iOS seems to be an outlier in how accessibility works, i.e. there are no nested accessibility elements (if the parent is an accessibility element its children can't be accessibility elements). 

Also the "layout-only view removal" optimization was added in Fabric, so for RN-macOS that would not make sense as the reason why the views are not being found by the e2e tests.

rurikoaraki:
Based on my fuzzy memory I'm guessing that, at least on Windows, the test fails when accessibility is not set to true because the testid we get is stored on the accessibility tree. If the element is not accessible, then it's not put in the accessibility tree, and so when the test tries to look for it via the accessibility tree, it fails to find it.

Using the accessibility tree (aka the UI Automation tree) to test UI is a fairly common practice in Windows, which means that sometimes we need to make things accessible to make sure they end up in a testable state via automation.

Saadnajmi (approved):

PR #2420, opened 2022-12-13
Update list of exports of main FURN package

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

@fluentui/react-native package is missing a few outputs from its input packages. Updating the index file.
Would be nice to have a linter check for this or something now that we can't export * but not within the scope of this PR.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

lyzhan7:
I've never touched this file before/knew it existed - while there isn't a linter check for now, when should we be updating this file? Is this something to consider in every PR, or just something when a component reaches a certain stage? Ex. should I add the shadow component to this file? Does that mean that partners don't have access to the shadow component until I add it to this file?

PR #2423, opened 2022-12-14
Overhaul Tests + Errors for Avatar and Callout

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

- Avatar
- Callout

#### Removal of Selectors

One big change to the tests across the board is the removal of Selector enums within tests. Previously, Selectors were used to choose which element was to be clicked, get an attribute from, etc. However, because many tests required different selectors from the base `ComponentSelector` in BasePage.ts to select test elements, this meant that different Page Objects ended up implementing the same methods (`sendKey()`, `click()`, and other attribute testing methods). 

In removing selectors, we can re-use methods for doing most things a test requires implemented in BasePage.ts and pass WebdriverIO elements directly (see [`compareAttribute()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L80-L93),[ `click()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L127-L130), and [`sendKeys()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L132-L134)). This way, future Page Objects remain light and only implement getters or other methods needed for the specific test.

This doesn't mean that Selectors can't be used. Taking the [FocusZone test](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/FocusZone/pages/FocusZonePageObject.ts#L18-L28), the large amount of elements used within the test let us use a Selector Enum to pick out specific elements to pass in. But for most cases, we can rely on a small amount of getters and the reusable BasePage methods for creating a test.

### Verification

Tests pass in the CI and locally. 

image

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
Let's clarify in the description why we are getting rid of the Selector enums on each PageObject. It's a pretty big pattern change improvement and it's important to note

lawrencewin:
> Let's clarify in the description why we are getting rid of the Selector enums on each PageObject. It's a pretty big pattern change improvement and it's important to note

Just updated my description!

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -52,31 +68,36 @@ describe('Checkbox Functional Testing', () => {
...
     await CheckboxPageObject.clickComponent();
 
     /* Validate the Checkbox is toggled OFF */
-    await expect(await CheckboxPageObject.isCheckboxChecked()).toBeFalsy();
+    await expect(await CheckboxPageObject.isCheckboxChecked()).toBeFalsy('The primary checkbox should have been toggled off.');
+
     await expect(await CheckboxPageObject.didAssertPopup()).toBeFalsy(CheckboxPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Click the "SPACE" on a Checkbox and verify it toggles', async () => {
```
ksiler, line 92: I think this should be renamed "Press "Space" on a Checkbox...." instead of "Click the..."

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -52,31 +68,36 @@ describe('Checkbox Functional Testing', () => {
...
     /* Presses the "space bar" to select the Checkbox */
-    await CheckboxPageObject.sendKey(CheckboxSelector.Primary, Keys.SPACE);
+    await CheckboxPageObject.sendKeys(CheckboxPageObject._primaryComponent, [Keys.SPACE]);
     await CheckboxPageObject.waitForCheckboxChecked(PAGE_TIMEOUT);
 
     /* Validate the Checkbox is selected */
-    await expect(await CheckboxPageObject.isCheckboxChecked()).toBeTruthy();
+    await expect(await CheckboxPageObject.isCheckboxChecked()).toBeTruthy(
+      "Primary checkbox failed to fire onChange() callback via 'Space' press.",
```
ksiler, line 99: This error message is inaccurate since we're testing whether or not the checkbox got toggled (not whether or not the onChange callback was called). 

Could you change this test for pressing "Space" to mimic the above test for click? (i.e. keep this test in but also add a test to verify the onChange callback was called and rename the test to "Press "Space" on a Checkbox -> Validate it toggles correctly AND calls the user's onChange")

apps/E2E/src/Checkbox/specs/Checkbox.spec.windows.ts
```diff
@@ -1,8 +1,7 @@
 import NavigateAppPage from '../../common/NavigateAppPage';
 import CheckboxPageObject from '../pages/CheckboxPageObject';
```
ksiler, line 2: Why do we have two test specs for ".win.ts" and ".windows.ts"? Can win32 and windows not use the same test spec?

apps/E2E/src/CheckboxExperimental/specs/ExperimentalCheckbox.spec.win.ts
```diff
@@ -65,31 +75,38 @@ describe('Checkbox Functional Testing', () => {
...
   it('Click the "SPACE" on a Checkbox and verify it toggles', async () => {
     /* Presses the "space bar" to select the Checkbox */
-    await ExperimentalCheckboxPageObject.sendKey(ExperimentalCheckboxSelector.Primary, Keys.SPACE);
+    await ExperimentalCheckboxPageObject.sendKeys(ExperimentalCheckboxPageObject._primaryComponent, [Keys.SPACE]);
     await ExperimentalCheckboxPageObject.waitForCheckboxChecked(PAGE_TIMEOUT);
 
     /* Validate the Checkbox is selected */
-    await expect(await ExperimentalCheckboxPageObject.isCheckboxChecked()).toBeTruthy();
+    await expect(await ExperimentalCheckboxPageObject.isCheckboxChecked()).toBeTruthy(
+      "Primary checkbox failed to fire onChange() callback via 'Space' press.",
```
ksiler, line 108: same comment from Checkbox.spec.win.ts

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -35,29 +30,46 @@ describe('Avatar Accessibility Testing', () => {
...
+    await expect(
+      await AvatarPageObject.compareAttribute(AvatarPageObject._primaryComponent, Attribute.AccessibilityHint, AVATAR_ACCESSIBILITY_HINT),
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Validate accessibilityRole', async () => {
-    await expect(await AvatarPageObject.getPrimaryComponentAttribute(ACCESSIBILITY_ROLE_ATTRIBUTE)).toEqual(ACCESSIBILITY_ROLE_LINK);
+    await expect(
```
samuelfreiberg, line 61: These two tests are a bit confusing - both testing accessibility role, but two different expected values? What's different between the two components and why are they different?

Let's clarify the test descriptions here

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -53,24 +69,32 @@ describe('Button Functional Testing', () => {
   });
 
   it('Validate OnClick() callback was fired -> Click', async () => {
-    await ButtonPageObject.clickComponent();
-    await expect(await ButtonPageObject.didOnClickCallbackFire()).toBeTruthy();
+    await ButtonPageObject.click(ButtonPageObject._primaryComponent);
+    await expect(await ButtonPageObject.waitForOnClickCallbackToFire()).toBeTruthy(
+      'Mouse click failed to fire the onClick callback for the primary button.',
```
samuelfreiberg, line 74: Nit: Let's re-word this a bit:

"Clicking on the primary button failed to fire the onClick() callback"

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -53,24 +69,32 @@ describe('Button Functional Testing', () => {
...
 
     await ButtonPageObject.clickComponent(); // Reset Button State
   });
 
   it('Validate OnClick() callback was fired -> Type "Enter"', async () => {
-    await ButtonPageObject.sendKey(ButtonSelector.PrimaryButton, Keys.ENTER);
-    await expect(await ButtonPageObject.didOnClickCallbackFire()).toBeTruthy();
+    await ButtonPageObject.sendKeys(ButtonPageObject._primaryComponent, [Keys.ENTER]);
+    await expect(await ButtonPageObject.waitForOnClickCallbackToFire()).toBeTruthy(
+      "'Enter' keypress failed to fire the onClick callback for the primary button.",
```
samuelfreiberg, line 85: Same as above:

"Pressing the "Enter" key on the primary button failed to fire the onClick() callback"

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -53,24 +69,32 @@ describe('Button Functional Testing', () => {
...
 
     await ButtonPageObject.clickComponent(); // Reset Button State
   });
 
   it('Validate OnClick() callback was fired -> Type "SPACE"', async () => {
-    await ButtonPageObject.sendKey(ButtonSelector.PrimaryButton, Keys.SPACE);
-    await expect(await ButtonPageObject.didOnClickCallbackFire()).toBeTruthy();
+    await ButtonPageObject.sendKeys(ButtonPageObject._primaryComponent, [Keys.SPACE]);
+    await expect(await ButtonPageObject.waitForOnClickCallbackToFire()).toBeTruthy(
+      "'Space' keypress failed to fire the onClick callback for the primary button.",
```
samuelfreiberg, line 95: "Pressing the "Spacebar" key on the primary button failed to fire the onClick() callback"

apps/E2E/src/Callout/pages/CalloutPageObject.win.ts
```diff
@@ -14,32 +14,20 @@ class CalloutPageObject extends BasePage {
...
-    await (await this._buttonToOpenCallout).addValue('');
-  }
-
-  async openCallout(): Promise<void> {
+  // This both opens and waits for it to go in view
+  async openCalloutAndWaitForLoad(): Promise<void> {
     await (await this._buttonToOpenCallout).click();
+    await this.waitForCondition(
+      async () => await this.didCalloutLoad(),
+      'The testing callout failed to display after attempting to open it.',
```
samuelfreiberg, line 22: Nit: Let's get more specific:

"Clicked on the button to open the Callout, but the Callout did not open correctly"

apps/E2E/src/Callout/specs/Callout.spec.win.ts
```diff
@@ -45,13 +53,12 @@ describe('Callout Accessibility Testing', () => {
...
-    await CalloutPageObject.scrollToTestElement();
-    await CalloutPageObject.openCallout();
-    await CalloutPageObject.waitForCalloutComponentInView(PAGE_TIMEOUT);
+    await CalloutPageObject.scrollToTestElement(await CalloutPageObject._buttonToOpenCallout);
+    await CalloutPageObject.openCalloutAndWaitForLoad();
   });
 
   it('Open the callout and validate it loaded correctly (visible)', async () => {
-    await expect(await CalloutPageObject.didCalloutLoad()).toBeTruthy();
+    await expect(await CalloutPageObject.didCalloutLoad()).toBeTruthy('The callout failed to visibly display.');
```
samuelfreiberg, line 61: Let's get more specific - What action did we take to open the callout?

apps/E2E/src/Checkbox/pages/CheckboxPageObject.ts
```diff
@@ -22,11 +22,7 @@ class CheckboxPageObject extends BasePage {
   }
 
   async waitForCheckboxChecked(timeout?: number): Promise<void> {
-    await browser.waitUntil(async () => await this.isCheckboxChecked(), {
-      timeout: timeout ?? this.waitForUiEvent,
-      timeoutMsg: 'The Checkbox was not toggled correctly',
-      interval: 1000,
-    });
+    await this.waitForCondition(async () => await this.isCheckboxChecked(), 'The Checkbox was not toggled correctly.', timeout);
```
samuelfreiberg, line 25: Let's consolidate all these functions to one function:

```
toggleCheckbox(toggleTo: ToggleState) {
  if(await (await this._primaryComponent).isSelected() != toggleTo) {
     ClickCheckbox()
  }
  waitForCondition
```

apps/E2E/src/Checkbox/specs/Checkbox.spec.windows.ts
lawrencewin: The .windows.ts spec states that an appium method for checking whether the checkbox is toggled does not work on the UWP app. Therefore, functional tests in the .win.ts spec will always fail.

ksiler: Gotcha, thanks for clarifying!

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
     await AvatarPageObject.scrollToTestElement();
   });
 
-  it('Validate accessibilityLabel', async () => {
-    await expect(await AvatarPageObject.getPrimaryComponentAttribute(ACCESSIBILITY_LABEL_ATTR)).toEqual(AVATAR_ACCESSIBILITY_LABEL);
+  it('Set accessibilityLabel -> Validate set accessibilityLabel is correct', async () => {
```
samuelfreiberg, line 32: Nit: Remove "set" in 2nd part of error message 

"Set accessibilityLabel -> Validate accessibilityLabel is correct"

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
...
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Validate accessibilityLabel from `name` prop', async () => {
-    await expect(await AvatarPageObject.getSecondaryComponentAttribute(ACCESSIBILITY_LABEL_ATTR)).toEqual(
-      AVATAR_ACCESSIBILITY_LABEL_BY_NAME,
-    );
+  it("Set 'name' prop -> Validate accessibilityLabel is correct", async () => {
```
samuelfreiberg, line 40: What does "correct" mean here? Most people probably don't know that if we only set the 'name' prop and not the accessibilityLabel, then accessibilityLabel defaults to the 'name' prop.

"Correct" doesn't add any value, we should add a statement about what _makes_ it correct:

"Set 'name' prop without setting 'accessibilityLabel' prop -> 'accessibilityLabel' should default to 'name'"

apps/E2E/src/Callout/specs/Callout.spec.win.ts
```diff
@@ -45,13 +53,12 @@ describe('Callout Accessibility Testing', () => {
...
-    await CalloutPageObject.scrollToTestElement();
-    await CalloutPageObject.openCallout();
-    await CalloutPageObject.waitForCalloutComponentInView(PAGE_TIMEOUT);
+    await CalloutPageObject.scrollToTestElement(await CalloutPageObject._buttonToOpenCallout);
+    await CalloutPageObject.openCalloutAndWaitForLoad();
   });
 
-  it('Open the callout and validate it loaded correctly (visible)', async () => {
-    await expect(await CalloutPageObject.didCalloutLoad()).toBeTruthy();
+  it('Open the callout via button press -> Validate that it is visible and loaded', async () => {
```
samuelfreiberg, line 60: Nit: Replace "press" with "click".

samuelfreiberg: "click" is specific to a mouse event

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
...
+    await expect(
+      await AvatarPageObject.compareAttribute(AvatarPageObject._primaryComponent, Attribute.AccessibilityRole, LINK_A11Y_ROLE),
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Validate default accessibilityRole', async () => {
-    await expect(await AvatarPageObject.getSecondaryComponentAttribute(ACCESSIBILITY_ROLE_ATTRIBUTE)).toEqual(ACCESSIBILITY_ROLE_IMAGE);
+  it('Do NOT set accessibilityRole -> Validate default accessibilityRole is correct', async () => {
```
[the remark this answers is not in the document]
ksiler: The user does have the ability to set the accessibilityRole. See [useAvatar.ts](https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Avatar/src/useAvatar.ts#L80):

`accessibilityRole: accessibilityRole ?? 'image'`

I believe most (if not all) of the FURN controls have the accessibilityRole prop exposed (coming from View props) but not all of them allow the default accessibilityRole to be overridden. I'm not sure if that's an oversight or by design.

I think allowing the default accessibilityRole prop to be overridden makes sense for some controls but not all.  A user of Text, for example, should be able to set is as either a 'text', 'link', or 'header' role which are all valid and makes a difference in how it's exposed to screen readers. On the other hand, it may only make sense for a RadioButton to be set as a 'radio' and nothing else. I don't think there should be an issue with letting the user override the accessibilityRole on a control if they want as long as we provide the correct accessibilityRole in the default case so it works properly "out of the box" and with any other role(s) that make sense.

For Avatar, I've seen it be represented as an 'image' or 'button' in different scenarios and I think (according to the spec) there's plans to add an onClick() handler in which case setting it as a 'button' role would make sense. As far as testing goes, I think as long as it makes sense to have a role other than the default role set, we should test our ability to set a different role.

apps/E2E/src/Checkbox/pages/CheckboxPageObject.ts
```diff
@@ -21,45 +15,29 @@ class CheckboxPageObject extends BasePage {
...
+    return (await this.isCheckboxChecked()) === newState;
   }
 
-  /* Useful in beforeEach() hook to reset the checkbox before every test */
-  async toggleCheckboxToUnchecked(): Promise<void> {
-    if (await this.isCheckboxChecked()) {
-      await (await this._primaryComponent).click();
+  async toggleCheckbox(newState: boolean) {
+    if ((await this.isCheckboxChecked()) !== newState) {
+      await this.click(this._primaryComponent);
```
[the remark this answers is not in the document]
lawrencewin: in this case, the `click()` method in BasePage accepts a `Promise<WebdriverIO.Element>` so we don't have to pass in an `await` for convenience sake.

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
...
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Validate accessibilityLabel from `name` prop', async () => {
-    await expect(await AvatarPageObject.getSecondaryComponentAttribute(ACCESSIBILITY_LABEL_ATTR)).toEqual(
-      AVATAR_ACCESSIBILITY_LABEL_BY_NAME,
-    );
+  it("Set 'name' prop -> 'accessibilityLabel' should default to '<name>, available'", async () => {
```
samuelfreiberg, line 40: Confused about "'<name>, available'"

lawrencewin: When the name prop is set to something like "Larry Win", the accessibilityLabel should read "Larry Win, available". This is according to the test.

samuelfreiberg: Awesome :) Did not know that

samuelfreiberg, line 40: Clarify more:

"Set 'name' prop without setting 'accessibilityLabel' prop -> 'accessibilityLabel' should default to '<name>, available'"

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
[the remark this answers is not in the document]
samuelfreiberg: Thanks for the info!

apps/E2E/src/Callout/pages/CalloutPageObject.win.ts
```diff
@@ -14,32 +14,20 @@ class CalloutPageObject extends BasePage {
...
-  /* OVERRIDE: This must scroll to the button that opens the callout, not the callout (since it's not visible on load.) */
-  async scrollToTestElement(): Promise<void> {
-    // on win32, adding a blank value to an element automatically scrolls to it
-    await (await this._buttonToOpenCallout).addValue('');
-  }
-
-  async openCallout(): Promise<void> {
+  // This both opens and waits for it to go in view
+  async openCalloutAndWaitForLoad(): Promise<void> {
     await (await this._buttonToOpenCallout).click();
```
samuelfreiberg, line 19: I think we should check if it's already open (doesn't hurt). I could see it being a headache for someone if they're calling this and it's already open (then this function will close it and the `wait()` will fail

apps/E2E/src/Checkbox/pages/CheckboxPageObject.ts
```diff
@@ -21,45 +15,29 @@ class CheckboxPageObject extends BasePage {
     return await (await this._primaryComponent).isSelected();
   }
 
-  async waitForCheckboxChecked(timeout?: number): Promise<void> {
-    await browser.waitUntil(async () => await this.isCheckboxChecked(), {
-      timeout: timeout ?? this.waitForUiEvent,
-      timeoutMsg: 'The Checkbox was not toggled correctly',
-      interval: 1000,
-    });
```
samuelfreiberg, line 29: Let's add a comment here giving a quick explanation of the function (we should get in the habit of documenting each method even if it's seems redundant. It never hurts giving a brief explanation)

lawrencewin: done

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
...
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Validate accessibilityLabel from `name` prop', async () => {
-    await expect(await AvatarPageObject.getSecondaryComponentAttribute(ACCESSIBILITY_LABEL_ATTR)).toEqual(
-      AVATAR_ACCESSIBILITY_LABEL_BY_NAME,
-    );
+  it("Set 'name' prop without setting 'accessibilityLabel -> 'accessibilityLabel' should default to '_, available'", async () => {
```
samuelfreiberg, line 40: "'_, available'" is still confusing. Let's do:

"{name}, available"

apps/E2E/src/Checkbox/pages/CheckboxPageObject.ts
```diff
@@ -21,45 +15,34 @@ class CheckboxPageObject extends BasePage {
...
+    if ((await this.isCheckboxChecked()) !== newState) {
+      await this.click(this._primaryComponent);
+      await this.waitForCheckboxToggle(
+        newState,
+        `Clicked the primary checkbox to turn it ${newState ? 'on' : 'off'}, but it failed to change states.`,
+      );
     }
   }
 
+  /* The onChange callback for the checkbox shows a hidden Text component near the checkmark. This checks if the text
```
samuelfreiberg, line 38: I'd recommend providing more info on why we need the text in the first place (that's the confusing part about this function).

"Unlike snapshot testing, we can't spy on functions to determine if they get called or not. So, in order to determine if the onChange() callback gets fired, we show/hide a text component as the onChange() callback is fired. Then, we can check to see if this text exists to determine if the onChange() fired."

apps/E2E/src/CheckboxExperimental/pages/ExperimentalCheckboxPageObject.ts
```diff
@@ -21,45 +15,34 @@ class ExperimentalCheckboxPageObject extends BasePage {
...
+    if ((await this.isCheckboxChecked()) !== newState) {
+      await this.click(this._primaryComponent);
+      await this.waitForCheckboxToggle(
+        newState,
+        `Clicked the primary checkbox to turn it ${newState ? 'on' : 'off'}, but it failed to change states.`,
+      );
     }
   }
 
+  /* The onChange callback for the checkbox shows a hidden Text component near the checkmark. This checks if the text
```
samuelfreiberg, line 38: Same as above

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -53,24 +69,32 @@ describe('Button Functional Testing', () => {
   });
 
   it('Validate OnClick() callback was fired -> Click', async () => {
-    await ButtonPageObject.clickComponent();
-    await expect(await ButtonPageObject.didOnClickCallbackFire()).toBeTruthy();
+    await ButtonPageObject.click(ButtonPageObject._primaryComponent);
+    await expect(await ButtonPageObject.waitForOnClickCallbackToFire()).toBeTruthy(
+      'Clicking on the primary button failed to fire the onClick() callback.',
```
ksiler, line 74: You may have to double check, but does this error message actually get used at all? I recall testing this out and only saw the error message from `waitForOnClickCallbackToFire()` ("The Button onClick callback did not fire."). I think the error messages in the spec file are more helpful since they're more specific. Any way we can get these error messages to appear instead of the generic one)?

lawrencewin: You are correct. The new commit should fix this by passing the messages into the waiting methods. This was an issue with Button, ButtonExperimental, Checkbox, and CheckboxExperimental tests, which should be good now.

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -31,17 +30,34 @@ describe('Button Accessibility Testing', () => {
   });
 
   it('Button - Validate accessibilityRole is correct', async () => {
```
ksiler, line 32: nit no need for "Button - " prefix (same for the next three tests). We already describe this as "Button Accessibility Testing".

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -31,17 +30,34 @@ describe('Button Accessibility Testing', () => {
...
-    await expect(await ButtonPageObject.getAccessibilityRole()).toEqual(BUTTON_A11Y_ROLE);
+    await expect(
+      await ButtonPageObject.compareAttribute(ButtonPageObject._primaryComponent, Attribute.AccessibilityRole, BUTTON_A11Y_ROLE),
+    ).toBeTrue();
+
     await expect(await ButtonPageObject.didAssertPopup()).toBeFalsy(ButtonPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Button - Set accessibilityLabel', async () => {
```
ksiler, line 40: nit I think we should follow the same wording as the rest of the spec files i.e. "Set accessibilityLabel -> Validate accessibilityLabel is correct"

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -31,17 +30,34 @@ describe('Button Accessibility Testing', () => {
...
+        ButtonPageObject._primaryComponent,
+        Attribute.AccessibilityLabel,
+        BUTTON_ACCESSIBILITY_LABEL_DEPRECATED,
+      ),
+    ).toBeTrue();
+
     await expect(await ButtonPageObject.didAssertPopup()).toBeFalsy(ButtonPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Button - Do not set accessibilityLabel -> Default to Button label', async () => {
```
ksiler, line 52: nit "Do not set accessibilityLabel -> Validate accessibilityLabel defaults to label"

apps/E2E/src/Button/specs/Button.spec.win.ts
```diff
@@ -53,24 +69,36 @@ describe('Button Functional Testing', () => {
   });
 
   it('Validate OnClick() callback was fired -> Click', async () => {
```
ksiler, line 71: nit this reads a bit backwards to me. I think it should be "condition/action -> expected result" so:

"Click on Button -> Validate OnClick() callback was fired"
"Type "Enter" on Button -> Validate OnClick() callback was fired"
"Type "Space" on Button -> Validate OnClick() callback was fired"

same for ExperimentalButton

apps/E2E/src/ButtonExperimental/pages/ButtonExperimentalPageObject.ts
```diff
@@ -10,14 +10,9 @@ class ButtonExperimentalPageObject extends BasePage {
   /******************************************************************/
   /**************** UI Element Interaction Methods ******************/
   /******************************************************************/
-  async didOnClickCallbackFire(): Promise<boolean> {
+  async didOnClickCallbackFire(errorMsg?: string): Promise<boolean> {
```
ksiler, line 13: This function looks/reads cleaner to me than `waitForOnClickCallbackToFire` in ButtonPageObject.ts. Could we replace `waitForOnClickCallbackToFire` in ButtonPageObject with this function (`didOnClickCallbackFire`)? Or at least make them both the same?

apps/E2E/src/Callout/specs/Callout.spec.win.ts
```diff
@@ -23,18 +22,27 @@ describe('Callout Testing Initialization', function () {
...
   it('Validate accessibilityRole is correct', async () => {
-    await expect(await CalloutPageObject.getAccessibilityRole()).toEqual(CALLOUT_A11Y_ROLE);
+    await expect(
+      await CalloutPageObject.compareAttribute(CalloutPageObject._primaryComponent, Attribute.AccessibilityRole, CALLOUT_A11Y_ROLE),
+    ).toBeTrue();
+
     await expect(await CalloutPageObject.didAssertPopup()).toBeFalsy(CalloutPageObject.ERRORMESSAGE_ASSERT); // Ensure no asserts popped up
   });
 
   it('Set accessibilityLabel', async () => {
```
ksiler, line 37: nit "Set accessibilityLabel -> Validate accessibilityLabel is correct"

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -27,17 +26,34 @@ describe('Checkbox Accessibility Testing', () => {
   });
 
   it('Checkbox - Validate accessibilityRole is correct', async () => {
```
ksiler, line 28: nit same as Button, no need for the "Checkbox - " prefix (same for next two tests)

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -27,17 +26,34 @@ describe('Checkbox Accessibility Testing', () => {
...
-    await expect(await CheckboxPageObject.getAccessibilityRole()).toEqual(CHECKBOX_A11Y_ROLE);
+    await expect(
+      await CheckboxPageObject.compareAttribute(CheckboxPageObject._primaryComponent, Attribute.AccessibilityRole, CHECKBOX_A11Y_ROLE),
+    ).toBeTrue();
+
     await expect(await CheckboxPageObject.didAssertPopup()).toBeFalsy(CheckboxPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Checkbox - Set accessibilityLabel', async () => {
```
ksiler, line 36: nit "Set accessibilityLabel -> Validate accessibilityLabel is correct"

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -27,17 +26,34 @@ describe('Checkbox Accessibility Testing', () => {
...
+        CheckboxPageObject._primaryComponent,
+        Attribute.AccessibilityLabel,
+        CHECKBOX_ACCESSIBILITY_LABEL,
+      ),
+    ).toBeTrue();
+
     await expect(await CheckboxPageObject.didAssertPopup()).toBeFalsy(CheckboxPageObject.ERRORMESSAGE_ASSERT);
   });
 
   it('Checkbox - Do not set accessibilityLabel -> Default to Checkbox label', async () => {
```
ksiler, line 48: nit "Do not set accessibilityLabel -> Validate accessibilityLabel defaults to label"

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
```diff
@@ -47,36 +63,57 @@ describe('Checkbox Functional Testing', () => {
...
-    await expect(await CheckboxPageObject.isCheckboxChecked()).toBeFalsy();
+    await expect(
+      await CheckboxPageObject.waitForCheckboxToggle(false, 'The primary checkbox should have been toggled off via click.'),
+    ).toBeTruthy();
+
     await expect(await CheckboxPageObject.didAssertPopup()).toBeFalsy(CheckboxPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Click the "SPACE" on a Checkbox and verify it toggles', async () => {
+  it("Press 'SPACE' on a Checkbox and verify it toggles correctly AND calls the user's onChange", async () => {
```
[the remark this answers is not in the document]
ksiler: Same for ExperimentalCheckbox

apps/E2E/src/Checkbox/specs/Checkbox.spec.win.ts
ksiler: Same for ExperimentalCheckbox

apps/E2E/src/Checkbox/specs/Checkbox.spec.windows.ts
```diff
@@ -26,15 +25,29 @@ describe('Checkbox Accessibility Testing', () => {
   });
 
   it('Checkbox - Validate accessibilityRole is correct', async () => {
-    await expect(await CheckboxPageObject.getAccessibilityRole()).toEqual(CHECKBOX_A11Y_ROLE);
+    await expect(
+      await CheckboxPageObject.compareAttribute(CheckboxPageObject._primaryComponent, Attribute.AccessibilityRole, CHECKBOX_A11Y_ROLE),
+    ).toBeTrue();
   });
 
   it('Checkbox - Set accessibilityLabel', async () => {
```
ksiler, line 33: nit "Checkbox -" prefix not needed

ksiler (approved):

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -34,30 +29,47 @@ describe('Avatar Accessibility Testing', () => {
...
+    ).toBeTrue();
+
     await expect(await AvatarPageObject.didAssertPopup()).toBeFalsy(AvatarPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Validate accessibilityLabel from `name` prop', async () => {
-    await expect(await AvatarPageObject.getSecondaryComponentAttribute(ACCESSIBILITY_LABEL_ATTR)).toEqual(
-      AVATAR_ACCESSIBILITY_LABEL_BY_NAME,
-    );
+  it("Set 'name' prop without setting 'accessibilityLabel -> 'accessibilityLabel' should default to '{name}, available'", async () => {
```
samuelfreiberg, line 40: Nit: Missing ' after 'accessibilityLabel

samuelfreiberg (approved):

PR #2432, opened 2022-12-20
Remove inline links from win32

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We've decided to not support inline links on the win32 platform as the implementation is taking a lot of time, but we do not anticipate lots of usage of inline links due to how UI is generally designed on the platform.

This PR:
1. Forces inline links to not be used on win32. This will result in failures rendering components if you try using inline links on win32. 
2. Specifies that inline links are not supported on win32 in documentation
3. Removes the inline test link case for the win32 platform.

### Verification

Verified that if inline link is attempted in win32, that an error is thrown.
Verified that Link test renders as expected without inline link test cases.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/LinkExperimental/LinkTest.tsx
```diff
@@ -25,52 +26,24 @@ const DefaultLinks: React.FunctionComponent = () => {
...
-        to navigate.
-      </Text>
-    </Stack>
-  );
-};
-
 const SubtleLinks: React.FunctionComponent = () => {
   const doPress = React.useCallback(() => Alert.alert('Alert.', 'You have been alerted.'), []);
   const doAllyTap = React.useCallback(() => Alert.alert('Alert.', 'You have invoked onAllyTap.'), []);
+  const supportsInlineLink = Platform.OS !== ('win32' as any);
```
Saadnajmi, line 32: I'm not sure any platform can properly support inline links for similar reasons. Maybe just... Remove it altogether?

On the macOS side, some of our collaborators at Meta have actually investigated adding an `href` prop directly to text in React Native macOS too very support links.

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/LinkExperimental/InlineLinksTest.win32.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import { TextV1 as Text } from '@fluentui/react-native';
+
+// Platform.select not available for win32
+// Forking test instead
+export const InlineLinks: React.FunctionComponent = () => {
+  return <Text>Inline links are not supported on win32.</Text>;
+};
```
Saadnajmi, line 8: Nit: I've given feedback in other PRs to try and fork tests as little as possible for consistency of people reading code (and also try to standardize where we fork by platform). To that effect, I wonder if it's better to just put this check in the default file instead of forking for win32. Not a strong preference either way since we haven't documented how we want to handle platform specific tests yet.

rurikoaraki: I'm in camp fork tests 😅Personally I think the code is eventually going to get really messy with platform checks. I'll WF for now, we can decide when documentation is written up.

PR #2435, opened 2022-12-20
Fork SubmenuItem pressable

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

PR #2403 reverted SubmenuItem pressable to useAsPressable, which caused a regression on Win32 where submenu doesn't open on mouse click. Note that this regression does not repro in FluentTester but only in our client app. Since the intent of the initial fix was to address a bug on macOS, let's fork the SubmenuItem pressable code to mitigate the regression.
### Verification

No behavioral changes on CM.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-contextual-menu-59d1d1ea-40db-437d-80ab-b615797ee124.json
+++ b/change/@fluentui-react-native-contextual-menu-59d1d1ea-40db-437d-80ab-b615797ee124.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fork SubmenuItem pressable",
+  "packageName": "@fluentui-react-native/contextual-menu",
+  "email": "amchiu@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/components/ContextualMenu/src/SubmenuItem.tsx
+++ b/packages/components/ContextualMenu/src/SubmenuItem.tsx
@@ -1,6 +1,6 @@
 /** @jsx withSlots */
 import * as React from 'react';
-import { I18nManager, Platform, View } from 'react-native';
+import { I18nManager, Platform, Pressable, View } from 'react-native';
 import {
   SubmenuItemSlotProps,
   SubmenuItemState,
@@ -15,7 +15,7 @@ import { Text } from '@fluentui-react-native/text';
 import { settings } from './SubmenuItem.settings';
 import { backgroundColorTokens, borderTokens, textTokens, foregroundColorTokens, getPaletteFromTheme } from '@fluentui-react-native/tokens';
 import { mergeSettings } from '@uifabricshared/foundation-settings';
-import { useKeyDownProps, useViewCommandFocus, useAsPressable } from '@fluentui-react-native/interactive-hooks';
+import { useKeyDownProps, useViewCommandFocus, useAsPressable, usePressableState } from '@fluentui-react-native/interactive-hooks';
 import { CMContext } from './ContextualMenu';
 import { Icon, SvgIconProps, createIconProps } from '@fluentui-react-native/icon';
 import { Svg, G, Path, SvgProps } from 'react-native-svg';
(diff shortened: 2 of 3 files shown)


packages/components/ContextualMenu/src/SubmenuItem.types.ts
```diff
@@ -14,7 +15,7 @@ export type SubmenuItemProps = ContextualMenuItemProps;
 export type SubmenuItemState = ContextualMenuItemState;
 
 export interface SubmenuItemSlotProps {
-  root: React.PropsWithRef<IViewProps>;
+  root: React.PropsWithRef<PressablePropsExtended>;
```
rurikoaraki, line 18: You might need to fork this based on platform, too

Saadnajmi:
Uh.. do we know what the difference is between client code and our test app is that is causing this issue? I would really prefer if we don't have to fork like this..

packages/components/ContextualMenu/src/SubmenuItem.types.ts
chiuam: The file SubmenuItem.types.macos.ts I added should take care of that, doesn't it?

chiuam:
> Uh.. do we know what the difference is between client code and our test app is that is causing this issue? I would really prefer if we don't have to fork like this..

I don't know what is causing this issue precisely, but it's odd that win32 has their own SubmenuItem implementation. It's not a pretty fix, I'm hoping we can transition to Menu V1 as soon as we can!
 
cc @ksiler might have more knowledge on this

ksiler:
> > Uh.. do we know what the difference is between client code and our test app is that is causing this issue? I would really prefer if we don't have to fork like this..
> 
> I don't know what is causing this issue precisely, but it's odd that win32 has their own SubmenuItem implementation. It's not a pretty fix, I'm hoping we can transition to Menu V1 as soon as we can!
> 
> cc @ksiler might have more knowledge on this

The client code uses a customized version of SubmenuItem on win32 which uses the previous `usePressableState` hook instead of `useAsPressable`. Switching the client code to use `useAsPressable` (to match the current SubmenuItem implementation) fixes the issue but there's a type error that occurs:

```
Types of property 'cancelable' are incompatible.
      Type 'boolean | null | undefined' is not assignable to type 'boolean | undefined'.
        Type 'null' is not assignable to type 'boolean | undefined'.

     const pressable = useAsPressable({`
 ```
 
This is because the cancelable prop on SubmenuItemProps is defined as `cancelable?: null | boolean | undefined;` (from types/react-native) and useAsPressable uses `cancelable?: boolean;` (from the PressabilityConfig type in @fluentui-react-native/interactive-hooks). This is resolved by adding `"strictNullChecks": false` in the client code's tsconfig.json.

Saadnajmi:
> > > Uh.. do we know what the difference is between client code and our test app is that is causing this issue? I would really prefer if we don't have to fork like this..
> > 
> > I don't know what is causing this issue precisely, but it's odd that win32 has their own SubmenuItem implementation. It's not a pretty fix, I'm hoping we can transition to Menu V1 as soon as we can!
> > 
> > cc @ksiler might have more knowledge on this
> 
> The client code uses a customized version of SubmenuItem on win32 which uses the previous `usePressableState` hook instead of `useAsPressable`. Switching the client code to use `useAsPressable` (to match the current SubmenuItem implementation) fixes the issue but there's a type error that occurs:
> 
> ```
> Types of property 'cancelable' are incompatible.
>       Type 'boolean | null | undefined' is not assignable to type 'boolean | undefined'.
>         Type 'null' is not assignable to type 'boolean | undefined'.
> 
>      const pressable = useAsPressable({`
>  ```
>  
> This is because the cancelable prop on SubmenuItemProps is defined as `cancelable?: null | boolean | undefined;` (from types/react-native) and useAsPressable uses `cancelable?: boolean;` (from the PressabilityConfig type in @fluentui-react-native/interactive-hooks). This is resolved by adding `"strictNullChecks": false` in the client code's tsconfig.json.

Couldn't we also fix the type in PressabilityConfig in interactive hooks since we own that?

Saadnajmi:
Also, is the intention with this PR to switch the client over to our implementation of submenuitem? Or are they keeping their customized version?

chiuam:
> Also, is the intention with this PR to switch the client over to our implementation of submenuitem? Or are they keeping their customized version?

No, merging this PR shouldn't require any additional changes in the client code

Saadnajmi:
> > Also, is the intention with this PR to switch the client over to our implementation of submenuitem? Or are they keeping their customized version?
> 
> No, merging this PR shouldn't require any additional changes in the client code

I'm confused then. We determined that `usePressableState` causes issues with submenuitem because of how we override both onHoverIn and onMouseEnter. That shouldn't be platform dependent afaik as that logic is all JS. Wouldn't reverting just win32 in fluent tester mean we get that bug on win32? And also, what's the point of reverting win32 in FIR. if client code is using a custom implementation anyway?

Saadnajmi (approved):
While I'm not convinced this is the right fix, approving because client team wants a fix quickly and this may help mitigate.

PR #2441, opened 2022-12-21
Update Aliascolortoken interface

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

In preparation of adding iOS alias color tokens, updating the existing AliasColorTokens interface to specify which platforms each token is on, make all the properties optional (there are only 30 tokens that are currently defined on all platforms, and that still seems to be changing) and also reorganized the tokens into the following:

// Foreground
1. Neutralforeground 1-4
2. Neutralforegrounddisabled
3. Neutralforegroundoncolor and onboard
4. Neutralforegroundinverted
5. Brandforegroundlink
6. Compoundbrandforeground
7. Brandforeground 1 and 2
8. Brandforegroundtint
9. Brandforegrounddisabled

// Background
10. Neutralbackground 1-6
11. Neutralbackground other
12. Subtle background
13. Transparent background
14. Brandbackground

// Stroke
15. Neutralstroke
16. Stroke focus
17. Brand stroke
18. Transparent stroke

// Red
16. Red background/foreground/border

Changes are split up by commit.

### Verification

There are no functional changes in this PR - no properties were removed/added to this interface.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-types/src/Color.types.ts
```diff
@@ -451,152 +451,362 @@ export interface ControlColorTokens {
 }
 
 export interface AliasColorTokens {
+  /// Neutral colors
+
+  /** Defined on platforms: android, iOS, macOS, win32, windows */
   neutralForeground1: ColorValue;
+
+  /** Defined on platforms: macOS, win32, windows */
```
lyzhan7, line 459: Realized maybe I should delete all the empty lines in between each token definition, to be consistent with some comments above

Saadnajmi: I personally like the empty lines better fwiw

ankraj12 (approved):

packages/theming/theme-types/src/Color.types.ts
rurikoaraki: Same. It's more readable.

rurikoaraki:
Just a quick note about the regrouping - The current ordering should be similar to how tokens are listed by design (generally grouped by foreground/background/stroke). Reorganizing is fine, but it does make it harder to follow the list provided by design since the ordering will be different from how design lists tokens.

lyzhan7:
> 

I see, the iOS/android designs are ordered by neutral, brand, and shared colors (which we don't have yet). If win32 is ordered different I can revert back to what it was before, there were a couple tokens that seemed out of place with either order so I'll just move those

lyzhan7:
Tried ordering the tokens again, this time following foreground/background/stroke/red. Updated the PR description with a more detailed list of how things are grouped. I don't have a strong preference for what the groupings are if designs differ across platforms, just want there to be some kind of consistent organization so tokens are easier to find. If reordering the tokens should be discussed offline I can also scope this PR to just adding the platforms

lyzhan7:
Another thought - should all the properties in AliasColorTokens be optional? I will have to mark some of them as optional since iOS doesn't define them, but wondering if I should only update the necessary ones or if I should just update all of them. I guess any token that isn't on all the platforms will need to be marked as optional (which, currently seems like there are only 30 tokens that are defined on all platforms). Since token definitions still seem to be changing, thinking I can just go ahead and mark them all as optional.

rurikoaraki (approved):

rurikoaraki:
> Another thought - should all the properties in AliasColorTokens be optional? I will have to mark some of them as optional since iOS doesn't define them, but wondering if I should only update the necessary ones or if I should just update all of them. I guess any token that isn't on all the platforms will need to be marked as optional (which, currently seems like there are only 30 tokens that are defined on all platforms). Since token definitions still seem to be changing, thinking I can just go ahead and mark them all as optional.

No strong opinion from me.

ankraj12:
> Another thought - should all the properties in AliasColorTokens be optional? I will have to mark some of them as optional since iOS doesn't define them, but wondering if I should only update the necessary ones or if I should just update all of them. I guess any token that isn't on all the platforms will need to be marked as optional (which, currently seems like there are only 30 tokens that are defined on all platforms). Since token definitions still seem to be changing, thinking I can just go ahead and mark them all as optional.

We can skip marking those properties as optional where are defined for all endpoints. I am hoping those should already be marked optional post the integration for android colors. You would additionally need to override mapping file for iOS - similar to https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/theming-utils/src/mapPipelineToTheme.android.ts. I feel this is a clearer way to indicate which properties work for which platform, and this also allows easily accounting for name conventions changes (different alias names in design tokens package) for a platform.

lyzhan7:
> > Another thought - should all the properties in AliasColorTokens be optional? I will have to mark some of them as optional since iOS doesn't define them, but wondering if I should only update the necessary ones or if I should just update all of them. I guess any token that isn't on all the platforms will need to be marked as optional (which, currently seems like there are only 30 tokens that are defined on all platforms). Since token definitions still seem to be changing, thinking I can just go ahead and mark them all as optional.
> 
> We can skip marking those properties as optional which are defined for all endpoints. I am hoping those should already be marked optional post the integration for android colors. You would additionally need to override mapping file for iOS - similar to https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/theming-utils/src/mapPipelineToTheme.android.ts. I feel this is a clearer way to indicate which properties work for which platform, and this also allows easily accounting for name conventions changes (different alias names in design tokens package) for a platform.

For the properties that were defined on all endpoints, I was debating between marking them as optional or keeping them required. I ended up just marking them as optional since the tokens are continuing to change and it seems possible we'd just end up marking them as optional again. I don't have a strong opinion though, if you think we should make them optional I can make that change. There were actually a few tokens that I needed to mark as optional - Android had some that weren't defined on Android (ex. the hover ones) - I have a change I'm working on for mapPipelineToTheme.ios.ts where it will be more clear what additional changes I needed to make for iOS. 

I added the @platform indicator to AliasColorTokens because if you right click on a reference to t.colors.<token> it will take you to AliasColorTokens, not one of the mapPipelineToToken files. The mapPipelineToToken files are also currently out of date, some of them reference tokens that aren't actually defined for that platform in the design-tokens repo. Maybe it could cause some confusion that the mapPipelineToTheme files should match the platforms that are now specified in AliasColorTokens, but for now at least there's one place that's up to date with the design-tokens repo.

ankraj12:
> > > Another thought - should all the properties in AliasColorTokens be optional? I will have to mark some of them as optional since iOS doesn't define them, but wondering if I should only update the necessary ones or if I should just update all of them. I guess any token that isn't on all the platforms will need to be marked as optional (which, currently seems like there are only 30 tokens that are defined on all platforms). Since token definitions still seem to be changing, thinking I can just go ahead and mark them all as optional.
> > 
> > 
> > We can skip marking those properties as optional which are defined for all endpoints. I am hoping those should already be marked optional post the integration for android colors. You would additionally need to override mapping file for iOS - similar to https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/theming-utils/src/mapPipelineToTheme.android.ts. I feel this is a clearer way to indicate which properties work for which platform, and this also allows easily accounting for name conventions changes (different alias names in design tokens package) for a platform.
> 
> For the properties that were defined on all endpoints, I was debating between marking them as optional or keeping them required. I ended up just marking them as optional since the tokens are continuing to change and it seems possible we'd just end up marking them as optional again. I don't have a strong opinion though, if you think we should make them optional I can make that change. There were actually a few tokens that I needed to mark as optional - Android had some that weren't defined on Android (ex. the hover ones) - I have a change I'm working on for mapPipelineToTheme.ios.ts where it will be more clear what additional changes I needed to make for iOS.
> 
> I added the @platform indicator to AliasColorTokens because if you right click on a reference to t.colors. it will take you to AliasColorTokens, not one of the mapPipelineToToken files. The mapPipelineToToken files are also currently out of date, some of them reference tokens that aren't actually defined for that platform in the design-tokens repo. Maybe it could cause some confusion that the mapPipelineToTheme files should match the platforms that are now specified in AliasColorTokens, but for now at least there's one place that's up to date with the design-tokens repo.

Sounds good since there already are additional changes you are integrating in for iOS.

PR #2446, opened 2022-12-21
Add spacing E2E tests

amgleitman:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Add E2E automation tests for Spacing page.

### Verification

If the tests pass, this should be good.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


lyzhan7:
The test itself look good to me but I don't think the Spacing test page has been added to the other platforms yet, it's still only on iOS., so the e2e tests will probably fail on win32/macOS. Also you can also try running the tests locally - instructions for iOS here: https://github.com/microsoft/fluentui-react-native/blob/main/apps/E2E/README.md#ios-steps

rurikoaraki (approved):

rurikoaraki:
I put @samuelfreiberg and @lawrencewin on the PR as an FYI since they're in the middle of revamping the E2E tests, but I don't think this is blocked on their review.

PR #2425, opened 2022-12-16
Overhaul ContextualMenu, Link, and LinkV1 tests with better errors and refactored spec code

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Continuation of PRs #2423 and #2405. 

Currently, for most developers, it is difficult to figure out what the root issue is when a FURN E2E test fails because of opaque error messages. The existing E2E tests are also in need of some clean up in terms of code and PageObject design. 

This commit is part of a larger effort to address the above. Here, I've updated error messages, use new BasePage methods, and did general cleanup for the following tests:
- ContextualMenu
- Link (deprecated)
- Link V1 (experimental)

#### Removal of Selectors (copied from #2423)

One big change to the tests across the board is the removal of Selector enums within tests. Previously, Selectors were used to choose which element was to be clicked, get an attribute from, etc. However, because many tests required different selectors from the base `ComponentSelector` in BasePage.ts to select test elements, this meant that different Page Objects ended up implementing the same methods (`sendKey()`, `click()`, and other attribute testing methods). 

In removing selectors, we can re-use methods for doing most things a test requires implemented in BasePage.ts and pass WebdriverIO elements directly (see [`compareAttribute()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L80-L93),[ `click()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L127-L130), and [`sendKeys()`](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/common/BasePage.ts#L132-L134)). This way, future Page Objects remain light and only implement getters or other methods needed for the specific test.

This doesn't mean that Selectors can't be used. Taking the [FocusZone test](https://github.com/microsoft/fluentui-react-native/blob/3500d25b492a0fba0df3ab478b09d2775b78e452/apps/E2E/src/FocusZone/pages/FocusZonePageObject.ts#L18-L28), the large amount of elements used within the test let us use a Selector Enum to pick out specific elements to pass in. But for most cases, we can rely on a small amount of getters and the reusable BasePage methods for creating a test.

### Verification

Tests pass locally and via CI. 

[...]

PR #2447, opened 2022-12-21
Overhaul Menu E2E tests with better errors, refactored test code

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Continuation of PRs https://github.com/microsoft/fluentui-react-native/pull/2423, https://github.com/microsoft/fluentui-react-native/pull/2405, and #2425.

- Menu

### Verification

Tests pass locally and within CI. 

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/Menu/pages/MenuPageObject.ts
```diff
@@ -4,23 +4,20 @@ import {
...
- * The spec file should import this enum and use it when wanting to interact with different elements on the page. */
-export const enum MenuComponentSelector {
-  MenuTrigger = 0, //this._primaryComponent
-  FirstMenuItem,
-  SecondMenuItem,
-  ThirdMenuItem,
-  FourthMenuItem,
+import { Keys } from '../../common/consts';
+
+/** Convenience enum for an external caller to get individual menu items using .getMenuItem() */
```
samuelfreiberg, line 15: Nit: Remove word "Convenience". 

"Allows caller to get specific menu items using .getMenuItem()"

apps/E2E/src/Menu/pages/MenuPageObject.ts
```diff
@@ -35,32 +32,35 @@ class MenuPageObject extends BasePage {
...
+      await this.click(this._menuTrigger);
       await this.waitForMenuToOpen();
     }
   }
 
   async closeMenu(): Promise<void> {
     if (await this.menuIsExpanded()) {
-      await this.sendKey(MenuComponentSelector.FirstMenuItem, Keys.ESCAPE);
+      await this.sendKeys(this.getMenuItem(MenuItem.First), [Keys.ESCAPE]);
       await this.waitForCondition(async () => (await this.menuIsExpanded()) === false, 'The Menu did not close.', this.waitForUiEvent, 500);
```
samuelfreiberg, line 43: Improve error message:

"Typed 'Escape' on a MenuItem, Menu should close"

apps/E2E/src/Menu/pages/MenuPageObject.ts
```diff
@@ -70,57 +70,25 @@ class MenuPageObject extends BasePage {
     );
   }
 
+  /* When menu item 1 is clicked, it increments a counter variable in the test component page, which is displayed in a text box. */
```
samuelfreiberg, line 73: Add reasoning - Why do we need this? What value is added by incrementing a counter variable?

apps/E2E/src/Menu/specs/Menu.spec.win.ts
```diff
@@ -53,27 +74,27 @@ describe('Menu Accessibility Testing', () => {
...
-    await MenuPageObject.scrollToTestElement();
+    await MenuPageObject.scrollToTestElement(await MenuPageObject._menuTrigger);
 
     await MenuPageObject.resetTest();
   });
 
   it('Click MenuTrigger -> Validate Menu is opened by checking if MenuItems are visible', async () => {
-    await MenuPageObject.clickItem(MenuComponentSelector.MenuTrigger);
+    await MenuPageObject.click(MenuPageObject._menuTrigger);
     await expect(await MenuPageObject.waitForMenuToOpen()).toBeTruthy();
```
samuelfreiberg, line 84: Add error message to waiter

lawrencewin: For this specific test case, if the waiter fails, it returns the error `The Menu did not open`. Combined with the name of the case itself, would this be clear enough? Adding an extra error feels a bit redundant.

apps/E2E/src/Menu/pages/MenuPageObject.ts
```diff
@@ -70,57 +80,27 @@ class MenuPageObject extends BasePage {
     );
   }
 
+  /* When menu item 1 is clicked, it increments a counter variable in the test component page, which is displayed in a text box. Incrementing a counter within a label
+   * allows us to test whether the onClick() callback fires for click and keypress inputs across individual test cases without having to close the menu and reset the
+   * label per case (as you see in the button + checkbox tests). This decreases test time and improves performance for this spec. */
   async itemOnClickHasFired(timesFired: number): Promise<boolean> {
     return (await (await this._callbackLabel).getText()).includes(timesFired.toString());
```
ksiler, line 87: I think we should compare the strings via "===" instead of includes in case we're firing more onClick callbacks than we expect. So in the event we happen to fire "12" onClick callbacks but we only expect "2", we would flag that as an error. Otherwise, the includes function would return true.

apps/E2E/src/Menu/pages/MenuPageObject.ts
```diff
@@ -35,32 +32,45 @@ class MenuPageObject extends BasePage {
...
       await this.waitForMenuToOpen();
     }
   }
 
   async closeMenu(): Promise<void> {
     if (await this.menuIsExpanded()) {
-      await this.sendKey(MenuComponentSelector.FirstMenuItem, Keys.ESCAPE);
-      await this.waitForCondition(async () => (await this.menuIsExpanded()) === false, 'The Menu did not close.', this.waitForUiEvent, 500);
+      await this.sendKeys(this.getMenuItem(MenuItem.First), [Keys.ESCAPE]);
+      await this.waitForCondition(
```
ksiler, line 43: nit could we create a `waitForMenuToClose()` function to mimic the `openMenu()` and `waitForMenuToOpen()` functions? I could see us adding a test to close the menu by clicking somewhere else (instead of pressing 'ESC') in the future and having `waitForMenuToClose()` as a separate function would help with that. I also think it makes the code easier to read since there would be an `openMenu/closeMenu` function and `waitForMenuToOpen/waitFoMenuToClose` function instead of just a `closeMenu` function but maybe that's just me.

samuelfreiberg (approved):

ksiler (approved):

PR #2449, opened 2022-12-22
Export Text for android

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
Exported text variants for android, and used in tester.

### Verification
Dark:
DarkText

Light:
LightText

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/text/src/Variants.android.ts
```diff
@@ -0,0 +1,45 @@
...
+export const Body1Strong = Text.customize({
+  variant: 'body1Strong',
+} as any);
+export const Body2 = Text.customize({
+  variant: 'body2',
+} as any);
+export const Body2Strong = Text.customize({
+  variant: 'body2Strong',
+} as any);
+export const Subtitle1 = null; // Not supported on iOS
```
amgleitman, line 26: Copying the variants over from iOS is a good rough estimate, but if we have a definitive set of tokens for Android typography, we should use those.

ankraj12: Validated, the same design variants exist for android - I copied this to a common file for iOS and Android.

lyzhan7:
When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync. 

With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work. 

Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same

Saadnajmi:
> When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync. 
> 
> With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work. 
> 
> Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same

Having a ".mobile" file that the .ios and .android is an approach we've taken before and been a useful one. We use it FAB and Activity indicator iirc. I don't have issues with it.

rurikoaraki:
> > When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync.
> > With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi
> >             
> >             Saad Najmi
> >             FTE do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work.
> > Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same
> 
> Having a ".mobile" file that the .ios and .android is an approach we've taken before and been a useful one. We use it FAB and Activity indicator iirc. I don't have issues with it.

We might also be able to edit the bundler to look for .mobile as part of a mobile platform. I think there's a client who does that.

apps/fluent-tester/src/TestComponents/Text/TextTest.tsx
```diff
@@ -30,7 +30,7 @@ export const TextTest: React.FunctionComponent = () => {
     uwpStatus: 'Experimental',
     iosStatus: 'Experimental',
     macosStatus: 'Experimental',
-    androidStatus: 'Beta',
+    androidStatus: 'Experimental',
```
rurikoaraki, line 33: This is the wrong test folder for Text V1. Please look at TextExperimental instead.

Saadnajmi:
> > > When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync.
> > > With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi
> > > ```
> > >         Saad Najmi
> > >         FTE do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same
> > 
> > 
> > Having a ".mobile" file that the .ios and .android is an approach we've taken before and been a useful one. We use it FAB and Activity indicator iirc. I don't have issues with it.
> 
> We might also be able to edit the bundler to look for .mobile as part of a mobile platform. I think there's a client who does that.

That sounds like something the end app/client would have to do and not a component library like us can do though? I might be wrong, I'm not 100% sure about bundlers. Regardless, I think how we have it set up we're safe with Metro and the standard RN flow.

rurikoaraki:
> > > > When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync.
> > > > With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi
> > > > ```
> > > >         Saad Najmi
> > > >         FTE do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same
> > > 
> > > 
> > > Having a ".mobile" file that the .ios and .android is an approach we've taken before and been a useful one. We use it FAB and Activity indicator iirc. I don't have issues with it.
> > 
> > 
> > We might also be able to edit the bundler to look for .mobile as part of a mobile platform. I think there's a client who does that.
> 
> That sounds like something the end app/client would have to do and not a component library like us can do though? I might be wrong, I'm not 100% sure about bundlers. Regardless, I think how we have it set up we're safe with Metro and the standard RN flow.

That's fair, forgot that downstream clients can have different bundler settings that affect our code.

ankraj12:
> > > > > When iOS/android share code we've usually just duplicated files, i.e. have one file for iOS and one for Android. The benefit of this is that .ios/.android files will only be included for the relevant platform, ex. .ios files will be included on the iOS platform only, which saves memory - but the tradeoff is that we end up with duplicated code and sometimes we end up having to keep multiple files in sync.
> > > > > With the approach you took here there's less duplicated code but I think the V2Usage.mobile.tsx file will be available on every platform, not just iOS/android. @Saadnajmi
> > > > > ```
> > > > >         Saad Najmi
> > > > >         FTE do you have any thoughts on this approach? I was actually thinking of doing something similar for some of the token work.
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > Also I guess this approach is only something we can do when iOS/Android code are the exact same...I guess a lot of the places where we have a separate .ios vs. .android file, the code is similar but not exactly the same
> > > > 
> > > > 
> > > > Having a ".mobile" file that the .ios and .android is an approach we've taken before and been a useful one. We use it FAB and Activity indicator iirc. I don't have issues with it.
> > > 
> > > 
> > > We might also be able to edit the bundler to look for .mobile as part of a mobile platform. I think there's a client who does that.
> > 
> > 
> > That sounds like something the end app/client would have to do and not a component library like us can do though? I might be wrong, I'm not 100% sure about bundlers. Regardless, I think how we have it set up we're safe with Metro and the standard RN flow.
> 
> That's fair, forgot that downstream clients can have different bundler settings that affect our code.

I see `V2Usage.win32.tsx`, 
`V2Usage.ios.tsx`, `V2Usage.android.tsx`, `V2Usage.tsx` and `V2Usage.mobile.tsx`. My understanding was that mobile endpoints should bundle `V2Usage.ios.tsx` and `V2Usage.android.tsx`, win32 should use *.win32 and macOS should use the stubbed *.tsx.  The *.mobile 'might' not end up in any bundles, but I could be wrong on this part.  
I think this is any okay approach for code sharing between iOS and Android where the files are common to avoid huge amounts of duplicates. If there is any other magic around configurations that Ruriko mentioned possible, that would be a good addition on top.

Saadnajmi:
@ankraj12 it's basically what is documented here https://reactnative.dev/docs/0.63/platform-specific-code

What you said is correct. .mobile will not auto map to any platform, but it's a convention we've used where we import it into .ios and . android .

ayush547 (approved):

apps/fluent-tester/src/TestComponents/TextExperimental/V2Usage.mobile.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <TextV1 variant="caption2">Caption2sdkjskdj v1</TextV1>
+        <Caption2>Caption2 v2</Caption2>
+        <TextV1 variant="caption1">Caption1 v1</TextV1>
+        <Caption1>Caption1 v2</Caption1>
+        <TextV1 variant="caption1Strong">Caption1Strong v1</TextV1>
+        <Caption1Strong>Caption1Strong v2</Caption1Strong>
+        <TextV1 variant="body2">Body2 v1</TextV1>
+        <Body2>Body2 v2</Body2>
+        <TextV1 variant="body2Strong">Body2Strong v1</TextV1>
+        <Body2Strong>Body2Strongv2 </Body2Strong>
```
rohanprasadofficial, line 34: nit : spacing between

rohanprasadofficial:
just a quick question, we're comparing v1, v2 here, why is that? also they look exactly the same 🤔

ankraj12:
> just a quick question, we're comparing v1, v2 here, why is that? also they look exactly the same 🤔

It's intended to demonstrate that the variants work similar in the usages. They are expected to look same.

PR #2456, opened 2022-12-22
Add spacing test page and E2E tests for other platforms

amgleitman:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

As a follow-up to #2446, we should also add the test page for the rest of the platforms.

### Verification

If the CI passes, this should be good :-)

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki (approved):

Saadnajmi (commented):
Do we need an iOS spec?

lyzhan7:
> Do we need an iOS spec?

I think in #2446 it was added to iOS, this PR is adding it for other platforms

lyzhan7 (approved):

Saadnajmi:
> > Do we need an iOS spec?
> 
> I think in #2446 it was added to iOS, this PR is adding it for other platforms

Ah, good catch! Thanks =)

Saadnajmi (approved):

PR #2457, opened 2022-12-23
[Switch] Animation using Animated API , Removing LayoutAnimation

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
This PR targets to make work animation on switch using Animated API instead of Layout Animations. 
Layout Animations are very heaving and have effect on parent as well which using the Switch , there were some known bug like controlled switch was not working with Layout Animations. 

**Note :** 
This is only target Android for now but it will work on iOS , I have tested the animation on iOS as well. 

### Verification
Visual Verification. 

Added the screenshot on both Android and iOS. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rohanprasadofficial:
Animation work with Animated API as well on iOS. 
I tested it on iOS tester app with some broken tokens but animation looks good. When iOS tokens are in , seems it should look fine , maybe some small changes required. 


https://user-images.githubusercontent.com/30728574/209326709-4dbfb1ac-65c4-4f13-9595-9a9dd91a7d3c.mov

rohanprasadofficial:
Screen recording for Android - Switch Animation 

https://user-images.githubusercontent.com/30728574/209326909-89ad145c-1db0-4dfc-a789-671effaf826d.mov

rohanprasadofficial:
Latest screenshot with correct token values : 

| Light Mode | Dark Mode |
|--------------|-----------|
| Screenshot_1672204526 | Screenshot_1672204567 |

packages/experimental/Switch/src/Switch.tsx
```diff
@@ -29,19 +31,108 @@ export const switchLookup = (layer: string, state: SwitchState, userProps: Switc
...
+          if (checked) {
+            startTrackBackgroundAnimation(checked, trackBackgroundAnimation);
+            startTrackAnimation(false, animationConfig, animation, checked);
+          } else {
+            startTrackBackgroundAnimation(checked, trackBackgroundAnimation);
+            startTrackAnimation(false, animationConfig, animation, checked);
+          }
+        }
+      },
+      [userProps.onChange],
```
rurikoaraki, line 98: should include trackBackgroundAnimation, animation

packages/experimental/Switch/src/Switch.tsx
```diff
@@ -29,19 +31,108 @@ export const switchLookup = (layer: string, state: SwitchState, userProps: Switc
...
+    );
+
+    //Setting the initial position of the know on track when page loads.
+
+    // For Mobile platform we are passing extra data to useSwitch for Aninated API
+    const switchInfo = useSwitch({ ...userProps, onChange: onChangeWithAnimation });
+
+    const switchAnimationStyles = {
```
rurikoaraki, line 106: should be memo'ed

packages/experimental/Switch/src/Switch.tsx
```diff
@@ -29,19 +31,108 @@ export const switchLookup = (layer: string, state: SwitchState, userProps: Switc
...
+
+    const animationConfig: IAnimationConfig = {
+      toggleOnBgColor: switchOnSlot.track({}).props.style.backgroundColor,
+      toggleOffBgColor: switchOffSlot.track({}).props.style.backgroundColor,
+      trackWidth: switchOnSlot.track({}).props.style.width,
+      thumbWidth: switchOnSlot.thumb({}).props.style.width,
+      thumbMargin: switchOnSlot.thumb({}).props.style.margin,
+    };
+
+    const [animation] = React.useState(new Animated.Value(0));
```
rurikoaraki, line 82: I'd prefer if all of the hook calls were under useSwitch so that they're all in one place.

packages/experimental/Switch/src/Switch.tsx
```diff
@@ -29,19 +31,108 @@ export const switchLookup = (layer: string, state: SwitchState, userProps: Switc
...
     toggleContainer: View,
     onOffText: Text,
   },
   useRender: (userProps: SwitchProps, useSlots: UseSlots<SwitchType>) => {
-    const switchInfo = useSwitch(userProps);
+    const switchOnSlot = useSlots(userProps, (layer) => switchLookup(layer, { toggled: true, disabled: userProps.disabled }, {}));
+    const switchOffSlot = useSlots(userProps, (layer) => switchLookup(layer, { toggled: false, disabled: userProps.disabled }, {}));
+
+    const animationConfig: IAnimationConfig = {
+      toggleOnBgColor: switchOnSlot.track({}).props.style.backgroundColor,
```
rurikoaraki, line 75: This seems like the Switch will do a lot of work in order to get at the token values. Compressible is probably better for this kind of thing.

Since the value shouldn't change between renders (assuming Switch has no tokensThatAreAlsoProps), you could probably make this once and then just reuse it on every render, minimizing the times that you call React.createElement

rohanprasadofficial: just a quick question here: I tried with tokensThatAreAlsoProps , seems the token specified also becomes a props and user can pass any value to props that is taken by component as token change and is applied to the token. 

When I try to access it in ,tsx file it doesn't comes as part of userProps or final props. 🤔  , correct me if I am anything wrong.

rurikoaraki: Yeah, your understanding is correct, `tokensThatAreAlsoProps` are component tokens that are exposed as props so that users can change the token value without using `customize`. This is helpful for tokens that are expected to be customized frequently, such as the background color of an Avatar.

The props themselves should be coming through as part of userProps, and are accessed within useStyling when calling useSlots, if I remember correctly.

The way you're doing it now is, I think, the only way to access a token value that is not passed in as a prop. This is where compressible is a better choice as a framework, since you can get direct access to token values inside the render function. You can look at Text as an example.

rurikoaraki: I guess actually you can't assume that the token values will stay constant throughout an instance of the component since interactions can change things like backgroundColor. So I guess this is the way that it has to be if using compose :\

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: Actually , yes at a particular point while working on this , I had everything on the useSwitch browse here :

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: Let me try using compressible, seems there're few hacks that is current code of this PR but again it has to do lot of work to just the token value.

rohanprasadofficial: Also current code having some issue with re-rendering and styling seems the tests starts failing.

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: test failiong
 moved everything into useSwitch. Seems 2 test fails.  
 
 
 
 But upon making animatedStyles into a callback function like this :
   const switchAnimationStyles = React.useCallback(() => {
    return {
      // transform over toggled on position to toggled off position and vice versa
      thumbAnimatedStyle: {
        transform: [
          {
            translateX: animation,
          },
        ],
      },
      // Interpolate over toggled on color to toggled off color and vice versa
      trackBackgroundStyle: {
        backgroundColor: trackBackgroundAnimation.interpolate({
          inputRange: [0, 1],
          outputRange: checkedState
            ? [animationConfig.toggleOnBgColor, animationConfig.toggleOnBgColor]
            : [animationConfig.toggleOffBgColor, animationConfig.toggleOffBgColor],
        }),
      },
    };
  }, []); 
  
  
  Passes the test but the animation doesn't work anymore.

packages/experimental/Switch/src/Switch.types.ts
```diff
@@ -126,6 +126,14 @@ export interface SwitchTokens extends LayoutTokens, FontTokens, IBorderTokens, I
   above?: SwitchTokens;
 }
 
+export interface IAnimationConfig {
```
rurikoaraki, line 129: Interface names don't need to start with `I` anymore

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -45,12 +40,78 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
...
+      backgroundColor: trackBackgroundAnimation.interpolate({
+        inputRange: [0, 1],
+        outputRange: checkedState
+          ? [animationConfig.toggleOnBgColor, animationConfig.toggleOnBgColor]
+          : [animationConfig.toggleOffBgColor, animationConfig.toggleOffBgColor],
+      }),
+    },
+  };
+
+  //Setting the initial position of the know on track when page loads.
```
rurikoaraki, line 104: typo: know -> knob, I think?

packages/experimental/Switch/src/Switch.tsx
```diff
@@ -29,19 +29,34 @@ export const switchLookup = (layer: string, state: SwitchState, userProps: Switc
...
   },
   useRender: (userProps: SwitchProps, useSlots: UseSlots<SwitchType>) => {
-    const switchInfo = useSwitch(userProps);
+    const switchOnSlot = useSlots(userProps, (layer) => switchLookup(layer, { toggled: true }, {}));
+    const switchOffSlot = useSlots(userProps, (layer) => switchLookup(layer, { toggled: false }, {}));
+
+    // For Mobile platform we are passing extra data to useSwitch for Animated API
+    const switchInfo = useSwitch(
+      userProps,
+      isMobile && {
```
Saadnajmi, line 52: I think using `Animated` for the animation would also benefit Windows and macOS, so doesn't have to be constrained to mobile. More, "not-win32".

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -45,12 +40,78 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
...
+  if (isMobile) {
+    if (checkedState) {
+      startTrackAnimation(false, animationConfig, animation, checkedState);
```
rurikoaraki, line 107: It looks like you're calling the animations during the render phase, not in response to a state change like checkedState changing or toggle being called?

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -6,25 +6,20 @@ import {
...
-      UIManager.setLayoutAnimationEnabledExperimental(false);
-    });
-  }
-};
+const isMobile = Platform.OS === 'android' || Platform.OS == 'ios';
 
-export const useSwitch = (props: SwitchProps): SwitchInfo => {
+export const useSwitch = (props: SwitchProps, animationConfig?: IAnimationConfig): SwitchInfo => {
   const defaultComponentRef = React.useRef(null);
+  const [animation] = React.useState(new Animated.Value(0));
```
rurikoaraki, line 20: Wouldn't you need to be able to change the value of the animation state at some point? Does the Animation API do that for you?

The example on the RN website uses refs to track the current state: https://reactnative.dev/docs/animated

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: added

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: Taking it to compressible might take some time as there are many things on switch to tackle with animations.

packages/experimental/Switch/src/useSwitch.ts
rohanprasadofficial: correct.

packages/experimental/Switch/src/Switch.types.ts
rohanprasadofficial: Removed

packages/experimental/Switch/src/useSwitch.ts
rohanprasadofficial: animation value is changed using the apis such as interpolate or timing provided by AnimatedAPI , setting up toValue property. 
I think it's better to use useRef for now then.

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: @Saadnajmi , agree on that point. `Animated` work cross-plat but I would like to take it forward with mobile first in its first implementation. 
There're couple of things are need from the design and for mobile it has clear definition, as animating using Animated is doing using `transform` not given the react native to handle such as Layout Animation does.  

There will few tokens changes/value might require on win32 as they're used in calculations for x-axis transformation according to padding between track * knob and width of the track and other things such as speed & duration of the time, hence it would be great if we could do that in a separate PR. For iOS i will test and add screen recordings.

rohanprasadofficial:
Latest screenshot with correct token values : 

| Dark Mode | Light Mode |
|--------------|-----------|
| Screenshot_1672827239 | Screenshot_1672827232 |


## IPhone/iOS screenshot with incorrect token : 

Screenshot 2023-01-04 at 4 45 30 PM

rohanprasadofficial:
## IPhone/iOS screen current behaviour with incorrect tokens. 

https://user-images.githubusercontent.com/30728574/210544281-780323ed-2d5f-4351-a4a5-19deb44908d2.mov

## Android screen current behaviour with correct tokens. 

https://user-images.githubusercontent.com/30728574/210544286-8943b9c0-2257-4203-be79-ad118a380148.mov

packages/experimental/Switch/src/__tests__/Switch.test.tsx
```diff
@@ -13,9 +16,11 @@ it('Switch Disabled', () => {
   expect(tree).toMatchSnapshot();
 });
 
-it('Simple switch rendering does not invalidate styling', () => {
-  checkRenderConsistency(() => <Switch label="Default Switch" />, 2);
-});
+// Temporary commented out to check on Animation , seems AnimatedAPI brings difference in render tree.
```
rurikoaraki, line 19: Yeah, I think this is expected with Animated. We can remove this test.

packages/experimental/Switch/src/useSwitch.ts
rurikoaraki: It probably makes sense to put this in a useEffect hook that depends on checkedState changing instead of running it on every render

packages/experimental/Switch/src/Switch.tsx
Saadnajmi: Fair enough if you want to keep it limited to the platforms you've tested.

rurikoaraki:
Tested on win32, behavior looks good

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -45,12 +40,91 @@ export const useSwitch = (props: SwitchProps): SwitchInfo => {
   const onChangeWithAnimation = React.useCallback(
     (e: InteractionEvent, checked?: boolean) => {
       onChange && onChange(e, checked);
-      animateSwitchKnob();
+      if (isMobile) {
```
rurikoaraki, line 43: With line 118 this seems like it would run more often than it should, we should probably remove the duplicate. I think running the animation in a useEffect hook would be best since that would cover controlled switches, too.

packages/experimental/Switch/src/__tests__/Switch.test.tsx
rohanprasadofficial: Removed.

packages/experimental/Switch/src/useSwitch.ts
rohanprasadofficial: Moved into useEffect() , animation runs good.

packages/experimental/Switch/src/useSwitch.ts
rohanprasadofficial: Moved into useEffect()

rohanprasadofficial:
> Tested on win32, behavior looks good

Thanks for testing it @rurikoaraki , really appreciate your time 😃

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: Moved to useSwitch.

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: done

packages/experimental/Switch/src/Switch.tsx
rurikoaraki: Yeah, for now this is fine.

rurikoaraki (approved):

packages/experimental/Switch/src/Switch.tsx
rohanprasadofficial: Sure , marking this resolved.

PR #2459, opened 2022-12-25
Exporting E2E Data as NuGet Package

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

The purpose of this PR is to create a NuGet package that exports all E2E testing data (mainly, everything we need to consume the spec files).

Previously, we had exported a JS Bundle containing all the specs. However, we need to give the consumer the ability to split up the spec files (the JS bundle executed every spec file - could not run any individually). Exporting the whole package allows the consumer to pick and choose which specs to run.

This requires moving all E2E testing constants from /apps/fluent-tester to /apps/E2E (hence the amount of files changed)

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/package.json
```diff
@@ -45,34 +44,5 @@
...
-        },
-        "typescriptValidation": false
-      }
-    ],
-    "capabilities": [
-      "core",
-      "react",
-      "metro-config",
-      "babel-preset-react-native"
-    ]
```
ksiler, line 76: Is there a reason why we're removing the entire rnx-kit config here? I would expect you'd just want to remove the "bundle" portion.

samuelfreiberg: Initially, I thought that rnx-kit was strictly for bundling in this package, but looking back, I am wrong. Adding back in newest commit (without bundle portion). Thanks

ksiler (approved):

PR #2461, opened 2022-12-27
RadioGroup Narrator Selection State Bug Fix

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Disabling keyboarding on the native side for Arrow Left/Right/Up/Down and handling those keyboard events on the JS side. This way, the selection state is set before focus is set so that narrator receives the correct state before narrating. 

Code Changes:
- The values array is populated to hold all of the Radio children values of a RadioGroup and the enabledValues array is populated to hold all of the clickable Radio children values of a RadioGroup (essentially excluding any invalid or disabled Radios). The values array is used for accessibilitySetSize and enabledValues is used for win32 directional arrow keys keyboarding logic, which is now moved to the JS-side. Originally, these arrays were populated in RadioGroup.tsx upon render using the list of children React identifies. However, when there are View components or the Radio children are wrapped in a View, React would not identify the children causing the values array to be populated as undefined or empty. This was fixed by populating the values array at mount and un-mount time directly from the Radio component and storing the arrays in the context.
- Commented out Radio rendering jest test due to a common issue of incorrect memoization of keyDownEvents prop. Will re-visit and uncomment this test once this issue is resolved.
- Added platform specific win32 useRadio file that contains keyboarding logic to be handled on the JS-side. The focus is set after the selection state is set.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -51,10 +52,59 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+        e.nativeEvent.key == 'ArrowDown' ||
+        e.nativeEvent.key == 'ArrowRight' ||
+        e.nativeEvent.key == 'ArrowUp' ||
+        e.nativeEvent.key == 'ArrowLeft'
+      ) {
+        const length = radioGroupContext.enabledValues.length;
+        const currRadioIndex = radioGroupContext.enabledValues.indexOf(radioGroupContext.value);
+        let newCurrRadioIndex;
+        if (e.nativeEvent.key === 'ArrowDown' || e.nativeEvent.key == 'ArrowRight') {
+          if (isCircularNavigation || !(currRadioIndex + 1 == length)) {
```
samuelfreiberg, line 69: If circularNav is always true, do we need to check for it here? That's mostly FocusZone code. Since RG always has circularNav, I don't think it's necessary

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -51,10 +52,59 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+            newCurrRadioIndex = (currRadioIndex + 1) % length;
```
samuelfreiberg, line 70: Since we know (currRadioIndex + 1) is less than length, do we need the %? I don't think it changes anything?

When X < Y, X % Y = X always (I believe?)

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -51,10 +52,59 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+            radioGroupContext.value = radioGroupContext.enabledValues[newCurrRadioIndex];
```
samuelfreiberg, line 71: Are we allowed to explicitly change RadioGroup context values? I'm not fresh on how Context works, but on the two lines below, we're using a callback so the RadioGroup itself can change the values. Here, we're explicitly setting radioGroupContext.value.

Maybe we try making a `changeValue(newVal)` function in the context and call that here?

Again, not sure entirely if this is correct or not, if you think it's ok, we can leave it

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -51,10 +52,59 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+      if (
+        e.nativeEvent.key == 'ArrowDown' ||
+        e.nativeEvent.key == 'ArrowRight' ||
+        e.nativeEvent.key == 'ArrowUp' ||
+        e.nativeEvent.key == 'ArrowLeft'
+      ) {
+        const length = radioGroupContext.enabledValues.length;
+        const currRadioIndex = radioGroupContext.enabledValues.indexOf(radioGroupContext.value);
+        let newCurrRadioIndex;
+        if (e.nativeEvent.key === 'ArrowDown' || e.nativeEvent.key == 'ArrowRight') {
```
ksiler, line 68: Replace "==" with "===" for 'ArrowRight'

Saadnajmi, line 68: Does this respect RTL?

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -51,10 +52,59 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+  const isCircularNavigation = true;
+
+  const onInvoke = React.useCallback(
+    (e: KeyPressEvent) => {
+      if (
+        e.nativeEvent.key == 'ArrowDown' ||
+        e.nativeEvent.key == 'ArrowRight' ||
+        e.nativeEvent.key == 'ArrowUp' ||
+        e.nativeEvent.key == 'ArrowLeft'
+      ) {
```
ksiler, line 64: Is this if statement necessary? I would just do `if (e.nativeEvent.key === 'ArrowDown' || e.nativeEvent.key === 'ArrowRight')` and `else if (e.nativeEvent.key === 'ArrowUp' || e.nativeEvent.key === 'ArrowLeft')` down below.

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
```diff
@@ -31,6 +31,12 @@ export interface RadioGroupState extends RadioGroupProps {
    * Array of radio keys in the group
    */
   values?: string[];
+
+  /**
+   * Array of enabled radio keys in the group
+   * Windows-Specific Prop.
```
Saadnajmi, line 37: Nit, the convention is more like `@platform win32`

packages/experimental/RadioGroup/src/Radio/Radio.tsx
```diff
@@ -4,7 +4,7 @@ import { radioName, RadioType, RadioProps } from './Radio.types';
 import { TextV1 as Text } from '@fluentui-react-native/text';
 import { stylingSettings } from './Radio.styling';
 import { compose, mergeProps, withSlots, UseSlots } from '@fluentui-react-native/framework';
-import { useRadio } from './useRadio';
+import { useRadio } from './useRadio.win32';
```
Saadnajmi, line 7: Imports should generally not have the platform suffix imo. I'm not sure what this will do on another platform like macOS where ".win32" files might be filtered out. 

But also, does this need to be win32? I didn't see that much platform specific logic in it.

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: That's a good point, win32 should have circular navigation so I can remove that check here. Updated, thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Reworked the logic a little bit by removing the check for length, so I'll need to keep the % for circular navigation to work correctly.

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Replaced, thanks!

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Thanks for the suggestion, simplified!

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
gulnazsayed: Replaced, thanks!

packages/experimental/RadioGroup/src/Radio/Radio.tsx
gulnazsayed: Reverted the files so that useRadio is the only one with win32 specific code because that's where the new JS keyboarding logic lives. The old useRadio code is still available for other platforms (like macOS) to use, where native is still handling keyboarding for arrow keys. Thanks for pointing this out!

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Other controls, like Tab, have a few instances where they explicitly change context value. However, in this case, the RadioGroup context value is updated with the onChange callback so I removed this line as it is redundant. Thanks for pointing this out!

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Based on my understanding, I think yes because arrow key behavior would remain the same regardless of RTL since Radio ordering would remain the same.

Saadnajmi: If the script (and label and all the views..) was reversed, you wouldn't want arrowLeft to go to the next radio button?

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -0,0 +1,177 @@
...
+  React.useEffect(() => {
+    radioGroupContext.addRadioValue(value);
+
+    return () => {
+      radioGroupContext.removeRadioValue(value);
+    };
+  }, []);
+
+  // Explicitly only run on mount and unmount
+  React.useEffect(() => {
```
rurikoaraki, line 70: Seems like you could run both `radioGroupContext.addRadioValue(value);` and `radioGroupContext.addRadioEnabledValue(value);` in the same `useEffect`

rurikoaraki (approved):

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
rurikoaraki: +1, I think we should double check RTL behavior.

rurikoaraki (changes requested):
I;d double check RTL behavior w.r.t left and right arrows

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Fixed, thanks!

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
```diff
@@ -27,10 +27,41 @@ export interface RadioGroupState extends RadioGroupProps {
    */
   selectedButtonRef?: React.MutableRefObject<any>;
 
+  /**
+   * Populate the values arrays
```
samuelfreiberg, line 31: What are value arrays? What values are being added/removed from the Radio's? 

(clarify comments :) )

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
```diff
@@ -27,10 +27,41 @@ export interface RadioGroupState extends RadioGroupProps {
...
   /**
    * Array of radio keys in the group
    */
   values?: string[];
+
+  /**
+   * Populate the enabledValues arrays
+   * @platform win32
+   */
+  addRadioEnabledValue?: (value: string) => void;
```
samuelfreiberg, line 45: What's the difference between `addRadioEnabledValue()` and `addRadioValue()`? I'd like the comments to be a little more in-depth

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.tsx
```diff
@@ -56,17 +56,6 @@ export const RadioGroup = compose<RadioGroupType>({
         </Slots.label>
       );
 
-      // Populate the buttonKeys array
-      if (children) {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore - TODO, fix typing error
-        contextValue.values = React.Children.map(children, (child: React.ReactChild) => {
```
samuelfreiberg, line 63: It might be helpful to a reviewer to note this change in the PR description rather than having to read through the code and figure out why we're deleting this important bit of code.

I try to always note all changes in the PR description in the case that a reviewer is confused by a block of code, they can go to the description and figure out the change as a whole.

gulnazsayed: Thanks for the feedback! Updated the PR description.

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -0,0 +1,171 @@
...
+      radioGroupContext.removeRadioEnabledValue(value);
+    };
+  }, []);
+
+  const onInvoke = React.useCallback(
+    (e: KeyPressEvent) => {
+      const length = radioGroupContext.enabledValues.length;
+      const currRadioIndex = radioGroupContext.enabledValues.indexOf(radioGroupContext.value);
+      let newCurrRadioIndex;
+      if (e.nativeEvent.key === 'ArrowDown' || e.nativeEvent.key === 'ArrowRight') {
```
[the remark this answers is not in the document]
gulnazsayed: Thanks for the suggestion! I created an enum in the same file for now since I couldn't find a more central place to put it.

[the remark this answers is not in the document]
Saadnajmi: Good suggestion that maybe this type can move up to React Native so we can reference the enum / union from there.

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
gulnazsayed: Thanks for the suggestion! The values array is an array of all of the Radios' keys. Updated the comments with a little more description.

packages/experimental/RadioGroup/src/RadioGroup/RadioGroup.types.ts
gulnazsayed: `addRadioEnabledValue()` adds the Radio key to the `enabledValues` array and `addRadioValue()` adds the Radio key to the values array. Updated the description!

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
gulnazsayed: Thanks for the callout! Was able to check for RTL and verify changes in FURN tester app via Lime. It should work correctly now.

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -78,17 +78,25 @@ export const useRadio = (props: RadioProps): RadioInfo => {
...
+
   const onInvoke = React.useCallback(
     (e: KeyPressEvent) => {
       if (e.nativeEvent.key in DirectionalArrowKeys) {
         const length = radioGroupContext.enabledValues.length;
+        const previous =
+          (!isRTL && (e.nativeEvent.key === DirectionalArrowKeys.ArrowUp || e.nativeEvent.key === DirectionalArrowKeys.ArrowLeft)) ||
+          (isRTL && (e.nativeEvent.key === DirectionalArrowKeys.ArrowDown || e.nativeEvent.key === DirectionalArrowKeys.ArrowRight));
+        const next =
+          (isRTL && (e.nativeEvent.key === DirectionalArrowKeys.ArrowUp || e.nativeEvent.key === DirectionalArrowKeys.ArrowLeft)) ||
```
rurikoaraki, line 91: RTL shouldn't reverse up/down, just left/right

rurikoaraki, line 91: RTL shouldn't reverse up/down, just left/right

gulnazsayed: Fixed, thanks!

rurikoaraki (approved):

PR #2463, opened 2022-12-27
[Menu] Mark certain key events as handled for MenuPopover

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In the MenuPopover, certain key events are automatically handled by native, but the key event gets fired on the JS side anyway (this is by design for RN). This can cause bugs as keys could get double handled if a user has a separate key handler set up.

Since the key events are handled on native, it would make sense to automatically mark those key events as handled. That way, they don't get handled on the JS side, too.

### Verification

Tested change against client code

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -47,6 +50,42 @@ export const useMenuPopover = (_props: MenuPopoverProps): MenuPopoverState => {
...
+      // Mark key events that move selection as handled.
+      // These key events are handled on the native side.
+      switch (e.nativeEvent.key) {
+        case 'ArrowUp':
+        case 'ArrowDown':
+        case 'Tab':
+        case 'Home':
+        case 'End':
+          e.stopPropagation();
+      }
```
Saadnajmi, line 66: Might be worth putting these in a `Platform === 'win32'` check, if just for clarity. The events only bubble on the JS side on win32. On macOS these events never get sent to JS, as they are handled natively by FocusZone. 

I'm a little confused about the win32 side though. iirc, every keyboard event is sent to JS, and MenuV1 doesn't do any custom handling for any of these events. Instead they are handled by native netui, which moves focus, which moves selection. 

Is the intention that the client can override these events on Popover to do something on the JS side, but also still have native do it's thing?

Saadnajmi: Without calling `e.stopPropogation()`, does the event just keep walking up the JS tree even if native already handled it? That seems bad..

Saadnajmi (approved):

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
Saadnajmi: I take back comment on platform check. If anything this is extra safety so yay

rurikoaraki: > Without calling `e.stopPropogation()`, does the event just keep walking up the JS tree even if native already handled it? That seems bad..

Yeah, this is basically what's happening.

PR #2416, opened 2022-12-13
[Checkbox - Android] Ripple addition and review feedbacks

rohanprasadofficial:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
This PR tackles few changes for enabling ripple on Android and address feedbacks on PR https://github.com/microsoft/fluentui-react-native/pull/2346  

- apps/fluent-tester/src/TestComponents/CheckboxExperimental/* - Changes on unsupported warning and removal for sections 
- packages/components/Checkbox/SPEC.md
- packages/components/Checkbox/SPEC.md - Changes from Unsupported on Android to Unsupported on Mobile 
- packages/components/Checkbox/src/Checkbox.styling.ts - changes for ripple and props non removal.   
### Verification

**Note : Replacement of the hard coded ripple color would be done as a part of this issue : https://github.com/microsoft/fluentui-react-native/issues/2429**   

(how the change was tested, including both manual and automated tests)

| Light                                       | Dark                                      |
|----------------------------------------------|--------------------------------------------|
|MicrosoftTeams-image (10) | MicrosoftTeams-image (9)|

### Pull request checklist

This PR has considered (when applicable):
- [ ] Automated Tests
- [x] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #2466, opened 2022-12-28
Revert Checkbox back to using View instead of Pressable

ksiler:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This change reverts Checkbox back to using View instead of Pressable (which was introduced by #2416). This change caused a regression on win32 where the user is unable to toggle a checkbox via click if they clicked on the checkbox itself (not the label). Clicking on the label still toggles the checkbox which is why E2E test did not catch this. The click() function clicks the test element at the top center of the element. 

I'm working on adding an additional test which will test if we can toggle a checkbox via click on a checkbox without a label which should catch this regression if it ever occurs in the future. The additional test will come in a future commit.

This change was manually tested in the Fluent Tester on win32.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-checkbox-27225660-45fe-4215-aa25-6a06e0cbfc6d.json
+++ b/change/@fluentui-react-native-checkbox-27225660-45fe-4215-aa25-6a06e0cbfc6d.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Revert checkbox back to using View from Pressable",
+  "packageName": "@fluentui-react-native/checkbox",
+  "email": "krsiler@microsoft.com",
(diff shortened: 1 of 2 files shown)


ksiler:
Closing this in favor of doing a whole revert of #2416

rohanprasadofficial:
[ksiler](https://github.com/ksiler), can you verify on win32 on your end as well.

rohanprasadofficial:
Created a new PR for fixing this : https://github.com/microsoft/fluentui-react-native/pull/2467

PR #2467, opened 2022-12-28
[Checkbox] Fix regression on win32

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

https://github.com/microsoft/fluentui-react-native/pull/2416 , in this PR there was a change from View to Pressable to accommodate ripple behavior on Android as View doesn't support ripple. 

But there was an regression on win32 ,  https://github.com/microsoft/fluentui-react-native/pull/2466 . 
I have provided a fix for the same in this PR. 

### Verification
Ran win32 tester as well and verified with pre-existing behavior of keyboard navigation and click points. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Checkbox/src/Checkbox.tsx
+++ b/packages/components/Checkbox/src/Checkbox.tsx
@@ -54,7 +54,7 @@ export const Checkbox = compose<CheckboxType>({
       return (
         <Slots.root {...mergedProps} {...(Platform.OS == 'android' && { accessible: !disabled, focusable: !disabled })}>
           {Checkbox.state.labelIsBefore && labelComponent}
-          <Slots.checkbox {...(Platform.OS == 'android' && { onPress, disabled })} accessible={false} focusable={false}>
+          <Slots.checkbox accessible={false} onPress={onPress} disabled focusable={false}>
             <Slots.checkmark key="checkmark" viewBox="0 0 12 12">
               {checkmarkPath}
             </Slots.checkmark>
--- a/change/@fluentui-react-native-checkbox-a1ee3952-b272-40e5-9131-ba47501c063c.json
+++ b/change/@fluentui-react-native-checkbox-a1ee3952-b272-40e5-9131-ba47501c063c.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "fix regression on win32",
+  "packageName": "@fluentui-react-native/checkbox",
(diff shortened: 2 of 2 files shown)


rohanprasadofficial:
[ksiler](https://github.com/ksiler), can you verify on win32 on your end as well to make sure it doesn't break any existing features.

rohanprasadofficial:
seems there might be one more issue like when you click on checkbox, the whole area with label and checkbox is selected, is it expected?

ksiler:
@rohanprasadofficial thanks for the fix! I checked this on win32 and it looks good. Are you referencing the focus indicator that appears around the checkbox and label on click? I believe that's intentional. At least, that was the behavior I saw on checkbox before you made your changes for android so I think it's okay.

ksiler (approved):

rohanprasadofficial:
@ksiler ,  
Thanks for testing this :) 
there is small break on android but that might need more investigation , as we have same use-case on Radio Group as well to fix it on Android. 
For now we can merge and unblock win32. 


We might your help again sometime later to test win32 once we fix that.

ksiler:
@rohanprasadofficial thanks, would be happy to help test future changes.

packages/components/Checkbox/src/Checkbox.tsx
```diff
@@ -54,7 +54,7 @@ export const Checkbox = compose<CheckboxType>({
       return (
         <Slots.root {...mergedProps} {...(Platform.OS == 'android' && { accessible: !disabled, focusable: !disabled })}>
           {Checkbox.state.labelIsBefore && labelComponent}
-          <Slots.checkbox {...(Platform.OS == 'android' && { onPress, disabled })} accessible={false} focusable={false}>
+          <Slots.checkbox accessible={false} onPress={onPress} disabled focusable={false}>
```
Saadnajmi, line 57: Won't this make it always disabled, and therefore onPress won't fire?

Saadnajmi: Nevermind! This is fine =)

rohanprasadofficial: hmm correct as we extracted `disabled`, so it takes either undefined or true, hence makes it work.

PR #2364, opened 2022-11-28
Change jest endpoint to win32 for Checkbox and Switch

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Changing the jest endpoint to win32 for the aforementioned components, as they are mainly used in win32.
This consists of changing the config file to win32 and adding RN Win32 as a dev dependency to these packages.

The tests are failing due to two bugs, so this will stay a draft for now.

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #2468, opened 2022-12-28
Better types and tests for useKeyProps

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This PR makes the following improvements to `useKeyProps`:

1) Moves the associated types to platform specific type files. That should set us up better for when we enable Typescript's platform specific type checking
2) Combines the `useKeyProps.<platform>.ts` files into one. There was starting to be a lot of duplicated logic. Now that the types are separated out, I think the main file can be one combined one.
3) Adds some simple tests that ensure that we can use our hook without causing unnecessary extra renders
4) Removes some uses of a type from our deprecated copy of Pressability in favor of the _actual_ type from React Native Core / macOS / windows.

All in all, this is mostly a quality of life  improvement.

### Verification

CI should pass

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/utils/interactive-hooks/src/__tests__/useKeyProps.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+it('Simple Pressable with useKeyProps rendering does not invalidate styling', () => {
+  const keyboardProps = useKeyProps(dummyFunction, ' ', 'Enter');
+  checkRenderConsistency(() => <PressableWithDesktopProps {...keyboardProps} />, 2);
+});
+
+it('Pressable with useKeyProps re-renders correctly', () => {
+  const keyboardProps = useKeyProps(dummyFunction, ' ', 'Enter');
+  checkReRender(() => <PressableWithDesktopProps {...keyboardProps} />, 2);
+});
```
Saadnajmi, line 31: These tests pass... so I think https://github.com/microsoft/fluentui-react-native/pull/2364 is not actually blocked by more Pressable / useKeyProps changes. 

However, I did speak with @rurikoaraki and an alternate way to land that PR is to remove unnecessary uses of `useKeyProps` from some of our components. On Windows/win32, Pressable by default calls `onPress` with Space/Enter. That's the majority of our use case in FURN, so I can just remove them. I need to implement that default handling in macOS too though, which I'm doing here: https://github.com/microsoft/react-native-macos/pull/1614

rurikoaraki: This is passing in the result of useKeyProps and then using the same result twice to render. This isn't calling useKeyProps twice, which is what's causing the issue in #2364.

rurikoaraki: I think the test that would help is a test that calls useKeyProps twice with the same inputs and verifies that the output is the same object.

packages/utils/interactive-hooks/src/useKeyProps.ts
```diff
@@ -23,38 +29,106 @@ const noOp = () => {};
...
 
 /**
  * Re-usable hook for keyboard events. on macOS, this is onKeyDown, while on windows this is onKeyUp.
  * @param userCallback The function you want to be called once the key has been activated on key down
  * @param keys A string of the key you want to perform some action on. If undefined, always invokes userCallback
- * @returns KeyPressProps: An object containing the correct platform specific props to  handle key press
+ * @returns KeyPressProps: An object containing the correct platform specific props to handle key press
  */
-export const useKeyProps = noOp2;
+export const useKeyProps = memoize(getKeyUpPropsWorker);
```
rurikoaraki, line 128: Doesn't seem like this is following "on macOS, this is onKeyDown, while on windows this is onKeyUp." Is that intentional?

packages/utils/interactive-hooks/src/useKeyProps.ts
```diff
@@ -23,38 +29,106 @@ const noOp = () => {};
...
-  return noOp;
+export function useKeyCallback(userCallback?: KeyCallback, ...keys: string[]) {
+  const onKeyEvent = React.useCallback(
+    (e: KeyPressEvent) => {
+      if (userCallback !== undefined && (keys === undefined || keys.includes(e.nativeEvent.key))) {
+        userCallback(e);
+        e.stopPropagation();
+      }
+    },
+    [keys, userCallback],
```
Saadnajmi, line 40: Worth noting this used to be `...keys` and eslint complained, so I updated. That is _probably_ a functionality change.

packages/utils/interactive-hooks/src/useKeyProps.ts
Saadnajmi: Good catch, fixed. I guess the reason this didn't fail snapshots is none of our snapshots with keyboard events are using macOS as their platform (iirc, they're all win32).

packages/utils/interactive-hooks/src/__tests__/useKeyProps.test.tsx
Saadnajmi: I added that test. I'm somewhat confused why I can run a hook outside of a function component, but maybe jest tests are special?

packages/utils/interactive-hooks/src/useKeyProps.types.macos.ts
```diff
@@ -0,0 +1,33 @@
...
+  ArrowUp: boolean;
+  ArrowDown: boolean;
+  key: string;
+}
+
+export type KeyPressEvent = NativeSyntheticEvent<NativeKeyEvent>;
+
+export type KeyCallback = (e?: KeyPressEvent) => void;
+
+export type KeyPressProps = {
```
amgleitman, line 26: Why do we have Windows type definitions in a macOS-specific file?

packages/utils/interactive-hooks/src/useKeyProps.ts
amgleitman: What exactly was eslint complaining about?

packages/utils/interactive-hooks/src/useKeyProps.types.macos.ts
Saadnajmi: Not windows. I converted the macOS flow type to typescript and put it here. Post 0.71, I hope to import the type directly from RN-macOS

packages/utils/interactive-hooks/src/useKeyProps.ts
Saadnajmi: It can't statically check a spread operation, so it asked me to use "keys" instead of "...keys"

rurikoaraki (approved):

PR #2471, opened 2022-12-29
Update Compressible inline documentation

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Super simple change to add some inline documentation to compressible so that VS Code Intellisense can pick it up. The words I added are a rehash of the PR description of #819 (the PR that added compressible).

This is...  very little documentation, but better than nothing. 

### Verification

Hovering over compressible in VS Code works. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/framework/framework/src/useTokens.ts
+++ b/packages/framework/framework/src/useTokens.ts
@@ -5,6 +5,7 @@ import { TokenSettings } from './useStyling';
 
 export { applyTokenLayers, applyPropsToTokens, customizable, patchTokens } from '@fluentui-react-native/use-tokens';
 
+// A hook function to build a set of tokens from a passed in theme as well as a cache object
 export type UseTokens<TTokens> = UseTokensCore<TTokens, Theme>;
 
 export function buildUseTokens<TTokens>(...tokens: TokenSettings<TTokens>[]): UseTokens<TTokens> {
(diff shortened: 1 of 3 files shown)


packages/framework/framework/src/useTokens.ts
```diff
@@ -5,6 +5,7 @@ import { TokenSettings } from './useStyling';
 
 export { applyTokenLayers, applyPropsToTokens, customizable, patchTokens } from '@fluentui-react-native/use-tokens';
 
+// A hook function to build a set of tokens from a passed in theme as well as a cache object
```
amgleitman, line 8: Is it possible to add an example of how this might be invoked?

amgleitman (approved):

packages/framework/framework/src/useTokens.ts
Saadnajmi: I tried. It gets really big and wordy for VS Code. I think that would be better with a README.md somewhere. I'll look into that.

rurikoaraki (approved):

packages/framework/framework/src/useTokens.ts
Saadnajmi: I think I'll leave it as a follow-up. Any README with info on compressible should also have info on stagedComponent and compose. That becomes a bigger task than the simple purpose of this PR.

PR #2472, opened 2022-12-30
Use a V2 variant style for Notification's action button

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This simplifies the way tokens are applied to a V2 variant.

One thing to note is that the Notification snapshot has changed to not reflect letter spacing or line height. This is because our jest tests take styles from [the default theme package](https://github.com/microsoft/fluentui-react-native/blob/c225025e9ccfbf3e3b910796f1628062f8d3c30b/packages/theming/default-theme/src/defaultTheme.ts#L49), which doesn't define letter spacing or line height.

### Verification

* Confirmed that the appearance of Notification doesn't change on iOS.
* Confirmed that we're actually using these tokens. If we replace the letter spacing [here](https://github.com/microsoft/fluentui-react-native/blob/c225025e9ccfbf3e3b910796f1628062f8d3c30b/packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts#L159) with a modest positive value (such as `10`), that value gets clearly reflected in the test app as seen below.
  * As for why this spacing isn't reflected in "Listen to Emails...", that may be a separate issue. I believe it's worth investigating how we can ensure that our components are using the platform-appropriate styles instead of some random default.

Simulator Screen Shot - iPhone 14 Pro - 2022-12-29 at 17 40 40

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Notification/src/Notification.helper.tsx
+++ b/packages/components/Notification/src/Notification.helper.tsx
@@ -60,10 +60,6 @@ export const NotificationButton = stagedComponent((props: NotificationButtonProp
       backgroundColor: 'transparent',
       color: props.color,
       iconColor: props.color,
-      fontSize: 15,
-      fontWeight: '600',
-      fontLineHeight: 20,
-      fontLetterSpacing: -0.23, // iOS only prop
       disabled: {
         color: props.disabledColor,
       },
@@ -76,6 +72,7 @@ export const NotificationButton = stagedComponent((props: NotificationButtonProp
         minWidth: 0,
         padding: globalTokens.sizeNone,
         paddingHorizontal: globalTokens.sizeNone,
+        variant: 'body2Strong',
       },
     },
   });
--- a/apps/fluent-tester/ios/Podfile.lock
+++ b/apps/fluent-tester/ios/Podfile.lock
@@ -10,10 +10,10 @@ PODS:
     - React-jsi (= 0.68.5)
     - ReactCommon/turbomodule/core (= 0.68.5)
   - fmt (6.2.1)
-  - FRNAvatar (0.16.30):
+  - FRNAvatar (0.17.2):
     - MicrosoftFluentUI (= 0.10.0)
     - React
-  - FRNDatePicker (0.7.3):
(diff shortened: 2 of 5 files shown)


rurikoaraki (approved):

packages/components/Notification/src/Notification.helper.tsx
```diff
@@ -60,10 +60,6 @@ export const NotificationButton = stagedComponent((props: NotificationButtonProp
       backgroundColor: 'transparent',
       color: props.color,
       iconColor: props.color,
-      fontSize: 15,
-      fontWeight: '600',
-      fontLineHeight: 20,
-      fontLetterSpacing: -0.23, // iOS only prop
```
lyzhan7, line 66: Hmm shouldn't these props be getting replaced with something inside the 'subtle' object, rather than the 'medium" object?

lyzhan7:
I actually didn't know our jest tests take styles from the default theme rather than from the platform-specific theme that the jest config specifies - ex. for Notification, since jest config platform is set to iOS, I thought it would be using the apple theme.

packages/components/Notification/src/Notification.helper.tsx
amgleitman: If we put these values in 'subtle', they get overridden further down the line. It's similar to what we saw in #2448.

rurikoaraki:
> I actually didn't know our jest tests take styles from the default theme rather than from the platform-specific theme that the jest config specifies - ex. for Notification, since jest config platform is set to iOS, I thought it would be using the apple theme.

The defaultTheme doesn't use the apple theme. The apple theme is a separate package/theme that clients need to use explicitly  and feed into a ThemeProvider. Since these tests don't use a ThemeProvider, they fall back onto the defaultTheme. It might make sense to make the defaultTheme different per platform, but we need to do it in a way that won't bloat other platforms.

PR #2473, opened 2022-12-30
Rename test files to Legacy/V1 instead of using "experimental"

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The word "experimental" causes confusion as to what components and certain test pages are testing, so in an attempt to clarify the test pages I am renaming them to stop using the word "Experimental" and use "V1" instead.

All usages of "experimental" have been changed to "V1"
"Legacy" has been tacked on to the ends of components that also have a V1 test page to denote which page is testing the old version of the component.

The only functional change is that I removed a redundant Text E2E test. We have the same test running in the TextV1 page.

### Verification

Build passes
CI will verify E2E tests.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/ButtonV1/pages/ButtonV1PageObject.ts
```diff
@@ -4,9 +4,9 @@ import {
   BUTTON_NO_A11Y_LABEL_COMPONENT,
   HOMEPAGE_BUTTON_BUTTON,
   BUTTON_ON_PRESS,
-} from '../../Button/consts';
+} from '../../ButtonLegacy/consts';
```
ksiler, line 7: Should we create a consts file for ButtonV1?

ksiler:
I think you need to edit the RegExp in NativeTestingPageObject.win.ts to include numbers with the addition of 'V1' to some of the testIDs:

`const reg = new RegExp('Homepage_[a-zA-Z1-9]*_Button');`

apps/fluent-tester/src/TestComponents/TextV1/TextV1Test.tsx
```diff
@@ -32,11 +32,11 @@ const textSections: TestSection[] = [
   },
   {
     name: 'E2E Testing for Experimental Text',
```
ksiler, line 34: E2E Testing for TextV1

apps/E2E/src/ButtonV1/pages/ButtonV1PageObject.ts
ksiler: Same for IconV1

rurikoaraki: I would say that's out of the scope of this PR. For the Button, there's only one const page because there was only one test folder for the button where the consts used to live, so both legacy and V1 button consts were in the same file.

Saadnajmi (approved):

PR #2480, opened 2023-01-03
Refactor RadioGroup (legacy) tests and add better errors

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- RadioGroup (legacy)

### Verification

Tests pass in the CI and locally.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/RadioGroup/pages/RadioGroupPageObject.ts
```diff
@@ -24,52 +24,36 @@ class RadioGroupPage extends BasePage {
...
   }
 
-  async waitForRadioButtonSelected(radioButtonSelector: RadioButtonSelector, timeout?: number): Promise<void> {
-    await browser.waitUntil(async () => await this.isRadioButtonSelected(radioButtonSelector), {
-      timeout: timeout ?? this.waitForUiEvent,
-      timeoutMsg: 'RadioButton was not selected correctly.',
-      interval: 1000,
-    });
+  async isRadioButtonSelected(selector: RadioButton): Promise<boolean> {
+    return await (await this.getRadioButton(selector)).isSelected();
```
samuelfreiberg, line 35: You mentioned in a different PR that `isSelected()` didn't work on UWP. Is it the same here? Or does it work here?

apps/E2E/src/RadioGroup/pages/RadioGroupPageObject.ts
```diff
@@ -24,52 +24,36 @@ class RadioGroupPage extends BasePage {
...
   }
 
-  /* Sends a Keyboarding command on a specific UI element */
-  async sendKey(key: string, radioButtonSelector: RadioButtonSelector): Promise<void> {
-    await (await this.getRadioButton(radioButtonSelector)).addValue(key);
+  async waitForRadioButtonSelected(selector: RadioButton, errorMsg?: string): Promise<boolean> {
+    await this.waitForCondition(
+      async () => await this.isRadioButtonSelected(selector),
+      errorMsg ?? `RadioButton ${selector} failed to be selected.`,
```
samuelfreiberg, line 41: Let's make `errorMsg` a mandatory parameter here. We want to make sure the user passes in their specific scenario here

apps/E2E/src/RadioGroup/specs/RadioGroup.spec.win.ts
```diff
@@ -29,80 +28,133 @@ describe('RadioGroup/RadioButton Testing Initialization', function () {
 describe('RadioGroup/RadioButton Accessibility Testing', () => {
   /* Scrolls and waits for the RadioGroup to be visible on the Test Page */
   beforeEach(async () => {
-    await RadioGroupPageObject.scrollToTestElement();
+    await RadioGroupPageObject.scrollToTestElement(await RadioGroupPageObject._firstRadioGroup);
   });
 
-  it("Validate RadioGroup's accessibilityRole is correct", async () => {
-    await expect(await RadioGroupPageObject.getAccessibilityRole()).toEqual(RADIOGROUP_A11Y_ROLE);
+  it('Validate RadioGroup\'s "accessibilityRole" defaults to List "ControlType" element attribute.', async () => {
```
samuelfreiberg, line 34: Nit: 'Validate RadioGroup\'s "accessibilityRole" property defaults to "ControlType".'

apps/E2E/src/RadioGroup/specs/RadioGroup.spec.win.ts
```diff
@@ -29,80 +28,133 @@ describe('RadioGroup/RadioButton Testing Initialization', function () {
...
+    await RadioGroupPageObject.click(RadioGroupPageObject.getRadioButton(RadioButton.Second));
 
     /* Validate the RadioButton is selected */
-    await expect(await RadioGroupPageObject.isRadioButtonSelected(RadioButtonSelector.Second)).toBeTruthy();
+    await expect(
+      await RadioGroupPageObject.waitForRadioButtonSelected(
+        RadioButton.Second,
+        'Clicked the second RadioButton, but it failed to be selected.',
+      ),
+    ).toBeTruthy();
```
samuelfreiberg, line 126: Using `.toBeTruthy()` here and `.toBeTrue()` above. We should use the same one throughout the repo.

Is there a functional difference?

apps/E2E/src/RadioGroup/pages/RadioGroupPageObject.ts
lawrencewin: I can't confirm anything. The page is inaccessible on the UWP test app, and both RadioGroups are not part of the tests run on UWP. For now, this should not be an issue.

apps/E2E/src/RadioGroup/specs/RadioGroup.spec.win.ts
lawrencewin: "Truthy-ness", if I recall, encompasses values such as true booleans, numbers > 0 and non-empty strings whereas True only means a true boolean. Since this returns true on success, both will pass, though we should indeed stick to `.toBeTrue()`.

apps/E2E/src/RadioGroup/specs/RadioGroup.spec.win.ts
lawrencewin: For here, would it be better to include which ControlType the element should default to? Especially in this case where we're testing both RadioGroup and RadioButton which have different ControlTypes

apps/E2E/src/RadioGroup/specs/RadioGroup.spec.win.ts
lawrencewin: On closer inspection, `.toBeTruthy()` accepts an optional parameter that allows you to throw an error message if the truthy test fails. Maybe sticking to `.toBeTruthy` might be the better move.

samuelfreiberg (approved):

Issue #2190, opened 2022-10-03
Refactoring - Make theme tokens flow consistent


ankraj12:
### Platform

- [X] iOS
- [X] macOS
- [X] win32
- [ ] windows
- [ ] android

### Describe the feature that you would like added

As per #2159 , there was code move to start using [theme-tokens](https://github.com/microsoft/fluentui-react-native/tree/main/packages/theming/theme-tokens) for creation of all tokens that use fluent token pipeline.  default-theme changes has been merged in above PR, [win32](https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/win32-theme/src/createOfficeAliasTokens.ts), [macOS](https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/apple-theme/src/createMacOSAliasTokens.ts) and [iOS](url) can be restructured to follow the same.

### Have you discussed this feature with our team, and if so, who

@rurikoaraki

PR #2481, opened 2023-01-04
Set up iOS alias color tokens

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Set up iOS alias color tokens in FURN (just neutral and brand colors for now; shared colors will be done in a future PR), so the tokens specified in updated Fluent2 designs can be accessed through the theme.

1. The pipeline output is retrieved by getTokens.ios.ts
   - Also moved apple-theme/getiOSTokens.ts to theme-tokens/getTokens.ios.ts, as per https://github.com/microsoft/fluentui-react-native/issues/2190. Slightly unrelated but also recombined the existing android/windows getTokens/getShadowTokens files - they were separated to make android token integration more simple but since that work is complete, should be fine to recombine them and have less files 
3. The output gets mapped in an AliasColorTokens object through mapPipelineToTheme
   - The AliasColorTokens object was updated to include some new mobile tokens
   - The existing mapPipelineToTheme method in the mapPipelineToTheme.ios.ts file was copied from the original mapPipelineToTheme.ts file. Updated it to match designs - for this file reviewing commit by commit should make the changes more clear
4. The AliasColorTokens object gets added to the iOS color palette in paletteFromAppleColors  by createiOSColorAliasTokens

Out of scope for this PR:
1. Shared colors (ex. dangerbackground1) - there's been some questions raised about the definitions for these tokens so they haven't been added to the design-tokens package yet
2. Moving all createAliasTokens files from platform-specific themes into the theme-tokens folder - this seemed to have been planned for all the platforms but none of them do this yet. Planning to do this in a future PR

### Verification

Initial verification since the actual tokens aren't part of FURN yet - copied some tokens into the IOS alias-tokens files in node_modules, and checked that they were part of the final iOS palette.

No visual changes otherwise. Will add these tokens to the Notification component in a future PR for additional verification.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theme-types/src/Color.types.ts
```diff
@@ -522,6 +522,14 @@ export interface AliasColorTokens {
   /** @platform android, iOS */
   neutralForegroundOnColor?: ColorValue;
 
+  // TODO #2440: Add to android
+  /** @platform iOS */
+  neutralForegroundDarkStatic?: ColorValue;
+
+  // TODO #2440: Add to android
+  /** @platform iOS */
+  neutralForegroundLightStatic?: ColorValue;
```
harrieshin, line 531: do we actually know for sure android needs it?

packages/theming/theme-tokens/src/getTokens.ios.ts
```diff
@@ -1,9 +1,24 @@
...
 import { assertNever } from 'assert-never';
 
-export function getiOSShadowTokens(mode: AppearanceOptions) {
+export function getAliasTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return iOSLightAliasTokens;
+  } else if (mode === 'dark') {
+    return iOSDarkAliasTokens;
+  } else if (mode === 'highContrast') {
+    throw new Error('highContrast is not a valid AppearanceOptions on iOS');
```
harrieshin, line 15: can we also give comment to documentation why?

harrieshin (approved):

packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts
```diff
@@ -1,134 +1,116 @@
 import { AliasColorTokens, Variants, VariantValue } from '@fluentui-react-native/theme-types';
 import { FontStyleTokens } from '@fluentui-react-native/tokens';
 
+// API that translates tokens coming for android to Theme color values.
```
ankraj12, line 4: nit n rename to iOS

packages/theming/theme-types/src/Color.types.ts
ankraj12: There typically is a lag from iOS - Android figma on the tokens and values owing to design team functioning, but in almost all cases the tokens remain same for both mobile endpoints. So I would take these as valid changes for android.

packages/theming/theme-tokens/src/getTokens.ios.ts
Saadnajmi: Adding quick notes for some context
I did a quick google search, and it looks like RN Core is missing the ability to look that up: https://reactnative.dev/docs/accessibilityinfo
The only way _right now_ to do it is via PlatformColor: https://reactnative.dev/docs/dynamiccolorios
I imagine @mischreiber 's work of exposing trait collection may work into feeding this option into RN Core.

packages/theming/theme-types/src/Color.types.ts
lyzhan7: I would make the change for Android myself but the android token values in the design-tokens repo needs to be updated first, and there are tokens there that should be removed as well as added. I think it would make more sense to update Android altogether since there are updated controls that could be impacted by these token updates, which is why I made a github issue. @ankraj12 if there's a place in ADO that would make sense for me to create a work item, let me know.

ankraj12: Sure Lynn, please feel free to add in a tracking item on me for this and android team an re-base the tokens we are using across deigns tokens repo. and FURN.

packages/theming/theme-tokens/src/getTokens.ios.ts
```diff
@@ -1,9 +1,24 @@
...
 import { AppearanceOptions } from '@fluentui-react-native/theme-types';
 import { assertNever } from 'assert-never';
 
-export function getiOSShadowTokens(mode: AppearanceOptions) {
+export function getAliasTokens(mode: AppearanceOptions) {
+  if (mode === 'light') {
+    return iOSLightAliasTokens;
+  } else if (mode === 'dark') {
+    return iOSDarkAliasTokens;
+  } else if (mode === 'highContrast') {
```
ankraj12, line 14: Curious - how did you hot this assert at runtime flows from the tester?

packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts
```diff
@@ -1,134 +1,116 @@
...
 
-    brandBackgroundStatic: pipelineOutput.brandBackgroundStatic.fillColorRest,
-    brandBackground2: pipelineOutput.brandBackground2.fillColorRest,
+    neutralForeground1: pipelineOutput.neutralForeground1.fillColorRest,
+
+    neutralForeground2: pipelineOutput.neutralForeground2.fillColorRest,
+
+    neutralForeground3: pipelineOutput.neutralForeground3.fillColorRest,
+
+    neutralForegroundDisabled1: pipelineOutput.neutralForegroundDisabled1.fillColorRest,
```
ankraj12, line 57: Re-looking a the tokens, I think `neutralForegroundDisabled1` can be completely removed to use `neutralForegroundDisabled` for all endpoints ( or reverse). They are present in both endpoints with different names, and could be merged already.

```
/** @platform macOS, win32, windows */
  neutralForegroundDisabled?: ColorValue;

  /** @platform android, iOS */
  neutralForegroundDisabled1?: ColorValue;

```

packages/theming/theme-tokens/src/getTokens.ios.ts
lyzhan7: The default theme was always getting created on startup for iOS - actually, all three types of default theme (light mode, dark mode, and high contrast mode) were getting created (https://github.com/microsoft/fluentui-react-native/blob/main/packages/theming/default-theme/src/createDefaultTheme.ts), regardless of what mode iOS was on. After the getiOSTokens.ts to getTokens.ios.ts rename, the default theme started using iOS tokens to get created rather than web/win32 tokens. So we were trying to create a high contrast default theme using iOS tokens, which doesn't define any. 

Android should be hitting this too, but rather than throwing an error, Android just returns the light mode tokens. I might make the same change for iOS, until we can come up with an overall plan for deciding what logic each theme should contain and how the default theme should work

packages/theming/theme-tokens/src/getTokens.ios.ts
lyzhan7: Yes iOS currently doesn't have the ability to look it up, also we don't have high contrast tokens for iOS. Currently just light/dark mode, and dark elevated coming up. Ended up changing this to return the light mode tokens - needed because the default theme will always try to create a high contrast theme, and with this change it'll try to create a high contrast theme with iOS tokens. More context here: https://github.com/microsoft/fluentui-react-native/issues/2492

packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts
lyzhan7: Updated. I thought some parts of the original comment were redundant so I replaced it/the Android comment with something that made more sense to me - happy to update it again if you think I removed some important information

packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts
lyzhan7: Thanks for pointing this out - asked about how to resolve this inconsistency on Teams. I think the designs themselves and the design-tokens repo should be updated to be consistent across platforms to avoid confusion

lyzhan7:
Currently waiting on https://github.com/microsoft/fluentui-react-native/pull/2502 to go in first - that should be the last blocker for the PR checks to pass

packages/theming/theming-utils/src/mapPipelineToTheme.ios.ts
lyzhan7: Going to leave as is for now and follow up once this has been discussed on Teams

Issue #2492, opened 2023-01-06
Make default theme use platform-specific alias tokens


lyzhan7:
### Platform

- [X] iOS
- [X] macOS
- [X] win32
- [X] windows
- [X] android

### Describe the feature that you would like added

Current state of the default theme:
- The defaultTheme was set up before all of the other platforms were fully supported, and was based on the web palette as a fallback (which currently windows in FURN also uses). It works like a web theme, which means that it has light/dark/hc as its modes.
- Up until this [PR](https://github.com/rohanprasadofficial/fluentui-react-native), the default theme has been using web/windows tokens. After that change, on Android only, the default theme is actually using Android tokens. 
- iOS will likely also be following Android and will also be in a mixed state (default theme is getting created with iOS tokens)
- One of the issues with how iOS/Android currently work, is that the default theme creation code assumes there is a light, dark, and high contrast mode version. iOS and android currently do not support light or dark mode, so as a workaround both currently return use light mode tokens to populate the default theme 
- So currently, how each platform's default-theme works is different

Whether the default theme should be using web/windows tokens or should be using platform specific tokens was brought up with the recent iOS color tokens work. We are thinking of changing the existing behavior so that the default theme is more specific to each platform. An engineering plan will need to be drawn up for this work, and we should ensure that the changes do not impact package size.

Some open questions to consider
- Is it possible to not pull in the defaultTheme if a client provides a ThemeProvider

PR #2500, opened 2023-01-09
Change default theme jest config platform from iOS -> win32

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR is related to https://github.com/microsoft/fluentui-react-native/pull/2481, decided to make this a separate PR in order to keep the original as small as possible.

Why this change is required for PR 2481:
- Currently the default-theme tests run on iOS, and before PR 2481 goes in iOS will have been using web/windows alias tokens. So technically, the default-theme tests were never really testing iOS, they've been generally testing web/windows theme creation. Also, the default-theme tests include tests for high contrast mode and 'dynamic' mode - neither of these exist on iOS yet. 
- After PR 2481 iOS will be using iOS alias tokens. This means the default-theme tests will be using iOS alias tokens. It doesn't make sense to be running high contrast tests on iOS, so updating these tests to actually be running on a platform that has high contrast
- Note - these tests still do run on iOS, I had to make a change in PR 2481 so that on iOS, if we try to access high contrast mode, instead of throwing an error we return the light mode tokens. I could keep the platform here set to iOS, just the high contrast mode tests will be really testing light mode tokens

For more context behind the current state of the default theme see https://github.com/microsoft/fluentui-react-native/issues/2492

### Verification

No visual changes

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/theming/default-theme/package.json
+++ b/packages/theming/default-theme/package.json
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
     "@types/react": "^17.0.2",
     "react": "17.0.2",
     "react-native": "^0.68.0"
--- a/change/@fluentui-react-native-default-theme-46304d79-3780-47f1-8df7-438af921a622.json
+++ b/change/@fluentui-react-native-default-theme-46304d79-3780-47f1-8df7-438af921a622.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update jest.config",
+  "packageName": "@fluentui-react-native/default-theme",
+  "email": "78454019+lyzhan7@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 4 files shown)


harrieshin (approved):

packages/theming/default-theme/package.json
```diff
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@fluentui-react-native/eslint-config-rules": "^0.1.1",
     "@fluentui-react-native/scripts": "^0.1.1",
+    "@office-iss/react-native-win32": "^0.68.0",
```
Saadnajmi, line 44: I'd also add it peer dependencies while we're at it.

lyzhan7:
I realized another option is to change the platform to windows instead of win32, maybe this would make more sense since the default-theme was created with with web/windows tokens in mind? Not sure what the difference is between web/windows and win32 tokens. Thinking I'll leave it as is for now though since win32 is used more often & might be better to have more test coverage on win32

packages/theming/default-theme/package.json
lyzhan7: In all of the package.json files that include this, it's only under devDependencies. Should we also be adding to to peer dependencies for all of the package.json files?

Saadnajmi: Oh, if that's the case then nvm. IMO yes anything that takes this fork as a devDependency should also take it as a peer dependency. But it's not a big deal.

Issue #2501, opened 2023-01-09
Fix apple-theme jest test on macOS


lyzhan7:
### Platform

- [ ] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [ ] android

### Describe the feature that you would like added

The apple-theme jest tests have been running on iOS, even though they are testing macOS-specific methods. After trying to change the apple-theme jest tests to run on macOS instead, the tests start failing due to this issue:

![image](https://user-images.githubusercontent.com/78454019/211388466-6dac5496-29c9-4420-8091-95d44c7c4672.png)

Spent several hours trying to figure out why this error was coming up only in the tests (running the code in the tester works fine). My best guess was that maybe AccessibiltyInfo/IsHighContrastEnabled needs to be mocked out, but wasn't able to figure out how to stop the error.

In order to get a separate PR unblocked, temporarily commented out the test but we should figure out how to get this error to stop showing up on macOS and get the test working again.

Saadnajmi:
@chiuam any idea? I think you worked last on the high contrast getter?

rurikoaraki:
Might need a mock - a lot of RN module functionality tends to need to get mocked out. esp if it accesses native info

rurikoaraki:
https://jestjs.io/docs/tutorial-react-native#mock-native-modules-using-jestmock

lyzhan7:
Think I found a fix for this, which was to make sure the AccessibilityInfo module was not getting mocked out. My working understanding of why this fixed the issue is that the AccessibilityInfo module was getting mocked out automatically (I think because it's a node module, and [node modules get mocked out automatically](https://jestjs.io/docs/manual-mocks#mocking-node-modules)). When I would try printing out what was in AccessibilityInfo this is what I would get:

```
{
        addEventListener: [Function: mockConstructor] {
          _isMockFunction: true,
          getMockImplementation: [Function (anonymous)],
          mock: [Getter/Setter],
          mockClear: [Function (anonymous)],
          mockReset: [Function (anonymous)],
          mockRestore: [Function (anonymous)],
          mockReturnValueOnce: [Function (anonymous)],
          mockResolvedValueOnce: [Function (anonymous)],
          mockRejectedValueOnce: [Function (anonymous)],
          mockReturnValue: [Function (anonymous)],
          mockResolvedValue: [Function (anonymous)],
          mockRejectedValue: [Function (anonymous)],
          mockImplementationOnce: [Function (anonymous)],
[...]

PR #2502, opened 2023-01-09
Change apple theme jest config platform from iOS -> macOS

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

PR is related to https://github.com/microsoft/fluentui-react-native/pull/2481, making this a separate PR in order to keep the original as small as possible.

Why this change is required for PR 2481:
- Currently the apple-theme tests are running on iOS, so some of the code that's getting run is iOS-specific code, but the tests are testing macOS-specific methods and are using macOS tokens
- One of the reasons this has worked is that the mapPipelineToTheme method has been the same for both iOS/macOS
- After 2481 we've introduced an iOS specific mapPipelineToTheme.ios.ts file, which is written for the new iOS tokens. The apple-theme tests fail after this change because now we're trying to create an iOS alias token object from macOS tokens
- **Given that the tests seem to have been written with macOS in mind, update the tests to run on the macOS.** 
- ~~Ran into an issue with updating the macOS tests, filed here #2501. Justification for commenting out this test - this test hasn't been testing either macOS or iOS correctly - we've been using a mix of code from the two platforms. This test didn't run into the same issue before because it was actually running the iOS version of that code (createAppleTheme.ios.ts, instead of createAppleTheme.macos.ts)~~ found a fix for this issue
- Another option I considered was to change the tests to work for iOS- I would remove all the macOS specific tests (anything related to high contrast mode/dynamic). Decided it would be better to keep the testing coverage the same as before, and later find a way to add iOS-specific tests in addition to the existing macOS tests

Note: planning to write iOS tests in the future

### Verification

No visual changes

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/apple-theme/src/__tests__/apple-theme.test.ts
```diff
@@ -1,9 +1,14 @@
-import { createAppleTheme } from '../createAppleTheme';
 import { createMacOSColorAliasTokens, createMacOSShadowAliasTokens } from '../createMacOSAliasTokens';
 import { getIsHighContrast, setIsHighContrast } from '../appleHighContrast.macos';
 import { AppearanceOptions } from '@fluentui-react-native/theme-types';
+import { createAppleTheme } from '../createAppleTheme';
+
+// For some reason the automatically mocked AccessibilityInfo object is the react-native version, not the
+// react-native-macos version, which doesn't contain the isHighContrastEnabled function that is accessed in
+// createAppleTheme. Use the actual module instead
+jest.dontMock('react-native-macos/Libraries/Components/AccessibilityInfo/AccessibilityInfo');
```
Saadnajmi, line 9: Maybe make an issue and leave the number here for follow-up?

Saadnajmi (approved):

packages/theming/apple-theme/src/__tests__/apple-theme.test.ts
lyzhan7: Going to check this in as is to get the color token work unblocked for now

PR #2504, opened 2023-01-09
Add Retractable E2E Sections to Fluent Tester 

lawrencewin:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

In the Fluent Tester, some component pages have sections at the bottom of the page dedicated for E2E testing. For users playing around with the app, these don't serve much purpose. For E2E testing as well, one bottleneck with testing speed is the time it takes to load components into view. 

This PR adds an expanding / retracting section dedicated for E2E testing at the top of of the following pages (these are all the pages with dedicated components for A11y and E2E testing):
- Avatar
- Button(s)
- Callout
- Checkbox(es)
- ContextualMenu
- FocusZone
- IconV1
- Link(s)
- Menu
- MenuButton(s)
- RadioGroup(s)
- Switch
- Tab(s)
- Text(s)

### Verification

| Win32 (E2E off) | Win32 (E2E on) |
|------------------|------------------|
|image|image|

| Windows (E2E off) | Windows (E2E on) |
|----------------------|--------------------|
| image | image |

| Mac (E2E off) | Mac (E2E on) |
|----------------|----------------|
| MicrosoftTeams-image (3) | MicrosoftTeams-image (4) |

| Android (E2E off) | Android (E2E on) |
|--------------------|--------------------|
| Screenshot_2023-01-24-12-49-56-52_3238b23fbbdef38ccf4a465a34706bab | Screenshot_2023-01-24-12-50-00-11_3238b23fbbdef38ccf4a465a34706bab |

| iOS (E2E off)        | iOS (E2E on)       |
|--------------------|-------------------|
| MicrosoftTeams-image (1)     | MicrosoftTeams-image (2)    |

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


samuelfreiberg:
I think I'd rather see the E2E test switch on each specific test page, rather than next to the Theme dropdowns.

Reason being, those three drop-downs aren't specific to a test page, it's for the whole app. But the E2E testing switch is specific for each test app. I think we should move the switch to be next to the "spec" link on each test page

lawrencewin:
> I think I'd rather see the E2E test switch on each specific test page, rather than next to the Theme dropdowns.
> 
> Reason being, those three drop-downs aren't specific to a test page, it's for the whole app. But the E2E testing switch is specific for each test app. I think we should move the switch to be next to the "spec" link on each test page

Makes sense. Updated the Test component to render the switch if an E2E section is passed. Currently trying to see how it looks per platform. Looks good on win32 and seems to be fine on windows. Lets see how android, iOS, and mac look.

apps/fluent-tester/src/TestComponents/Test.tsx
```diff
@@ -113,23 +133,48 @@ export const Test = (props: TestProps): React.ReactElement<Record<string, never>
...
+            {...testProps('Focus_Button')}
+            style={styles.e2eFocusButton}
+          >
+            E2E Button
+          </Button>
+        )}
+        {props.e2eSections && (
+          <View style={styles.e2eSwitch}>
+            <Text style={styles.e2eSwitchLabel} {...(isMobile ? {} : { variant: 'body1Strong' })}>
+              Show E2E
```
samuelfreiberg, line 158: Nit: Let's change to "E2E Mode"

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
```diff
@@ -16,6 +16,9 @@ describe('Avatar Testing Initialization', function () {
     await AvatarPageObject.waitForPageDisplayed(PAGE_TIMEOUT);
 
     await expect(await AvatarPageObject.isPageLoaded()).toBeTruthy(AvatarPageObject.ERRORMESSAGE_PAGELOAD);
+
+    await AvatarPageObject.enableE2ETesterMode();
```
samuelfreiberg, line 20: I think we can make some API improvements here. 

We know that every spec is going to need these 3 functions - `NavigateAppPage.clickAndGoToAvatarPage();`, `AvatarPageObject.waitForPageDisplayed(PAGE_TIMEOUT);`, and `await AvatarPageObject.enableE2ETesterMode();`.

Can we consolidate these into one method?

We can just have `NavigateAppPage.navigateToAvatarE2EPage();`. This method is responsible for clicking on the Avatar page, waiting for page load, and enabling E2E tester mode.

I think this makes writing specs a lot cleaner to write and easier to read.

samuelfreiberg (changes requested):

apps/E2E/src/Avatar/specs/Avatar.spec.win.ts
lawrencewin: Added a method to BasePage: `navigateToPageAndLoadTests()` which handles navigation, waits for the page to load, and if passed a flag, enables e2e mode. 

So far, I only refactored the win32 tests that have e2e sections. Should I do the same for all specs or put that in a separate PR?

lawrencewin:
Finally got pictures of each tester and the new E2E switch. Thus, I can mark this ready for review.

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +80,44 @@ export abstract class BasePage {
...
+   * For any given page object, this method automates navigating to the page, waiting for it to load, and revealing its E2E components.
+   * This also contains error checking through its waiters, removing the extra 'expect()' calls during test setup.
+   */
+  async navigateToPageAndLoadTests(showE2ESection = false) {
+    // Desktop platforms automatically scroll to a page's navigation button - this extra step is purely for mobile platforms.
+    if (this.platform === MobilePlatform.Android || this.platform === MobilePlatform.iOS) {
+      await this.mobileScrollToComponentButton();
+    }
+
+    await (await this._pageButton).click();
```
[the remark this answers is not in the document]
lawrencewin: We can use this method in the rest of the tests and probably remove `NavigateAppPage`. 

If we remove `NavigateAppPage`, we still need to check whether the first tester page loads, and that can be done by `BasePage`.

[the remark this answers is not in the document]
samuelfreiberg: Let's do that

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +82,48 @@ export abstract class BasePage {
   /**************** UI Element Interaction Methods ******************/
   /******************************************************************/
 
+  /**
+   * For any given page object, this method automates:
+   *  - Navigating to the page by clicking its component button
+   *  - Waiting for the component page to load
+   *  - Revealing its E2E components IF a boolean flag is passed into the method
```
samuelfreiberg, line 89: Why do we need a boolean flag to show it's E2E components?

samuelfreiberg, line 89: Nit:

"Expanding the E2E testing sections" rather than "Revealing its e2e components"

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +82,48 @@ export abstract class BasePage {
...
+    await this.waitForPageDisplayed();
+
+    if (showE2ESection) {
+      await this.enableE2ETesterMode();
+    }
+  }
+
+  async enableE2ETesterMode(): Promise<void> {
+    const e2eSwitch = await this._e2eSwitch;
+    switch (this.platform) {
```
samuelfreiberg, line 109: Can we add a comment explaining the difference between Android vs. other platforms here?

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +82,48 @@ export abstract class BasePage {
...
+            async () => (await e2eSwitch.getAttribute(AndroidAttribute.Checked)) === 'true',
+            'Clicked the E2E Mode Switch, but it failed to toggle.',
+          );
+        }
+        break;
+      default:
+        if (!(await e2eSwitch.isSelected())) {
+          await e2eSwitch.click();
+          await this.waitForCondition(async () => e2eSwitch.isSelected(), 'Clicked the E2E Mode Switch, but it failed to toggle.');
+        }
```
samuelfreiberg, line 123: I think we should also add a wait outside of the case statement that validates the E2E section is visibly shown correctly.

This might require adding another E2E testing constant on the title of the E2E section

apps/E2E/src/common/BasePage.ts
lawrencewin: Some pages don't have a dedicated section for E2E tests and don't render the switch. If we try and click the unrendered switch, appium won't find the element and throw an error.

samuelfreiberg: jw what pages don't have a dedicated E2E section? Do those components have E2E tests running?

lawrencewin: The pages that only check for baseline E2E crash tests, such as ActivityMonitor, Badge, IconLegacy, don't have dedicated E2E sections. Those tests only check if the page title becomes visible.

samuelfreiberg: Let's add that explanation in the comments for the prop description

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +82,48 @@ export abstract class BasePage {
   /**************** UI Element Interaction Methods ******************/
   /******************************************************************/
 
+  /**
+   * For any given page object, this method automates:
+   *  - Navigating to the page by clicking its component button
+   *  - Waiting for the component page to load
+   *  - Revealing its E2E components IF a boolean flag is passed into the method
+   *
+   * This also contains error checking through its waiters, removing the extra `expect()` calls during test setup.
```
samuelfreiberg, line 91: Add:

Parameter: showE2ESection: boolean (default false) - Explanation of why we need this param

samuelfreiberg:
@lawrencewin I believe the E2E ReadMe has references to NavigateAppPage. We need to update the ReadMe

apps/E2E/src/common/BasePage.ts
```diff
@@ -72,6 +83,59 @@ export abstract class BasePage {
...
 
+  /**
+   * For any given page object, this method automates:
+   *  - Navigating to the page by clicking its component button
+   *  - Waiting for the component page to load
+   *  - Optionally expanding its E2E sections IF a boolean flag is passed.
+   *
+   * This also contains error checking through its waiters, removing the extra `expect()` calls during test setup.
+   *
+   * @param {boolean} showE2ESection (default = `false`) Some pages have E2E tests that only check if the component causes
```
samuelfreiberg, line 94: Nit:

Some components' E2E tests check only if the test page loads correctly or not. Others (majority), perform UI manipulation tests on UI components on the test page. In these scenarios, these UI components have their own section on the test page (by default, it's hidden so partners don't see it). If this parameter is true, this method opens up that testing section.

apps/E2E/README.md
```diff
@@ -210,9 +241,7 @@ Spec documents should be put in apps/fluent-tester/src/E2E/\_ _ComponentToBeTest
 ```
 describe('Click on each test page and check if it renders', function() {
   it('Checkbox Test Page', () => {
-    NavigateAppPage.clickAndGoToCheckboxPage();
-    expect(CheckboxPageObject.isPageLoaded()).toBeTruthy();
-  });
+    CheckboxPageObject.navigateToPageAndLoadTests(true); // check true if you have a dedicated section for E2E tests
```
samuelfreiberg, line 244: Nit: "pass in" instead of "check"

apps/E2E/src/ActivityIndicator/specs/ActivityIndicator.spec.ios.ts
```diff
@@ -1,22 +1,12 @@
...
   it('Click and navigate to Activity Indicator test page', async () => {
-    await ActivityIndicatorPageObject.mobileScrollToComponentButton();
-    await ActivityIndicatorPageObject.waitForButtonDisplayed(PAGE_TIMEOUT);
-
-    /* Click on component button to navigate to test page */
-    await NavigateAppPage.clickAndGoToActivityIndicatorPage();
-    await ActivityIndicatorPageObject.waitForPageDisplayed(PAGE_TIMEOUT);
-
-    await expect(await ActivityIndicatorPageObject.isPageLoaded()).toBeTruthy(ActivityIndicatorPageObject.ERRORMESSAGE_PAGELOAD);
+    await ActivityIndicatorPageObject.navigateToPageAndLoadTests();
```
samuelfreiberg, line 10: Does this cause an issue since there's no `expect` call here?

lawrencewin: From testing, a `waitFor` timing out causes a test to fail because it will throw an error. The spec still acts the same, but putting an `expect` call here would be good for sanity.

apps/E2E/src/common/BasePage.ts
```diff
@@ -65,6 +75,65 @@ export abstract class BasePage {
...
+    // Desktop platforms automatically scroll to a page's navigation button - this extra step is purely for mobile platforms.
+    if (this.platform === 'android' || this.platform === 'ios') {
+      await this.mobileScrollToComponentButton();
+    }
+
+    await (await this._pageButton).click();
+    // Wait for page to load
+    await this.waitForCondition(
+      async () => await this.isPageLoaded(),
+      this._pageName + ' did not render correctly. Please see /errorShots for more information.',
```
samuelfreiberg, line 101: Can we just use `ERRORMESSAGE_PAGELOAD` here?

apps/E2E/src/common/BasePage.ts
```diff
@@ -177,25 +241,12 @@ export abstract class BasePage {
...
-  async waitForButtonDisplayed(timeout?: number): Promise<void> {
-    await browser.waitUntil(async () => await this.isButtonInView(), {
-      timeout: timeout ?? this.waitForUiEvent,
-      timeoutMsg: 'Could not find the button to navigate to ' + this._pageName + '. Please see /errorShots for more information.',
-      interval: 1500,
-    });
-  }
-
-  /* @deprecated, only use `scrollToTestElement()` instead */
+  /** @deprecated, only use `scrollToTestElement()` instead */
```
samuelfreiberg, line 244: Do we really need this here? Are there any uses of waitForPrimaryElementDisplayed() currently?

Since this isn't an external package, I don't think we really need to use deprecated here. We should just replace all instances of this with the new method and delete it

samuelfreiberg:
Just have one more unresolved comment + CI is failing

apps/E2E/src/ActivityIndicator/specs/ActivityIndicator.spec.ios.ts
lawrencewin: The expects have been re-added

samuelfreiberg (approved):

PR #2506, opened 2023-01-10
[Proposal] Jest multi plat testing

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change demos what configuring and implementing multi-plat jest tests could look like.

Main things:
- I'm using Button as the component of choice
- I moved all the tests under a `__tests__` file so that they're in one place
- I moved all the tests out of `src` because this is a lot of files and they don't need to be in the published package
- I added a jest.config file for each platform
- I added test files for each platform. I think this is unfortunately necessary to get the snapshot comparison to work per-platform, because the path to the snapshot file for a test has to be  able to be derived from the path to the test file, and vice versa. See [this](https://jestjs.io/docs/27.x/configuration#snapshotresolver-string) for more info on the API.
- Snapshots were added for each platform
- yarn test now calls yarn test-iOS and yarn-win32, which then find the appropriate config file
- jest task has been modified to account for -c arg

### Verification

Verified yarn test runs both sets of tests locally. win32 doesn't pass at this moment.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/package.json
```diff
@@ -15,7 +15,9 @@
     "just": "fluentui-scripts",
     "clean": "fluentui-scripts clean",
     "lint": "fluentui-scripts eslint",
-    "test": "fluentui-scripts jest",
+    "test": "yarn test-iOS && yarn test-win32",
+    "test-iOS": "fluentui-scripts jest -c __tests__/iOS/jest.config.js",
+    "test-win32": "fluentui-scripts jest -c __tests__/win32/jest.config.js",
```
Saadnajmi, line 20: I think the convention is that these are always lowercase. Also I think `test:ios` / `test:win32` is closer to what I've seen elsewhere in similar Microsoft repos

Saadnajmi, line 20: If we follow the same structure everywhere, could we modify the script invoked with `fluentui_scripts` to take a `-- platform` arg that then chooses the right config? Guessing on how to make this a little more ergonomic. If we can make it an argument like `--platform` then we don't need extra `test-<platform>` scripts

packages/components/Button/__tests__/win32/ButtonLegacy.test.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
```
lyzhan7, line 1: It looks like the tests are duplicated in iOS and win32 - I wonder if it's possible to move the contents of the file (not the file itself) into another file ex. we'd have a ButtonLegacy.test.tsx file outside of iOS and win32 that contains any tests that are shared amongst all the platforms, and it exports those tests, and then there's a ButtonLegacy.test.tsx file in each of iOS/win32 that's referencing those tests. 

Not sure if this would be more complicated to navigate vs. potentially having to add a test in 5 different places

packages/components/Button/package.json
rurikoaraki: I think it could also be nice to have a per-platform command so that you don't have to run /all the tests/ in order to iterate on a failure if you need to look into something, so I would say there's value in having per-platform commands

Saadnajmi: Let me rephrase:

`yarn test` would still do some default platform
`yarn test --platform ios` would test iOS specifically, and pass the iOS jest config.

rurikoaraki: Ah okay. Would you be open to that change being a follow up? I think I'll need to add a new script in order to implement it, since the args currently have to be an option that the jest CLI accepts.

packages/components/Button/__tests__/win32/ButtonLegacy.test.tsx
rurikoaraki: Hmm yeah good idea. I'll try it out.

rurikoaraki: I think the lack of code duplication is a win over having a hard time navigating to the test. Better that all our tests are consistent, imo.

packages/components/Button/package.json
Saadnajmi: Yeah that's fine!

lyzhan7:
"I moved all the tests out of src because [...] they don't need to be in the published package" - I didn't know that everything outside of src gets removed from the published package. Is that behavior something we specified for FURN or is that from something else (yarn maybe?)

lyzhan7:
I was wondering if you've looked into the [rnx-kit/jest-preset package](https://microsoft.github.io/rnx-kit/docs/tools/jest-preset) to see if anything there would be useful? It looks like they add the ability to specify multiple jest.config.js files, like jest.config.ios.js - not exactly sure what this is meant for though, is it so that you can have tests on different platforms or is it just so that locally when you run a test it automatically chooses the platform you're on

rurikoaraki:
> I was wondering if you've looked into the [rnx-kit/jest-preset package](https://microsoft.github.io/rnx-kit/docs/tools/jest-preset) to see if anything there would be useful? It looks like they add the ability to specify multiple jest.config.js files, like jest.config.ios.js - not exactly sure what this is meant for though, is it so that you can have tests on different platforms or is it just so that locally when you run a test it automatically chooses the platform you're on

https://microsoft.github.io/rnx-kit/docs/tools/jest-preset#multiple-jestconfigjs

This looks basically like what this PR is setting up

rurikoaraki:
> "I moved all the tests out of src because [...] they don't need to be in the published package" - I didn't know that everything outside of src gets removed from the published package. Is that behavior something we specified for FURN or is that from something else (yarn maybe?)

Looking at it again I was wrong, the files will get published, but they won't be included in the lib/lib-commonjs files (i.e. transpiled from TS to JS). This is specified https://github.com/microsoft/fluentui-react-native/blob/main/scripts/src/tasks/ts.js I think

lyzhan7:
> > I was wondering if you've looked into the [rnx-kit/jest-preset package](https://microsoft.github.io/rnx-kit/docs/tools/jest-preset) to see if anything there would be useful? It looks like they add the ability to specify multiple jest.config.js files, like jest.config.ios.js - not exactly sure what this is meant for though, is it so that you can have tests on different platforms or is it just so that locally when you run a test it automatically chooses the platform you're on
> 
> https://microsoft.github.io/rnx-kit/docs/tools/jest-preset#multiple-jestconfigjs
> 
> This looks basically like what this PR is setting up. Since we have snapshot files I don't think it's enough to only fork the jest files per endpoint. ETA: Were you suggesting something else?

Ah no I guess I was thinking maybe the way the jest files got forked per endpoint could help with getting snapshot comparison to work per platform but maybe not



> > "I moved all the tests out of src because [...] they don't need to be in the published package" - I didn't know that everything outside of src gets removed from the published package. Is that behavior something we specified for FURN or is that from something else (yarn maybe?)
> 
> Looking at it again I was wrong, the files will get published, but they won't be included in the lib/lib-commonjs files (i.e. transpiled from TS to JS). This is specified https://github.com/microsoft/fluentui-react-native/blob/main/scripts/src/tasks/ts.js I think

Hmm okay, maybe the tests shouldn't be part of the package if we're trying to reduce binary size?

rurikoaraki:
@lyzhan7 @saadnajmi I was having trouble getting all of the following to work:
1. Only one copy of the test pages
2. Snapshots being found correctly per platform
3. Snapshots from other platforms not being reported as obsolete. Jest has functionality that basically keeps track of whether snapshot files have been checked while running tests, and if it finds any snapshot files that aren't used, it'll mark them as obsolete and exit with an error code.
4. Snapshots being separated per test page

The current structure that I have the tests results in the following:
1. Only one copy of the tests, but they are all wrapped in a function so that they can be exported and run in a different file
2. Each platform has its own main test page and jest config. This allows the snapshots for each platform to not affect other platforms.
3. Side effect of the current setup is that all the snapshots are in one file. If we want them to be in separate files, then we'll need to have a File.test.tsx file for each test file under each platform folder. The content could still be shared crossplat but it does add a lot of clutter.

packages/components/Button/package.json
Saadnajmi: Leaving things to be more ergonomic as a follow up makes sense! Thanks =)

packages/components/Button/.npmignore
```diff
@@ -0,0 +1,2 @@
+# Don't include test files in package
+__tests__/
```
Saadnajmi, line 2: If we need this in every package, can we also update the component-generator later, if not in this PR? I know that isn't heavily used but I think some people may still try to use that as their "first new package" template

Saadnajmi (approved):

packages/components/Button/.npmignore
rurikoaraki: yeah, that makes sense. I can also add it top level instead.

JasonVMo:
We have this functionality now, it is done automatically with module suffixes, tests can be shared by creating shared test files.

PR #2509, opened 2023-01-10
Fixed Horizontal-Stacked RadioGroup Padding

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Left padding was uneven for horizontal-stacked RadioGroup. The fix was updating the alignment for the labelContent slot and making the marginLeft and labelMarginLeft tokens symmetrical to the right counterparts. 

### Verification

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
How does this fix look with a label that's really thin like the letter "I"? Basically something narrower than the radio above? My worry is that because the fix isn't symmetrical, it'll break in cases like that.

It's a bit weird that alignItems isn't working to place the content correctly. I wonder if there's some extra padding in the Text element? You can wrap the Text in a View and color the View's background to find out what its actual bounds are.

gulnazsayed:
> How does this fix look with a label that's really thin like the letter "I"? Basically something narrower than the radio above? My worry is that because the fix isn't symmetrical, it'll break in cases like that.
> 
> It's a bit weird that alignItems isn't working to place the content correctly. I wonder if there's some extra padding in the Text element? You can wrap the Text in a View and color the View's background to find out what its actual bounds are.

That's a good point. I tried the case where the text is narrower than the button and it breaks. Will re-work the padding for the horizontal-stacked case.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
```diff
@@ -18,6 +18,11 @@ export interface RadioTokens extends FontTokens, IColorTokens, IForegroundColorT
    */
   rootVerticalPadding?: number;
 
+  /**
+   * Left padding of the root view
+   */
+  rootLeftPadding?: number;
+
```
Saadnajmi, line 25: Should this be `rootStartPadding` for RTL support?

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -17,11 +17,20 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
           flexDirection: tokens.flexDirection,
           paddingHorizontal: tokens.rootHorizontalPadding,
           paddingVertical: tokens.rootVerticalPadding,
+          paddingLeft: tokens.rootLeftPadding,
```
Saadnajmi, line 20: Feels like if we're exposing left, we should also expose right in case some other context / client / platform has a different kind of uneven padding. Or expose `start`/`end`.

packages/experimental/RadioGroup/src/Radio/Radio.types.ts
gulnazsayed: Removed this token

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Removed the rootLeftPadding token so left padding is no longer exposed

rurikoaraki (approved):

PR #2511, opened 2023-01-11
[E2E Tests] Expand Android tests

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

- Android scroll requires the id for the scrolling to happen. Added _primaryComponentName() and _secondComponentName() in place of _primaryComponent() and _secondaryComponent().
- Updated _primaryComponent() and _secondaryComponent() to internally use _primaryComponentName() and _secondComponentName().
- Added complete E2E tests for ButtonV1, CheckboxV1, RadioGroupV1.
- Added AndroidAttributes class to verify their values / mobileScrollToTestElement() to scroll to E2E section / Crash detection.
- CheckboxV1 had extra selector (CheckboxV1Selector) that was not adding any functionality, and extra scrollToTestElement() calls. Removed them.

### Verification

Pipeline E2E tests run as expected.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/ButtonV1/specs/ButtonV1.spec.android.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('ButtonV1 - Verify accessibilityLabel', async () => {
+    await expect(
+      await ButtonV1PageObject.compareAttribute(
+        ButtonV1PageObject._primaryComponent,
+        AndroidAttribute.AccessibilityLabel,
+        BUTTON_TEST_COMPONENT,
+      ),
+    ).toBeTrue();
```
samuelfreiberg, line 37: Let's align with the rest of the file and use `.toBeTruthy()` here. Consistency across APIs. `.toBeTruthy()` also allows us to pass in an error message param

apps/E2E/src/CheckboxV1/specs/CheckboxV1.spec.android.ts
```diff
@@ -1,21 +1,76 @@
...
+  });
+
+  it('CheckboxV1 - Verify accessibilityLabel', async () => {
+    await expect(
+      await CheckboxV1PageObject.compareAttribute(
+        CheckboxV1PageObject._primaryComponent,
+        AndroidAttribute.AccessibilityLabel,
+        CHECKBOXV1_TEST_COMPONENT,
+      ),
+    ).toBeTrue();
```
samuelfreiberg, line 38: Same as above

apps/E2E/src/common/BasePage.ts
```diff
@@ -238,6 +242,49 @@ export abstract class BasePage {
...
+  async mobileScrollToTestElement(componentIdentifier?: string): Promise<void> {
+    const componentToScrollTo = componentIdentifier ?? this._primaryComponentName;
+    if (await (await By(componentToScrollTo)).isDisplayed()) {
+      return;
+    }
+
+    const errorMsg = 'Could not scroll to the ' + componentToScrollTo + '. Please see Pipeline artifacts for more debugging information.';
+
+    switch (this.platform) {
+      case MobilePlatform.iOS: {
```
samuelfreiberg, line 255: Is this actually needed for iOS? iOS hasn't been using this before, and it's been passing successfully.

apps/E2E/src/common/BasePage.ts
```diff
@@ -238,6 +242,49 @@ export abstract class BasePage {
     await FocusButton.addValue(scrollDownKeys);
   }
 
+  /* Scrolls to the specified or primary UI test element until it is displayed. */
+  async mobileScrollToTestElement(componentIdentifier?: string): Promise<void> {
```
samuelfreiberg, line 246: This has some code duplication from `scrollToTestElement()`.

Let's combine the two methods - this also aligns the specs better by calling the same function:

`scrollToTestElement(component?: WebDriverIO.Element)`

And then, we can simply use the `switch(this.platform)` to do the correct thing cross-plat

apps/E2E/src/ButtonV1/specs/ButtonV1.spec.android.ts
ayush547: Done. Also replaced occurrences of toBeTrue() in other files

apps/E2E/src/common/consts.ts
```diff
@@ -124,3 +139,5 @@ export const BASE_TESTPAGE = 'Base_TestPage';
 // ScrollViewer that contains the list of test page buttons
 export const TESTPAGE_BUTTONS_SCROLLVIEWER = 'TestPage_Buttons_ScrollViewer';
 
+// ScrollViewer that contains the list of test page buttons
```
samuelfreiberg, line 142: Update comment

apps/E2E/src/common/consts.ts
```diff
@@ -124,3 +139,5 @@ export const BASE_TESTPAGE = 'Base_TestPage';
...
+export const TESTPAGE_TESTS_SCROLLVIEWER = 'TestPage_Tests_ScrollViewer';
```
samuelfreiberg, line 143: I'd also change this to:

  `TESTPAGE_CONTENT_SCROLLVIEWER`

Better represents that this is the scrollviewer on each test page that wraps the content

apps/E2E/src/common/BasePage.ts
```diff
@@ -209,33 +213,82 @@ export abstract class BasePage {
...
+   * Either the component identifier or the component itself can be passed in as a parameter.
+   * On Android only the component identifier can be passed in.
+   * If no parameter is passed in, the primary component is used.
+   */
+  async scrollToTestElement(component?: WebdriverIO.Element | string): Promise<void> {
+    let componentIdentifier: string;
+    let componentToScrollTo: WebdriverIO.Element;
+
+    if (typeof component == 'string') {
+      componentIdentifier = component ?? this._primaryComponentName;
```
samuelfreiberg, line 226: I don't think we need the `?? this._primaryComponentName`? If the parameter is a string, then we know the user passed in a string value (non-null value). We should have a separate check to see if the parameter is null or not, and if it is, then set equal to `this._primaryComponentName`

samuelfreiberg: `
if(typeof component === 'undefined')
{
  componentToScrollTo = await this._primaryComponent;
}
else if(if (typeof component == 'string') {}
else { componentToScrollTo = component};
}

ayush547: Updated

apps/E2E/src/common/BasePage.ts
```diff
@@ -209,33 +213,86 @@ export abstract class BasePage {
...
+    let componentToScrollTo: WebdriverIO.Element;
+
+    if (typeof component == 'undefined') {
+      componentIdentifier = this._primaryComponentName;
+      componentToScrollTo = await this._primaryComponent;
+    } else if (typeof component == 'string') {
+      componentIdentifier = component;
+      componentToScrollTo = await By(componentIdentifier);
+    } else {
+      componentIdentifier = null;
```
samuelfreiberg, line 232: Won't this cause an issue with line 262?

apps/E2E/src/common/BasePage.ts
```diff
@@ -209,33 +213,86 @@ export abstract class BasePage {
...
-  /* Scrolls to the specified or primary UI test element until it is displayed. */
-  async scrollToTestElement(component?: WebdriverIO.Element): Promise<void> {
-    const ComponentToScrollTo = component ?? (await this._primaryComponent);
-    if (await ComponentToScrollTo.isDisplayed()) {
+  /* Scrolls to the specified or primary UI test element until it is displayed.
+   * Either the component identifier or the component itself can be passed in as a parameter.
+   * On Android only the component identifier can be passed in.
+   * If no parameter is passed in, the primary component is used.
+   */
+  async scrollToTestElement(component?: WebdriverIO.Element | string): Promise<void> {
```
samuelfreiberg, line 221: I don't think we should allow the `| string` here. It adds complexity and confusion as to what a user _should_ use.

Is this specific to Android? If you need the testID, I believe you can reverse-engineer from the WebDriverIO.Element to get the automationID

ayush547: The string is required for Android.
It won't be possible to get the ID from the component because the component itself is not retrievable when using By() when it is not on the screen!

ayush547: I could fork the function just for Android.
Alternatively, I can change the function to only take a string, the webdriverio.element will be accessible from it, and we do have string IDs available everywhere. 
Which approach do you suggest?

apps/E2E/src/common/BasePage.ts
ayush547: On Android only string would be supported.
If the webdriverio.element is retrievable to be passed into the scroll function, then the element is already visible, no need to scroll, hence line 236 would cause the function to return prematurely, never reaching line 262

apps/E2E/src/common/BasePage.ts
samuelfreiberg: Let's create an Overload of the function. One takes WebDriverIO.Element and one takes a string

samuelfreiberg: And let's make sure the comments explain why we have two different functions, and which platforms should use which

samuelfreiberg (approved):

apps/E2E/src/common/BasePage.ts
ayush547: Moving ahead as discussed offline with separate mobile scroll function since the parameter types are different and have limited code duplication. Also simplifies the functions and will be easier to diagnose misused function platform wise while writing tests.

apps/E2E/src/common/BasePage.ts
ayush547: Moving ahead as discussed offline with separate mobile scroll function since the parameter types are different and have limited code duplication. Also simplifies the functions and will be easier to diagnose misused function platform wise while writing tests.

PR #2521, opened 2023-01-13
macOS: Update Menu font

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Updating the font token value for MenuItem to align with ContextualMenu visuals.

### Verification

| image |image |
ContextualMenu:
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
+++ b/packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
@@ -9,7 +9,7 @@ export const defaultMenuItemTokens: TokenSettings<MenuItemTokens, Theme> = (t: T
   checkmarkSize: 16,
   color: t.colors.menuItemText, // matches ContextualMenu
   fontFamily: t.typography.families.primary,
-  fontSize: 13, // aligning with ContextualMenu font size
+  fontSize: 13, // aligning with NSMenu font size
   fontWeight: globalTokens.font.weight.regular as FontWeightValue,
   gap: globalTokens.size40,
   paddingHorizontal: 5, // hardcoded for now to match ContextualMenu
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
```diff
@@ -9,7 +9,7 @@ export const defaultMenuItemTokens: TokenSettings<MenuItemTokens, Theme> = (t: T
   checkmarkSize: 16,
   color: t.colors.menuItemText, // matches ContextualMenu
   fontFamily: t.typography.families.primary,
-  fontSize: globalTokens.font.size300,
+  fontSize: 13, // aligning with ContextualMenu font size
```
mischreiber, line 12: Is it worth noting that, not only are we aligning to ContextualMenu, but also to the native macOS menu?

Saadnajmi, line 12: Update comment with :and/or align with NSMenu?

Saadnajmi (approved):

packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
Saadnajmi: We have the same comment :D

Saadnajmi: Confirmed with HIG that 13 is correct. Probably why I used it for ContextualMenu. If you wanted to be very exact, you could also check with this: https://developer.apple.com/documentation/appkit/nsfont/1533068-menufont

chiuam: @Saadnajmi Thank you for checking! I'm thinking replacing ContextualMenu with just NSMenu assuming that CM will soon be deprecated?

Saadnajmi: Yeah that sounds fine. Remove references to CM and replace with NSMenu

PR #2526, opened 2023-01-17
Android Avatar - Presence Activity Ring

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- Updated tokens file in first commit - removing extra tokens and fixing values.
- Updated the type definitions to allow size 16, a RingConfig for typecheck and documentation edits.
- Added a additional view to be rendered only in android, that shows the outer most gap for active state.
- Style updates to support new slot and values.
- Tester updates.

Pending on subsequent commits:
- "ring" border needs to be a shared colored reference in android - working on the change, need some inputs from desing.
- size16 renders weirdly.
- "inactive" state should show only "innerGap" - still checking on how best to integrate this in current code.
- Cleanups, previous code review comments on memoizing styles in tester.

### Verification

| Screenshot or description before this [change](url) | Screenshot or description with this change |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


.vscode/settings.json
```diff
@@ -51,10 +51,10 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
ankraj12, line 54: Would revert this.

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
```diff
@@ -3,109 +3,145 @@ import { Avatar } from '@fluentui-react-native/avatar';
...
 
-        {/* Variation - No Ring and badge. */}
-        <Avatar size={56} avatarColor={'brand'} />
+  return (
+    <View style={{ paddingBottom: 10, paddingTop: 5 }}>
+      <Switch label={'Show Activity Ring'} defaultChecked={true} onChange={toggleActivityState} />
 
-        {/* Variation - With Ring.*/}
-        <Avatar active="active" activeAppearance="ring" size={56} avatarColor={'brand'} />
+      <View style={mobileStyles.testVariants}>
```
ankraj12, line 20: Would review this if there is scope for caching across this code as next commits.

packages/components/Avatar/src/Avatar.styling.ts
```diff
@@ -151,14 +153,32 @@ export const stylingSettings: UseStylingOptions<AvatarProps, AvatarSlotProps, Av
...
             borderColor: ringColor,
+            ...getRingSpacing(tokens),
+            backgroundColor: ringBackgroundColor || 'transparent',
             aspectRatio: 1,
           },
         };
       },
       ['size', 'ringColor', 'ringBackgroundColor', 'ringThickness', ...borderStyles.keys],
     ),
+    outerRing: buildProps(
```
rohanprasadofficial, line 165: is outer ring also needed on other platforms ? just a question 🤔  

if not we can do something like this :  ...(Platform.OS === 'android' && outerRing : () )

ankraj12: This is not needed in win32, updated the check.

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
ankraj12: Resolving this, I could not understand a case of useMemo here. If there is a case, please add in reviews and I would update it.

packages/components/Avatar/src/Avatar.types.ts
```diff
@@ -21,7 +22,7 @@ export const AvatarSizesForTokens = [
   'size120',
 ] as const;
 
-export const AvatarSizes = [20, 24, 28, 32, 36, 40, 48, 56, 64, 72, 96, 120] as const;
+export const AvatarSizes = [16, 20, 24, 28, 32, 36, 40, 48, 56, 64, 72, 96, 120] as const;
```
ValentinaKozlova, line 25: Does win32 also has 16 size?

packages/components/Avatar/src/Avatar.types.ts
```diff
@@ -219,17 +230,34 @@ export interface AvatarTokens extends IBackgroundColorTokens, IForegroundColorTo
   /**
    * Avatar sizes:
    */
+  size16?: AvatarTokens;
   size20?: AvatarTokens;
   size24?: AvatarTokens;
+
+  /** Not applicable for @platform android */
```
ValentinaKozlova, line 237: Maybe we should have tokens based on the platform? E.g. AvatarWin32Tokens and AvatarAndriodTokens. And AvatarTokens will have tokens based on the platform

packages/components/Avatar/src/Avatar.tsx
```diff
@@ -53,10 +54,17 @@ export const Avatar = compose<AvatarType>({
       const { badge, ...mergedProps } = avatar.props;
       const svgIconsEnabled = ['ios', 'macos', 'win32', 'android'].includes(Platform.OS as string);
       const RingComponent = showRing && !transparentRing ? Slots.ring : Fragment;
+      const hasOuterRing = Platform.OS === 'android';
```
ValentinaKozlova, line 57: Does android always have an outer ring?

packages/components/Avatar/src/Avatar.styling.ts
```diff
@@ -173,43 +194,91 @@ export const stylingSettings: UseStylingOptions<AvatarProps, AvatarSlotProps, Av
   },
 };
 
-function getRingConfig(tokens: AvatarTokens): any {
+function getRingConfig(tokens: AvatarTokens): RingConfig {
```
ValentinaKozlova, line 197: Maybe it'll be good to split this method into two: getRingConfigWin32 and getRingConfigAngroid for better readability?

rurikoaraki: You could create a file that holds stylingUtils.android/win32.ts or something, to minimize code that isn't used in other platforms.

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
```diff
@@ -3,109 +3,145 @@ import { Avatar } from '@fluentui-react-native/avatar';
...
-        {/* With badge. */}
-        <Avatar badge={{ status: 'available' }} avatarColor={'brand'} size={40} />
+  const toggleActivityState = (_e: InteractionEvent, isActive: boolean) => {
+    setActivityRing(isActive);
+  };
 
-        {/* Variation - No Ring and badge. */}
-        <Avatar size={56} avatarColor={'brand'} />
+  return (
+    <View style={{ paddingBottom: 10, paddingTop: 5 }}>
```
rurikoaraki, line 17: nit: Please avoid object literals as props

packages/components/Avatar/src/Avatar.tsx
ankraj12: Yes, in case of presence active - there always is an outer ring for android.

packages/components/Avatar/src/Avatar.types.ts
ankraj12: This pattern to have a superset type files is consistent across other code in repo, using documentation for indicating subset support matrix. I would prefer keeping it this way for similarity.

packages/components/Avatar/src/Avatar.styling.ts
ankraj12: Agrees, the split by platform files to clean this code piece for readability.

apps/fluent-tester/src/TestComponents/Avatar/BasicAvatar.android.tsx
ankraj12: Updated to refer to a const in style file.

packages/components/Avatar/src/Avatar.types.ts
ValentinaKozlova: But there are a lot of changes and other platforms might add even more changes so it'll become hard to read

Saadnajmi:
A note: one of the requirements we've had on iOS (and maybe also Android?) was to add a rainbow gradient border for the ring. This is achievable with the image background component: https://reactnative.dev/docs/imagebackground 

However, I think we ran into issues because the ring is generally the border of a view and not the view itself. Does this PR change that?
Also feel free to leave out of scope, I was forward thinking a bit.

packages/components/Avatar/src/Avatar.types.ts
ankraj12: Taking this offline - this approach was the recommended/ agreed upon approach for handling type differences in endpoints.

ankraj12:
> A note: one of the requirements we've had on iOS (and maybe also Android?) was to add a rainbow gradient border for the ring. This is achievable with the image background component: https://reactnative.dev/docs/imagebackground
> 
> However, I think we ran into issues because the ring is generally the border of a view and not the view itself. Does this PR change that? Also feel free to leave out of scope, I was forward thinking a bit.

You are right on the actual glow ring being border of a View. I am not aware if Android requested this yet - would you have a screen shot of what was the expected UX on iOS maybe - I can take a look if that somehow works now. I haven't really changed the component structure for inner rings although.

rohanprasadofficial (approved):

packages/components/Avatar/src/Avatar.types.ts
ankraj12: I am going to merge this PR changes. The suggestions on handling type differences are not specific to this code - and if needed there can be another change review to update this guidance across. There is more active work for this component pending, so I would account for needed change in those! Please feel free to add any additional post merge comments that I can track back.

PR #2532, opened 2023-01-19
[Spinner V1] Define spinner props and slots

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

Adding the Spinner package with bare minimum code.
Defining props, tokens, and slots only

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Spinner/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@fluentui-react-native/experimental-spinner",
```
rurikoaraki, line 2: I think we can just use "spinner" here since that name isn't taken already

rurikoaraki (approved):

ankraj12:
Hi Lena, could you help share snippets of how the UI looks for this. We have work for Spinner Loading UI in android, and so trying to understand the current Spinner here, vs the implementation in :
`https://github.com/microsoft/fluentui-reactnative/blob/main/packages/experimental/ActivityIndicator/src/ActivityIndicator.mobile.tsx`

lenahong:
> Hi Lena, could you help share snippets of how the UI looks for this. We have work for Spinner Loading UI in android, and so trying to understand the current Spinner here, vs the implementation in : `https://github.com/microsoft/fluentui-reactnative/blob/main/packages/experimental/ActivityIndicator/src/ActivityIndicator.mobile.tsx`

This pr actually doesn't render any UI yet actually. For win32, I'm following similar design guidelines for web.

ankraj12:
> > Hi Lena, could you help share snippets of how the UI looks for this. We have work for Spinner Loading UI in android, and so trying to understand the current Spinner here, vs the implementation in : `https://github.com/microsoft/fluentui-reactnative/blob/main/packages/experimental/ActivityIndicator/src/ActivityIndicator.mobile.tsx`

> This pr actually doesn't render any UI yet actually. For win32, I'm following similar design guidelines for web.

> Okay! Would you happen to know how same/ different this Spinner and  the one in experimental/ActivityIndicator/src/ActivityIndicator would look like?

PR #2535, opened 2023-01-20
[Switch] Fix initial page load animation 

rohanprasadofficial:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

This PR targets to fix the animation behavior that was seen on page load itself which was not expected. 
Animation should only occur when the switch changes the state from ON to OFF & vice-versa.  

This also target to clean up some functions and reduce few code. 

Adding the videos below. 

### Verification

| https://user-images.githubusercontent.com/30728574/213670626-b76b982b-6841-4218-97d0-6cc32aa24f37.mov | https://user-images.githubusercontent.com/30728574/213670678-8380a6c0-7475-49ee-b14e-8600d323b2f8.mov |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Switch/src/__tests__/__snapshots__/Switch.test.tsx.snap
+++ b/packages/experimental/Switch/src/__tests__/__snapshots__/Switch.test.tsx.snap
@@ -168,7 +168,7 @@ exports[`Switch Disabled 1`] = `
         ],
       },
       "trackBackgroundStyle": Object {
-        "backgroundColor": "rgba(180, 214, 250, 1)",
+        "backgroundColor": "rgba(240, 240, 240, 1)",
       },
     }
   }
@@ -199,7 +199,7 @@ exports[`Switch Disabled 1`] = `
       style={
         Object {
           "alignItems": "center",
-          "backgroundColor": "rgba(180, 214, 250, 1)",
+          "backgroundColor": "rgba(240, 240, 240, 1)",
           "borderRadius": 100,
           "flexDirection": "row",
           "height": 32,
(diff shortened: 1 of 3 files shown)


packages/experimental/Switch/src/__tests__/__snapshots__/Switch.test.tsx.snap
```diff
@@ -168,7 +168,7 @@ exports[`Switch Disabled 1`] = `
         ],
       },
       "trackBackgroundStyle": Object {
-        "backgroundColor": "rgba(180, 214, 250, 1)",
+        "backgroundColor": "rgba(240, 240, 240, 1)",
```
rohanprasadofficial, line 171: This results in the change as before it was taking initial animation value before moving to the correct state , Now it take correct animation value directly.

rohanprasadofficial:
This bug fix targets the completion of all the pending task for switch for Android. 

FYI - @vishnu2410 @ankraj12

packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -37,62 +31,27 @@ export const useSwitch = (props: SwitchProps, animationConfig?: AnimationConfig)
...
+      const toValue = checkedState ? 0 : 1;
+      Animated.timing(trackBackgroundAnimation, {
+        toValue,
+        duration: isInit ? 0 : 500,
+        useNativeDriver: false,
+      }).start();
+
+      setIsInit(false);
+    }
+  }, [checked, checkedState]);
```
rurikoaraki, line 54: nit: doesn't look like this depends on checked anymore?

rurikoaraki (approved):

packages/experimental/Switch/src/useSwitch.ts
rohanprasadofficial: @rurikoaraki , I tried with it but we have removed function with all the animation code in it and will be handling with this function itself.So when also checked state chnages we will need call this.

PR #2547, opened 2023-01-24
Fix MenuPopover props not getting called

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

onDismiss/onShow overrides aren't getting called, let's make sure we map them in the useMenuPopover hook.
### Verification
Overrides are getting called (macOS & win32)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
+++ b/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
@@ -26,7 +26,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
   const onDismiss = React.useCallback(() => {
     props.onDismiss();
     setOpen(undefined, false /* isOpen */), [setOpen];
-  });
+  }, [props.onDismiss, setOpen]);
   const dismissBehaviors = isControlled ? controlledDismissBehaviors : undefined;
   const directionalHint = getDirectionalHint(isSubmenu, I18nManager.isRTL);
 
--- a/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
+++ b/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
@@ -101,7 +101,8 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
     props: {
       accessibilityRole,
       target: triggerRef,
-      onDismiss,
+      onDismiss: props.onDismiss ?? onDismiss,
+      onShow: props.onShow,
       directionalHint,
       dismissBehaviors,
       doNotTakePointerCapture,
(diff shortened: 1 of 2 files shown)


mischreiber (approved):

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -101,7 +101,8 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
     props: {
       accessibilityRole,
       target: triggerRef,
-      onDismiss,
+      onDismiss: props.onDismiss ?? onDismiss,
```
Saadnajmi, line 104: Can we override our local `onDismiss` to call `props.onDismiss if available?

Saadnajmi: Speaking of... do we need to do a similar thing for `onShow`? 

AKA, onDismiss is currently defined as
>const onDismiss = React.useCallback(() => setOpen(undefined, false /* isOpen */), [setOpen]);

should onShow be:

>const onShow = React.useCallback(() => setOpen(undefined, **true** /* isOpen */), [setOpen]);

Saadnajmi: Edit: I think the answer to that is no because setOpen(true) is done by some timers...

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -23,7 +23,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
 
   const { onKeyDown: onKeyDownProp, onKeyUp: onKeyUpProp } = props;
 
-  const onDismiss = React.useCallback(() => setOpen(undefined, false /* isOpen */), [setOpen]);
+  const onDismiss = props.onDismiss ?? React.useCallback(() => setOpen(undefined, false /* isOpen */), [setOpen]);
```
Saadnajmi, line 26: Sorry, I meant that `onDismiss` should call `props.onDismiss`. I imagine we always need to call `setOpen` so it would be bad if the user providing a callback broke that?

chiuam: I thought user overriding onDismiss implied that the setOpen logic would be handled in the override. But then I see CM does call props.onDismiss on top of that so let's follow suit

Saadnajmi (approved):

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -101,7 +104,8 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
     props: {
       accessibilityRole,
       target: triggerRef,
-      onDismiss,
+      onDismiss: onDismiss,
```
Saadnajmi, line 107: Nit: I think you can just do `onDismiss,` now :)

PR #2549, opened 2023-01-24
Menu v1 - scrolling in win32

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes
The following change was scoped to mac only due to ScrollView behaving abnormal in win32.
[https://github.com/microsoft/fluentui-react-native/pull/1175](Add ScrollView in MacOS)

Fixes 
- Support MaxHeight on MenuPopover (adds scrollview) 

Issues remaining after changes applies:
- If the user never sets a maxHeight but content did overflow, overflown contents would be cut off. Omitting the MaxHeight prop will be render the menu without a ScrollView.
- Width of the menu isn’t dynamic 

Using current control: 
- Always set a maxHeight value on MenuPopover when you want a scrollView 
- Current default width for a menu with a minWidth of 175, which is 15 pixels larger than the default menuItem width. If your menuItems’ content require a larger minWidth, you’d have to set a larger minWidth (Not supported yet) 

### Verification
menu without scrollview

menu with scrollview

### Pull request checklist

- [X] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuList/MenuList.types.ts
```diff
@@ -12,6 +12,7 @@ export interface MenuListTokens extends LayoutTokens, IBackgroundColorTokens {
    * Space between items in pixels
    */
   gap?: number;
+  hasMaxHeight?: MenuListTokens;
```
chiuam, line 15: Hmm, was it intentional to have it as a token?

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -30,19 +30,22 @@ const MenuStack = stagedComponent((props: React.PropsWithRef<IViewProps> & { gap
...
-    ...(Platform.OS === 'macos' && { scrollView: ScrollView }),
+    scrollView: ScrollView,
     ...(Platform.OS === 'macos' && { focusZone: FocusZone }),
   },
   useRender: (userProps: MenuListProps, useSlots: UseSlots<MenuListType>) => {
     const menuList = useMenuList(userProps);
     const menuContext = useMenuContext();
     const menuListContextValue = useMenuListContextValue(menuList);
-    const Slots = useSlots(menuList.props);
+    const Slots = useSlots(menuList.props, (layer) => menuListLookup(layer, menuList, userProps));
```
chiuam, line 48: For my own knowledge, why is this filter needed?

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -68,6 +71,10 @@ export const MenuList = compose<MenuListType>({
               </Slots.focusZone>
             </Slots.scrollView>
           </Slots.root>
+        ) : menuContext.hasMaxHeight ? (
```
chiuam, line 74: I'm contemplating if macOS should follow the same logic here.. @harrieshin @Saadnajmi @mischreiber

packages/components/Menu/src/MenuList/MenuList.types.ts
lenahong: Yes, in order to support this as a state for styling, the type should be a token

packages/components/Menu/src/MenuList/MenuList.tsx
lenahong: This looks up the tokens to apply styling to. In this case, `hasMaxHeight` value needs to be read in and forwarded to the appropriate slot.

packages/components/Menu/src/MenuList/MenuList.tsx
chiuam: opened a task for followup

chiuam (approved):

PR #2552, opened 2023-01-25
RadioGroup Style Fixes

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- Updated Radio indicator design using new design redlines. It has a "punch through" design now. Updated screenshots in spec to reflect change.
- Increased bottom margin between label subtext and the focus border so that the spacing matches that between the top of the indicator and the focus border. Reduced the spacing between the primary label and the subtext.
- Changed the focus rect to only be around the label, not the whole Radio.
- Removed enableFocusRing prop and manually applied focus border around labelContainer of Radio. Focus is handled on the root view, not labelContainer, so we need to manually set focus border to only appear on one part of the Radio.

### Verification

| image image image | image image image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
```diff
@@ -71,34 +71,38 @@ export const stylingSettings: UseStylingOptions<RadioProps, RadioSlotProps, Radi
...
           flexDirection: 'column',
           marginRight: tokens.labelMarginRight,
           marginLeft: tokens.labelMarginLeft,
           padding: tokens.labelPadding,
+          paddingRight: 0,
+          ...borderStyles.from(tokens, theme),
         },
       }),
-      ['labelAlignItems', 'labelMarginRight', 'labelMarginLeft', 'labelPadding'],
+      ['labelAlignItems', 'labelMarginRight', 'labelMarginLeft', 'labelPadding', ...borderStyles.keys],
```
rurikoaraki, line 85: For my own understanding, this would make it so that borders always apply around the label container instead of around the whole control, even though the whole control accepts mouse events. Does that design make sense?

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
```diff
@@ -7,7 +7,7 @@ import type { RadioTokens } from './Radio.types';
 export const defaultRadioTokens: TokenSettings<RadioTokens, Theme> = (t: Theme) =>
   ({
     rootHorizontalPadding: globalTokens.size40,
-    borderColor: t.colors.transparentStroke,
+    borderColor: t.colors.background,
```
rurikoaraki, line 10: This isn't a Fluent Design Token, can you use one of those instead? We should be fully using Fluent Tokens for V1 components.

rurikoaraki: Same comment for the win32 file

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
```diff
@@ -19,12 +19,12 @@ export const defaultRadioTokens: TokenSettings<RadioTokens, Theme> = (t: Theme)
     color: t.colors.neutralForeground3,
     radioOuterCircleSize: 20,
```
rurikoaraki, line 20: You should be able to use a global size token for these numbers.

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
gulnazsayed: The transparentStroke token is fixed to map to the correct color. Reverted back to using this design token.

packages/experimental/RadioGroup/src/Radio/RadioTokens.ts
gulnazsayed: Replaced.

packages/experimental/RadioGroup/src/Radio/Radio.styling.ts
gulnazsayed: Good question. In the options dialog, native has some limitations with being able to render a focus rect around the whole control. Design came to the conclusion that we want to have the borders always apply around the label container for consistency with native, so this is in line with the updated designs.

rurikoaraki (approved):

PR #2554, opened 2023-01-25
Adding FRNAppearanceAdditions native module

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adding a new native module to iOS, `FRNAppearanceAdditions`. This will be a hosting ground for APIs that wish were part of `react-native/appearance`, but aren't yet. The medium-term goal is to move equivalent logic from here upstream, but this will unblock FURN scenarios in the meantime.

### Verification

This code is not part of this PR (coming next), but here's some future code that rely on this native module, each doing their thing:

#### Switching betweek dark and dark-elevated at runtime, before:

https://user-images.githubusercontent.com/4934719/214698307-1ff4cbf3-3ae3-42a4-9717-1670e3900d93.mov

#### Switching betweek dark and dark-elevated at runtime, after:

https://user-images.githubusercontent.com/4934719/214698265-d72767d8-7c44-47c8-85f4-e37721f8ed9d.mov

#### Notification responding to size classes, before:

https://user-images.githubusercontent.com/4934719/214698302-f557f421-569b-4dcb-a915-3bfe1de8515a.mov

#### Notification responding to size classes, after:

https://user-images.githubusercontent.com/4934719/214698305-2d7b0fd2-597e-4331-9cea-f7531ca5659c.mov

### Pull request checklist

This PR has considered (when applicable):
- [x] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #2558, opened 2023-01-26
Full dark elevated theme support for iOS

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

As a follow-up to #2554, this PR does the actual work of enabling the dark elevated theme in iOS, as well as replacing some old size-class-guessing code with a real native hook.

### Verification

These videos were a preview event in #2554, but are now fully functional in this PR:

#### Switching betweek dark and dark-elevated at runtime, before:

https://user-images.githubusercontent.com/4934719/214698307-1ff4cbf3-3ae3-42a4-9717-1670e3900d93.mov

#### Switching betweek dark and dark-elevated at runtime, after:

https://user-images.githubusercontent.com/4934719/214698265-d72767d8-7c44-47c8-85f4-e37721f8ed9d.mov

#### Notification responding to size classes, before:

https://user-images.githubusercontent.com/4934719/214698302-f557f421-569b-4dcb-a915-3bfe1de8515a.mov

#### Notification responding to size classes, after:

https://user-images.githubusercontent.com/4934719/214698305-2d7b0fd2-597e-4331-9cea-f7531ca5659c.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

packages/components/Notification/src/__tests__/Notification.test.tsx
```diff
@@ -4,6 +4,12 @@ import * as renderer from 'react-test-renderer';
 import { checkRenderConsistency, checkReRender } from '@fluentui-react-native/test-tools';
 
 describe('Notification component tests', () => {
+  beforeAll(() => {
+    jest.mock('@fluentui-react-native/experimental-appearance-additions', () => ({
+      useHorizontalSizeClass: 'regular',
```
lyzhan7, line 9: Just curious, any reason why you chose regular over compact?

packages/theming/apple-theme/src/appleColors.ios.ts
```diff
@@ -215,11 +215,14 @@ function getFluentUIAppleDarkPalette(): ApplePalette {
...
 
 /** Creates a palette of colors for the apple theme, using the appropriate FluentUI Apple Palette based on appearance */
-export function paletteFromAppleColors(isLightMode: boolean): ThemeColorDefinition {
+export function paletteFromAppleColors(isLightMode: boolean, isElevated: boolean): ThemeColorDefinition {
   const fluentApple = isLightMode ? getFluentUIAppleLightPalette() : getFluentUIAppleDarkPalette();
 
   const appearance = Appearance.getColorScheme();
-  const mode = getCurrentAppearance(appearance, 'light');
+  let mode = getCurrentAppearance(appearance, 'light');
+  if (mode === 'dark' && isElevated === true) {
```
lyzhan7, line 223: Optional, I don't think you need the === true

packages/theming/theme-tokens/src/getTokens.ts
```diff
@@ -13,6 +13,8 @@ export function getAliasTokens(mode: AppearanceOptions) {
     return lightAliasTokens;
   } else if (mode === 'dark') {
     return darkAliasTokens;
+  } else if (mode === 'darkElevated') {
+    return darkAliasTokens;
```
lyzhan7, line 17: Optional could do mode === 'dark' && mode === 'darkElevated' on line 14 above

packages/theming/android-theme/src/createAndroidTheme.ts
```diff
@@ -5,9 +5,9 @@ import { Theme, ThemeOptions } from '@fluentui-react-native/theme-types';
 
 export function createAndroidTheme(options: ThemeOptions = {}): ThemeReference {
   const themeRef = new ThemeReference({} as Theme, () => {
-    // Stub out HC on Android
+    // Stub out HC and darkElevated on Android
     const current =
-      options.appearance == 'dynamic' || options.appearance == 'highContrast'
+      options.appearance == 'dynamic' || options.appearance == 'highContrast' || options.appearance == 'darkElevated'
```
lyzhan7, line 10: Should these all be strict equality

lyzhan7:
Another optional comment - so in this PR https://github.com/microsoft/fluentui-react-native/pull/2485 I refactored macOS a bit so that we call getCurrentAppearance once in createAppleTheme and then pass the mode down to the other theme creation calls. I was thinking we could do something similar for iOS, but that can also be done in a follow up pr

lyzhan7 (approved):

lyzhan7:
Also - was wondering if you were planning to add either jest or e2e tests to test that the colors are actually changing when dark elevated is selected?

packages/theming/apple-theme/src/appleColors.ios.ts
mischreiber: Ah you're right, that can go.

packages/theming/android-theme/src/createAndroidTheme.ts
mischreiber: Yes, almost certainly. Probably no real difference in this case, but any day we can remove a `==` is a good day. I might even consider adding it as a linting rule.

packages/components/Notification/src/__tests__/Notification.test.tsx
mischreiber: Not especially - I was testing on iPad and `'regular'` is the default there. Do you or @joannaqu happen to know what the default was in the tests before I started poking things? Looks like it was probably `'compact'` (in which case I'd like to switch it back for safety), though I haven't tested.

packages/theming/theme-tokens/src/getTokens.ts
mischreiber: `||` not `&&` I think, but otherwise yes!

packages/components/Notification/src/__tests__/Notification.test.tsx
mischreiber: Well I tested both ways, and it seems to come out the same regardless, so I'm sticking with `regular` for now. Lemme know if you think I should change it.

mischreiber:
> Another optional comment - so in this PR #2485 I refactored macOS a bit so that we call getCurrentAppearance once in createAppleTheme and then pass the mode down to the other theme creation calls. I was thinking we could do something similar for iOS, but that can also be done in a follow up pr

Yeah let's talk offline after this PR. It's a little trickier for iOS because we currently need `isLight` separately from the `mode` in `paletteFromAppleColors`.

Ultimately I'd like to talk about reworking `AppearanceOptions` to be a `struct` with all the interesting stuff we need (including elevation, light-vs-dark, high contrast, etc.) because that will also allow us to fully match method signatures across all endpoints, which means more code reuse opportunities!

mischreiber:
> Also - was wondering if you were planning to add either jest or e2e tests to test that the colors are actually changing when dark elevated is selected?

@lyzhan7 I don't see any real tests for any of the iOS theme stuff, including light vs dark mode. Do you think that should be a requirement for this PR (I'm open to it), or something to follow up with later?

packages/components/Notification/src/__tests__/Notification.test.tsx
lyzhan7: I think it was mostly tested on iPhone - the designs also primarily show examples of notification on an iPhone. Slight preference to use compact but fine leaving it too!

lyzhan7:
> > Another optional comment - so in this PR #2485 I refactored macOS a bit so that we call getCurrentAppearance once in createAppleTheme and then pass the mode down to the other theme creation calls. I was thinking we could do something similar for iOS, but that can also be done in a follow up pr
> 
> @lyzhan7 Yeah let's talk offline after this PR. It's a little trickier for iOS because we currently need `isLight` separately from the `mode` in `paletteFromAppleColors`.
> 
> Ultimately I'd like to talk about reworking `AppearanceOptions` to be a `struct` with all the interesting stuff we need (including elevation, light-vs-dark, high contrast, etc.) because that will also allow us to fully match method signatures across all endpoints, which means more code reuse opportunities!

Yeah I was thinking about switching it to a struct as well! In the meantime I had actually planned to switch the isLightMode parameter to mode sometime so doesn't need to be done in this PR - in `paletteFromAppleColors` I think we can also check for light mode using mode==='light'

> > Also - was wondering if you were planning to add either jest or e2e tests to test that the colors are actually changing when dark elevated is selected?
> 
> @lyzhan7 I don't see any real tests for any of the iOS theme stuff, including light vs dark mode. Do you think that should be a requirement for this PR (I'm open to it), or something to follow up with later?

Follow up later! Just wanted to make sure tests checking that other appearance options work as expected is on our radar. Doesn't need to be its own set of tests, but I think we should at least have a test somewhere (could be in a component whose appearance changes if the user used dark elevated) that fails if there was a regression to switching to/from dark elevated. I recently almost checked in a change for iOS that would have broken macOS dark mode since we didn't have any tests for it, so just want to make sure we guard against that for dark elevated

PR #2577, opened 2023-01-31
Not for check-in

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
+++ b/apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
@@ -1,0 +1,64 @@
+import * as React from 'react';
+import { Theme, useTheme } from '@fluentui-react-native/theme-types';
+import { themedStyleSheet } from '@fluentui-react-native/themed-stylesheet';
+import { THEME_TESTPAGE } from '../../../../E2E/src/Theme/consts';
+import { Test, TestSection, PlatformStatus } from '../Test';
+import { Text, View } from 'react-native';
+
+import { useFluentTheme } from '@fluentui-react-native/framework'
+
+const getThemedStyles = themedStyleSheet((theme: Theme) => {
+  return {
+    swatch: {
+      color: theme.colors.neutralForeground1,
+    },
+    rootView: {
+      backgroundColor: theme.colors.brandBackground
+    },
+  };
+});
+
+const Panel: React.FunctionComponent = () => {
+  const theme = useFluentTheme();
+  const theme2 = useTheme();
+  const themedStyles = getThemedStyles(theme);
+
+  return (
+    <View style={themedStyles.rootView}>
+      <Text
+        style={{ color: theme.colors.neutralForeground1, fontSize: 30 }}>RN Text</Text>
+      {/* Expected this to show blue text, but just showed normal RN default style text */}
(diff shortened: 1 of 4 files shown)


apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const themedStyles = getThemedStyles(theme);
+
+  return (
+    <View style={themedStyles.rootView}>
+      <Text
+        style={{ color: theme.colors.neutralForeground1, fontSize: 30 }}>RN Text</Text>
+      {/* Expected this to show blue text, but just showed normal RN default style text */}
+      <Text style={themedStyles.swatch} >Styled Text</Text>
+
+      <Text style={{ color: theme2.colors.brandStroke1}} >Styled Text</Text>
```
ankraj12, line 33: RNStockTest

@ejlayne  - This is how it looks on android, the properties in theme are accessible. The scenario is not clear although as the `theme.colors` object has a lot of extra definitions like [this](https://github.com/microsoft/fluentui-react-native/blob/391f3d691e286300dfad37cad539bed54d4436b4/packages/theming/theme-types/src/Color.types.ts#L25) -  which are [not filled correctly ](https://github.com/microsoft/fluentui-react-native/blob/609f7988754b77b86f2ed51d0ce4714fb2152293/packages/theming/android-theme/src/colorsTokens.ts#L71)for android, so accessing `theme.colors.primaryButtonText` would map to some stale color. In my understanding, they exist for the non V2 desings. At one point I check with @rurikoaraki on removing the non V2 entries from colors, but at this point they exist in code.

packages/theming/theming-utils/src/mapPipelineToTheme.android.ts
```diff
@@ -75,7 +77,10 @@ export function mapPipelineToTheme(pipelineOutput: any): AliasColorTokens {
     neutralStrokeDisabled: pipelineOutput.neutralStrokeDisabled.strokeColorRest,
 
     //Brand tokens.
-    brandBackground: pipelineOutput.brandBackground1.fillColorRest,
+    brandBackground: 'green',
+
+    // brandBackground: pipelineOutput.brandBackground1.fillColorRest,
```
ankraj12, line 82: This is just indicating the colors I overrode to easy check that the tester react native controls pick the overrides correctly for the theme property being accessed. The 3 updated values here are the ones referred in styles for the react native components in the test sample code.

apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
ankraj12: There is documentation here - https://github.com/microsoft/fluentui-react-native/blob/961b183e22df3bb9eac1eadc7d6b268ce3b71d52/docs/pages/Theming/Basics.md#accessing-theme-properties

Linking to confirm if you see need for any extensions of documentation. The only easily accessible property is colors, others like typography values do not seem to map very straight.

PR #2584, opened 2023-02-01
Polish E2E Tests: Homogenized test names and deprecated bad BasePage methods.

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR is a follow up to the effort to make E2E errors more informative and improve E2E test names.

For example, with later PRs, we decided to rephrase the names of AccessibilityRole tests - this PR goes to even out AccessibilityRole test naming on components whose changes were early in the error + name overhaul effort. 

This PR also fixes some errors that were missed, and it deprecates methods in the BasePage from previous E2E test iterations that should not be used anymore.
 
### Verification

Tests pass in CI.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/common/BasePage.ts
```diff
@@ -392,4 +373,39 @@ export abstract class BasePage {
 
   // Default timeout to wait until page is displayed (10s)
   waitForUiEvent = 25000;
+
+  /************************/
+  /** Deprecated Methods **/
+  /************************/
+
+  /** @deprecated Use the `click` method defined in BasePage. */
```
samuelfreiberg, line 381: I don't think we need this section. This isn't an external package, so partners don't need to change their implementation. All we need to do is replace all instances of these methods with the new method. 

That should suffice. And make sure the comments on the new methods provide enough detail on what the method does

apps/E2E/src/common/BasePage.ts
```diff
@@ -392,4 +373,39 @@ export abstract class BasePage {
 
   // Default timeout to wait until page is displayed (10s)
   waitForUiEvent = 25000;
+
+  /************************/
+  /** Deprecated Methods **/
```
samuelfreiberg, line 378: I don't think we need this section. This isn't an external package, so partners don't need to change their implementation. All we need to do is replace all instances of these methods with the new method.

That should suffice. And make sure the comments on the new methods provide enough detail on what the method does

samuelfreiberg (changes requested):

samuelfreiberg (approved):

PR #2585, opened 2023-02-01
Remove compound button from iOS

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Remove the compound button from iOS, following how Android did this (see[ this PR](https://github.com/microsoft/fluentui-react-native/pull/2207/files))
- Move the Android code returning null for the compound button into a shared .mobile.tsx file, return that for iOS/Android
- Update the tester
  - Compound button was being shown in multiple sections. In some places I added a check for iOS where there was an existing check for Android, to not show any compound buttons. In another place Android had its own custom section, so i just used that for iOS as well

Context for removing this - we got a bug about the compound button looking strange on iOS (not enough margins) but realized the compound button should just be removed altogether. The original bug will be moved to a platform that still offers the button.

Follow up work
- Android removed the toggle button in a similar way as the compound button. I was planning to do that in this PR as well but realized that would break the Platform status collapsible section (which opens/closes with a toggle button). We also use a toggle button in the Color Tokens section. I can start a conversation about whether this is an issue on Android as well and figure out what to replace the toggle button, and follow up in a future PR

### Verification

https://user-images.githubusercontent.com/78454019/216190181-b81ee180-6523-40bc-a556-a656bbe54482.mp4

https://user-images.githubusercontent.com/78454019/216190189-c645a86f-e988-4622-b551-0c3e4346e8b8.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
```diff
@@ -61,7 +61,7 @@ export const ButtonSizeTest: React.FunctionComponent = () => {
       <Button size="large" style={commonTestStyles.vmargin}>
         Large
       </Button>
-      {Platform.OS !== 'android' && (
+      {Platform.OS !== 'android' && Platform.OS !== 'ios' && (
```
Saadnajmi, line 64: nit: Can we move this to a 'isMobile' variable?

apps/fluent-tester/src/TestComponents/Button/ButtonTest.tsx
```diff
@@ -25,7 +25,8 @@ const buttonSections: TestSection[] = [
     component: ButtonIconTest,
   },
   ...Platform.select({
-    android: [null], //Following sections are not supported from Fluent Android
+    ios: [],
+    android: [], //Following sections are not supported from Fluent Android
```
Saadnajmi, line 29: Update comment?

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
lyzhan7: What do you think about making an isMobile method in @fluentui-react-native/framework? I was thinking I could do that and update all the references to android/ios platform checks then

apps/fluent-tester/src/TestComponents/Button/ButtonTest.tsx
```diff
@@ -53,7 +54,7 @@ const buttonSections: TestSection[] = [
     component: E2EButtonTest,
   },
   ...Platform.select({
-    android: [null], //Following sections are not supported from Fluent Android
+    android: [], //Following sections are not supported from Fluent Android
```
lyzhan7, line 57: Replaced [null] with [] because I got an error on iOS "Each child in a list should have a unique "key" prop" and this change got rid of the error

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
Saadnajmi: I answered this elsewhere, but I don't like that mainly because iPad is weird. Sometimes we want to filter if we iPad or not. For that reason, I've preferred just doing the checks individually

Saadnajmi: I'm not super against it, but it's a small enough thing that I wasn't sure it was necessary to add a new package for it / add it to an existing package

Saadnajmi: Actually with Mike's recent work with trait collections, that might be worth revising once that is upstreamed

lyzhan7: We could have separate isMobile/isMobileSmallScreen checks? isMobile would always include iPad, isMobileSmallScreen would only include ipad compact mode. 

@mischreiber do you have any preferences?

apps/fluent-tester/docs/android.md
```diff
@@ -43,7 +43,7 @@ adb reverse tcp:8081 tcp:8081
 yarn start
 ```
 
-- APKs: The emulator is running on arm64 so any APKs that need to be installed need to be for arm64. Here is a [link](https://www.apkmirror.com/apk/google-inc/android-accessibility-suite/android-accessibility-suite-13-1-0-501229322-release/android-accessibility-suite-13-1-0-501229322-2-android-apk-download/) to the accessibility APK which includes TalkBack (voiceover)
```
Saadnajmi: Thanks!

lyzhan7: Hmm neither of these are working :/

lyzhan7: Also, pretty sure I should be able to test this locally without having to push commits but running yarn test-links just hangs for me

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
Saadnajmi: To further elaborate my thought: an `isMobile` check could include the horizontal size class (i.e: `Platform.OS === 'iOS' && sizeClass === 'compact')`. But it's still context dependent whether we care about the size class. In this case, I don't think we'd want the compound button even on iPad.


I'd prefer that any "consolidation" of platform checks happen as a separate PR. I feel like it's something we get in a loop about and I'd rather it not block a PR.

apps/fluent-tester/docs/android.md
Saadnajmi: I just realized.. we shouldn't really have a link to apkmirror IMO, that's not an official Android website. Can we just ignore or remove it?

lyzhan7:
The Button jest tests are currently running on iOS, so with this change we will be losing coverage on the compound button. There is already a PR proposal here for multi platform jest tests: https://github.com/microsoft/fluentui-react-native/pull/2506/files and that should fix this issue

apps/fluent-tester/src/TestComponents/Button/ButtonSizeTestSection.tsx
lyzhan7: Yep I'll bring it up in the next sync, will leave this PR as is

PR #2590, opened 2023-02-02
chore: enable `@rnx-kit/no-const-enum`

tido64:
### Platforms Impacted

- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Disallow use of `const enum`. More on why here: https://hackmd.io/bBcd6R-1TB6Zq95PSquooQ

Luckily, they are only used in E2E. I've tried replacing the ones found there, but there are a few I couldn't find a satisfactory replacement for. For now, I've added suppressors since this package doesn't get published anyway.

Also noticed that we use `await expect(…)` in lots of places throughout the repository. `expect` is not an async function. I'll clean those up in a separate PR.

### Verification

Existing tests should pass.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/E2E/.eslintrc.js
+++ b/apps/E2E/.eslintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ['@fluentui-react-native/eslint-config-rules'],
+};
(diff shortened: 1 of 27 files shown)


Saadnajmi:
I think this touches some files that Andrew Coates's PRs already touch, so order of operations is rather those merge first

Saadnajmi (approved):

PR #2598, opened 2023-02-03
macOS: Update submenu indicator color

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We are showing the submenu indicator color as blue which is wrong, let's set it with the same alias token as the default one.
### Verification

|before |after |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
+++ b/packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
@@ -14,6 +14,7 @@ export const defaultMenuItemTokens: TokenSettings<MenuItemTokens, Theme> = (t: T
   gap: globalTokens.size40,
   paddingHorizontal: 5, // hardcoded for now to match ContextualMenu
   paddingVertical: 3, // hardcoded for now to match ContextualMenu
+  submenuIndicatorColor: t.colors.neutralForeground2,
   submenuIndicatorPadding: globalTokens.sizeNone,
   submenuIndicatorSize: 16,
   focused: {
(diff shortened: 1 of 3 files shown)


harrieshin:
do we still have a separate bug tracking > doesn't align correctly next to a text? also are we using the right icon? feels like it should be thicker  
image

harrieshin (approved):

Saadnajmi (approved):

packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
```diff
@@ -14,6 +14,7 @@ export const defaultMenuItemTokens: TokenSettings<MenuItemTokens, Theme> = (t: T
   gap: globalTokens.size40,
   paddingHorizontal: 5, // hardcoded for now to match ContextualMenu
   paddingVertical: 3, // hardcoded for now to match ContextualMenu
```
Saadnajmi, line 16: Nit: ContextualMenu -> NSMenu?

chiuam:
> do we still have a separate bug tracking > doesn't align correctly next to a text? also are we using the right icon? feels like it should be thicker image

Created a followup task, thanks!

packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts
chiuam: yup yup

Saadnajmi (approved):

PR #2600, opened 2023-02-07
Update submenu indicator svg

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Updating the submenu indicator SVG to use the Fluent chevron icon (size 16 fill). This change increases the chevron thickness and center aligns to the menu item text.

### Verification

| before-win32 |after-win32 |
| before | after|

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuItem/MenuItem.styling.ts
```diff
@@ -52,11 +52,7 @@ export const stylingSettings: UseStylingOptions<MenuItemProps, MenuItemSlotProps
           color: tokens.submenuIndicatorColor,
           height: tokens.submenuIndicatorSize,
           width: tokens.submenuIndicatorSize,
-          viewBox:
-            '0 0 ' +
-            (tokens.submenuIndicatorSize - tokens.submenuIndicatorPadding * 2) +
-            ' ' +
-            (tokens.submenuIndicatorSize - tokens.submenuIndicatorPadding * 2),
+          viewBox: '0 0 ' + tokens.submenuIndicatorSize + ' ' + tokens.submenuIndicatorSize,
```
amgleitman, line 55: Any reason why we're not using [string interpolation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#string_interpolation) here?

amgleitman (approved):

harrieshin:
why does win32 > shows up at the end of the menu while macOS > shows up right next to the text?

mischreiber (commented):
Color of the chevron and the text don't seem to match (the chevron is much darker than the text) -- is that intentional?

packages/components/Menu/src/MenuItem/MenuItem.styling.ts
chiuam: Thanks for the suggestion! fixed in 56ec502348e57cf01aaef0249993cb3444cc752a

chiuam:
> why does win32 > shows up at the end of the menu while macOS > shows up right next to the text?

Thanks for pointing that out. Created a task under the Menu crew to address it in a separate PR :)

chiuam:
> Color of the chevron and the text don't seem to match (the chevron is much darker than the text) -- is that intentional?

Agreed and not intentional at all. The font color matches ContextualMenu but seems to deviate from NSMenu:
image

Might be a good idea to update it with a darker shade of black?

chiuam:
> > Color of the chevron and the text don't seem to match (the chevron is much darker than the text) -- is that intentional?
> 
> Agreed and not intentional at all. The font color matches ContextualMenu but seems to deviate from NSMenu: image
> 
> Might be a good idea to update it with a darker shade of black?

Created a task for followup. Will merge the changes for now.

PR #2614, opened 2023-02-10
iOS Get initial trait collection correctly on boot

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

With the recent work to expose iOS trait collection through a new Native Module, one of the remaining issues was that we weren't getting some of the right initial trait collection values at boot. When we tried to access traits specific to a view - horizontal size class (compact vs regular) and user interface level (base vs elevated) - we would always just get undefined or the default value. Note that the other trait we're exposing, accessibility contrast, is a system wide trait so we were always getting the correct value for that.

After further investigation, the issue was that we were relying on getting the RCTUserInterfaceStyleDidChangeNotification to get fired from the RCTRootView traitCollectionDidChange method, which would contain the right values on boot. It turns out that this notification _was_ getting posted as expected, and contained the right values, but it gets posted before the AppearanceAdditions native module has even set up the ability to listen to any notifications (i.e. before startObserving gets called), so it didn't receive the initial notification. 

This fix 1) initializes the native module with the right values for trait collection, which needed to be retrieved from the rootView (added a comment inline for why), and 2) adds a call from JS to access these values when the JS constructor is called. 

Notes
- We're initializing the native module's trait collection values in startObserving, but maybe this should be done in an init method?
- Making all the methods get called on the main thread needed to be added since [UIApplication sharedApplication] can only be accessed on the main thread

### Verification

Tested by visually verifying the size class. On iPad compact mode, if we've determined the right size class (compact) at boot, we'll render the single pane view (like what you would see on mobile). If we haven't determined the right size class, we render the double pane view (like what you would see on desktop)

In the before screen recording you can see that at boot we render with the double pane view which isn't correct - only when we switch back to it do we get the single pane view. In the after screen recording the app boots with the single pane view at the start.

https://user-images.githubusercontent.com/78454019/218217167-5f7d4acc-4f34-4339-a607-1f0b9ebcb025.mov

https://user-images.githubusercontent.com/78454019/218217195-c68a748e-df80-4162-9e46-573ab524ffa2.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
+++ b/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
@@ -124,15 +124,6 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(accessibilityContrastOption)
 
 #pragma mark - Event processing
 
-RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getInitialTraitCollection)
-{
-    return @{
-        @"horizontalSizeClass": _horizontalSizeClass,
-        @"userInterfaceLevel": _userInterfaceLevel,
-        @"accessibilityContrastOption": _accessibilityContrastOption,
-    };
-}
-
 - (void)appearanceChanged:(NSNotification *)notification {
     if (_hasListeners) {
         UITraitCollection *traitCollection = [[notification userInfo] valueForKey:RCTUserInterfaceStyleDidChangeNotificationTraitCollectionKey];
--- a/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
+++ b/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
@@ -105,14 +105,22 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(accessibilityContrastOption)
 - (void)startObserving {
     _hasListeners = YES;
     
-    // [UITraitCollection currentTraitCollection] never returns the correct trait collection, presumably because FRNAppearanceAdditions isn't a view
-    // so it never gets updated with the right traitCollection (which happens when a view gets added to the view hierachy). In order
-    // to get the right trait collection we need to access the root view.
-    UITraitCollection *rootViewTraitCollection = [[[[[UIApplication sharedApplication] delegate] window] rootViewController] traitCollection];
+    // Note that [UITraitCollection currentTraitCollection] always returns the same default trait collection,
+    // presumably because FRNAppearanceAdditions isn't a view, so it never gets updated with the right traitCollection
+    // (which happens when a view gets added to the view hierachy). In order to get the right trait collection,
+    // we need to access a view that's been added to the view hierarchy
+    UIViewController *viewControllerWithInitialTraitCollection = RCTPresentedViewController();
+    UITraitCollection *initialTraitCollection;
     
-    _horizontalSizeClass = RCTHorizontalSizeClassPreference(rootViewTraitCollection);
-    _userInterfaceLevel = RCTUserInterfaceLevelPreference(rootViewTraitCollection);
-    _accessibilityContrastOption = RCTAccessibilityContrastPreference(rootViewTraitCollection);
+    if (viewControllerWithInitialTraitCollection != nil) {
+        initialTraitCollection = [viewControllerWithInitialTraitCollection traitCollection];
+    } else {
+        initialTraitCollection = [UITraitCollection currentTraitCollection];
+    }
+    
+    _horizontalSizeClass = RCTHorizontalSizeClassPreference(initialTraitCollection);
+    _userInterfaceLevel = RCTUserInterfaceLevelPreference(initialTraitCollection);
+    _accessibilityContrastOption = RCTAccessibilityContrastPreference(initialTraitCollection);
     
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(appearanceChanged:)
--- a/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
+++ b/packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
@@ -104,6 +104,13 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(accessibilityContrastOption)
 
 - (void)startObserving {
     _hasListeners = YES;
+    
+    UITraitCollection *rootViewTraitCollection = [[[[[UIApplication sharedApplication] delegate] window] rootViewController] traitCollection];
+    
+    _horizontalSizeClass = RCTHorizontalSizeClassPreference(rootViewTraitCollection);
+    _userInterfaceLevel = RCTUserInterfaceLevelPreference(rootViewTraitCollection);
+    _accessibilityContrastOption = RCTAccessibilityContrastPreference(rootViewTraitCollection);
+    
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(appearanceChanged:)
                                                  name:RCTUserInterfaceStyleDidChangeNotification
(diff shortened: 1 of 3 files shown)


packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -103,11 +97,23 @@ + (BOOL)requiresMainQueueSetup {
...
 - (void)startObserving {
     _hasListeners = YES;
-    _horizontalSizeClass = RCTHorizontalSizeClassPreference(nil);
-    _userInterfaceLevel = RCTUserInterfaceLevelPreference(nil);
-    _accessibilityContrastOption = RCTAccessibilityContrastPreference(nil);
+    
+    // [UITraitCollection currentTraitCollection] never returns the correct trait collection, presumably because FRNAppearanceAdditions isn't a view
+    // so it never gets updated with the right traitCollection (which happens when a view gets added to the view hierachy). In order
+    // to get the right trait collection we need to access the root view.
+    UITraitCollection *rootViewTraitCollection = [[[[[UIApplication sharedApplication] delegate] window] rootViewController] traitCollection];
```
harrieshin, line 111: we can't rely on uiapplication sharedapplication. this will not work reliably with multiwindow in in ipad or external screens.

lyzhan7: Thanks for pointing that out, I think @mischreiber actually also mentioned to me at some point to consider multiwindow. Will look into a better way of getting the right rootviewcontroller that works when there are multiple windows

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -103,11 +97,31 @@ + (BOOL)requiresMainQueueSetup {
...
     _hasListeners = YES;
-    _horizontalSizeClass = RCTHorizontalSizeClassPreference(nil);
-    _userInterfaceLevel = RCTUserInterfaceLevelPreference(nil);
-    _accessibilityContrastOption = RCTAccessibilityContrastPreference(nil);
+    
+    // Note that [UITraitCollection currentTraitCollection] always returns the same default trait collection,
+    // presumably because FRNAppearanceAdditions isn't a view, so it never gets updated with the right traitCollection
+    // (which happens when a view gets added to the view hierachy). In order to get the right trait collection,
+    // we need to access a view that's been added to the view hierarchy
+    UIViewController *viewControllerWithInitialTraitCollection = RCTPresentedViewController();
```
Saadnajmi, line 112: I have a vague feeling this method might have a diff in it between React Native and React Native macOS. Might be worth double checking

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: I replaced `[UIApplication sharedApplication]` with something I found in `RCTUtils`, that we're also using the datepicker (`RCTPresentedViewController`). Here is a snippet of code from `RCTUtils` that describes what `RCTPresentedViewController` is returning:


```objective-c
UIApplication *__nullable RCTSharedApplication(void)
{
  if (RCTRunningInAppExtension()) {
    return nil;
  }
  return [[UIApplication class] performSelector:@selector(sharedApplication)];
}

UIWindow *__nullable RCTKeyWindow(void)
{
  if (RCTRunningInAppExtension()) {
    return nil;
  }

  // TODO: replace with a more robust solution
  for (UIWindow *window in RCTSharedApplication().windows) {
    if (window.keyWindow) {
      return window;
    }
  }
  return nil;
}

UIViewController *__nullable RCTPresentedViewController(void)
{
  if ([RCTUtilsUIOverride hasPresentedViewController]) {
    return [RCTUtilsUIOverride presentedViewController];
  }

  UIViewController *controller = RCTKeyWindow().rootViewController;
  UIViewController *presentedController = controller.presentedViewController;
  while (presentedController && ![presentedController isBeingDismissed]) {
    controller = presentedController;
    presentedController = controller.presentedViewController;
  }

  return controller;
}
```

I tried turning on the iPad multiwindow setting locally, and while the tester generally works as expected, there are some situations that don't behave as expected. Ex. can't have a single pane fluenttester instance on one side, and a double fluenttester instance on the other


https://user-images.githubusercontent.com/78454019/218613360-607e988d-ca41-4203-bb21-e42a22c09ffc.mov

I think supporting that scenario and others (ex. dark theme on one side, dark elevated on the other) will require a lot more thought that's beyond the scope of this PR. Also the issue seen above isn't a regression introduced by this PR, this has always been the behaviour (even before FRNAppearanceAdditions was added)

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: These look the same to me but let me know if I'm not comparing the right things

https://github.com/facebook/react-native/blob/main/React/Base/RCTUtils.m
https://github.com/microsoft/react-native-macos/blob/main/React/Base/RCTUtils.m#L612

Saadnajmi: I guess not. Thanks for confirming!

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -103,11 +97,31 @@ + (BOOL)requiresMainQueueSetup {
     return @[ @"appearanceChanged" ];
 }
 
+- (dispatch_queue_t)methodQueue
```
harrieshin, line 100: can we rename this to be more obvious like ui thread or main thread?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -103,11 +97,31 @@ + (BOOL)requiresMainQueueSetup {
...
+{
+    return dispatch_get_main_queue();
```
harrieshin, line 102: is it worth checking if we are already on mainthread just return the current thing and if we are on the background thread use get the main queue kind of logic? or are you trying to always do it on the next main queue cycle?

harrieshin (approved):

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: Do you mean the name "methodQueue"? This is an overridden method from RCTBridgeModule so don't think I can change the method name

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: This bit of code comes from https://reactnative.dev/docs/native-modules-ios#threading 

Documentation on methodQueue:

```objective-c
/**
 * The queue that will be used to call all exported methods. If omitted, this
 * will call on a default background queue, which is avoids blocking the main
 * thread.
 *
 * If the methods in your module need to interact with UIKit methods, they will
 * probably need to call those on the main thread, as most of UIKit is main-
 * thread-only. You can tell React Native to call your module methods on the
 * main thread by returning a reference to the main queue, like this:
 *
 * - (dispatch_queue_t)methodQueue
 * {
 *   return dispatch_get_main_queue();
 * }
 *
 * If you don't want to specify the queue yourself, but you need to use it
 * inside your class (e.g. if you have internal methods that need to dispatch
 * onto that queue), you can just add `@synthesize methodQueue = _methodQueue;`
 * and the bridge will populate the methodQueue property for you automatically
 * when it initializes the module.
 */
@property (nonatomic, strong, readonly) dispatch_queue_t methodQueue;
```

Without this code (if I comment out this block) I get this error

Screenshot 2023-02-16 at 1 47 27 PM

"is it worth checking if we are already on mainthread just return the current thing and if we are on the background thread use get the main queue kind of logic" do you mean something like this?

Screenshot 2023-02-16 at 1 45 46 PM

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
Saadnajmi: I think this has to stay as is, yeah.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
Saadnajmi: I always thought this method was a "What thread should I run native code on?". So I'm not sure that check makes sense. This is more of a declaration "Run this module on X thread".

mischreiber: Yeah, this is a necessary bit of native module logic and should not be changed.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
mischreiber: Yeah, I agree that this is not the *final* word on these changes, but @lyzhan7's changes represent a significant improvement to the status quo.

mischreiber (approved):
As mentioned earlier, this still does not cover 100% of multiwindow cases, but I believe that it's a sufficient improvement over the current version that we should still take the change as-is.

PR #2616, opened 2023-02-12
Android Tokens -> Mobile Tokens

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

A number of components were tokenized for Android. These designs are largely similar if not identical to the iOS designs. Let's copy over the tokens to iOS so that both mobile platforms benefit. Most of these components were in the backlog / experimental for iOS, so this change should be relatively safe.

I did the change by moving the contents of `<component>Tokens.android.ts` into a shared `<component>Tokens.mobile.ts` file for components that share designs. For the ones that don't (Checkbox, Radio), I just copied the tokens for now.

Affected components:
- Avatar
- Checkbox
- Radio / RadioGroup
- Switch
- Shimmer

Caveats:
- The Switch "knob" for iOS does not have a Shadow, as I discovered `<Shadow>` doesn't support wrapping Animated components. As such, I added a note to the Shadow spec that Animated components aren't supported.
- Checkbox and Radio actually have different designs for iOS (they are circular with a checkmark). For now.. I think copying the Android design still leaves us in a better state as we're not shipping these controls yet anyway.

### Verification

Took EJ's "FURN Android Playground", and had it render on iOS. I thought it looks much better :).

#### Before 
Screenshot 2023-02-12 at 12 38 21 AM 

#### After
Screenshot 2023-02-12 at 12 37 59 AM

https://user-images.githubusercontent.com/6722175/218294648-05bf8ea3-3a0c-4ee9-af75-6092f9e94ed1.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


ankraj12 (commented):
Thank you for these changes! Going ahead I also think we can keep mobile endpoint components coded common from start, and diverge only when needed between android and iOS for consistency across codebase.

Do you also have any thoughts around how the Fluent tester in iOS should show up these components? Android has been mostly removing examples that are non-fluent variations from tester as well - but probably that comes when iOS plans on releasing these components.

ankraj12 (approved):

rohanprasadofficial:
nit description update :  <component>Tokens.android.ts into a shared <component>Tokens.android.ts file.  to <component>Tokens.android.ts into a shared <component>Tokens.mobile.ts file. 


I am ok with this approach as i am also seeing most of the design are almost same with minor differences , i think we can have a quick platform check such cases until there is large divergence. 

For eg : Currently i changed the new Divider for Android felt it is same as iOS so handled both the parts here : https://github.com/microsoft/fluentui-react-native/pull/2617  

but if you check:  packages/experimental/Divider/src/DividerTokens.android/ios.ts , there is no much difference only the thickness differ , that might be handled using Platform check and don't need the divergence on the file.

rohanprasadofficial (approved):

chiuam (approved):

Saadnajmi:
Thanks @ankraj / @rohanprasadofficial for the feedback! This was also a learning experience for me of what works well when designs are similar. A few notes:


1) I think as a starting point for platforms where designs are mostly identical, this kind of consolidation into a shared `.mobile` file makes sense. This could also make sense for a desktop control where maybe `windows` and `macOS` share a spec. I haven't seen a case of that yet though :D. As soon as the platform differences become more than a few lines, I think it's fine to split them up again.

2) I actually found out that the iOS designs for Checkbox and Radio are actually quite different between iOS and Android. I think I'll keep my change as is because I still think the components look better with my change than without. But it further emphasizes that I should be clear these components are not production ready :).

3) Differences in Fluent Tester -> I didn't notice if they were "if (Platform.OS === 'android')" checks or "isMobile" checks. I'll take a second look. I don't have a strong opinion yet on how to properly handle that.. I know different people have handled it in different ways. I haven't felt that one way was strong enough to make an official recommendation.

packages/components/Avatar/src/AvatarTokens.android.ts
```diff
@@ -1,171 +1,3 @@
-import type { Theme } from '@fluentui-react-native/framework';
-import type { TokenSettings } from '@fluentui-react-native/use-styling';
-import type { AvatarTokens } from '.';
-import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import { defaultAvatarTokens } from './AvatarTokens.mobile';
```
lyzhan7, line 1: Does just `export { defaultAvatarTokens } from './AvatarTokens.mobile' work

packages/experimental/Shadow/SPEC.md
```diff
@@ -48,9 +48,10 @@ If `undefined` is passed into the `shadowToken` prop, no shadow will be rendered
 
 ## Implementation details
 
-In order to get around the React Native restriction of one Shadow per View, the Shadow component adds an extra View around the original View. One of the shadows is placed on the original View and the other is placed on the extra View.  The Shadow component smartly extracts the original View's style props so that the extra View isn't noticable (ex. will always be the same size as the original View, padding/margins are set correctly, etc.)
+In order to get around the React Native restriction of one Shadow per View, the Shadow component adds an extra View around the original View. One of the shadows is placed on the original View and the other is placed on the extra View. The Shadow component smartly extracts the original View's style props so that the extra View isn't noticable (ex. will always be the same size as the original View, padding/margins are set correctly, etc.)
 
 ## Notes
 
+- Known issue: The Shadow component does not currently work if you are wrapping an Animated component.
```
lyzhan7, line 55: Is this something we'd want to eventually support? If so can you file a bug & assign to me

Saadnajmi: I don't know if it is. It might be possible to work around it, but I intentionally kept that out of this PR. If/when we start Switch for iOS, I think we can revisit.

packages/components/Checkbox/src/CheckboxTokens.mobile.ts
```diff
@@ -0,0 +1,48 @@
+import type { Theme } from '@fluentui-react-native/framework';
+import type { TokenSettings } from '@fluentui-react-native/use-styling';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import type { CheckboxTokens } from './Checkbox.types';
+import { Platform } from 'react-native';
+
+// No Tokens for Sizes , hovered and circular checkbox as they're not supported for Android Platform
```
lyzhan7, line 7: Should we change comment to be "not supported on mobile platforms"

packages/experimental/RadioGroup/src/Radio/RadioTokens.mobile.ts
```diff
@@ -0,0 +1,58 @@
...
+    marginTop: 24 - 20, // Size - Outer Circle Size
+    marginBottom: 24 - 20, // Size - Outer Circle Size
+    marginLeft: 24 - 20, // Size - Outer Circle Size
+    marginRight: 24 - 20, // Size - Outer Circle Size
+    radioOuterCircleSize: 20,
+    radioInnerCircleSize: 10,
+    radioBorderStyle: 'solid',
+    radioBorderWidth: globalTokens.stroke.width15,
+    rippleRadius: globalTokens.size160,
+    rippleColor: '#D4D4D4',
```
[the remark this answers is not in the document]
lyzhan7: Although, thinking about it more, if iOS/android differ in terms of whether or not they have a ripple, maybe we shouldn't be combining them into a single file

packages/components/Avatar/src/AvatarTokens.android.ts
Saadnajmi: Somehow I get the error `@rnx-kit/no-const-enum` when I do so. I'll take a deeper look later.

packages/components/Checkbox/src/CheckboxTokens.mobile.ts
Saadnajmi: Yes, good catch!
I was actually thinking of splitting this into `.android` and `.ios` because these controls actually have very different specs on Android/iOS.

packages/experimental/RadioGroup/src/Radio/RadioTokens.mobile.ts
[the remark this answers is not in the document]
Saadnajmi: Yeah, we can put this in a `Platform.OS` check. Thanks for the catch!

packages/components/Avatar/src/AvatarTokens.android.ts
lyzhan7: Ah okay interesting. I think for compoundbutton I was able to just export it in one line

lyzhan7 (approved):

packages/components/Avatar/src/AvatarTokens.android.ts
Saadnajmi: I take it back, I think it works!

packages/components/Checkbox/src/CheckboxTokens.mobile.ts
Saadnajmi: I ended up splitting Checkbox and Radio back to separate files altogether.  @ankraj / @rohanprasadofficial FYI.

packages/experimental/RadioGroup/src/Radio/RadioTokens.mobile.ts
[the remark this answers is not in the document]
Saadnajmi: No longer relevant

PR #2617, opened 2023-02-13
[Mobile] Divider Implementation & insetSize fix

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
This PR targets the addition of Divider V1 on mobile platforms. 
This consists of Tester app changes for demonstration which is file separated because the usage of Divider on Mobile is not as rich as Desktop due to screen size constraint 
This also consist of a fix in the insetsize prop which will be applied to only paddingStart on mobile instead of both side padding. 

**Android Tokens :** 
Color of Divider : neutralStroke2
thickness : 1 (strokewidth10)

**iOS Tokens :** 
Color of Divider : neutralStroke2
thickness : 0.5 (strokewidth05)

### Verification

Visual verification and RTL verification as well for both platforms (ios and android) 

| Android  - Dark                                      | Android - Light                                      | iOS - Light  |
|----------------------------------------------|--------------------------------------------|-----------------------------------|
| MicrosoftTeams-image (21) | MicrosoftTeams-image (20) | MicrosoftTeams-image (22) |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rohanprasadofficial:
@Saadnajmi @lyzhan7 , can you review this from iOS perspective. 
@lawrencewin , can you review this so it doesn't break win32. 
Thanks

packages/experimental/Divider/src/Divider.styling.ts
```diff
@@ -49,7 +52,7 @@ export const useDividerSlotProps = (props: DividerProps, tokens: DividerTokens)
             }
           : {
               flexDirection: 'row',
-              paddingHorizontal: props.insetSize,
+              ...(isMobile ? { paddingStart: props.insetSize } : { paddingHorizontal: props.insetSize }), // insetSize ono iOS and Android is on start only.
```
Saadnajmi, line 55: Nit: spelling.

packages/experimental/Divider/src/Divider.types.ts
```diff
@@ -14,56 +14,67 @@ export interface DividerProps {
   /**
    * If a text or icon is passed, this dictates where content appears in the divider: at the start, centered, or towards the end.
    * @default 'center'
+   * @platform win32
```
Saadnajmi, line 17: Rather than this, could we add a note saying it doesn't apply to mobile? Reason being I imagine it might still apply to windows/macOS.

Saadnajmi (approved):
Approved with minor comments

lawrencewin (commented):
Is there currently logic that keeps the `after` slot from rendering on mobile? We currently render two lines using the `before` and `after` slots if content is passed in and only render `before` if no content is passed.

lawrencewin:
Other than that, nothing seems to break on windows.

lawrencewin (changes requested):
There doesn't seem to be anything broken on windows.

I think this looks good, but some logic to prevent people from passing in an icon / rendering text in `Divider.tsx` would be good. Plus, since the mobile divider is a single line, not rendering the `after` slot would also be a good change.

packages/experimental/Divider/src/Divider.types.ts
rohanprasadofficial: It makes sense , added that.

rohanprasadofficial:
> There doesn't seem to be anything broken on windows.
> 
> I think this looks good, but some logic to prevent people from passing in an icon / rendering text in `Divider.tsx` would be good. Plus, since the mobile divider is a single line, not rendering the `after` slot would also be a good change.

@lawrencewin I approach we're following for a component on Android (or Mobile) is that , if it is implemented on the other platform , we explicitly are not removing the props or stopping any props/slots from rendering. 
The source of the truth for platform what we support is spec.md and tester app , Anything we show on the tester app are supported for that <platform>. 

For example : If you see checkbox packages/components/Checkbox/src/Checkbox.tsx , we don't supported label before the checkbox and only support the after label but we have not removed explicitly on the implementation and documented it accurately on the tester and spec. Also Size as well. 
https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Checkbox/SPEC.md 


i don't have strong opinion for removing it but this what we're following for all components & would like keep it consistent for now.

packages/experimental/Divider/src/DividerTokens.ios.ts
```diff
@@ -0,0 +1,11 @@
+import { buildUseTokens } from '@fluentui-react-native/framework';
+import type { Theme } from '@fluentui-react-native/framework';
+import type { DividerTokens } from './Divider.types';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+
+export const useDividerTokens = buildUseTokens<DividerTokens>((theme: Theme) => ({
+  // base tokens
+  insetSize: 0,
+  lineColor: theme.colors.neutralStroke2,
+  thickness: globalTokens.stroke.width05,
```
ankraj12, line 10: This is from iOS designs? I would understand that we are not calling the iOS component completed unless the team validates back on designs and implementation - is that a fair statement?

packages/experimental/Divider/src/Divider.types.ts
```diff
@@ -14,56 +14,67 @@ export interface DividerProps {
...
    * @default false
+   * Note: This prop is not supported on mobile platforms(Android & iOS).
    */
   vertical?: boolean;
 }
 
 export interface DividerTokens extends LayoutTokens {
   /**
    * The color of the content passed into the divider. This is overriden if the `color` prop is set.
+   * Note: This prop is not supported on mobile platforms(Android & iOS).
```
ankraj12, line 48: I think all the properties in tokens do not need this note, tokens are customisation support out of the example Fluent variations. These properties are not props for the component, so a developer building a Divider would not reference them unless explicitly looking to style differently. This also aligns current guidance to document props and SPEC correctly, and to other components like - https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Checkbox/src/Checkbox.types.ts
 where the props

apps/fluent-tester/src/testPages.ts
```diff
@@ -117,7 +117,7 @@ export const tests: TestDescription[] = [
     platforms: ['ios'],
   },
   {
-    name: 'Divider',
+    name: 'Divider V1',
```
ankraj12, line 120: nit:  we need to mark this as V1 - I am of understanding that are adding V1 for components that have a non-v1 variant existing.

apps/fluent-tester/src/TestComponents/Divider/DividerTest.tsx
```diff
@@ -100,17 +102,36 @@ const dividerSections: TestSection[] = [
...
-    iosStatus: 'Backlog',
+    iosStatus: 'Experimental',
     macosStatus: 'Backlog',
-    androidStatus: 'Backlog',
+    androidStatus: 'Experimental',
   };
 
-  const description =
-    'A Divider is a visual separator that can contain content (text or an icon). Dividers can be horizontal or vertical. As of now, the component is being designed for win32 only.';
+  const description = 'A Divider is a visual separator that can contain content (text or an icon). Dividers can be horizontal or vertical';
```
ankraj12, line 127: Maybe update this to remove the note of text and icon for mobile?

ankraj12 (changes requested):
Could you confirm if you would track SPEC file and E2E tests for this component in android? How does the component look in RTL?

packages/experimental/Divider/src/DividerTokens.ios.ts
rohanprasadofficial: Correct. I have the behaviour and tokens files added but iOS engineers needs to validate on there end to mark it complete on iOS + Adding the E2E tests.

rohanprasadofficial:
> Could you confirm if you would track SPEC file and E2E tests for this component in android? How does the component look in RTL?


Yes i have task for it on the internal task board. 

RTL seems working on it. 

image

packages/experimental/Divider/src/Divider.types.ts
rohanprasadofficial: Done.

apps/fluent-tester/src/TestComponents/Divider/DividerTest.tsx
rohanprasadofficial: My understanding was different , thanks for the clarification.

apps/fluent-tester/src/testPages.ts
rohanprasadofficial: My understanding was different , thanks for the clarification. Changed it.

ankraj12:
> > Could you confirm if you would track SPEC file and E2E tests for this component in android? How does the component look in RTL?
> 
> Yes i have task for it on the internal task board.
> 
> RTL seems working on it.
> 
> image

Thanks for checking and confirming.

ankraj12 (approved):

lawrencewin (approved):

PR #2620, opened 2023-02-13
Update appium and node version

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Updating appium to latest to align to a version that doesn't use jsdoc package, which uses a package with a security risk and has no maintainer.
This requires updating the repo to node 16 to avoid issues with running appium. It doesn't work with node v14.
Updating node to 16 required removing using midgard-yarn and using plain yarn in our CI instead. I'm ok with this change since midgard-yarn doesn't seem to be maintained, and we'd have to move away from it if we want to move to yarn v3 anyway.
Also needed to add back script commands to install appium drivers in order to run the e2e tests.

Changes to node version and yarn install were made by @saadnajmi

### Verification

Using PR to test CI.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

lyzhan7 (approved):

.ado/templates/e2e-testing-uwp.yml
```diff
@@ -57,6 +57,12 @@ steps:
     condition: succeeded()
     displayName: 'Create success build variable'
 
+  - script: |
+      yarn appium driver install --source=npm appium-windows-driver
```
Saadnajmi, line 61: This one needs the extra "-source=npm" and others don't?

Saadnajmi: (same for win32)

Saadnajmi (approved):

.ado/templates/e2e-testing-uwp.yml
rurikoaraki: Yeah, because the windows driver is not in maintained by appium (though it kind of seems like it is), so it's not in their list of short names they recognize.

.ado/templates/e2e-testing-win32.yml
```diff
@@ -25,6 +25,12 @@ steps:
     condition: succeeded()
     displayName: 'Create success build variable'
 
+  - script: |
+      yarn appium driver install --source=npm appium-windows-driver
```
samuelfreiberg, line 29: Curious about this - why do we need to add it now?

We have the "appium-windows-driver" in apps/win32 package.json which should download the driver to apps/win32/node_modules

rurikoaraki: You still need to tell appium to install the driver, it doesn't automatically pick up the driver. I wasn't able to find a way to tell appium "use the already downloaded folder" without running into problems.

samuelfreiberg: Interesting, that seems like a weird regression with Appium then. Previously, I found it would automatically search in the current node_modules folder.

If your running it locally, does it also require you to run that driver install command?

rurikoaraki: Yeah, it refused to run unless I told it to install the driver.
It does have some behavior to automatically find the driver, but only using npx. I'm not sure if it's a appium regression or if it's a difference in behavior due to changing our node version.

.ado/templates/e2e-testing-android.yml
```diff
@@ -39,3 +45,9 @@ steps:
       platform: 'android'
       buildArtifacts: variables['task.Build.status']
       directory: $(Build.SourcesDirectory)/apps/E2E
+
+  - script: |
+      yarn appium driver uninstall uiautomator2
```
ayush547, line 50: Do we _need_ to uninstall the drivers? We could skip this right?
cc - @samuelfreiberg

rurikoaraki: No, we don't necessarily need to, assuming the VM fully resets itself between tests. I figure it's a good practice to clean up state before exiting the script, though.

ayush547: Alright, asked because we have not followed this for other setup templates

rurikoaraki:
> Any reason we can't merge this?

Yeah the Windows PR deterministically fails at E2E tests

PR #2629, opened 2023-02-15
[Proposal] Nest files with platform overrides in VS Code

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Inspired by https://twitter.com/Baconbrix/status/1625564914034085922?t=WaD1ZAg7BvMlA10EnLlz-Q&s=19

If we like this, we can perhaps add it to the official React Native VS Code extension :D 

VS Code lets you nest files together. Let's add the settings to do that in this project 😄. I modified the example in the tweet above to exclude `.test / .native / .node` and include `.macos / .windows / .win32`. I specifically didn't include stuff like `.mobile / .desktop` to hopefully make it more clear that metro won't automatically recognize those as platform overrides.

### Verification
Screenshot 2023-02-15 at 11 05 12 AM

https://user-images.githubusercontent.com/6722175/219103402-4cfb85f8-eaa4-48b4-9c30-08e9fef51924.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,13 @@
   "eslint.enable": true,
   "eslint.workingDirectories": [{ "mode": "auto" }], // infer working directory based on .eslintrc/package.json location
 
+  "explorer.fileNesting.enabled": true,
+  "explorer.fileNesting.patterns": {
+    "*.js": "${capture}.js.map, ${capture}.d.ts, ${capture}.d.ts.map",
+    "*.ts": "${capture}.android.tsx, ${capture}.ios.tsx, ${capture}.macos.tsx, ${capture}.windows.tsx, ${capture}.win32.tsx, ${capture}.android.ts, ${capture}.ios.ts, ${capture}.macos.ts, ${capture}.windows.ts, ${capture}.win32.ts",
+    "*.tsx": "${capture}.android.tsx, ${capture}.ios.tsx, ${capture}.macos.tsx, ${capture}.windows.tsx, ${capture}.win32.tsx, ${capture}.android.ts, ${capture}.ios.ts, ${capture}.macos.ts, ${capture}.windows.ts, ${capture}.win32.ts"
+  },
+
   "files.trimTrailingWhitespace": true,
   "files.associations": {
     "**/package.json.hbs": "json",


.vscode/settings.json
```diff
@@ -8,6 +8,13 @@
   "eslint.enable": true,
   "eslint.workingDirectories": [{ "mode": "auto" }], // infer working directory based on .eslintrc/package.json location
 
+  "explorer.fileNesting.enabled": true,
+  "explorer.fileNesting.patterns": {
+    "*.js": "${capture}.js.map, ${capture}.d.ts, ${capture}.d.ts.map",
+    "*.ts": "${capture}.android.tsx, ${capture}.ios.tsx, ${capture}.macos.tsx, ${capture}.windows.tsx, ${capture}.win32.tsx, ${capture}.android.ts, ${capture}.ios.ts, ${capture}.macos.ts, ${capture}.windows.ts, ${capture}.win32.ts",
```
rurikoaraki, line 14: Do we want to add mobile to this list?

rurikoaraki: Also I think it makes sense to separate *.ts and *.tsx. I don't think a file makes sense to be forked and not have the same extension.

Saadnajmi: Yeah, this is a proposal. The goal should be to make it easier for FURN/RN devs to understand RN structure. But I'm also perfectly happy to leave this as a personal setting that individual devs have to add themselves..

I think the reason ".ts" and ".tsx" are grouped together (according to the tweet) is because the files will get transpiled to JS, and be picked up by metro all the same. I haven't confirmed but it sounds like these 3 files like 
- AvatarTokens.js
- "AvatarTokens.ios.**tsx**
- AvatarTokens.android.**ts**
Speaking of, I omitted the `${capture}.<platform>.js` variants from this list but I could add them back.

Saadnajmi: Also note, enabling file nesting means other side effects happen,  like "yarn.lock" being nested under "package.json"...

rurikoaraki: Hm. Not sure how much I like that, but I guess it's just something you get used to.

rurikoaraki (approved):

.vscode/settings.json
Dewsk: Adding Tommy and Coates for some thinking towards this. In general, I like the idea but would need to play more with how 'discoverable' the platform forks are for more junior folks.

acoates-ms: I saw the original tweet, thought it was an interesting idea.  Wont know for sure what I think of it until I've used it.  -- It's probably a fairly harmless experiment to try anyway.

acoates-ms (approved):

.vscode/settings.json
tido64: I have some concerns with discoverability as well, but it does look nice. Maybe we can enable `explorer.fileNesting.expand` as well?

Saadnajmi: My other concern is that file nesting looks too much like a directory, and that will confuse people into thinking there's another level in the folder hierarchy. But I guess that was always a problem with this setting ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

Saadnajmi:
I'll go ahead and merge this. We can always undo if needed.

PR #2632, opened 2023-02-16
[Proposal] Order imports using the eslint import plugin

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

We currently don't have any linting rules to enforce the order our imports are in, so attempting to add one for more organized code.

Documentation on import/order is [here](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md). The current set up for the order configuration:
- separates the imports into groups, and adds newlines between each group
- within each group, the import is sorted alphabetically based on where the import is coming from

Happy to make any changes to the order configuration (ex. what the groups should be/what order they should be in) if anyone has any opinions.

ESLint does have a [sort-imports rule](https://eslint.org/docs/latest/rules/sort-imports) but I found that the functionality wasn't sufficient for how I thought the imports should to be ordered - I think with this rule you can only sort by the name of the object that's getting imported, rather than where the object is coming from (ex. for the statement import { a } from `react`, you can only sort by "a" and not be "react". The documentation has a couple of [examples](https://eslint.org/docs/latest/rules/sort-imports#examples))

Main change is in commit 1 (updating eslintrc.json, package.json, yarn.lock). Commit 2 is all the changes to the imports from running yarn lint --fix. I was also considering adding this rule for just a single package at first, but decided to add it for all of FURN since the change just involves `running yarn lint --fix`. If this PR is too big to check in at once I can start with making this change in just one package (ex. Button).

### Verification

See commit 2 for the changes this PR makes to the imports

Example of what imports will look like if there's a violation of the rule & how to fix:

https://user-images.githubusercontent.com/78454019/219808028-dcb5f643-b570-45c1-aa7e-57c5ad51ddf2.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
"so for some files there are now some warnings"

Could we suppress these warnings for now? It would otherwise be hard to notice new instances of the rule being broken.

mischreiber:
I do really like this approach - clean imports are much easier to reason about.

Pardon my JS failings, but in C languages, import order can and frequently does have a tangible impact on code behavior. Is the same true in TypeScript, or does import order only affect code readability?

lyzhan7:
> 
> Could we suppress these warnings for now? It would otherwise be hard to notice new instances of the rule being broken.

I changed the rule from "warn" to "error" - and for some reason changing it to an error caused eslint to catch & fix a lot more files. I don't see any files that have any extra warnings/errors anymore.

>does import order only affect code readability

That's a good question, I've been looking into this. I've seen mixed answers about it, but generally I think it can matter in certain situations - but I assumed that if the CI checks passed on this PR it would be fine for FURN

Saadnajmi:
> > Could we suppress these warnings for now? It would otherwise be hard to notice new instances of the rule being broken.
> 
> I changed the rule from "warn" to "error" - and for some reason changing it to an error caused eslint to catch & fix a lot more files. I don't see any files that have any extra warnings/errors anymore.
> 
> > does import order only affect code readability
> 
> That's a good question, I've been looking into this. I've seen mixed answers about it, but generally I think it can matter in certain situations - but I assumed that if the CI checks passed on this PR it would be fine for FURN

Imports can have side effects, something @lyzhan7 discovered while doing theming work. So yes, _theoretically_ order matters, but I think it should be OK?

Saadnajmi (approved):

rurikoaraki (approved):

rurikoaraki:
> > > Could we suppress these warnings for now? It would otherwise be hard to notice new instances of the rule being broken.
> > 
> > 
> > I changed the rule from "warn" to "error" - and for some reason changing it to an error caused eslint to catch & fix a lot more files. I don't see any files that have any extra warnings/errors anymore.
> > > does import order only affect code readability
> > 
> > 
> > That's a good question, I've been looking into this. I've seen mixed answers about it, but generally I think it can matter in certain situations - but I assumed that if the CI checks passed on this PR it would be fine for FURN
> 
> Imports can have side effects, something @lyzhan7 Lynn Zhang FTE discovered while doing theming work. So yes, _theoretically_ order matters, but I think it should be OK?

I think our E2E tests would catch the worst behaviors around things like crashes due to weird import behaviors. Based on what I've observed, importing a file will "run" it and execute each line of the file, which leads to more imports and setting up const variables and such.

lyzhan7:
Update - changed the order of the imports, I found some article that suggested ordering from standard to local, i.e imports from react/react-native, third party imports which would include imports from @fluentui-react-native, and then local modules/components.

PR #2641, opened 2023-02-20
[Bug fixes] Button-FAB/Shimmer/Ripple Color

ayush547:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Bug fixes -
- Make Button test page more accessible (Android).
- Fix Button/FAB icon being accessible (All plats).
- Rename FABCore to FAB.mobile maintaining parity with other files.
- Change rippleColor to '#D4D4D4' for visibility on dark mode and consistency (Android).
- Fix Shimmer background not applying (All plats except win32).
- Fix Shimmer.customize not working (All plats except win32).

### Verification

Accessibility works as expected.
Shimmer customize works as expected.

| Screenshot_2023-02-23-10-59-54-92_3238b23fbbdef38ccf4a465a34706bab | Screenshot_2023-02-23-10-56-56-14_3238b23fbbdef38ccf4a465a34706bab |

The background was not applied before, it is now being applied.
The first two shimmers in the 'Customized Shimmer' section are supposed to look the same. One uses customize API and the other uses props to achieve the same result.
Note - The 'Customized Shimmer' section does not render on Android tester app, images are added for reference.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
+++ b/apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
@@ -32,11 +32,11 @@ export const ButtonVariantTest: React.FunctionComponent = () => {
       <Button appearance="subtle" disabled style={commonTestStyles.vmargin}>
         Subtle Disabled
       </Button>
-      <FAB icon={iconProps} iconOnly disabled style={commonTestStyles.vmargin} />
+      <FAB icon={iconProps} iconOnly disabled style={commonTestStyles.vmargin} accessibilityLabel="FAB" />
       <FAB icon={iconProps} style={commonTestStyles.vmargin} showContent={showFABText} onClick={flipFABcontent}>
         Click Me!
       </FAB>
-      <FAB appearance="subtle" iconOnly disabled icon={iconProps} style={commonTestStyles.vmargin} />
+      <FAB appearance="subtle" iconOnly disabled icon={iconProps} style={commonTestStyles.vmargin} accessibilityLabel="FAB" />
       <FAB appearance="subtle" icon={iconProps} style={commonTestStyles.vmargin} showContent={showFABText} onClick={flipFABcontent}>
         Click Me!
       </FAB>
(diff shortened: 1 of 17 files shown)


apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
```diff
@@ -32,11 +32,11 @@ export const ButtonVariantTest: React.FunctionComponent = () => {
       <Button appearance="subtle" disabled style={commonTestStyles.vmargin}>
         Subtle Disabled
       </Button>
-      <FAB icon={iconProps} iconOnly disabled style={commonTestStyles.vmargin} />
+      <FAB icon={iconProps} iconOnly disabled style={commonTestStyles.vmargin} accessibilityLabel="FAB" />
```
rohanprasadofficial, line 35: is this related to a11y issue or just E2E fix ? 
I think we should add it in component not only on tester app, can you give bit context related to this ?

packages/experimental/Shimmer/src/Shimmer.styling.ts
```diff
@@ -6,11 +6,13 @@ import { defaultShimmerTokens } from './ShimmerTokens';
 
 export const stylingSettings: UseStylingOptions<ShimmerProps, ShimmerSlotProps, ShimmerTokens> = {
   tokens: [defaultShimmerTokens, shimmerName],
+  tokensThatAreAlsoProps: 'all',
   slotProps: {
     root: buildProps(
       (tokens: ShimmerTokens) => ({
         accessibilityRole: 'progressbar',
         accessible: true,
+        style: { overflow: 'hidden', backgroundColor: tokens.backgroundColor },
```
rohanprasadofficial, line 15: Seems Shimmer has only one slot as 'root'. 
Shouldn't we add this directly token files would be good idea ?  any specific reason to add it here 🤔

apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
ayush547: A11y issue, this isn't the issue with other components on the page since they either have a text and that is picked up as the label

ankraj12 (approved):
Could you add behaviour before and after for Shimmer color and angle changes?

ayush547:
> Could you add behaviour before and after for Shimmer color and angle changes?

Updated description

PR #2644, opened 2023-02-21
[Switch] Font truncated a11y issue fix

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

On font size increase the label text was getting truncated. 
In this examples the fourth example disabled default unchecked switch.  "FalseTest" was getting hidden inside screen. 

### Verification

Visual Verification after font size increase. 

| MicrosoftTeams-image (26)| MicrosoftTeams-image (25) |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Switch/src/Switch.styling.ts
+++ b/packages/experimental/Switch/src/Switch.styling.ts
@@ -20,6 +20,8 @@ export const switchStates: (keyof SwitchTokens)[] = [
   'disabled',
 ];
 
+const isMobile = Platform.OS === 'android' || Platform.OS === 'ios';
+
 export const stylingSettings: UseStylingOptions<SwitchProps, SwitchSlotProps, SwitchTokens> = {
   tokens: [defaultSwitchTokens, switchName],
   states: switchStates,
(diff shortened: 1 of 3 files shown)


packages/experimental/Switch/src/Switch.styling.ts
```diff
@@ -103,6 +105,7 @@ export const stylingSettings: UseStylingOptions<SwitchProps, SwitchSlotProps, Sw
       (tokens: SwitchTokens, theme: Theme) => ({
         style: {
           color: tokens.color,
+          ...(isMobile && { flexShrink: 1 }),
```
ankraj12, line 108: I am not sure if we should do this as part of Switch component. What would be the behaviour for a text which say spans 4 lines - do we expands the component to take vertical space to render all the text? Or truncate after a specific limit. 
Can user not instead customise/style the properties for the label if it needs to be too long?

ankraj12 (commented):
We should check the guidance from PM/ design if this is coded as a component behaviour.

rohanprasadofficial:
> We should check the guidance from PM/ design if this is coded as a component behaviour.

@ankraj12 , We had PM confirmation for label to be shown as all spanning across lines and not trucating it and this PR complies the same.

ankraj12 (approved):

PR #2649, opened 2023-02-22
[Shadow] Only add style props that were actually defined when creating the Shadow views

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

See commits for breakdown of changes
- Update the shadow component to only set props that are defined
   - There was some discussion in this previous [PR](https://github.com/microsoft/fluentui-react-native/pull/2117#discussion_r971135416) that introduced this change about whether or not we should only set props that are defined or just pass everything through; ultimately we decided to go with the latter since we didn't think it would make a difference. WIth some of the new test scenarios (ex. shadow on iOS button) it looks like there is a functional difference that this makes, so updating the Shadow component to go with the former approach.
- This fixes issue where shadow on an iOS Button looked strange (see screenshots below)
  - I added some jest tests + FluentTester tests to show this issue has been fixed
- This also cleans up our jest snapshots by removing all the undefined props we were adding

### Verification

| | Before any changes | After updating the test page (to show the bug) | After making the fix in Shadow |
| - | - | - | - |
| iOS | ios_before | ios_after_adding_button_to_testpafe | ios_after |
| macOS | macOS_before | macos_after_adding_button_to_testpage | macOS_after |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
```diff
@@ -14,25 +18,31 @@ const CustomFABShadow64 = FAB.customize({
...
+          </FAB>
+          <CustomFABShadow64 icon={iconProps} style={styles.marginBetweenComponentsWithShadow}>
+            Custom FAB with shadow64
+          </CustomFABShadow64>
+          <CustomFABNoShadow icon={iconProps} style={styles.marginBetweenComponentsWithShadow}>
+            Custom FAB with no shadow
+          </CustomFABNoShadow>
+        </View>
+      )}
+      {Platform.OS !== 'android' && (
```
Saadnajmi, line 36: Could we make the platform checks a little more obvious? Maybe pull out to separate variables with names like `supportsFABShadow` or something? It's not clear to me what's special about Android and iOS to me in this test.

packages/experimental/Shadow/src/Shadow.tsx
```diff
@@ -92,51 +92,53 @@ function getStylePropsForShadowViewsWorker(childStyleProps: ViewStyle = {}, shad
...
-        paddingTop,
-        paddingVertical,
-
-        alignItems,
-
-        flexWrap,
-        flexDirection,
+        // Only add the prop if it is defined, otherwise every prop will be added and be set to
+        // undefined, which can cause unexpected behaviour with some of the styles
+        ...(borderBottomWidth && { borderBottomWidth }),
```
Saadnajmi, line 97: I wonder there's a better way to do this, but I am not sure 😅. I'm not sure of the perf implications of so many spread operators. Maybe there's something like `mergeStyle` that's useful in `@fluentui-react-native/framework`?

Not opposed though, I think this is a good overall change :)

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/Shadow/ShadowButtonTestSection.tsx
lyzhan7: Just updated

packages/experimental/Shadow/src/Shadow.tsx
lyzhan7: Ended up taking out all the undefined props after creating the shadow inner/outer style objects - if you had a different idea in mind let me know.

PR #2661, opened 2023-02-24
Only open menu when shouldFocusOnContainer is defined

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The Menu would put focus on the first item when opened via hover, which is not aligned with win32 Menu behavior.

The issue was that the popover would get initialized with shouldFocusOnContainer false and then changed to true after first render, but by then focus would already be placed on the first item.
In order to prevent this behavior, this change prevents the popover from rendering until shouldFocusOnContainer is a defined value of true or false, and changed shouldFocusOnContainer to be undefined while the popover is not opened. This makes sure that shouldFocusOnContainer is the correct value when the popover is initialized and focus is put on the correct component.

### Verification

Verified via tester. @lenahong also tested.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/Menu/Menu.tsx
+++ b/packages/components/Menu/src/Menu/Menu.tsx
@@ -28,6 +28,8 @@ export const Menu = stagedComponent((props: MenuProps) => {
     return (
       <MenuProvider value={contextValue}>
         {menuTrigger}
+        {/* GH#2661: Make sure that shouldFocusOnContainer is defined before initializing
+            the popover so that focus is put in the correct place */}
         {state.open && state.shouldFocusOnContainer !== undefined && menuPopover}
       </MenuProvider>
     );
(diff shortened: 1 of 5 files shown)


lenahong (approved):

chiuam (approved):

packages/components/Menu/src/Menu/Menu.tsx
```diff
@@ -28,7 +28,7 @@ export const Menu = stagedComponent((props: MenuProps) => {
     return (
       <MenuProvider value={contextValue}>
         {menuTrigger}
-        {state.open && menuPopover}
+        {state.open && state.shouldFocusOnContainer !== undefined && menuPopover}
```
Saadnajmi, line 31: Nit: can we add a comment, as I feel like this is otherwise not clear?

Saadnajmi (approved):

packages/components/Menu/src/Menu/Menu.tsx
rurikoaraki: I added a comment, does it help?

PR #2662, opened 2023-02-24
Remove remaining references of experimental-button to stop warnings

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Noticed that the tester always triggers the same warnings, which makes it harder to notice any new warnings, so trying to get rid of the ones that consistently show up.

This PR removes this warning: 

`The @fluentui-react-native/experimental-button package is deprecated. The contents of this package have been moved to @fluentui-react-native/button. If you need to use the Button component from this package, please use ButtonV1 from @fluentui-react-native/button` by replacing the remaining uses of experimental-button with button gets rid of one of the warnings.

### Verification

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/package.json
+++ b/apps/fluent-tester/package.json
@@ -48,7 +48,6 @@
     "@fluentui-react-native/experimental-activity-indicator": ">=0.7.20 <1.0.0",
     "@fluentui-react-native/experimental-appearance-additions": "^0.3.4",
     "@fluentui-react-native/experimental-avatar": ">=0.17.18 <1.0.0",
-    "@fluentui-react-native/experimental-button": "0.16.92",
     "@fluentui-react-native/experimental-checkbox": "0.13.40",
     "@fluentui-react-native/experimental-expander": "0.5.20",
     "@fluentui-react-native/experimental-menu-button": ">=0.6.33 <1.0.0",
--- a/apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
+++ b/apps/fluent-tester/src/TestComponents/Button/ButtonVariantTestSection.mobile.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { View } from 'react-native';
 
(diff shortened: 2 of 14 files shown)


rurikoaraki:
The use of Button from the experimental-button package was intentional, since I wanted to make sure that the Button would still render if pulled from experimental-button while it's still in deprecation mode. It's to make sure we don't break the experimental-button package. If you could replace that verification with something else, then I'm ok with this change

lyzhan7:
> 

I see okay, thanks for explaining. I'll think about it but can also just close the PR. Is there a planned timeline for removing experimental-button/no longer supporting it in deprecation mode? Looks like that console warning was added a year ago

rurikoaraki:
> > The use of Button from the experimental-button package was intentional, since I wanted to make sure that the Button would still render if pulled from experimental-button while it's still in deprecation mode. It's to make sure we don't break the experimental-button package while making changes to Button in the main package. If you could replace that verification with something else, then I'm ok with this change
> 
> I see okay, thanks for explaining. I'll think about it but can also just close the PR. Is there a planned timeline for removing experimental-button/no longer supporting it in deprecation mode? Looks like that console warning was added a year ago
> 
> I was also planning on making the same changes for experimental-checkbox/experimental-text, I'm assuming the same reasoning for why we haven't fully removed those from FURN apply to those components as well?

Yeah, it's because removing a package would be a breaking change. We would need to give ample warning to clients before doing that so that anyone who's been depending on those packages has time to transition to the new packages.

I was thinking that the timing would be to do it at the same time as removing the v0 component, which we can't do until all platforms have finished migrating the component to v1. Button still needs work from a few platforms before we can claim that moving to v1 would be stable for clients. We could remove the experimental package before removing v0 though, since usually once we move the code for one platform out of experimental it stays in the main package anyway.

PR #2664, opened 2023-02-27
[Badge] Add warning in tester if shadow is applied to ghost/outline badge variants

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

During a bug bash we noticed on the macOS tester that some shadows on the badge tester page didn't look right/was getting applied to the text inside:

image

After further investigation, shadows should not be applied to certain badge variants. In the Figma, the shadow property is optional for badge, except for the outline/ghost variants where there should be no shadow. Since shadows get added to badge through customize, it's up to the user to enforce that a shadow doesn't get added to the outline/ghost variants. This PR adds a warning to the tester that shadows on the outline/ghost variants are not supported.

Notes:
- also added a jest test for shadow on a valid badge variant
- removed some code in Badge that seems to try to check the variant to see if a shadow was added or not, but doesn't work

### Verification

Before

https://user-images.githubusercontent.com/78454019/221663971-7335983f-a009-4a28-bd0b-8c0087cab515.mov

After

https://user-images.githubusercontent.com/78454019/225386876-b23b289b-7a9e-4a10-8526-6652d1d0c604.mov

Notes:
- The shadow component was added to Badge initially in [this PR](https://github.com/microsoft/fluentui-react-native/pull/2149)

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Badge/src/Badge.tsx
```diff
@@ -23,8 +23,7 @@ export const badgeLookup = (layer: string, userProps: BadgeProps): boolean => {
     layer === userProps['shape'] ||
     (!userProps['shape'] && layer === 'circular') ||
     layer === userProps['badgeColor'] ||
-    (I18nManager.isRTL && layer === 'rtl') ||
-    (layer === 'shadowToken' && (!userProps.appearance || userProps.appearance === 'filled' || userProps.appearance === 'tint'))
+    (I18nManager.isRTL && layer === 'rtl')
```
lyzhan7, line 26: So it seemed like this code was supposed to be checking the variant to see if a shadow was added or not, but this didn't seem to work. I didn't see any difference removing it so i just removed it, but if anyone has any insight what this code should be doing/how to fix it let me know

packages/components/Badge/src/Badge.tsx
```diff
@@ -44,27 +43,33 @@ export const Badge = compose<BadgeType>({
...
+              ),
+            )}
+          {icon && showIcon && iconPosition === 'after' && <Slots.icon accessible={false} {...iconProps} />}
+        </Slots.root>
       );
+
+      if (showOptionalShadow) {
+        return <Slots.shadow>{badgeWithoutShadow}</Slots.shadow>;
+      } else {
+        return badgeWithoutShadow;
```
Saadnajmi, line 71: `Shadow.tsx` will only render the children if no shadow token is applied
```
  if (!props.shadowToken) {
    return <>{children}</>;
  }
```
Could we do something where we only pass the shadow Token if our appearance is `filled` or `tint` so that we can just take advantage of the existing optimization in Shadow?

Saadnajmi (approved):
Would be nice if we can verify win32 or iOS isn't broken too.

packages/components/Badge/src/Badge.tsx
rurikoaraki: Yeah I'm guessing that's how the code was meant to work based on what was written - always use the shadow slot but only apply shadow tokens if the appearance is one of the correct ones.

lyzhan7: Hmm so I had briefly looked into this but wasn't sure where the logic to "only pass the shadow Token if our appearance is filled or tint" would go.

In Notification we could define the shadow tokens in NotificationTokens, because by default Shadows should always be applied to certain Notification variants. For badge the shadow is only optional for certain badge variants (by default it's not applied), so there aren't any shadows defined in BadgeTokens. The way to add a shadow in Badge is by using Badge.customize

Maybe if a user really wants to use customize to add a shadow to a badge variant that shouldn't have a shadow, we could just let them? And maybe just print out a warning.

lyzhan7: Talked offline - decided we should leave it up the user to not add a shadow to certain variants. I've reverted most of the changes in the badge component, and added a warning in the tester.

Saadnajmi (approved):

PR #2665, opened 2023-02-27
[Menu] Override onHoverOut to call onBlur

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change addresses a bug on Mac where keyboard focus is lost when mouse hover moves away from menu, which results in broken keyboard navigation. Calling .blur() on menu item fully removes keyboard focus which causes FocusZone to no-op, so let's override onBlur from Pressable that disables focus instead.

Note: This behavior aligns with Contextual Menu, though ideally we'd want focus to be reset to the first item.
### Verification

Before:

https://user-images.githubusercontent.com/67026167/222189733-b840f191-117d-4458-b50f-a81ccf336e9b.mov

After:

https://user-images.githubusercontent.com/67026167/222189970-ed44a92a-b725-4183-953c-7ac849b8da50.mov

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/components/Menu/src/MenuItem/useMenuItem.ts
@@ -35,7 +35,7 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
         ((isRtl && e.nativeEvent.key === 'ArrowLeft') || (!isRtl && e.nativeEvent.key === 'ArrowRight'));
 
       if (!disabled && (!isArrowKey || isArrowOpen)) {
-        onBlur();
+        onBlur(e);
         onClick?.(e);
       }
 
--- a/packages/components/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/components/Menu/src/MenuItem/useMenuItem.ts
@@ -52,6 +52,20 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
   );
 
   const pressable = usePressableState({ ...rest, onPress: onInvoke });
+  const { onBlur, onHoverOut, ...restPressableProps } = pressable.props;
+  const onHoverOutModified = React.useCallback(
+    (e) => {
+      onBlur(e);
+      onHoverOut && onHoverOut(e);
+    },
+    [onBlur, onHoverOut],
+  );
+
+  const pressablePropsModified = {
+    onBlur,
+    onHoverOut: onHoverOutModified,
+    ...restPressableProps,
+  };
   const itemRef = useViewCommandFocus(componentRef);
   const keys = disabled ? [] : isSubmenu ? submenuTriggerKeys : triggerKeys;
 
@@ -62,7 +76,7 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
 
   return {
     props: {
-      ...pressable.props,
+      ...pressablePropsModified,
       accessible: true,
       accessibilityRole: 'menuitem',
       onAccessibilityTap: props.onAccessibilityTap || onInvoke,
@@ -100,8 +114,6 @@ export const useHoverFocusEffect = (hovered: boolean, componentRef: React.Mutabl
   React.useLayoutEffect(() => {
     if (hovered) {
       componentRef?.current?.focus();
-    } else {
-      componentRef?.current?.blur();
     }
   }, [hovered, componentRef]);
 };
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -52,6 +52,23 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
...
+      onHoverOut && onHoverOut(e);
+    },
+    [onBlur, onHoverOut],
+  );
+
+  const pressablePropsModified = {
+    onBlur,
+    onHoverOut: onHoverOutModified,
+    ...restPressableProps,
+  };
```
Saadnajmi, line 71: Could we do this before `usePressableState` do that we don't have to modify pressable.props?
I did that in ContextualMenu because we had to do a lot of shennanigans of wanting both `onMouseEnter` and `onHoverIn`, but it doesn't seem like that applies here.

chiuam: Is there a way where I can override `onBlur` and `onHoverOut` without modifying pressable.props?

Saadnajmi: Passing on our / onHoverOut into usePressableState doesn't work?

Saadnajmi: Passing on our / onHoverOut into usePressableState doesn't work?

chiuam: I don't think so since onBlur is called inside the onHoverOut override?

Saadnajmi:
Could you also attach before and after videos? I think that would help

chiuam:
> Could you also attach before and after videos? I think that would help

Updated PR description with before/after videos, hope that helps

Saadnajmi (commented):
Talked offline, and leaving a comment with my thoughts here:

By calling the `onBlur` returned by `usePressableState` manually when blur hasn't occurred, we're "tricking" React Native into thinking the menu item lost focus, while instead it still has it. This is visible in the "after" video by the fact that ArrowDown doesn't move focus to the first menu item, but instead the menu item below the last focused menu item. This is true in ContextualMenu too, in fact it seems I wrote a lengthy explanation about it:

```
/**
 * On Desktop, focus gets moved to the root of the menu, so hovering off the menu does not automatically call onBlur as we expect it to.
 * onMouseLeave is overridden to explicitly call onBlur to simulate removing focus
 * To achieve this, we override the onMouseLeave handler returned by useAsPressable, and replace it with our own. Inside our own
 * onMouseLeave handler, we call useAsPressable's onMouseLEave handler,
 */
const { onBlur, onMouseLeave, ...restPressableProps } = pressable.props;
const onMouseLeaveModified = React.useCallback(
  (e) => {
    onBlur(e);
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore IViewWin32 doesn't have the event as an argument, while macOS does
    onMouseLeave && onMouseLeave(e);
  },
  [onBlur, onMouseLeave],
);
const pressablePropsModified = {
  onBlur: onBlur,
  onMouseLeave: onMouseLeaveModified,
  ...restPressableProps,
};
```

But I'm not sure I like the behavior better, over just leaving the focused menu item highlighted, for two reasons:

1) I don't like "tricking" React Native in case we run into weird bugs because of that. It seems we have not, given ContextualMenu is fine
2) When the mouse hovers out and nothing is highlighted, the user might be expecting that focus is on the menu, and ArrowDown moves it to the first menu item. That matches what NSMenu does. Instead, it's somewhere else. I understand that ContextualMenu has the same behavior, but I don't think I like it there as much either.

The "ideal" experience is that `onHoverOut` moves focus back to the Menu/MenuList, but that's a more involved change that I think ran into issues. Since #2669, the manual call to `.blur()` no-ops and now the menu item stays highlighted after onHoverOut. I think I personally like that better as an experience, rather than following CM and manually calling `onBlur`.

This is an edge case and not a hard disagreement, so I'm totally OK if someone else approves it and this merges.

mischreiber (approved):

mischreiber (commented):
Currently learning about the enter/space behavior while nothing is selected

chiuam:
Talked with @mischreiber and @Saadnajmi offline, closing this PR for now

PR #2668, opened 2023-02-28
Adding theme recipes example in tester

ankraj12:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Added a test flow to demonstrate custom theme colors and usage in android.
### Verification

darkColors
themelightColors

Tester in android tester.
### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
```diff
@@ -0,0 +1,67 @@
...
+  },
+];
+
+export const ThemeTest: React.FunctionComponent = () => {
+  const status: PlatformStatus = {
+    win32Status: 'Beta',
+    uwpStatus: 'Experimental',
+    iosStatus: 'Experimental',
+    macosStatus: 'Experimental',
+    androidStatus: 'Backlog',
```
rurikoaraki, line 60: Probably should update androidStatus here

packages/theming/theme-types/src/Color.types.ts
```diff
@@ -19,7 +19,9 @@ export type ThemeColorDefinition = Palette & {
   bodyText: ColorValue;
   subText: ColorValue;
   disabledText: ColorValue;
-  [key: string]: ColorValue;
+
+  // There is this alias value existing to take custom color names.
```
rurikoaraki, line 23: Sorry, this comment doesn't make much sense to me... could you try explaining what you're trying to convey?

apps/fluent-tester/src/TestComponents/Theme/ThemeTest.android.tsx
```diff
@@ -0,0 +1,67 @@
...
+    },
+    rootView: {
+      backgroundColor: theme.colors.brandBackground,
+    },
+  };
+});
+
+const Panel: React.FunctionComponent = () => {
+  const theme = useFluentTheme();
+  const theme2 = useTheme();
```
rurikoaraki, line 24: useFluentTheme should have everything on useTheme. the difference useFluentTheme has is that it returns a fallback theme if the ThemeProvider isn't used.

ankraj12: Ok, thanks for clarifying. I was initially thinking to demonstrate usage of both as doing the same thing - but I think the guidance would be to use `useFluentTheme`() - so kept only that.

ankraj12: I am not also aware if there is another place that has a similar example on `ThemeRecipes` usage - if there is another one that does similar/ I should extend this example to other endpoints, let me know and I can make the changes.

rurikoaraki: I usually point people to the win32 theme (the impl of createOfficeTheme) as an example of using ThemeRecipes, but other than that I don't think we have an example of it.

rurikoaraki (approved):

PR #2669, opened 2023-02-28
Fix various issues with checkbox focus on windows

acoates-ms:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

enableFocusRing defaults to true on non-win32 platforms, so it needs to be set to false to prevent multiple focus visuals showing in CheckBox.

Added an additional step in how to run fluent-tester for windows.

Modified useViewCommandFocus to prevent overriding the implementation of ref.focus.  -- The current implementation does not work on windows.

### Verification

in progress....

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/docs/windows.md
+++ b/apps/fluent-tester/docs/windows.md
@@ -16,7 +16,8 @@
 ## Running the App
 
 1. Run `yarn` from the root of the repo
-2. From this directory, simply `yarn windows`
+2. From this directory, run `yarn install-windows-test-app`
+3. From this directory, simply `yarn windows`
 
 ## Debugging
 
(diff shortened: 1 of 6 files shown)


rurikoaraki (approved):

Saadnajmi:
I think I'll want to validate this change on macOS, just to make sure that I was correct that we don't need the `useViewCommandFocus` hook.

chiuam:
> I think I'll want to validate this change on macOS, just to make sure that I was correct that we don't need the `useViewCommandFocus` hook.

Thanks, I remember making a change to `useViewCommandFocus` to unblock a focus issue on Mac. I'm going to validate it against Menu

Saadnajmi (approved):
Did some quick macOS testing and seems fine.. fingers crossed!

acoates-ms:
Thanks @chium, let me know when you've done your validation.

chiuam (approved):
Did a quick a test on keyboard navigation, seems like the call to .blur() no-ops on menu item 

https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Menu/src/MenuItem/useMenuItem.ts#L39

Though I think the proper way of doing it is disabling focus, not removing it which is what .blur() does - similar to this change #2665. I'm okay with checking in this change with a followup fix to address the fallout.

PR #2671, opened 2023-03-01
Make 'Escape' stop event propagation in `MenuPopover` key handlers.

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

For 'Escape' keypresses in an opened MenuPopover, the keypress event should not propagate to parent components on bubbling up. Currently, parents to Menus will receive keyboard events if a user presses 'Escape' in a MenuPopover. 

This PR fixes this by adding 'Escape' to a list of keys that stop keyboard event propagation in MenuPopover to parent components.

### Verification

Video of corrected behavior. I first cycle through escape on the menutrigger, which calls the parent view's keydown handler, changing the border color. I then open the popover and press escape, while not cycling through the border colors because the event is not propagated to the parent view.

https://user-images.githubusercontent.com/15683103/222019176-dd05332a-4013-4fa7-8eb4-e0900a101076.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
+++ b/packages/components/Menu/src/MenuPopover/useMenuPopover.ts
@@ -67,6 +67,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
         case 'Tab':
         case 'Home':
         case 'End':
+        case 'Escape':
           e.stopPropagation();
       }
     },
@@ -85,6 +86,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
         case 'Tab':
         case 'Home':
         case 'End':
+        case 'Escape':
           e.stopPropagation();
       }
     },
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):
Looks good to me, though would still like quick verification on macOS. We don't send that keyboard event to JS currently, so there shouldn't be any effect.

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -67,6 +67,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
         case 'Tab':
         case 'Home':
         case 'End':
+        case 'Escape':
```
PPatBoyd, line 70: Can we factor out these shared values?

chiuam (approved):
Validated on macOS, looks good!

PR #2672, opened 2023-03-01
[Menu] Add icon support to MenuItems

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Not fully done, but pushing this out to start review.

Adding icon support to MenuItems. This change:

1. Adds "hasIcons" as a prop to Menu. This controls whether a space should be added to items without icons for alignment purposes
2. Pipes that prop down to MenuItems
3. Adds imgIcon, fontOrSvgIcon, and iconPlaceholder slots to MenuItem and MenuItemCheckbox.
4. Adds color and size as props to icon so that these props can be passed along to icons through compose without needing to know the kind of icon it is.
5. Adds default tokens
6. Adds a test page.

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
Nice :) 
image

PPatBoyd:
Should the menu items text align (left) regardless of icon size, or should it be left to usage?

Do we handle scaling automatically for display scaling and text scaling, or is manual scaling required of the user for one or both?

rurikoaraki:
> Should the menu items text align (left) regardless of icon size, or should it be left to usage?
> 
> Do we handle scaling automatically for display scaling and text scaling, or is manual scaling required of the user for one or both?

1. I have a bit of a limitation here on styling for the icon from the MenuItem due to the fact that svg and font icons are bundled into the same icon component. At the moment font icons don't default to being centered and it's pretty easy to change that via the font icon props so I'm leaving it to usage (see test), but I can ask PM about it.

2. Display scaling is handled automatically but text scaling only affects text automatically. I can change the View's size to be minHeight/Width to handle that case. Thanks for catching

Saadnajmi:
(Because I still had it open) yay it looks better!
image

PPatBoyd:
> (Because I still had it open) yay it looks better! image

JW for both platforms, is the checkbox expected in option 5 implemented or TBD? I'm a little surprised it doesn't have a visual indicator -- it makes it look misaligned at a glance

Saadnajmi:
> 

Theres's a bug where there's mismatched font size between MenuItem and MenuItemCheckbox on macOS, we're working on that. I think there's no checkmark because it isn't checked. If I click and re-open, it'll have the checkmark.

rurikoaraki:
> > (Because I still had it open) yay it looks better! image
> 
> JW for both platforms, is the checkbox expected in option 5 implemented or TBD? I'm a little surprised it doesn't have a visual indicator -- it makes it look misaligned at a glance

Yeah as @saadnajmi said, there's a blank there because the option isn't checked. If I check it it looks like this:

image

There's another prop that will add spacing in place of a checkmark if the item isn't selectable, I've locally checked that that case works as well.

Saadnajmi:
Nit: Can the default menu in the test page (the first menu you see) have a superset of all these options? Like have it have MenuItem, MenuItemCheckbox, MenuDivider, MenuItem with icons? Or at least one menu item with an icon?

Saadnajmi:
> > > (Because I still had it open) yay it looks better! image
> > 
> > 
> > JW for both platforms, is the checkbox expected in option 5 implemented or TBD? I'm a little surprised it doesn't have a visual indicator -- it makes it look misaligned at a glance
> 
> Yeah as @Saadnajmi said, there's a blank there because the option isn't checked. If I check it it looks like this:
> 
> image
> 
> There's another prop that will add spacing in place of a checkmark if the item isn't selectable, here's a screenshot of both props set: image

Does the client have to know to add that prop for the spacing? I guess that's how web does it and I'm not sure there's a better way, but I feel like I'd rather the default have the right spacing from the get go. But if we're aligned with web, then that's best.

rurikoaraki:
> Nit: Can the default menu in the test page (the first menu you see) have a superset of all these options? Like have it have MenuItem, MenuItemCheckbox, MenuDivider, MenuItem with icons? Or at least one menu item with an icon?

Hmm I was kind of thinking the default menu test case would represent "What things would look like if you didn't add anything" (although i guess disabled is in there). Maybe it would make sense to redesign the Menu test page since it's getting pretty long? Like have a bunch of switches that would modify the first example? That would be its own PR.

rurikoaraki:
> > > > (Because I still had it open) yay it looks better! image
> > > 
> > > 
> > > JW for both platforms, is the checkbox expected in option 5 implemented or TBD? I'm a little surprised it doesn't have a visual indicator -- it makes it look misaligned at a glance
> > 
> > 
> > Yeah as @Saadnajmi said, there's a blank there because the option isn't checked. If I check it it looks like this:
> > image
> > There's another prop that will add spacing in place of a checkmark if the item isn't selectable, here's a screenshot of both props set: image
> 
> Does the client have to know to add that prop for the spacing? I guess that's how web does it and I'm not sure there's a better way, but I feel like I'd rather the default have the right spacing from the get go. But if we're aligned with web, then that's best.

Yeah, the client has to pass that prop in to add the spacing. I'll double check if web has some special handling of the icon/checkmark case.

Saadnajmi:
> 

That's fine. It's a minor nit, and not blocking.

Saadnajmi:
OK, another screenshot with `hasCheckmarks` as well, plus me manually fixing the MenuItemCheckbox styling:

image

Note, I had to remove the `maxWidth={160}` for MenuPopover in the test because it was clipping the menu items. Do we need that in the test?

Saadnajmi (approved):

PR #2673, opened 2023-03-01
[Presence Badge] Add support for different svgs based on size

ayush547:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Android/iOS require different svgs for different presence badge sizes. This PR adds support for the same.
Added new types where relevant.

image
This image shows how the "available" svg for small (10x10) looks when resized and overlapped with the large (16x16) "available" svg.

### Verification

| image | image | 

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


PPatBoyd:
Can you add screenshots for before/after, and share more on the need for different SVGs instead of scaling the SVGs?

As an extension of scaling SVGs, semantic icons (providing information to the user) should scale with accessibility text scaling; I'm not sure of the specifics addressing this for Android, but for context a link to [iOS dev docs on the topic](https://developer.apple.com/design/human-interface-guidelines/foundations/accessibility#:~:text=Increase%20the%20size%20of%20meaningful%20interface%20icons%20as%20font%20size%20increases.%20If%20you%20use%20interface%20icons%20to%20communicate%20important%20information%2C%20make%20sure%20they%20are%20easy%20to%20view%20at%20larger%20font%20sizes%2C%20too.%20When%20you%20use%20SF%20Symbols%2C%20you%20get%20icons%20that%20scale%20automatically%20with%20Dynamic%20Type%20size%20changes.).

ayush547:
> Can you add screenshots for before/after, and share more on the need for different SVGs instead of scaling the SVGs?
> 
> As an extension of scaling SVGs, semantic icons (providing information to the user) should scale with accessibility text scaling; I'm not sure of the specifics addressing this for Android, but for context a link to [iOS dev docs on the topic](https://developer.apple.com/design/human-interface-guidelines/foundations/accessibility#:~:text=Increase%20the%20size%20of%20meaningful%20interface%20icons%20as%20font%20size%20increases.%20If%20you%20use%20interface%20icons%20to%20communicate%20important%20information%2C%20make%20sure%20they%20are%20easy%20to%20view%20at%20larger%20font%20sizes%2C%20too.%20When%20you%20use%20SF%20Symbols%2C%20you%20get%20icons%20that%20scale%20automatically%20with%20Dynamic%20Type%20size%20changes.). Mentioning to raise awareness; I wouldn't block this PR on text scaling but I am interested on if multiple vector graphics are needed for scaling purposes.

For mobile (both android and iOS) the designer guidelines require us to have different svgs for different presence badge sizes. 

Multiple vector graphics are NOT required for scaling and that is a benefit of using svgs! In this case the svgs themselves are different hence this PR (adding an image showing the 10x10 available svg scaled up to the 16x16 size and overlapped with the actual 16x16 available svg to show difference)

packages/components/Badge/src/PresenceBadge/stylingUtils.mobile.ts
```diff
@@ -0,0 +1,95 @@
...
+        doNotDisturbOutOfOffice:
+          'M4.99769 0C2.23754 0 0 2.23754 0 4.99769C0 7.75784 2.23754 9.99538 4.99769 9.99538C7.75784 9.99538 9.99538 7.75784 9.99538 4.99769C9.99538 2.23754 7.75784 0 4.99769 0ZM1 4.99769C1 2.78983 2.78983 1 4.99769 1C7.20555 1 8.99538 2.78983 8.99538 4.99769C8.99538 7.20555 7.20555 8.99538 4.99769 8.99538C2.78983 8.99538 1 7.20555 1 4.99769ZM3 5C3 4.72386 3.22386 4.5 3.5 4.5H6.5C6.77614 4.5 7 4.72386 7 5C7 5.27614 6.77614 5.5 6.5 5.5H3.5C3.22386 5.5 3 5.27614 3 5Z',
+      };
+  }
+}
+
+export function getIconPath(status: PresenceBadgeStatus, isOutOfOffice: boolean, size: BadgeSize): string {
+  console.log(size);
```
rohanprasadofficial, line 74: remove console.log

packages/components/Badge/src/PresenceBadge/stylingUtils.ts
```diff
@@ -13,3 +16,26 @@ export const presenceIconPaths = {
   availableOutOfOffice:
     'M11.7 6.7a1 1 0 00-1.4-1.4L7 8.58l-1.3-1.3a1 1 0 00-1.4 1.42l2 2a1 1 0 001.4 0l4-4zM0 8a8 8 0 1116 0A8 8 0 010 8zm8-6a6 6 0 100 12A6 6 0 008 2z',
 };
+
+// Mobile requires different svgs based on BadgeSize, not used for other platforms.
+export function getIconPath(status: PresenceBadgeStatus, isOutOfOffice: boolean, _size?: BadgeSize): string {
```
rohanprasadofficial, line 21: is this function getIconPath() is identical to the getIconPath present in stylingUtils.mobile.ts ?

packages/components/Badge/src/PresenceBadge/stylingUtils.mobile.ts
ayush547: Thank you for catching this, removed.

ayush547: Thanks for catching this, removed

packages/components/Badge/src/PresenceBadge/stylingUtils.ts
ayush547: No, this function does not vary the icon according to size!

rurikoaraki:
I would say that it's worth being aware of the tradeoff here - adding multiple SVGs will increase the size of the component quite a bit compared to scaling it but having it be off of design guidelines. I would make sure that the cost of the increase of size is worth it to be pixel-perfect.

packages/components/Badge/src/PresenceBadge/stylingUtils.ts
rohanprasadofficial: No i mean seems the logic inside can be resued.

ayush547:
> It's worth being aware of the tradeoff here - adding multiple SVGs will increase the size of the component quite a bit compared to scaling one svg but having it be off of design guidelines. I would make sure that the cost of the increase of size is worth it to be pixel-perfect, unless there is an additional reason to have there be multiple svgs.
> 
> ETA: Web is able to have multiple SVGs as they are able to lazy load assets, which is something native cannot do (we must bundle all the assets that can be used regardless of whether they're used)

Yes, I did the size cost analysis and based on what I saw from bundle size, Presence Badge as a component goes from about 5.5KB to about 7.9KB, which is not much in terms of KB but is a whopping ~50% increase (for only presence badge NOT entire avatar)!
Had a discussion with design, the instructions we received are that we need to have the different svgs, same one being resized is not good enough!

packages/components/Badge/src/PresenceBadge/stylingUtils.ts
```diff
@@ -13,3 +16,26 @@ export const presenceIconPaths = {
   availableOutOfOffice:
     'M11.7 6.7a1 1 0 00-1.4-1.4L7 8.58l-1.3-1.3a1 1 0 00-1.4 1.42l2 2a1 1 0 001.4 0l4-4zM0 8a8 8 0 1116 0A8 8 0 010 8zm8-6a6 6 0 100 12A6 6 0 008 2z',
 };
+
+// Mobile requires different svgs based on BadgeSize, not used for other platforms.
```
ankraj12, line 20: nit: _size not used

ankraj12 (approved):
Approving per your note on another thread that ascertains the need for adding up the multiple images.

packages/components/Badge/src/PresenceBadge/stylingUtils.ts
ayush547: Updated

PR #2678, opened 2023-03-02
Fix min-height token value being overriden in second render stage + update default values to align with spec.

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

After taking another look at the original Divider spec, I realized that  the min-height token value for childless vertical dividers was wrong. This commit changes the height from 24 -> 20, removes references to  that in the tester, and adds documentation.

This commit also fixes logic for determining the divider minHeight. Previously, if a user set the minHeight token via customize, that value would be overriden for the defaults determined in the second render stage.

### Verification

Before:
Before

After:
After

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/experimental/Divider/SPEC.md
+++ b/packages/experimental/Divider/SPEC.md
@@ -220,7 +220,7 @@ type DividerLayoutTokens = {
   minWidth?: string | number;
   maxWidth?: string | number;
   /**
-   * For min-height, there are different default values:
+   * For minHeight, there are different default values:
    * @default 0 for horizontal dividers
    * @default 20 for vertical dividers without children
    * @default 84 for vertical divicers with children
--- a/packages/experimental/Divider/SPEC.md
+++ b/packages/experimental/Divider/SPEC.md
@@ -223,7 +223,7 @@ type DividerLayoutTokens = {
    * For minHeight, there are different default values:
    * @default 0 for horizontal dividers
    * @default 20 for vertical dividers without children
-   * @default 84 for vertical divicers with children
+   * @default 84 for vertical dividers with children
    */
   minHeight?: string | number;
   maxHeight?: string | number;
--- a/packages/experimental/Divider/SPEC.md
+++ b/packages/experimental/Divider/SPEC.md
@@ -219,6 +219,12 @@ type DividerLayoutTokens = {
    */
   minWidth?: string | number;
   maxWidth?: string | number;
+  /**
+   * For min-height, there are different default values:
+   * @default 0 for horizontal dividers
+   * @default 20 for vertical dividers without children
+   * @default 84 for vertical divicers with children
+   */
   minHeight?: string | number;
   maxHeight?: string | number;
   /**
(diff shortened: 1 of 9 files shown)


rurikoaraki:
Screenshots would be great to see for purely visual changes, since it's hard to envision them through code.

packages/experimental/Divider/SPEC.md
```diff
@@ -219,6 +219,12 @@ type DividerLayoutTokens = {
    */
   minWidth?: string | number;
   maxWidth?: string | number;
+  /**
+   * For min-height, there are different default values:
+   * @default 0 for horizontal dividers
+   * @default 20 for vertical dividers without children
+   * @default 84 for vertical divicers with children
```
PPatBoyd, line 226: \#nit s/divicers/dividers

packages/experimental/Divider/src/Divider.tsx
```diff
@@ -35,7 +35,7 @@ export const Divider = compressible<DividerProps, DividerTokens>((props: Divider
   [tokens, cache] = patchTokens(tokens, cache, {
     flexAfter: props.alignContent === 'end' ? 0 : 1,
     flexBefore: props.alignContent === 'start' ? 0 : 1,
-    minHeight: props.vertical ? globalTokens.size240 : 0,
+    minHeight: props.vertical ? globalTokens.size200 : 0,
```
PPatBoyd, line 38: \#nit can we bring the `props.vertical && hasContent` check down on line 68 and this logic together to evaluate the default minHeight in one place?

packages/experimental/Divider/SPEC.md
PPatBoyd: \#nit s/min-height/minHeight

packages/experimental/Divider/src/Divider.tsx
lawrencewin: "Yes we can" - Bob the Builder

packages/experimental/Divider/src/Divider.tsx
```diff
@@ -63,9 +62,13 @@ export const Divider = compressible<DividerProps, DividerTokens>((props: Divider
     const hasContent = textContent !== undefined || props.icon !== undefined;
 
     // This style must be set here because we need to know if text content is passed in the final render to set the height correctly
+    let minHeight = 0;
+    if (props.vertical) {
+      minHeight = hasContent ? 84 : globalTokens.size200;
+    }
     const mergedRootProps = {
```
rurikoaraki, line 69: This const should probably be returned from a memoized function.

rurikoaraki: That would probably fix your rerender test.

PPatBoyd (approved):

PR #2679, opened 2023-03-03
[Button] HC fixes button

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Client reported some issues with the Button in HC.

It looks like we're using colors for plain text on the Button when we should be using ButtonText instead. This works for primary buttons but for other ones we're using the wrong color.

The reason this happens is because the alias tokens don't differentiate between usage in a canvas vs. usage in a button, so most alias tokens point to canvas/text instead of button/button text. Web fixes this by overriding the colors in HC specifically, so we'll do the same. Seems like a failure of the alias token system, but that's a separate conversation.

Changes:
1. Add a utility function to check if we're in HC. This function operates differently depending on the platform. On mobile it no-ops, on macOS it retrieves a stored value, on windows it uses AppTheme and on win32 it reads a property on the Theme object.
2. Replace current checks for high contrast with this new one
3. Update colors on Button, Compound Button and ToggleButton

### Verification

| image image image | image image image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/theming-utils/src/isHighContrast.macos.ts
```diff
@@ -0,0 +1,26 @@
...
+let isHighContrastEnabled = false;
+
+/**
+ * Used to set isHighContrast state for FURN
+ * Specifically for macOS, where you cannot ask for HC state,
+ * so you have to store the value to keep track of it.
+ *
+ * @platform macOS
+ * @param isHighContrast New value of high contrast state
+ */
```
Saadnajmi, line 12: It's occurring to me we could probably just... fix this in React Native macOS so you can ask for the value. Would that simplify things for FURN? @chiuam fyi

packages/theming/theming-utils/src/isHighContrast.windows.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Gives the state of high contrast in FURN.
+ * On win32, the t parameter is required, but on other platforms
+ * it's not needed.
+ *
+ * @param t The Theme object. Required on win32.
+ */
+export function isHighContrast(_t?: Theme): boolean {
+  return AppTheme.isHighContrast;
```
Saadnajmi, line 24: I think this is deprecated?
https://microsoft.github.io/react-native-windows/docs/apptheme-api

packages/theming/theming-utils/src/isHighContrast.macos.ts
rurikoaraki: I mean up to you? I don't think it particularly matters since macOS' implementation would probably be different from RNW anyway..

Saadnajmi: I wasn't sure how bad the workaround in FURN was. It seems like RNW is a custom implementation so converging won't happen anytime soon...

I think more likely we'll push a change to React Native Core for iOS HC, and align to that.

Saadnajmi (approved):

packages/theming/theming-utils/src/isHighContrast.macos.ts
chiuam: re: fixing HC in react-native-macos - Agreed that we should expose an API for detecting HC, I ended up implementing that in FURN as a workaround to unblock ButtonV1.

chiuam (approved):

packages/theming/theming-utils/src/isHighContrast.macos.ts
Saadnajmi: @chiuam with filing a bug? Maybe we can get it with 0.71

Saadnajmi: Oh, cool! I'd be happy to leave that as a followup outside of this PR to be clear, to reduce risk.

chiuam: Sounds good! Maybe when we tap into some of the iOS HC work?

PR #2681, opened 2023-03-03
Only call `ref.current.focus()` if the platform supports it.

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

After #2669 we were getting a Redbox on iOS that `focus()` didn't exist as a ViewManager command. This is because our hook `useOnPressWithFocus` was calling focus() when that is only valid on desktop platforms. This fix is a simple platform check to ensure we only call focus() if the platform supports it. 

### Verification

iOS no longer Redboxes on any button press.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-interactive-hooks-7e7228ba-5e22-41e2-8b7a-9697ad2f07ff.json
+++ b/change/@fluentui-react-native-interactive-hooks-7e7228ba-5e22-41e2-8b7a-9697ad2f07ff.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Only call focus if the platform supports it.",
+  "packageName": "@fluentui-react-native/interactive-hooks",
+  "email": "sanajmi@microsoft.com",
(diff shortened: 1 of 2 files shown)


chiuam (approved):

rurikoaraki (approved):

mischreiber (approved):

acoates-ms:
Just checking that this hook doesn't get used on TextInputs?  Since those do support focus on iOS and android.

Saadnajmi:
> Just checking that this hook doesn't get used on TextInputs? Since those do support focus on iOS and android.

We don't have any components based on TextInput at the moment in FURN. Unless a downstream user it using this hook with their TextInput on mobile, I think we should be OK.

acoates-ms (approved):

PR #2682, opened 2023-03-04
[DRAFT]Remove hover effect on menuItem when keyboard focusing on other MenuItem

lenahong:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes
Adding a useLayoutEffect hook that will check for isMenuItemHovered state and remove hover effect if keyboard focus is another menuItem.

### Verification

Needs further testing.

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -99,12 +114,23 @@ function getAccessibilityStateWorker(disabled: boolean, accessibilityState?: Acc
   return { disabled };
 }
 
-export const useHoverFocusEffect = (hovered: boolean, componentRef: React.MutableRefObject<any>) => {
+export const useHoverFocusEffect = (hovered: boolean, componentRef: React.MutableRefObject<any>, setIsMenuItemHovered: any) => {
```
rurikoaraki, line 117: I feel like you could simplify this to:

```
export const useHoverFocusEffect = React.useLayoutEffect((hovered: boolean, componentRef: React.MutableRefObject<any>, setIsMenuItemHovered: any) => {
//etc.
```

instead of wrapping the useLayoutEffect hook in another function

Same with useRemoveHoverEffect

packages/components/Menu/src/MenuList/MenuList.types.ts
```diff
@@ -68,6 +68,8 @@ export interface MenuListState extends Omit<MenuListProps, 'checked' | 'onChecke
   removeRadioItem: (name: string) => void;
   hasMaxHeight?: boolean;
   hasMaxWidth?: boolean;
+  isMenuItemHovered?: boolean | undefined;
+  setIsMenuItemHovered?: (isMenuItemHovered: boolean | undefined) => void;
```
rurikoaraki, line 72: nit: Could be isMenuItemHovered?: boolean

packages/components/Menu/src/MenuItem/useMenuItem.ts
rurikoaraki, line 117: type of setIsMenuItemHovered should probably be defined instead of using `any`

same for onHoverOut.

packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -60,11 +61,25 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
...
-  useHoverFocusEffect(pressable.state.hovered, componentRef);
+  useHoverFocusEffect(pressable.state.hovered, componentRef, setIsMenuItemHovered);
+  useRemoveHoverEffect(pressable.state.hovered, menuListContext.isMenuItemHovered, pressable.props.onHoverOut);
+
+  const onFocusOverride = React.useCallback(
+    (e) => {
+      pressable.props.onFocus(e);
+      // if it's not a hover focus
+      if (!pressable.state.focused && !pressable.state.hovered) {
+        // remove hover state from everywhere else
```
rurikoaraki, line 72: I'm guessing that onBlur doesn't fire on win32? Otherwise that seems like a better solution (listen for when focus moves somewhere else and clear hover in that case)

packages/components/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts
```diff
@@ -120,9 +122,22 @@ export const useMenuCheckboxInteraction = (
     hasTooltips,
   };
 
+  const onFocusOverride = React.useCallback(
+    (e) => {
+      pressable.props.onFocus(e);
+      // if it's not a hover focus
+      if (!pressable.state.focused && !pressable.state.hovered) {
```
Saadnajmi, line 129: Is `hover focus` a windows term? I assume it means "mouse hover". I feel like I'd find this to be more clear if it was worded like "If not focused via mouse hover"

packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -99,12 +114,23 @@ function getAccessibilityStateWorker(disabled: boolean, accessibilityState?: Acc
...
+      setIsMenuItemHovered(undefined);
+    }
+  }, [hovered, componentRef, setIsMenuItemHovered]);
+};
+
+export const useRemoveHoverEffect = (hovered: boolean, isMenuItemHovered: boolean, onHoverOut: any) => {
+  React.useLayoutEffect(() => {
+    if (hovered && isMenuItemHovered != undefined && !isMenuItemHovered) {
+      // remove Focus from this menuITem
+      onHoverOut();
```
Saadnajmi, line 133: I had a similar comment in a previous PR (not one you made). 
Manually calling `onHoverOut` feels like tricking React Native that mouse hover is gone, but doesn't actually reflect the native state (you could call `onHoverOut` while the mouse is over the component for example, and now native / RN state are out of sync). Why do we need to manually call it?

lenahong: That's exactly what we are trying to do actually, calling onHoverOut while the mouse is over the component but keyboard focus is on a different component. Then, even if the mouse is still over the component, as long as you're keyboard focusing through other components, we don't want hover sate.

Saadnajmi: On macOS, we do that by not having any special styling for `hover`. We only set tokens for `focus` and `onHoverIn/Out` calls `ref.current.focus` via the `useFocusHoverEffect` hook. We effectively don't use hover to inform any styling, focus is the "one" source of truth, and hover updates focus. https://github.com/microsoft/fluentui-react-native/blob/4430c6d93553e25870ab350fb7304adbfd2d456b/packages/components/Menu/src/MenuItem/MenuItemTokens.macos.ts#L23-L27

Is there different styling for hover state and focus state on menu items in win32? On macOS they look the same. I had assumed till now that on win32, you could have a menu item hovered and another menu item keyboard focused and they would have different looking selection states. It sounds like from this PR, we actually only want one state: selected. If so, would what we do on macOS work for win32?

Saadnajmi: or (maybe too much for this PR) have both hover and focus update some state at the MenuList level of the currently "selected" menu item? That way the MenuList is what manages styling, and each menu item is just using its callbacks to update the entire menu's state?

lenahong: We actually do have separate styling defined for hovered state. Focused state and hovered state styles are identical. Saad, I think you just made me realize this could have been a one liner fix LOL

PR #2687, opened 2023-03-07
[Menu] Implement N of M pattern on items

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Implementing population of N of M props for MenuItems so that

This change is done in the MenuList. The render function calculates n and m and fills out the appropriate props for both windows and win32.

### Verification

Checked with tester, need to test a little more.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuList/MenuList.tsx
+++ b/packages/components/Menu/src/MenuList/MenuList.tsx
@@ -58,6 +58,19 @@ export const MenuList = compose<MenuListType>({
     }, []);
 
     return (_final: MenuListProps, children: React.ReactNode) => {
+      const childrenWithSet = React.Children.map(children, (child, index: number) => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, {
+            accessibilityPositionInSet: index, // win32
+            accessibilityPosInSet: index, // windows
+            accessibilitySetSize: React.Children.toArray(children).length, //win32
+            accessibilitySizeOfSet: React.Children.toArray(children).length, //windows
+          } as any);
+        }
+
+        return child;
+      });
+
       const content =
         Platform.OS === 'macos' ? (
           <Slots.root>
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -58,6 +58,19 @@ export const MenuList = compose<MenuListType>({
     }, []);
 
     return (_final: MenuListProps, children: React.ReactNode) => {
+      const childrenWithSet = React.Children.map(children, (child, index: number) => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, {
+            accessibilityPositionInSet: index, // win32
+            accessibilityPosInSet: index, // windows
+            accessibilitySetSize: React.Children.toArray(children).length, //win32
+            accessibilitySizeOfSet: React.Children.toArray(children).length, //windows
```
Saadnajmi, line 67: Would this be better as a Platform.Select()?

rurikoaraki: Platform.select doesn't work for win32. I can do a conditional but unless the extra props are a problem, it seems fine to pass everything in

Saadnajmi: I don't think it's a problem, though it'll be something to keep in mind for RN upgrades (in case win32 implements the windows props in the future, or macOS does, etc..). I'll consider it not blocking I guess.

Side note, do you know why these differ between the windows platforms?

rurikoaraki: Yeah, in the future I expect the win32 props to align with windows.

I don't know why they don't align, my guess would be that they were both created around the same time and didn't consult each other.

Saadnajmi (approved):

chiuam (approved):

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -58,6 +58,19 @@ export const MenuList = compose<MenuListType>({
     }, []);
 
     return (_final: MenuListProps, children: React.ReactNode) => {
+      const childrenWithSet = React.Children.map(children, (child, index: number) => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, {
+            accessibilityPositionInSet: index, // win32
+            accessibilityPosInSet: index, // windows
+            accessibilitySetSize: React.Children.toArray(children).length, //win32
```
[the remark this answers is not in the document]
rurikoaraki: Mm yeah that's a good catch, I changed the logic to filter out all non-elements.

PR #2693, opened 2023-03-09
[ButtonV1][Win32] Fix focus borders on primary buttons

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Our focus border on win32 was deemed not meeting a11y color contrast guidelines, so we needed to make a change to them. Currently the design is that there is a two-tone border for primary buttons, one color contrasting against the background and one color contrasting against the background of the primary button. Native accomplishes this with an outer and inner border applied to the control, but this is not possible via our RN APIs.

The solution I've come up with is to have an absolutely positioned View above the Button with the inner border color. This is accomplished by adding a new Slot which is only created in win32 to minimize impact to other platforms. The size is calculated via a layout callback, otherwise there isn't a way for the View to know it's size since it's absolutely positioned.

The caveat of this change is that it has rendering issues on higher DPI levels. This seems to be due to how the layout is done at higher DPIs, and results in uneven distribution of the border. Because we're limited by the RN APIs, this is something we have to live with. PM has signed off on how it looks.

Other options:
1. Have a View child inside of the root View instead of having it absolutely positioned. This results in different visual bugs at other DPI levels that looked worse to other people, so I went with this option instead.
2. Have an SVG draw the border. This idea was brought up by @saadnajmi. Might be worth exploring later but not something we have time for right now.

This change mainly affects win32 primary buttons.

### Verification

| image | image image image image image image image |

125:
image

150:
image

200:
image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/Button.styling.ts
```diff
@@ -37,40 +37,41 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
...
-              borderWidth: tokens.borderInnerWidth,
-              borderRadius: tokens.borderRadius,
-              overflow: 'hidden',
-            },
-          };
-        },
-        ['borderRadius'],
-      ),
-    }),
+    rippleContainer: buildProps(
```
rurikoaraki, line 40: Whoops. Need to pue this back.

packages/components/Button/src/Button.styling.ts
```diff
@@ -15,19 +15,19 @@ import { defaultButtonTokens } from './ButtonTokens';
...
-  'subtle',
-  'outline',
-  'hovered',
   'small',
   'medium',
   'large',
   'hasContent',
   'hasIconAfter',
   'hasIconBefore',
+  'primary',
```
rurikoaraki, line 24: This was moved because now primary has to take size into account when adjusting padding.

packages/components/Button/src/Button.tsx
```diff
@@ -110,9 +111,26 @@ export const Button = compose<ButtonType>({
...
+            )}
           </Slots.root>
         );
       }
     };
   },
 });
+
+const getFocusBorderStyleWorker = (height: number, width: number) => {
+  const adjustment = 2; // width of border * 2
```
PPatBoyd, line 129: nit is the comment accurate w/ `*`?

packages/components/Button/src/Button.types.ts
```diff
@@ -166,14 +166,23 @@ export interface ButtonProps extends ButtonCoreProps {
   loading?: boolean;
 }
 
+interface ButtonState {
+  height?: number;
```
PPatBoyd, line 170: should these be `measuredHeight` `measuredWidth` for variable-sized buttons e.g. localization?

Saadnajmi:
I think I already asked this but for my own sake:

Why doesn't the default focus ring from `enableFocusRing` have this two-tone effect? Is this a special case we need only for primary button because of a11y/contrast? And if so, we'll probably need it for _any_ focusable solidly filled control?

packages/components/Button/src/Button.tsx
rurikoaraki: Yeah 2 is the width of the border twice.

rurikoaraki:
> I think I already asked this but for my own sake:
> 
> Why doesn't the default focus ring from `enableFocusRing` have this two-tone effect? Is this a special case we need only for primary button because of a11y/contrast? And if so, we'll probably need it for _any_ focusable solidly filled control?

Yes it's specifically for primary buttons in win32 for color contrast. It's true that you would need it for any focusable solidly filled control, but in practice I can only think of primary buttons having that issue. All other controls use neutral backgrounds, which don't have this issue.

packages/components/Button/src/Button.types.ts
rurikoaraki: Sure, I can rename them.

Saadnajmi (approved):

packages/components/Button/src/Button.types.ts
```diff
@@ -166,14 +166,23 @@ export interface ButtonProps extends ButtonCoreProps {
...
+  measuredWidth?: number;
+
+  // win32 only. Whether the component should use a tone-tone focus border instead of single-tone
+  shouldUseTwoToneFocusBorder?: boolean;
+}
+
 export interface ButtonInfo {
   props: ButtonProps & React.ComponentPropsWithRef<any>;
-  state: PressableState;
+  state: PressableState & ButtonState;
```
Saadnajmi, line 179: nit: should ButtonState just extend PressableState? There was a pattern I followed when converting other views to `usePressableState`, lemme look it up..

Saadnajmi: Yeah, that's what we did for Checkbox and Switch

PPatBoyd (approved):

packages/components/Button/src/Button.types.ts
rurikoaraki: Sure, can do.

PR #2694, opened 2023-03-10
[Menu] Implement N of M pattern on items

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Implementing population of N of M props for MenuItems. N of M props can still be overwritten using N of M props on MenuItems and Menu (for sub menus).

This change is done in the MenuList. The render function calculates n and m and fills out the appropriate props for win32.

### Verification

Manually tested and added an example of how N of M props can be overwritten in the Fluent Tester.

N of M prop on MenuItem
NofMOnMenuItem

N of M Prop on SubMenu Trigger Item
NofMOnSubMenuTriggerItem

N of M Prop Overwritten on Menu Item
NofMOverwrittenOnMenuItem

N of M Prop Overwritten on SubMenu Trigger Item
NofMOverwrittenOnSubmenuTriggerItem

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


packages/components/Menu/src/MenuItem/MenuItem.tsx
```diff
@@ -54,7 +54,11 @@ export const MenuItem = compose<MenuItemType>({
               {icon && (icon.svgSource || icon.fontSource) && <Slots.fontOrSvgIcon {...icon} />}
             </Slots.iconPlaceholder>
           )}
-          {children && <Slots.content tooltip={tooltipResult}>{children}</Slots.content>}
+          {children && (
```
rurikoaraki, line 57: Thanks 😅 Do we need to pass accessible={false} to the icons as well?

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -58,17 +59,27 @@ export const MenuList = compose<MenuListType>({
...
-        return React.cloneElement(
-          child as React.ReactElement<unknown, string | React.JSXElementConstructor<any>>,
-          {
-            accessibilityPositionInSet: index + 1, // win32
-            accessibilityPosInSet: index, // windows
-            accessibilitySetSize: filteredChildren.length, //win32
-            accessibilitySizeOfSet: filteredChildren.length, //windows
-          } as any,
-        );
+      const itemCount = React.Children.toArray(children).filter(
```
rurikoaraki, line 62: Maybe store the result of the filtered array so you can reuse it to create childrenWithSet?

packages/components/Menu/src/Menu/Menu.tsx
```diff
@@ -25,9 +26,25 @@ export const Menu = stagedComponent((props: MenuProps) => {
     const menuTrigger = childrenArray[0];
     const menuPopover = childrenArray[1];
 
+    const menuTriggerChild = menuTrigger.props.children;
+
+    const menuTriggerChildWithSet = React.cloneElement(
+      menuTriggerChild as React.ReactElement<unknown, string | React.JSXElementConstructor<any>>,
+      {
+        accessibilityPositionInSet: menuTriggerChild.props.accessibilityPositionInSet ?? accessibilityPositionInSet, // win32
```
rurikoaraki, line 34: I wonder if it would be more robust to put these on the MenuContext instead... This would work when the MenuTrigger is the immediate child but I think we have clients who wrap the MenuTrigger in another component..

packages/components/Menu/src/MenuList/MenuList.tsx
ksiler: The reason I didn't do this is because we're filtering out MenuDividers. If I were to reuse this array, any MenuDivider wouldn't be included in childrenWithSet.

packages/components/Menu/src/MenuItem/MenuItem.tsx
ksiler: They didn't show up in the a11y tree so I didn't add them but I guess it wouldn't hurt. I'm not seeing that icons aren't accessible by default.

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -58,6 +59,29 @@ export const MenuList = compose<MenuListType>({
...
 
     return (_final: MenuListProps, children: React.ReactNode) => {
+      const itemCount = React.Children.toArray(children).filter(
+        (child) => React.isValidElement(child) && child.type !== MenuDivider,
+      ).length;
+      let itemPosition = 0;
+
+      const childrenWithSet = React.Children.toArray(children).map((child) => {
+        if (React.isValidElement(child)) {
+          if ((child as any).type !== MenuDivider) {
```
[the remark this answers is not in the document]
rurikoaraki: I think that's fine. It's possible that the check could be more successful if you check for type.displayName instead, but also if as a client customizing controls comes with the risk that you might break default behaviors.

packages/components/Menu/src/MenuList/MenuList.tsx
rurikoaraki: Ah, that makes sense.

packages/components/Menu/src/MenuList/MenuList.tsx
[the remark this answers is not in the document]
ksiler: Huh, I tried type.displayName before and thought it wasn't working but I just tried it again and now it works :) I'll update this so customized dividers will get filtered out.

PR #2697, opened 2023-03-10
iOS FRNAppearanceAdditions: support for multiple trait collections

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Second attempt to add support for multiple trait collections (first attempt [here](https://github.com/microsoft/fluentui-react-native/pull/2675), was closed after realizing that native modules are not singletons/one instance is created per scene)

Context: Before this change, the FRNAppearanceAdditions native module assumed that there was only one window/root view at a time, i.e. one trait collection at a time. This assumption doesn't hold for iPad multiwindow scenarios where there can be more than one window at a time, each having different trait collections.

Getting the iPad multiwindow scenario to work can be broken down into two parts:
1. Making sure that the right windows have the right traits when traits change during runtime
   - Since each native module instance corresponds to one iOS scene, and we've made the assumption that there is only one unique trait collection per scene (for the traits we care about), we could keep the instance variables (`_horizontalSizeClass` `_userInterfaceLevel` `_accessibilityContrastOption`)  as is. The behavior that was happening before this change was every time the `RCTUserInterfaceStyleDidChangeNotification` was fired from _any_ RCTRootView, it would get sent to _every_ native module instance, and update every scene. So all that was needed was to add a check in the `appearanceChanged` event in `RCTAppearanceAdditions.m` that the view that the notification was fired from corresponded to the right native module instance before passing along the message to the JS side
2. Making sure that the right windows have the right traits on boot
   - Ideally, there would be a way to determine which RCTRootView/window/scene corresponds to the current instance of a native module, so we can initialize the traits before the first time we ask for the any traits from the JS side. However, since we didn't find a way to do this, the next option was to allow the user to pass in the viewTag/rootTag from the JS side when getting any of the traits, and if a viewTag is provided we'll look up the right trait collection from the UIManager and return that. The remaining changes in this PR are to support this.
     - Note: the react native rootTag identifier returns the react tag of the root view. More documentation about rootTag [here](https://reactnative.dev/docs/roottag)

State of the different traits we support at this point:
1. Accessibility contrast (normal vs. high)
    1. Already worked, since it is a system wide setting/not specific to a window
2. Appearance size class (compact vs. regular)
    1. Requires support for multiple trait collections - now works after this PR. See before & after videos below
3. User interface level (base vs. elevated)
    1. Requires support for multiple trait collections - now works in this PR.
    2. ~~Also requires support for multiple themes (out of scope for this PR)~~

### Verification

This change was tested using multiple windows in compact/regular mode:

Before (can't have one side be compact mode, other side be regular mode - both windows will render the same way, which means that one side is sometimes incorrect)

https://user-images.githubusercontent.com/78454019/222306286-97cedf7e-a69e-4691-a08d-415c8aac3ee4.mov

After (each window will render in compact more or regular mode correctly - both at boot, and at runtime)

https://user-images.githubusercontent.com/78454019/229221239-2c1cee4a-22b7-4e53-b18b-93999967a8f4.mov

Other scenarios tested:
- slideover view
- opening lots of windows in varying states of compact vs. regular width

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -105,23 +107,20 @@ - (dispatch_queue_t)methodQueue
...
-    UITraitCollection *initialTraitCollection;
-    
-    if (viewControllerWithInitialTraitCollection != nil) {
-        initialTraitCollection = [viewControllerWithInitialTraitCollection traitCollection];
-    } else {
-        initialTraitCollection = [UITraitCollection currentTraitCollection];
+    for (UIWindow *window in RCTSharedApplication().windows) {
+        RCTRootView *rootView = (RCTRootView *)[[window rootViewController] view];
+        
+        if ([rootView isKindOfClass:[RCTRootView class]]) {
```
Saadnajmi, line 113: Feels kinda redundant? Maybe line 111 should just be `UIView`?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -105,23 +107,20 @@ - (dispatch_queue_t)methodQueue
...
+            if ([[rootView bridge] bundleURL] == [[self bridge] bundleURL]) {
```
Saadnajmi, line 114: The bundleURL isn't the same across windows?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
mischreiber: I'm cool with either pattern - if you declare `rootView` as `UIView` you'll just have to cast later anyway.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
mischreiber: Can we literally check `if ([[rootView bridge] isEqual:[self bridge]])` ?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -137,9 +136,21 @@ - (void)stopObserving {
...
     if (_hasListeners) {
+        RCTBridge *currentBridge = [self bridge];
+        RCTBridge *notificationBridge = [[notification object] bridge];
+        
+        if (![notificationBridge isKindOfClass:[RCTBridge class]]) {
+            return;
+        }
+        
+        // Don't send the appearanceChanged event if the notification didn't originate from the same react native instance
+        if ([currentBridge bundleURL] != [notificationBridge bundleURL]) {
```
mischreiber, line 147: Same as earlier: should be able to just check literal object equality.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: So at runtime, `[rootView bridge]` is an RCTBridge, but [self bridge] is an RCTCxxBridge, so that check alone always fails. I looked up what RCTCxxBridge is - it gets initialized with a parent RCTBridge. I had to import <React/RCTBridge+Private.h> in order to be able to access the parent bridge, which I'm not sure if I should do, but this works the same as before.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: Updated to check is the pointers to the bridges refer to the same instance

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -97,31 +97,27 @@ + (BOOL)requiresMainQueueSetup {
...
-    // we need to access a view that's been added to the view hierarchy
-    UIViewController *viewControllerWithInitialTraitCollection = RCTPresentedViewController();
-    UITraitCollection *initialTraitCollection;
-    
-    if (viewControllerWithInitialTraitCollection != nil) {
-        initialTraitCollection = [viewControllerWithInitialTraitCollection traitCollection];
-    } else {
-        initialTraitCollection = [UITraitCollection currentTraitCollection];
+    for (UIWindow *window in RCTSharedApplication().windows) {
+        RCTRootView *rootView = (RCTRootView *)[[window rootViewController] view];
```
tido64, line 108: I also mentioned this on Discord: With hybrid apps, I don't think this is always true. `RCTRootView` can live further down the tree.

Getting the node handle should give you a more accurate result. The issue with this approach is that it may require thread jumps. You need to resolve the tag inside `RCTExecuteOnMainQueue` to safely access `RCTUIManager`.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: Resolving since I ended up removing this code

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: Thanks for pointing that out timmy - we also realized that trying to find a RCTRootView from UISharedApplication wasn't the right approach since RCTRootView might not be the actual root view. I've removed this code & now this PR relied on passing the viewTag from the JS side to get the right trait collection. 

I was able to use `RCTExecuteOnMainQueue` to get this working - thanks for letting me know about that - and ended up not needing to get the node handle

lyzhan7:
Update: after the recent changes to this PR, trait collection with iPad multiwindow works as expected for most scenarios. I've tried testing with split window, slideover window, and opening a bunch of windows. 

I am still trying to understand why the current state of the PR is getting the right behavior - I thought that after we execute the view lookup on the main queue, we would need to return some kind of callback/fire some event to let the JS side know what the trait value is. Somehow this wasn't necessary; just updating the stored trait value in the native module instance seems to already trigger the JS side to re-render with the right value, with no flashing. 

There's one specific scenario that doesn't work as expected, and it's when a new window is created in an existing compact window. Sometimes this works as expected (the new FluentTester window is correctly in compact mode), but sometimes it doesn't (gets rendered in regular mode). Video here:


https://user-images.githubusercontent.com/78454019/229242638-414019db-2de1-4bb6-a5a6-3dc73c42b935.mov



You can see that the first window we open is in compact mode, but subsequent windows are in regular mode.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -76,18 +80,29 @@ + (BOOL)requiresMainQueueSetup {
...
+            NSString *horizontalSizeClass = RCTHorizontalSizeClassPreference([view traitCollection]);
+            self->_horizontalSizeClass = horizontalSizeClass;
+        }
+    });
     return _horizontalSizeClass;
 }
 
-RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(userInterfaceLevel)
-{
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(userInterfaceLevel:(id)reactTag) {
```
lyzhan7, line 94: Note - I tried overloading this method so that we can call userInterfaceLevel() instead of userInterfaceLevel(null) but ran into issues, it seems like with RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD only one of these methods can be recognized at a time.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -137,9 +134,20 @@ - (void)stopObserving {
...
+        // Don't send the appearanceChanged event if the notification didn't originate from the same react native instance
+        RCTBridge *currentBridge = [[self bridge] parentBridge];
+        if (![currentBridge isEqual:notificationBridge]) {
+            return;
+        }
+
         UITraitCollection *traitCollection = [[notification userInfo] valueForKey:RCTUserInterfaceStyleDidChangeNotificationTraitCollectionKey];
         if (![traitCollection isKindOfClass:[UITraitCollection class]]) {
-            traitCollection = nil;
+            return;
```
Saadnajmi, line 150: Should we do some sort of assert or log here?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -76,18 +80,29 @@ + (BOOL)requiresMainQueueSetup {
...
 }
 
-RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(userInterfaceLevel)
-{
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(userInterfaceLevel:(id)reactTag) {
+    RCTUnsafeExecuteOnMainQueueSync(^{
+        if ([reactTag isKindOfClass:[NSNumber class]]) {
+            UIView *view = [[[self bridge] uiManager] viewForReactTag:reactTag];
+            NSString *horizontalSizeClass = RCTHorizontalSizeClassPreference([view traitCollection]);
+            self->_horizontalSizeClass = horizontalSizeClass;
```
Saadnajmi, line 99: should be `userInterfaceLevel`, not `horizontalSizeClass`?

lyzhan7: Oooh yes thank you

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: I don't expect us to ever get into this state, It shouldn't be possible. I can add an assert if you think there should still be on. There's similar logic in the RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD calls as well (just flipped), do you think it should be added there as well?

lyzhan7:
Update: after the most recent commit (replacing RCTExecuteOnMainQueue with RCTUnsafeExecuteOnMainQueueSync), this PR works for all the multiwindow scenarios I've tested (including the specific scenario I mentioned in the comment above). Tests include split view, slideover view, opening new windows in split/slideover view, and opening and closing and reopening lots of windows in varying states of compact vs. regular width

Saadnajmi (approved):

lyzhan7:
Update: I refactored the code a bit since the root tag really only needs to be sent to the native module once when it gets initialized. I made an initializeTraitCollection method, and was keep the previous horizontalSizeClass/userInterfaceLevel getters as they were before. 

This actually also seems to work for all the dark elevated scenarios I've tested (note if a FluentTester window is in dark elevated mode, the background is dark grey, if it's in dark mode, the background is black)
- When there's only one window, FluentTester is in dark mode
- In split view mode, any FluentTester windows should be dark elevated
- In slideover mode, and the top window is FluentTester, FluentTester is in dark elevated mode
- In slideover mode, and the bottom window is FluentTester, FluentTester should be in dark mode


https://user-images.githubusercontent.com/78454019/230217934-9574649b-6637-45cf-9b2b-f70a0a741d19.mov


One additional change I had to make (most recent commit) was add null checks in appearanceAdditions.ios.ts when receiving appearanceChanged events. RCTAppearance also fires an event called appearanceChanged, and for some reason that only started causing issues after the recent changes. I might also be able to just change the event name in FRNAppearanceAdditions

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -76,18 +80,35 @@ + (BOOL)requiresMainQueueSetup {
...
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(initializeTraitCollection:(id)reactTag) {
+    RCTUnsafeExecuteOnMainQueueSync(^{
+        if ([reactTag isKindOfClass:[NSNumber class]]) {
+            UIView *view = [[[self bridge] uiManager] viewForReactTag:reactTag];
+            self->_horizontalSizeClass = RCTHorizontalSizeClassPreference([view traitCollection]);
+            self->_userInterfaceLevel = RCTUserInterfaceLevelPreference([view traitCollection]);
+            self->_accessibilityContrastOption = RCTAccessibilityContrastPreference([view traitCollection]);
+        }
+    });
+    return nil;
```
Saadnajmi, line 100: Do we need this?

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -137,9 +166,20 @@ - (void)stopObserving {
 
 - (void)appearanceChanged:(NSNotification *)notification {
     if (_hasListeners) {
+        RCTBridge *notificationBridge = [[notification object] bridge];
+        if (![notificationBridge isKindOfClass:[RCTBridge class]]) {
+            return;
+        }
+
+        // Don't send the appearanceChanged event if the notification didn't originate from the same react native instance
+        RCTBridge *currentBridge = [[self bridge] parentBridge];
```
Saadnajmi, line 175: where does parentBridge come from? I've never seen that before.

Saadnajmi (approved):

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: [self bridge] actually returns a RCTxxBridge, which is only apparent at runtime. It's the C++ version of RCTBridge, and gets initialized with an RCTBridge (which is accessed with parentBridge)

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: Yeah that's a good question - I don't need to return anything in this method, but RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD requires a return value. I thought we would still want the init method to be synchronous on the JS thread so I just returned null, but the other option might be to change it to RCT_EXPORT_METHOD. I can try it and see if it works

lyzhan7: Just tried RCT_EXPORT_METHOD, and it doesn't work - I think the code here needs to block the JS thread. If there's an alternative to returning nil in RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD let me know, but I don't see anything else I could do

Saadnajmi: returning nil is fine, didn't realize it was required.

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
```diff
@@ -97,37 +118,45 @@ + (BOOL)requiresMainQueueSetup {
...
+ When this native module is first initialized, we don't have access to the information necessary to consistently retrieve all the right traits.
+ We should still attempt to initialize the traits with the right values with the information we do have access to.
+ 
+ The traits retrieved from RCTPresentedViewController() should be correct for non-multiwindow scenarios.
+ 
+ The traits retrived from [UITraitCollection currentTraitCollection] will always be the same default trait collection,
+ presumably because FRNAppearanceAdditions isn't a view, so it never gets updated with the right traitCollection
+ (which happens when a view gets added to the view hierachy).
+ */
+-(UITraitCollection *)attemptTraitCollectionInitialization {
```
harrieshin, line 150: super nit space

harrieshin (approved):

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
mischreiber: That's a very heavily debated issue - personally I'm a bigger fan of failing safe, like here, though if there was some way to log that'd be nice. But personally I wouldn't crash.

mischreiber (approved):

packages/experimental/AppearanceAdditions/ios/FRNAppearanceAdditions.m
lyzhan7: I'll start a conversation about logging in RN because I haven't seen any usages of it, and can revisit this

PR #2701, opened 2023-03-13
Add Menu "Home" and "End" key functionality

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, the Menu lacks the ability to use the "HOME" and "END" keys to switch focus to the beginning and end items. This PR adds a JS implementation of this functionality

### Verification

Behavior in action:
https://user-images.githubusercontent.com/15683103/233462693-18faae53-6e57-46c4-9033-2ce5050cb4eb.mp4

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuList/useMenuList.ts
```diff
@@ -129,7 +170,27 @@ export const useMenuList = (_props: MenuListProps): MenuListState => {
...
+      ref,
+      disabled,
+    }),
+    [ref, disabled],
+  );
+
+  React.useEffect(() => {
+    trackMenuItem(item);
+    return () => untrackMenuItem(item);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rurikoaraki, line 194: nit: I'd add a comment that this is intentional since you only want it to run once

packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -62,6 +63,9 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
 
   useHoverFocusEffect(pressable.state.hovered, componentRef);
 
+  // Track the ref and disabled props on this menu item so the MenuList can handle Home and End keypresses.
+  useMenuItemTracking(componentRef, disabled);
```
rurikoaraki, line 67: Is this guaranteed to run in order of the children array?

packages/components/Menu/src/MenuList/useMenuList.ts
```diff
@@ -112,6 +118,41 @@ export const useMenuList = (_props: MenuListProps): MenuListState => {
...
+        trackedMenuItems.findIndex((x) => x.ref === item.ref),
+        1,
+      ),
+    [trackedMenuItems],
+  );
+
+  const onListKeyDown = (e: InteractionEvent) => {
+    const key = e.nativeEvent.key;
+    if (handledKeys.includes(key)) {
+      // For iOS and macOS, 'Home' and 'End' must set focus on the first and last enabled item.
```
rurikoaraki, line 139: Looks like this includes win32 too?

packages/components/Menu/src/MenuList/useMenuList.ts
```diff
@@ -112,6 +118,41 @@ export const useMenuList = (_props: MenuListProps): MenuListState => {
...
+        increment = 1;
+        idx = 0;
+      } else if (key === 'End') {
+        increment = -1;
+        idx = trackedMenuItems.length - 1;
+      }
+      while (platformsWithoutFocusOnDisabled.includes(Platform.OS) && trackedMenuItems[idx].disabled) {
+        idx += increment;
+      }
+      trackedMenuItems[idx].ref.current.focus();
```
rurikoaraki, line 152: We should check that ref.current is still defined (ref.current?.)

Saadnajmi:
Suggestion (feel free to reject if it doesn't work):

Handle focus natively rather than JS, via FocusZone. 

Reasons being:
1) We add FocusZone for macOS but not win32. Having FocusZone on both platforms would reduce platform differences. When I asked about adding FocusZone for win32, most people just kinda shrugged whether it was a good or bad idea. 
2) Handling focus in JS is somewhat hard. All we really have access to is the imperative `ref.current.focus()` method. I didn't take too deep a look at the PR, but it felt like there were some additional hoops to jump through to get stuff working. Also handling keyboard events in JS is not the best either (platform differences sad..)
3) I thought about the existing use cases of FocusZone (Radiogroup, lists) and it didn't.. feel that bad to also just get Home/End for free there?

Downsides:
1) Would have to be natively implemented for win32 and macOS. I'm familiar with the macOS half, but not the win32.
2) It may end up being more complicated / change behavior onexisting clients

Again. Feel free to completely reject this! I propose it as an alternative if you feel like doing this in JS is too cumbersome :)

packages/components/Menu/src/MenuItem/useMenuItem.ts
lawrencewin: Are there scenarios where React children aren't rendered in order? I was under the impression that the children are rendered inorder in the tree, but I could be wrong.

packages/components/Menu/src/MenuList/useMenuList.ts
lawrencewin: Mistakenly left in during testing.

lawrencewin:
> > Suggestion (feel free to reject if it doesn't work):
> > Handle focus natively rather than JS, via FocusZone.
> > Reasons being:
> > 
> > 1. We add FocusZone for macOS but not win32. Having FocusZone on both platforms would reduce platform differences. When I asked about adding FocusZone for win32, most people just kinda shrugged whether it was a good or bad idea.
> > 2. Handling focus in JS is somewhat hard. All we really have access to is the imperative `ref.current.focus()` method. I didn't take too deep a look at the PR, but it felt like there were some additional hoops to jump through to get stuff working. Also handling keyboard events in JS is not the best either (platform differences sad..)
> > 3. I thought about the existing use cases of FocusZone (Radiogroup, lists) and it didn't.. feel that bad to also just get Home/End for free there?
> > 
> > Downsides:
> > 
> > 1. Would have to be natively implemented for win32 and macOS. I'm familiar with the macOS half, but not the win32.
> > 2. It may end up being more complicated / change behavior onexisting clients
> > 
> > Again. Feel free to completely reject this! I propose it as an alternative if you feel like doing this in JS is too cumbersome :)
> 
> Yeah as I was looking at this I was wondering if FocusZone had Home/End handling. @PPatBoyd would you happen to know?

From testing and from an offline conversation, I don't think there is native handling on Windows. @PPatBoyd can you elaborate here?

Saadnajmi:
> > > Suggestion (feel free to reject if it doesn't work):
> > > Handle focus natively rather than JS, via FocusZone.
> > > Reasons being:
> > > 
> > > 1. We add FocusZone for macOS but not win32. Having FocusZone on both platforms would reduce platform differences. When I asked about adding FocusZone for win32, most people just kinda shrugged whether it was a good or bad idea.
> > > 2. Handling focus in JS is somewhat hard. All we really have access to is the imperative `ref.current.focus()` method. I didn't take too deep a look at the PR, but it felt like there were some additional hoops to jump through to get stuff working. Also handling keyboard events in JS is not the best either (platform differences sad..)
> > > 3. I thought about the existing use cases of FocusZone (Radiogroup, lists) and it didn't.. feel that bad to also just get Home/End for free there?
> > > 
> > > Downsides:
> > > 
> > > 1. Would have to be natively implemented for win32 and macOS. I'm familiar with the macOS half, but not the win32.
> > > 2. It may end up being more complicated / change behavior onexisting clients
> > > 
> > > Again. Feel free to completely reject this! I propose it as an alternative if you feel like doing this in JS is too cumbersome :)
> > 
> > 
> > Yeah as I was looking at this I was wondering if FocusZone had Home/End handling. @PPatBoyd would you happen to know?
> 
> From testing and from an offline conversation, I don't think there is native handling on Windows. @PPatBoyd can you elaborate here?

I didn't expect there to be native handling of Home/End for FocusZone on either platform. That would need to be something added to the native module.

lawrencewin:
> Suggestion (feel free to reject if it doesn't work):
> 
> Handle focus natively rather than JS, via FocusZone.
> 
> Reasons being:
> 
> 1. We add FocusZone for macOS but not win32. Having FocusZone on both platforms would reduce platform differences. When I asked about adding FocusZone for win32, most people just kinda shrugged whether it was a good or bad idea.
> 2. Handling focus in JS is somewhat hard. All we really have access to is the imperative `ref.current.focus()` method. I didn't take too deep a look at the PR, but it felt like there were some additional hoops to jump through to get stuff working. Also handling keyboard events in JS is not the best either (platform differences sad..)
> 3. I thought about the existing use cases of FocusZone (Radiogroup, lists) and it didn't.. feel that bad to also just get Home/End for free there?
> 
> Downsides:
> 
> 1. Would have to be natively implemented for win32 and macOS. I'm familiar with the macOS half, but not the win32.
> 2. It may end up being more complicated / change behavior onexisting clients
> 
> Again. Feel free to completely reject this! I propose it as an alternative if you feel like doing this in JS is too cumbersome :)

I'm leaning towards using this JS implementation, but if you are willing to help with a MacOS implementation of the feature FocusZone, I'm can try my hand at a windows version.

packages/components/Menu/src/MenuItem/useMenuItem.ts
rurikoaraki: I'm not sure, I think it'd be worth doing some research to see if you could confirm.

PPatBoyd:
> > From testing and from an offline conversation, I don't think there is native handling on Windows. @PPatBoyd can you elaborate here?
> 
> I didn't expect there to be native handling of Home/End for FocusZone on either platform. That would need to be something added to the native module.

So I looked into this and to answer the question that was asked: a FocusZone currently won't add new Home/End behaviors.  

That said I actually expected Home/End to work in the first place, since it's part of the native platform focus management already.  Two quick examples: The old ContextMenu and general React-Native content such as the FURN tester list of pages or the FURN tester test pages already honor Home/End correctly, skipping to the last/first focusable element.  I was surprised that MenuV1 did not do that automatically but in the time that I debugged it last week I didn't see reason for the MenuV1 to be acting differently from the FURN tester surfaces (I did see a reason for ContextMenu to be special cased, so disregarding that).

Home/End behaviors should be coming correctly from the native platform, it'd be great to better understand why it isn't working natively for MenuV1 which as I see the layout (i.e. all menu items are focusable in the tree) I would expect to work without needing JS-side intervention.

Saadnajmi:
> > > From testing and from an offline conversation, I don't think there is native handling on Windows. @PPatBoyd can you elaborate here?
> > 
> > I didn't expect there to be native handling of Home/End for FocusZone on either platform. That would need to be something added to the native module.
> 
> So I looked into this and to answer the question that was asked: a FocusZone currently won't add new Home/End behaviors.  
> 
> That said I actually expected Home/End to work in the first place, since it's part of the native platform focus management already.  Two quick examples: The old ContextMenu and general React-Native content such as the FURN tester list of pages or the FURN tester test pages already honor Home/End correctly, skipping to the last/first focusable element.  I was surprised that MenuV1 did not do that automatically but in the time that I debugged it last week I didn't see reason for the MenuV1 to be acting differently from the FURN tester surfaces (I did see a reason for ContextMenu to be special cased, so disregarding that).
> 
> Home/End behaviors should be coming correctly from the native platform, it'd be great to better understand why it isn't working natively for MenuV1 which as I see the layout (i.e. all menu items are focusable in the tree) I would expect to work without needing JS-side intervention.

Any chance Contextualmenu was getting that from the native ScrollView? I think MenuV1 doesn't _always_ render a scroll view, unlike CM.

PPatBoyd:
> Any chance Contextualmenu was getting that from the native ScrollView? I think MenuV1 doesn't _always_ render a scroll view, unlike CM.

Good call, that is where it's coming from in the examples I'd compared!

Saadnajmi:
> So I looked into this and to answer the question that was asked: a FocusZone currently won't add new Home/End behaviors.

@PPatBoyd Could you elaborate on why Native FocusZone shouldn't / won't get Home/End functionality? Mostly for my own knowledge :)

Context: I've been stuck on how to handle these particular keys for a while (mostly while working on Flatlist). 

Home/End is handled natively in win32 ScrollView but not every list is guaranteed to have a ScrollView. Perhaps we should just enforce that for MenuV1? I think doing so will cause some other side effects though. macOS handle's its equivalent hotkeys in  a _different_ native component that React Native doesn't use (NSTableView). So regardless, we have to do some special handling _somewhere_ to get it on macOS too. That's either in JS (usually complicated ☹️) or some other native module like FocusZone.

I've been having a hard time figuring out how to reconcile this particular platform difference in RN. If we want to do it natively, maybe we enforce that MenuV1 always has a ScrollView, and then FocusZone for macOS can implement those extra keys? Or maybe we just go with the original PR and do it in JS 😅

packages/components/Menu/src/MenuList/useMenuList.ts
Saadnajmi: Thoughts on defining this as a separate hook `useMountEffect` so it's got a nice semantic name? 
https://stackoverflow.com/a/56767883

Saadnajmi:
With upcoming "MenuGroup" work, there was talk that Home/End take you to the start and end of groups. Might be worth making sure this change would fit with that as well?

packages/components/Menu/src/MenuList/useMenuList.ts
rurikoaraki: sure, that seems fine to me.

rurikoaraki (approved):

Saadnajmi (approved):

Saadnajmi (approved):
(Assuming this works on macOS, I didn't test)

lawrencewin:
> With upcoming "MenuGroup" work, there was talk that Home/End take you to the start and end of groups. Might be worth making sure this change would fit with that as well?

Not too familiar with the upcoming MenuGroup work, but if the behavior is similar to sub-menus, then this should be compatible. Even so, I don't think the changes will be impossible to integrate.

lawrencewin:
Commenting to say that MacOS behavior is currently bugged. Merging as is and will find a MacOS solution later.

PR #2702, opened 2023-03-14
Make accessible prop customizable for all controls

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

A lot of our controls hardcode accessible to true, but there can be cases where people want to make accessible false. Changing our controls to have accessible be editable.

### Verification

Tested with tester

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Switch/src/useSwitch.ts
```diff
@@ -128,7 +129,7 @@ export const useSwitch = (props: SwitchProps, animationConfig?: AnimationConfig)
 
   return {
     props: {
-      accessible: isMobile ? !disabled : true,
+      accessible: isMobile ? !disabled : accessible ?? true,
```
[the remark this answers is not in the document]
rurikoaraki: Ah good catch, yeah let me go change that.

packages/components/Avatar/src/useAvatar.ts
```diff
@@ -18,6 +18,7 @@ import { titles, multiWordTitles } from './titles';
 export const useAvatar = (props: AvatarProps): AvatarInfo => {
   const {
     avatarColor,
+    accessible,
```
ecreeth, line 21: We can use default parameters instead.
```js
accessible = true
```

rurikoaraki: I'm going to leave this as is since most of our other props follow the current pattern.

(Also personally I'm not a huge fan of default props, since it makes it less obvious where the value of a variable is coming from when reading code, but that's a personal opinion).

PR #2705, opened 2023-03-15
Fix host theming not persisting across multiple instances of win32 theme

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Client had an issue where if you changed the host theme and then created a new instance of the win32 theme, it would use the wrong host theme as the basis.

The issue is that the win32 theme uses the constant initialHostThemeSetting from the ThemingModule as the host theme but this value is only read when the theming module is created (since it's a constant). If the host theme changes after the constant is created and nothing is keeping track of the new value, then new instances of the win32 theme will use the wrong host theme becayse initialHostThemeSetting will be outdated.

Solution is to keep track of the current host theme when the module is created and reroute the win32 theme to use the new state as its host theme when it's created. This is less ideal than a change to the native module which would allow us to query for the host theme, but since native changes take more time to propagate and this solution is needed quickly, this JS solution was coded instead.

### Verification

Tested client bug is fixed

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

rurikoaraki:
> nit: `getThemingModule.native.ts`: I don't think we need the `.native` bit? That's usually for distinguishing server and client code I think.

Yeah I think it's leftover from when we supported web maybe? Probably worth cleaning up but I'd rather keep that a separate change.

PR #2706, opened 2023-03-16
Apply "width" token to Switch root

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In re-testing Switch tokens to try and publish it as V1, I found that the "width" token was not being applied. This PR applies the token to the root while cleaning up the existing code. (minWidth and minHeight get applied twice - edited to be applied once)

### Verification

Tested in FURN tester. Below is the CustomizedSwitch section with width set to 500px. 

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Switch/src/Switch.styling.ts
```diff
@@ -34,24 +34,14 @@ export const stylingSettings: UseStylingOptions<SwitchProps, SwitchSlotProps, Sw
           alignItems: 'center',
           flexDirection: tokens.flexDirection,
           alignSelf: 'flex-start',
-          minHeight: tokens.minHeight,
-          minWidth: tokens.minWidth,
+          width: tokens.width,
           borderColor: tokens.focusStrokeColor,
           borderWidth: tokens.focusBorderWidth,
           borderRadius: tokens.focusBorderRadius,
           ...layoutStyles.from(tokens, theme),
```
Saadnajmi, line 41: Shouldn't these be part of `...layoutStyles` ?

lawrencewin: minWidth, and minHeight are. Width currently isn't, but it probably should be.

rurikoaraki: Yeah you can add it. It's mostly a list of tokens that have manually been added over time, there isn't really a reason why a particular token has been left out.

rurikoaraki (approved):

packages/utils/tokens/src/layout-tokens.ts
```diff
@@ -6,6 +6,7 @@ import type { OperationSet } from './token.types';
 import { tokenBuilder } from './tokenBuilder';
 
 export interface LayoutTokens {
+  width?: ViewStyle['width'];
```
Saadnajmi, line 9: should we add height as well? Also, did we confirm that no other set of tokens uses width/height?

Saadnajmi (approved):

packages/utils/tokens/src/layout-tokens.ts
Saadnajmi: OK I don't see any other token sets that use width/height. But I would prefer we _also_ add height.

PR #2724, opened 2023-03-22
[Button][win32] Only fire Button onClick with keyboard if keyDown event was received first

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The current button behavior on key press in win32 is causing issues. Specifically, there are cases where Buttons can be invoked accidentally because focus moves to the Button due to another action and then the enter or space button is released.
The native button would only fire for key presses if it detected a keyDown first, so we're going to mimic that behavior for Buttons in win32. ~Native buttons also adopt pressed coloring while keyDown, so that has been applied as well.~ Undoing this change since it's causing E2E failures

The mimicking is done by adding another state which tracks if the button is pressed via keys. This state is cleared when the key is released or when focus moves away from the button.

Another change in behavior is that key presses on win32 buttons will go through with shift or ctrl pressed, so making that behavior change for win32.

### Verification

Tested with tester against behavior of native Button

- Tested tabbing
- Tested clicking
- Tested enter invokes a Button, adds pressed coloring
- Tested pressing enter, tabbing from button1 to button2, and then releasing enter doesn't execute button2
- Tested that pressing enter, tabbing away and then back, and then releasing enter doesn't execute button
- Tested keyboard with mouse interactions (like pressing enter, clicking, and then releasing enter) align with native
- Tested invoking a button with shift & enter
- Tested space interactions
- It is expected that the button executes twice if you press enter, press down via mouse, release enter, and release mouse. This aligns with native behavior.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Button/src/useButton.ts
+++ b/packages/components/Button/src/useButton.ts
@@ -32,14 +32,14 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
   } = props;
 
   const isDisabled = !!disabled || !!loading;
-  const [isKeyPressed, setIsKeyPressed] = React.useState<boolean>(false);
+  const [isProcessingKeyboardInvocation, setIsProcessingKeyboardInvocation] = React.useState<boolean>(false);
 
   // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
   const focusRef = isDisabled ? null : componentRef;
   const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
   const onBlurInner = React.useCallback(
     (e) => {
-      setIsKeyPressed(false);
+      setIsProcessingKeyboardInvocation(false);
       onBlur?.(e);
     },
     [onBlur],
@@ -48,8 +48,8 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
 
   const onKeyDownInner = React.useCallback(
     (e) => {
-      if ((!disabled && e.nativeEvent.key === 'Enter') || e.nativeEvent.key === ' ') {
-        setIsKeyPressed(true);
+      if (!disabled && (e.nativeEvent.key === 'Enter' || e.nativeEvent.key === ' ')) {
+        setIsProcessingKeyboardInvocation(true);
       }
       onKeyDown?.(e);
     },
@@ -57,12 +57,12 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
   );
   const onKeyPress = React.useCallback(
     (e) => {
-      if (isKeyPressed) {
+      if (isProcessingKeyboardInvocation) {
         onClick?.(e);
-        setIsKeyPressed(false);
+        setIsProcessingKeyboardInvocation(false);
       }
     },
-    [isKeyPressed, onClick],
+    [isProcessingKeyboardInvocation, onClick],
   );
   const onKeyProps = useKeyProps(shouldOnlyFireIfPressed ? onKeyPress : onClick, ' ', 'Enter');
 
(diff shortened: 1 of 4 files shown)


packages/components/Button/src/useButton.ts
```diff
@@ -8,31 +8,68 @@ import { isHighContrast } from '@fluentui-react-native/theming-utils';
...
+      setIsKeyPressed(false);
+      onBlur?.(e);
+    },
+    [onBlur],
+  );
+  const pressable = usePressableState({ ...rest, onPress: onClickWithFocus, onBlur: shouldOnlyFireIfPressed ? onBlurInner : onBlur });
+
+  const onKeyDownInner = React.useCallback(
+    (e) => {
+      if ((!disabled && e.nativeEvent.key === 'Enter') || e.nativeEvent.key === ' ') {
```
Saadnajmi, line 51: Intentional that Space will always trigger, but Enter will only trigger if not disabled?

Saadnajmi (approved):
Testing on macOS made me realize it feels like there's weirdness about not clearing the pressed state there either. But it's on `main` and this branch, so probably unrelated.

packages/components/Button/src/useButton.ts
```diff
@@ -8,31 +8,68 @@ import { isHighContrast } from '@fluentui-react-native/theming-utils';
...
+    onClick,
+    onKeyDown,
     loading,
     enableFocusRing,
     focusable,
     ...rest
   } = props;
 
   const isDisabled = !!disabled || !!loading;
+  const [isKeyPressed, setIsKeyPressed] = React.useState<boolean>(false);
```
PPatBoyd, line 35: \#supernitty wording could indicate relevant functionality e.g. keyboardInvocationStarted but not a big deal

packages/components/Button/src/useButton.ts
```diff
@@ -8,31 +8,68 @@ import { isHighContrast } from '@fluentui-react-native/theming-utils';
...
+    [onBlur],
+  );
+  const pressable = usePressableState({ ...rest, onPress: onClickWithFocus, onBlur: shouldOnlyFireIfPressed ? onBlurInner : onBlur });
+
+  const onKeyDownInner = React.useCallback(
+    (e) => {
+      if ((!disabled && e.nativeEvent.key === 'Enter') || e.nativeEvent.key === ' ') {
+        setIsKeyPressed(true);
+      }
+      onKeyDown?.(e);
```
PPatBoyd, line 54: should other keydowns invalidate the pressed state?

packages/components/Button/src/useButton.ts
rurikoaraki: nope, thanks for catching.

packages/utils/interactive-hooks/src/useKeyProps.ts
```diff
@@ -5,23 +5,30 @@ import { memoize } from '@fluentui-react-native/memo-cache';
 
 import type { KeyCallback, KeyPressEvent, KeyPressProps } from './useKeyProps.types';
 
+const shouldExcludeAllModifierKeys = Platform.OS === 'macos';
+
```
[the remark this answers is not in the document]
rurikoaraki: Sure I can make it win32 only

packages/components/Button/src/useButton.ts
rurikoaraki: Native doesn't seem to do that so I don't think so.

packages/components/Button/src/useButton.ts
rurikoaraki: changed to isProcessingKeyboardInvocation

PR #2727, opened 2023-03-23
[Menu] Fix for keyboard focus is lost when hovering out of a menu

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

~~The proposed fix is to let FocusZone handle the removal of focus by calling `resignFirstResponder` on the focused element. That requires creating a tracking area for the FocusZone window and overriding `mouseExited`.~~

~~Followup: In the after video, you can see that pressing the down arrow key sets keyboard focus on the subsequent item of the last focused item, where it should be the top item (last item w/ up arrow key).~~

Edit: A better fix was suggested by @Saadnajmi, that is to set focus on `onMouseLeave`, a simple one line fix :)

### Verification

Before:
  

https://user-images.githubusercontent.com/67026167/227336554-285357c8-ebf2-41ce-ba85-747eaea79e15.mov

After:

https://user-images.githubusercontent.com/67026167/227336582-de66cc6a-478c-4790-85ca-ceea42e333ca.mov

### Pull request checklist

- [ ] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-focus-zone-bb03f1cd-8bb5-4100-a714-2ae966d3310c.json
+++ b/change/@fluentui-react-native-focus-zone-bb03f1cd-8bb5-4100-a714-2ae966d3310c.json
@@ -1,7 +1,0 @@
-{
-  "type": "patch",
-  "comment": "Fix for keyboard focus is lost when hovering out of a menu",
-  "packageName": "@fluentui-react-native/focus-zone",
-  "email": "email not defined",
-  "dependentChangeType": "patch"
-}
--- a/change/@fluentui-react-native-focus-zone-bb03f1cd-8bb5-4100-a714-2ae966d3310c.json
+++ b/change/@fluentui-react-native-focus-zone-bb03f1cd-8bb5-4100-a714-2ae966d3310c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for keyboard focus is lost when hovering out of a menu",
+  "packageName": "@fluentui-react-native/focus-zone",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -19,6 +19,9 @@ typedef BOOL (^IsViewLeadingCandidateForNextFocus)(NSView *candidateView);
 static const CGFloat FocusZoneBuffer = 3;
 
 @implementation RCTFocusZone
+{
+	NSTrackingArea *trackingArea;
+}
 
 static inline CGFloat GetDistanceBetweenPoints(NSPoint point1, NSPoint point2)
 {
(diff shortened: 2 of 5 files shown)


packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -186,6 +189,39 @@ static BOOL ShouldSkipFocusZone(NSView *view)
...
+
+- (void)mouseExited:(NSEvent *)event
+{
+	[super mouseExited:event];
+	NSView *viewWithFocus = GetFirstResponder([self window]);
+	if (viewWithFocus != nil)
+	{
+		[viewWithFocus resignFirstResponder];
+	}
+}
```
Saadnajmi, line 223: This would affect all consumers of FocusZone, not just menu right? Would this negatively impact stuff like Radiogroup?

chiuam: That's right, I did a quick test on RadioGroup and Tabs before publishing the PR, didn't find anything weird... will do some more testing and report back

harrieshin: just to be sure, lets make sure we test various edge cases we have been fixing over the years in focus zone so we aren't regressing anything.

packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -104,7 +104,11 @@ export const useHoverFocusEffect = (hovered: boolean, componentRef: React.Mutabl
     if (hovered) {
       componentRef?.current?.focus();
     } else {
-      componentRef?.current?.blur();
+      // Since MenuList uses FocusZone on macOS, we need to handle the .blur() logic there in order to
```
harrieshin, line 107: for my own understanding? why does blur logic interfere with keyboard navigation?

Saadnajmi: When we blur on mouseExit, focus is lost from the menu overall. Then, the user can't ArrowUp/Down back between menu items until they put their mouse back over the menu.

packages/components/FocusZone/macos/RCTFocusZone.m
chiuam: Debugged Tabs and RadioGroup in Xcode, I saw that `[viewWithFocus resignFirstResponder];` does get called on mouseExited. I think this is fine because it only removes the highlight, not the focus ring that those controls use.

source: https://developer.apple.com/documentation/uikit/uiresponder/1621097-resignfirstresponder

harrieshin: that link is for uikit though. not app kit

packages/components/Menu/src/MenuItem/useMenuItem.ts
harrieshin: can we make sure we capture that in comment and the pr description please?

harrieshin (approved):
sign off with comments

packages/components/FocusZone/macos/RCTFocusZone.m
harrieshin: that is a good point. in the radio group, if it removes the highlight? what is focused? then?

chiuam: Scratch all that, a much simpler fix is to set focus on the menu container when onMouseLeave is called. Sorry for the noise :/

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -83,7 +83,7 @@ export const MenuList = compose<MenuListType>({
 
       const content =
         Platform.OS === 'macos' ? (
-          <Slots.root>
+          <Slots.root onMouseLeave={() => focusZoneRef.current.focus()}>
```
Saadnajmi, line 86: Might be good to extract this to a a separate function / React.useCallback because we try to not pass lamba functions directly in jsx.

Saadnajmi (approved):

change/@fluentui-react-native-focus-zone-bb03f1cd-8bb5-4100-a714-2ae966d3310c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix for keyboard focus is lost when hovering out of a menu",
+  "packageName": "@fluentui-react-native/focus-zone",
```
Saadnajmi, line 4: We can delete this file?

Saadnajmi (approved):

Saadnajmi (approved):

PR #2731, opened 2023-03-24
Properly render switch on/off text when `labelPosition` prop = "before"

lawrencewin:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Change made in response to feedback from partners. 

Up until now, the `onText` and `offText` props were mutually exclusive with the `labelPosition` prop being set to = "before". This is because the Switch position would horizontally shift if the sizes of the on / off texts were uneven, which in most cases are. 

This commit changes how `onText` and `offText` are rendered by rendering both texts at the same time but hiding the inactive text via height = 0. This allows the on / off text to maintain the same width no matter which text is showing, removing the shifting switch issue. 

TODO: update documentation

### Verification

The videos below show the before / after behaviors on win32. Testing on other platforms would be appreciated. 

| https://user-images.githubusercontent.com/15683103/227612668-04bb4196-3990-4319-a2da-209d6f432848.mp4 |  https://user-images.githubusercontent.com/15683103/227612862-9c2142f2-dc01-4fe1-a019-216edb87335e.mp4 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Switch/src/SwitchTokens.ts
+++ b/packages/components/Switch/src/SwitchTokens.ts
@@ -22,8 +22,6 @@ export const defaultSwitchTokens: TokenSettings<SwitchTokens, Theme> = (t: Theme
   trackMarginLeft: 2,
   trackMarginRight: 2,
   thumbMargin: 2,
-  onOpacity: 0,
-  offOpacity: 0,
 
   beforeContent: {
     trackMarginLeft: 8,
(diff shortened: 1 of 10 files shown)


packages/components/Switch/src/SwitchTokens.ts
```diff
@@ -22,6 +22,8 @@ export const defaultSwitchTokens: TokenSettings<SwitchTokens, Theme> = (t: Theme
   trackMarginLeft: 2,
   trackMarginRight: 2,
   thumbMargin: 2,
+  onOpacity: 0,
```
PPatBoyd, line 25: talked offline about using absolute positioning to stack the two texts and use opacity, avoid layout change possibilities from the height changing

PPatBoyd (changes requested):

PPatBoyd (approved):

PR #2747, opened 2023-04-03
<Spinner win32> - Add TrackSvg and implement platform specific design

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- Added win32 specific props and tokens, and move sharables to newly created shared files
- Added win32 specific tests for track only (tail and animation not supported yet)
- Updated SPEC file to include win32 usage.

### Verification
spinner just track

### Pull request checklist

- [X] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Spinner/src/Spinner.tsx
```diff
@@ -1,35 +1,26 @@
...
 import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
-import { TextV1 as Text } from '@fluentui-react-native/text';
 import { Svg } from 'react-native-svg';
 
-import { RCTNativeAnimatedSpinner } from './consts.win32';
 import type { SpinnerProps, SpinnerType } from './Spinner.types';
 import { spinnerName } from './Spinner.types';
 
-/* TODO: Implement Spinner with following slots */
+export const AnimatedSvg = Animated.createAnimatedComponent(Svg);
```
rurikoaraki, line 11: >Animated.createAnimatedComponent [](http://example.com/codeflow?start=27&length=32)

ANimated doesn't work for win32, I thought?

packages/experimental/Spinner/src/Spinner.styling.win32.ts
```diff
@@ -0,0 +1,62 @@
+import type { UseStylingOptions, Variant } from '@fluentui-react-native/framework';
+import { buildProps } from '@fluentui-react-native/framework';
+
+import type { SpinnerProps, SpinnerSlotProps, SpinnerTokens } from './Spinner.types.win32';
+import { spinnerName } from './Spinner.types.win32';
+import { defaultSpinnerTokens } from './SpinnerTokens';
+
+export const diameterSizeMap: { [key: string]: number } = {
```
rurikoaraki, line 8: As noted in the comment on SpinnerTokens, this would usually be done in the tokens file and styling would map tokens to style prop entries.

packages/experimental/Spinner/src/Spinner.styling.win32.ts
```diff
@@ -0,0 +1,62 @@
...
+  huge: 'subtitle1',
+};
+
+export const stylingSettings: UseStylingOptions<SpinnerProps, SpinnerSlotProps, SpinnerTokens> = {
+  tokens: [defaultSpinnerTokens, spinnerName],
+  tokensThatAreAlsoProps: 'all',
+  slotProps: {
+    root: buildProps(
+      (tokens: SpinnerTokens) => ({
+        accessibilityRole: 'progressbar',
```
rurikoaraki, line 43: >ccessibilityRole: 'progressbar' [](http://example.com/codeflow?start=9&length=31)

nit: Since this is a styling file I would put non-styling props in a useSpinner hook or a different place.

packages/experimental/Spinner/src/Spinner.win32.tsx
```diff
@@ -0,0 +1,67 @@
...
+    label: Text,
+  },
+  useRender: (props: SpinnerProps, useSlots: UseSlots<SpinnerType>) => {
+    const Slots = useSlots(props, (layer) => spinnerLookup(layer, props));
+
+    return (rest: SpinnerProps) => {
+      const { ...mergedProps } = mergeProps(props, rest);
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.track viewBoxHeight={diameterSizeMap[props.size]} viewBoxWidth={diameterSizeMap[props.size]} />
```
rurikoaraki, line 62: >viewBoxHeight={diameterSizeMap[props.size]} viewBoxWidth={diameterSizeMap[props.size]} [](http://example.com/codeflow?start=23&length=86)

I think these could be set in the styling file

packages/experimental/Spinner/src/Spinner.tsx
lenahong: Yeah so I forked Spinner.win32 which doesn't use the Animated library. This file is being restored for other platforms

rurikoaraki: Whoops my bad. Thanks

packages/experimental/Spinner/src/Spinner.styling.win32.ts
```diff
@@ -1,47 +1,16 @@
...
 export const stylingSettings: UseStylingOptions<SpinnerProps, SpinnerSlotProps, SpinnerTokens> = {
   tokens: [defaultSpinnerTokens, spinnerName],
   tokensThatAreAlsoProps: 'all',
   slotProps: {
     root: buildProps(
       (tokens: SpinnerTokens) => ({
-        accessibilityRole: 'progressbar',
-        accessible: true,
         style: {
           width: diameterSizeMap[tokens.size],
```
rurikoaraki, line 15: For other controls we've set the width token to the output of diameterSizeMap in the Tokens file and then accessed tokens.width here. This allows the user to specify a particular width if they want to. Can we follow that pattern?

lenahong: So I'm a little confused on how I would do this. Do you mean that I define width and height tokens under SpinnerTokens, and set their default values based on size in the Tokens file? But you mentioned that size is not usually set as component tokens themselves, so size isn't accessible in the tokens file

rurikoaraki: So the pattern for other components has been to set size as a state for a component, which we would then set defaults for in the tokens file. For example state would be something like "small, medium, large", and then in the tokens file it would do something like:

"small": {
    width: foo1, height: foo2,
},
"medium" :{
    width: foo3, height: foo4,
}

etc.

rurikoaraki: Badge and Avatar should have examples.

lenahong: Gotcha! Thanks :)

PR #2757, opened 2023-04-11
[Radio] A11y fix - Make radio announce unselected on Android 

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

On Android, unselected state of the radio is not announced. Added that. 

### Verification

Manual verification of talkback.

### Pull request checklist

- [x] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/experimental/RadioGroup/src/Radio/useRadio.ts
```diff
@@ -97,7 +99,7 @@ export const useRadio = (props: RadioProps): RadioInfo => {
       ref: buttonRef,
       ...pressable.props,
       accessibilityRole: 'radio',
-      accessibilityLabel: accessibilityLabel ?? label,
+      accessibilityLabel: (shouldAnnounceUnselected && state.selected ? '' : 'unselected') + (accessibilityLabel ?? label),
```
rurikoaraki, line 102: This won't work outside of English languages. We don't localize strings in FURN, so either this will have to be something done by the client, or you'll need to pass this information via a prop which will localize the info on the native side.

ankraj12: If the behaviour can be achieved by clients using current "accessibilityLabel" to pass "unselected" a prefix in the label, that should suffice I believe. Having a prop. like "`unselectedLabel`" and "`accessibilityLabel`" is probably avoidable in that case.

ayush547: Updating Android tester with a sample of how this can be used

change/@fluentui-react-native-tester-f6c032cd-7173-4bf1-aa7e-23fc5400b4f0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
ankraj12, line 2: typo

ankraj12 (approved):

PR #2778, opened 2023-04-19
[Menu] Ensure menu and submenus have same pointer capture logic

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Bug: On a menu that opens via click that has a submenu that opens on hover with a scroll bar, the submenu dismisses early when dragging the scrollbar.

Having different pointer capture logic for both the parent menu and submenu causes a bug where we're dismissing the submenu earlier than we want. More specifically, when we have a parent menu that opens via click and a submenu that opens via hover, the parent menu takes pointer capture (by default on win32) and the submenu is set to not take pointer capture since it opens on hover. I was not able to determine the exact order of events that causes the submenu to dismiss early when clicking on the scrollbar but the issue is resolved when the parent menu and submenu share the same pointer capture logic.

To do this, I added `parentDoNotTakePointerCapture` to the menu context so child menus can set their pointer capture logic accordingly. I also had to ensure we cleared out the hoverOut timers when a submenu that opens on hover is opened while we're still hovered on the menu trigger to prevent the submenu from flickering open and closed when the menus take pointer capture.

### Verification

Added an extra example case that has a menu with a submenu that opens on hover with a scrollview and manually tested the changes to verify the issue was fixed.

|SubmenuDismissBug|SubmenuDissmisFix|

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
@@ -192,7 +192,7 @@ const SubmenuWithScrollView: React.FunctionComponent<MenuProps> = (props: MenuPr
 const MenuSubMenuWithScrollView: React.FunctionComponent = () => {
   return (
     <Stack style={stackStyle}>
-      <Menu openOnHover>
+      <Menu>
         <MenuTrigger>
           <Button>Test</Button>
         </MenuTrigger>
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
@@ -160,6 +160,53 @@ const MenuSubMenu: React.FunctionComponent = () => {
   );
 };
 
+const SubmenuWithScrollView: React.FunctionComponent<MenuProps> = (props: MenuProps) => {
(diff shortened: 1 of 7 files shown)


PPatBoyd (approved):

ksiler:
Just found a bug where the submenu won't dismiss if I quickly hover over another menu item before the submenu shows up. Will update with a fix soon.

ksiler:
The issue is I'm calling `clearTimeout(triggerHoverOutTimer);` when we place focus in the submenu. This is to avoid another bug where the submenu will open/close repeatedly since the menu trigger is receiving a call to onHoverOut as a result of the submenu taking focus when it opens and then returning focus to the menu trigger when it is closed.  When mouse hover is quickly moved to another menu item in the parent menu, onHoverOut is called on the menu trigger but is cancelled when the submenu shows up by `clearTimeout(triggerHoverOutTimer);`.

ksiler:
I'm still noticing some buggy behavior due to these changes so closing this PR for now and will open up another PR that's less risky.

PR #2779, opened 2023-04-19
Create new package for `TabList` component

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR copies the current contents of the experimental Tabs component into its own package, TabList. This package will be iterated on with changes to align it with the web TabList component. The main changes involve changing names of types to reflect the component's new name. 

### Verification

Pipeline passes local tests successfully. Note that jest tests for the TabList are commented out. This will be fixed later when the component API is updated. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/TabList/CHANGELOG.json
gulnazsayed, line 1: nit: is there supposed to be a changelog description in this file?

packages/experimental/TabList/CHANGELOG.md
gulnazsayed, line 1: nit: same thing as the other changelog file

packages/experimental/TabList/CHANGELOG.json
lawrencewin: I removed the changelog description since this is a new package separate from the experimental Tabs component. To me, it doesn't make sense for there to be an existing log of changes for this, but let me know if you disagree.

packages/experimental/TabList/CHANGELOG.md
lawrencewin: Same as above.

packages/experimental/TabList/CHANGELOG.json
gulnazsayed: I think it should be fine either way, I usually just put something like "new TabList package."

packages/experimental/TabList/src/TabListTokens.macos.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { TokenSettings } from '@fluentui-react-native/use-styling';
+
+import type { TabListTokens } from '.';
+
+export const defaultTabListTokens: TokenSettings<TabListTokens, Theme> = (t: Theme) =>
+  ({
+    color: t.colors.menuItemText,
+    variant: 'bodySemibold',
+    fontWeight: 'bold',
+    fontSize: 14,
```
gulnazsayed, line 11: Can you use a token here?

packages/experimental/TabList/src/__tests__/TabList.test.tsx
```diff
@@ -0,0 +1,83 @@
+// import * as React from 'react';
```
gulnazsayed, line 1: Will you go back and un-comment this later?

packages/experimental/TabList/src/TabListTokens.macos.ts
lawrencewin: For the text styling? I do plan on updating tokens to be in line with web and using tokens here, but I was hoping to keep this PR limited to simply creating the package.

packages/experimental/TabList/src/__tests__/TabList.test.tsx
lawrencewin: Yes. For now, I just need the package to build.

packages/experimental/TabList/CHANGELOG.json
lawrencewin: I have "add TabList package" in my change folder. That should get applied later by beachball, correct?

packages/experimental/TabList/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@fluentui-react-native/tablist",
```
samuelfreiberg, line 2: Should this contain "experimental" anywhere? Not sure, genuine question. What do other experimental controls do?

packages/experimental/TabList/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "version": "0.1.0",
```
samuelfreiberg, line 3: I think this should start at "1.0.0" and go from there

Saadnajmi: I think if this is not labelled experimental in its path/package name, then versioning it 1.0.0 means we have committed to the API. If we expect more iteration before it's V1 ready, I think it may be better to keep it as is.

packages/experimental/TabList/package.json
rurikoaraki: We've used the experimental name in the past to avoid name clashes with existing packages while we rebuilt the control with the new framework. You don't need to add it to the package name if it's not going to clash with anything.

packages/experimental/TabList/src/Tab.types.ts
```diff
@@ -0,0 +1,162 @@
...
+   * The icon color.
+   */
+  iconColor?: string;
+
+  /**
+   * Source URL or name of the icon to show on the TabsItem.
+   */
+  icon?: IconSourcesType;
+
+  /**
```
lenahong, line 38: I don't headerText should be a token. We generally want to only set styling props as tokens

packages/experimental/TabList/src/TabList.windows.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
+
+import { stylingSettings } from './TabList.styling';
+import type { TabListType, TabListProps, TabListContextData } from './TabList.types';
+import { tabListName } from './TabList.types';
+import { useTabList } from './useTabList';
+
+export const TabListContext = React.createContext<TabListContextData>({
+  selectedKey: null,
+  onTabsClick: (/* key: string */) => {
```
lenahong, line 15: remove unused parameter?

packages/experimental/TabList/src/TabListTokens.macos.ts
lenahong: For next time, you should separate out prs for defining prop and tokens types, and for setting specific platform style tokens, instead of hard coding it like this and updating it later. It's helpful for when we have to look back at each pr

packages/experimental/TabList/src/__tests__/TabList.test.tsx
lenahong: We shouldn't include entirely commented out text files into prs. Remove them and add a simple test that will build with this pr

PR #2792, opened 2023-04-22
Don't bundle in JS PR

Saadnajmi:
Our JS PR has been taking a while. We can probably skip bundling, since we do that in each platform check. 
Feel free to reject this PR if it's not one we like. I was curious how much time it saved us.


Saadnajmi:
From 52-59 minutes down to 31 minutes ._.

rurikoaraki (approved):

PR #2797, opened 2023-04-24
[Avatar] A11y talkback fixes enabling status announcements 

rohanprasadofficial:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Talkback verifications . 

Approach : 
Name , Status , Out of Office.

if name is passed, we annouce Name else we annouce image accessibilityRole.
if the avatar has out of office prop , then we also annouce out of office.
If avatar is active then we annouce , active else inactive 
if status is given ,it announces the presencebadge status 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


packages/components/Avatar/src/useAvatar.ts
```diff
@@ -38,16 +38,16 @@ export const useAvatar = (props: AvatarProps): AvatarInfo => {
 
   const showRing = active === 'active' && activeAppearance === 'ring';
   const showBadge = (!active || active === 'unset') && !!badge && !!badge.status;
-  const accessibilityText = `${name || ''}${showBadge ? `, ${badge.status}` : ''}`;
-
+  const accessibilityText = `${accessibilityLabel ? `${accessibilityLabel}` : undefined || `${name || ''}`}
+   ${showBadge ? `, ${badge.status}` : ''}${active != undefined ? active : ''} ${badge?.outOfOffice ? 'Out of Office' : ''} `;
```
Saadnajmi, line 42: I think this is starting concatenation (sorry, my brain is slow right now), which I think wouldn't be good for localization, because it assumes the structure of what you want to read out (maybe some language doesn't like space as the delineator, etc)

packages/components/Avatar/src/useAvatar.ts
```diff
@@ -38,16 +38,16 @@ export const useAvatar = (props: AvatarProps): AvatarInfo => {
 
   const showRing = active === 'active' && activeAppearance === 'ring';
   const showBadge = (!active || active === 'unset') && !!badge && !!badge.status;
-  const accessibilityText = `${name || ''}${showBadge ? `, ${badge.status}` : ''}`;
-
+  const accessibilityText = `${accessibilityLabel ? `${accessibilityLabel}` : undefined || `${name || ''}`}
+   ${showBadge ? `,${badge.status}` : ''}${active != undefined ? active : ''} ${badge?.outOfOffice ? 'Out of Office' : ''}`;
```
rurikoaraki, line 42: This would only work for English. We don't support localized stings in FURN, so you'll have to either have the client feed in this information or use a native prop which will localize the information for you.

packages/components/Avatar/src/useAvatar.ts
```diff
@@ -80,8 +80,8 @@ export const useAvatar = (props: AvatarProps): AvatarInfo => {
   return {
     props: {
       accessible: accessible ?? true,
-      accessibilityLabel: accessibilityLabel || accessibilityText,
-      accessibilityRole: accessibilityRole ?? 'image',
+      accessibilityLabel: accessibilityText,
+      accessibilityRole: accessibilityRole ?? (name ? 'none' : 'image'),
```
rurikoaraki, line 84: This is not a good change for win32. The Avatar will be treated as an image with a name prop on win32, especially since it's possible that someone can pass in a name and an image as props.

packages/components/Avatar/src/useAvatar.ts
Saadnajmi: Yeah this is one of the reasons we still have the iOS native native Avatar around :/ 
We either have to set up some sort of localization pipeline for FURN or instruct clients to localize this themselves. I hadn't found a way to concatenate info in an accessibility label that's internationalization friendly

rohanprasadofficial: Totally, make sense. 
i think the issue is across many of our components, We need to have some structure how we should handle accessibility labels and localisation.

One challenge i see here (may exist with other components as well) is sometime we map the component to types support 'active' , 'inactive' , or any other property we used from types.

rurikoaraki: I think in the short term the guidance we give is that clients feed in the full string. A longer term solution should be worked on, but we don't have bandwidth to support a proper solution for at least the next few months.

packages/components/Avatar/src/useAvatar.ts
rohanprasadofficial: ok , on Android we're trying to match native behaviours, so this would be required here with platform check.

packages/components/Avatar/src/useAvatar.ts
```diff
@@ -38,16 +38,16 @@ export const useAvatar = (props: AvatarProps): AvatarInfo => {
...
+   ${showBadge ? `,${badge.status}` : ''}${active != undefined ? active : ''} ${badge?.outOfOffice ? 'Out of Office' : ''}`;
   const state: AvatarState = {
     showRing,
     transparentRing: !!transparentRing,
     showBadge,
   };
 
   let imageProps: ImageProps = {
-    accessibilityLabel,
+    accessibilityLabel: accessibilityLabel != undefined ? accessibilityLabel : '',
```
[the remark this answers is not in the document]
rohanprasadofficial: added.

rohanprasadofficial:
PR Update: 
We have decided to go with the only changes that don't need localization & feedback on this PR.
Reverted changes where localization is required for now and we can have proper solution for localization as per team feedbacks. 

Currently changes have android tester app changes and minor fixes on avatar core code. 

cc: @ankraj12

ankraj12 (approved):

rurikoaraki (approved):

packages/components/Avatar/src/useAvatar.ts
```diff
@@ -39,15 +39,14 @@ export const useAvatar = (props: AvatarProps): AvatarInfo => {
...
-
   const state: AvatarState = {
     showRing,
     transparentRing: !!transparentRing,
     showBadge,
   };
 
   let imageProps: ImageProps = {
-    accessibilityLabel,
+    accessibilityLabel: accessibilityLabel !== undefined ? accessibilityLabel : '',
```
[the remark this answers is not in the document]
rohanprasadofficial: It announces undefined , so added a check. 

added a shorthand operator.

PR #2798, opened 2023-04-24
Conditionally set enableFocusRing when hovered and keyboarded focused.

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes
- This pr is to address the bug where the user keyboard navigates to an element that is already hovered.  Instead of last action winning (since last action is keyboard focus, we expect both highlight and focus indicators upon keyboarding to it), our previous implementation was preventing the focus indicators from showing when it's already hovered.
- I added focused state checks in onHoverIn and onFocus to set enableFocusRing to true when focus state is triggered in onFocus callback, which indicates that when the item is keyboard focused. I set enableFocusRing to false when focus state is set onHoverIn.

### Verification

https://user-images.githubusercontent.com/12578599/233988804-a0f750da-5622-4717-a84d-def28cec40cf.mp4

### Pull request checklist

- [X] Automated Tests
- [X] Keyboard Accessibility
- [ ] Voiceover


lenahong:
> What's preventing the focus indicator from showing when keyboard focus moves? Does `usePressableState` still report `focused` as a true state? If it's a tokens thing, would something like nested tokens also work? I.E:
> 
> ```
> // MenuItemTokens: {
> hovered : {
>   focused: {
>     ...
>   }
> }
> ```

Styling on nested tokens would not fix this specific scenario. This might work if we're trying to style only the hover focused, but the issue we're addressing is trying to recognize keyboard focus. The tricky thing is, focus state in RN cannot differentiate mouse and keyboard focus. Our keyboard focus indicator is actually drawn on the native side. On the JS side, we just set the enableFocusRing prop in order to tell native to show it. We originallly had enableFocusRing set to !pressed.state.hover but this was neglecting the case when the mouse remains on the previously hovered focus item and keyboard focus comes back to it.

Saadnajmi:
> > What's preventing the focus indicator from showing when keyboard focus moves? Does `usePressableState` still report `focused` as a true state? If it's a tokens thing, would something like nested tokens also work? I.E:
> > 
> > ```
> > // MenuItemTokens: {
> > hovered : {
> >   focused: {
> >     ...
> >   }
> > }
> > ```
> 
> Styling on nested tokens would not fix this specific scenario. This might work if we're trying to style only the hover focused, but the issue we're addressing is trying to recognize keyboard focus. The tricky thing is, focus state in RN cannot differentiate mouse and keyboard focus. Our keyboard focus indicator is actually drawn on the native side. On the JS side, we just set the enableFocusRing prop in order to tell native to show it. We originallly had enableFocusRing set to !pressed.state.hover but this was neglecting the case when the mouse remains on the previously hovered focus item and keyboard focus comes back to it. 

Ah.. the whole win32  mouse hover vs mouse focus thing that I don't quite understand. I think I won't be helpful for this change then 😅.

rurikoaraki (approved):

PR #2801, opened 2023-04-25
[Menu] Implement N of M on submenu trigger

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Re-implements N of M on submenu triggers that was previously reverted by #2725. This implementation does not cause any rendering issues that were present with the previous implementation and implements it in a better way. The n of m properties can be overridden by passing n of m props to the submenu (as demonstrated in the N of M example in the test app).

### Verification

Added examples to the test app that demonstrate default n of m properties are set on submenu triggers and that those props can be overridden. Also added test cases that show that is works for customized menu triggers.

Default N of M set on submenu trigger:

Screenshot 2023-04-27 at 2 19 24 PM

Overridden N of M values set on submenu trigger:

Screenshot 2023-04-27 at 2 18 38 PM

Without Change:

Screenshot 2023-04-27 at 2 24 15 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
+++ b/packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
@@ -76,21 +76,19 @@ export const useMenuTrigger = (childProps: MenuTriggerChildProps): MenuTriggerSt
   );
 
   const accessibilityPositionInSet = React.useMemo(() => {
-    const baseAccessibilityPositionInSet = context.accessibilityPositionInSet;
     if (childAccessibilityPositionInSet) {
       return childAccessibilityPositionInSet;
     }
 
-    return baseAccessibilityPositionInSet;
+    return context.accessibilityPositionInSet;
   }, [childAccessibilityPositionInSet, context.accessibilityPositionInSet]);
 
   const accessibilitySetSize = React.useMemo(() => {
-    const baseAccessibilitySetSize = context.accessibilitySetSize;
     if (childAccessibilitySetSize) {
       return childAccessibilitySetSize;
     }
 
-    return baseAccessibilitySetSize;
+    return context.accessibilitySetSize;
   }, [childAccessibilitySetSize, context.accessibilitySetSize]);
 
   const onHoverIn = React.useCallback(
--- a/packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
+++ b/packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
@@ -75,22 +75,6 @@ export const useMenuTrigger = (childProps: MenuTriggerChildProps): MenuTriggerSt
     [childOnAccessibilityAction, setOpen],
   );
 
-  const accessibilityPositionInSet = React.useMemo(() => {
-    if (childAccessibilityPositionInSet) {
-      return childAccessibilityPositionInSet;
-    }
-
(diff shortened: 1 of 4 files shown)


packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -73,6 +75,24 @@ export const useMenuTrigger = (childProps: MenuTriggerChildProps): MenuTriggerSt
     [childOnAccessibilityAction, setOpen],
   );
 
+  const accessibilityPositionInSet = React.useMemo(() => {
+    const baseAccessibilityPositionInSet = context.accessibilityPositionInSet;
```
samuelfreiberg, line 79: Recommend moving this line after the `if (childAccessibilityPositionInSet) {` statement. No need to initialize this variable here if it's not going to be used.

packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -73,6 +75,24 @@ export const useMenuTrigger = (childProps: MenuTriggerChildProps): MenuTriggerSt
...
+    if (childAccessibilityPositionInSet) {
+      return childAccessibilityPositionInSet;
+    }
+
+    return baseAccessibilityPositionInSet;
+  }, [childAccessibilityPositionInSet, context.accessibilityPositionInSet]);
+
+  const accessibilitySetSize = React.useMemo(() => {
+    const baseAccessibilitySetSize = context.accessibilitySetSize;
```
samuelfreiberg, line 88: Same here

packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
ksiler: Thanks, I updated it to return the value from the context instead of setting a value and returning it. Same for the other comment.

packages/components/Menu/src/MenuTrigger/useMenuTrigger.ts
```diff
@@ -73,6 +75,22 @@ export const useMenuTrigger = (childProps: MenuTriggerChildProps): MenuTriggerSt
     [childOnAccessibilityAction, setOpen],
   );
 
+  const accessibilityPositionInSet = React.useMemo(() => {
```
rurikoaraki, line 78: I don;t think this needs to be memo'ed. Comparing primitives to determine if a component should rerender is generally safe. It's objects that need to be memoized.

rurikoaraki (approved):

Saadnajmi:
Not blocking, but I noticed this line in the web menu storyboard:
> Don't render focusable or clickable items inside menu items.

It looks like this PR is adding a button, should we use some other kind of trigger?
Note: this guidance in web is new to me, so I'm not sure if we want to / plan to follow it for RN

https://react.fluentui.dev/?path=/docs/components-menu-menu--default

ksiler:
Thanks for the catch @Saadnajmi. I just chose to render a button arbitrarily but the guidance makes sense to me and we should adhere to best practices so I'm happy to change it.

PR #2805, opened 2023-04-26
removed other layouts tests for win32

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Removed "other" layouts (horizontal and horizontal-stacked) from win32 tester app. Added a comment in the spec and to the layout and labelPosition props that these layouts are not supported in win32. We decided that these "other" layouts are not really seen in win32 so it is better not to expose these layouts to partners until a use case is identified specifically for the win32 platform. The logic and props are left as is for other platforms to use and in the case we want to make it available for win32 in the future.

### Verification

Manually tested that the horizontal and horizontal-stacked layout tests are removed from the win32 tester app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/RadioGroupV1/RadioGroupV1Test.tsx
+++ b/apps/fluent-tester/src/TestComponents/RadioGroupV1/RadioGroupV1Test.tsx
@@ -30,6 +30,13 @@ const radioGroupV1Sections: TestSection[] = [
   },
   ...Platform.select({
     android: [null],
+    native: [
+      {
+        name: 'RadioGroup with Label Subtext',
+        testID: RADIOGROUPV1_TESTPAGE,
+        component: SubtextRadioGroup,
+      },
+    ],
     default: [
       {
         name: 'RadioGroup with Label Subtext',
--- a/packages/experimental/RadioGroup/SPEC.md
+++ b/packages/experimental/RadioGroup/SPEC.md
@@ -281,7 +281,7 @@ export interface RadioGroupProps extends Pick<FocusZoneProps, 'isCircularNavigat
   /**
    * The position of the label relative to the indicator.
    *
-   * 'horizontal' and 'horizontal-stacked' are not supported from Fluent Android, renders as-is.
+   * 'horizontal' and 'horizontal-stacked' are not supported from Fluent Android and Win32, renders as-is.
    *
    * @default vertical
    */
@@ -324,7 +324,7 @@ export interface RadioProps extends PressablePropsExtended {
    * This defaults to 'after' unless the Radio is inside a RadioGroup with layout horizontal-stacked,
    * in which case it defaults to 'below'
    *
-   * 'below' is not supported from Fluent Android, renders as-is.
+   * 'below' is not supported from Fluent Android and Win32, renders as-is.
    * @default after
    */
   labelPosition?: 'after' | 'below';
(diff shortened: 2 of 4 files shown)


rurikoaraki:
Are you also removing the horizontal layout capability from the radio group? This change doesn't seem to restrict anyone from using the props.

gulnazsayed:
> Are you also removing the horizontal layout capability from the radio group? This change doesn't seem to restrict anyone from using the props.

We want to keep the props so that the other platforms can use them and/or we can add the functionality back easily if we identify a use case later down the line. Android doesn't support these layouts either, which seems to only be enforced by the props' comments and not exposing it in the platform's tester app. 

In the latest commit, I added a check for win32 to enforce that the layout and labelPosition props are always the default. I added one for android in useRadioGroup as well.

rurikoaraki:
> 

We're only shipping this so far for Android and win32 right? My opinion would be that it's better to take them out and let someone add them back in if they need it for another platform. It would still be an extra code to have to upkeep if we leave it in even if nobody uses it right now.

But if you want to leave it in, the way to enforce it in win32/android would be to probably force the layer to always resolve to a different option if one of the horizontal layouts is picked. If it's not supported on the win32 or android platforms we shouldn't be letting the control resolve to unsupported layouts.

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
```diff
@@ -100,7 +102,7 @@ export const useRadioGroup = (props: RadioGroupProps): RadioGroupInfo => {
     props: {
       ...props,
       required,
-      layout,
+      layout: layout,
```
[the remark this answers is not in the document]
gulnazsayed: Fixed and confirmed it works as expected

packages/experimental/RadioGroup/src/Radio/useRadio.win32.ts
```diff
@@ -29,7 +29,7 @@ export const useRadio = (props: RadioProps): RadioInfo => {
     subtext,
     value,
     disabled,
-    labelPosition = radioGroupContext.layout === 'horizontal-stacked' ? 'below' : 'after',
+    labelPosition = 'after', // disables labelPosition in win32
```
rurikoaraki, line 32: I think this should be a hardcoded const. If we take it from props someone can override it by passing a different value

gulnazsayed: Fixed, thanks for the comment. We are shipping it only on win32 and android for now, but I decided to keep the horizontal layout code so we can easily re-enable it later if needed to align with Fluent. I changed the fix to force the layers to resolve the horizontal layouts to the default vertical, like you suggested.

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Platform } from 'react-native';
+// import { Platform } from 'react-native';
```
rurikoaraki: nit: remove?

rurikoaraki (approved):

packages/experimental/RadioGroup/src/RadioGroup/useRadioGroup.ts
gulnazsayed: Removed!

PR #2821, opened 2023-05-05
[Draft] Default theme: if no theme is provided to iOS, use actual iOS theme instead of the web default theme

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The main change in this file is the first commit - i.e. if no theme is provided, return the apple theme. 

The problem with this is that now any jest tests that are configured to run on iOS start to fail, because createAppleTheme is now being run. One of the problems in createAppleTheme is the call to the recent AppearanceAdditions native module. Some refactoring to the native module was needed in order to be able to mock these calls properly. 

Currently there are three test suites configured to run on iOS. The Shadow and Notification test suites now pass in this PR, but the Button one is running into issues. Since the Button test suite includes tests that run on other platforms, it 1) doesn't really make much more sense to make all the Button tests use iOS theme over the default web theme, and 2) i run into a lot of errors in Button when trying to so.

I'm trying to find a way for the Button tests to continue to use the default theme, but am pretty stuck with this. 

### Verification

- Tested that appearanceAdditions still work as expected
- PR checks will fail until a way to fix the Button tests is found

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Shadow/src/__tests__/Shadow.test.tsx
```diff
@@ -54,12 +54,18 @@ const TestShadowOnChildViewWithProps: React.FunctionComponent<ShadowOnChildViewW
   );
 };
 
+jest.mock('@fluentui-react-native/experimental-appearance-additions', () => ({
+  useUserInterfaceLevel: jest.fn(),
+  useHorizontalSizeClass: jest.fn(),
+}));
+
```
lyzhan7, line 61: I don't know why - but this mock just wouldn't work unless I put it outside the describe block. I tried putting it in the beforeAll section, also tried putting it directly into each test. The only way it worked was when I took it out.

lyzhan7: Also not sure why - I first tried to do useUserInterfaceLevel: 'base' but got an error saying that the userInterfaceLevel call wasn't a function. Will try making the jest.fn() return 'base' but doesn't seem to matter for this test

PR #2822, opened 2023-05-05
Update iOS menu picker to use @react-native-menu/menu instead of @react-native-picker/picker

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Replaced usage of @react-native-picker/picker with @react-native-menu/menu.

@react-native-picker/picker was not very usable in iOS. It took up a lot more space than necessary, making it difficult to scroll down any pages that had a lot of pickers. In dark mode the text would not show up.

@react-native-menu/menu was already being used for the iOS/android theme picker. Was able to use this to make the picker look more like the macOS dropdown component. 

Will be easier to review commit by commit.

### Verification

https://user-images.githubusercontent.com/78454019/236397204-8de381b8-eb62-4f90-8021-eb59d1b7b144.mov

https://user-images.githubusercontent.com/78454019/236397215-2ee70262-84ec-4bca-94b5-50da8dc47867.mov

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Common/MenuPicker.tsx
+++ b/apps/fluent-tester/src/TestComponents/Common/MenuPicker.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 
-import { Picker } from '@react-native-picker/picker';
+import { ButtonV1 as Button } from '@fluentui-react-native/button';
+import { Text } from '@fluentui-react-native/text';
+import { MenuView } from '@react-native-menu/menu';
+import type { MenuAction } from '@react-native-menu/menu';
 
 import type { CollectionItem, MenuPickerProps } from './MenuPicker.types';
 import { testProps } from './TestProps';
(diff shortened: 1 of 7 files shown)


lyzhan7:
@ankraj12 @ayush547 @rohanprasadofficial if one of you could check to see if this looks okay on Android that would be great - if not I'll fork off the iOS version into its own file.

ayush547:
> @ankraj12 @ayush547 @rohanprasadofficial if one of you could check to see if this looks okay on Android that would be great - if not I'll fork off the iOS version into its own file.

Hey Lynn,
Thanks for considering android upgrades as well, but looks like some more work is required to get this working for Android. I would suggest restricting this to iOS alone for now. 

https://user-images.githubusercontent.com/49569838/236414013-660d2d74-5db0-4066-9bbc-d16ae25458a1.mp4

lyzhan7:
> > @ankraj12 @ayush547 @rohanprasadofficial if one of you could check to see if this looks okay on Android that would be great - if not I'll fork off the iOS version into its own file.
> 
> Hey Lynn, Thanks for considering android upgrades as well, but looks like some more work is required to get this working for Android. I would suggest restricting this to iOS alone for now.
> 
>  Record_2023-05-05-14-10-11.mp4

Thanks for testing! I've just split them into separate files

apps/fluent-tester/src/TestComponents/Common/MenuPicker.tsx
```diff
@@ -1,43 +1,11 @@
...
-          label={item.label}
-          key={index}
-          value={item.value} /* For Android E2E testing purposes, testProps must be passed in after accessibilityLabel. */
-          {...testProps(item.testID)}
-        />
-      ))}
-    </Picker>
-  );
+export const MenuPicker: React.FunctionComponent<MenuPickerProps> = (_props: MenuPickerProps) => {
+  console.warn('Platform not supported');
```
Saadnajmi, line 9: At this point, do we have every platform we need? I'm OK with the fallback being react-native-picker tbh

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/Common/MenuPicker.tsx
lyzhan7: Yeah we have a MenuPicker file for every platform. Personally for cases like this it's more intuitive to me to just have a file per platform, with the fallback just returning null (or even not having a fallback file at all if possible) - it's easier to know what platforms you're affecting rather than having to deduce which ones don't have a platform specific file. 

If we don't want to do it like this, I was also thinking we could make the win32/macOS one the fallback - both of those currently just refer to the same MenuGroup.desktop.ts file. At least the benefit of that is that iOS/Android/windows won't be bundling the MenuGroup.desktop.ts file anymore

Saadnajmi: If it feels more intuitive to you, then we can leave as is. Thanks!

PR #2824, opened 2023-05-05
macOS: Implement MenuGroup 

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This change implements and integrates two new components into Menu: MenuGroup and MenuGroupHeader. MenuGroup provides a View container for the header text and menu items. 

To dos:
Accessibility: Header text needs to be VO accessible, its accessibleLabel may need to announce "heading" as its role.
MenuDivider: Having MenuDivider between menu groups is a nice to have.

For clarification, MenuGroup is not ready for consumption on win32. The look and feel of MenuGroup needs to be verified and accessibility requirements are not met with this change.

### Verification

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

packages/components/Menu/src/MenuGroupHeader/MenuGroupHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+  displayName: menuGroupHeaderName,
+  ...stylingSettings,
+  slots: {
+    root: Text,
+  },
+  useRender: (userProps: MenuGroupHeaderProps, useSlots: UseSlots<MenuGroupHeaderType>) => {
+    const Slots = useSlots(userProps);
+
+    return (_final: MenuGroupHeaderProps, children: React.ReactNode) => {
+      return <Slots.root accessible={userProps.accessible}>{children}</Slots.root>;
```
Saadnajmi, line 22: Can we just pass all the props to root?

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -0,0 +1,44 @@
...
+    root: View,
+  },
+  useRender: (userProps: MenuGroupProps, useSlots: UseSlots<MenuGroupType>) => {
+    const Slots = useSlots(userProps);
+    return (_final: MenuGroupProps, children: React.ReactNode) => {
+      let itemPosition = 0;
+      const childrenWithSet = React.Children.toArray(children).map((child) => {
+        if (React.isValidElement(child)) {
+          const itemCount = React.Children.toArray(children).filter(
+            (child) => React.isValidElement(child) && (child as any).type.displayName !== 'MenuGroupHeader',
```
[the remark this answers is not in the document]
ksiler: same for itemPosition++ below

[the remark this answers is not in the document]
ksiler: I guess you could argue that the user should use a MenuDivider in between two MenuGroups  (inside of within a MenuGroup like I was thinking). In that case, you wouldn't need the check here and Menu will take care of not including it as part of n-of-m.

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -0,0 +1,44 @@
...
+              accessibilitySetSize: child.props.accessibilitySetSize ?? itemCount, //win32
+            } as any,
+          );
+        }
+        return child;
+      });
+      return <Slots.root>{childrenWithSet}</Slots.root>;
+    };
+  },
+});
```
ksiler, line 44: I'm not sure where the issue lies but n-of-m properties aren't being set on submenu triggers on win32. I'll continue digging but the issue doesn't appear in the main branch.

ksiler: Oh sorry - this is because your branch doesn't include the latest changes. I merged in the code that enabled n-of-m for submenu triggers last week so it should be there once you merge them in from main.

packages/components/Menu/src/MenuGroup/MenuGroup.styling.ts
```diff
@@ -0,0 +1,19 @@
+import type { UseStylingOptions } from '@fluentui-react-native/framework';
+import { buildProps } from '@fluentui-react-native/framework';
+
+import { menuGroupName } from './MenuGroup.types';
+import type { MenuGroupProps, MenuGroupTokens, MenuGroupSlotProps } from './MenuGroup.types';
+
+export const stylingSettings: UseStylingOptions<MenuGroupProps, MenuGroupSlotProps, MenuGroupTokens> = {
+  tokens: [menuGroupName],
+  slotProps: {
+    root: buildProps(
```
rurikoaraki, line 10: nit: You can just pass an object, you don't need to call build props.

packages/components/Menu/src/MenuGroupHeader/MenuGroupHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const MenuGroupHeader = compose<MenuGroupHeaderType>({
+  displayName: menuGroupHeaderName,
+  ...stylingSettings,
+  slots: {
+    root: Text,
+  },
+  useRender: (userProps: MenuGroupHeaderProps, useSlots: UseSlots<MenuGroupHeaderType>) => {
+    const Slots = useSlots(userProps);
+
+    return (_final: MenuGroupHeaderProps, children: React.ReactNode) => {
```
rurikoaraki, line 21: I think you'll need to merge props from final into the component. Otherwise any settings to props passed to the MenuGroupHeader in the inner render will be dropped

Not likely to happen, but since it's easy to cover that case by copying code from other components, we should cover that case

samuelfreiberg:
We should add automated testing for these new components

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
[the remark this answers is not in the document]
chiuam: Ah I didn't think of that scenario, thanks for pointing that out. Yeah it should be valid to not have a header inside MenuGroup and separate menu items with MenuDivider. Agree with you on MenuDivider should be injected outside MenuGroup, the n-of-m behavior should be the same w/wo menu headers.

``` javascript
   <MenuList>
     <MenuGroup>
       <MenuItem>Item</MenuItem>
     </MenuGroup>
     <MenuDivider/>
     <MenuGroup>
       <MenuItem>Item</MenuItem>
     </MenuGroup>
   </MenuList>
```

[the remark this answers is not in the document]
ksiler: Sounds good, could be left as-is then. Thanks!

rurikoaraki (approved):

packages/components/Menu/src/MenuGroup/MenuGroup.styling.ts
chiuam: Yep confirmed that display is flex by default - https://reactnative.dev/docs/layout-props#display. I ended up removing the styling file altogether as it seems unnecessary

PR #2829, opened 2023-05-10
[Input] Change default svg to component - Android

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Using svg as a file poses some issues while importing and requires some setup. Changing svg into a component instead helps remove these issues.

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

packages/components/Input/src/dismissSvg.tsx
rohanprasadofficial, line 1: Putting this file inside a 'assets' folder inside src only would be good.

rohanprasadofficial (approved):
Putting this file inside a 'assets' folder.

packages/components/Input/package.json
```diff
@@ -40,11 +40,13 @@
     "@fluentui-react-native/test-tools": ">=0.1.1 <1.0.0",
     "@fluentui-react-native/scripts": "^0.1.1",
     "react": "17.0.2",
-    "react-native": "^0.68.0"
+    "react-native": "^0.68.0",
+    "react-native-svg": "^12.3.0"
```
ankraj12, line 44: Should this be specified in dependencies instead?

ankraj12:
Can you share why kind of issues were observed in current code..

packages/components/Input/src/dismissSvg.tsx
ayush547: Done

ayush547:
> Can you share what kind of issues were observed in current code.

SVGs require some [transformer](https://yarnpkg.com/package/react-native-svg-transformer) to be recognized as a module. Converting SVGs to component allows direct import. This should be better for partner consumption.

packages/components/Input/package.json
ayush547: Intentional

PR #2834, opened 2023-05-12
[Drawer - Android] Implement Structure and Basic Working Drawer

rohanprasadofficial:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes
Trying to keep the PR into part and small so easy to review 😄, still this might be big as there is boilerplate 

This is first PR for a series for PR Drawer Implementation and includes: 
- Basic structuring of the Drawer into Repo ( Component , E2E, Tester)
- Initial working version of POC , Still needs work on tokens , props , testers - will try to keep each of them into seperate PR for easy reviews. 

Next Steps: 
- Implement PanResponders for Drag 
- Tokenization for width , height and colors 

DEMO 
---------

https://github.com/microsoft/fluentui-react-native/assets/30728574/c31949e0-82f6-4f6a-85c0-39130c302c10

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Drawer/src/Drawer.styling.ts
```diff
@@ -0,0 +1,62 @@
+import type { Theme, UseStylingOptions } from '@fluentui-react-native/framework';
```
ankraj12, line 1: This implementation need not lie in experimental I think. Per last talk with PM on this, things can be in no experimental if:
- It's a V1 component.
- It's API is stable for the released endpoint (even if it's currently just android)

packages/experimental/Drawer/src/Drawer.types.ts
```diff
@@ -0,0 +1,195 @@
...
+   * */
+  onClose?: (e: InteractionEvent) => void;
+
+  /**
+   * Callback when the Drawer is opened
+   * */
+  onOpen?: (e: InteractionEvent) => void;
+
+  /**
+   * Whether the Drawer is open on mount
```
ankraj12, line 149: How is this prop. intended to be used? If this new component is within another full screen RN experience - and drawer is expected to open on mount of some other root, would the developer need to set "`visible`" on mount or "`defaultOpen`" - and if latter, how do you plan to handle that..

apps/fluent-tester/src/TestComponents/Drawer/DrawerDefault.tsx
```diff
@@ -0,0 +1,96 @@
...
+          <Button appearance="outline" onClick={() => handleDrawerOpen('bottom')}>
+            Open Drawer (Bottom)
+          </Button>
+        </View>
+        <Drawer visible={isDrawerVisible} onClose={handleDrawerClose} onBackdropClick={handleDrawerClose} position={drawerPosition}>
+          <View style={drawerContentstyles.drawerContent}>
+            <View style={drawerContentstyles.flexRow}>
+              <Avatar active={'active'} activeAppearance="ring" size={56} name="John Doe" avatarColor={'colorful'} />
+              <Text variant="headerSemibold" style={drawerContentstyles.text}>
+                John Doe
```
ankraj12, line 51: Is there any restriction on what can be placed within drawer from design guidance?

packages/experimental/Drawer/src/Drawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+      return (
+        <Slots.modal
+          {...rest}
+          visible={visible}
+          onRequestClose={onClose}
+          supportedOrientations={['portrait', 'portrait-upside-down', 'landscape', 'landscape-left', 'landscape-right']}
+          animationType="none"
+          transparent
+        >
+          <Slots.backdrop onPress={onBackdropClick}>
```
ankraj12, line 37: More for spec review query - duplicated from it.. could you more description on what these slots would handle each

packages/experimental/Drawer/src/Drawer.types.ts
```diff
@@ -0,0 +1,195 @@
...
+  shadowOffset?: { width: number; height: number };
+
+  /**
+   * The shadow opacity of the Drawer
+   * @platform iOS
+   * */
+  shadowOpacity?: number;
+
+  /**
+   * The shadow radius of the Drawer
```
ankraj12, line 120: cleanups pre merge

ankraj12 (changes requested):
Left open feedbacks

packages/experimental/Drawer/src/Drawer.styling.ts
rohanprasadofficial: This is intentional to put on experimental as there are many things are left on this such as: 
1. More Tokenization 
2. Handling of RTL, A11y 
3. E2E Tests 
4. Tests 
5. Implementation of PanResponder for Drag behaviours 

i don't think we can call it ready with merge of this PR, hence keeping it now in experimental til next implementation goes it, 

i will move this all to components folder once we have everything inplace and think Drawer is ready for consumption.

apps/fluent-tester/src/TestComponents/Drawer/DrawerDefault.tsx
rohanprasadofficial: Now Developers can put anything inside drawer and they handle every styles & layout how it looks. 
There are guidance for the width,height of Drawer that will be tokens only but users has ability to modify it.
It is treated as a blackbox and we will match what native android has implemented.

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: hmm , didn't get it, do you mean remove this ?

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: visible is the main criteria that will be used to define the state of Drawer - 'open' or 'close' .  

visible & defaultOpen will be working separately. 
defaultOpen depends upon visible : 

if visible is set of false - it becomes a control component and openDefault doesn't define anything.  

openDefault is applicable when the visible is true and screen is opened and Drawer visible = true.


defaultOpen is a property that decide the Drawer should open or not itself if the screen is loader where Drawer is used. 

Drawer will be not be tied to any sub rootview and will be always attached to topmost (zindex) view. 
if another Drawer comes it will be on top of previous Drawer (This is inlined with Native Drawer)

packages/experimental/Drawer/src/Drawer.tsx
rohanprasadofficial: Will add description to spec + code.

rohanprasadofficial: Description as added to the SPEC.md. 
I have already added them to spec doc and will be copied same to SPEC.md when we check in here.

packages/experimental/Drawer/src/DrawerTokens.ts
```diff
@@ -0,0 +1,23 @@
...
+    handleWidth: globalTokens.size360,
+    handleHeight: globalTokens.size40,
+    handleCornerRadius: globalTokens.size20,
+    handleAlignment: 'center',
+    handleMarginTop: globalTokens.size80,
+    handleMarginBottom: globalTokens.sizeNone,
+    handleMarginEnd: globalTokens.sizeNone,
+    handleMarginStart: globalTokens.sizeNone,
+    drawerElevation: globalTokens.size100,
+    position: 'left',
```
lenahong, line 21: Since position changes the tokens that get applied to the component, they are usually not set as component tokens themselves.

packages/experimental/Drawer/src/Drawer.types.ts
```diff
@@ -0,0 +1,200 @@
...
+   * The elevation of the Drawer
+   * @default 5
+   * */
+  drawerElevation?: number;
+
+  /**
+   * The position of the Drawer
+   * @default 'left'
+   * */
+  position?: 'left' | 'right' | 'bottom';
```
lenahong, line 90: Instead of having position itself a token, have left, right, bottom as individual tokens with value `DrawerTokens` and in Drawer.styling treat each token as a state condition for applying different style tokens. See Spinner size tokens for reference

packages/experimental/Drawer/src/Drawer.styling.ts
```diff
@@ -0,0 +1,64 @@
...
+      [],
+    ),
+    content: buildProps(
+      (tokens: DrawerTokens, _theme: Theme) => ({
+        style: {
+          position: 'absolute',
+          top: 0,
+          elevation: 18,
+          backgroundColor: tokens.drawerBackgroundColor,
+          borderTopRightRadius: tokens.position === 'bottom' ? tokens.drawerCornerRadius : 0,
```
lenahong, line 44: Instead of conditionally setting borderRadius based on other token values, have each of `left` `right` `botton` tokens style them separately.

packages/experimental/Drawer/src/Drawer.types.ts
lenahong: Since the value of the position token itself doesn't map to a specific style prop, you can remove this from DrawerTokens.

packages/experimental/Drawer/src/Drawer.types.ts
ankraj12: the @platform iOS *

rohanprasadofficial: But it is the property required for iOS for the shadow, i.e - elevation on android.  
I have them so that they can reused when we have a iOS component, For the spec i think the guidance is to consider other platform as well.

packages/experimental/Drawer/src/DrawerTokens.ts
```diff
@@ -0,0 +1,40 @@
...
+    handleHeight: globalTokens.size40,
+    handleCornerRadius: globalTokens.size20,
+    handleAlignment: 'center',
+    handleMarginTop: globalTokens.size80,
+    handleMarginBottom: globalTokens.sizeNone,
+    handleMarginEnd: globalTokens.sizeNone,
+    handleMarginStart: globalTokens.sizeNone,
+    drawerElevation: globalTokens.size100,
+    drawerCornerRadius: globalTokens.corner.radius120,
+    top: 0,
```
lenahong, line 22: I don't see this token defined in your Drawer.types so this probably doesn't do anything. Also, setting `left` `right` and `bottom` to numeric values in the conditions below isn't doing anything either.

If you want to set these as tokens on the Drawer, you want to define them in types first (possibly rethink naming conventions since it's confusing with the position values). Then also in Drawer.styling, you want to map each token to the corresponding style prop for each slot

packages/experimental/Drawer/src/Drawer.styling.ts
```diff
@@ -0,0 +1,54 @@
...
+      ['backdropColor'],
+    ),
+    content: buildProps(
+      (tokens: DrawerTokens, _theme: Theme) => ({
+        style: {
+          position: 'absolute',
+          height: tokens.height,
+          width: tokens.width,
+          elevation: tokens.drawerElevation,
+          ...(tokens.position === 'right' && { right: 0 }), // This is required to make the drawer stick to the right side of the screen as left overrides right position
```
lenahong, line 34: You shouldn't have to do this. Let DrawerTokens set the appropriate default tokens for each position state, and simply set right: tokens.[what you named the token] here

packages/experimental/Drawer/src/DrawerTokens.ts
rohanprasadofficial: Added to useDrawer.

packages/experimental/Drawer/src/DrawerTokens.ts
rohanprasadofficial: I have tokens for each position in the types and reference them in styling files worked out.

packages/experimental/Drawer/src/Drawer.types.ts
```diff
@@ -0,0 +1,221 @@
...
+  behavior?: DrawerBehaviorType;
+
+  /**
+   * Positions/Behaviours of the Drawer
+   */
+
+  leftSlideOver?: DrawerTokens;
+  rightSlideOver?: DrawerTokens;
+  bottomSlideOver?: DrawerTokens;
+  bottom?: DrawerTokens;
```
lenahong, line 141: duplicate?

packages/experimental/Drawer/src/Drawer.types.ts
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * The content of the Drawer
+   * */
+  children?: React.ReactNode;
+
+  /**
+   * The behavior of the Drawer
+   * @default 'leftSlideOver'
+   * */
+  behavior?: DrawerBehaviorType;
```
lenahong, line 193: nit: drawerPosition prop with values 'left' 'right' 'bottom' seems more intuitive imo

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: We had a discussion over spec and there are few variation on the bottom drawer. 
We have replaced the position with DrawerType now which consist of above 4 with the variation on bottom Drawer.


These tokens are for the same.

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: We tried to match the native naming convention with design suggested. 
But let me go through once and maybe we can remove slideOver and change the name of behaviour that is different than normal one.

packages/experimental/Drawer/src/Drawer.styling.ts
rohanprasadofficial: Yes took. [token Name] approach.

packages/experimental/Drawer/src/Drawer.styling.ts
rohanprasadofficial: Corrected.

packages/experimental/Drawer/src/Drawer.tsx
rohanprasadofficial: Sounds good.

apps/fluent-tester/src/TestComponents/Drawer/DrawerDefault.tsx
rohanprasadofficial: Discussed this on this over spec review.

ankraj12: Please add a note to validate different content sizes an their impact on the drawer size

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: I think this is good suggestion on the naming. 
Changing to drawerPosition with value only to 'left' 'right' 'bottom'  , while bottom drawer that have extra behaviour which can be take care by another prop if passed true.

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: Changed to right,left,bottom tokens for specific styling according to the position.

rohanprasadofficial: leftPosition?: number;
 rightPosition?: number;
 bottomPosition?: number;
 topPosition?: number; 
 
 These token are different than left, right, bottom - these are used to capture actual coordinates of the drawer the drawer is currently present.

packages/experimental/Drawer/src/DrawerTokens.ts
rohanprasadofficial: leftPosition?: number;
  rightPosition?: number;
  bottomPosition?: number;
  topPosition?: number;
  
  
  Added this to take the number coordinates of according to the position.

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: Removed having it on in props.

packages/experimental/Drawer/src/Drawer.types.ts
ankraj12: The guidance has been to test basic functionality for any new pure JS control in iOS, so that it validates bare-bones code sharing. If this token is not used currently - IMO it's something to skip for when the iOS fit and finish comes. Keeping types available which are non functional/ do not have any impact might add usage confusion.

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: Removed defaultOpen prop as Drawer seems be always controlled component which can be address via 'open' prop.

ankraj12 (approved):
Can you confirm the testing result if you have checked on iOS

packages/experimental/Drawer/src/Drawer.types.ts
rohanprasadofficial: Taking it to next PR  to make it sanitized to Android.

PR #2860, opened 2023-06-07
Declare missing dependencies

acoates-ms:
A couple of packages have dependencies that they use in code that are only declared as devDependencies.


Saadnajmi (approved):

acoates-ms:
Dup of #2868

Saadnajmi:
I thought I saw this change twice :D

PR #2861, opened 2023-06-07
Define TabList Package

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR defines the new TabList package and outlines its types and tokens. As I iterate on the component, I will bring in code from the old experimental tabs. 

### Verification

Tests should pass locally and on pipeline.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/TabList/src/TabListMargins.ts
gulnazsayed, line 1: I would exclude this file from the initial set-up PR since these functions aren't used anywhere yet

lawrencewin: Done

apps/fluent-tester/src/TestComponents/TabList/TabListTest.tsx
```diff
@@ -0,0 +1,36 @@
...
+const sections: TestSection[] = [
+  {
+    name: 'Main Test',
+    component: TabListMainTest,
+  },
+];
+
+export const TabListTest: React.FunctionComponent = () => {
+  const status: PlatformStatus = {
+    win32Status: 'Production',
```
lenahong, line 26: Should this be experimental?

lawrencewin: Fixed.

rohanprasadofficial:
@lawrencewin can you add few screenshots of the component ?

packages/experimental/TabList/src/TabList.types.ts
```diff
@@ -0,0 +1,113 @@
...
+import type { FocusZoneProps } from '@fluentui-react-native/focus-zone';
+import type { FontTokens, IForegroundColorTokens, IBackgroundColorTokens } from '@fluentui-react-native/tokens';
+
+export const tabListName = 'TabList';
+
+export interface TabListContextData {
+  /**
+   * The currently selected TabsItem's key
+   */
+  selectedValue: unknown;
```
lenahong, line 14: I don't think you should leave type as unknown. Would they be strings if they are keys?

lenahong: if they are TabItems, at least defined it and mark it with /* TODO */ tags

lenahong:
Props shouldn't be defined with unknown types. Maybe set them to primitive types such as string or boolean until you defined their correct types.

packages/experimental/TabList/src/TabList.types.ts
lawrencewin: > I don't think you should leave type as unknown. Would they be strings if they are keys?

This is a change that web made in their updated Tab component. `unknown` is used here because it's a more type-safe version of any. This would allow me to assign a tab's value to be a string, a number, an object, etc.

lenahong: On web, it has to be set to unknown because they are pltform specific HTML values. but for RN we type our props to be more specific. Also, use selectedKey instead of selectedValue (see Menu, RadioGroup for reference), and you can also follow the type convention there.

packages/experimental/TabList/src/TabList.types.ts
```diff
@@ -0,0 +1,113 @@
...
+export interface TabListProps extends Pick<FocusZoneProps, 'isCircularNavigation' | 'defaultTabbableElement'>, IViewProps {
+  /**
+   * Visual appearance of the TabList, affecting header hover / selection background.
+   */
+  appearance?: TabListAppearance;
+
+  /**
+   * The key of the TabsItem that will initially be selected
+   */
+  defaultSelectedValue?: unknown;
```
lenahong, line 52: defaultSelectedKey

packages/experimental/TabList/src/TabList.types.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Flag to disable all tabs
+   */
+  disabled?: boolean;
+
+  /**
+   * Callback for receiving a notification when the choice has been changed
+   */
+  onTabSelect?: (value: unknown) => void;
```
lenahong, line 62: onTabSelect?: (key: string) => void;

packages/experimental/TabList/src/TabList.types.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * The value of the selected option. If you provide this, you must maintain selection state by observing
+   * onTabsClick events and passing a new value in when changed. This overrides defaultSelectedValue
+   * and makes the Tabs a controlled component. This prop is mutually exclusive to defaultSelectedValue.
+   */
+  selectedValue?: unknown;
```
lenahong, line 69: selectedKey?: string

lenahong (approved):

PR #2866, opened 2023-06-09
[Callout][macOS] Implement setInitialFocus

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We have a scenario where we want to present a Callout without it becoming the Key Window (so that say, it doesn't take keyboard focus away from a TextInput). We can re-use the previously win32 only prop `setInitialFocus` to implement this.

### Verification

Re-used the existing test page test. Also verified that <Menu> still got focus.

https://github.com/microsoft/fluentui-react-native/assets/6722175/f6e35fa5-080a-4c16-8a83-8b0a72c9bd50

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/CalloutView.swift
+++ b/packages/components/Callout/macos/CalloutView.swift
@@ -26,7 +26,7 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
-    @objc public var setInitialFocus: Bool = false
+	@objc public var setInitialFocus: Bool = false
 
 	@objc public var onShow: RCTDirectEventBlock?
 
@@ -94,9 +94,9 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 
 		updateCalloutFrameToAnchor()
 		calloutWindow.orderFront(self)
-        if (setInitialFocus) {
-            calloutWindow.makeKey()
-        }
+		if (setInitialFocus) {
+		    calloutWindow.makeKey()
+		}
 
 		// Dismiss the Callout if the window is no longer active.
 		NotificationCenter.default.addObserver(self, selector: #selector(dismissCallout), name: NSApplication.didResignActiveNotification, object: nil)
--- a/packages/components/Callout/macos/CalloutView.swift
+++ b/packages/components/Callout/macos/CalloutView.swift
@@ -26,6 +26,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
+    @objc public var setInitialFocus: Bool = false
+
 	@objc public var onShow: RCTDirectEventBlock?
 
 	@objc public var onDismiss: RCTDirectEventBlock?
@@ -91,7 +93,10 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 
 		updateCalloutFrameToAnchor()
-		calloutWindow.makeKeyAndOrderFront(self)
+		calloutWindow.orderFront(self)
+        if (setInitialFocus) {
+            calloutWindow.makeKey()
+        }
 
 		// Dismiss the Callout if the window is no longer active.
 		NotificationCenter.default.addObserver(self, selector: #selector(dismissCallout), name: NSApplication.didResignActiveNotification, object: nil)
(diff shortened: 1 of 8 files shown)


packages/components/Callout/macos/CalloutView.swift
```diff
@@ -26,6 +26,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
+    @objc public var setInitialFocus: Bool = false
+
 	@objc public var onShow: RCTDirectEventBlock?
```
lyzhan7, line 31: does this need to be indented to match the other variables

lyzhan7: Also some indents below don't seem right

lyzhan7 (approved):

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -26,6 +26,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
+    @objc public var setInitialFocus: Bool = false
```
lyzhan7, line 29: Hmm would setting the default value to true preserve the existing behaviour if setInitialFocus isn't set?

packages/components/Callout/src/Callout.types.ts
```diff
@@ -199,8 +199,9 @@ export interface ICalloutProps extends IViewProps, ICalloutTokens {
   onShow?: () => void;
 
   /**
-   * @platform win32
-   * If true then the callout will attempt to focus the first focusable element that it contains.
+   * Determines whether the Callout sets focus when displayed.
+   * On macOS: this determines whether the Callout becomes the key window.
+   * On win32: If true then the callout will attempt to focus the first focusable element that it contains.
```
chiuam, line 204: For my own knowledge, does the prop behave differently between platforms?

chiuam (approved):

packages/components/Callout/src/Callout.types.ts
Saadnajmi: It does. It seems like on windows, it will put focus on the first focusable view inside the window. On macOS as far as I can tell, it's valid for the window to have first responder status (if it's a key window). So I didn't necessarily want to re-implement the windows behavior of having it focus on the first focusable window inside. I'd be open to making that change in the future.

packages/components/Callout/macos/CalloutView.swift
Saadnajmi: It would, but I prefer the default to be false (AKA, as if the prop was never passed in). Both Menu and ContextualMenu set the prop to true so I don't think we'll see a regression there.

packages/components/Callout/macos/CalloutView.swift
Saadnajmi: Yeah it's a tabs vs spaces thing. Good catch.

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -26,6 +26,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 		}
 	}
 
+	@objc public var setInitialFocus: Bool = false
```
mischreiber, line 29: Nit: `Bool -> BOOL, false -> NO`

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -33,9 +33,7 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
   const dismissBehaviors = isControlled ? controlledDismissBehaviors : undefined;
   const directionalHint = getDirectionalHint(isSubmenu, I18nManager.isRTL);
 
-  // Initial focus behavior differs per platform, Windows platforms move focus
-  // automatically onto first element of Callout
-  const setInitialFocus = Platform.OS === ('win32' as any) || Platform.OS === 'windows';
+  const setInitialFocus = true;
```
mischreiber, line 36: Yay fewer diffs!

PR #2879, opened 2023-06-16
[macOS] Components don't focus on click

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Native macOS Appkit controls don't take focus on click. Let's not have FURN do the same. 

### Verification

Clicked around and things seem fine

https://github.com/microsoft/fluentui-react-native/assets/6722175/749a9230-afb0-4bfd-ac94-f1eebd4e8bf2

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-interactive-hooks-00476c7a-9659-4dcd-ae01-130251e854a1.json
+++ b/change/@fluentui-react-native-interactive-hooks-00476c7a-9659-4dcd-ae01-130251e854a1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "useOnPressWihFocus doesn't take focus on macOS",
+  "packageName": "@fluentui-react-native/interactive-hooks",
+  "email": "sanajmi@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


packages/components/Button/src/useButton.ts
```diff
@@ -36,15 +42,19 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
 
   // GH #1336: Set focusRef to null if button is disabled to prevent getting keyboard focus.
   const focusRef = isDisabled ? null : componentRef;
-  const onClickWithFocus = useOnPressWithFocus(focusRef, onClick);
```
rurikoaraki, line 39: Seems like it would be less invasive to have the macos version of this not call focus (and maybe rename the function to be something a bit more cross plat friendly) rather than having to remember to fork this behavior on every platform

Saadnajmi: Hooks can't be conditionally called is one issue. The other is we export this hook and there's probably a case where someone wants focus with click. My thought was it would be weird if the hook didn't work for them suddenly and silently.

Saadnajmi: Though I can undo the 'inner' rename here..

rurikoaraki: Could add a wrapper and leave the core function alone if you want.

I think it's likely someone will forget that macOS needs different behavior (or windows needs particular behavior) when adding a new control

Saadnajmi: I did a code search internally, and nobody is directly referencing this method. So I think it'll be safe for me to change the implementation and update docs...

rurikoaraki (approved):

PR #2881, opened 2023-06-18
[Callout][macOS] Add new prop `vibrant`

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Callout on macOS was implemented to always have an NSVisualEffectView in its view heirarchy. This was necessary for ContextualMenu and Menu, but is problematic when you want to host non-vibrant content. For whatever reason, there is a slight inset between the Callouts' VisualEffectView and its' content, which can lead to a. weird border vibrancy effect. See the screenshot below:
Screenshot 2023-06-18 at 2 46 24 PM

A proper fix would be to make a new native module for NSVisualEffectView, and add it to FURN.
.... or we could just add a prop and switch between having an NSVisualEffectView or a regular NSView as the root.

One edge case is the borderRadius: We need to set one on the NSVisualEffectView itself, but don't need to for a regular NSView (which will just be transparent). So as is, vibrant callouts will still always have a border radius of 5.

### Verification

Added a new switch to the Callout test. I also verified that Menu and ContextualMenu behave as expected.

https://github.com/microsoft/fluentui-react-native/assets/6722175/a507003b-002f-417a-9480-54698a16a2b1

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


lyzhan7 (approved):

apps/fluent-tester/src/TestComponents/Callout/CalloutTest.tsx
```diff
@@ -394,14 +405,15 @@ const CustomCallout: React.FunctionComponent = () => {
...
+          vibrant={true}
           anchorRect={myRect}
           onDismiss={onDismissCustomizedCallout}
           onShow={onShowCustomizedCallout}
           accessibilityLabel="Customized Callout"
           accessibilityRole="alert"
           accessibilityOnShowAnnouncement="Be informed that a customized callout has been opened."
         >
-          <View style={{ padding: 20, borderWidth: 2, borderColor: 'black' }}>
+          <View style={{ padding: 20, borderWidth: 2, borderColor: 'black', backgroundColor: theme.colors.neutralBackground1 }}>
```
mischreiber, line 416: Is `neutralBackground1` the correct default color for a callout? I think I recall in some design spec somewhere that we wanted `neutralBackground4` by default, to better align with NSMenu/NSPopover.

mischreiber (approved):

apps/fluent-tester/src/TestComponents/Callout/CalloutTest.tsx
Saadnajmi: Good callout (no pun intended), I just chose a random color.

Saadnajmi:
Note: I think it might be worth doing the "proper" fix of making a separate native module for material / NSVisualEffectView and having Callout reference that. I'll revisit when I get more info

Saadnajmi:
Talked with @ppatboyd offline, and decided we don't need this prop for now. Re-working the branch for a different prop instead: `material` :)

PR #2892, opened 2023-06-21
Port old Tab implementation code into new TabList

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Refactor old Tab component and hook code, and port it into new TabList. So far, the component behaves like the old Tabs component without the header or panel. Work still needs to be done to get styling and tokens aligned with the new spec. 

Notes:
- Had to change prop name for Tab component from `key` to `tabKey` - `key` is reserved in React.
- Removed some unneeded tokens on the TabList component, will do the same pruning on Tab. 

### Verification

Tested on win32 locally. 

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/TabList/src/TabList.styling.ts
```diff
@@ -8,7 +8,21 @@ import { defaultTabListTokens } from './TabListTokens';
...
+      }),
+      [],
+    ),
+    stack: buildProps(
+      (tokens: TabListTokens) => ({
+        style: {
+          flexDirection: tokens.flexDirection,
+        },
+      }),
+      [],
```
gulnazsayed, line 25: add 'flexDirection'

packages/experimental/TabList/src/TabList.tsx
```diff
@@ -39,8 +39,46 @@ export const TabList = compose<TabListType>({
...
+        });
+      }
+
+      return (
+        <TabListContext.Provider
+          // Passes in the selected key and a hook function to update the newly selected tabsItem and call the client's onTabsClick callback.
+          value={tabs.state.context}
+        >
+          <Slots.root {...mergedProps}>
+            <Slots.container defaultTabbableElement={defaultTabbableElement} isCircularNavigation={isCircularNavigation}>
```
gulnazsayed, line 76: Replace 'isCircularNavigation={isCircularNavigation}' with 'isCircularNavigation'

packages/experimental/TabList/src/TabList.tsx
```diff
@@ -39,8 +39,46 @@ export const TabList = compose<TabListType>({
...
-    return () => {
-      return <Slots.root></Slots.root>;
+    return (final: TabListProps, ...children: React.ReactNode[]) => {
+      if (!tabs.state) {
+        return null;
+      }
+
+      const { defaultTabbableElement, isCircularNavigation, ...mergedProps } = mergeProps(tabs.props, final);
+
+      // Populate the tabsItemKeys array.
```
ksiler, line 49: nit tabsItemKeys -> tabKeys ?

packages/experimental/TabList/src/TabList.tsx
```diff
@@ -39,8 +39,46 @@ export const TabList = compose<TabListType>({
...
+      if (children) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore - TODO, fix typing error
```
ksiler, line 52: any idea what the typing error is here? maybe we can fix it?

packages/experimental/TabList/src/TabList.tsx
ksiler: same with defaultTabbableElement

packages/experimental/TabList/src/useTab.ts
```diff
@@ -8,8 +14,66 @@ import type { TabProps, TabInfo } from './Tab.types';
...
+
   return {
-    props: props,
-    state: {},
+    props: {
+      ...pressable.props,
+      accessible: accessible ?? true,
+      accessibilityRole: 'tab',
+      focusable: !disabled ?? true,
+      accessibilityState: { disabled: disabled, selected: info.selectedKey === tabKey },
```
ksiler, line 66: Can we use the memoized getAccessibilityStateWorker function for this like we do for other controls (see [useMenuItem.ts](https://github.com/microsoft/fluentui-react-native/blob/d1fc0c774448d103085b5bb5ab0103678a62cbbc/packages/components/Menu/src/MenuItem/useMenuItem.ts#L129))? That way, someone can pass in accessibilityState if they want to and override these values.

packages/experimental/TabList/src/useTabList.ts
```diff
@@ -8,8 +13,38 @@ import type { TabListProps, TabListInfo } from './TabList.types';
  * @returns configured props and state for Tabs
  */
 export const useTabList = (props: TabListProps): TabListInfo => {
+  const defaultComponentRef = React.useRef(null);
+  const { accessible, componentRef = defaultComponentRef, isCircularNavigation, selectedKey, onTabSelect, defaultSelectedKey } = props;
+
+  const data = useSelectedKey(selectedKey || defaultSelectedKey || null, onTabSelect);
+
+  // selectedTabRef should be set to default tabbale element.
```
ksiler, line 21: tabbale should be tabbable

packages/experimental/TabList/src/Tab.types.ts
```diff
@@ -97,7 +97,7 @@ export interface TabProps extends Omit<PressablePropsExtended, 'onPress'> {
   /**
    * A unique key-identifier for each option
    */
-  key: string;
+  tabKey: string;
```
PPatBoyd, line 100: this is a breaking change, not a patch change

PPatBoyd (changes requested):

packages/experimental/TabList/src/TabList.tsx
lawrencewin: I think this had to do with the old child param being a ReactChild type. Switching to ReactNode seems to fix the issue.

packages/experimental/TabList/src/TabList.tsx
rurikoaraki: No I think we want the current code. If you pass in isCircularNavigation directly then it will always be true. The spec is that the user is able to configure it. defaultTabbableElement is not a boolean prop, so it must take a value. Passing in the name of the prop only works for boolean props and will set it to true.

This would work fine if you're setting the properties on an object of FocusZoneProp type, though.

ksiler: Ah, thanks for the clarification!

packages/experimental/TabList/src/TabList.types.ts
```diff
@@ -35,7 +34,12 @@ export interface TabListContextData {
   focusZoneRef?: React.RefObject<any> | null;
 }
 
-export interface TabListTokens extends IForegroundColorTokens, FontTokens, IBackgroundColorTokens {}
+export interface TabListTokens {
+  /**
+   * Flex direction of Tab items. Controlled by TabList `vertical` prop.
+   */
+  flexDirection?: ViewStyle['flexDirection'];
```
PPatBoyd, line 41: flexDirection shouldn't be needed as a token if we have vertical as a prop

PPatBoyd (changes requested):

packages/experimental/TabList/src/TabList.styling.ts
```diff
@@ -8,7 +8,21 @@ import { defaultTabListTokens } from './TabListTokens';
...
+        style: {
+          display: 'flex',
+        },
+      }),
+      [],
+    ),
+    stack: buildProps(
+      () => ({
+        style: {
+          flexDirection: 'row',
```
rurikoaraki, line 22: Doesn't this make it so that the TabList is always horizontal?

packages/experimental/TabList/src/TabList.types.ts
rurikoaraki: Vertical is the prop that would drive the styling but flexDirection needs to be a specified token for the appropriate styling to be applied when vertical is set.

The flow of props to styles is generally the following:
Prop applies a layer of style tokens so that the output of tokens has the appropriate values given the current set of props.
Style tokens are applied at the useStyling stage and converted into props for each slot.
Slot props are passed to slots at inner render.

rurikoaraki: @lawrencewin I would say that it doesn't make sense for vertical to be a token, since "vertical" is not a style. Vertical should be a prop which applies style tokens.

rurikoaraki (approved):

gulnazsayed (approved):

PPatBoyd (approved):

PR #2910, opened 2023-06-30
macOS: Fix for menu items should be accessible when VoiceOver is enabled

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently menu items are not accessible through the VoiceOver command VO-shift-down, most likely due to the menu items are not recognized as children of the parent view. Let's make the scrollview not accessible so that VO focus would land on the first menu item instead.

### Verification

Before:
(Pressing the down arrow key moves VO focus to the first menu item here because I have the synchronize focus + VO cursor setting on, without it would get stuck instead)

https://github.com/microsoft/fluentui-react-native/assets/67026167/d3608cb5-fe3a-4a10-aace-94df04dd1dd3

After:

https://github.com/microsoft/fluentui-react-native/assets/67026167/4c3adbfa-55d9-4324-867e-36b8c5c4fb0e

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuList/MenuList.tsx
+++ b/packages/components/Menu/src/MenuList/MenuList.tsx
@@ -89,7 +89,6 @@ export const MenuList = compose<MenuListType>({
         Platform.OS === 'macos' ? (
           <Slots.root onMouseLeave={setFocusZoneFocus} onKeyDown={menuList.onListKeyDown}>
             <Slots.scrollView
-              accessibilityRole="menu"
               showsVerticalScrollIndicator={menuContext.hasMaxHeight}
               showsHorizontalScrollIndicator={menuContext.hasMaxWidth}
             >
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -89,7 +89,6 @@ export const MenuList = compose<MenuListType>({
         Platform.OS === 'macos' ? (
           <Slots.root onMouseLeave={setFocusZoneFocus} onKeyDown={menuList.onListKeyDown}>
             <Slots.scrollView
-              accessibilityRole="menu"
```
lyzhan7, line 92: Should we gate this for win32? Some accessibilityRoles are only available on netui, so this might be intended for win32

lyzhan7: Nvm there's already a platform check!

chiuam: Talked offline - this block of code should only execute on macOS.

harrieshin (approved):

PR #2911, opened 2023-07-01
Remove default palette for win32

lenahong:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
- remove call to createAliasesFromPalette in createOfficeTheme
- avoid setting default palette tokens that will override previous color settings

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/theming/win32-theme/src/createOfficeTheme.ts
```diff
@@ -66,7 +66,6 @@ export function createOfficeTheme(options: ThemeOptions = {}): ThemeReference {
       }
 
       return {
-        colors: createAliasesFromPalette(theme.host.palette, ref.themeName === 'HighContrast'),
```
rurikoaraki, line 69: We need this code to run for when a palette is added. Running this code should be conditional on whether a palette is passed in.

rurikoaraki: (One of your past commits looked like it had what I had in mind but for some reason it's showing as the entire line has been deleted)

lenahong: Ah ic, I can revert it to the conditional setting. Thanks!!

rurikoaraki (approved):

PR #2921, opened 2023-07-07
[E2E Tests] Waiters fix - Android

ayush547:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

PR #2917 updated verifyTextContent function to use compareAttribute fn from basePage instead of a getAttribute fn. But since compareAttribute function throws an error when values don't match, the waiters in Input & Chip were not getting run. Removing the initial verifyTextContent call to prevent this.

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/Chip/pages/ChipPageObject.ts
```diff
@@ -13,9 +13,7 @@ class ChipPageObject extends BasePage {
   /* Waits for the text content to get updated to new string.
    * Returns true if new string is attained. */
   async waitForStringUpdate(newState: string, errorMessage: string): Promise<boolean> {
```
samuelfreiberg, line 15: Why do we need this wrapper function?

Why not just call `await this.waitForCondition(async () => await this.verifyTextContent(newState), errorMessage);` from the spec?

ayush547: I had added it initially to follow the same pattern as that followed in Input tests, but here since there is only one call made to this function, removing wrapper fn

samuelfreiberg (approved):

PR #2926, opened 2023-07-07
Fix logic for non-main publish

acoates-ms:
While running a test run of publishing a patch version of FURN I realized that the logic I copied from another repo only works on windows machines.  While the FURN publish runs on ubuntu.  So I'm changing the logic to use ADO expressions rather than CMD variable expansion.

(Also removing the duplicate specification of the branch parameter)

### DIFF
--- a/.ado/azure-pipelines.publish.yml
+++ b/.ado/azure-pipelines.publish.yml
@@ -66,7 +66,7 @@ jobs:
         condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/main'))
 
       - script: |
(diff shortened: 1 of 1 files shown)


rurikoaraki (approved):

acoates-ms:
Waiting on https://office.visualstudio.com/ISS/_build/results?buildId=21271511&view=results to verify this fix.

acoates-ms:
Previous build failed, now waiting on https://office.visualstudio.com/ISS/_build/results?buildId=21272316&view=results to verify.

PR #2951, opened 2023-07-14
[Callout] Fix callout reanchoring from `target` to `anchorRect`

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The callout doesn't reanchor when it has rendered with a `target` anchor and new props specify `anchorRect`. It would seem native code never gets an update that `target` isn't specified anymore, because we don't update internal state used to populate that prop -- so fixing that.

Also making changes to the test page to validate the dynamic changing of anchor.

### Verification

Manually, using the updated test page. New behavior video: https://github.com/microsoft/fluentui-react-native/assets/72474613/d64107b7-caa3-4301-81bc-3f9b50d87a74

### Pull request checklist

This PR has considered (when applicable):
- [ ] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #2952, opened 2023-07-14
[macOS] [FRNCallout] Fix switching betwen target and anchorRect

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Native code isn't equipped to handle dynamically changing between these if the callout is already rendered (on screen).

* We assume given a `target`, there's always a valid `NSView`. That's a reasonable assumption, but when switching from `target` to `anchorRect`, we need to handle `nil` for target
* `calculateAnchorViewScreenRect` assumes there's a valid `anchorView`. Again a reasonable assumption, but then code also assumes if there's no `target`, there MUST be a `anchorRect`. We can remove the assumption that there's an `anchorView` when `calculateAnchorViewScreenRect` is called, but I think I like switching the precedence better, because that makes the Mac behavior consistent with Win32 (Office), which will only fall back to `anchorRect` if other methods of obtaining an anchor (like `target`) fail.
* There's an unfortunate assumption that every time a property changes, we're going to be in a valid state. That depends on the ordering the property changes come in -- for example, in the scenario from going from `target` to `anchorRect`, if first the new `anchorRect` is specified, and then the old `target` is cleared, we're good. In the other sequence, we have no anchor in the transient state. Let's not update the anchor in such a scenario, until we get to a "good" state.

### Verification

Testing:

* No more crashes when switching between anchors, with #2951 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/CalloutView.swift
+++ b/packages/components/Callout/macos/CalloutView.swift
@@ -6,7 +6,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 
 	@objc public var target: NSNumber? {
 		didSet {
-			guard let targetView = bridge?.uiManager.view(forReactTag: target) else {
+			let targetView = bridge?.uiManager.view(forReactTag: target)
+			if (targetView == nil && target != nil) {
 				preconditionFailure("Invalid target react tag")
 			}
 			anchorView = targetView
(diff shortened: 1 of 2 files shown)


packages/components/Callout/macos/CalloutView.swift
```diff
@@ -6,7 +6,8 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 
 	@objc public var target: NSNumber? {
 		didSet {
-			guard let targetView = bridge?.uiManager.view(forReactTag: target) else {
+			let targetView = bridge?.uiManager.view(forReactTag: target)
+			if (targetView == nil && target != nil) {
 				preconditionFailure("Invalid target react tag")
```
Saadnajmi, line 11: Should we remove the "react tag" bit now and just say "invalid target"?

Saadnajmi (approved):

packages/components/Callout/macos/CalloutView.swift
nakambo: sure

PR #2953, opened 2023-07-17
[iOS] Tokenize Chip and Input

ayush547:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x] android

### Description of changes

Added tokens for Chip and Input
Enabled them in tester app
Fixed font token not applying in Input

### Verification

Chip normal -
image

Chip selected - 
image

Chip with close button - 
image

Input - 
image

Input Error state -
image

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Input/src/Input.tsx
```diff
@@ -22,16 +18,7 @@ import { useInput } from './useInput';
...
-  return (
-    (layer === 'focused' && state.focused && !state.text) ||
-    (layer === 'filled' && !state.focused && !!state.text) ||
-    (layer === 'typing' && state.focused && !!state.text) ||
-    userProps[layer] ||
-    (layer === 'hasIcon' && userProps.defaultIcon) ||
-    (layer === 'error' && !!userProps.error && !!state.text)
-  );
-};
+export const inputLookup = () => null;
```
Saadnajmi, line 21: ?

packages/components/Input/src/Input.tsx
```diff
@@ -48,60 +35,9 @@ export const Input = compose<InputType>({
...
-                </Slots.accessoryIconPressable>
-              )}
-            </Slots.input>
-          </IconWrapper>
-          {(assistiveText || error != undefined) && <Slots.assistiveText>{error ? error : assistiveText}</Slots.assistiveText>}
-        </Slots.root>
-      );
+  useRender: (_userProps: InputProps, _useSlots: UseSlots<InputType>) => {
+    return (_final: InputProps) => {
+      return null;
```
Saadnajmi, line 40: ?

packages/components/Input/src/InputTokens.ios.ts
```diff
@@ -0,0 +1,86 @@
...
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import type { TokenSettings } from '@fluentui-react-native/use-styling';
+
+import type { InputTokens } from './Input.types';
+
+export const defaultInputTokens: TokenSettings<InputTokens, Theme> = (t: Theme) =>
+  ({
+    iconSize: 24,
+    hasIcon: {
+      //spacingIconContent + iconSize
```
Saadnajmi, line 11: Remove commented out bit?

packages/components/Input/src/InputTokens.ios.ts
```diff
@@ -0,0 +1,86 @@
...
+      spacingAssistiveTextStart: globalTokens.size160 + 24,
+      spacingLabelStart: globalTokens.size160 + 24,
```
Saadnajmi, line 13: Should we move 24 to a constant?

packages/components/Input/src/Input.mobile.tsx
```diff
@@ -0,0 +1,107 @@
+/** @jsxRuntime classic */
```
rohanprasadofficial, line 1: If Input is not planned to Desktop platform we can have only one file as Input.tsx and that should be fine.

packages/components/Input/src/Input.tsx
ayush547: Disabling for other platforms similar to Chip

packages/components/Input/src/Input.tsx
ayush547: Disabling for other platforms similar to Chip

packages/components/Input/src/InputTokens.ios.ts
ayush547: Have added this to explain the value (In figma it is two separate values which translate to one single here)

packages/components/Input/src/InputTokens.ios.ts
ayush547: All places that use icon size currently use hard coded values

ayush547: Should possibly be replaced everywhere though, not including it here

packages/components/Input/src/Input.mobile.tsx
ayush547: Chip was in a similar situation, it was recommended there to move it out to a mobile file, following same precedence here.

rurikoaraki:
Side effect of this is that Input will render nothing for non-mobile platforms. Is that ok @saadnajmi?

Saadnajmi:
> Side effect of this is that Input will render nothing for non-mobile platforms. Is that ok @Saadnajmi?

Yeah , I guess that's the standard we've kinda coalesced on

rurikoaraki (approved):

PR #2966, opened 2023-07-25
Fix TabList keyboarding, focus, and accessibility + Icon fix. 

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR makes multiple fixes to TabList:
- Fix Icon not rendering properly. Icon should now have correct color tokens and margins.
- Fix keyboarding and focus. TabList root is no longer focusable, and users are now no longer able to focus on and control disabled TabLists and Tab elements.
- Set and expose a11y properties for Tab and TabList. 

### Verification

Informal testing through Tester app. Below is an image of a TabList with icons.
image

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/packages/experimental/TabList/src/Tab/Tab.styling.ts
+++ b/packages/experimental/TabList/src/Tab/Tab.styling.ts
@@ -42,14 +42,6 @@ export const stylingSettings: UseStylingOptions<TabProps, TabSlotProps, TabToken
       }),
       ['iconColor', 'iconSize'],
     ),
-    iconPadding: buildProps(
-      (tokens: TabTokens) => ({
-        style: {
-          width: tokens.iconMargin,
-        },
-      }),
-      ['iconMargin'],
-    ),
     stack: buildProps(
       (tokens: TabTokens) => ({
         style: {
--- a/packages/experimental/TabList/src/Tab/Tab.styling.ts
+++ b/packages/experimental/TabList/src/Tab/Tab.styling.ts
@@ -28,20 +28,27 @@ export const stylingSettings: UseStylingOptions<TabProps, TabSlotProps, TabToken
       (tokens: TabTokens, theme: Theme) => ({
         style: {
           color: tokens.color,
+          marginLeft: tokens.contentMarginLeft,
+          marginRight: tokens.contentMarginRight,
           ...fontStyles.from(tokens, theme),
         },
       }),
-      ['color', ...fontStyles.keys],
+      ['color', 'contentMarginLeft', 'contentMarginRight', ...fontStyles.keys],
     ),
     icon: buildProps(
       (tokens: TabTokens) => ({
         color: tokens.iconColor,
         size: tokens.iconSize,
+      }),
+      ['iconColor', 'iconSize'],
+    ),
+    iconPadding: buildProps(
+      (tokens: TabTokens) => ({
         style: {
-          marginRight: tokens.iconMargin,
+          width: tokens.iconMargin,
         },
       }),
-      ['iconColor', 'iconSize', 'iconMargin'],
+      ['iconMargin'],
     ),
     stack: buildProps(
       (tokens: TabTokens) => ({
--- a/packages/experimental/TabList/src/Tab/Tab.styling.ts
+++ b/packages/experimental/TabList/src/Tab/Tab.styling.ts
@@ -24,16 +24,24 @@ export const stylingSettings: UseStylingOptions<TabProps, TabSlotProps, TabToken
       }),
       ['flexDirection', 'backgroundColor', ...borderStyles.keys],
     ),
+    contentContainer: buildProps(
+      (tokens: TabTokens) => ({
+        style: {
+          flexDirection: 'row',
+          paddingStart: tokens.contentMarginStart,
+          paddingEnd: tokens.contentMarginEnd,
+        },
+      }),
+      ['contentMarginEnd', 'contentMarginStart'],
+    ),
     content: buildProps(
       (tokens: TabTokens, theme: Theme) => ({
         style: {
           color: tokens.color,
-          marginLeft: tokens.contentMarginLeft,
-          marginRight: tokens.contentMarginRight,
           ...fontStyles.from(tokens, theme),
         },
       }),
-      ['color', 'contentMarginLeft', 'contentMarginRight', ...fontStyles.keys],
+      ['color', ...fontStyles.keys],
     ),
     icon: buildProps(
       (tokens: TabTokens) => ({
--- a/packages/experimental/TabList/src/Tab/Tab.tsx
+++ b/packages/experimental/TabList/src/Tab/Tab.tsx
@@ -1,11 +1,12 @@
 /** @jsxRuntime classic */
 /** @jsx withSlots */
+/** @jsxFrag */
 import * as React from 'react';
 import { Pressable, View } from 'react-native';
 
 import type { UseSlots } from '@fluentui-react-native/framework';
 import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
-import { Icon } from '@fluentui-react-native/icon';
+import { IconV1 as Icon } from '@fluentui-react-native/icon';
 import { TextV1 as Text } from '@fluentui-react-native/text';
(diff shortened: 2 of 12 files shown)


packages/experimental/TabList/src/Tab/Tab.tsx
```diff
@@ -59,11 +62,15 @@ export const Tab = compose<TabType>({
...
+      if (__DEV__ && !text && !icon) {
+        console.warn('A Tab component must render content. Text, an icon, or both should at least be passed in.');
+      }
+
       return (
         <Slots.root {...mergedProps}>
           <Slots.stack>
             {icon && <Slots.icon {...icon} />}
-            <Slots.content>{text}</Slots.content>
+            {text && <Slots.content accessible={false}>{text}</Slots.content>}
```
gulnazsayed, line 73: Why is accessible set to false here?

ksiler: So it doesn't get included in the accessibility tree. Text and/or images in teh accessibility tree isn't necessary unless it contains additional information. In most cases, the accessibilityLabel set is enough to describe the tab component. If an accessibilityLabel isn't set, we set it to the value of the text component. We do this for button too.

packages/experimental/TabList/src/TabList/useTabList.ts
```diff
@@ -1,28 +1,30 @@
...
     appearance = 'transparent',
+    accessibilityState,
     componentRef = defaultComponentRef,
+    defaultSelectedKey,
+    disabled = false,
     isCircularNavigation,
-    selectedKey,
     onTabSelect,
-    defaultSelectedKey,
+    selectedKey,
```
ksiler, line 27: We should probably add accessibilityLabel to this list (and below) so it can be set. Currently, the tablist component is showing up with no accessibilityLabel if one is set.

packages/experimental/TabList/src/Tab/Tab.tsx
lawrencewin: This is so that the Tab items in the UIA tree don't render the text node under the tab node, sort of like what the FURN button does.

packages/experimental/TabList/src/TabList/useTabList.ts
ksiler: Actually, I take that back. It is being set. Sorry!

packages/experimental/TabList/src/Tab/Tab.tsx
```diff
@@ -59,11 +62,15 @@ export const Tab = compose<TabType>({
 
       const { icon, tabKey, ...mergedProps } = mergeProps(tab.props, final, { accessibilityLabel: final.accessibilityLabel || text });
```
ksiler, line 63: I can't seem to set a custom accessibilityLabel on individual tabs (it defaults to text). Could you check to see if it works?

packages/experimental/TabList/src/Tab/Tab.tsx
```diff
@@ -50,6 +52,7 @@ export const Tab = compose<TabType>({
         return null;
       }
 
+      // Only support text as content for now.
       let text = '';
       React.Children.forEach(children, (child) => {
         if (typeof child === 'string') {
```
rurikoaraki, line 58: What does this do? It seems to set text to whatever the last child is?

packages/experimental/TabList/src/Tab/Tab.tsx
rurikoaraki: Yeah this will only catch the case where accessibilityLabel is passed as a prop to an inner render of a component using Tab. It won't catch accessibilityLabel passed in as part of normal use. Probably what you want to do is merge the props first, and then set accessibilityLabel using the result of the merge

packages/experimental/TabList/src/Tab/Tab.types.ts
```diff
@@ -19,9 +19,14 @@ export const tabName = 'Tab';
 
 export interface TabTokens extends FontTokens, IBorderTokens, IForegroundColorTokens, IBackgroundColorTokens, LayoutTokens {
   /**
-   * Horizontal margin of the tab text.
+   * Horizontal left margin of the tab text. If an icon and text are both rendered, this is the margin beween the icon and text.
    */
-  contentMarginHorizontal?: number;
+  contentMarginLeft?: number;
```
rurikoaraki, line 24: Should probably be contentMarginStart instead of Left so that it signifies that it adjusts for RTL properly.

packages/experimental/TabList/src/Tab/Tab.styling.ts
```diff
@@ -28,20 +28,19 @@ export const stylingSettings: UseStylingOptions<TabProps, TabSlotProps, TabToken
       (tokens: TabTokens, theme: Theme) => ({
         style: {
           color: tokens.color,
+          marginLeft: tokens.contentMarginLeft,
```
rurikoaraki, line 31: Should use marginStart/marginEnd

lawrencewin:
> > This PR makes multiple fixes to TabList:
> > Fix Icon not rendering properly. Icon should now have correct color tokens and margins.
> > Fix keyboarding and focus. TabList root is no longer focusable, and users are now no longer able to focus on and control disabled TabLists and Tab elements.
> > Set and expose a11y properties for Tab and TabList.
> 
> In the future, you should consider making each of these its own PR. Smaller changes are easier to review, and it makes incremental progress easier to make (ie an issue with one part of the change won't hold up an unrelated part of the change)

Fair. I've had these changes locally for a while so I put them together in a PR. I can def uncouple the slot changes vs the keyboarding and accessibility changes into separate PRs right now.

packages/experimental/TabList/src/Tab/Tab.tsx
lawrencewin: This gets the text for the accessibility label. For the text content in the Tab button, I want text to render similar to how the FURN button does it, but comparing code, I'm realizing my mistake. Let me fix this.

rurikoaraki:
> > > This PR makes multiple fixes to TabList:
> > > Fix Icon not rendering properly. Icon should now have correct color tokens and margins.
> > > Fix keyboarding and focus. TabList root is no longer focusable, and users are now no longer able to focus on and control disabled TabLists and Tab elements.
> > > Set and expose a11y properties for Tab and TabList.
> > 
> > 
> > In the future, you should consider making each of these its own PR. Smaller changes are easier to review, and it makes incremental progress easier to make (ie an issue with one part of the change won't hold up an unrelated part of the change)
> 
> Fair. I've had these changes locally for a while so I put them together in a PR. I can def uncouple the slot changes vs the keyboarding and accessibility changes into separate PRs right now.

You could leave it as is (splitting it up now would be a little confusing), but take it into account for future changes.

packages/experimental/TabList/src/Tab/Tab.tsx
gulnazsayed: Makes sense, thanks for the context!

rurikoaraki (approved):

packages/experimental/TabList/src/Tab/useTab.ts
```diff
@@ -49,7 +63,12 @@ export const useTab = (props: TabProps): TabInfo => {
     [changeSelection],
   );
 
-  /* We use the componentRef of the currently selected tabsItem to maintain the default tabbable
+  const accessibilityActionsProp = React.useMemo(
```
ksiler, line 66: Since we're allowing the user to pass their own value for `accessibilityActions`, we should also call their `onAccessibilityAction` prop. Similar to: https://github.com/microsoft/fluentui-react-native/blob/edebdac97d38c85002ccee7f074f288dfd629c0c/packages/components/Menu/src/MenuItemCheckbox/useMenuItemCheckbox.ts#L103C9-L103C35

I'd also check if we're disabled before we call 'Select' in that function.

ksiler (approved):

PR #2979, opened 2023-07-28
Fix TabList Double Narration Bug on Selection via Arrow Keys and Mouse

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, the TabList has an accessibility bug where switching selection by both mouse and keyboard causes the narrator to announce the interacted Tab twice. This PR fixes this in two separate ways: 

- Mouse: The reason double narration occurs on win32 is because focus is set before the selection state of the pressable is updated. This means the narrator announces the Tab, once when focus is set, and once when selection is set. To fix this, we need to set focus after the selection state is updated. I implemented the same fix in the `onPress` callback as #2461 uses in the `onKeyDown` callback. 
- Keyboarding: It turns out that the implementation of arrow key interaction was not in line with the current TabList spec. Previously, pressing arrow keys switched both focus and selection. Now, arrow keys only switch focus between Tabs, and ENTER / SPACE are used to select the current focused Tab. This indirectly fixes the narrator bug. 

This PR also populates the `tabKeys` list in the TabList context via the `useTab` hook rather than in the final render of `TabList`. 

### Verification

Tested in the win32 Fluent Tester app. 

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/experimental/TabList/src/TabList/TabList.tsx
```diff
@@ -36,36 +25,23 @@ export const TabList = compose<TabListType>({
   },
   useRender: (userProps: TabListProps, useSlots: UseSlots<TabListType>) => {
     // configure props and state for tabs based on user props
-    const tabs = useTabList(userProps);
+    const tablist = useTabList(userProps);
```
gulnazsayed, line 28: nit: camel-case to tabList

packages/experimental/TabList/src/TabList/TabList.types.ts
```diff
@@ -11,37 +11,66 @@ export type TabListAppearance = 'transparent' | 'subtle';
...
+  invoked?: boolean;
 
   /**
    * Updates the selected Tab and calls the client’s onTabSelect callback
    */
   onTabSelect?: (key: string) => void;
 
   /**
-   * Updates the selected Tab's ref to set as the default tabbable element
+   * Method to remove a tabKey from a context's `tabKeys` list. Run once when the tab umounts.
```
ksiler, line 41: "unmounts"

packages/experimental/TabList/src/Tab/useTab.win32.ts
```diff
@@ -0,0 +1,149 @@
...
+    ...rest
+  } = props;
+  // Grabs the context information from Tabs (currently selected Tab and client's onTabSelect callback).
+  const { addTabKey, invoked, onTabSelect, removeTabKey, setInvoked, setSelectedTabRef, selectedKey, tabKeys, ...tablist } =
+    React.useContext(TabListContext);
+
+  const isDisabled = disabled || tablist.disabled;
+
+  const changeSelection = React.useCallback(() => {
+    if (tabKey != selectedKey) {
```
ksiler, line 42: nit: !==

ksiler: also in useTab.ts

ksiler (approved):

PR #2984, opened 2023-07-31
Add alwaysShowToolTip prop to Menu items

rrfang:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Add the alwaysShowToolTip prop to menu items (Button, MenuItem, MenuItem*) where a tooltip exists so that tooltips will show up in Win32.  

Wasn't sure if the prop needed to be added to the props interface, so I can remove them if they aren't needed.  Also wasn't sure if we should enforce the Platform.OS check as well since the prop will only be read on Win32, so I can add that in if it's needed.

### Verification

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/SPEC.md
+++ b/packages/components/Menu/SPEC.md
@@ -332,13 +332,6 @@ export interface MenuItemProps extends Omit<PressablePropsExtended, 'onPress'> {
    * Do not dismiss the menu when a menu item is clicked
    */
   persistOnClick?: boolean;
-
-  /**
-   * Display the tooltip if one exists
-   * Note the capital "T" in tip as it is needed to match the same prop name in win32 where this is read
-   * @platform win32
-   */
-  alwaysShowToolTip?: boolean;
 }
 ```
 
--- a/packages/components/Menu/SPEC.md
+++ b/packages/components/Menu/SPEC.md
@@ -332,6 +332,13 @@ export interface MenuItemProps extends Omit<PressablePropsExtended, 'onPress'> {
    * Do not dismiss the menu when a menu item is clicked
    */
   persistOnClick?: boolean;
+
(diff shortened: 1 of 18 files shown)


packages/components/Menu/SPEC.md
```diff
@@ -332,6 +332,13 @@ export interface MenuItemProps extends Omit<PressablePropsExtended, 'onPress'> {
    * Do not dismiss the menu when a menu item is clicked
    */
   persistOnClick?: boolean;
+
+  /**
+   * Display the tooltip if one exists
+   * Note the capital "T" in tip as it is needed to match the same prop name in win32 where this is read
+   * @platform win32
+   */
+  alwaysShowToolTip?: boolean;
```
rurikoaraki, line 341: My two cents is that we should set this in FURN and not let the client control it. I don't really see a case where the client would set a tooltip and not want it to show, so we should just set alwaysShowTooltip for the client if tooltip is defined. But I think I would talk to PM for clarification

rrfang: I removed the entry from the spec.md file, but should they also get removed from the ***.types.ts files as well?

rurikoaraki: Yeah. We should remove it from the types file, since that would expose the prop

rurikoaraki (approved):

rurikoaraki:
It might make sense to port this logic to components like Button or Checkbox, too. In theory they can show up in menus

rurikoaraki:
@rrfang Anything holding up this PR from being committed?

PR #2995, opened 2023-08-04
Add E2E A11y and Functional Tests for TabList

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, there isn't any testing for TabList within any pipeline. 

This PR adds accessibility and functional end to end tests that run on win32.

### Verification

Changes pass locally and in pipeline.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+  async resetListSelection(): Promise<void> {
+    return (await this.getTab('First')).click();
+  }
+
+  async isTabSelected(selector: Tab): Promise<boolean> {
+    return await (await this.getTab(selector)).isSelected();
+  }
+
+  async isTabFocused(selector: Tab): Promise<boolean> {
+    return await this.compareAttribute(this.getTab(selector), Attribute.IsFocused, AttributeValue.true);
```
samuelfreiberg, line 27: Missing `await` before this.getTab()

apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+class TabListPageObject extends BasePage {
+  async resetListSelection(): Promise<void> {
+    return (await this.getTab('First')).click();
+  }
+
+  async isTabSelected(selector: Tab): Promise<boolean> {
+    return await (await this.getTab(selector)).isSelected();
+  }
+
+  async isTabFocused(selector: Tab): Promise<boolean> {
```
samuelfreiberg, line 26: I don't think we need to populate functions like this when it's just a wrapper over ` this.compareAttribute()`. It adds unnecessary technical debt.

It's perfectly valid to use `.compareAttribute()` in the spec file

apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+} from '../consts';
+
+type Tab = 'First' | 'Second' | 'Third' | 'Fourth' | 'Fifth';
+
+class TabListPageObject extends BasePage {
+  async resetListSelection(): Promise<void> {
+    return (await this.getTab('First')).click();
+  }
+
+  async isTabSelected(selector: Tab): Promise<boolean> {
```
samuelfreiberg, line 22: I don't think we need to populate functions like this when it's just a wrapper over  this.compareAttribute(). It adds unnecessary technical debt.

It's perfectly valid to use .compareAttribute() in the spec file

apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+    return await (await this.getTab(selector)).isSelected();
+  }
+
+  async isTabFocused(selector: Tab): Promise<boolean> {
+    return await this.compareAttribute(this.getTab(selector), Attribute.IsFocused, AttributeValue.true);
+  }
+
+  async waitForTabSelected(selector: Tab, errorMsg: string, timeout?: number): Promise<boolean> {
```
samuelfreiberg, line 30: Instead of returning `return await this.isTabSelected(selector);`, we can do:
`return await this.waitForCondition(async () => await this.isTabSelected(selector), errorMsg, timeout);`

This improves perf by not having unnecessary calls

apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+    await this.waitForCondition(async () => await this.isTabSelected(selector), errorMsg, timeout);
+    return await this.isTabSelected(selector);
+  }
+
+  async waitForTabFocused(selector: Tab, errorMsg: string, timeout?: number): Promise<boolean> {
```
samuelfreiberg, line 35: Same as above

apps/E2E/src/TabList/pages/TabListPageObject.ts
```diff
@@ -0,0 +1,77 @@
...
+    await this.waitForCondition(async () => await this.isTabFocused(selector), errorMsg, timeout);
+    return await this.isTabFocused(selector);
+  }
+
+  async waitForCallbackToFire(tabKeyPressed: string, errorMsg: string, timeout?: number) {
+    const callbackText = await By(TABLIST_CALLBACK_TEXT);
+    await this.waitForCondition(async () => (await callbackText.getText()) === tabKeyPressed, errorMsg, timeout);
```
samuelfreiberg, line 42: Same as above

samuelfreiberg (changes requested):

apps/E2E/src/TabList/pages/TabListPageObject.ts
lawrencewin: `waitForCondition` returns void, but it does throw an error if the condition isn't met. I removed the return anyway for one less await call.

apps/E2E/src/TabList/pages/TabListPageObject.ts
lawrencewin: `compareAttribute` takes a promise, so no need for an await here.

apps/E2E/src/TabList/specs/TabList.spec.win.ts
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  it('Click and navigate to TabList test page', async () => {
+    await TabListPageObject.navigateToPageAndLoadTests(true);
+    expect(await TabListPageObject.isPageLoaded())
+      .withContext(TabListPageObject.ERRORMESSAGE_PAGELOAD)
+      .toBeTruthy();
+
+    expect(await TabListPageObject.didAssertPopup())
+      .withContext(TabListPageObject.ERRORMESSAGE_ASSERT)
```
samuelfreiberg, line 27: What is `.withContext()` for? Should we make a task to normalize this pattern?

lawrencewin: Before, we were passing error messages in the `.toBeTruthy` / `.toBeFalsy` methods within specs. This is deprecated in Jasmine - `withContext` is the correct method to pass the error message if the test fails.

lawrencewin: It would probably be a good idea to normalize this across e2e tests.

samuelfreiberg: Awesome! Room to improve - do you mind creating a task for this? Good error messages are crucial :)

apps/E2E/src/TabList/specs/TabList.spec.win.ts
```diff
@@ -0,0 +1,190 @@
...
+  beforeEach(async () => {
+    await TabListPageObject.scrollToTestElement();
+  });
+
+  it('Validate TabList\'s "accessibilityRole" defaults to "ControlType.Tab"', async () => {
+    expect(
+      await TabListPageObject.compareAttribute(TabListPageObject._primaryComponent, Attribute.AccessibilityRole, TAB_A11Y_ROLE),
+    ).toBeTruthy();
+
+    expect(await TabListPageObject.didAssertPopup())
```
samuelfreiberg, line 42: We don't need these Assert checks for tests where we're only querying elements. We're not performing any UI patterns. Shouldn't cause any asserts. Improves performance by removing unnecessary calls.

samuelfreiberg (approved):

PR #3001, opened 2023-08-08
[Callout][macOS] Add new prop `material`

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We have a hardcoded [NSVisualEffectView](https://developer.apple.com/documentation/appkit/nsvisualeffectview) as the root / content view of macOS Callout. We also hardcoded it's [NSVisualEffectView.material](https://developer.apple.com/documentation/appkit/nsvisualeffectview/1526623-material) to be of type `.menu`. We now have a use case where we want a different material type (`popover`). Let's just expose the enum as a prop, so that JS can set it. Let's also set the default to `.menu` for backwards compatibility.

### Verification

The same Callout with two different types of materials: The second one (popover) is _slightly_ less opaque

Screenshot 2023-08-07 at 10 54 29 PM
Screenshot 2023-08-07 at 10 55 14 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Callout/src/Callout.types.ts
```diff
@@ -212,6 +228,13 @@ export interface ICalloutProps extends IViewProps, ICalloutTokens {
    * anchoring to a point inside the component.
    */
   target?: React.RefObject<React.Component> | string;
+
+  /**
+   * Determines the material of the NSVisualEffectView that is the contentView of the Callout.
```
PPatBoyd, line 233: After our discussion yesterday, did you find that we couldn't rely on the content to set the NSVisualEffectView because having a default at one material value e.g. `menu` was [multiplicative or otherwise not compatible] with the content being able to achieve the desired non-default material effect e.g. `popover`?

Saadnajmi: I haven't tested that yet, I'll spin up a test app and see. It will depend on the [blending mode](https://developer.apple.com/documentation/appkit/nsvisualeffectview#1674173) of the two VisualEffectViews in question. In general, that feels a bit like undefined-behavior-land and once that could change between OS releases (It wouldn't be the first time), too I'm not a fan of a solution that involves placing two NSVisualEffectViews on top of each other.

Saadnajmi: OK, I did some testing. There are two blending modes: `.behindWindow` and `.withinWindow`. I have a long vertical red column, with two VisualEffectViews on top to test what happens:

## 1) behindWindow

Screenshot 2023-08-08 at 4 31 16 PM

VisualEffectView ignores the view heirarchy and blends content directly behind the window. This means we can layer as many VisualEffectViews as we want, and is the default that Callout uses currently.

## 2) withinWindow 

Screenshot 2023-08-08 at 4 31 57 PM

There is some multiplicativeness happening, which is undesirable.

Saadnajmi: Conclusion: It seems that for VisualEffectViews with `.behindWindow` blending (the Callout default), you CAN stack them and the top-most one wins. That being said, I would still like to add this prop to control the Callout's root VisualEffectView.. I don't like relying on the stacking behavior + feels weird to need to add a new view when I could just modify the existing one.

Saadnajmi: Gonna go with closing this PR and making https://github.com/microsoft/fluentui-react-native/pull/3026 instead

Saadnajmi:
Closing in favor of https://github.com/microsoft/fluentui-react-native/pull/3026

PR #3023, opened 2023-08-10
Color tokens test page updates

lyzhan7:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

1. The color tokens test page was showing either 1) tokens from the win32 theme if the theme was a win32 theme or 2) tokens created using default theme code. For macOS, we were showing macOS tokens, but these tokens were going through default theme code rather than macOS theme code. This resulted in some expected tokens not showing up, and in some tokens showing up as undefined. Let's just show all the tokens in theme - this should at least make sense for every platform. 
2. I noticed that the swatch color wasn't updating when switching from light/dark mode. This was because when we were re-rendering after a light/dark mode switch, we would check first to see if a color with that same colorName was defined already (which it would be, but with the wrong light/dark mode). I updated this code so that we just update the color of the swatch every time we render - if there's a better way to do this though please let me know. 

Notes: 
- After this change, at least on macOS, there are more tokens shown in the color token test page. This is because the tokens we expose through the theme contain 1) tokens from the pipeline/design-tokens repo, as well as 2) tokens we define in FURN, which we should remove at some point. Previous to this change we were only showing tokens from 1), but now we are showing both. All the 2) tokens should be removed at some point.
- Some of the 2) tokens (defined in FURN) show up as [object Object] - either because they are DynamicColorMacOS or they are PlatformColor. Given that we should remove them at some point, I've left them as is. 

### Verification

Before:

https://github.com/microsoft/fluentui-react-native/assets/78454019/f26c31a3-e44a-4d92-b6da-12ea9f16d37e

After:

https://github.com/microsoft/fluentui-react-native/assets/78454019/70318b6c-eb1f-4c20-84cb-f0c55fc1bec9

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/src/TestComponents/ColorTokens/ColorTokenTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/ColorTokens/ColorTokenTest.tsx
@@ -53,11 +53,6 @@ const styles = StyleSheet.create({
   colorDescriptionNamePadding: { paddingRight: 5 },
 });
 
-const getSwatchColorStyle = (colorName: string, colorValue: ColorValue): ViewStyle => {
-  styles[colorName] = styles[colorName] || { backgroundColor: colorValue };
-  return styles[colorName];
-};
-
 type ColorTokenProps = { colorValue: ColorValue; colorName: string };
 const ColorToken: React.FunctionComponent<ColorTokenProps> = (p: ColorTokenProps) => {
   if (p.colorValue === undefined) {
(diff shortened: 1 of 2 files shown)


Saadnajmi:
> - either because they are DynamicColorMacOS or they are PlatformColor. Given that we should remove them at some point, I've left them as is.

I don't think that's necessarily true? I thought a goal might be to have support for them in the token system the way windows/win32 do right now.

Saadnajmi (approved):

lyzhan7:
> > * either because they are DynamicColorMacOS or they are PlatformColor. Given that we should remove them at some point, I've left them as is.
> 
> I don't think that's necessarily true? I thought a goal might be to have support for them in the token system the way windows/win32 do right now.

Not sure about PlatformColor - but pretty sure we want to remove DynamicColorMacOS and rely on the light/dark theme tokens instead

PR #3026, opened 2023-08-11
[macOS] Introduce VibrancyView

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On macOS, a large part of the design system is the concept of [Vibrancy](https://developer.apple.com/design/human-interface-guidelines/materials#macOS), which is largely controlled and defined by [NSVisualEffectView](https://developer.apple.com/documentation/appkit/nsvisualeffectview). We don't have a generic native component exposed for NSVisualEffectView, so let's create one: VibrancyView! The name was chosen as it matches well with the iOS equivalent React Native component in [@react-native-community/react-native-blur](https://github.com/Kureev/react-native-blur/tree/master), which might be a potential place to move or copy this native component, should we want to.

This also gives us the side benefit that, with `blendingMode={behindWindow}`, we can overlay our VibrancyView on top of Callout (which had a hardcoded NSVisualEffectView with material `'menu`) and effectively render a different material, making #3001 redundant. 

There were two special considerations taken to make our native component more usable on React Native macOS:

1) Our Native `VibrancyView` is actually an `RCTView` with a `NSVisualEffectView` subview. This is done so that our native component inherits all of ViewProps. In order to make sure all children are subviews of our `NSVisualEffectView`, we override `insertReactSubview`.

2) While writing this native component, I found there were lots of issues with getting rendering and register clicks properly. This is due to differences in how UIKit (used by React Native on iOS) and Appkit (used by React Native macOS) handle hit testing and coordinate systems. In React Native macOS, we have custom code in RCTView / RCTTouchHandler/etc to mitigate that, but an arbitrary native component derived from `NSView` will not. Therefore, instead of wrapping `NSVisualEffectView` directly, I wrote a `FixedVisualEffectView` class with some extra code to work with React Native macOS and other RCTViews properly. I put in as much comments as I can to explain what is happening. Many thanks to @lenaic for helping me find solutions here.

### Verification

Wrote a test page with a lot of tests

https://github.com/microsoft/fluentui-react-native/assets/6722175/a2dd751e-445f-4bf7-9eb2-4180bc57de0f

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/VisualEffectView/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "lint": "fluentui-scripts eslint",
+    "test": "fluentui-scripts jest",
+    "update-snapshots": "fluentui-scripts jest -u",
+    "prettier": "fluentui-scripts prettier",
+    "prettier-fix": "fluentui-scripts prettier --fix true"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/experimental/popover"
```
rurikoaraki, line 27: Please update

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
```diff
@@ -91,6 +92,10 @@ export const useMenuPopover = (props: MenuPopoverProps): MenuPopoverState => {
     };
   });
 
+  const vibrancyViewProps: VibrancyViewProps = {
```
rurikoaraki, line 95: These changes could have been in their own PR separate from implementing VibrancyView

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,224 @@
...
+});
+
+const VibrancyViewDefault: React.FunctionComponent = () => {
+  const viewProps: ViewProps = {
+    focusable: true,
+    style: [styles.box, styles.borderProps],
+  };
+  return (
+    <Stack style={stackStyle}>
+      <View style={[commonStyles.root, styles.spacing]}>
```
rurikoaraki, line 72: nit: should be a const defined outside of the component or memoized

rurikoaraki: Apply to other Views with inline style arrays as well

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,224 @@
...
+  });
+
+  return (
+    <Stack style={stackStyle}>
+      <View style={styles.Rectangle1}>
+        <VibrancyView material={material} blendingMode={blendingMode} state={state} style={styles.Rectangle2}>
+          <MenuPicker
+            prompt="Material"
+            selected={material || 'undefined'}
+            onChange={(material) => setMaterial(material)}
```
rurikoaraki, line 148: should be a useCallback hook

rurikoaraki: Same with other onChange inline functions

rurikoaraki:
Only reviewed the TS

packages/components/Menu/src/MenuPopover/useMenuPopover.ts
Saadnajmi: I can factor out any changes to Callout and menu I think and keep it simpler

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
Saadnajmi: I didn't think that applied to array styles? But sure

rurikoaraki: The object passed in would be regenerated on every render (since the array object would be newly created), causing a wasted render assuming the style/other props didn't change

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,225 @@
...
+        <Text>Vibrancy View with no props set</Text>
+        <VibrancyView style={styles.box} />
+      </View>
+    </Stack>
+  );
+};
+
+const VibrancyViewWithViewProps: React.FunctionComponent = () => {
+  const descriptionString = '<View> and <VibrancyView> with the same props set';
+  const viewProps: ViewProps = {
```
rurikoaraki, line 76: nit: this will still be a new object on every render, so it should be memoized or a const declared outside of the component.

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,225 @@
...
+        </VibrancyView>
+      </View>
+    </Stack>
+  );
+};
+
+const HitTestButton = (props: { title: string }) => {
+  return (
+    <Button
+      onClick={(_e) => {
```
rurikoaraki, line 168: nit: should be useCallback

lyzhan7 (approved):

packages/experimental/VibrancyView/macos/VibrancyView.swift
```diff
@@ -0,0 +1,40 @@
...
+	@objc public func setMaterial(_ material: NSVisualEffectView.Material) {
+		visualEffectView.material = material
+	}
+
+	@objc public func setBlendingMode(_ blendingMode: NSVisualEffectView.BlendingMode) {
+		visualEffectView.blendingMode = blendingMode
+	}
+
+	@objc public func setState(_ state: NSVisualEffectView.State) {
+	  visualEffectView.state = state
```
lyzhan7, line 36: nit - whitespace on this line doesn't match

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,225 @@
...
+      focusable: true,
+      style: [styles.box, styles.borderProps],
+    };
+  }, []);
+  return (
+    <Stack style={stackStyle}>
+      <View style={styles.rootWithSpacing}>
+        <Text>{descriptionString}</Text>
+        <VibrancyView {...viewProps} />
+        <View {...viewProps} />
```
lyzhan7, line 87: Maybe switch the order of VibrancyView and View here to match the description? And also would be great if they were side by side instead of stacked.

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
```diff
@@ -0,0 +1,225 @@
...
+      <View style={styles.rootWithSpacing}>
+        <Text>Vibrancy View with no props set</Text>
+        <VibrancyView style={styles.box} />
+      </View>
+    </Stack>
+  );
+};
+
+const VibrancyViewWithViewProps: React.FunctionComponent = () => {
+  const descriptionString = '<View> and <VibrancyView> with the same props set';
```
lyzhan7, line 75: Might make it clearer what we're looking at if you specified what props are being set?

lyzhan7:
Add video for dark mode?

apps/fluent-tester/src/TestComponents/VibrancyView/VibrancyViewTest.tsx
Saadnajmi: Done! Couldn't figure out how to get a space between the two though..

Saadnajmi:
> Add video for dark mode?

updated

PR #3031, opened 2023-08-14
Have postBump also regenerate lockfile

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

As part of moving to yarn 3, the local packages are now in the lockfile, which means every time we bump versions we need to regenerate the lockfile. Having to do this manually would be tedious and probably block things, so adding a step to beachball to do it for us.

I think this change will work, but if not I can migrate this to use postPublish instead. The benefit of using postBump is that changes will be committed automatically, while with postPublish I'd have to manually commit them instead as part of the hook.


Saadnajmi (approved):

beachball.config.js
```diff
@@ -18,6 +18,8 @@ module.exports = {
       if (name === '@fluentui-react-native/dependency-profiles') {
         console.log(`Updating ${name} to use latest published versions`);
         execSync(`yarn update-profile`, { cwd: packagePath });
+        console.log('Updating lockfile');
+        execSync(`yarn install  --mode update-lockfile`);
       }
     },
```
Saadnajmi, line 24: Note: This will only update if `@fluentui-react-native/dependency-profiles` is updated

rurikoaraki: This should be ok since the hook is called after bump has been run on all packages.

Saadnajmi (approved):

PR #3056, opened 2023-08-23
Implement Tablist Animations on win32

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR implements a key feature that was not a part of previous FURN Tab implementations - an animated indicator for win32. 

- Wrote a new component, TabListAnimatedIndicator. This is a simple component that is displayed and animated as the selected tab's indicator. 
- Rewrote the Tab component to use the compressible framework. This was done to get access the Tab's tokens, which are needed to calculate the layout values of the TabListAnimatedIndicator and to style the Tabs appropriately so their static indicators don't clash with the animated one. 
  - Before, the static indicator was a separate component for organization purposes. With the compressible rewrite, this component was removed and its logic was consolidated into Tab. 
- Wrote hook logic to calculate the animated indicator's styles, used in `useTabList()`.
- Reorganized the package.

### Verification

https://github.com/microsoft/fluentui-react-native/assets/15683103/3dc999ff-f27c-4cac-8987-07f9b069f7de

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/TabList/src/Tab/Tab.styling.ts
```diff
@@ -1,77 +1,120 @@
...
+import type { TabProps, TabSlotProps, TabTokens } from './Tab.types';
+import type { TabListContextData } from '../TabList/TabList.types';
+
+/**
+ * Hook to get the style props for each Tab slot.
+ */
+export const useTabSlotProps = (props: TabProps, tokens: TabTokens, theme: Theme, context: TabListContextData): TabSlotProps => {
+  const { vertical } = context;
+
+  const indicatorColor = React.useMemo(() => {
```
rurikoaraki, line 19: nit: you don't need to useMemo for primitives

rurikoaraki:
FWIW I think you can mix and match compose and compressible. I don't think you need to convert the entire component to use compressible if only one component needs to access token values.

packages/experimental/TabList/src/TabListAnimatedIndicator/TabListAnimatedIndicator.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * This component renders as the indicator for the selected tab. Its styles are calculated and passed
+ * from the useTabList hook, so it doesn't need to use the compose or compressible franework.
+ */
+export const TabListAnimatedIndicator = stagedComponent((props: TabListAnimatedIndicatorProps) => {
+  return (final: TabListAnimatedIndicatorProps) => {
+    const { styles, ...finalProps } = final;
+    const rootProps = mergeProps(props, finalProps, { style: styles.container });
+    const indicatorProps = getIndicatorProps('Ribbon_TabUnderline', styles.indicator);
```
rurikoaraki, line 25: This will only work for win32, so it might be worth forking this component to return null on non-win32 platformw

packages/experimental/TabList/src/Tab/useTabAnimation.ts
```diff
@@ -0,0 +1,67 @@
...
+        animatedIndicatorState?.tablistLayout &&
+        animatedIndicatorState.tablistLayout.width > 0 &&
+        e.nativeEvent.layout.height <= animatedIndicatorState.tablistLayout.height &&
+        e.nativeEvent.layout.height < RENDERING_HEIGHT_LIMIT
+      ) {
+        let width: number, height: number;
+        // we can calculate the dimensions of the indicator using token values we can access via the compressable framework
+        if (vertical) {
+          width = tokens.indicatorThickness;
+          height = e.nativeEvent.layout.height - 2 * (tokens.indicatorInset + tokens.borderWidth + 1); // the '1' is to account for padding between the focus border and tab
```
rurikoaraki, line 48: nit: magic numbers. 2 and 1 in this line can be defined as consts with meaningful names

packages/experimental/TabList/src/TabListAnimatedIndicator/TabListAnimatedIndicator.types.ts
```diff
@@ -0,0 +1,33 @@
...
+export type ListLayoutMap = { [tabKey: string]: TabLayoutInfo };
+
+export interface AnimatedIndicatorStyles {
+  container: ViewStyle;
+  indicator: ViewStyle;
+}
+export type AnimatedIndicatorStylesUpdate = Partial<AnimatedIndicatorStyles>;
+
+export interface AnimatedIndicatorState {
+  addToLayoutMap: (tabKey: string, layout: TabLayoutInfo) => void;
```
rurikoaraki, line 29: Hmm... something seems weird to me to have an object named "state" to have functions on it. I feel like this could be an object with styles and layout info and addToLayoutMap and updateStyles could be standalone functions. Neither of the functions depend on component information so they shouldn't need useCallback

lawrencewin:
> FWIW I think you can mix and match compose and compressible. I don't think you need to convert the entire component to use compressible if only one part needs to access token values.

How would mixing and matching the two frameworks work? Is there an example already on FURN?

rurikoaraki:
> > FWIW I think you can mix and match compose and compressible. I don't think you need to convert the entire component to use compressible if only one part needs to access token values.
> 
> How would mixing and matching the two frameworks work? Is there an example already on FURN?

In theory any compressible component having children that are made of the compose framework would be doing something similar, like MenuPopover having a MenuList child. What I'm saying is that just because TabIndicator needs compressible doesn't mean Tab needs to also. I don't really see why Tab had to be converted, although, since you did the work I think it's fine to leave it that way

packages/experimental/TabList/src/Tab/useTabAnimation.ts
```diff
@@ -0,0 +1,67 @@
...
+  }, [tabKey, selectedKey, tokens.indicatorColor]);
+
+  // onLayout callbacks to help calculate positioning of the animated indicator
+  const onTabLayout = React.useCallback(
+    (e: LayoutEvent) => {
+      // If the layout values of the tab aren't set or are wonky, don't update state
+      if (
+        e?.nativeEvent?.layout &&
+        animatedIndicatorState?.tablistLayout &&
+        animatedIndicatorState.tablistLayout.width > 0 &&
```
PPatBoyd, line 40: did you ever determine the cause of getting a layout result of width = 0 for horizontal tabs?  Did vertical tabs have their own version of this issue?

packages/experimental/TabList/src/Tab/__tests__/__snapshots__/Tab.test.tsx.snap
```diff
@@ -94,9 +95,6 @@ exports[`Tab component tests Customized Tab 1`] = `
     >
       <Text
         accessible={false}
-        ellipsizeMode="tail"
-        numberOfLines={0}
```
PPatBoyd, line 98: so intrigued on where the numberOfLines=0 came from before 🤔

packages/experimental/TabList/src/Tab/Tab.tsx
```diff
@@ -28,74 +33,84 @@ const tabLookup = (layer: string, state: TabState, props: TabProps, tablistConte
...
-    };
-  },
-});
+    return (
+      <RootSlot {...mergedProps}>
+        <StackSlot>
+          {icon && <IconSlot {...icon} />}
+          {hasChildren && (
+            <ContentContainerSlot>
+              {React.Children.map(children, (child, i) =>
```
PPatBoyd, line 95: this is a bit of a drive-by comment since it's the same as before, but what scenario was this used for? I'm intrigued at why we can't natively pass in {children} and the effect on varying combinations (more than one string child or mixed types)

PPatBoyd (commented):
Ideally we'd root out or at least understand why we're getting intermediate layout values that are "wonky" so that we can clarify if it's win32-only or a cross-platform issue, why we aren't fixing the issue now (e.g. if there's an endemic issue with layout on win32 that we can't resolve), etc setting ourselves up for future success

lawrencewin:
> Ideally we'd root out or at least understand why we're getting intermediate layout values that are "wonky" so that we can clarify if it's win32-only or a cross-platform issue, why we aren't fixing the issue now (e.g. if there's an endemic issue with layout on win32 that we can't resolve), etc setting ourselves up for future success

The issue stems from rendering the test components on the tester app inside a ScrollView. I'm not too familiar with win32 ScrollView logic, but it looks like when figuring layout, the ScrollView first sets its childrens' width to 0 before later setting an actual width value. Layout values appear normal when rendering outside the ScrollView on win32, and they appear normal when testing on Android.

packages/experimental/TabList/src/Tab/useTabAnimation.ts
lawrencewin: See comment below.

packages/experimental/TabList/src/Tab/Tab.tsx
lawrencewin: Talked about offline, this is used to pass in both string and non-string children and have them render correctly.

PPatBoyd (approved):

packages/experimental/TabList/src/TabList/useTabList.ts
```diff
@@ -53,9 +54,17 @@ export const useTabList = (props: TabListProps): TabListInfo => {
     [setTabKeys],
   );
 
+  // Logic to style the animated indicator
+  const { addTabLayout, onTabListLayout, tabListLayout, styles, updateStyles } = useAnimatedIndicator(
```
rurikoaraki, line 58: Why is the tab list generating the animated indicator's styles? shouldn't the indicator do that itself?

lawrencewin: Initially, I tried calling the hook inside the indicator. The hook needs to know the selected key and whether the TabList is vertical or horizontal. This is info passed through the TabList context, but the issue is that when trying to access the context in the indicator component, the context never updates - its values are the same as they are when we initially define it via `React.createContext()`. The indicator is also a child of the context provider, unless I'm hallucinating which makes this issue all the more strange. 

The hook also has to run in the first render, so my work around is to run it in the TabList component and pass its results to the indicator.

rurikoaraki: That sounds like a bug that you should look into. To me that would mean that the context provider isn't getting the value prop as a new object when the properties update.

lawrencewin: It's strange because the TabList context inside of the Tab component updates as expected. It's just for this specific component. I'm not really sure what to make of this bug.

lawrencewin: @rurikoaraki Did some investigation - it seems like the bug comes from calling `useContext` inside a slot component which uses the framework (`stagedComponent`, `compose`, `compressible`). Making the indicator a non-slot component fixes the issue, so I was able to move the styling logic into the `useAnimatedIndicator` hook

packages/experimental/TabList/src/TabList/useTabList.ts
```diff
@@ -53,21 +55,35 @@ export const useTabList = (props: TabListProps): TabListInfo => {
...
+  // State variables and functions for saving layout info and other styling information to style the animated indicator.
+  const [listLayoutMap, setListLayoutMap] = React.useState<{ [key: string]: TabLayoutInfo }>({});
+  const [tabListLayout, setTabListLayout] = React.useState<LayoutRectangle>();
+  const [userDefinedAnimatedIndicatorStyles, setUserDefinedAnimatedIndicatorStyles] = React.useState<AnimatedIndicatorStyles>({
+    container: {},
+    indicator: {},
+  });
+
+  const addTabLayout = (tabKey: string, layoutInfo: TabLayoutInfo) => {
+    setListLayoutMap((prev) => ({ ...prev, [tabKey]: { ...prev[tabKey], ...layoutInfo } }));
```
rurikoaraki, line 67: >...prev[tabKey] [](http://example.com/codeflow?start=55&length=15)

Is there a reason why the old results should be kept around?

packages/experimental/TabList/src/TabList/useTabList.ts
```diff
@@ -53,21 +55,35 @@ export const useTabList = (props: TabListProps): TabListInfo => {
...
+  };
+
+  const updateStyles = (update: AnimatedIndicatorStylesUpdate) =>
+    setUserDefinedAnimatedIndicatorStyles((prev) => {
+      const newStyles: AnimatedIndicatorStyles = { ...prev };
```
rurikoaraki, line 72: >const newStyles: AnimatedIndicatorStyles = { ...prev }; [](http://example.com/codeflow?start=6&length=55)

Before running setUserDefinedAnimatedIndicatorStyles, we should probably make sure that one of update.container or update.indicator is defined, otherwise this will cause wasted work

packages/experimental/TabList/src/TabListAnimatedIndicator/useAnimatedIndicatorStyles.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { animatedIndicatorStyles, layout, selectedKey, vertical } = React.useContext(TabListContext);
+
+  const selectedIndicatorLayout = React.useMemo<TabLayoutInfo | null>(() => {
+    return selectedKey ? layout.tabs[selectedKey] : null;
+  }, [selectedKey, layout.tabs]);
+
+  // Calculate styles using both layout information and user defined styles
+  const styles = React.useMemo<AnimatedIndicatorStyles | null>(() => {
+    // if not all layout props have been recorded for the current selected indicator, don't render the animated indicator
+    if (selectedIndicatorLayout) {
```
rurikoaraki, line 21: Code will be a little less nested if you instead put the "bail" condition at the top of the function. Ie

if (!selectedIndicatorLayout) {
    return null;
}

... rest of function


Imo it also makes the function easier to read since it will list all the reasons for exiting early first.

packages/experimental/TabList/src/Tab/Tab.tsx
```diff
@@ -28,74 +33,84 @@ const tabLookup = (layer: string, state: TabState, props: TabProps, tablistConte
...
-
-    // Return the handler to finish render.
-    return (final: TabProps, ...children: React.ReactNode[]) => {
-      if (!tab.state) {
-        return null;
-      }
+const getRootProps = memoize(rootPropsWorker);
+function rootPropsWorker(props: PressablePropsExtended, onLayout: (e: LayoutEvent) => void) {
+  return mergeProps(props, { onLayout });
+}
```
rurikoaraki, line 39: nit: Probably should be a hook in useTab since it's not in the inner render?

packages/experimental/TabList/src/Tab/Tab.styling.ts
```diff
@@ -1,77 +1,116 @@
...
         style: {
-          display: 'flex',
-          alignItems: 'center',
-          flexDirection: 'row',
-          alignSelf: 'flex-start',
-          justifyContent: 'center',
-          marginHorizontal: tokens.stackMarginHorizontal,
-          marginVertical: tokens.stackMarginVertical,
+          flex: 1,
+          borderRadius: 99,
```
rurikoaraki, line 107: > 99, [](http://example.com/codeflow?start=23&length=4)

nit: could be global token

packages/experimental/TabList/src/Tab/Tab.styling.ts
```diff
@@ -100,7 +100,7 @@ export const useTabSlotProps = (props: TabProps, tokens: TabTokens, theme: Theme
     // if we're the selected tab and we've generated styles for the animated indicator, render the static tab indicator as transparent.
     // The animated indicator will receive styling instead via useTabAnimation hook.
     () => {
-      const hideStaticIndicator = props.tabKey === selectedKey && !!context.animatedIndicatorStyles;
+      const hideStaticIndicator = props.tabKey === selectedKey && indicatorHasRendered;
```
rurikoaraki, line 103: `indicatorHasRendered` is confusing to me here, it's not clear it's talking about the animatedIndicator. Can we rename this to something more obvious like "canShowAnimatedIndicator"?

packages/experimental/TabList/src/TabListAnimatedIndicator/useAnimatedIndicatorStyles.ts
```diff
@@ -45,5 +45,7 @@ export function useAnimatedIndicatorStyles(): AnimatedIndicatorStyles {
     };
   }, [vertical, selectedIndicatorLayout, animatedIndicatorStyles]);
 
+  React.useEffect(() => setIndicatorHasRendered(styles !== null), [setIndicatorHasRendered, styles]);
```
rurikoaraki, line 48: Is there a comment somewhere that explains that there's a static indicator rendered to cover the late render of the animated indicator? If not it would be great to add a comment to make the intent of this variable clear.

lawrencewin: There's a comment in Tab.styling.tsx, but I can add one here as well.

rurikoaraki: Ah ok. I would say it would be nice to add more about the intent of the static indicator, i.e. it exists to cover the animated indicator "popping in," rather than just that it'll be swapped out once styles are generated.

rurikoaraki: New comment looks great, thanks!

rurikoaraki (approved):

PR #3076, opened 2023-08-31
Fail E2E Tests on First Failure

samuelfreiberg:
### Platforms Impacted
- [X] iOS
- [X] macOS
- [X] win32 (Office)
- [X] windows
- [X] android

### Description of changes

Previously, we didn't stop E2E tests from running after a failure. In my opinion, we should stop the testing on the first failure to fail swiftly. Rather than continuing test execution when we know the suite will fail, we'll stop test execution on the first failure and save developer time.

Also, this allows us to retry the spec directly after the failure, rather than running all subsequent tests and **then** re-running.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
I do think there can be value in letting the test continue running since it will mean you can see all the failures at once, but I think the likelihood that a change will cause more than one test to fail is low so I agree that change makes sense.

rurikoaraki (approved):

samuelfreiberg:
> I do think there can be value in letting the test continue running since it will mean you can see all the failures at once, but I think the likelihood that a change will cause more than one test to fail is low so I agree that change makes sense.

Thanks for the input! We can see how it goes for a little while - if we feel as though running all tests is better, we can change it back :)

PR #3083, opened 2023-09-05
Native DatePicker Android

ankarm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [x ] android

### Description of changes

(Work in Progress) Wrapping FluentUI-Android native datepicker control as a react native Native Module.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
@Saadnajmi can you chime in from an iOS perspective?

rurikoaraki:
Please run "yarn change" at repo root and push the result.

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
```diff
@@ -1,2 +1,11 @@
-export const NativeDatePicker = null;
-console.warn('DatePicker not supported');
+export const NativeDatePicker = {
```
rurikoaraki, line 1: You nshould make a .android version of this file and leave the NativeDatePicker.tsx version as is. Otherwise the NativeDatePicker will not error on other unsupported platforms like macOS.

packages/experimental/NativeDatePicker/src/NativeDatePicker.types.shared.ts
```diff
@@ -0,0 +1,9 @@
+export type DatePickerSharedParameterObject = {
+  startDate?: Date;
+  endDate?: Date;
+  callback: (startDate: string, endDate: string) => void
+}
+
+export type NativeDatePickerSharedInterface = {
```
rurikoaraki, line 7: Please use ".mobile.ts" file name for files shared on mobile.

packages/experimental/NativeDatePicker/src/NativeDatePicker.tsx
rurikoaraki, line 1: This should not be an object when it's defined as a component on other endpoints. Otherwise code shared between endpoints will probably have build errors.

I'm also weary of the types being very different between iOS and Android for the same reason - if a client uses the same code to access the NativeDatePicker for both iOS and Android, there will be build issues for them. If having different type interfaces is strictly necessary, it's probably worth splitting the NativeDatePicker into NativeDatePickerIos and NativeDatePickerAndroid. However, in general in FURN, we try to keep the props API aligned between all platforms of a component.

@saadnajmi in case he has additional thoughts

Saadnajmi: I think the iOS one is actually a native module, not a native component. It's a module with a method that will present a sheet with a popover, but the actual view / presentation happens all in native. If we're going to have different types, I agree it should just be DatePickerIOS and DatePickerAndroid.

ankarm: While working with the two date pickers (i.e. Android and iOS), I noticed that the iOS native component from FluentUI-Apple repo takes in a lot of customizable parameters like `referenceStartDate`, `referenceEndDate` (to customize the allowed dates to be selected in the calendar), and also `startTitle`, `endTitle`, etc for custom titles in the calendar modal. 
The Android native component from FluentUI-Android repo does not have so much customizability. It only takes as input the type of picker (i.e. Date / DateTime), and the range mode (i.e. None, Start, End).

Due to this significant disparity in the number of parameters accepted by the date pickers in the two platforms, I too believe it would be best to split this into `DatePickerIOS` and `DatePickerAndroid`. 

I will push an update after refactoring.

ankarm: Pushed a change with code refactored to move common params out into a common types file. I decided against splitting the component into Android and iOS specific ones because of the fact that most of the iOS parameters don't work (probably due to a bug) leaving only the common set of parameters working for both.
* mode
* dateRangeMode (android only)
* startDate
* endDate
* callback
* referenceStartDate (iOS only)
* referenceEndDate (iOS only)

apps/fluent-tester/src/TestComponents/NativeDatePicker/NativeDatePickerTest.tsx
```diff
@@ -11,7 +11,114 @@ import { stackStyle, commonTestStyles as commonStyles } from '../Common/styles';
 import type { PlatformStatus, TestSection } from '../Test';
 import { Test } from '../Test';
 
-const NativeDatePickerMainTest: React.FunctionComponent = () => {
+const NativeDatePickerAndroidUsage: React.FunctionComponent = () => {
```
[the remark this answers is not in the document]
ankarm: Noted! Will have a followup PR for this refactoring.

rurikoaraki (approved):

Saadnajmi (approved):

Saadnajmi:
@rohanprasadofficial / @ayush547 in case y'all have followup comments, please let us know and we can log them!

PR #3091, opened 2023-09-14
E2E Reliability + Perf Improvements

samuelfreiberg:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

_I know this PR contains a lot of files changed, but that's because I changed some common patterns used in all specs. Read below for more info_

This PR does **not** add or remove any tests. It's purely aimed to clean up the code, improve code readability, and reliability.

There are many cases where the pattern we use do not efficiently utilize WebDriverIO helpers to their max capability. For example, `waitUntil()` is the most common helper we use. This method either returns true (if the condition is met) or throws an error if it does not. We do not utilize this correctly - in many cases, we call this method, don't read the return value, and then make another API call to return the same exact condition it checked for. Additionally, we double-up on the error message (we don't need an error message in the `expect()` call if the method we're calling contains `waitUntil()` because waitUntil will throw an error if the condition is not met.

Additionally, some logic in the tests were not sound or efficient. Making extra API calls, missing `await`'s, and performing unnecessary UI actions.

Also removes didAssertPopup() checks from accessibility tests. In accessibility tests, we don't perform any UI manipulation, we're only querying the app for values. No need to check for asserts.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki (approved):

Saadnajmi (approved):

Saadnajmi:
> Also removes didAssertPopup() checks from accessibility tests. In accessibility tests, we don't perform any UI manipulation, we're only querying the app for values. No need to check for asserts.

Don't we do that just to be sure no Redbox showed up ?

PR #3099, opened 2023-09-21
Remove four unused size tokens from test page

TravisSpomer:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

FURN was the first to implement the Fluent design tokens for size. Since then, the largest four have been removed from Fluent. Since they aren't used anywhere except for this page, it seems acceptable to remove them from FURN even though it's *technically* a breaking change. This just updates the test page so that it doesn't expect `globalTokens.size640` to exist so I don't break your build; an upcoming change to the source token data will then actually remove `size640` from what eventually gets turned into `globalTokens`.


lenahong (approved):

samuelfreiberg (approved):

gulnazsayed (approved):

Saadnajmi (approved):

TravisSpomer:
Thanks all! I added a change file so I think we're good now but someone else will have to click the merge button for me.

PR #3104, opened 2023-09-26
Add FocusZone around MenuList in win32

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

If there are non-MenuList items in a Menu, they are included in the menu's innate arrow loop due to how win32 native configures the native arrowing. In order to ensure that MenuLists get their own grouping for focus, we should add the FocusZone in win32.

### Verification

Added test. Also tested scroll view case and group case.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -44,23 +43,13 @@ export const MenuList = compose<MenuListType>({
   slots: {
     root: MenuStack,
     scrollView: ScrollView,
-    ...(Platform.OS === 'macos' && { focusZone: FocusZone }),
+    focusZone: Platform.OS !== 'android' ? FocusZone : React.Fragment,
```
Saadnajmi, line 46: Might be more accurate to do like a `supportsFocusZone = ['macos', 'win32'] and check those two so that other platforms like windows and iOS don't immediately Redbox.

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -85,33 +74,34 @@ export const MenuList = compose<MenuListType>({
         return child;
       });
 
-      const content =
-        Platform.OS === 'macos' ? (
-          <Slots.root onMouseLeave={setFocusZoneFocus} onKeyDown={menuList.onListKeyDown}>
-            <Slots.scrollView
-              showsVerticalScrollIndicator={menuContext.hasMaxHeight}
-              showsHorizontalScrollIndicator={menuContext.hasMaxWidth}
+      const shouldHaveScrollView = Platform.OS === 'macos' || menuList.hasMaxHeight || menuList.hasMaxWidth;
```
Saadnajmi, line 77: We can revisit later, but it might actually be an improvement on macOS to only have a scrollview if maxWidth//height are defined.. aka we could maybe take out the platform check from here lol

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -85,33 +74,34 @@ export const MenuList = compose<MenuListType>({
...
+      const ScrollViewWrapper = shouldHaveScrollView ? Slots.scrollView : React.Fragment;
+      const shouldHaveFocusZone = Platform.OS !== 'android';
```
Saadnajmi, line 79: would flip this check, like I said above

packages/components/Menu/src/MenuList/MenuList.tsx
```diff
@@ -85,33 +74,34 @@ export const MenuList = compose<MenuListType>({
...
+          >
+            <Slots.focusZone
+              // avoid error that fires when props are passed into React.fragment
+              {...(shouldHaveFocusZone && {
+                componentRef: menuList.focusZoneRef,
+                focusZoneDirection: 'vertical',
+                defaultTabbableElement: menuList.focusZoneRef,
+                enableFocusRing: false,
+                isCircularNavigation: Platform.OS !== 'macos',
+              })}
```
Saadnajmi, line 98: At this point, could we just move this to wherever we calculate slot props?

Saadnajmi (approved):

packages/components/Menu/src/MenuList/MenuList.tsx
Saadnajmi: oh right this is the framework where the slot props are basically styling nvm

packages/components/Menu/src/MenuList/MenuList.tsx
rurikoaraki: I unfortunately can't test this so I can't confirm the behavior of the Menu on macOS. So I would say out of scope, if that's ok

apps/E2E/src/Menu/specs/Menu.spec.win.ts
```diff
@@ -144,17 +144,12 @@ describe('Menu Functional Testing', () => {
     expect(await MenuPageObject.didAssertPopup()).toBeFalsy(MenuPageObject.ERRORMESSAGE_ASSERT);
   });
 
-  it('Press "Tab" to navigate between MenuItems. Validate that focus switches correctly between MenuItems.', async () => {
+  it('Press "Tab" to navigate between MenuItems. Validate that focus does not switch between MenuItems.', async () => {
```
Saadnajmi: heehee

rurikoaraki:
Think I'm going to abandon this for now

PR #3172, opened 2023-10-23
[Menu win32] Add focus zone around menu groups

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In win32, menu groups should be keyboard arrow-able while you use tab to move between groups. Wrapping a focus zone around the Menu Group should align the keyboard behavior to be closer to the ideal spec.

### Verification

Added more items to the menu group test to test keyboarding behavior

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuGroup/MenuGroup.tsx
+++ b/packages/components/Menu/src/MenuGroup/MenuGroup.tsx
@@ -10,10 +10,8 @@ import type { UseSlots } from '@fluentui-react-native/framework';
 import type { MenuGroupProps, MenuGroupType } from './MenuGroup.types';
 import { menuGroupName } from './MenuGroup.types';
 
-// macOS has focus zone around entire list, including it for groups may
-// cause bad keyboard behavior. Will review in the future.
+// Intentionally not enabled on macOS to match system context menus
 const hasFocusZone = ['win32'].includes(Platform.OS as string);
-const useCircularNav = Platform.OS !== 'macos';
 
 export const MenuGroup = compose<MenuGroupType>({
   displayName: menuGroupName,
@@ -54,7 +52,7 @@ export const MenuGroup = compose<MenuGroupType>({
             {...(hasFocusZone && {
               focusZoneDirection: 'vertical',
               enableFocusRing: false,
-              isCircularNavigation: useCircularNav,
+              isCircularNavigation: true,
             })}
           >
             {childrenWithSet}
--- a/packages/components/Menu/src/MenuGroup/MenuGroup.tsx
+++ b/packages/components/Menu/src/MenuGroup/MenuGroup.tsx
@@ -1,18 +1,25 @@
 /** @jsxRuntime classic */
 /** @jsx withSlots */
 import React from 'react';
-import { View } from 'react-native';
+import { Platform, View } from 'react-native';
 
+import { FocusZone } from '@fluentui-react-native/focus-zone';
 import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
 import type { UseSlots } from '@fluentui-react-native/framework';
 
 import type { MenuGroupProps, MenuGroupType } from './MenuGroup.types';
 import { menuGroupName } from './MenuGroup.types';
 
+// macOS has focus zone around entire list, including it for groups may
+// cause bad keyboard behavior. Will review in the future.
+const hasFocusZone = ['win32'].includes(Platform.OS as string);
+const useCircularNav = Platform.OS !== 'macos';
+
 export const MenuGroup = compose<MenuGroupType>({
   displayName: menuGroupName,
   slots: {
     root: View,
+    focusZone: hasFocusZone ? FocusZone : React.Fragment,
   },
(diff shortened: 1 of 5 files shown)


packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -1,18 +1,25 @@
...
+import { FocusZone } from '@fluentui-react-native/focus-zone';
 import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
 import type { UseSlots } from '@fluentui-react-native/framework';
 
 import type { MenuGroupProps, MenuGroupType } from './MenuGroup.types';
 import { menuGroupName } from './MenuGroup.types';
 
+// macOS has focus zone around entire list, including it for groups may
+// cause bad keyboard behavior. Will review in the future.
+const hasFocusZone = ['win32'].includes(Platform.OS as string);
```
Saadnajmi, line 15: Nice, silence typescript check lol

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -1,18 +1,25 @@
...
 
+import { FocusZone } from '@fluentui-react-native/focus-zone';
 import { compose, mergeProps, withSlots } from '@fluentui-react-native/framework';
 import type { UseSlots } from '@fluentui-react-native/framework';
 
 import type { MenuGroupProps, MenuGroupType } from './MenuGroup.types';
 import { menuGroupName } from './MenuGroup.types';
 
+// macOS has focus zone around entire list, including it for groups may
+// cause bad keyboard behavior. Will review in the future.
```
Saadnajmi, line 14: Shhhhouuullld we file a Github issue and ref here so it's tracked somewhat?

Saadnajmi:
I'll double check change on macOS before approving

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
rurikoaraki: Ah I think I had this as "[macos, win32].includes" before and then I removed the "macOS". but this'll make it easy to add back if it ever needs to be

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
rurikoaraki: I should probably update this comment but from what you showed me it seems like not having a FZ in macOS would be the desired behavior? So maybe not necessary to review this later? What do you think?

Saadnajmi: Yeah I think we can re-word as "Intentionally not enabled on macOS to match system context menus"

Saadnajmi (approved):

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -39,7 +44,21 @@ export const MenuGroup = compose<MenuGroupType>({
...
         return child;
       });
-      return <Slots.root {...mergedProps}>{childrenWithSet}</Slots.root>;
+
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.focusZone
+            // avoid error that fires when props are passed into React.fragment
+            {...(hasFocusZone && {
+              focusZoneDirection: 'vertical',
```
samuelfreiberg, line 53: Just confirming - you don't want the left/right arrows to do anything?

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -39,7 +44,21 @@ export const MenuGroup = compose<MenuGroupType>({
...
+              enableFocusRing: false,
+              isCircularNavigation: true,
```
ksiler, line 55: Does this make it so when you reach the end of the menu group, down arrow takes you to the first menu item in the current menu group? If so, this doesn't align with the native behavior on win32. Up/down arrows should allow you to move to every item in the menu, even between groups, but maybe I'm misunderstanding what this does and it will allow you to do that.

rurikoaraki: oh ok. I misunderstood the spec then. Yeah this will make it so that hitting the end of the group will bring you back to the beginning. I thought that's what the bug was? We'll take this offline

rurikoaraki:
Wrong implementation

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
PPatBoyd, line 55: @samuelfreiberg check me if I'm out of date but I believe `isCircularNavigation` can be replaced with `NavigateAtEnd='NavigateWrap'`

samuelfreiberg: We expose the prop as isCircularNavigation, so this is correct. But under the hood, you are correct.

PR #3173, opened 2023-10-24
Fix Legacy SVG Icon `viewBox` prop being explicitly set as undefined

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

For rendering SVGs, the legacy Icon is mistakenly written to set the `viewBox` prop as undefined if no viewbox is passed. This is problematic because an un-set viewbox causes issues scaling across DPIs. ButtonV1 also consumes the legacy Icon which makes this even more of an issue.

This PR fixes this by not setting the prop if nothing is passed.

### Verification

Local testing and verification.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Icon/src/legacy/Icon.tsx
+++ b/packages/components/Icon/src/legacy/Icon.tsx
@@ -5,7 +5,7 @@ import type { ImageStyle, TextStyle } from 'react-native';
 import { mergeStyles, useFluentTheme } from '@fluentui-react-native/framework';
 import { stagedComponent, mergeProps, getMemoCache } from '@fluentui-react-native/framework';
 import { Text } from '@fluentui-react-native/text';
-import type { SvgProps, UriProps } from 'react-native-svg';
+import type { SvgProps } from 'react-native-svg';
 import { SvgUri } from 'react-native-svg';
 
 import type { IconProps, SvgIconProps, FontIconProps } from './Icon.types';
@@ -71,7 +71,7 @@ function renderSvg(iconProps: IconProps) {
   const { accessible, accessibilityLabel, width, height, color } = iconProps;
   const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width, height }, [width, height])[0]);
 
-  let svgProps: SvgProps | UriProps = { width, height, color };
+  const svgProps: SvgProps = { width, height, color };
   if (svgIconProps.viewBox) {
     svgProps.viewBox = svgIconProps.viewBox;
   }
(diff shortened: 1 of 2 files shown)


rurikoaraki:
> @rurikoaraki should this PR also include a change to move the button to IconV1?

No. That will 1. make this change more complicated and increase the scope of it unnecessarily. Changes like that should be made separately especially in case we need to revert this change due to a bug surfacing. 2. It's close to the fork, we should avoid larger changes like that at this time

packages/components/Icon/src/legacy/Icon.tsx
```diff
@@ -68,19 +68,23 @@ function renderFontIcon(iconProps: IconProps) {
...
       <View style={style} accessible={accessible} accessibilityRole="image" accessibilityLabel={accessibilityLabel}>
-        <svgIconProps.src viewBox={viewBox} width={width} height={height} color={color} />
+        <svgIconProps.src {...svgProps} />
       </View>
     );
   } else if (svgIconProps.uri) {
     return (
       <View style={style} accessible={accessible} accessibilityRole="image" accessibilityLabel={accessibilityLabel}>
-        <SvgUri uri={svgIconProps.uri} viewBox={viewBox} width={width} height={height} color={color} />
+        <SvgUri uri={svgIconProps.uri} {...svgProps} />
```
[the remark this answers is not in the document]
lawrencewin: done

packages/components/Icon/src/legacy/Icon.tsx
```diff
@@ -68,19 +69,24 @@ function renderFontIcon(iconProps: IconProps) {
 function renderSvg(iconProps: IconProps) {
   const svgIconProps: SvgIconProps = iconProps.svgSource;
   const { accessible, accessibilityLabel, width, height, color } = iconProps;
-  const viewBox = iconProps.svgSource.viewBox;
   const style = mergeStyles(iconProps.style, rasterImageStyleCache({ width, height }, [width, height])[0]);
 
+  let svgProps: SvgProps | UriProps = { width, height, color };
```
Saadnajmi, line 74: Sorry, I didn't realize that `SvgProps` and UriProps` are actually two different types. I think what was earlier is probably better 😅

lawrencewin: No worries, just reverted it!

Saadnajmi (approved):

PR #3179, opened 2023-10-24
Update iOS and Android Button spacing fonts and values

gremlin529:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [x] windows
- [ ] android

### Description of changes

Bringing the button controls in line with the latest figmas

### Verification

running the test suite and it exhibits only the failures that are present in main.  Also looked at the test harness apps

| 
image
 | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Please run "yarn change" and push the result

gremlin529:
> Please run "yarn change" and push the result

Done I believe

packages/components/Button/src/ButtonTokens.ios.ts
```diff
@@ -10,75 +10,72 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
...
-          spacingIconContentAfter: globalTokens.size100,
+          spacingIconContentAfter: globalTokens.size80,
         },
         hasIconBefore: {
-          spacingIconContentBefore: globalTokens.size100,
+          spacingIconContentBefore: globalTokens.size80,
         },
         focused: {
-          paddingHorizontal: globalTokens.size160,
+          paddingHorizontal: globalTokens.size120,
```
rurikoaraki, line 30: Seems like the focused override for paddingHorizontal can be removed?

packages/components/Button/src/ButtonTokens.ios.ts
```diff
@@ -10,75 +10,72 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
...
         },
       },
     },
     small: {
-      padding: globalTokens.size80 - globalTokens.stroke.width10,
+      paddingHorizontal: globalTokens.size60,
       borderWidth: globalTokens.stroke.width10,
```
rurikoaraki, line 36: Is the width of the button what you expect to see if paddingHorizontal and borderWidth are set? borderWidth might be adding an additional pixel that you're not expecting.

packages/components/Button/src/ButtonTokens.ios.ts
gremlin529: I wasn't sure you could have an empty or removed "Focused" section and it would behave properly so I thought the safest course was to leave it and it would have no effect.

rurikoaraki: You can remove them safely. Each of the subgroups of tokens are layers which are applied on top of any base or previous layer's values. So if there's a value already defined, it's not required to redefine it to the same value at a later layer. It will have the value that was set by any previous layer, so you only need to define explicit overides when certain layers are applied. The ordering of the layers are usually defined in a <component>.styling.ts file

rurikoaraki:
CI is failing, but otherwise change looks good.

packages/components/Button/src/ButtonTokens.ios.ts
Saadnajmi: I should also note that you'll basically never get the `focused` state on iOS, given that React Native mobile doesn't really support keyboard focus. So defining tokens here is not really necessary

packages/components/Button/src/ButtonTokens.ios.ts
```diff
@@ -10,75 +10,72 @@ export const defaultButtonTokens: TokenSettings<ButtonTokens, Theme> = () =>
...
+          spacingIconContentBefore: globalTokens.size80,
         },
         focused: {
           paddingHorizontal: globalTokens.size160,
         },
       },
     },
-    rounded: {
-      borderRadius: globalTokens.corner.radius40,
-    },
```
Saadnajmi, line 81: Do we just not have a rounded button on iOS?

packages/components/Button/src/ButtonTokens.ios.ts
gremlin529: Yeah i've prepped a review to remove this but having trouble running the tests.

packages/components/Button/src/ButtonTokens.ios.ts
gremlin529: Rounded button is the default, the others are variants, so we still have rounded, this is per the docs.

packages/components/Button/src/ButtonTokens.ios.ts
gremlin529: and pushed

packages/components/Button/src/ButtonTokens.ios.ts
gremlin529: the size looks good in the inspector

rurikoaraki (approved):

rurikoaraki (approved):

rurikoaraki (approved):

PR #3189, opened 2023-10-27
[win32] Fix keyboarding in MenuList

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Third time's the charm?

This change tries to fix keyboarding for menus in win32. The spec as I understand it is the following:

- Items in a list should be able to be fully arrow navigable, and focus travels across menu groups
- If you press tab while in a list with groups, you should stop at the start of the next group
- Things that are outside of a homogenous list should only be able to be tabbed to, not arrowable.

To follow this, we do the following:

- MenuList now has a focus zone in win32. It allows for tab navigation to flow through the FZ and will not trap it, allowing things in a menu but outside a menu list to also be tabbable. This preserves arrow navigation to just menu list items
- MenuGroups also have a focus zone in win32. This focus zone only allows arrowing and lets focus fall through to the next focusable element outside the FZ. This allows for Menu groups to be tab stops without interfering with arrowing.

This gets us closer to the win32 keyboarding behavior (though not perfectly aligned, since the win32 spec is fairly intricate).

### Verification

Added two test cases to the tester and verified that keyboarding behaves as intended.

### DIFF
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuFocusZone.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuFocusZone.tsx
@@ -1,0 +1,26 @@
+import * as React from 'react';
+
+import { ButtonV1 as Button } from '@fluentui/react-native';
+import { Menu, MenuItem, MenuTrigger, MenuPopover, MenuList } from '@fluentui-react-native/menu';
+import { Stack } from '@fluentui-react-native/stack';
+
+import { stackStyle } from '../Common/styles';
+
+export const MenuFocusZone: React.FunctionComponent = () => {
+  return (
+    <Stack style={stackStyle}>
+      <Menu>
+        <MenuTrigger>
+          <Button>The Menu</Button>
+        </MenuTrigger>
+        <MenuPopover>
+          <MenuList>
+            <MenuItem>A plain MenuItem</MenuItem>
+            <MenuItem>A plain MenuItem 2</MenuItem>
+          </MenuList>
+          <Button>Not in arrow loop</Button>
+        </MenuPopover>
+      </Menu>
+    </Stack>
+  );
+};
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
@@ -20,6 +20,7 @@ import { Switch } from '@fluentui-react-native/switch';
 import { TextV1 as Text } from '@fluentui-react-native/text';
 
 import { E2EMenuTest } from './E2EMenuTest';
+import { MenuFocusZone } from './MenuFocusZone';
 import { MenuIcons } from './MenuIcons';
 import { MenuTriggerChildRef } from './MenuRefs';
 import { MenuScrollView } from './MenuScrollView';
@@ -527,6 +528,13 @@ const menuSections: TestSection[] = [
     name: 'Menu with groups',
     component: MenuWithGroups,
   },
+  Platform.select({
+    android: null,
+    default: {
+      name: 'MenuList FocusZone test',
+      component: MenuFocusZone,
+    },
+  }),
 ];
 
 const e2eSections: TestSection[] = [
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
@@ -377,9 +377,11 @@ const MenuWithGroups: React.FunctionComponent = () => {
             <MenuGroup>
               <MenuGroupHeader>Section 1</MenuGroupHeader>
               <MenuItem>A plain MenuItem</MenuItem>
+              <MenuItem>A plain MenuItem</MenuItem>
             </MenuGroup>
             <MenuGroup>
               <MenuGroupHeader>Section 2</MenuGroupHeader>
+              <MenuItem>A plain MenuItem</MenuItem>
               <MenuItem>A plain MenuItem</MenuItem>
             </MenuGroup>
           </MenuList>
--- a/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
+++ b/apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
@@ -20,7 +20,7 @@ import { Switch } from '@fluentui-react-native/switch';
 import { TextV1 as Text } from '@fluentui-react-native/text';
 
 import { E2EMenuTest } from './E2EMenuTest';
-import { MenuFocusZone } from './MenuFocusZone';
+import { MenuComponentOutsideMenuList } from './MenuComponentOutsideMenuList';
 import { MenuIcons } from './MenuIcons';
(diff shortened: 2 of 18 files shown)


ksiler:
When testing these changes, it doesn't seem like menu groups are tab stops. So, if I'm in the first group, tabbing does not take me to the first item in the next group like I would expect. Instead, it just takes me to the first item in the list. I'm not sure what's causing this or how to fix it.

rurikoaraki:
> When testing these changes, it doesn't seem like menu groups are tab stops. So, if I'm in the first group, tabbing does not take me to the first item in the next group like I would expect. Instead, it just takes me to the first item in the list. I'm not sure what's causing this or how to fix it.

You might need to update the client to be running latest win32 code. The checked in version is outdated.

rurikoaraki:
@samuelfreiberg @Saadnajmi had some concerns about the nested FZ, is the behavior still considered risky/undefined or was there work to make nested FZs behave in a reasonable way?

ksiler (approved):

PPatBoyd:
> @samuelfreiberg @Saadnajmi had some concerns about the nested FZ, is the behavior still considered risky/undefined or was there work to make nested FZs behave in a reasonable way?

@Saadnajmi what were the concerns? :) Hopefully we can clarify and document for the future

samuelfreiberg:
> > @samuelfreiberg @Saadnajmi had some concerns about the nested FZ, is the behavior still considered risky/undefined or was there work to make nested FZs behave in a reasonable way?
> 
> @Saadnajmi what were the concerns? :) Hopefully we can clarify and document for the future

Nested FZ behavior is undefined - we never did any work to integrate. Behavior is unknown to me. When I made the FZ spec, I talked with Web and they had so many bugs/issues with nested FZs - it turned out to be a huge burden. Since then, we never planned to look into nested FZs (this was a couple years ago).

It could work, it could not work. No one has ever investigated behavior, so it's risky

Saadnajmi:
@ppayboyd the last I spoke with @samuelfreiberg about it, Nested FocusZones was explicitly undefined behavior on win32 because it was too complex to support at the time. My concern is whatever the behavior is right now is not stable. It also seems to be opposite of how we handle it on macOS.

apps/fluent-tester/src/TestComponents/Menu/MenuFocusZone.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+
+import { ButtonV1 as Button } from '@fluentui/react-native';
+import { Menu, MenuItem, MenuTrigger, MenuPopover, MenuList } from '@fluentui-react-native/menu';
+import { Stack } from '@fluentui-react-native/stack';
+
+import { stackStyle } from '../Common/styles';
+
+export const MenuFocusZone: React.FunctionComponent = () => {
```
Saadnajmi, line 9: Maybe a different name like "ComponentOutsideMenuList"? The name made me think I'd see a FZ inside the Menu somewhere, so it wasn't obvious what this was testing

apps/fluent-tester/src/TestComponents/Menu/MenuTest.tsx
```diff
@@ -525,6 +528,13 @@ const menuSections: TestSection[] = [
     name: 'Menu with groups',
     component: MenuWithGroups,
   },
+  Platform.select({
+    android: null,
+    default: {
+      name: 'MenuList FocusZone test',
+      component: MenuFocusZone,
```
Saadnajmi, line 535: Similar comment about naming here

Saadnajmi:
@p

> > > @samuelfreiberg @Saadnajmi had some concerns about the nested FZ, is the behavior still considered risky/undefined or was there work to make nested FZs behave in a reasonable way?
> > 
> > 
> > @Saadnajmi what were the concerns? :) Hopefully we can clarify and document for the future
> 
> Nested FZ behavior is undefined - we never did any work to integrate. Behavior is unknown to me. When I made the FZ spec, I talked with Web and they had so many bugs/issues with nested FZs - it turned out to be a huge burden. Since then, we never planned to look into nested FZs (this was a couple years ago).
> 
> It could work, it could not work. No one has ever investigated behavior, so it's risky

To this point, I want to highlight the behavior difference. When you press Tab inside a nested FocusZone:
- macOS FocusZone will take you out of the _outermost_ FocusZone. This was explicit behavior I added as a partner request
- Win32 (as it would appear) takes you outside the current focuszone, and to the next item. It does not consider whether there was a parent FZ or not.

I could see an argument for either behavior, I think the latter (win32 version) might be more correct. But further cements we're in undefined territory.

packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -39,7 +44,21 @@ export const MenuGroup = compose<MenuGroupType>({
         }
         return child;
       });
-      return <Slots.root {...mergedProps}>{childrenWithSet}</Slots.root>;
+
+      return (
+        <Slots.root {...mergedProps}>
+          <Slots.focusZone
```
Saadnajmi, line 50: I know the slot is a Fragment on non win32 platforms, but it still feels a bit weird to see the name here. I don't know I have a better idea, so I guess not blocking

packages/components/Menu/src/MenuList/useMenuList.ts
```diff
@@ -162,9 +162,21 @@ export const useMenuList = (_props: MenuListProps): MenuListState => {
...
   });
 
+  // focus management
+  const focusZoneRef = React.useRef<View>();
+  const setFocusZoneFocus = () => {
+    focusZoneRef?.current?.focus();
+  };
+
+  React.useEffect(() => {
+    setFocusZoneFocus();
```
Saadnajmi, line 172: Why not just call `focusZoneRef?.current?.focus();` directly here?

Saadnajmi:
@ppatboyd / @samuelfreiberg I talked to @rurikoaraki offline, and I think actually the behavior difference of nested FZs between win32 and macOS could be handled by supporting the `tabNavigation` prop that was recently added on macOS. I'll table that discussion for later then.

packages/components/Menu/src/MenuList/useMenuList.ts
rurikoaraki: I think it was a bit of code sharing done with the onMouseLeave call below. This code was moved from MenuList.tsx to consolidate the hooks into one place. I think it's fine as is, since we want the same thing to happen on mount as onMouseLeave, it seems.

packages/components/Menu/src/MenuList/useMenuList.ts
```diff
@@ -162,9 +162,21 @@ export const useMenuList = (_props: MenuListProps): MenuListState => {
...
+  }, []);
+
   return {
     props: {
       ...context,
+      onMouseLeave: setFocusZoneFocus,
```
rurikoaraki, line 178: note to self probably don't want this on win32

Saadnajmi (approved):

PR #3193, opened 2023-10-28
Mark tests to satisfy compliance

dannyvv:
### Description of changes

Compliance requires we run tests, we don't use build-in ADO test runners so we need to annotate which tasks run tests.

### Verification
N/A

### DIFF
--- a/.ado/azure-pipelines.publish.yml
+++ b/.ado/azure-pipelines.publish.yml
@@ -81,7 +81,7 @@ extends:
                 displayName: 'yarn install'
 
               - script: |
-                  yarn buildci
+                  yarn buildci [test]
                 displayName: 'yarn buildci'
 
               - script: |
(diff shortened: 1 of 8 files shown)


.ado/azure-pipelines.publish.yml
```diff
@@ -81,7 +81,7 @@ extends:
                 displayName: 'yarn install'
 
               - script: |
-                  yarn buildci
+                  yarn buildci [test]
```
Saadnajmi, line 84: This doesn't get in the way of running the test? It's passing "[test]" as an argument now

Saadnajmi (approved):

.ado/azure-pipelines.publish.yml
dannyvv: Yes, my bad. fixed.

PR #3208, opened 2023-11-03
Expose Style Prop for FocusZone

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [X] win32 (Office)
- [X] windows
- [ ] android

### Description of changes

In Layout, FocusZone acts as a View. However, we don't expose Style props for FocusZone. This caused a bug where a ScrollView wrapped in a FocusZone couldn't dynamically grow because the default Flex values are 0 (FZ won't grow with space). In this PR, we expose the Style prop to FocusZone and pass the values directly to native.

### Verification

Manual testing

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


PPatBoyd (approved):

packages/components/FocusZone/src/FocusZone.types.ts
```diff
@@ -63,6 +64,11 @@ export type FocusZoneProps = React.PropsWithChildren<{
    * Callback called when “focus” event triggered in FocusZone
    */
   onFocus?: (e?: any) => void;
+
+  /**
+   * Allow consumers to pass in Style props
+   */
+  style?: ViewStyle;
```
Saadnajmi, line 71: Could we just extend ViewProps too?

Saadnajmi: At least on macOS, I think FocusZone accepts all or most View props. I never got around to changing the types though

samuelfreiberg: FocusZone doesn't actually have a View slot, so I don't think it's viable to also extend ViewProps. We just want to give the consumer access to some layout properties.

Saadnajmi: What I meant was, macOS FocusZone extends RCTView and RCTViewManager, so the native component itself accepts most view props. I think when I asked (and we looked) at the win32 FZ's implementation, that was also true

samuelfreiberg: There's an added complexity to some of the props exposed by ViewProps as it relates to FocusZone - collapsable, focusable are some.

For now, adding Style solves the problems of some of our partners. If someone brings this up in the future, definitely open for a discussion :)

samuelfreiberg: Thanks for the callout though - it's a valid thought.

PR #3216, opened 2023-11-07
Update to react-native 0.72

acoates-ms:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Mostly bumping package versions.
Modified a few types that were using number|string for dimensions.  In RN these types now use `${number}%` instead of string, so we need our types to be similarly restrictive.

Heavy rewrite of the stack components.  Switched it to use the newer gap properties.  Also substantially reduced the valid values of some of its properties.  Many of the value it was trying to use never worked in RN. (Setting padding to "5 10" doesn't work in RN), nor do the "calc" values.

### Verification

Build/test.  Built the win32 bundle and launched the test app.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Menu/src/MenuItemCheckbox/MenuItemCheckbox.styling.ts
```diff
@@ -43,7 +43,8 @@ export const stylingSettings: UseStylingOptions<MenuItemCheckboxProps, MenuItemC
             justifyContent: 'center',
           },
           ...(hasPressRententionForA11y && {
-            pressRetentionOffset: typeof tokens.padding === 'number' ? tokens.padding : parseFloat(tokens.padding), /// Retention of the press area outside of the checkbox equal to padding to match accessibility requirement
+            pressRetentionOffset:
+              typeof tokens.padding === 'number' ? tokens.padding : typeof tokens.padding === 'number' ? parseFloat(tokens.padding) : 0, /// Retention of the press area outside of the checkbox equal to padding to match accessibility requirement
```
ksiler, line 47: should the second typeof be `typeof tokens.padding === 'string'` ?

acoates-ms: Good catch

PPatBoyd (approved):

PR #3236, opened 2023-11-15
Re-structure TabList Animated Indicator code to work better with RTL layouts

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Currently, we style and animate the TabList's animated indicator by calculating, setting, and adjusting its `start` layout position. On RTL layouts, the `start` value is calculated by flipping the left offset of a given tab by subtracting it from the width of the tablist. On mac, this also means that the translate transform for JS animations needs to be flipped. 

While this works, the extra logic to set the `start` layout style results in code that is more complex than it needs to be. This PR changes the approach for positioning the indicator. Instead of calculating the `start` of the indicator, we set its position to be absolute and calculate its `left` / `top` offsets relative to the entire tablist. This removes the need for extra logic on RTL layouts. The code is also re-organized to include all of the layout math in `useTabAnimation` and remove it from the win32's `useAnimatedIndicatorStyles` (unfortunately, math still has to be done on math because we're calculating transforms 😞). 

Other changes:
- Miscellaneous styling fixes 
- Refactor different animated indicator related types
- Fix TabList test page title to be "TabList" instead of "TabsV1"

### Verification

| image | image |

Mac video: 

https://github.com/microsoft/fluentui-react-native/assets/15683103/b61ee923-50e4-45eb-b854-786e13099355

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/TabList/src/Tab/useTabAnimation.ts
```diff
@@ -58,24 +58,30 @@ export function useTabAnimation(
...
         } else {
-          width = e.nativeEvent.layout.width - totalIndicatorInset;
-          height = tokens.indicatorThickness;
+          indicatorWidth = tabWidth - totalIndicatorInset;
+          indicatorHeight = tokens.indicatorThickness;
+        }
+        let tabX = e.nativeEvent.layout.x;
+        // For RTL users, we adjust the x position of each tab to be relative to the entire tablist starting right to left
+        if (Platform.OS == ('win32' as any) && I18nManager.isRTL) {
+          tabX = layout.tablist.width - (tabX + tabWidth);
```
rurikoaraki, line 78: Kind of surprised you have to do this considering that x eventually gets set to the start style property, which should take RTL into account. am I misreading the code?

lawrencewin: The intuitive fix would initially be to just use `left` instead of `start` for the given x, but on both mac and win32, setting `left: x` when RTL is enabled, moves the transform origin towards the right side of the tablist rather than the left. This makes the indicator fly way off screen.

rurikoaraki (approved):
:shipit:

rurikoaraki (approved):

packages/experimental/TabList/src/Tab/useTabAnimation.ts
```diff
@@ -49,40 +89,37 @@ export function useTabAnimation(
...
+  React.useEffect(() => {
+    if (
+      (tabLayoutRect &&
+        // Following checks are for win32 only, will be removed after addressing scrollview layout bug
+        Platform.OS !== ('win32' as any)) ||
+      (layout?.tablist &&
+        layout.tablist.width > 0 &&
+        tabLayoutRect.height <= layout.tablist.height &&
+        tabLayoutRect.height < RENDERING_HEIGHT_LIMIT)
+    ) {
```
rurikoaraki: nit: refactor this check into a function?

packages/experimental/TabList/src/Tab/useTabAnimation.ts
```diff
@@ -49,40 +89,37 @@ export function useTabAnimation(
...
+      calculateAndUpdateAnimationLayoutInfo(tabLayoutRect, tokens);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rurikoaraki: >// eslint-disable-next-line react-hooks/exhaustive-deps [](http://example.com/codeflow?start=4&length=55)

comment explaining warning suppression please

rurikoaraki (approved):

packages/experimental/TabList/src/TabList/TabList.tsx
```diff
@@ -43,22 +44,24 @@ export const TabList = compose<TabListType>({
...
-            )}
-          </Slots.container>
+          <Slots.root {...mergedProps}>
+            <Slots.container
+              disabled={disabled || tablistDisabledState}
+              defaultTabbableElement={defaultTabbableElement}
+              focusZoneDirection={vertical ? 'vertical' : 'horizontal'}
+              isCircularNavigation={isCircularNavigation}
+            >
+              <Slots.stack>{children}</Slots.stack>
```
PPatBoyd, line 54: Slightly questioning if we need the `Slots.stack` View in the UI tree but can be taken as a followup

PPatBoyd (approved):

PR #3237, opened 2023-11-16
Publish Initial Tooltip Package

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Publishing the initial Tooltip experimental package with basic setup for testing now that there is a working baseline native module to call. Will be iterating on this with additional testing and documentation.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Please consider breaking up changes in the future into smaller PRs. This is a lot of code to review at once.

rurikoaraki:
> 

Are E2E tests a follow up?

---
Refers to: apps/E2E/src/Tooltip/consts.ts:3 in e8de9c8. [](commit_id = e8de9c80f1a2ebd8a12ad350540764f17ae75e75, deletion_comment = False)

apps/fluent-tester/src/TestComponents/Tooltip/consts.ts
```diff
@@ -0,0 +1,2 @@
+export const HOMEPAGE_TOOLTIP_BUTTON = 'Homepage_Tooltip_Button';
```
rurikoaraki, line 1: Do we need this and the consts file in the E2E folder? They contain the same code

apps/fluent-tester/src/TestComponents/Tooltip/TooltipTest.tsx
```diff
@@ -0,0 +1,29 @@
...
+    uwpStatus: 'Backlog',
+    iosStatus: 'Backlog',
+    macosStatus: 'Backlog',
+    androidStatus: 'Backlog',
+  };
+
+  const description =
+    'A Tooltip component displays additional information about another component. Tooltip is not expected to handle interactive content.';
+
+  return <Test name="Tooltip Test" description={description} sections={tooltipSections} status={status}></Test>;
```
rurikoaraki, line 28: >}></Test>; [](http://example.com/codeflow?start=102&length=10)

Should be a self-closing tag if the component has no children

apps/fluent-tester/src/TestComponents/Tooltip/TooltipDefault.tsx
```diff
@@ -0,0 +1,24 @@
...
+      </Tooltip>
+      <Tooltip content="Test 2">
+        <Button>Test 2</Button>
+      </Tooltip>
+      <Tooltip content="Test 3">
+        <Button>Test 3</Button>
+      </Tooltip>
+      <Tooltip content="Test 4">
+        <Button>Test 4</Button>
+      </Tooltip>
```
rurikoaraki, line 21: Could you explain what Test 3 and Test 4 are testing for that's different from Test 2?

apps/fluent-tester/src/TestComponents/Tooltip/TooltipDefault.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { Button } from '@fluentui-react-native/experimental-button';
+import { Tooltip } from '@fluentui-react-native/tooltip';
+
+export const TooltipDefault: React.FunctionComponent = () => {
+  return (
+    <View>
+      <Tooltip content="Test" positioning="topCenter">
+        <Button>Test</Button>
+      </Tooltip>
```
rurikoaraki, line 12: suggestion: Positioning could be its own test section

packages/experimental/Tooltip/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui-react-native.git",
+    "directory": "packages/experimental/Tooltip"
+  },
+  "dependencies": {
+    "@fluentui-react-native/framework": "0.12.0",
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@fluentui-react-native/component-cache": "^1.6.0",
```
rurikoaraki, line 33: Should be a dependency, since it's directly referenced in prod code.

packages/experimental/Tooltip/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@fluentui-react-native/eslint-config-rules": "^0.1.1",
+    "@fluentui-react-native/experimental-button": "^0.19.0",
```
rurikoaraki, line 35: experimental-button is a placeholder package to not break users, we should use button instead since that's where the really code lives.

packages/experimental/Tooltip/src/Tooltip.types.ts
```diff
@@ -0,0 +1,63 @@
+import type * as React from 'react';
+
+export const tooltip = 'Tooltip';
```
rurikoaraki, line 3: >tooltip [](http://example.com/codeflow?start=13&length=7)

we should add "Name" to this variable to make it clear it's the name of the component.

packages/experimental/Tooltip/src/useTooltip.ts
```diff
@@ -0,0 +1,10 @@
+import type { TooltipProps } from './Tooltip.types';
+
+export const useTooltip = (props: TooltipProps): TooltipProps => {
+  const { ...rest } = props;
```
rurikoaraki, line 4: Seems like this isn't used, do we need this file?

packages/experimental/Tooltip/src/Tooltip.types.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Properties and Tokens for FluentUI React Native Tooltip
+ */
+export type DirectionalHint =
```
rurikoaraki, line 8: >DirectionalHint [](http://example.com/codeflow?start=12&length=15)

Is this shared with Callout or is this different?

packages/experimental/Tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param layer The name of the state that is being checked for
+ * @param userProps The props that were passed into the tooltip
+ * @returns Whether the styles that are assigned to the layer should be applied to the tooltip
+ */
+export const tooltipLookup = (layer: string, userProps: TooltipProps): boolean => {
+  return userProps[layer];
+};
+
+export const Tooltip = compose<TooltipType>({
```
rurikoaraki, line 26: >compose [](http://example.com/codeflow?start=23&length=7)

Seems like this is meant to be a wrapper around a native component. In that case, compose might be too much if styling isn't going to be driven by JS. If this component is meant to be a wrapper around a native component it might make more sense to use stagedComponent instead which is a lighter framework.

apps/fluent-tester/src/TestComponents/Tooltip/TooltipDefault.tsx
gulnazsayed: Updated!

gulnazsayed:
> > 
> 
> Are E2E tests a follow up?
> 
> Refers to: apps/E2E/src/Tooltip/consts.ts:3 in [e8de9c8](https://github.com/microsoft/fluentui-react-native/commit/e8de9c80f1a2ebd8a12ad350540764f17ae75e75). [](commit_id = [e8de9c8](https://github.com/microsoft/fluentui-react-native/commit/e8de9c80f1a2ebd8a12ad350540764f17ae75e75), deletion_comment = False)

Yes, they will be a follow-up. Including it in another testing PR for easier review once this one is checked in.

apps/fluent-tester/src/TestComponents/Tooltip/TooltipDefault.tsx
gulnazsayed: I had included these tests for manually testing different variations and testing how it works when quickly hovering over several components. Removed the duplicate cases and intending to have more tests in follow-up PR so that it is easier to review.

packages/experimental/Tooltip/src/Tooltip.tsx
gulnazsayed: Thanks for the suggestion! Updated to using stagedComponent.

packages/experimental/Tooltip/src/useTooltip.ts
gulnazsayed: Removed!

packages/experimental/Tooltip/src/Tooltip.types.ts
gulnazsayed: Shared with Callout so changed to import from there

apps/fluent-tester/src/TestComponents/Tooltip/TooltipTest.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const tooltipSections: TestSection[] = [
+  {
+    name: 'Default',
+    testID: TOOLTIP_TESTPAGE,
+    component: TooltipDefault,
+  },
+  {
+    name: 'Target + Positioning',
+    testID: TOOLTIP_TESTPAGE,
```
rurikoaraki, line 17: >testID: TOOLTIP_TESTPAGE, [](http://example.com/codeflow?start=4&length=25)

Only the first test needs an ID. This is so that the e2e test is able to fine the element via the UIA tree.

rurikoaraki (approved):

PR #3238, opened 2023-11-16
Button tokens was missing iOS File and had a typo in Default file

gremlin529:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

### Verification

ran local tests and test code

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/Button/src/ButtonColorTokens.ios.ts
```diff
@@ -0,0 +1,96 @@
+import type { Theme } from '@fluentui-react-native/framework';
+import type { TokenSettings } from '@fluentui-react-native/use-styling';
+
+import type { ButtonTokens } from './Button.types';
+
+export const defaultButtonColorTokens: TokenSettings<ButtonTokens, Theme> = (t: Theme) =>
+  ({
+    backgroundColor: t.colors.buttonBackground,
```
rurikoaraki, line 8: Is this a copy of the base ButtinColorTokens file? If so this file isn't necessary, it should pick up the base version if no override has been specified.

rurikoaraki:
Please add a description of your changes.

packages/components/Button/src/ButtonColorTokens.ios.ts
gremlin529: let me check that, I actually noticed this was missing and the bugs in the other one several weeks ago then forgot to go through with the check in so I'm resurecting it.

rurikoaraki: If you do need iOS tokens, we should use Fluent tokens to specify the styling. The current colors listed here are older colors that aren't used in the Fluent2 system.

PR #3281, opened 2023-12-04
Update packages to version in NPM (again)

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

The NPM publish pipeline successfully published new versions of these packages but did not update them in the repo due to an auth issue. Now that the auth issue has been resolved, we need to bump these packages in the repo so they match the version that exists in NPM and unblocks the NPM publish pipeline.

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-dependency-profiles-9b7c2763-ecdc-43b6-b097-c5ec8bda8d9c.json
+++ b/change/@fluentui-react-native-dependency-profiles-9b7c2763-ecdc-43b6-b097-c5ec8bda8d9c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "update versions to match NPM (again)",
+  "packageName": "@fluentui-react-native/dependency-profiles",
+  "email": "krsiler@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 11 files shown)


change/@fluentui-react-native-dependency-profiles-9b7c2763-ecdc-43b6-b097-c5ec8bda8d9c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
rurikoaraki, line 2: These files would cause another bump right? Is that intended?

ksiler: That's right - I'm not sure that we need another bump but it's what we've done in the past. Otherwise, the PR build will complain there are no change files present.

ksiler: I guess I could mark it as "no change" instead of "patch" so it doesn't actually bump?

rurikoaraki (approved):

PR #3301, opened 2023-12-12
Legacy RadioGroup Fix - Narrator announcing selected RadioButton as "non-selected"

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This is an accessibility bug fix for legacy RadioGroup where narrator was reading selected RadioButtons as "non-selected." The fix for keyboard focus is similar to the workaround for RadioGroup V1. However, there appears to be an additional timing issue for onPress now where narrator reads "non-selected" for selected RadioButtons where you mouse click on a RadioButton. This issue was not present in RadioGroup V1, even though the events are the same between both legacy and V1, which leads me to think it's a timing issue in legacy RadioGroup. Hence, the fix for onPress is similar to keyboarding, essentially calling changeSelection and setting that it was invoked to set the focus on the current RadioButton.

### Verification

No visual changes. Tested manually using Narrator.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki:
Curious, is there anything stopping the client from using the new Radio group instead?

gulnazsayed:
> Curious, is there anything stopping the client from using the new Radio group instead? In general we shouldn't be making fixes to the legacy controls, especially if it requires a lot of additional code.

An accessibility bug was filed for this issue that was pushing us out of compliance and since we haven't deprecated the old RadioGroup yet we thought it would be best to fix the issue in the meantime!

packages/components/RadioGroup/src/legacy/RadioButton.win32.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+enum DirectionalArrowKeys {
+  ArrowUp = 'ArrowUp',
+  ArrowDown = 'ArrowDown',
+  ArrowRight = 'ArrowRight',
+  ArrowLeft = 'ArrowLeft',
+}
+
+export const RadioButton = compose<IRadioButtonType>({
+  displayName: radioButtonName,
```
rurikoaraki, line 36: What of this file is a fork from the base file?

gulnazsayed: The win32 specific parts added on to the contents of the base file are primarily lines 77-139.

packages/components/RadioGroup/src/legacy/RadioButton.win32.tsx
```diff
@@ -0,0 +1,217 @@
...
+          let newCurrRadioButtonIndex;
+          if (next) {
+            newCurrRadioButtonIndex = (currRadioButtonIndex + 1) % length;
+            info.onChange && info.onChange(info.enabledButtonKeys[newCurrRadioButtonIndex]);
+            info.updateInvoked && info.updateInvoked(true);
+          } else if (previous) {
+            newCurrRadioButtonIndex = (currRadioButtonIndex - 1 + length) % length;
+            info.onChange && info.onChange(info.enabledButtonKeys[newCurrRadioButtonIndex]);
+            info.updateInvoked && info.updateInvoked(true);
+          }
```
rurikoaraki, line 112: Is there any case where both next and previous are false? If not, making this an if/else instead could be better so that we make sure that the code falls into one case or the other.

Also the last two lines of each block are identical, we could pull them out.

rurikoaraki (approved):

rurikoaraki (approved):

PR #3331, opened 2023-12-29
VoiceOver should ignore contextual menu item children

amgleitman:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

On macOS, VoiceOver reads "interactive" when looking at `ContextualMenuItem`s and `SubmenuItem`s, which is not on par with what a native `NSMenu` does. The reason for this appears to be that we were previously able to navigate to the FURN contextual menu item child views (such as the icon, the text, or the chevron indicating a submenu). `NSMenu` doesn't allow for this, so we can replicate this by setting `accessible: false` on all the subcomponents that make up a `ContextualMenuItem` or `SubmenuItem`.

### Verification

Validated that VoiceOver treats native `NSMenu`s and FURN contextual menus more similarly as per the specs above.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/packages/components/ContextualMenu/src/ContextualMenuItem.tsx
+++ b/packages/components/ContextualMenu/src/ContextualMenuItem.tsx
@@ -128,8 +128,8 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
         children: text,
       },
       icon: {
-        accessible: false,
         ...createIconProps(icon),
+        accessible: false,
       },
     });
 
--- a/packages/components/ContextualMenu/src/ContextualMenuItem.tsx
+++ b/packages/components/ContextualMenu/src/ContextualMenuItem.tsx
@@ -123,8 +123,14 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
         testID,
         ...rest,
       },
-      content: { children: text },
-      icon: createIconProps(icon),
+      content: {
+        accessible: false,
+        children: text,
+      },
+      icon: {
+        accessible: false,
+        ...createIconProps(icon),
+      },
     });
 
     return { slotProps, state };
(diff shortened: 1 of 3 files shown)


Saadnajmi:
Do we need to do this for MenuV1 too?

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -123,8 +123,14 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
         testID,
         ...rest,
       },
-      content: { children: text },
-      icon: createIconProps(icon),
+      content: {
+        accessible: false,
```
Saadnajmi, line 127: You might want to filter to only set this on macOS so as to not affect win32 and Android

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
```diff
@@ -123,8 +123,14 @@ export const ContextualMenuItem = compose<ContextualMenuItemType>({
...
+        children: text,
+      },
+      icon: {
+        accessible: false,
+        ...createIconProps(icon),
```
Saadnajmi, line 132: Nit: reverse the order so that `createIconProps` doesn't accidentally overwrite `accessible`

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
ksiler: I think these changes make sense for win32 and we're already doing it for Menu V1: https://github.com/microsoft/fluentui-react-native/blob/7943ac41f1e71920e3f959c73aab6ebe0f43fdbc/packages/components/Menu/src/MenuItem/MenuItem.tsx#L53C55-L53C55

Saadnajmi: Oh, perfect! I guess that means the bug doesn't repro on MenuV1 then for macOS?

Saadnajmi (approved):

packages/components/ContextualMenu/src/ContextualMenuItem.tsx
Saadnajmi: Talked offline, seems not.

amgleitman: Correct, MenuV1 doesn't give us "interactive" as intended.

PR #3333, opened 2024-01-02
Specify disabled=false explicitly in accessibility state for Contextual Menu items and subitems

amgleitman:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We've seen a bug where Voice Control doesn't label items that come out of a Menu Button V1.

Some basic experimentation shows that the bug ultimately lies in React Native macOS where specifying an accessibility state without explicitly saying we're not disabled causes the OS to not recognize an item as interactable. We can fix this by loosening [this check](https://github.com/microsoft/react-native-macos/blob/50e1a4e040649f05136814ea799b4cdc3f6e6b2c/packages/react-native/React/Views/RCTView.m#L514-L518) in React Native macOS, but this change limited to FURN is a valid workaround too.

### Verification

Validated that Voice Control will show numbers next to enabled items.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-contextual-menu-0e8e7823-58b6-4621-b98a-1241fe072b8a.json
+++ b/change/@fluentui-react-native-contextual-menu-0e8e7823-58b6-4621-b98a-1241fe072b8a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Specify disabled=false explicitly for contextual menu items and subitems",
+  "packageName": "@fluentui-react-native/contextual-menu",
+  "email": "adgleitm@microsoft.com",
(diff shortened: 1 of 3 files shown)


rurikoaraki (approved):

Saadnajmi:
Do we also need to update MenuV1?

amgleitman:
> Do we also need to update MenuV1?

The bug does repro in MenuV1. I'll look into fixing that as well in a separate PR.

PR #3339, opened 2024-01-04
Delete colors pass 5

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Remove unused colors. Checked some clients for usage. 
This is the final pass of color removal. All other colors are used by either our own controls or by clients.


Saadnajmi:
"All other colors are used by either our own controls or by clients."
 I assume this includes the legacy components too. If there are a few that can be refactored to use new colors, would that be virtuous? Or would that maybe cause issues with people using `.customize()`?

Saadnajmi (approved):

rurikoaraki:
> "All other colors are used by either our own controls or by clients." I assume this includes the legacy components too. If there are a few that can be refactored to use new colors, would that be virtuous? Or would that maybe cause issues with people using `.customize()`?

Most usages in clients are to fill in theme colors or assigning colors directly, so I think customize should be ok, but I don't think the work to move old components to use new colors would be trivial. I think it would be better to get clients to move to new components, since we want that to happen anyway.

Issue #3163, opened 2023-10-19
ThemeProvider not updating the state when themeRef changes


imbaky:
### Willing to submit a PR to fix?

- [X] I am willing to submit a PR to fix

### Requested priority

High

### Package version(s)

@fluentui-react-native/theme": "^0.9.1

### Platform

- [X] iOS
- [ ] macOS
- [ ] win32
- [ ] windows
- [X] android

### Please provide a reproduction of the bug

If you have a code similar to the one below which can change the themeRef passed to the themeProvider:
```ts
render(){

  let currentThemeRef = useMemo(()=>{

    return defaultTheme ? new ThemeReference(defaultFluentTheme) :  new ThemeReference(defaultFluentTheme, ()=>{
                 return {colors: {  background: '#f8b0b0' /* pink */,} };
      });
  }, [props.Theme]);


return (
<ThemeProvider theme={currentThemeRef}>
           <Button />
<ThemeProvider>
);

}

```

### Actual behavior

When the themeRef changes the ThemeProvider state doesn't change.

https://github.com/microsoft/fluentui-react-native/blob/bb3ab25ec0411eba461d733be17d6c1147de80d6/packages/framework/theme/src/ThemeProvider.tsx#L18
```ts
export const ThemeProvider: React.FunctionComponent<ThemeProviderProps> = (props: ThemeProviderProps) => {
  const { theme: themeRef, children } = props;
  const [theme, setThemeState] = React.useState(() => themeRef.theme);

  React.useEffect(() => {
    const onInvalidate = () => {
      setThemeState(themeRef.theme);
    };

    themeRef.addOnThemeChanged(onInvalidate);
    return () => {
      themeRef.removeOnThemeChanged(onInvalidate);
    };
  }, [themeRef, setThemeState]);
```

### Expected behavior

When the themeRef changes the ThemeProvider state should change.

 Proposed fix:

```ts
export const ThemeProvider: React.FunctionComponent<ThemeProviderProps> = (props: ThemeProviderProps) => {
  const { theme: themeRef, children } = props;
  const [theme, setThemeState] = React.useState(() => themeRef.theme);

  React.useEffect(() => {

    // Proposed fix
    setThemeState(themeRef.theme);

    const onInvalidate = () => {
      setThemeState(themeRef.theme);
    };

    themeRef.addOnThemeChanged(onInvalidate);
    return () => {
      themeRef.removeOnThemeChanged(onInvalidate);
    };
[...]

PR #3370, opened 2024-01-18
ThemeProvider now correctly updates state when the passed ThemeReference changes

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Bug: #3163 

Currently, a ThemeProvider updates its stored theme whenever its ThemeReference prop listens to a change in its theme. However, when the ThemeReference prop itself changes, the ThemeProvider does not update its stored theme. 

This PR fixes this by setting the theme state variable first whenever the ThemeReference prop updates in a useEffect hook.

### Verification

Locally tested and verified. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/framework/theme/src/ThemeProvider.tsx
+++ b/packages/framework/theme/src/ThemeProvider.tsx
@@ -16,6 +16,9 @@ export const ThemeProvider: React.FunctionComponent<ThemeProviderProps> = (props
   const [theme, setThemeState] = React.useState(() => themeRef.theme);
 
   React.useEffect(() => {
+    // Result of https://github.com/microsoft/fluentui-react-native/issues/3163
+    setThemeState(themeRef.theme);
+
     const onInvalidate = () => {
       setThemeState(themeRef.theme);
     };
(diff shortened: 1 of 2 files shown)


packages/framework/theme/src/ThemeProvider.tsx
```diff
@@ -16,6 +16,9 @@ export const ThemeProvider: React.FunctionComponent<ThemeProviderProps> = (props
   const [theme, setThemeState] = React.useState(() => themeRef.theme);
 
   React.useEffect(() => {
+    // Result of https://github.com/microsoft/fluentui-react-native/issues/3163
```
rurikoaraki, line 19: nit: I don't think we need to reference the issue in the code, we can explain what the intent of the code is directly, or just leave it out. I think it's a bit cumbersome to direct a code reader to an issue to gain context for some code.

rurikoaraki (approved):

PR #3376, opened 2024-01-20
[CI] Update to Xcode 15.2

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Update our CI to Xcode 15.2 and macOS-13 images. Unfortunately, this means disabling the iOS E2E tests as they continue to fail (simulator won't boot) and I haven't found a resolution. 

### Verification

CI should pass
### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/.ado/variables/vars.yml
+++ b/.ado/variables/vars.yml
@@ -1,6 +1,6 @@
 variables:
   - name: VmImageApple
-    value: macos-13
+    value: macos-13-arm64
   - name: xcode_version
     value: 'Xcode 15.2'
   - name: xcode_path
--- a/.ado/variables/vars.yml
+++ b/.ado/variables/vars.yml
@@ -1,11 +1,11 @@
 variables:
   - name: VmImageApple
-    value: internal-macos12
+    value: macos-13
   - name: xcode_version
-    value: 'Xcode 14.2'
+    value: 'Xcode 15.2'
   - name: xcode_path
-    value: '/Applications/Xcode_14.2.app'
+    value: '/Applications/Xcode_15.2.app'
   - name: ios_version
-    value: '16.2.0'
+    value: '17.2.0'
   - name: ios_simulator
-    value: 'iPhone 14'
+    value: 'iPhone 15 (17.2.0)'
--- a/.ado/variables/vars.yml
+++ b/.ado/variables/vars.yml
@@ -1,6 +1,6 @@
 variables:
   - name: VmImageApple
-    value: macos-13-arm64
+    value: macos-13
   - name: xcode_version
     value: 'Xcode 15.2'
   - name: xcode_path
--- a/.ado/azure-pipelines.yml
+++ b/.ado/azure-pipelines.yml
@@ -115,7 +115,7 @@ jobs:
       vmImage: $(VmImageApple)
     variables:
       platform: 'ios'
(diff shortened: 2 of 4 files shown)


rurikoaraki (approved):

.ado/variables/vars.yml
```diff
@@ -1,11 +1,11 @@
...
+    value: macos-13
   - name: xcode_version
-    value: 'Xcode 14.2'
+    value: 'Xcode 15.2'
   - name: xcode_path
-    value: '/Applications/Xcode_14.2.app'
+    value: '/Applications/Xcode_15.2.app'
   - name: ios_version
-    value: '16.2.0'
+    value: '15.2.0'
```
rurikoaraki, line 9: was this meant to downgrade?

Saadnajmi: This was supposed to be 17.. nice catch

Saadnajmi:
@rurikoaraki @tido64 I confirmed from https://github.com/microsoft/fluentui-react-native/pull/3499 that it's not macOS 13 or Xcode 15 that is the problem, but the Azure Devops public image pools for `macos-12` and `macos-13`. I'm not sure why still, if I had to assume. probably something to do with SIP. I've also mostly confirmed that we don't have many alternative pools we can use for macOS images (@dannyvv fyi). So my thoughts are as follows:

- Dig deeper into why iOS E2E tests specifically fail on these images
- Run a test to see if the GitHub Action images _do_ allow the E2E tests to pass
- Keep folllowing through to see if we can get a different pool to use (I doubt this one).

Saadnajmi:
> @rurikoaraki @tido64 I confirmed from #3499 that it's not macOS 13 or Xcode 15 that is the problem, but the Azure Devops public image pools for `macos-12` and `macos-13`. I'm not sure why still, if I had to assume. probably something to do with SIP. I've also mostly confirmed that we don't have many alternative pools we can use for macOS images (@dannyvv fyi). So my thoughts are as follows:
> 
> * Dig deeper into why iOS E2E tests specifically fail on these images
> * Run a test to see if the GitHub Action images _do_ allow the E2E tests to pass
> * Keep folllowing through to see if we can get a different pool to use (I doubt this one).

EDIT:

By increasing timeouts it seems to be working now. So actually... I think the new VMs were just that much slower........ 🤦

Saadnajmi:
OK, post FHL update #2:

Between this PR and #3499 , I think I've only had like.. 2 of 10 successful runs with iOS E2E tests. I'm pretty sure it's because with the newer VMs, they are much more hardware constrained and the iOS simulator launches _much_ slower. Even when it passes, the total job takes ~1-1.5 hours, rather then the ~30 minutes it takes currently. Given that, I feel uneasy "upgrading" as it would be a devX decrease. I think my next test would be to rewrite the job in GitHub Actions, and see if that pool is any better (I suspect much better, and also macOS 14). @rurikoaraki @tido64 FYI. Also I'm not sure if React Native 0.73 requires Xcode 15, but this will eventually become blocking. React Native macOS most likely does.

tido64:
FWIW, 0.73 _does_ require Xcode 15. I tried compiling it with Xcode 14 when I was debugging the visionOS condition and it did not like it at all.

PR #3382, opened 2024-01-24
[mac/ios] Ensure explicit module imports and fix a few warnings

anandrajeswaran:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

- fluentui-react-native has lots of usage of other modules but doesn't import them. This seems to work in cocoapods (likely through the unfortunate magic of the module cache), but doesn't work in other build systems and is likely to break in a future version of Xcode. Add explicit imports for things we use.
  - Wrap in USE_REACT_AS_MODULE, which is not defined in this repo and thus doesn't actually add the imports. This can be defined in other repos which don't get the free module usage magically afforded by cocoapods' setup), but we can we remove these wrappers when rn 0.73 is merged and React is importable.
- Warnings are not treated as errors, but may be so in other build systems and are good to fix anyway
  - Use "" for path-relative intra-project imports (FRNCalloutManager.m)
  - Don't have multiple declarations of the same symbol in the same scope (RCTFocusZone.m)
  - Match nullability specifications between function declaration and implementation (FRNFontMetrics.m)
  
### Verification
All build-time, not run-time impact, so just local build

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Callout/macos/CalloutManager.swift
+++ b/packages/components/Callout/macos/CalloutManager.swift
@@ -1,4 +1,6 @@
+import AppKit
 import Foundation
+import React
 
 @objc(FRNCalloutManager)
 class CalloutManager: RCTViewManager {
--- a/packages/components/Callout/macos/CalloutManager.swift
+++ b/packages/components/Callout/macos/CalloutManager.swift
@@ -1,6 +1,8 @@
 import AppKit
 import Foundation
+#if USE_REACT_AS_MODULE
 import React
+#endif // USE_REACT_AS_MODULE
 
 @objc(FRNCalloutManager)
 class CalloutManager: RCTViewManager {
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -421,7 +421,7 @@ static BOOL ShouldSkipFocusZone(NSView *view)
 
 		// If the previous view is in a FocusZone, focus on its defaultKeyView
 		// (For FocusZoneActionTab, this is handled by becomeFirstResponder).
-		RCTFocusZone *focusZoneAncestor = GetFocusZoneAncestor(nextViewToFocus);
+		focusZoneAncestor = GetFocusZoneAncestor(nextViewToFocus);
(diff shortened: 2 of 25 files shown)


packages/components/Callout/macos/CalloutManager.swift
```diff
@@ -1,4 +1,6 @@
+import AppKit
 import Foundation
```
Saadnajmi, line 2: Should we still have the Foundation import?

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -421,7 +421,7 @@ - (NSView *)nextViewToFocusOutsideZone:(FocusZoneAction)action
 
 		// If the previous view is in a FocusZone, focus on its defaultKeyView
 		// (For FocusZoneActionTab, this is handled by becomeFirstResponder).
-		RCTFocusZone *focusZoneAncestor = GetFocusZoneAncestor(nextViewToFocus);
+		focusZoneAncestor = GetFocusZoneAncestor(nextViewToFocus);
```
Saadnajmi, line 424: Oof..

packages/experimental/Avatar/ios/AvatarViewManager.swift
```diff
@@ -1,5 +1,7 @@
+import AppKit
```
Saadnajmi, line 1: Remove / replace with UIKit

Saadnajmi (changes requested):

packages/components/Callout/macos/CalloutManager.swift
anandrajeswaran: They are separate module so good to import what we need

Saadnajmi (approved):
Approved, last change I ask is that we mention somewhere that the preprocessor macro `USE_REACT_AS_MODULE` is not defined anywhere in RN / Coocoapods but a different, stricter build system would set it

PR #3392, opened 2024-01-29
Remove win32 specific override for focus command

acoates-ms:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes
Remove win32 specific override for focus command.  Win32 correctly implements the focus method on UIManager and has for quite a while.


Saadnajmi (approved):

acoates-ms:
Looks like this uncovered another bug.  This fix will need to wait for https://github.com/microsoft/react-native-windows/pull/12692.

rurikoaraki:
@acoates-ms you can probably rebase the change now? We should have RNW/RNWin32 updated...

acoates-ms:
Yeah, giving it some time to allow it to settle in Office.  -- The new code will be incompatible with older RNWin32 versions - so going to wait for an additional Office fork to avoid cases where someone tries to ship a newer FURN version to an older fork.

(Basically, I will wait till next month before I push this in)

PR #3398, opened 2024-01-31
Fix bug where macOS CalloutView renders offscreen in RTL edge case

FalseLobster:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In LTR layouts, we do work to make sure that our callouts, when anchored, are fully rendered on screen when snapping to the edge would make part of the control render off screen.  However, while we do make sure this works if we render off screen past the extents of the far right X dimension, we do not take into account RTL situations when we render off screen in the negative X dimension relative to the screen.

This simply adds a check when such a situation happens and pushes the origin on the new callout to 0 to ensure the contents are fully visible, as other macOS menus behave.

### Verification

|Screenshot 2024-01-30 at 3 51 29 PM |Screenshot 2024-01-30 at 3 52 07 PM|
Screenshot 2024-01-30 at 3 50 45 PM|Screenshot 2024-01-30 at 3 47 34 PM|

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/macos/Podfile.lock
+++ b/apps/fluent-tester/macos/Podfile.lock
@@ -1,25 +1,25 @@
 PODS:
   - boost (1.76.0)
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.72.12)
-  - FBReactNativeSpec (0.72.12):
+  - FBLazyVector (0.72.15)
+  - FBReactNativeSpec (0.72.15):
     - RCT-Folly (= 2021.07.22.00)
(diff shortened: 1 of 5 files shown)


amgleitman (approved):

Saadnajmi (approved):

Saadnajmi:
@falselobster Look like this needs change files, could you push those when you get a chance?

PR #3407, opened 2024-02-02
Use Correct checkmarkPadding for MenuItemCheckbox

JunielKatarn:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Fixes size and padding of the checkmark icon used in MenuItemCheckboxTokens.

### Verification

1. Built and started the tester app.
1. Opened the `Menu` section.
1. Clicked on the only button in the `Menu Radioitem` section.
1. Enabled the last item (the only check box control).

| 
image
 | 
image
 |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/apps/fluent-tester/macos/Podfile.lock
+++ b/apps/fluent-tester/macos/Podfile.lock
@@ -13,11 +13,11 @@ PODS:
   - FRNAvatar (0.20.7):
     - MicrosoftFluentUI (= 0.13.1)
     - React
-  - FRNCallout (0.25.8):
+  - FRNCallout (0.25.7):
     - React
   - FRNCheckbox (0.16.7):
     - React
-  - FRNMenuButton (0.12.13):
+  - FRNMenuButton (0.12.12):
     - React
   - FRNRadioButton (0.20.9):
     - React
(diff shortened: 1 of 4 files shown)


apps/fluent-tester/macos/Podfile.lock
```diff
@@ -704,4 +704,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 0e740f9fca1901b692b84c05390764ca72e2fb41
 
-COCOAPODS: 1.14.2
+COCOAPODS: 1.15.0
```
Saadnajmi, line 707: Could you revert this file from the change? cocoapods 1.15 actually isn't entirely compatible with React Native. I also usually don't update the podfile lock because it's kinda useless and causes lots of conflicts.

packages/components/Menu/src/MenuItemCheckbox/MenuItemCheckboxTokens.macos.ts
```diff
@@ -7,7 +7,7 @@ import type { MenuItemCheckboxTokens } from './MenuItemCheckbox.types';
 export const defaultMenuItemCheckboxTokens: TokenSettings<MenuItemCheckboxTokens, Theme> = (t: Theme): MenuItemCheckboxTokens => ({
   backgroundColor: t.colors.transparentBackground,
   borderRadius: 5,
-  checkmarkPadding: globalTokens.sizeNone,
+  checkmarkPadding: globalTokens.size20,
```
Saadnajmi, line 10: Do we need any changes to `MenuItemRadioTokens` too?

Saadnajmi (changes requested):

packages/components/Menu/src/MenuItemCheckbox/MenuItemCheckboxTokens.macos.ts
JunielKatarn: Not for this bug, and not that I can tell for sure.
I took that value precisely from `MenuItemRadioTokens`.

Other properties also differ between these controls but I don't know if they are justified to do so.

Saadnajmi (approved):

PR #3423, opened 2024-02-14
Add safety checks for `MenuList` max height/width specifications

amgleitman:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

We've seen some weird behavior in FURN menus where there would be some extra padding on the end or bottom. We believe this is due to `hasMaxWidth` or `hasMaxHeight` being undefined. This sort of thing has caused us headaches in the past (see #3333). This fix is along the same vein.

### Verification

Validated that this part of the code no longer complains if we turn on `strictNullChecks` in our tsconfig.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
The extra padding is a side effect of the RN macOS scroll view adding padding to its' shadow view based on whether we always show scroll bars, or only show scroll bars while scrolling. It's a bit complicated, but I can make a summary if needed.

Saadnajmi (approved):

PR #3429, opened 2024-02-15
Rework package json version notation

rrfang:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This makes a broad change reworking the notation that we're using to specify which package versions to use for dependencies.  Instead of using the wide range ">=xxx <yyy", we use the carat notation "^xxx" because the wide range notation was causing duplicate dependency imports (when mixing with the carat notation).  

### Verification

Everything still works in the tester app.  I would be interested to hear folks' thoughts on how I should go about testing to make sure we're not causing any issues for downstream users of FURN components.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi:
> the wide range notation was causing duplicate dependency imports (when mixing with the carat notation)

Oh wow, thanks for catching that! Do you know why that happens? Overall I'm in favor of this change, but I'll let others also comment.

Saadnajmi (approved):

rrfang:
> > the wide range notation was causing duplicate dependency imports (when mixing with the carat notation)
> 
> Oh wow, thanks for catching that! Do you know why that happens? Overall I'm in favor of this change, but I'll let others also comment.

From what I understand about the issue is that when we have two competing dependency imports say "^0.5.0" and ">=0.5.0 <1.0.0", we run into an issue because "^0.5.0" is actually treated as "0.5.x" where the leading zero is ignored and the first value, 5 in this case, is locked in.  

When we upgrade the dependency to 0.6.0, any package.json with the ^0.5.0 will keep using version 0.5.0, where as packages that specify ">=0.5.0 <1.0.0" will use the upgraded 0.6.0.

rurikoaraki:
@rrfang looks like this ran into some conflicts. It sounds like some partners are blocked on this work, could take another look please?

NikoAri (approved):

Saadnajmi:
Thanks for the change @rrfang! 🎆

NikoAri:
> > the wide range notation was causing duplicate dependency imports (when mixing with the carat notation)
> 
> Oh wow, thanks for catching that! Do you know why that happens? Overall I'm in favor of this change, but I'll let others also comment.

Tommy and I found this problem where users weren't able to auto generate proper Yarn.lock files. And manual Yarn.lock crafting is pretty tedious.

PR #3434, opened 2024-02-20
Adding useControllableValue hook

riosn-n:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

This PR adds a new hook used to manage state in components that can be controlled or uncontrolled based on which props the consumer provides (such as the Input component's value vs. defaultValue).

The implementation and tests are direct copies from:

https://github.com/microsoft/fluentui/blob/master/packages/react-hooks/src/useControllableValue.ts

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Input/src/useInput.ts
+++ b/packages/components/Input/src/useInput.ts
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { createIconProps } from '@fluentui-react-native/icon';
 import type { IconProps } from '@fluentui-react-native/icon';
-import { usePressableState } from '@fluentui-react-native/interactive-hooks';
+import { usePressableState, useControllableValue } from '@fluentui-react-native/interactive-hooks';
 
 import { DismissSvg } from './assets/dismissSvg';
 import type { InputProps, InputInfo } from './Input.types';
@@ -31,10 +31,10 @@ export const useInput = (props: InputProps): InputInfo => {
     ...rest
   } = props;
   const pressable = usePressableState({ onBlur, onFocus });
-  const [text, setText] = React.useState<string>(defaultValue ? defaultValue : '');
   const defaultIconProps = createIconProps(defaultIcon);
   const focusedIconProps = createIconProps(focusedStateIcon);
   const [iconProps, setIconProps] = React.useState<IconProps>(defaultIconProps);
+
   React.useEffect(() => {
     if (pressable.state.focused && !error && focusedIconProps) {
       setIconProps(focusedIconProps);
@@ -42,6 +42,15 @@ export const useInput = (props: InputProps): InputInfo => {
       setIconProps(defaultIconProps);
     }
   }, [error, pressable.state.focused, defaultIconProps, focusedIconProps]);
+
+  const onChangeText = React.useCallback(
+    (_ev: unknown, text: string) => {
+      onChange?.(text);
+    },
+    [onChange],
+  );
+
+  const [text, setText] = useControllableValue(value, defaultValue, onChangeText);
 
   return {
     props: {
@@ -53,17 +62,15 @@ export const useInput = (props: InputProps): InputInfo => {
       onChange,
       accessoryIcon,
       accessoryButtonOnPress,
-      value,
+      value: text,
       // Directly applied onto the TextInput
       textInputProps: {
         ...textInputProps,
         keyboardType: type,
         placeholder,
-        defaultValue,
-        value: value ? value : text,
+        value: text,
         onChangeText: (text) => {
-          !value && setText(text);
-          onChange && onChange(text);
+          setText(text);
         },
         ref: componentRef,
         accessibilityLabel,
@@ -78,6 +85,6 @@ export const useInput = (props: InputProps): InputInfo => {
       componentRef,
       ...rest,
     },
-    state: { ...pressable.state, text: value ? value : text },
+    state: { ...pressable.state, text },
   };
 };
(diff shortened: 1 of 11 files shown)


packages/utils/interactive-hooks/src/useConst.ts
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+
+/**
+ * Hook to initialize and return a constant value. Unlike `React.useMemo`, this is guaranteed to
```
Saadnajmi, line 4: I'm slightly confused why this needs to be a hook rather than `const var = <constant>`, but I know this is coming from FluentUI React.

packages/utils/interactive-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,69 @@
...
+  defaultUncontrolledValue: TValue | undefined,
+): Readonly<[TValue | undefined, (update: React.SetStateAction<TValue | undefined>) => void]>;
+
+export function useControllableValue<TValue, TElement, TEvent extends React.SyntheticEvent<TElement> | undefined>(
+  controlledValue: TValue | undefined,
+  defaultUncontrolledValue: TValue | undefined,
+  onChange: ValueChangeCallback<TElement, TValue, TEvent> | undefined,
+): Readonly<[TValue | undefined, (update: React.SetStateAction<TValue | undefined>, ev?: React.FormEvent<TElement>) => void]>;
+
+export function useControllableValue<TValue, TElement extends HTMLElement, TEvent extends React.SyntheticEvent<TElement> | undefined>(
```
Saadnajmi, line 33: Can `TElement` not extend HTMLElement?

packages/utils/test-tools/src/enzymeTests.tsx
```diff
@@ -86,3 +88,51 @@ export function checkReRender(render: JSXProducer, depth: number = 1) {
...
+  it(testDescription || 'returns the same value(s) each time', () => {
+    let latestValues: TValues | undefined;
+    let callCount = 0;
+
+    const TestComponent: React.FunctionComponent = () => {
+      callCount++;
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      latestValues = callCount === 1 ? useHook() : (useHookAgain || useHook)();
+      return <div />;
+    };
```
Saadnajmi, line 115: Feels iffy we're conditionally calling a hook. Was this also copied from FluentUI React?

Saadnajmi: Also, return a `<View>`, not a `<div>`?

Saadnajmi (changes requested):
Could we attach a before and after video of how the `Input` control behavior changes with this PR?

packages/utils/interactive-hooks/src/useConst.ts
riosn-n: If you do something like `const address = { city: 'redmond', state: 'wa' }` the identity of that object will change (it will effectively be a different object stored in `address`) every time that line gets executed. 

`useConst` ensures that the object doesn't change.

packages/utils/test-tools/src/enzymeTests.tsx
riosn-n: Yes, the test was copied from here:

https://github.com/microsoft/fluentui/blob/6a2ebfd797c041225db94e6550e41e0e59672fb7/packages/react-hooks/src/testUtilities.tsx#L13

I changed the <div> for a <View>

riosn-n:
> Could we attach a before and after video of how the `Input` control behavior changes with this PR?

The behavior doesn't change at all (neither visually nor in the way callbacks are called). 

The goal here was to be fully backward compatible.

packages/utils/interactive-hooks/src/useConst.ts
Saadnajmi: Ah, that makes sense, thanks!

packages/utils/test-tools/src/enzymeTests.tsx
Saadnajmi: I'll resolve this then. I imagine because it's a test on the hook, it has to do weird stuff

packages/utils/interactive-hooks/src/useControllableValue.ts
riosn-n: Good catch. Changed.

packages/components/Input/src/useInput.ts
```diff
@@ -43,6 +43,15 @@ export const useInput = (props: InputProps): InputInfo => {
     }
   }, [error, pressable.state.focused, defaultIconProps, focusedIconProps]);
 
+  const onChangeText = React.useCallback(
+    (_ev: unknown, text: string) => {
```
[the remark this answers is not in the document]
riosn-n: I added the type to the callback 👍

Saadnajmi (approved):

packages/utils/interactive-hooks/src/__tests__/useControllableValue.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { mount } from 'enzyme';
+
+import { useControllableValue } from '../useControllableValue';
+
+describe('useControllableValue', () => {
+  it('respects controlled value', () => {
+    let resultValue: boolean | undefined;
+    const TestComponent: React.FunctionComponent<{ value?: boolean; defaultValue?: boolean }> = ({ value, defaultValue }) => {
+      [resultValue] = useControllableValue(value, defaultValue);
+      return <div />;
```
rurikoaraki, line 13: Does <div> make sense given that this is a React Native library? I think View seems like a better fit.

riosn-n: Removed. Thank you.

rurikoaraki (approved):

PR #3457, opened 2024-02-28
[FURN Menu] Fix MenuGroup Accessibility

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

#### Problem:
MenuGroup does not create a new group in the accessibility tree that's a parent to the MenuItems under it. This is a problem for users who use screen readers because they will not know the group context when navigating through the menu items.

#### Solution:
Assign the "group" accessibility role to MenuGroup so assistive technology is aware of the group context and assign the MenuGroupHeader text as the accessibilityLabel of the MenuGroup if it is present.

Note: On win32, an accessibilityLabel must be provided in order for the group control to be read out so we set the default accessibilityLabel to a blank space ("[space]").

### Verification

Manually tested with Narrator and verified with A11yInsights

| image | image |
| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [x] Voiceover


packages/components/Menu/src/MenuGroup/MenuGroup.tsx
```diff
@@ -48,8 +48,27 @@ export const MenuGroup = compose<MenuGroupType>({
...
+      // Check if this MenuGroup has a MenuGroupHeader. If so, use the MenuGroupHeader
+      // as the accessibilityLabel for the MenuGroup.
+      const menuGroupHeader = React.Children.toArray(children).find(
+        (child) => React.isValidElement(child) && (child as any).type.displayName === 'MenuGroupHeader',
+      );
+
+      // In order for assistive technology to read a group, it must have a name which is why
+      // we use a string with a space as the default accessibilityLabel for MenuGroup.
+      // If an empty string was used, the group context would not be read.
+      let menuGroupA11yLabel = ' ';
```
ksiler, line 60: @amgleitman @Saadnajmi would either or you know if an accessibilitylabel is required in order for a group to be announced on mac like it is on win32?

Saadnajmi: It seems to be announcing an extra group. I don't know what we want to do with groups on macOS sicne native menus don't really have them / don't expose them to VoiceOver at all. Even if there are groups and sections, macOS seems to just treat it like a flat list (see end of my video where I bring up a native menu.

https://github.com/microsoft/fluentui-react-native/assets/6722175/c18d669d-88f2-43b9-9a67-3187bfcb1d44

Saadnajmi: Without your change, it looks like this:


https://github.com/microsoft/fluentui-react-native/assets/6722175/e06a2e35-b4dc-4608-8868-f31c0744ac3a

ksiler: Thanks, this is helpful! In that case, I'll just make this change for win32.

PR #3461, opened 2024-02-29
[FURN Menu] Add Timestamp Per Instance

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Changing to use a new timestamp per instance to avoid disruptions in the opening behavior between two different menus. Previously, the global timestamp resulted in issues with opening a second menu when the time between clicks was very short. 

### Verification

Ran automation tests and manually checked that menu open behavior appeared correct.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/Menu/useMenu.ts
+++ b/packages/components/Menu/src/Menu/useMenu.ts
@@ -7,13 +7,6 @@ import { isMouseEvent } from '@fluentui-react-native/interactive-hooks';
 import type { MenuProps, MenuState } from './Menu.types';
 import { useMenuContext } from '../context/menuContext';
 
-// Due to how events get fired we get double notifications
-// for the same event causing us to immediately reopen
-// a menu when we close it. Adding in a delay to prevent
-// this behavior.
-const delayOpen = 150;
-let lastCloseTimestamp = -1;
-
 export const useMenu = (props: MenuProps): MenuState => {
   const triggerRef = React.useRef();
   const context = useMenuContext();
@@ -61,6 +54,13 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  // Due to how events get fired we get double notifications
+  // for the same event causing us to immediately reopen
+  // a menu when we close it. Adding in a delay to prevent
+  // this behavior.
+  const delayOpen = 150;
+  const [lastCloseTimestamp, setLastCloseTimestamp] = React.useState<number>(-1);
+
   const setOpen = React.useCallback(
     (e: InteractionEvent, isOpen: boolean, bubble?: boolean) => {
       const openPrev = state;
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/Menu/useMenu.ts
```diff
@@ -61,6 +54,13 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  // Due to how events get fired we get double notifications
+  // for the same event causing us to immediately reopen
+  // a menu when we close it. Adding in a delay to prevent
+  // this behavior.
+  const delayOpen = 150;
+  const [lastCloseTimestamp, setLastCloseTimestamp] = React.useState<number>(-1);
```
rurikoaraki, line 62: I think using a ref would be better - state causes a new render cycle, and the state wouldn't update to the new value until that render cycle is complete. This value isn't needed directly to render anything i.e. changing it shouldn't cause any renders

packages/components/Menu/src/Menu/useMenu.ts
```diff
@@ -61,6 +54,13 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  // Due to how events get fired we get double notifications
+  // for the same event causing us to immediately reopen
+  // a menu when we close it. Adding in a delay to prevent
+  // this behavior.
+  const delayOpen = 150;
```
rurikoaraki, line 61: This can stay as a global constant. This doesn't need to be stored separately per component instance.

packages/components/Menu/src/Menu/useMenu.ts
```diff
@@ -61,6 +54,13 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  // Due to how events get fired we get double notifications
+  // for the same event causing us to immediately reopen
+  // a menu when we close it. Adding in a delay to prevent
+  // this behavior.
```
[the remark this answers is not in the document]
gulnazsayed: It looks like it was added [here](https://github.com/microsoft/fluentui-react-native/pull/1733) to avoid flickering behavior for the keyboard.

packages/components/Menu/src/Menu/useMenu.ts
gulnazsayed: That makes sense! Making the changes and posting it in a new PR that's based off the main branch

packages/components/Menu/src/Menu/useMenu.ts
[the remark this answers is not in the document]
Saadnajmi: Is there a way we can read `event.target` , and perhaps ignore the extra events that are coming from a different react view? Reason being I think React doesn't like timestamps with renders much, and also because a timestamp feels like a workaround for a workaround of an issue.

PR #3462, opened 2024-03-01
Use Workspace for version for local packages

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Converting local packages to use workspace:* for our dependencies. This has the side effect of making our yarn.lock more stable between version bumps.

With appium not needing explicit install calls to work anymore, this might work now.

### Verification

CI will verify appium, which was the main issue last time this was attempted.


Saadnajmi (approved):

Saadnajmi (approved):

packages/dependency-profiles/package.json
```diff
@@ -56,7 +56,7 @@
     "@fluentui-react-native/popover": "*",
     "@fluentui-react-native/pressable": "*",
     "@fluentui-react-native/radio-group": "*",
-    "@fluentui-react-native/scripts": "^0.1.1",
+    "@fluentui-react-native/scripts": "workspace:*",
```
Saadnajmi, line 59: Why is this different?

Saadnajmi (commented):
Can we confirm with a publish dry run that beach ball handled this properly?

rurikoaraki:
> Can we confirm with a publish dry run that beach ball handled this properly?

Sooo I tried a dry run but I can't find any meaningful logs or the package that was produced... there is code in beachball to handle this case though.

Saadnajmi:
> > Can we confirm with a publish dry run that beach ball handled this properly?
> 
> 
> 
> Sooo I tried a dry run but I can't find any meaningful logs or the package that was produced... there is code in beachball to handle this case though.

Cool. We can back out if needed. Ship ittttt

PR #3467, opened 2024-03-04
[FURN Menu] Add Timestamp Per Instance

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Changing to use a new timestamp per instance to avoid disruptions in the opening behavior between two different menus. Previously, the global timestamp resulted in issues with opening a second menu when the time between clicks was very short.

### Verification

| flyout before | flyout after |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/Menu/useMenu.ts
+++ b/packages/components/Menu/src/Menu/useMenu.ts
@@ -12,7 +12,6 @@ import { useMenuContext } from '../context/menuContext';
 // a menu when we close it. Adding in a delay to prevent
 // this behavior.
 const delayOpen = 150;
-let lastCloseTimestamp = -1;
 
 export const useMenu = (props: MenuProps): MenuState => {
   const triggerRef = React.useRef();
@@ -61,10 +60,12 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  const lastCloseTimestampRef = React.useRef<number>(-1);
+
   const setOpen = React.useCallback(
     (e: InteractionEvent, isOpen: boolean, bubble?: boolean) => {
       const openPrev = state;
-      if (!isControlled && (!isOpen || lastCloseTimestamp + delayOpen <= Date.now())) {
+      if (!isControlled && (!isOpen || lastCloseTimestampRef.current + delayOpen <= Date.now())) {
         setOpenInternal(isOpen);
       }
 
@@ -78,7 +79,7 @@ const useMenuOpenState = (
 
       if (!isOpen) {
         setShouldFocusOnContainer(undefined);
-        lastCloseTimestamp = Date.now();
+        lastCloseTimestampRef.current = Date.now();
       }
 
       if (onOpenChange && openPrev !== isOpen) {
--- a/change/@fluentui-react-native-menu-3f6d4b03-ffff-45d0-9b72-8d668c0f8ab8.json
+++ b/change/@fluentui-react-native-menu-3f6d4b03-ffff-45d0-9b72-8d668c0f8ab8.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "timestamp per instance",
+  "packageName": "@fluentui-react-native/menu",
+  "email": "gulnazsayed@microsoft.com",
+  "dependentChangeType": "patch"
+}


gulnazsayed:
> Left the comment in the previous PR, leaving it here again here:
> 
> Could we possibly check `event.target` and ignore the events not coming from our View instead? Using timestamps feels like a workaround of a workaround.

@Saadnajmi Could you explain your suggestion a little more?

Spoke with @rurikoaraki offline and from our understanding it doesn't seem possible to do it that way because of the way the events are. For example, looking at [this PR](https://github.com/microsoft/fluentui-react-native/pull/1733), here's what we expect is happening in the lower GIF:
1. Menu is open
2. Click down on button, light dismisses menu. Controlled by native. Not sure what mouse event we get / send for this without investigating further, but it's likely "onMouseDown" event.
3. Releasing mouse click gives an "onClick" event on trigger, which reopens the Menu.
It's likely the same thing is happening for the keyboard invoke.

Saadnajmi:
@gulnaz I'll try to do some testing in a bit

rurikoaraki:
> @gulnazsayed I'll try to do some testing in a bit Edit: It seems I had a similar idea at that PR too :D [#1733 (comment)](https://github.com/microsoft/fluentui-react-native/pull/1733#discussion_r896227138)

"React doesn't like timestamps with renders" I don't understand what this means, can you explain why React doesn't like timestamps? I also don't quite understand how your suggestion here is similar to the one you left on my original PR... One seems to be about emitting a new custom RN event from JS while here you're suggesting checking metadata on the event? Could you be more specific?

Saadnajmi:
> > @gulnazsayed I'll try to do some testing in a bit Edit: It seems I had a similar idea at that PR too :D [#1733 (comment)](https://github.com/microsoft/fluentui-react-native/pull/1733#discussion_r896227138)
> 
> "React doesn't like timestamps with renders" I don't understand what this means, can you explain why React doesn't like timestamps? I also don't quite understand how your suggestion here is similar to the one you left on my original PR... One seems to be about emitting a new custom RN event from JS while here you're suggesting checking metadata on the event? Could you be more specific?

For React + timestamps, the issue I think does not apply here, so my apologies. The issue is if there's a timestamp mismatch between hydration and render (a bit above my head, but we're not rendering time stamps so I think we're good): https://react.dev/reference/react-dom/client/hydrateRoot#suppressing-unavoidable-hydration-mismatch-errors

I more have an issue with timestamps in general. What if we have a slow computer, or a very fast one? What if a thread is hanging. It's inherently machine dependent. And the real issue (if I understand correctly) is that we have an event coming from multiple sources and we want to ignore some of them. I feel like the "correct" solution would be to somehow read the events, see which ones we want to respond to, and filter out the ones we don't want. This relates to the "custom" event I mentioned earlier because then we have control over what events Menu is listening and responding to.

acoates-ms:
+1 Using time between timestamps to affect behavior sounds horrible.

packages/components/Menu/src/Menu/useMenu.ts
```diff
@@ -61,10 +60,12 @@ const useMenuOpenState = (
 
   const state = isControlled ? open : openInternal;
 
+  const lastCloseTimestampRef = React.useRef<number>(-1);
+
   const setOpen = React.useCallback(
     (e: InteractionEvent, isOpen: boolean, bubble?: boolean) => {
       const openPrev = state;
-      if (!isControlled && (!isOpen || lastCloseTimestamp + delayOpen <= Date.now())) {
+      if (!isControlled && (!isOpen || lastCloseTimestampRef.current + delayOpen <= Date.now())) {
```
Saadnajmi, line 68: An aside, you'd probably want to update `useMenu.android` too

rurikoaraki:
Ok, we can take this back to the drawing board and see if we can come up with a solution that doesn't rely on timestamps.

Saadnajmi (approved):
Approving since I didn't realize we were already using timestamps + a delay, and this PR just extends that solution. I'll keep digging to see if I find an alternate solution, but I don't want to block

packages/components/Menu/src/Menu/useMenu.ts
gulnazsayed: Fixed!

gulnazsayed:
> I would test that closing a submenu via keyboard invoke still works correctly with this change.

Confirmed it is working correctly.

rurikoaraki (approved):

PR #3475, opened 2024-03-06
Minor Tab adjustments for better compatibility with Overflow / OverflowItem components.

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR introduces some minor changes to the Tab component that it needs for compatibility with the upcoming Overflow component. 

- Add callback chaining to the `onLayout` callback returned by the `useTabAnimation` hook, such that the parent OverflowItem's `onLayout` runs. 
- Add `numberOfLines` token to allow for ellipsification of Tab's text slot + minor styling adjustments.  

### Verification

Tested locally. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/components/TabList/src/Tab/Tab.types.ts
```diff
@@ -109,6 +109,14 @@ export interface TabProps extends Omit<PressablePropsExtended, 'onPress'> {
    */
   icon?: IconProps;
 
+  /**
+   * Conditional flag to disable the default Tab's onLayout handler from sending animation values to the TabList.
+   *
+   * If you want to set the onLayout handler from useTabAnimation() on your own custom element, this must be set to
+   * true so that animations run correctly.
+   */
+  calculateAnimationValuesExternally?: boolean;
```
rurikoaraki, line 118: Why do we need a separate bool for this? Is it not possible to check if onLayout is defined and alter behavior accordingly? This feels like it's leaking implementation details.

packages/components/TabList/src/Tab/Tab.tsx
```diff
@@ -74,7 +74,14 @@ export const Tab = compressible<TabProps, TabTokens>((props: TabProps, useTokens
       }
     });
 
-    const { icon, tabKey, ...mergedProps } = mergeProps(tab.props, final, {
+    // `onLayout` is unused and excluded from the rest of the mergedProps to be passed into the RootSlot.
+    // This is to ensure that the chained layout callback created in useTabAnimation isn't overwritten.
+    const {
+      icon,
+      tabKey,
+      onLayout: _,
```
rurikoaraki, line 82: I think this is fine for now, and it's good that this comment is here, but we might get a request in the future to be able to handle children having onLayout working.

rurikoaraki (approved):

packages/components/TabList/src/Tab/Tab.tsx
lawrencewin: This is a consequence of an `onLayout` callback passed from `React.cloneElement` completely overriding the hook's `onLayout` callback. I still don't fully understand why that occurs with cloning vs just passing an `onLayout` prop, but this at least is a working work-around.

PR #3491, opened 2024-03-14
Add missing dependencies/devDependencies

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Fixing up a few yarn-related warnings w.r.t peer deps missing. Should help with some downstream clients.

If I'm honest I still don't quite know when it makes sense to declare something as a peerDep, so please let me know if an entry seems to be missing.

Notably the adapter package wasn't declaring the various RN packages as direct dependencies even though the code is using them directly. This should fix a few issues downstream.

### Verification

Most yarn warnings have disappeared. The main ones remaining relate to the fact that we're using an enzyme adapter package for React 17 while we're on react 18.

### DIFF
--- a/packages/utils/adapters/package.json
+++ b/packages/utils/adapters/package.json
@@ -25,19 +25,17 @@
     "url": "https://github.com/microsoft/fluentui-react-native.git",
     "directory": "packages/utils/adapters"
   },
-  "dependencies": {
+  "devDependencies": {
+    "@fluentui-react-native/eslint-config-rules": "workspace:*",
+    "@fluentui-react-native/scripts": "workspace:*",
     "@office-iss/react-native-win32": "^0.72.0",
+    "@react-native/metro-config": "^0.72.0",
+    "@types/jest": "^29.0.0",
+    "@types/react": "^18.2.0",
     "react": "18.2.0",
     "react-native": "^0.72.0",
     "react-native-macos": "^0.72.0",
     "react-native-windows": "^0.72.0"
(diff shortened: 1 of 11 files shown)


packages/utils/adapters/package.json
```diff
@@ -25,15 +25,19 @@
     "url": "https://github.com/microsoft/fluentui-react-native.git",
     "directory": "packages/utils/adapters"
   },
+  "dependencies": {
+    "@office-iss/react-native-win32": "^0.72.0",
```
acoates-ms, line 29: Wont this make it so that everyone using FURN will be required to install rn-win32, rn-macos and rn-windows?

That doesn't seem right?

Saadnajmi: I Think this is actually something that could be an optional peer dependency? which is not common but @Tido64 had me do it in RNTA so it's a supported thing: https://github.com/microsoft/react-native-test-app/pull/1803/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

rurikoaraki: Let me undo the changes to this package since we're chatting offline about how to properly fix it

Saadnajmi (approved):

rurikoaraki:
#3494 is a more comprehensive change, closing

PR #3469, opened 2024-03-05
Add prototyped Overflow package

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

This PR adds the Overflow and OverflowItem components, based off of customer demands, and it adds a test page for the component in the test app. 

The Overflow is a container which shows and hides `OverflowItems` based off of available width. It is based off of the FluentUI v9 web component with the same name: https://react.fluentui.dev/?path=/docs/components-overflow--default

### Verification

image

### Pull request checklist

This PR has considered (when applicable):
- [ ] Automated Tests
- [ ] Documentation and examples
- [ ] Keyboard Accessibility
- [ ] Voiceover
- [ ] Internationalization and Right-to-left Layouts

PR #3494, opened 2024-03-14
Add a bunch of missing peerDependencies

acoates-ms:
If package A depends on a package B that has a peerDependency then package A should also declare that package as a peerDependency so that the consumers of FURN can properly provide that peer.

In addition, peerDependencies on additional RN platforms should be marked as optional, so that users who do not care about those packages do not need to provide those platforms.


rurikoaraki:
Are the changes to adapters happening in a different change?

packages/components/Avatar/package.json
```diff
@@ -48,9 +48,12 @@
     "react-native-svg": "^13.14.0"
   },
   "peerDependencies": {
+    "@office-iss/react-native-win32": "^0.72.0",
     "react": "18.2.0",
     "react-native": "^0.72.0",
-    "react-native-svg": "^13.14.0"
+    "react-native-macos": "^0.72.0",
+    "react-native-svg": "^13.14.0",
+    "react-native-windows": "^0.72.0"
```
tido64, line 56: We should make the OOT packages optional with the `peerDependenciesMeta` field. Otherwise, you're telling consumers that they **must** also install them even though they don't care for the platform.

```
  "peerDependenciesMeta": {
    "@office-iss/react-native-win32": {
      "optional": true
    },
    "react-native-macos": {
      "optional": true
    },
    "react-native-windows": {
      "optional": true
    }
  },
```

rurikoaraki:
FYI another package was just added by #3469

packages/components/Avatar/package.json
tido64: Never mind, it looks like OOT platform types are being used all over the place. This will be difficult to untangle.

Saadnajmi: Most of the typings are centralized to the adapters package. Would you say untangling the types is a prerequisite to getting optional peer dependencies?

tido64: @acoates-ms is currently looking into it.

acoates-ms: https://github.com/microsoft/fluentui-react-native/pull/3497 will fix the adapter types to not require the rn-platform packages, so I've switched this back to making the rn-platform packages optional.

Saadnajmi (approved):

PR #3497, opened 2024-03-14
Make base adapter types be union of all platforms

acoates-ms:
This allows repos / build tasks which are not type checking against platform specific types the ability to use platform specific properties.


rurikoaraki (approved):

Saadnajmi (approved):
Long term, now that TS supports platform specific types, I feel like the best thing for FURN is to actually _not_ use the adapters at all, and just rely on the types the user is already importing?

Regardless, approving bc this is better. Though I hope it doesn't break mobile only users.

rurikoaraki (approved):

rurikoaraki (approved):

PR #3528, opened 2024-03-25
Remove a usage of findNodeHandle in fluent-tester

acoates-ms:
Removing a usage of `findNodeHandle` in `fluent-tester`.  This call was being used to call `UIManager.measure`, which is itself deprecated.

Replacing the call to `UIManager.measure` with `ref.measure`, also eliminates the call to `findNodeHandle`.


rurikoaraki (approved):

Saadnajmi (commented):
An aside.. pretty sure our slider is broken and we use the community module on platforms we actually need a slider 🤷

PR #3533, opened 2024-03-26
Use codegenNativeComponent so JS View configs can be generated for all native components

acoates-ms:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Currently all the FURN native components have no JS view config.  This is fine currently as the view config is provided by the native view mangers.  In the new architecture the view config will be provided by JS.  The view config will be codegen'd by default in 0.73 when we switch to using `@react-native/babel-preset`.

This is prep work for enabling the new architecture.

Moving all the NativeComponents into separate JS modules also removes the need for component-cache (which had questionable benefit anyway).  There are no more uses of the package in the repo, and I didn't see any uses in the internal repos either.  So I removed the whole package.

### Verification

I attempted to turn on the built in RN runtime config verification logic for the FURN components.  But in 0.72 at least, the view configs of RN-win32 are also quite out of sync.  So I got a lot of errors from the base view props.  But it did allow me to look through the list of missing properties to verify that there were none specific to the FURN components.

```js
const nativeComponentRegistry = require('react-native/Libraries/NativeComponent/NativeComponentRegistry');
// Modify the default RuntimeConfigProvider to verify that the JS view configs align with the native ones
// This helps keep them in sync so that when we switch to the JS view configs with fabric things keep working
nativeComponentRegistry.setRuntimeConfigProvider((name: string) => {
  // We dont need to verify configs that are not owned by FURN.
  const builtInConfigs = ['RCTView', 'RCTImageView', 'RCTActivityIndicatorView', 'RCTSwitch', 'RCTScrollView', 'RCTScrollContentView'];
  const builtInViewConfig = builtInConfigs.includes(name) || name.startsWith('RNSVG');
  return builtInViewConfig ? {
    native: true,
    strict: false,
    verify: false,
  } : {
    native: true,
    strict: true,
    verify: true,
  };
});
```

This change shouldn't affect anyone that hasn't either manually turned-on JS view specs by calling `setRuntimeConfigProvider` with `native: false`, or apps running with in bridgeless mode.


packages/components/MenuButton/src/MacOSMenuButtonNativeComponent.ts
```diff
@@ -0,0 +1,32 @@
...
+  identifier?: string;
+  hasSubmenu?: boolean;
+  submenu: MenuItem[];
+}
+
+export interface NativeProps extends ViewProps {
+  content?: string;
+  image?: ImageSourcePropType;
+  enabled?: boolean;
+  menu?: MenuItem;
```
rurikoaraki, line 26: tooltip missing?

packages/components/FocusZone/src/FocusZoneNativeComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { Int32, WithDefault } from 'react-native/Libraries/Types/CodegenTypes';
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+
+export interface NativeProps extends ViewProps {
+  navigateAtEnd?: WithDefault<'NavigateStopAtEnds' | 'NavigateWrap' | 'NavigateContinue', 'NavigateStopAtEnds'>;
+  defaultTabbableElement?: Int32;
+  focusZoneDirection?: WithDefault<'bidirectional' | 'vertical' | 'horizontal' | 'none', 'bidirectional'>;
+  use2DNavigation?: boolean;
+  tabKeyNavigation?: WithDefault<'None' | 'NavigateWrap' | 'NavigateStopAtEnds' | 'Normal', 'None'>;
+  disabled?: boolean;
```
rurikoaraki, line 18: isTabNavigation missing?

packages/components/FocusTrapZone/src/FocusTrapZoneNativeComponent.ts
```diff
@@ -0,0 +1,39 @@
...
+   */
+  focusPreviouslyFocusedInnerElement?: boolean;
+  /**
+   * By default, when the FocusTrapZone is unmounted or disabled focus returns to the element outside the FocusTrapZone that last had
+   * focus. Setting this to true disables this behavior, so the UI behaves as it normally would: If the FocusTrapZone is disabled, focus
+   * remains on the control inside the FocusTrapZone. If the FocusTrapZone is unmounted, focus usually moves to the nearest focusable
+   * control, but this can vary.
+   * @default false
+   */
+  ignoreExternalFocusing?: boolean;
```
rurikoaraki, line 36: forceFocusInsideTrap missing?

packages/components/MenuButton/src/MacOSMenuButtonNativeComponent.ts
acoates-ms: Looks like its toolTip in native:
https://github.com/microsoft/fluentui-react-native/blob/ced28105fa91b14cfe655651cc39799354adfba8/packages/components/MenuButton/macos/FRNMenuButtonManager.m#L51

but tooltip in JS:
https://github.com/microsoft/fluentui-react-native/blob/ced28105fa91b14cfe655651cc39799354adfba8/packages/components/MenuButton/src/MenuButton.macos.tsx#L111

Are properties not case sensitive? -- Is this a bug?

acoates-ms: looks like its tooltip - inherited from view.  -- So it shouldn't actually be part of this component spec.

Saadnajmi: I can't remember if the case sensitivity has bit us before. If we fix this on the native side with `RCT_REMAP_PROPERTY` or something, I'd be happy to do so.

Saadnajmi:
Any idea if the presence of the view configs affects paper? We could test this branch internally to check, mostly worried the macOS components might break in some way internally after this merges

packages/components/Callout/package.json
```diff
@@ -30,7 +30,6 @@
   },
   "dependencies": {
     "@fluentui-react-native/adapters": "workspace:*",
-    "@fluentui-react-native/component-cache": "workspace:*",
```
Saadnajmi, line 33: Mayyybe worth removing `component-cache` in a separate PR to keep this change smaller?

packages/components/Callout/src/Callout.tsx
```diff
@@ -11,10 +17,14 @@ import { mergeSettings } from '@uifabricshared/foundation-settings';
...
+import CalloutNativeComponent from './CalloutNativeComponent';
+import type { NativeProps as MacOSCalloutNativeProps } from './MacOSCalloutNativeComponent';
+import MacOSCalloutNativeComponent from './MacOSCalloutNativeComponent';
 
-const NativeCalloutView = Platform.select({
-  macos: ensureNativeComponent('FRNCallout'),
-  default: ensureNativeComponent('RCTCallout'), // win32
+const NativeCalloutView = Platform.select<HostComponent<MacOSCalloutNativeProps> | HostComponent<CalloutNativeProps>>({
+  macos: MacOSCalloutNativeComponent,
+  default: CalloutNativeComponent, // win32
```
Saadnajmi, line 27: or can we not remove component-cache easily because that means double changes in areas like this?

packages/components/Callout/src/CalloutNativeComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+  target?: UnsafeMixed;
+  // targetAnchor?: string; // Win32 only Callout can target an anchor registered in the anchor registry // Can be a node id or an anchor ID - This need to be reworked as Mixed types are not supported going forward
+  testID?: string;
+
+  onRestoreFocus?: DirectEventHandler<{ target: Int32; containsFocus: boolean }>;
+  onDismiss?: DirectEventHandler<{ target: Int32 }>;
+  onShow?: DirectEventHandler<{ target: Int32 }>;
+}
+
+export default codegenNativeComponent<NativeProps>('RCTCallout') as HostComponent<NativeProps>;
```
Saadnajmi, line 60: The reason macOS adopted `FRNCallout` instead of `RCTCallout` is because traditionally, the 3 letter prefix is a convention for name spacing in objective-c (which has a global namespace), and the `RCT` prefix was reserved by React Native, so we had to use something else. It would be nice to not have two different codegen configs here, but I'm not sure if it's better for win32 to rename their native control or for macOS to break the namespace convention.

packages/components/FocusZone/src/FocusZone.ts
```diff
@@ -1,15 +1,19 @@
...
-import { ensureNativeComponent } from '@fluentui-react-native/component-cache';
 import { useViewCommandFocus } from '@fluentui-react-native/interactive-hooks';
 import type { IUseStyling } from '@uifabricshared/foundation-composable';
 import { composable } from '@uifabricshared/foundation-composable';
 import { mergeSettings } from '@uifabricshared/foundation-settings';
 
 import type { FocusZoneProps, FocusZoneSlotProps, FocusZoneType } from './FocusZone.types';
-
-const RCTFocusZone = ensureNativeComponent('RCTFocusZone');
+import RCTFocusZone from './FocusZoneNativeComponent';
```
Saadnajmi, line 16: Probably out of scope, but could we also not leak the apple 3 letter prefix convention to JS code here? 😅 I know this is long existing code.

acoates-ms:
> Any idea if the presence of the view configs affects paper? We could test this branch internally to check, mostly worried the macOS components might break in some way internally after this merges

By default the JS view configs are only used in bridgeless.  You can opt in to using them on paper, but we dont anywhere.

Saadnajmi (approved):

PR #3543, opened 2024-04-03
Update to RN 0.73

acoates-ms:
Mostly package version bumps.

In 0.72 react-native-windows was not exporting its type info correctly, so the window type adapters were just using RN's types.  Now they are more complete. With 0.73, react-native-windows is more correctly exporting its TS types, which requires updates to our adapter types.  Now that we have both windows and win32 types in the adapters, it more clearly shows some of the inconsistencies between the two platforms which we should work to reduce.


Saadnajmi:
This will be blocked by https://github.com/microsoft/fluentui-react-native/pull/3376 as an FYI

packages/dependency-profiles/src/index.js
```diff
@@ -1,636 +1,946 @@
...
-    "@fluentui-react-native/tokens": {
-      "name": "@fluentui-react-native/tokens",
-      "version": "0.22.9"
-    }
   },
-  "0.71": {
-    "@fluentui-react-native/e2e-testing": {
-      "name": "@fluentui-react-native/e2e-testing",
-      "version": "1.32.8"
+  0.72: {
```
rurikoaraki, line 314: We have an existing 0.71 file I think? Do we need one for 0.72?

rurikoaraki:
FYI #3544 added a few more packages to dep profiles.

Saadnajmi:
For the macOS error, looks like codegen doesn't like the object `ImageSourcePropType` for MenuButton's native component. As a fix, we could probably just remove the native module altogether and use the JS version (which itself uses a Callout that _is_ implemented on macOS)

scripts/src/configs/configureJest.ts
```diff
@@ -18,6 +18,9 @@ export function configureJest(customConfig?: object): object {
       moduleNameMapper: {
         '\\.(scss)$': path.resolve(__dirname, 'jest/jest-style-mock.js'),
         KeyCodes: path.resolve(__dirname, 'jest/jest-mock.js'),
+        // Jest is picking up the hoisted version of lru-cache, which is
+        // incompatible from the version required by semver
+        'lru-cache': require.resolve('lru-cache', {paths:[require.resolve('semver')]}),
```
Saadnajmi, line 23: Could we fix this with a resolution? Or do we just _need_ to have two copies?

Saadnajmi: Actually, do we even need semver? I see it in our package.json but not actually imported in any JS

Saadnajmi:
The Android issue might be solved by updating the Android test app. Whenever RNTA makes updates to the template files, we don't automatically get them, unless we run `yarn init-test-app` again. I know there was a bunch of Android fixes recently.

rurikoaraki (approved):

.ado/templates/setup-repo.yml
```diff
@@ -1,9 +1,9 @@
 # This template install dependencies, builds, and bundles this repo
 steps:
-  - task: NodeTool@0
-    inputs:
-      versionSource: 'fromFile'
-      versionFilePath: '.node-version'
+  # - task: NodeTool@0
```
rurikoaraki, line 3: should this stay commented?

acoates-ms: The windows build doesn't work if we use NodeTool to install node (even though its already got node 18?  🤷.  -- And all the builds are working currently.  I can remove the code completely as a follow up PR.

PR #3580, opened 2024-04-18
FURN Tooltip Spec

gulnazsayed:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Added a spec file for FURN tooltip that outlines the expected behavior of the default tooltip. Added layoutEffect hook for setting the target on the JS side, this would be used later with the native implementation to get the target reference. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


packages/experimental/Tooltip/SPEC.md
```diff
@@ -2,4 +2,131 @@
...
+
+Not yet implemented.
+
+The `Tooltip` component can reference a target component that will be used as the position anchor.
+
+## API
+
+### Slots
+
+The `RadioGroup` component has one slot, or parts. The slots behave as follows:
```
rurikoaraki, line 57: nit: Should be `Tooltip`

packages/experimental/Tooltip/src/Tooltip.tsx
```diff
@@ -9,21 +10,33 @@ import { tooltipName } from './Tooltip.types';
...
+      setNativeTarget(findNodeHandle(target.current));
+    } else {
+      // Clear `target` so we may fall back on `anchorRect` if provided
+      setNativeTarget(null);
+    }
+  }, [target]);
+
+  const TooltipComponent = (rest: TooltipProps, children: React.ReactNode) => {
+    return (
+      <NativeTooltipView {...(nativeTarget && { target: nativeTarget })} {...mergeProps(props, rest)}>
```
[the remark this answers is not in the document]
gulnazsayed: We want to "process" the target, so the layout effect hook should process the target prop defined in props, then that's what we want to pass into the component.

rurikoaraki (approved):

Saadnajmi:
Related aside, the React Native team at Meta recently wrote some new architecture docs that included an example (with code) of a Tooltip :)

https://reactnative.dev/docs/the-new-architecture/landing-page#synchronous-layout-and-effects

packages/experimental/Tooltip/SPEC.md
```diff
@@ -2,4 +2,131 @@
...
+);
+```
+
+## Variants
+
+### Positioning
+
+Not yet implemented.
+
+The `Tooltip` component will appear in the specified position relative to the anchor (target). In Win32, the positions are denoted by the DirectionalHint: `topLeftEdge`, `topCenter`, `topRightEdge`, `rightTopEdge`, `rightCenter`, `rightBottomEdge`, `bottomLeftEdge`, `bottomCenter`, `bottomRightEdge`, `leftTopEdge`, `leftCenter`, and `leftBottomEdge`.
```
Saadnajmi, line 45: Can we align with FluentUI React and use the more react-style names? The names (which felt a bit win32 specific) was something I hard to hardcode into macOS Callout and didn't like
https://github.com/microsoft/fluentui/blob/4ad395276e01a2a04df8f0263e1af2467310bcfd/packages/react-components/react-positioning/src/types.ts#L244

```
export type PositioningShorthandValue =
  | 'above'
  | 'above-start'
  | 'above-end'
  | 'below'
  | 'below-start'
  | 'below-end'
  | 'before'
  | 'before-top'
  | 'before-bottom'
  | 'after'
  | 'after-top'
  | 'after-bottom';
  ```

File: packages/experimental/Tooltip/src/Tooltip.tsx [hunk 4]
```diff
@@ -9,21 +10,33 @@ import { tooltipName } from './Tooltip.types';
...
+
+  const [nativeTarget, setNativeTarget] = React.useState<number | string | null>(null);
+
+  React.useLayoutEffect(() => {
+    if (typeof target === 'string') {
+      // Pass string type `target` directly
+      setNativeTarget(target);
+    } else if (target?.current) {
+      // Pass the tagID for a valid ref `target`
+      setNativeTarget(findNodeHandle(target.current));
```
Saadnajmi, line 24: I know this is existing code and used elsewhere, but it's worth noting `findNodeHandle`, while it works in the new arch, isn't recommended and can break. I think that's why the link I sent earlier uses a Rect instead of a ref to position its' tooltip. It might be worth discussing whether we just go with using a Rect altogether

packages/experimental/Tooltip/SPEC.md
gulnazsayed: We had decided to align with the naming convention used in Callout to have consistency across the existing FURN controls. I'll make a note to get PM input on this though when we revisit Tooltip in the future.

packages/experimental/Tooltip/src/Tooltip.tsx
gulnazsayed: Thanks for sharing the link to the new architecture! I'll make a note to look into this / discuss this when we revisit Tooltip in the future.

PR #3589, opened 2024-04-23
[ios] Add missing include statement

anandrajeswaran:
### Platforms Impacted
- [x] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Problem:
We have a file which uses things from Foundation, but never actually declares that as an input. It's unclear how the compiler is able to resolve these symbols (guessing something related to module caching), but doesn't work when built with other build systems.

Fix:
Add the explicit include

### Verification

Just a local build and standard PR verification. Not product impacting.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

Saadnajmi:
My guess is we get it transitively through the RCTVIewManager.h import.

PR #3608, opened 2024-05-01
Use getConstants() when accessing constants from TurboModules/NativeModules

acoates-ms:
With NativeModules, constants are exposed both properties directly on the NativeModule, and as properties returned on an object from a getConstants() method.  With TurboModules the constants are only exposed from getConstants().

This change modifies our usage of constants to always use getConstants, to be compatible with the modules running as either NativeModules or TurboModules.

### DIFF
--- a/packages/experimental/Avatar/src/NativeAvatar.tsx
+++ b/packages/experimental/Avatar/src/NativeAvatar.tsx
@@ -26,7 +26,7 @@ interface ExportedConstants {
   sizes: { [key in Size]: number };
 }
 
-const ExportedNativeConstants: ExportedConstants = TurboModuleRegistry.get('FRNAvatarConstants') || NativeModules.FRNAvatarViewManager;
+const ExportedNativeConstants: ExportedConstants = TurboModuleRegistry.get('FRNAvatarConstants')?.getConstants() || NativeModules.FRNAvatarViewManager;
 
 export type NativeAvatarTokens = {
   /**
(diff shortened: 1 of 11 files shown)


packages/experimental/Avatar/src/NativeAvatar.tsx
```diff
@@ -26,7 +26,8 @@ interface ExportedConstants {
   sizes: { [key in Size]: number };
 }
 
-const ExportedNativeConstants: ExportedConstants = TurboModuleRegistry.get('FRNAvatarConstants') || NativeModules.FRNAvatarViewManager;
+const ExportedNativeConstants: ExportedConstants =
+  TurboModuleRegistry.get('FRNAvatarConstants')?.getConstants() || NativeModules.FRNAvatarViewManager;
```
PPatBoyd, line 30: Just curious; do you know what the ` || NativeModules.FRNAvatarViewManager` does for us here? Was it an older means for dealing with accessing the constants from turbo vs native modules?

PPatBoyd (approved):

packages/experimental/Avatar/src/NativeAvatar.tsx
Saadnajmi, line 30: I'll probably need to update the native constant-getting code here.

Saadnajmi: (Or just remove it and move the constants to JS)

Saadnajmi: Tested. Avatar just shows up as "Unimplemented Component".But also the size ramp seems to work so I think the constants are good. It doesn't crash so I'm happy. 

image

PR #3627, opened 2024-05-17
Move platform theming logic from button component into theme

acoates-ms:
Currently there is a lot of hidden platform checks in many of our components, which control the appearance outside of the theme.  This essentially means that our themes only work on their specific platform.

This change starts to move logic out of internal platform checks and into the theme objects.  Eventually this will allow themes to work better across platforms.  It also ensures that themes have maximum capabilities. Internally we will need this as we will want to use the win32-theme on the windows platform when we move to fabric.


Saadnajmi:
> Eventually this will allow themes to work better across platforms.

With some platform specific Platform Colors, I think that was explicitly not a goal when I touched this last. as an fyi

Saadnajmi:
Second comment. Moving files from platform specific to not platform specific will mean increased bundle size since metro won't filter them out. I'm not sure that's an issue anymore, but it was the reason we went for platform specific files for tokens everywhere.

acoates-ms:
> Second comment. Moving files from platform specific to not platform specific will mean increased bundle size since metro won't filter them out. I'm not sure that's an issue anymore, but it was the reason we went for platform specific files for tokens everywhere.

I dont think any production bundles are going to be including the android theme on win32 for example.  However, there will potentially be a larger bundle for users that are only using one component, which will now get a bundle including the theme for all components.

packages/theming/android-theme/src/components/Button/ButtonTheme.ts
```diff
@@ -0,0 +1,21 @@
...
+import { defaultButtonColorTokens } from './ButtonColorTokens';
+import { defaultButtonFontTokens } from './ButtonFontTokens';
+import { defaultButtonTokens } from './ButtonTokens';
+import { defaultToggleButtonColorTokens } from './ToggleButtonColorTokens';
+
+export const defaultButtonTheme = (theme: Theme) =>
+({
+  components: {
+    Button: immutableMerge<object>(defaultButtonColorTokens(theme), defaultButtonFontTokens(theme), defaultButtonTokens(theme)),
+    ToggleButton:
```
rurikoaraki, line 14: I believe FAB also applies to Android?

packages/theming/apple-theme/src/components/Button/FABTokens.ts
```diff
@@ -1,10 +1,7 @@
 import type { Theme } from '@fluentui-react-native/framework';
-import { globalTokens } from '@fluentui-react-native/theme-tokens';
-import type { TokenSettings } from '@fluentui-react-native/use-styling';
+import {default as globalTokens} from '@fluentui-react-native/theme-tokens/lib/tokens-global.ios';
 
-import type { FABTokens } from './FAB.types';
-
-export const defaultFABTokens: TokenSettings<FABTokens, Theme> = (t: Theme) =>
+export const defaultFABTokens = (t: Theme) =>
```
rurikoaraki, line 4: No FAB on macOS. Just iOS

packages/components/Button/src/ToggleButton/ToggleButton.tsx
```diff
@@ -27,9 +27,10 @@ export const ToggleButton = compose<ToggleButtonType>({
   useRender: (userProps: ToggleButtonProps, useSlots: UseSlots<ToggleButtonType>) => {
     const iconProps = createIconProps(userProps.icon);
     const toggleButton = useToggleButton(userProps);
+    const theme = useFluentTheme();
 
     // grab the styled slots
-    const Slots = useSlots(userProps, (layer) => buttonLookup(layer, toggleButton.state, userProps));
+    const Slots = useSlots(userProps, (layer) => buttonLookup(layer, toggleButton.state, userProps, theme));
```
rurikoaraki, line 33: Huh. I'm surprised this needed to be manually added.

packages/theming/default-theme/src/components/Button/ButtonTheme.ts
```diff
@@ -0,0 +1,34 @@
...
+import { defaultButtonTokens } from './ButtonTokens';
+import { defaultCompoundButtonColorTokens } from './CompoundButtonColorTokens';
+import { defaultCompoundButtonFontTokens } from './CompoundButtonFontTokens';
+import { defaultCompoundButtonTokens } from './CompoundButtonTokens';
+import { defaultToggleButtonColorTokens } from './ToggleButtonColorTokens';
+
+export const defaultButtonTheme = (theme: Theme) =>
+({
+  components: {
+    Button: immutableMerge<object>(
```
rurikoaraki, line 16: If you can, you should probably pull the names here from the exported buttonName, etc. from the component

packages/theming/default-theme/src/components/Button/ButtonTheme.ts
```diff
@@ -0,0 +1,34 @@
...
+import { defaultButtonFontTokens } from './ButtonFontTokens';
+import { defaultButtonTokens } from './ButtonTokens';
+import { defaultCompoundButtonColorTokens } from './CompoundButtonColorTokens';
+import { defaultCompoundButtonFontTokens } from './CompoundButtonFontTokens';
+import { defaultCompoundButtonTokens } from './CompoundButtonTokens';
+import { defaultToggleButtonColorTokens } from './ToggleButtonColorTokens';
+
+export const defaultButtonTheme = (theme: Theme) =>
+({
+  components: {
```
rurikoaraki, line 15: What's the thinking for the older version of the Button? Although it's marked as deprecated we haven;t been able to force everyone off of it yet.

acoates-ms: Do you think we have users outside of win32?
I had to move GDPR to the new button because the older version was relying on some layout behavior we have in win32 that isn't in the other RN platforms.  So, the older button would layout incorrectly when I switched to fabric.

So, I guess I'm suggesting forcing people to upgrade before they can move to fabric.

packages/theming/default-theme/src/components/Button/ButtonTheme.ts
acoates-ms: default-theme cannot depend on Button, because Button depends on framework which depends on default-theme. :(

packages/components/Button/src/ToggleButton/ToggleButton.tsx
acoates-ms: Surprised that it wasn't already there.. or surprised / questioning the need for the change?

PR #3648, opened 2024-06-07
Fix typo in `DirectionalHint` Callout type

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

There's a typo in the `DirectionalHint` type for Callout's `directionalHint` prop, where `"bottomLeftEdge"` is spelled as `"bottonLeftEdge"`. This PR fixes this so the linter stops spitting an incorrect error. 

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-callout-50213cd4-3583-401a-b8cc-d16cde4e24bc.json
+++ b/change/@fluentui-react-native-callout-50213cd4-3583-401a-b8cc-d16cde4e24bc.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix typo in `DirectionalHint` type",
+  "packageName": "@fluentui-react-native/callout",
+  "email": "winlarry@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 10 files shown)


PPatBoyd (dismissed):

amgleitman (dismissed):

Saadnajmi (changes requested):
You'll have to update this too: https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Callout/macos/FRNCalloutManager.m#L18-L33

Saadnajmi:
Which.. after that, we'll have to ensure that the change makes it internally and doesn't cause regressions for partners who don't update their bundles..

PPatBoyd:
> Which.. after that, we'll have to ensure that the change makes it internally and doesn't cause regressions for partners who don't update their bundles..

FWIW there are no current users of the enum value that I could find in any Office repos

PR #3658, opened 2024-06-18
fix(Apple): Render vector borders on Button

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Due to a [very obscure check](https://github.com/facebook/react-native/blob/d8739e1d37ceebb9643ca56d66a54aaca330665a/packages/react-native/React/Views/RCTView.m#L806-L813) inside of React Native's iOS (and macOS) implementation, borders on views are sometimes drawn as raster images. This means that when you zoom into a view (if it's a ScrollView for example), the corner radio don't scale well, and you can get blurriness. Luckily, the fix is to just add the style property `overflow: 'hidden'`. This seems like a safe/good thing to do for all platforms, so I didn't gate it behind a platform check

### Verification

Before:
Screenshot 2024-06-18 at 3 34 02 PM

After:
Screenshot 2024-06-18 at 3 34 21 PM

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Dewsk (approved):
Looks good - thanks for the fix!

packages/components/Button/src/Button.styling.ts
```diff
@@ -58,6 +58,7 @@ export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, Bu
       (tokens: ButtonTokens, theme: Theme) => ({
         style: {
           display: 'flex',
+          overflow: 'hidden',
```
nakambo, line 61: just for my own understanding -- if someone did want to revert this behavior, they still can using custom tokens, right?

nakambo (approved):

packages/components/Button/src/Button.styling.ts
Saadnajmi: Yeah, I think so. Or adding their own stiling on top via our slot stuff

PR #3661, opened 2024-06-20
Macos callout text run anchoring

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Enable anchoring to nested MacOS text runs

### Verification

Manually verified with new anchor targets on Callout test page

image

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


PPatBoyd:
Note: this PR will be broken until it can take a new version of RN-MacOS with dependent changes.  Making available for review and discussion ahead of that integration.

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -224,7 +383,7 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 	}
 
 	/// Calculates the NSRect of the anchorView in the coordinate space of the current screen
-	private func calculateAnchorViewScreenRect() -> NSRect {
+	private func calculateAnchorViewScreenRect(anchorView: NSView?, anchorBounds: NSRect) -> NSRect {
 		guard let anchorView = anchorView else {
 			preconditionFailure("No anchor view provided to position the Callout")
 		}
```
Saadnajmi, line 389: Feels weird to allow `anchorView` to be optional, and then fail if it isn't?

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -175,14 +227,121 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
+		// If we could find the bounding rect of our target view and it's a representative rect, return it as the anchor rect for the target
+		if !targetViewBounds.equalTo(zeroRect) {
+				return calculateAnchorViewScreenRect(anchorView: leafNSView, anchorBounds: targetViewBounds)
+		}
+
+		// Unfortunately our efforts could not determine a valid anchor rect for our target prop
+		return nil
+	}
+
+	/// Get the AnchorScreenRect to use for Callout anchoring, prioritizing the target prop is prioritized over th anchorRect prop
```
Saadnajmi, line 326:
```suggestion
	/// Get the AnchorScreenRect to use for Callout anchoring, prioritizing the target prop over the anchorRect prop
```

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -175,14 +227,121 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
+			return nil
+		}
+
+		// If we could find the bounding rect of our target view and it's a representative rect, return it as the anchor rect for the target
+		if !targetViewBounds.equalTo(zeroRect) {
+				return calculateAnchorViewScreenRect(anchorView: leafNSView, anchorBounds: targetViewBounds)
+		}
+
+		// Unfortunately our efforts could not determine a valid anchor rect for our target prop
+		return nil
```
Saadnajmi, line 323: Something feels bad about this.. but I'll have to think about it more and see if there's a more idiomatic way to do the same thing

PPatBoyd: Can split to separate functions at a minimum

PPatBoyd: I considered splitting the "TryGetBoundsAsNSBackedView" and "TryGetBoundsAsASubnodeOfALeafNode" portions into their own functions, it didn't make it a whole lot better.  I could be over-using guard conditions and stack a few `if let`s instead, but overall my dislikes seem mostly around the style and less in changing the general flow of the logic

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -174,15 +226,138 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
...
+
+		// Do not proceed if we don't have the NSView corresponding to the yoga leaf view
+		guard let leafView = uiManager.view(forReactTag: leafShadowView.reactTag) else {
+				return nil
+		}
+
+		// Do not proceed if somehow the leafView is not an RCTTextView
+		guard let textView = leafView as? RCTTextView else {
+				preconditionFailure("it should not be possible for the RCTTextShadowView to not be represented by an RCTTextView")
+		}
```
amgleitman, line 256: nit: We should be able to combine a number of these, at the very least the first and third ones. You could even chain it with the fourth one if you wanted unless you want to isolate that precondition failure case.

```swift
guard let leafView = bridge?.uiManager?.view(forReactTag: leafShadowView.reactTag) else {
    return nil
}
```

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -1,19 +1,71 @@
 import AppKit
 import Foundation
-#if USE_REACT_AS_MODULE
 import React
-#endif // USE_REACT_AS_MODULE
+
+/// Return the text length of a provided RCTShadowView
+func getLengthOfTextShadowNode(shadowView: RCTShadowView) -> Int {
```
amgleitman, line 6: nit: A number of these function names are a bit verbose. Swift tends to like to "interleave" a function's name with its arguments when it can, for example, [`convert(_ rect: NSRect, to view: NSView?)`](https://developer.apple.com/documentation/appkit/nsview/1483217-convert).

This is more about style than functionality; this should be totally functional as is.

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -224,16 +399,12 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 	}
 
 	/// Calculates the NSRect of the anchorView in the coordinate space of the current screen
-	private func calculateAnchorViewScreenRect() -> NSRect {
-		guard let anchorView = anchorView else {
-			preconditionFailure("No anchor view provided to position the Callout")
-		}
-
+	private func calculateAnchorViewScreenRect(anchorView: NSView, anchorBounds: NSRect) -> NSRect {
```
amgleitman, line 402: It's not immediately obvious why you'd want `anchorBounds` as a separate argument instead of just using `anchorView.bounds`. You should add a comment explaining why.

packages/components/Callout/macos/CalloutView.swift
```diff
@@ -370,7 +541,7 @@ class CalloutView: RCTView, CalloutWindowLifeCycleDelegate {
 	// MARK: Private variables
 
 	/// The view the Callout is presented from.
-	private var anchorView: NSView?
+	private var anchorReactTag: NSNumber?
```
amgleitman, line 544: We don't use this anywhere. Do we still need it?

packages/components/Callout/macos/CalloutView.swift
PPatBoyd: sought to clarify with a better variable name, per your recommendation of "Swiftier" function signatures :)

PR #3677, opened 2024-07-09
Export Shimmer consts

rurikoaraki:
Export constants for shimmer delay/duration/angle for use in other projects so that timings are consistent across usages


Saadnajmi (approved):
I guess we never really developed control tokens / a way to export them

PR #3688, opened 2024-07-22
Implement Focus Zone circular navigation and tab key behaviors for macOS

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Initial (protoype) implementation for `navigateAtEnd` and `tabKeyNavigation` for macOS implementation of focus zone.

### Verification

Manual testing performed:

* "FocusZone with Circular Navigation": all arrow keys (left, right, up, down), and tab (Tab, Shift+Tab)
* "FocusZone with Tab Key Abstraction": all keys for all of Normal, NavigateWrap, NavigateStopAtEnds, None

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover: n/a

### DIFF
--- a/apps/fluent-tester/src/TestComponents/FocusZone/styles.ts
+++ b/apps/fluent-tester/src/TestComponents/FocusZone/styles.ts
@@ -1,4 +1,4 @@
-import { StyleSheet, View } from 'react-native';
+import { Platform, StyleSheet, View } from 'react-native';
 
 import { Text } from '@fluentui/react-native';
 import { ButtonV1 as Button } from '@fluentui-react-native/button';
@@ -47,14 +47,16 @@ export const focusZoneTestStyles = StyleSheet.create({
   smallBoxStyle: {
     width: 20,
     height: 20,
-    backgroundColor: 'lightgrey',
     margin: 5,
+    // something is wrong with setting the background color on this style on the Mac, and until we fix that, let's not do this
+    ...Platform.select({ macos: { borderWidth: 1 }, default: { backgroundColor: 'lightgrey' } }),
   },
   wideBoxStyle: {
     width: 150,
     height: 20,
-    backgroundColor: 'lightgrey',
     margin: 5,
+    // something is wrong with setting the background color on this style on the Mac, and until we fix that, let's not do this
+    ...Platform.select({ macos: { borderWidth: 1 }, default: { backgroundColor: 'lightgrey' } }),
   },
 });
 
--- a/packages/components/FocusZone/macos/RCTFocusZone.m
+++ b/packages/components/FocusZone/macos/RCTFocusZone.m
@@ -308,6 +308,73 @@ static BOOL ShouldSkipFocusZone(NSView *view)
 	return [self nextViewToFocusForCondition:block];
 }
 
+- (NSView *)nextViewToFocusForCircularAction:(FocusZoneAction)action
+{
+	BOOL isAdvance = IsAdvanceWithinZoneAction(action);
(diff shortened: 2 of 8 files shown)


apps/fluent-tester/src/TestComponents/FocusZone/styles.ts
```diff
@@ -47,14 +47,16 @@ export const focusZoneTestStyles = StyleSheet.create({
   smallBoxStyle: {
     width: 20,
     height: 20,
-    backgroundColor: 'lightgrey',
     margin: 5,
+    // something is wrong with setting the background color on this style on the Mac, and until we fix that, let's not do this
+    ...Platform.select({ macos: { borderWidth: 1 }, default: { backgroundColor: 'lightgrey' } }),
```
Saadnajmi, line 52: Could you elaborate?

nakambo: image

This is all I see, there's nothing visible in the gray area above. However, the controls DO exist, they can be focused, and I bet you could hear them via VO.

packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -308,6 +308,73 @@ - (NSView *)nextViewToFocusForAction:(FocusZoneAction)action
...
+		? NSMakePoint(isAdvance ? 0 : [self bounds].size.width, NSMidY(firstResponderRect))
+		: NSMakePoint(NSMidX(firstResponderRect), isAdvance ? 0 : [self bounds].size.height);
+
+	__block CGFloat closestDistance = CGFLOAT_MAX;
+
+	IsViewLeadingCandidateForNextFocus block = ^BOOL(NSView *candidateView)
+	{
+		BOOL isLeadingCandidate = NO;
+		BOOL skip = NO;
+		NSRect candidateRect = [candidateView convertRect:[candidateView bounds] toView:self];
```
Saadnajmi, line 328: Was this code lifted from elsewhere, or is to new code? I didn't expect that we would need a new block and all here

nakambo: This was reworked off of the existing navigation code -- the conditions and checks are different -- for example to handle right arrow when we are already at the rightmost control (previous code to determine the control to focus returns nil in this scenario), we attempt to find the leftmost control in roughly the same "row". If we don't find any, then we look in the other (i.e. vertical) axis, again similar to existing navigation code.

It was written this way because this is how I thought about the algorithm in my head -- we're only going to try to wrap when we actually "run out of things". If we can't still find something in one axis, we go in the other axis. Conveniently it also seems less likely to break an existing scenario that doesn't use circular navigation written this way.

Saadnajmi (approved):

apps/fluent-tester/src/TestComponents/FocusZone/styles.ts
PPatBoyd: @Saadnajmi this is the issue with the Text backgroundColor support from https://github.com/microsoft/fluentui-react-native/blob/4d1c92173633ca23fc153890fda6472e732e9e50/apps/fluent-tester/src/TestComponents/FocusZone/FocusZoneTest.tsx#L131C1-L145C3
```
const TinyText = () => {
  return (
    <Text selectable style={focusZoneTestStyles.smallBoxStyle}>
      Hi
    </Text>
  );
};

const WideText = () => {
  return (
    <Text selectable style={focusZoneTestStyles.wideBoxStyle}>
      Hello world
    </Text>
  );
};
```

PR #3689, opened 2024-07-23
FocusZone navigationOrderInRenderOrder

PPatBoyd:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Adds a `navigationOrderInRenderOrder` property for MacOS keyboard focus management.  In some cases the auto-calculated key view loop may use the _center point_ of focusable controls to determine the focus order.  This property is available to make using the "render order" (the natural order of the accessible children as returned by `NSView.accessibilityChildren`) the order that should be used for navigation purposes.  Note that to participate in the key view loop process the FocusZone itself must be:
- `accessible`
- have an appropriate `accessibilityRole`

This PR also corrects the FocusZone prop type to incorporate the naturally relevant `ViewProps` since FocusZone is inherently a `View` underneath, and this eases maintenance and developer ergonomics.

### Verification

Manually tested.

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover


packages/components/FocusZone/macos/RCTFocusZone.m
```diff
@@ -186,6 +186,15 @@ static BOOL ShouldSkipFocusZone(NSView *view)
 	return NO;
 }
 
+/// Prevent order change
```
Saadnajmi, line 189: Coooulld we have a more descriptive comment here, especially if it's a triple slash, AKA, documentation? :)

packages/components/FocusZone/src/FocusZone.types.ts
```diff
@@ -7,7 +8,7 @@ export const focusZoneName = 'FocusZone';
 
 export interface FocusZoneState {}
 
-export type FocusZoneProps = React.PropsWithChildren<{
+export type FocusZoneProps = ViewProps & {
```
Saadnajmi, line 11: nit: VewProps from adapters

packages/components/FocusZone/macos/RCTFocusZone.m
amgleitman: As an aside, what assumptions can we make here?

Apple's documentation suggests (but doesn't seem to *explicitly* say) that [`accessibilityChildren`](https://developer.apple.com/documentation/appkit/nsaccessibility/1535018-accessibilitychildren) follows the view hierarchy. Is that good enough?

I don't know if we want to make other assumptions about [`accessibilityChildrenInNavigationOrder`](https://developer.apple.com/documentation/appkit/nsaccessibility/2869552-accessibilitychildreninnavigatio), where the docs only say "an easily navigable order," but for the latter I suppose we can lean back on "whatever the OS defaults to."

amgleitman (approved):

packages/components/FocusZone/macos/RCTFocusZone.m
amgleitman: New documentation comment looks good to me 👍

PPatBoyd: The simplicity of the implementation is by-design; I'm not seeking to define "render order" or "UI hierarchy order" any better than "the order of the UI elements as returned by `accessibilityChildren`, which is _an_ order corresponding with the render order and UI hierarchy but certainly not the only order possible.  It gives the RN author the opportunity to change the navigation order from the default calculated order in a predictable way but certainly lacks flexibility if they have a specific problem where neither order is their desired order.  

If someone else requires a different order, we can re-assess the API then with the concrete example; it isn't possible without Fabric for us to have a JSI-callback to allow order to be dynamically determined from JS, and there are flaws with many other possible designs.  If a given order is flawed based on say, the UI tree generated by native components changes the parent-child relationships of native children relative to the VDOM then either the use case is flawed (e.g. `Callout`) or there may be a platform bug worth addressing in the native component

Saadnajmi (approved):

packages/components/FocusZone/src/FocusZone.types.ts
```diff
@@ -1,14 +1,14 @@
 import type * as React from 'react';
-import type { ViewStyle } from 'react-native';
 
+import type { IViewProps } from '@fluentui-react-native/adapters';
```
amgleitman, line 3: This requires you to add `"@fluentui-react-native/adapters": "workspace:*"` to FocusZone's package.json. That should clear up the current CI failures.

amgleitman (changes requested):

amgleitman (approved):

Dewsk (approved):

PR #3692, opened 2024-07-26
[Mac] [Focus Zone] Restore broken direction = none functionality

nakambo:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

In a previous change, I incorrectly refactored some functionality that caused focus zone to respond to arrow keys even when direction was set to none. This change restores direction = none functionality.

### Verification

Manual testing:

* Disabled focus zone: tab works normally (as if no focus zone is present), arrow keys beep
* Arrow keys on direction = none zone: tab works like focus zone (move to outside the zone), arrow keys beep
* Arrow keys on direction = horizontal zone: tab works like focus zone (move to outside the zone), vertical arrow keys beep, horizontal arrow keys move in zone
* Arrow keys on direction = vertical zone: tab works like focus zone (move to outside the zone), horizontal arrow keys beep, vertical arrow keys move in zone
* Arrow keys on direction = bidirectional zone: tab works like focus zone (move to outside the zone), arrow keys move in zone

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover: n/a

### DIFF
--- a/change/@fluentui-react-native-focus-zone-1e63be00-70e8-4c8d-88be-e75acdcdaf39.json
+++ b/change/@fluentui-react-native-focus-zone-1e63be00-70e8-4c8d-88be-e75acdcdaf39.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "[Mac] [Focus Zone] Restore broken direction = none functionality",
+  "packageName": "@fluentui-react-native/focus-zone",
+  "email": "nakambo@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):

PPatBoyd:
`[non-blocking]` relevant but not for this PR: `FocusZone.types.ts` doesn't note the default `FocusZoneDirection` which probably _should_ be Direction=None but defaults to Bidirectional due to how the native enum is defined.

which would be a breaking change and need to be minded for uptake by consumers to adjust accordingly... woo...

PR #3728, opened 2024-08-22
Bump WebDriverIO

samuelfreiberg:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

A significant bug was fixed in 8.40.4. This PR updates to pull that in.

Previously, after test execution, there was an Appium process that lingered, that causes the next run to fail. Additionally, the runs throw an error with Appium exiting early.

All are fixed in this update

### Verification

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


Saadnajmi (approved):

apps/E2E/package.json
```diff
@@ -39,14 +39,14 @@
...
-    "@wdio/spec-reporter": "8.32.4",
-    "appium": "^2.5.1",
+    "@wdio/appium-service": "^8.40.0",
+    "@wdio/cli": "^8.40.0",
+    "@wdio/globals": "^8.40.0",
+    "@wdio/jasmine-framework": "^8.40.0",
+    "@wdio/json-reporter": "^8.40.0",
+    "@wdio/local-runner": "^8.40.0",
+    "@wdio/spec-reporter": "^8.40.0",
+    "appium": "^2.11.2",
```
Saadnajmi, line 49: Side note: I wonder if we can publish a dependency profile like we do with the rest of FURN for the stable sets of Webdriver deps?

samuelfreiberg: That's a good callout - right now, v8 is stable (they're already releasing v9 soon). Once v9 comes out, I'm sure we can do this.

PR #3744, opened 2024-09-10
Bringing windows platform more in line with more complete win32 impl

acoates-ms:
This change introduces our first .officewin file.  The idea of the .officewin extension is to allow us to continue to use our theme fragments that are picked up by the bundler, but allow two themes for the windows platform.  By default the windows platform should provide a windows fluent theme.  But for office we want the default theme to have various officeisms.  We will do that by having custom resolution logic when building against the windows platform within office.  (Office will resolve *.officewin, *.windows, *.native for FURN packages when building against the windows platform.

Currently the only difference between the Windows theme and Office's theme for the .windows platform is that we override the default button size to be small within Office.  But I expect that to expand as we dig into the details of the theming differences between Office and Windows.

Mostly this change unifies some theming of Button/Checkbox by sharing definitions of various parts of the theme between win32 and windows (by moving the win32 implementation to windows).  The win32 definitions are more complete and more tested than the current windows ones.  Over time as we work out that parts of those themes are more Office specific, we could move those parts of the definitions to .officewin files.


Dewsk:
@rurikoaraki and @PPatBoyd - any concerns here? Will review as well but this one should have a couple sets of eyes.

rurikoaraki:
> @rurikoaraki and @PPatBoyd - any concerns here? Will review as well but this one should have a couple sets of eyes.

Will make time to review today

packages/components/Button/src/ToggleButton/ToggleButtonColorTokens.windows.ts
```diff
@@ -33,7 +33,6 @@ export const defaultToggleButtonColorTokens: TokenSettings<ToggleButtonTokens, T
 const highContrastColors = {
   checked: {
     backgroundColor: PlatformColor('SystemColorHighlightColor'),
-    borderColor: PlatformColor('SystemColorHighlightColor'),
```
rurikoaraki, line 36: Was this change to the border intended?

packages/components/Button/src/useButton.ts
```diff
@@ -76,7 +76,8 @@ export const useButton = (props: ButtonProps): ButtonInfo => {
   const hasTogglePattern = props.accessibilityActions && !!props.accessibilityActions.find((action) => action.name === 'Toggle');
 
   const theme = useFluentTheme();
-  const shouldUseTwoToneFocusBorder = Platform.OS === ('win32' as any) && props.appearance === 'primary' && !isHighContrast(theme);
+  const shouldUseTwoToneFocusBorder =
```
rurikoaraki, line 79: > const shouldUseTwoToneFocusBorder = [](http://example.com/codeflow?start=1&length=36)

At the moment, this is an Officeism. Potentially this will come later to windows but I don't think the windows platform does this today.

packages/components/Button/src/CompoundButton/CompoundButtonTokens.win.ts
```diff
@@ -0,0 +1,166 @@
+import type { Theme } from '@fluentui-react-native/framework';
+import { globalTokens } from '@fluentui-react-native/theme-tokens';
+import { isHighContrast } from '@fluentui-react-native/theming-utils';
+import type { TokenSettings } from '@fluentui-react-native/use-styling';
+
+import type { CompoundButtonTokens } from './CompoundButton.types';
+
+export const defaultCompoundButtonTokens: TokenSettings<CompoundButtonTokens, Theme> = (theme: Theme): CompoundButtonTokens => ({
+  medium: {
+    padding: globalTokens.size120 - globalTokens.stroke.width10,
```
rurikoaraki, line 10: >globalTokens.size120 - globalTokens.stroke.width10, [](http://example.com/codeflow?start=13&length=51)

I wonder if this will be needed on windows... we end up having to subtract the width of the border of the button to figure out the padding on win32 because when Views are bound by minHeight/Width instead of just height/width, the border adds to the height of the button so it ends up being bigger than intended (and sometimes causing size changes on focus). This (seemingly?) doesn't happen on web, so I wonder if windows has this problem or not

rurikoaraki: (and we need minHeight for handling text scaling, and minWidth for just who knows how long strings are)

rurikoaraki (approved):
:shipit:

PR #3812, opened 2024-11-12
Enable inline links on macOS to respond to interaction events

rurikoaraki:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Use View wrapping around links in macOS for inline links so that they can respond to interaction events.

### DIFF
--- a/packages/components/Link/src/Link.tsx
+++ b/packages/components/Link/src/Link.tsx
@@ -43,6 +43,9 @@ export const Link = compose<LinkType>({
       // This is a workaround for the issue. Once those issues are resolved, supportsA11yTextInText can be removed.
       const supportsA11yTextInText = Platform.OS !== 'android';
 
+      // MacOS Text component doesn't handle interaction events like hover etc.
+      // which are needed to style links correctly. Since macOS can handle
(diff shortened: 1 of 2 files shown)


packages/components/Link/src/Link.tsx
```diff
@@ -43,7 +43,12 @@ export const Link = compose<LinkType>({
       // This is a workaround for the issue. Once those issues are resolved, supportsA11yTextInText can be removed.
       const supportsA11yTextInText = Platform.OS !== 'android';
 
-      return supportsA11yTextInText && (inline || mergedProps.selectable) ? (
+      // MacOS Text component doesn't handle interaction events like hover etc.
```
Saadnajmi, line 46: I thought it did? @amgleitman

Saadnajmi (approved):

packages/components/Link/src/Link.tsx
rurikoaraki: It's possible that this was fixed later but this is what we have in our downstream code right now.

PR #3841, opened 2025-01-24
Implement TabList CTRL + TAB keyboard shortcut on win32.

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Native Windows TabList / Pivot implement a CTRL + TAB keyboard shortcut to switch tab selection for the entire tablist. This keyboard shortcut was missing on the FURN TabList, and this PR implements this. 

### Verification

Tested locally in FURN app. Verified narration and announcements do not break - no repro of old double narration bug. 

https://github.com/user-attachments/assets/c7705bc9-4f9c-4832-920f-a60f8008a2a8

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [x] Voiceover

### DIFF
--- a/packages/components/TabList/src/TabList/useTabList.ts
+++ b/packages/components/TabList/src/TabList/useTabList.ts
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import type { View, AccessibilityState, LayoutRectangle } from 'react-native';
+import { type View, type AccessibilityState, type LayoutRectangle, Platform } from 'react-native';
 
 import { memoize, mergeStyles } from '@fluentui-react-native/framework';
 import type { LayoutEvent } from '@fluentui-react-native/interactive-hooks';
 import { useSelectedKey } from '@fluentui-react-native/interactive-hooks';
+import type { IKeyboardEvent } from '@office-iss/react-native-win32';
 
 import type { TabListInfo, TabListProps } from './TabList.types';
 import type { AnimatedIndicatorStyles } from '../TabListAnimatedIndicator/TabListAnimatedIndicator.types';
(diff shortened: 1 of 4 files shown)


packages/components/TabList/src/TabList/useTabList.ts
```diff
@@ -72,6 +73,49 @@ export const useTabList = (props: TabListProps): TabListInfo => {
...
+    (increment: number) => {
+      if (increment === 0) {
+        return;
+      }
+
+      const currentIndex = tabKeys.indexOf(selectedTabKey);
+
+      if (currentIndex === -1 && __DEV__) {
+        console.error(`Selected tab key does not exist in tab key array. Tab Key: ${selectedTabKey}. Keys: ${tabKeys}`);
+      }
```
Saadnajmi, line 86: Should we keep this, or just remove it?

packages/components/TabList/src/TabList/useTabList.ts
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import type { View, AccessibilityState, LayoutRectangle } from 'react-native';
+import { type View, type AccessibilityState, type LayoutRectangle, Platform } from 'react-native';
```
rurikoaraki, line 2: I think we prefer these to be separate import lines (import type and import) for tree shaking purposes

packages/components/TabList/src/TabList/useTabList.ts
rurikoaraki: i would say that if the API/function isn't exposed externally, it's a bit overkill to check for weird inputs since you should be able to know exactly what is coming... but not a strong opinion.

packages/components/TabList/src/TabList/useTabList.ts
```diff
@@ -72,6 +73,49 @@ export const useTabList = (props: TabListProps): TabListInfo => {
...
+
+      const currentIndex = tabKeys.indexOf(selectedTabKey);
+
+      if (currentIndex === -1 && __DEV__) {
+        console.error(`Selected tab key does not exist in tab key array. Tab Key: ${selectedTabKey}. Keys: ${tabKeys}`);
+      }
+
+      // We want to only switch selection to non-disabled tabs. This skips over disabled ones.
+      const direction = increment > 0 ? 1 : -1;
+      increment = increment / direction; // abs
```
rurikoaraki, line 90: can increment be its own variable so that we're not reusing the param passed in?

rurikoaraki: This function seems over-complicated by trying handle any amount of increment. My understanding is that the only use case right now is 1 or -1, so in order for simplicity I think it would be best to start with just handling those two values.

packages/components/TabList/src/TabList/useTabList.ts
lawrencewin: Went ahead and removed the log.

packages/components/TabList/src/TabList/useTabList.ts
lawrencewin: No that's fair criticism. Method has been changed to be simpler.

rurikoaraki (approved):

PR #3845, opened 2025-01-28
Potential fix for code scanning alert no. 3: Prototype-polluting assignment

rurikoaraki:
Potential fix for [https://github.com/microsoft/fluentui-react-native/security/code-scanning/3](https://github.com/microsoft/fluentui-react-native/security/code-scanning/3)

To fix the prototype pollution issue, we need to ensure that the `key` value used in the `ensureAndReturn` function is validated and does not include dangerous property names like `__proto__`, `constructor`, or `prototype`. We can achieve this by adding a check to reject these keys before performing the assignment.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

From Ruriko: I have checked the code and it's in line with the suggestions attached to the CodeQL alert for how to fix the issue.


rurikoaraki:
Doesn't seem like fix is required - build says that it's not possible, and looking at the code i'm inclined to agree...

Saadnajmi (approved):

PR #3848, opened 2025-01-29
chore: remove enzyme tests

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Enzyme has been deprecated / not compatible with the latest React versions, but we still have many tests running using it. As far as I can tell, there is no replacement for the types of tests we're running (`checkRenderConsistency` and `checkReRender`), since the replacement (react-native-testing-library) is more of an integration test library. Let's remove the enzyme tests altogether, and port the ones that are simply snapshot tests to use `react-test-renderer`.

Note, `react-test-renderer` (used for all our snapshot tests) is also deprecated, but it looks like [react-native-testing-library v14 will have a replacement](https://github.com/callstack/react-native-testing-library/pull/1705). So I'll hold off migrating those snapshot tests till we need to. 

### Verification

CI should pass

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover


rurikoaraki (approved):

packages/framework/framework/tsconfig.json
```diff
@@ -1,23 +1,8 @@
 {
+  "extends": "@fluentui-react-native/scripts/tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "target": "es5",
```
rurikoaraki, line 4: target values are different, not sure if that's concerning to you. seems like aligning es versions makes sense though

packages/framework/framework/tsconfig.json
```diff
@@ -1,23 +1,8 @@
...
-    "jsx": "react",
     "outDir": "lib",
-    "declaration": true,
-    "sourceMap": true,
-    "experimentalDecorators": true,
-    "importHelpers": true,
-    "strictNullChecks": true,
-    "noImplicitAny": true,
```
rurikoaraki, line 12: looks like several values are changing from what they were before. I guess any issues should show up in trying to build?

Saadnajmi: That was my thought, but I can do a smaller sublet of changes. There's a whole lot more "combine tsconfigs" I can do in a separate PR

rurikoaraki: quick glance tells me most of these are changing to be less strict settings so this particular change should be ok i think

Saadnajmi: Oh. Less strict not good. I undid the ones so they stay overrides.

rurikoaraki (approved):

PR #3859, opened 2025-02-20
[LinkV1] Simplify code to find child string of LinkV1 when setting a11yLabel

ksiler:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Previously, if we didn't find a string as a child of LinkV1 and called .toString(), we ran into an error. This fixes it by simplyfing the code to just return the first child that's a string (if it exists) instead of calling .toString() on the result. We use this to set the a11yLabel if one isn't defined.

### Verification

Tested by creating a Link without any children. Verified FURN test app no longer runs into an error.

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/change/@fluentui-react-native-link-c18251d2-05a6-485f-a6bc-f471efef105e.json
+++ b/change/@fluentui-react-native-link-c18251d2-05a6-485f-a6bc-f471efef105e.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Check if we find a string as a child before calling .toString()",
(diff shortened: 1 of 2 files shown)


rurikoaraki (approved):

rurikoaraki (approved):

ksiler:
Abandoning this PR in favor of a new one (for some reason the PR pipeline isn't running the latest updates).

PR #3904, opened 2025-06-16
Fix peer dep warnings when running yarn

rurikoaraki:
Fixing the last peer dep warnings that appear when running yarn:
- Update version of use-subscription
- Change our react-dom version to 18.2.0 to align with react version of RN. We do this via yarn.lock so that we can look into changing the depdencies of FURN to allow multiple RN versions, at which point we couldnt restrict react that tightly


Saadnajmi (approved):

Saadnajmi:
Congrats!

PR #3918, opened 2025-06-24
ci: link lockfile

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [ ] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

Similar to https://github.com/microsoft/react-native-test-app/pull/2496 , run `@rnx-kit/lint-lockfile` on CI.

While touching this, I realized we were running `align-deps` (a repo-wide task run once) as part of `depcheck` (a task we run with lage on every leaf project). I don't like that, so I separated the two so that repo-wide single tasks (`align-deps` and `lint-lockfile`) are not run as part of lage and run explicitly as separate steps. I would prefer them to run with Lage, but I don't know how to configure the pipeline for these repo-wide single tasks. 

### Verification

CI should pass.

### Pull request checklist

- [x] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "check-for-changed-files": "cd scripts && yarn fluentui-scripts checkForModifiedFiles",
     "checkchange": "beachball check --changehint \"Run 'yarn change' to generate a change file\"",
     "align-deps": "rnx-align-deps --no-unmanaged --requirements react-native@0.74",
-    "depcheck": "yarn align-deps && lage depcheck",
+    "depcheck": "lage depcheck",
     "lint": "lage lint",
     "preinstall": "node ./scripts/use-yarn-please.js",
     "prettier": "lage prettier",
@@ -40,6 +40,7 @@
     "@react-native/metro-babel-transformer": "^0.74.0",
     "@react-native/metro-config": "^0.74.0",
     "@rnx-kit/align-deps": "^3.0.0",
+    "@rnx-kit/lint-lockfile": "^0.1.0",
     "babel-jest": "^29.7.0",
     "beachball": "^2.20.0",
     "eslint": "^8.0.0",
@@ -76,6 +77,16 @@
       "presets": [
         "microsoft/react-native"
       ],
+      "lint": {
+        "lockfile": {
+          "noDuplicates": {
+            "packages": [
+              "#react-native",
+              "@babel/core"
+            ]
+          }
+        }
+      },
       "requirements": {
         "development": [
           "react-native@0.74"
--- a/yarn.lock
+++ b/yarn.lock
@@ -4114,6 +4114,7 @@ __metadata:
     "@react-native/metro-babel-transformer": "npm:^0.74.0"
     "@react-native/metro-config": "npm:^0.74.0"
     "@rnx-kit/align-deps": "npm:^3.0.0"
+    "@rnx-kit/lint-lockfile": "npm:^0.1.0"
     babel-jest: "npm:^29.7.0"
     beachball: "npm:^2.20.0"
     eslint: "npm:^8.0.0"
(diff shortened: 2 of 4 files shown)


yarn.lock
```diff
@@ -6849,6 +6876,19 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@rnx-kit/tools-workspaces@npm:^0.2.3":
```
Saadnajmi: This is a duplicate.. so is the lint not working?

Saadnajmi: No, because we're only configured to ignore duplicates of babel and react-native.

Saadnajmi:
I looked into enabling root-level targets with lage, and I need to do a couple of things:

1. Add the root as a workspace because of https://github.com/microsoft/lage/issues/862 (Feels like lage should do this automatically?)
2. Update our lage config so we have something like `['build-tools']: ['^build-tools', '@fluentui-react-native/root#align-deps', '@fluentui-react-native/root#link-lockfile'],`
3. Add inputs to both tasks so they can be used with a cache, because of https://github.com/microsoft/lage/blob/681710b4af96431c5245237ff7c9a50e56ce2fb9/packages/hasher/src/TargetHasher.ts#L173
  - For lint-lock file, the input is just the lock file.
  - for align-deps, it would be every package.json.. so we will need to glob that?


I think I'd be willing to do those 3, but feels like much?

tido64:
> 1. Add the root as a workspace because of [Cannot reference root package tasks? lage#862](https://github.com/microsoft/lage/issues/862) (Feels like lage should do this automatically?)

This is expected, even if a bit unintuitive. I think it's the same with other build orchestrators, at least with Nx.

> 2. Update our lage config so we have something like `['build-tools']: ['^build-tools', '@fluentui-react-native/root#align-deps', '@fluentui-react-native/root#link-lockfile'],`

Wouldn't this be sufficient if you add root as a workspace? `['build-tools']: ['^build-tools', 'align-deps', 'link-lockfile'],`

> 3. Add inputs to both tasks so they can be used with a cache, because of https://github.com/microsoft/lage/blob/681710b4af96431c5245237ff7c9a50e56ce2fb9/packages/hasher/src/TargetHasher.ts#L173

If `build-tools` is cached, its tasks are no longer run. I don't think we need to make any changes for this.

Saadnajmi:
> > 1. Add the root as a workspace because of [Cannot reference root package tasks? lage#862](https://github.com/microsoft/lage/issues/862) (Feels like lage should do this automatically?)
> 
> This is expected, even if a bit unintuitive. I think it's the same with other build orchestrators, at least with Nx.
> 
> > 2. Update our lage config so we have something like `['build-tools']: ['^build-tools', '@fluentui-react-native/root#align-deps', '@fluentui-react-native/root#link-lockfile'],`
> 
> Wouldn't this be sufficient if you add root as a workspace? `['build-tools']: ['^build-tools', 'align-deps', 'link-lockfile'],`
> 
> > 3. Add inputs to both tasks so they can be used with a cache, because of https://github.com/microsoft/lage/blob/681710b4af96431c5245237ff7c9a50e56ce2fb9/packages/hasher/src/TargetHasher.ts#L173
> 
> If `build-tools` is cached, its tasks are no longer run. I don't think we need to make any changes for this.

Once you add root as a workspace, any task run at root (including "build") must have inputs. I played around with it and git it half working but the build fails. Rather than debug, I think I'll just merge this change as is.

package.json
```diff
@@ -76,6 +77,16 @@
       "presets": [
         "microsoft/react-native"
       ],
+      "lint": {
+        "lockfile": {
+          "noDuplicates": {
+            "packages": [
+              "#react-native",
```
rurikoaraki, line 84: for knowledge what's "#" mean?

rurikoaraki (approved):

package.json
Saadnajmi: I think it's custom syntax, and similar to an align-deps profile. It was introduced here: https://github.com/microsoft/rnx-kit/pull/3652/files

Saadnajmi: It looks like because this package is still the rnx-kit "incubator" folder, it does not have it's README published to the docs website.

PR #3923, opened 2025-07-02
Update FURN builds to use node16 module resolution and export maps

JasonVMo:
### Platforms Impacted
- all

#### Additional Updates 7/3

Some more changes to fix issues with the bundles.

- QOL improvement added, changed vscode defaults to automatically fix the import ordering and other eslint errors on save.
- Changed reference to e2e package to be via package rather than direct references. The direct references were causing the internal paths to be extremely deep. More importantly it caused the lib depth to diverge from the src depth causing asset references to fail.
- Exported things specifically in e2e. Note generating e2e/src/index.ts was done with AI, extremely useful to automate the tediousness of manually exporting everything.

The way TS was set up in fluent-tester, reaching across to the e2e package, caused TS to create a lib structure with lib/e2e/src and lib/fluent-tester/src within the fluent-tester package. It's causes wonky behavior, adding as a reminder to folks to not do this.

#### Original Description

To update FURN to use the modern jsx runtime it needs to use more modern settings, in particular export maps, which require using node16. This caused a bit of a cascade:

- Updated the base tsconfig.json to use node16 for module and moduleResolution, this will build commonjs output unless type="module" (which we don't do)
- We still output ESM -> lib and CJS -> lib-commonjs, now with node16 settings.
- Updated target to something more modern, babel transforms can handle it as needed.
- Got rid of the OnPublish stuff, it was done originally to handle jest typing better but updates since, in particular the @rnx-kit/jest-config knows how to handle things correctly.
- Added export maps for all packages and updated main and module to reference the .js files.
- Found/fixed a number of typing issues being hidden by our previous settings. In particular the use of TS directly caused unexported types to not always be caught.
- Removed an unused tsconfig.json at the root.

### Verification

Normal build and pipeline runs.


rurikoaraki (approved):

packages/utils/tokens/package.json
```diff
@@ -2,13 +2,16 @@
   "name": "@fluentui-react-native/tokens",
   "version": "0.23.1",
   "description": "Token interface parts and helpers",
-  "main": "src/index.ts",
-  "module": "src/index.ts",
-  "typings": "lib/index.d.ts",
-  "onPublish": {
```
Saadnajmi, line 8: We had special logic in our scripts to handle this key, is it not there anymore?

packages/theming/win32-theme/package.json
```diff
@@ -7,13 +7,16 @@
...
-    "main": "lib-commonjs/index.js",
-    "module": "lib/index.js"
+  "main": "lib-commonjs/index.js",
+  "module": "lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib-commonjs/index.js",
+      "types": "./lib/index.d.ts"
+    }
```
Saadnajmi, line 17: Does the version of metro we use internally support this? I'm not 100% sure.

packages/utils/tokens/package.json
JasonVMo: It was a beachball key to swap out package.json values during the publish process. It is a blind replacement and the code to do that still exists, but it is no longer leveraged at the moment. So we still have the capability.

packages/theming/win32-theme/package.json
Saadnajmi: Talked offline

Saadnajmi (approved):

PR #3924, opened 2025-07-03
Update yarn to the latest stable version and add dynamic-extensions plugin

JasonVMo:
### Platforms Impacted

- build only change

### Description of changes

This is another engineering system change. I updated yarn's version to 4.9.2 which is the latest stable version. It has some performance fixes that Danny put in and has the latest features.

I also added the @rnx-kit/yarn-dynamic-extensions plugin which can push dependencies to all the packages in the repo. See `dynamic.extensions.mjs` for the starting configuration.

Note that while this plugin will add things as dependencies, that simply ensures the node_modules are configured and the packages are available within the package scope. Because they are not added to the package.json itself it doesn't actually cause a change in the dependency graph, it is simply a way of adding/updating dev dependencies without having to change every package.json in the project.

### DIFF
--- a/tester_deps/.yarnrc.yml
+++ b/tester_deps/.yarnrc.yml
@@ -1,1 +1,1 @@
-yarnPath: ../.yarn/releases/yarn-4.6.0.cjs
+yarnPath: ../.yarn/releases/yarn-4.9.2.cjs


Saadnajmi:
Could we separate the yarn update from the dynamic dependency update? The latter looks like it'd have more effect, and I'm not sure what it's for from just this PR, but the yarn update we should probably just take.

JasonVMo:
> Could we separate the yarn update from the dynamic dependency update? The latter looks like it'd have more effect, and I'm not sure what it's for from just this PR, but the yarn update we should probably just take.

As per the discussion we had just now, the yarn bump + plugin gives me an escape valve to push common dev dependencies to the packages without having to create a lot of churn, which was what motivated this change. This is also something we are using in rnx-kit, and something whos usage is planned in OMR. The depcheck pass will ensure dependencies are directly declared.

One other note. Because eslint is not included in the devDependencies (the version is being picked up from just) it is not actually being applied to the packages. So the impact of the change is to make typescript and just-scripts available in the packages in the repo.

rurikoaraki:
Btw the tester_deps package will need a similar change

.yarn/plugins/@rnx-kit/yarn-plugin-dynamic-extensions.cjs
```diff
@@ -0,0 +1,96 @@
+// @ts-check
+
```
JasonVMo, line 2: This file should be ignored except for informational purposes. It is the bundle produced by rnx-kit.

rurikoaraki (approved):

PR #3926, opened 2025-07-09
Update the build command to use tsc directly, pick targets based off package.json, and integrate codegen

JasonVMo:
### Platforms Impacted
- all platforms

### Description of changes

More engineering system changes and cleanup that I encountered while working on the jsx-runtime work and separated into a more focused PR. This includes a number of changes to the way we build TS in FURN.

This includes the following changes:

- build typescript by invoking tsc directly rather than going through justScripts
- cleaned up the typescript options so we are using the tsconfig.json to configure almost everything
- typescript builds now pick targets based on package.json settings. It uses main/exports.require and module/exports.import to determine whether to build cjs / esm output and where to put that output.
- did a more comprehensive implementation of external.dependencies.mjs to allow for configuring groups of dependencies to inject in a more extendable way.
- cleaned up a few package.json entries with things that were wrong.
- made the babel config more configurable (though the configs remain the same for all packages). The babel settings, in particular the runtime setting, is one of the handlers that determines whether or not to use the new or old jsx runtime behavior.
- now run the codegen step for both output targets. Previously we were only doing codegen for --outDir lib

### Verification

Ran builds, tests, bundles, win32 tester.


scripts/src/utils/codegenNativeComponents.js
```diff
@@ -0,0 +1,60 @@
+// @ts-check
+const glob = require('glob');
+
+const fs = require('fs');
+const path = require('path');
```
Saadnajmi, line 5: Could we prefix these with `node:`?

Saadnajmi: Sidenote.. I think we can put a codegen config into our package.json so that end apps who consume us run codegen instead of us doing so (see RN 0.79 release notes I think?). I'm not sure how far back that goes.

JasonVMo: Yeah, I think there is a larger followup on the codegen side of things. I'm not sure I understand the purpose as it doesn't do the native boilerplate code that react-native codegen does.

Issue #1596, opened 2022-04-14
SVGTest testing SVG from local network fails


rurikoaraki:
### Willing to submit a PR to fix?

- [ ] I am willing to submit a PR to fix

### Requested priority

Low

### Products/applications affected

Tester is affected.

### Package version(s)

@fluentui/react-native  v0.28.10

### OS version(s)

Win32

### Platform

- [ ] iOS
- [ ] macOS
- [X] win32
- [ ] windows
- [ ] android

### Please provide a reproduction of the bug

Undo changes in #1591
Open SVG test page in tester
Wait 2 minutes for assert

### Actual behavior

Assert pops up

### Expected behavior

No assert

JasonVMo:
The test that was disabled would have violated some security rules if it were enabled.

PR #3932, opened 2025-07-11
Fix linting: finish the conversion to v9 and flat configs, make lint not silently fail and fix a bunch of type errors

JasonVMo:
### Platforms Impacted
- all, changes should be type only in practice

### Description of changes

I realized lint runs were going too fast...when I realized it was silently hiding errors. This led to the realization that the switch to v9 and the new flat config went a bit too easily. So this has a series of fixes of note.

- changed to explicitly running eslint in each folder rather than using the just indirection (which was hiding things)
- included the necessary dependencies in the framework/eslint-config-rules package which defines the lint rules for FURN
- updated lint rules to format them correctly
- increased lint strictness by default, then disabled rules we had to
- removed the repo root config which wasn't actually used by eslint
- changed the vscode setting for linting to point to the one that we actually use, now vscode should reflect the correct rules.

After that it was fixing a number of linting issues:

- generally places where we have type Foo<T extends object> it should actually be extending a type defined to be {}. This was the intent but it was a rookie error on my part.
- included handling for null and made the typing more correct in immutable merge.
- fixed a lot of odds and ends, the rules that are explitly turned back on we should look at, particularly the exhaustive-deps one the the PlatformColor references.

### Verification

This should be an eng-sys + typing change so builds are the main validation mechanism.


apps/fluent-tester/src/TestComponents/Svg/SvgTest.tsx
```diff
@@ -248,6 +248,7 @@ const RemoteSvgTest: React.FunctionComponent = () => {
           style={styles.svg}
           width="100"
           height="100"
+          // eslint-disable-next-line @microsoft/sdl/no-insecure-url
```
rurikoaraki: For adding new suppressions, could you add a comment mentioning why it's ok to bypass?

packages/framework/immutable-merge/src/Merge.ts
```diff
@@ -117,6 +141,24 @@ function getHandlerForPropertyOfType(
...
+ * Assign properties of source objects to a new target object. This is just a type wrapper around Object.assign
+ * @param objs - array of objects to merge
+ * @returns the result of object assign on the objects, typed to T
+ */
+function assignToNewObject<T extends ObjectBase>(...objs: T[]): T {
+  return Object.assign({}, ...objs);
+}
+
+/**
+ * Filter a set of unknown values to only include those that extend ObjectBase, this actually
```
rurikoaraki: was this comment meant to have more text?

packages/theming/theme-tokens/src/highContrast/tokens-alias.ts
```diff
@@ -10,6 +10,7 @@ function processAliasTokens(aliasTokens: any) {
       const entry = aliasTokens[key][innerKey];
       if (typeof entry === 'string' && entry.includes('PlatformColor')) {
         const color = 'SystemColor' + entry.substring(14, entry.length - 1) + 'Color';
+        // eslint-disable-next-line @react-native/platform-colors
```
rurikoaraki, line 13: ...sorry why are platform colors a problem?

JasonVMo: The new lint error is giving a message that PlatformColor has to take a static value, likely because it will be replaced as part of bundling in some way. So:
- PlatformColor('colorName') is valid
- PlatformColor(myVariable) won't work

packages/framework/immutable-merge/src/Merge.ts
JasonVMo: I think I was going to explain more about type inferencing but then decided that wasn't necessary. I'll fix it.

apps/fluent-tester/src/TestComponents/Svg/SvgTest.tsx
JasonVMo: Actually this one I wanted to talk about. I'm not as familiar with the SvgTest but it feels like this should be removed, even for non-shipping code.

Saadnajmi: It's more a test of react native svg.. we can delete I think

JasonVMo: Yes, I just deleted this code. It was disabled already so removed this will address issue #1596 (or make it no longer relevant)

PR #3933, opened 2025-07-11
[MenuItem][MacOS] Fix "onClick" being called twice when invoking menu item via ENTER / SPACE

lawrencewin:
### Platforms Impacted
- [ ] iOS
- [x] macOS
- [ ] win32 (Office)
- [ ] windows
- [ ] android

### Description of changes

React Native MacOS pressables handle a special case for ENTER / SPACE keys for the `keyDown` / `keyDown` event handlers: these keys should also be interpreted as presses, thus causing the event to be passed to the `onPress` handler. The code for this is found here: https://github.com/microsoft/react-native-macos/blob/48e607706bfcc25b4c28eba1aa352f409c02975e/packages/react-native/Libraries/Pressability/Pressability.js#L580-L617

For the `MenuItem`, we pass `onInvoke`, the handler for clicks and keypresses, into both `press` and `keyDown` / `keyUp` handlers. The code snippet above, then causes `onClick` to be called twice on the platform. We can fix this by simply marking the event as handled, so `onClick` is only called once. 

### Verification

Tested locally, video: 

### Pull request checklist

- [ ] Automated Tests
- [x] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuItem/useMenuItem.ts
+++ b/packages/components/Menu/src/MenuItem/useMenuItem.ts
@@ -15,6 +15,8 @@ import { useMenuItemTracking } from '../MenuList/useMenuList';
 export const triggerKeys = [' ', 'Enter'];
 export const submenuTriggerKeys = ['ArrowLeft', 'ArrowRight', ...triggerKeys];
 
+const preventDuplicateKeyDownHandling = Platform.OS === 'macos';
+
 export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
   // attach the pressable state handlers
   const defaultComponentRef = React.useRef(null);
@@ -38,6 +40,12 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
       if (!disabled && (!isArrowKey || isArrowOpen)) {
         componentRef?.current?.blur();
         onClick?.(e);
+
+        // For macOS, the pressable keyDown handler also calls onPress() for 'Enter' / 'Space' keys. This means that
+        // this invoke() callback is called twice on the platform. preventDefault() makes it so that its handled once.
+        if (preventDuplicateKeyDownHandling) {
+          e.preventDefault();
+        }
       }
 
       if (!hasSubmenu && !isArrowKey && !shouldPersist) {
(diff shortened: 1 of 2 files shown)


packages/components/Menu/src/MenuItem/useMenuItem.ts
```diff
@@ -38,6 +40,12 @@ export const useMenuItem = (props: MenuItemProps): MenuItemInfo => {
       if (!disabled && (!isArrowKey || isArrowOpen)) {
         componentRef?.current?.blur();
         onClick?.(e);
+
+        // For macOS, the pressable keyDown handler also calls onPress() for 'Enter' / 'Space' keys. This means that
+        // this invoke() callback is called twice on the platform. preventDefault() makes it so that its handled once.
+        if (preventDuplicateKeyDownHandling) {
+          e.preventDefault();
+        }
```
Saadnajmi: Is this not also true for RNW? Also, wouldn't we rather have the default?

lawrencewin: Turns out it is true for win32, but the same check fails because the `nativeEvent` object doesn't have a `code` property https://github.com/microsoft/react-native-windows/blob/a93b35a8709ddb27b4c50925c3867363d3dd965d/packages/%40office-iss/react-native-win32/src-win/Libraries/Pressability/Pressability.win32.js#L619C12-L623C1 😨 

If we shouldn't prevent default here, then how can we mark the event as handled or in the very case skip the second check? Have a special case for ENTER / SPACE here?

rurikoaraki: i'm confused by "Turns out it is true for win32, but the same check fails." Are you saying that we see similar behavior on win32 but there's a different reason for it? Or that the code exists on RNWin32, but it doesn't trigger, so we don't see the double-invoke behavior?

FWIW I think preventDefault is fine, since it tells the platform (RN) to not do the default action of handling enter/space for us. But I would like to see the comment be clearer that we're preventing react-native's default behavior

lawrencewin: The latter - the code exists on RN Win32, but the check always fails because it reads the wrong property on the event, so we don't see the issue.

rurikoaraki: This looks fine to me, but @Saadnajmi if you have a strong opinion feel free to share

Saadnajmi: I think I opened a bug on RNW a while back about this. I'm going to trust the PR outcome.

PR #3934, opened 2025-07-14
Fix bundling to use typescript files rather than js files

JasonVMo:
### Platforms Impacted
- all

### Description of changes

We used to have the package.json main field pointing at src/index.ts, then we would change it to point to lib-commonjs/index.js on publish. This served to ensure that metro bundling used the typescript files directly when bundling. This avoids the need to build, while also ensuring fast-refresh worked correctly when changes were made. This was changed as part of modernizing the way modules were defined and referenced in the repo (which was previously hiding bugs) but had the downside of breaking fast-refresh in the repo.

Instead of the old system (with the custom publishing hook) instead I added "react-native": "src/index.ts" references to the packages. This will cause the metro bundler to use the typescript entry-point and causes fast-refresh to work again. This will also cause downstream consumers to use the typescript entry points in their builds as well which is good for type-checking, good for avoiding intermediate helpers being added as part of JS build targeting, and eventually good for allowing static hermes to better optimize the code.

### Verification

Normal build & pipeline checks, also checked that bundling runs without building the repo first.

### DIFF
--- a/apps/E2E/package.json
+++ b/apps/E2E/package.json
@@ -5,6 +5,7 @@
   "license": "MIT",
   "main": "lib-commonjs/index.js",
   "module": "lib/index.js",
+  "react-native": "src/index.ts",
   "typings": "lib/index.d.ts",
   "exports": {
     ".": {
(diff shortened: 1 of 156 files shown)


Saadnajmi (commented):
Should we have some linter or CI check enforce this?

Saadnajmi (approved):

JasonVMo:
> Should we have some linter or CI check enforce this?

Created issue #3936 to keep track of this.

PR #3935, opened 2025-07-14
Rework scripts to no longer require the build-tools step, and make eventual direct TS transition easier

JasonVMo:
### Platforms Impacted
- all

### Description of changes

While it is possible to run TS directly in node 22.x, it isn't 100% viable until 23.6 because it is too easy to loop back into scripts and inadvertently call without the extra command line args. At the same time the build-tools step is super tedious when changing engineering systems stuff so I want the ability to execute it directly. Finally the scripts directory was a mess. This PR makes the following changes:

- reorganize files, creating a scripts/configs directory to put common configs, a preinstall directory to put code that needs to run before package installs happen.
- hook up linting in scripts. Still a todo to override scripts linting to ensure only node imports are enabled in presintall.
- move the tsconfig the whole project builds off of to the configs directory.
- move the scripts tsconfig to pull in the @rnx-kit/tsconfig/tsconfig.node.json to put in the right compiler directives required to execute TS code directly. Also make it run on the whole scripts package and check JS files.
- move any TS files to be JS + JSDoc syntax
- centralize a bunch of functionality, particularly the root project stuff to ensure we can share that amongst scripts.
- move scripts to be type: "module" to allow modern import syntax (makes eventual TS conversion easier)
- remove dead/unused scripts
- convert the jest execution to a direct script rather than the just based task
- export ObjectBase from framework to ensure typescript inferencing doesn't give errant non-portable type warnings.
- modified the cli entry-point to not require a just.config.js in every folder. As part of that deleted those just.config.js files.

Ultimately many of the files that look brand new are actually renames from .ts to .js which are showing up as moves. Most of the changes are changing require to import statements, and changing config references.


Saadnajmi:
Do we still need the lage step "build-tools"?

JasonVMo:
> Do we still need the lage step "build-tools"?

Good catch. I had it in another version of it but forgot to do it this time. Updated with paring down of the lage config (and fixing the execution of prettier in the scripts folder).

lage.config.js
```diff
@@ -1,15 +1,14 @@
...
     buildci: ['build', 'test', 'depcheck'],
-    bundle: ['build-tools', 'build'],
+    bundle: ['build'],
     clean: [],
-    depcheck: ['build-tools', 'align-deps'],
-    lint: ['build-tools'],
-    prettier: ['build-tools'],
-    ['prettier-fix']: ['build-tools'],
-    test: ['build-tools', 'lint', 'build'],
+    depcheck: ['align-deps'],
```
Saadnajmi: Maybe a follow up, but these are technically 2 different verbs. The combination is a carryover from when align deps was named depchek except we had a linter named dep check already. Could we separate the verbs?

Saadnajmi (approved):

lage.config.js
JasonVMo: Yep, we already have a pipeline task to run it, I removed it from the depcheck step.

Saadnajmi: Follow up idea: we run the repo wide tasks (align-deps, lint-lockfile) in CI, but not in lage. It would be nice to add to lage. It requires adding "." As a workspace. I ran into hangs when I tried, so I scrapped the idea. Maybe it works better now with all the eng sys improvements

PR #3945, opened 2025-07-22
Fix `gap` token on `MenuList`

rurikoaraki:
### Platforms Impacted
- [x] iOS
- [x] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

`gap` token wasn't working on MenuList component. Turns out it was being applied to the wrong component, the outer root View which only has one child, instead of the focusZone slot which actually wraps the menu items.

Apply the gap token to the focus zone instead. If no focus zone is present then make it a View.

### Verification

Used the customized Menu test which applies a noticeable gap to the MenuList

| image | image |

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Menu/src/MenuList/MenuList.styling.ts
+++ b/packages/components/Menu/src/MenuList/MenuList.styling.ts
@@ -15,16 +15,29 @@ export const stylingSettings: UseStylingOptions<MenuListProps, MenuListSlotProps
   states: menuListStates,
   slotProps: {
     root: buildProps(
-      (tokens: MenuListTokens, theme: Theme) => ({
+      (tokens: MenuListTokens, theme: Theme) => {
+        console.log(tokens);
+        return {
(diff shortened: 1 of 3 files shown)


packages/components/Menu/src/MenuList/MenuList.styling.ts
```diff
@@ -15,16 +15,29 @@ export const stylingSettings: UseStylingOptions<MenuListProps, MenuListSlotProps
   states: menuListStates,
   slotProps: {
     root: buildProps(
-      (tokens: MenuListTokens, theme: Theme) => ({
+      (tokens: MenuListTokens, theme: Theme) => {
+        console.log(tokens);
```
Saadnajmi: remove?

PR #3946, opened 2025-07-23
Get rid of just-script and just-task entirely and switch to clipanion

JasonVMo:
### Platforms Impacted
- all

### Description of changes

This finishes moving the scripts package away from just and instead leverages clipanion, the same cli tool that yarn uses internally. Many of these scripts are adapted from what is used in rnx-kit as well. It's pretty straightforward to see how it is used.

The other main thing this does is clean up the lage.config.js file to declare inputs and outputs on most of the build tasks. The reason why this would sometimes grind to a halt is that it was often indexing the full scope of node_modules. It is possible that when node modules change drastically that a build --no-cache will have to be executed in case it doesn't pick everything up, but all other builds should run much faster.

This also has a few cleanup tasks:
- remove the unused (and likely not working for some time) copy task
- bump the @types/node package to match our version of node.
- dedupe the @types/node versions.


scripts/src/tasks/clean.js
```diff
@@ -1,9 +1,31 @@
 // @ts-check
 
+import { Command } from 'clipanion';
+import * as fs from 'fs';
```
Saadnajmi:
```suggestion
import * as fs from 'node:fs';
```

JasonVMo: This syntax (node:fs) doesn't seem to work in the repo right now, I need to look into it.

JasonVMo: Actually it looks like when I switched to the separate tsconfig with more modern definitions in scripts this now works. I've updated the references to use node:fs, node:path, etc.

PR #3947, opened 2025-07-25
JSX runtime (part 2 of 3) - more formal rendering logic for both classic and new jsx runtimes

JasonVMo:
### Platforms Impacted
- all

### Description of changes

This further centralizes the render logic and provides both classic and jsx-runtime implementations. It also does a few other things:

- Create more correct patterns around composition and staged rendering. It turns out the handling of children was incorrect, and in the case where the JSX handlers are omitted will cause indeterminate behavior with children. This also creates some type divergence.
- Fixed the typings on the render functions to correctly mirror createElement/jsx/jsxs
- Moved memo-cache, immutable-merge, and merge-props documentation into the correct package, and created a reference link to the new locations.

### Verification

Builds, tests, bundling, running the testers and trying the controls. Also inspection of .tsx -> .js output as well as snapshot test results for render trees (which remain unchanged).

### Other Notes

I will be creating unit tests as part of the final change where I turn on the new runtime. I need to figure out how to correctly unit test both classic and jsx scenarios at the same time.

### DIFF
--- a/packages/framework-base/src/component-patterns/README.md
+++ b/packages/framework-base/src/component-patterns/README.md
@@ -1,6 +1,6 @@
 # `fluentui-react-native` - Common component patterns
 
-These are the base component patterns shared across the deprecated framework, and the newer framework. This also includes the custom JSX handlers required to render them properly.
+These are the base component patterns shared across the deprecated or v0 framework (found under packages/deprecated), and the newer framework (found under packages/framework). This also includes the custom JSX handlers required to render them properly.
 
 There are two main patterns exposed here: direct rendering and staged rendering.
 
@@ -22,8 +22,8 @@ When this is rendered, there is an entry for `MyNewText` which contains a `MyTex
 
 There are two types of implementations in this folder:
 
-- `DirectComponent` - a functional component that marks itself as direct with a `_callDirect: true` attached static. This will then be called as a normal function component, with children included as part of props.
-- `LegacyDirectComponent` - the pattern currently used in this library that should be moved away from. In this case `_canCompose: true` is set as an attached static, and the function component will be called with children split from props.
+- `DirectComponent` - a functional component that marks itself as direct with a `_callDirect: true` attached property. This will then be called as a normal function component, with children included as part of props.
+- `LegacyDirectComponent` - the pattern currently used in this library that should be moved away from. In this case `_canCompose: true` is set as an attached property, and the function component will be called with children split from props.
 
 The internal logic of the JSX rendering helpers will handle both patterns. In the case of the newer `DirectComponent` pattern, the component will still work, even without any jsx hooks, whereas the `LegacyDirectComponent` pattern will have a somewhat undefined behavior with regards to children.
 
@@ -36,4 +36,4 @@ As above there is a newer and older version of the pattern.
 - `StagedComponent` - the newer version of the pattern, where the returned component function expects children as part of props.
 - `StagedRender` - the older version, where children are split out and JSX hooks are required to render correctly.
 
-Note that while the newer patterns work without any JSX hooks, the hooks will enable the layer removal.
+Note that while the newer patterns work without any JSX hooks, the hooks will enable the element flattening.
--- a/packages/framework/immutable-merge/README.md
+++ b/packages/framework/immutable-merge/README.md
@@ -2,4 +2,4 @@
 
 This package is now a pass-through package which references `@fluentui-react-native/framework-base` for backwards compatibility.
 
-The current documentation can be found [here](../../framework-base/immutable-merge/README.md).
+The current documentation can be found [here](../../framework-base/src/immutable-merge/README.md).
--- a/packages/framework/memo-cache/README.md
+++ b/packages/framework/memo-cache/README.md
@@ -2,4 +2,4 @@
 
 This package is now a pass-through package which references `@fluentui-react-native/framework-base` for backwards compatibility.
 
-The current documentation can be found [here](../../framework-base/memo-cache/README.md).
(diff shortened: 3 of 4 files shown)


packages/framework-base/src/component-patterns/stagedComponent.tsx
```diff
@@ -0,0 +1,45 @@
+/**
```
Saadnajmi, line 1: Any idea why this isn't recognized as a rename?

packages/framework-base/src/component-patterns/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```ts
+const MyNewText: React.FunctionComponent<MyTextProps> = (props) => {
+  return <MyText {...props, italicize: true} />;
+}
+```
+
+When this is rendered, there is an entry for `MyNewText` which contains a `MyText` (another entry), which might contains `Text` (for react-native usage). The direct rendering pattern is one where a component can denote that it is safe to be called directly as a function, instead operating as a prop transform that gets applied to the underlying component.
+
+- For the above to be safe, `MyNewText` should NOT use hooks. In the case of any conditional rendering logic this will break the rule of hooks.
```
Saadnajmi, line 21: It can totally safely use hooks though? At least, that one small example provided above?

packages/framework-base/src/component-patterns/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+There are two types of implementations in this folder:
+
+- `DirectComponent` - a functional component that marks itself as direct with a `_callDirect: true` attached static. This will then be called as a normal function component, with children included as part of props.
```
Saadnajmi: Nit: I don't know what statics are, and have not been able to find a definition when I last did FURN dev actively, so I didn't know what you were talking about when you referred to this.

packages/framework-base/src/component-patterns/README.md
```diff
@@ -0,0 +1,39 @@
...
+The internal logic of the JSX rendering helpers will handle both patterns. In the case of the newer `DirectComponent` pattern, the component will still work, even without any jsx hooks, whereas the `LegacyDirectComponent` pattern will have a somewhat undefined behavior with regards to children.
+
+## Staged Rendering
+
+The issue with the direct component pattern above, is that hooks are integral to writing functional components. The staged rendering pattern is designed to help with this. In this case a component is implemented in two stages, the prep stage where hooks are called, and the rendering stage where the tree is emitted.
+
+As above there is a newer and older version of the pattern.
+
+- `StagedComponent` - the newer version of the pattern, where the returned component function expects children as part of props.
+- `StagedRender` - the older version, where children are split out and JSX hooks are required to render correctly.
```
Saadnajmi, line 37: Could we just call it v0 and v1? I can't remember if we set @fluentui-react-native/framework as v1 yet.

packages/framework-base/src/component-patterns/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Note that while the newer patterns work without any JSX hooks, the hooks will enable the layer removal.
```
Saadnajmi: nit: layer removal might be better understood as View flattening, since there's docs for that: https://reactnative.dev/architecture/view-flattening

Or maybe just explain that more in detail somewhere.

NikoAri (approved):

packages/framework-base/src/component-patterns/README.md
JasonVMo: Yes, the pattern is designed with that in mind. I will make the documentation more comprehensive in the next stage of the change and remove the older documentation which is currently still existing.

packages/framework-base/src/component-patterns/README.md
JasonVMo: Ah, attached properties would be a better term for them. I'll update it when I redo the documentation for the next change.

packages/framework-base/src/component-patterns/stagedComponent.tsx
JasonVMo: It has the @jsx comment at the top, the types have been moved out of this file, and there is an additional implementation below, which I guess triggered the new logic rather than the rename logic.

packages/framework-base/src/component-patterns/README.md
JasonVMo: The newer ones mentioned here are actually not used internally yet, I need to figure out the best way to safely convert. Both frameworks use the older pattern. Again I will update this with the next change as I look into it more and build up more tests.

packages/framework-base/src/component-patterns/README.md
JasonVMo: Fixed

packages/framework-base/src/component-patterns/README.md
JasonVMo: Switched to calling them attached properties.

packages/framework-base/src/component-patterns/stagedComponent.tsx
rurikoaraki: generally if more than 50% of the file content changes, git won't recognize it as a rename. It acts on content and not on file name.

packages/framework-base/src/component-patterns/README.md
rurikoaraki: yeah some clarity here would be nice. This makes it sound like hooks can't be used with FURN components when used as direct components at all, which sounds bad?

rurikoaraki (approved):
:shipit: - with suggestions, which is apparently not an option with GH

Saadnajmi (approved):

PR #3948, opened 2025-08-01
Fix dependencies, create config packages, and implement depcheck --fix-errors

JasonVMo:
### Platforms Impacted
- all

### Description of changes

The repo had a large number of missing or incorrect dependencies, some of which were being hidden by the ignore list in depcheck, others being hidden by the use of the old-style node_modules layout in yarn. Because of this we often have to inject dependencies in OMR to make things work correctly.

This change includes a few key pieces:

#### More comprehensive `depcheck` command

The previous use of depcheck had a large list of ignored packages that were hiding a number of issues. For various issues I had run into the issue of needing to manually make changes to large number of packages to fix things, to address this I got rid of the special casing in depcheck and adding fix capabilities.

- centralized the tool versions into a new package in `scripts/src/preinstall` this is where various tools pull common versions.
- depcheck and the dynamic dependencies now rely on this.
- depcheck now takes the dependencies injected via dynamic dependencies into account and ignores them.
- depcheck does the ignore handling as post-processing rather than built-into the tool to not hide them
- `--verbose` option that will show ignored errors
- `--fix-errors` option that will automatically fix errors if it can. For packages that it can't determine the version it will fail.
- `--fix-warnings` option that will fix warnings. Use this with caution as it will flag some devDependencies that are actually needed.
- `--dry-run` option that will show what the fix mechanisms will do without making changes.
- added better coloring so you can clearly see what is going on.

The fix can be run from root by executing `yarn workspaces foreach --all --no-private run depcheck --fix-errors`. The no-private option is needed for now as it will fail on the root.

#### Dedicated packages for jest and babel configurations

This helps isolate the dependencies required for those packages so they aren't all jammed together in scripts. These were created under `packages/configs` and I will move the eslint config package there as a separate change and likely eventually create a separate tsconfig package as well. These are all private packages just used within the repo.

- Note that the checkbox component hasn't been running tests due to a missing jest configuration, I restored this and updated the snapshots to be current now.

#### Got rid of codegen as part of the build-step

This was unnecessary as it will be done either way as part of metro bundling as it is handled by the metro babel transforms.

#### Switched to pnpm mode

Once fixing depcheck and doing the other work this was mainly tweaking dependency declarations to be correct.

### Verification

- Automated tests.


apps/fluent-tester/src/TestComponents/PersonaCoin/CustomizeUsage.mobile.tsx
```diff
@@ -1,14 +1,17 @@
...
 import { PersonaCoin } from '@fluentui/react-native';
 import { useTheme } from '@fluentui-react-native/theme-types';
 import Slider from '@react-native-community/slider';
+import type { SliderProps } from '@react-native-community/slider';
 
 import type { undefinedText } from './styles';
 import { steveBallmerPhotoUrl } from './styles';
 import { commonTestStyles as commonStyles } from '../Common/styles';
 
+const NativeSlider = Slider as unknown as React.ComponentType<SliderProps>;
```
Saadnajmi, line 13: As unknown?

packages/framework-base/src/merge-props/mergeStyles.test.ts
```diff
@@ -1,8 +1,9 @@
-import type { ColorValue } from 'react-native';
-
 import { flattenStyle, mergeAndFlattenStyles, mergeStyles } from './mergeStyles';
 import type { StyleProp } from './mergeStyles.types';
 
+type OpaqueColorValue = symbol & { __TYPE__: 'Color' };
+type ColorValue = string | OpaqueColorValue;
```
Saadnajmi, line 5: Why not use the RN import?

JasonVMo: It was the only react-native dependency in this package, used for a simple type import for a test where the actual behavior wasn't important so it is simpler to remove it. I'm trying to be more clear in configuration for react-native vs. more generic packages.

apps/fluent-tester/src/TestComponents/PersonaCoin/CustomizeUsage.mobile.tsx
JasonVMo: It's a more correct way of having an intermediate cast. TS requires something like this when it can't figure out the overlap between the two types, in this case Slider and React.ComponentType<SliderProps>

apps/fluent-tester/ios/Podfile
```diff
@@ -1,15 +1,17 @@
...
 
 use_test_app! do |target|
   target.app do
     platform :ios, '15.0'
 
     # There is a bug where autolinking isn't working, so specify these manually.
-    pod 'FRNAppearanceAdditions', :path => '../../../packages/experimental/AppearanceAdditions/FRNAppearanceAdditions.podspec'
-    pod 'FRNFontMetrics', :path => '../../../packages/experimental/NativeFontMetrics/FRNFontMetrics.podspec'
+    pod 'FRNAppearanceAdditions', :path => '../node_modules/@fluentui-react-native/experimental-appearance-additions/FRNAppearanceAdditions.podspec'
+    pod 'FRNFontMetrics', :path => '../node_modules/@fluentui-react-native/experimental-native-font-metrics/FRNFontMetrics.podspec'
```
Saadnajmi: I'm not a fan of these changes, but I also suspect with autolinking, we can remove these lines now.

apps/win32/metro.config.js
```diff
@@ -5,10 +5,21 @@
...
+const excludeMixins = [];
+const extraNodeModules = {};
+function ensureUniqueModule(moduleName, excludeList, nodeModules) {
+  const [nmEntry, excludePattern] = resolveUniqueModule(moduleName);
+  excludeMixins.push(excludePattern);
+  extraNodeModules[moduleName] = nmEntry;
+}
+
+// build up the added excludes and extraNodeModules
+['react-native-svg'].forEach((moduleName) => ensureUniqueModule(moduleName));
```
Saadnajmi, line 21: Where does the duplicate come from? Can we just enforce this ourselves via a resolution or aligning every dependency on it now?

apps/fluent-tester/macos/Podfile
```diff
@@ -1,18 +1,20 @@
...
-    pod 'FRNAvatar', :path => '../../../packages/experimental/Avatar/FRNAvatar.podspec'
-    pod 'FRNCheckbox', :path => '../../../packages/experimental/Checkbox/FRNCheckbox.podspec'
-    pod 'FRNMenuButton', :path => '../../../packages/components/MenuButton/FRNMenuButton.podspec'
-    pod 'FRNRadioButton', :path => '../../../packages/components/RadioGroup/FRNRadioButton.podspec'
-    pod 'RCTFocusZone', :path => '../../../packages/components/FocusZone/RCTFocusZone.podspec'
+    pod 'FRNAvatar', :path => '../node_modules/@fluentui-react-native/experimental-avatar/FRNAvatar.podspec'
+    pod 'FRNCheckbox', :path => '../node_modules/@fluentui-react-native/experimental-checkbox/FRNCheckbox.podspec'
+    pod 'FRNMenuButton', :path => '../node_modules/@fluentui-react-native/menu-button/FRNMenuButton.podspec'
+    pod 'FRNRadioButton', :path => '../node_modules/@fluentui-react-native/radio-group/FRNRadioButton.podspec'
+    pod 'RCTFocusZone', :path => '../node_modules/@fluentui-react-native/focus-zone/RCTFocusZone.podspec'
```
Saadnajmi: Similar comment on "Maybe we don't need this anymore"

lage.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
       inputs: ['*', 'src/**/*', 'assets/**/*'],
       outputs: ['lib/**/*', 'lib-commonjs/**/*'],
     },
-    buildci: ['build', 'test', 'lint', 'depcheck'],
+    buildci: ['build', 'test', 'lint', 'depcheck', 'check-publishing'],
```
Saadnajmi, line 9: Can we add `lint-lock file` and `align-deps` here too? I couldn't because those run on the root and need the root to be a workspace, which then required us to specify inputs/outputs. Maybe now, this is better,

package.json
```diff
@@ -19,8 +19,11 @@
     "change": "beachball change",
     "check-for-changed-files": "cd scripts && yarn fluentui-scripts check-changes",
     "checkchange": "beachball check --changehint \"Run 'yarn change' to generate a change file\"",
+    "check-publishing": "node ./scripts/src/cli.mjs check-publishing",
```
Saadnajmi, line 22: Out of scope, could this be a yarn constraint?

apps/fluent-tester/ios/Podfile
JasonVMo: Removed in latest, though I'm also temporarily turning off pnpm mode as the appium issues are tricky.

apps/fluent-tester/macos/Podfile
JasonVMo: Removed

lage.config.js
JasonVMo: Added a new command "pr-check" that does everything. Because this command is referenced in the pipeline tasks I didn't want to change that now.

JasonVMo:
Note that I have reverted the pnpm mode changes for now, though the dependency cleanup, dedicated config packages and everything else is still valuable and needed for the next jsx change. The issue with pnpm mode was related to the e2e test packages not working right. I started to work through the configuration but it requires a bit more understanding.

apps/win32/metro.config.js
JasonVMo: I'll look at this more when I reenable pnpm.

package.json
JasonVMo: Definitely a good thing to look at later. I haven't fully looked at the yarn constraint code and what it can do yet.

PR #4058, opened 2026-03-07
Add a common config package with a strict tsconfig, then switch some core packages to have stronger type safety

JasonVMo:
### Platforms Impacted

- all

#### Add `@fluentui-react-native/config` package

This adds the `@fluentui-react-native/config` package which right now just contains a strict tsconfig.json file. I want to aggregate a bunch of the configs into this one package to cut down on noise and the sheer number of discreet packages in the repo. I'll consolidate as part of a future checkin.

#### Switch framework-base, adapters, interactive-hooks, and use-slot to strict types

These packages now build with the new stricter tsconfig, which forced me to improve the type safety at the core levels. This helps catch bugs but also is necessary if we want to ever have people build against typescript directly as it will be subject to their own tsconfig settings in some cases.

The main changes of note are in framework-base, see the comments for about things like using object as the base type for acceptance and record as the fallback type. These changes should make inference much stronger and cut down on both types becoming too loose, and on incorrect assignment errors.

### Verification

Builds and tests

### DIFF
--- a/packages/utils/interactive-hooks/src/useAsToggle.ts
+++ b/packages/utils/interactive-hooks/src/useAsToggle.ts
@@ -15,7 +15,7 @@ export type OnChangeCallback = () => void;
  *         state.isChecked - Whether or not component is currently checked or selected
  */
 export function useAsToggle(defaultChecked?: boolean, checked?: boolean, userCallback?: OnToggleCallback): [boolean, OnChangeCallback] {
-  const [isChecked, setChecked] = React.useState(defaultChecked ?? checked);
+  const [isChecked = false, setChecked] = React.useState(defaultChecked ?? checked);
 
   const onChange = React.useCallback(() => {
     userCallback && userCallback(!isChecked);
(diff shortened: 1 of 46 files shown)


Saadnajmi:
EDIT: I see it was in scripts. Ok

I thought we had a package for the tsconfig already? I'll check next week as I'm ending my day now.

packages/utils/interactive-hooks/src/useAsToggle.ts
```diff
@@ -15,7 +15,7 @@ export type OnChangeCallback = () => void;
  *         state.isChecked - Whether or not component is currently checked or selected
  */
 export function useAsToggle(defaultChecked?: boolean, checked?: boolean, userCallback?: OnToggleCallback): [boolean, OnChangeCallback] {
-  const [isChecked, setChecked] = React.useState(defaultChecked ?? checked);
+  const [isChecked = false, setChecked] = React.useState(defaultChecked ?? checked);
```
Saadnajmi, line 18: technically this change might warrant a minor release? We are changing the default behavior, as evidenced by the tests changing as well

PR #4087, opened 2026-04-14
Remove overflow hidden from Button for Win32/Windows

rurikoaraki:
### Platforms Impacted

- [ ] iOS
- [ ] macOS
- [x] win32 (Office)
- [x] windows
- [x] android

### Description of changes

Turns out adding overflow hidden to all our Buttons is a perf problem for win32. So removing the property from win32 (and windows, in case they have a similar issue).

### Verification

Tested the button in the win32 app. Corners look fine, double focus border also looks fine

### Pull request checklist

- [ ] Automated Tests
- [ ] Keyboard Accessibility
- [ ] Voiceover

### DIFF
--- a/packages/components/Button/src/Button.styling.ts
+++ b/packages/components/Button/src/Button.styling.ts
@@ -33,6 +33,8 @@ export const buttonStates: (keyof ButtonTokens)[] = [
   'disabled',
 ];
 
+const applyOverflowHiddenToRoot = Platform.OS === 'ios' || Platform.OS === 'macos';
+
 export const stylingSettings: UseStylingOptions<ButtonProps, ButtonSlotProps, ButtonTokens> = {
   tokens: [defaultButtonTokens, defaultButtonFontTokens, defaultButtonColorTokens, buttonName],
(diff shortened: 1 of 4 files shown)


packages/components/Button/src/Button.styling.ts
```diff
@@ -33,6 +33,8 @@ export const buttonStates: (keyof ButtonTokens)[] = [
   'disabled',
 ];
 
+const applyOverflowHiddenToRoot = Platform.OS === 'ios' || Platform.OS === 'macos';
```
Saadnajmi, line 36: Wouldn't it be the other way around, specifically don't apply to windows? Mostly bc Android

rurikoaraki: Android is already handled by a different component (the ripple). Specifically I wanted to keep the intended effect of #3658 unaffected but it does seem that applying it too much is a problem, so if that fix is mainly targeted for Apple then I would like it to stay that way

Saadnajmi (approved):

PR #4101, opened 2026-05-07
Switch tsconfig settings for all packages in the repo

JasonVMo:
### Platforms Impacted

-- all

### Description of changes

This change has two parts, the config level and package level.

At the config level:
- add two new tsconfig.json files to scripts, a general one and a strict one
- adds an exports map to scripts, and exposes those two files via the map
- module/moduleResolution changed to "nodenext" this allows for requiring ESM which is supported in all versions of node we are using.
- add `erasableSyntaxOnly` and `noDefaultOverrides` to the looser config

At the package level:
- update tsconfig.json files to refer to the right source location
- ensure we have a rootDir set if noEmit is not true. This is required to move forward with newer versions of tsgo/tsc.
- standardize the outDir setting if it wasn't already set
- ensure noEmit is declared at the package level when appropriate.

For packages the type errors fixed fall into several buckets:
- add explicit overrides on class elements that are overridden.
- get rid of const enums, generally replacing them with constant records. const enums are a typescript only thing.
- get rid of some older forms to type assertions.


apps/win32-81/src/index.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+import { AppRegistry } from 'react-native';
+import { FluentTesterApp } from '@fluentui-react-native/tester-core';
+AppRegistry.registerComponent('FluentTester', () => FluentTesterApp);
+export default FluentTesterApp;
+//# sourceMappingURL=index.js.map
```
Saadnajmi: remove?

apps/win32/src/index.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+Object.defineProperty(exports, "__esModule", { value: true });
+const react_native_1 = require("react-native");
+const tester_core_1 = require("@fluentui-react-native/tester-core");
+react_native_1.AppRegistry.registerComponent('FluentTester', () => tester_core_1.FluentTesterApp);
+exports.default = tester_core_1.FluentTesterApp;
+//# sourceMappingURL=index.js.map
```
Saadnajmi: remove?

packages/components/Menu/src/Menu/Menu.types.ts
```diff
@@ -67,8 +67,9 @@ export interface MenuState extends MenuProps {
...
-  Hidden,
-  Animating,
-  Shown,
-}
+export const AndroidMenuStates = {
+  Hidden: 0,
+  Animating: 1,
+  Shown: 2,
+} as const;
+export type AndroidMenuStates = (typeof AndroidMenuStates)[keyof typeof AndroidMenuStates];
```
Saadnajmi, line 75: Probably out of scope for this PR, but all of these would be better served as string unions right?

packages/utils/interactive-hooks/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
-  "extends": "@fluentui-react-native/scripts/configs/tsconfig.json",
+  "extends": "@fluentui-react-native/scripts/tsconfig",
   "compilerOptions": {
     "outDir": "lib",
-    "lib": ["es2022", "dom"]
+    "lib": ["es2022", "dom"],
```
Saadnajmi: THis one is special?

Saadnajmi (approved):

packages/components/Menu/src/Menu/Menu.types.ts
JasonVMo: Potentially yes. This pattern is effectively what a const enum is behind the scenes. There is a question of validation for user content. If you want validation (e.g. turn a string into an enumerated type) you need to use the pattern of:
```ts
const MY_VALUES = ["val1", "val2"] as const;
export type MyType = typeof MY_VALUES[number];

export function isMyType(test: string): test is MyType {
  return MY_VALUES.includes(test);
}
```
The syntax above might be slightly off as I'm doing it from memory. The object with numbers has some benefits for comparisons if there is a prominence to the levels. The array pattern in my comment is slightly more optimal if not.

The direct string enum type is mainly good for internal only code where type checking is sufficient.

packages/utils/interactive-hooks/tsconfig.json
JasonVMo: Legacy, but I just removed it.

PR #4109, opened 2026-05-09
Remove last config package and move to root level package linting

JasonVMo:
### Platforms Impacted

- configuration only

### Description of changes

- This moves the kit-config into scripts (which gets rid of the last dedicated config package)
- lint-package is deprecated and no longer run package by package
- There is now a `lint-repo` / `lint-repo --fix` command at the root which runs through the various repo wide linting operations. This takes ~4 seconds on my machine instead of the > 1 minute that the lint-package routine used to take.
- Depcheck is replaced by a knip run from the root (in lint-repo)
- package.json validation now runs during yarn constraints using the lint-package context.
- align-deps runs from the root (previously blocked due to the extends bug I recently fixed)
- format runs and will always run last in fix mode to clean up editing issues from the other tools
- lint-lockfile runs as part of this as well

lintPackage is still present as some of the code in there will need to be moved over, but I will likely not keep all of it so that will be done later.

### Verification

Automation


yarn.config.cjs
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = defineConfig({
+  constraints: async ({ Yarn }) => {
+    for (const workspace of Yarn.workspaces()) {
+      if (workspace.cwd !== '.') {
+        constrain(workspace);
+      }
+    }
+  },
+});
```
Saadnajmi, line 14: Can we add a comment to make it a little more obvious that instead of implementing constraints dirctly, we are using our helper package?

scripts/src/tasks/lintRepo.ts
```diff
@@ -0,0 +1,99 @@
...
+  const key: keyof CommandVariants = fix ? 'fix' : 'check';
+  const ops: Promise<boolean>[] = [];
+  for (const cmdName of Object.keys(lintCommands)) {
+    const command = lintCommands[cmdName][key];
+    if (command) {
+      if (fix) {
+        if (!(await execCommand(cmdName, command))) {
+          return 1;
+        }
+      } else {
```
Saadnajmi, line 47: It feels like this could be handled by Lage?

JasonVMo: It's a bit easier to manage this way. With lage we would have to add a script command for each part as they differ in fix and non-fix mode. This also does output suppression except when errors happen which makes the commands much noisier.

Saadnajmi: Could it be a "worker" script then, and keep Lage's scheduler?
https://github.com/microsoft/fluentui-react-native/blob/c9ea2d963a43d33f6ca13fd0efa2c4696880e871/lage.config.mjs#L76

https://github.com/microsoft/fluentui-react-native/blob/c9ea2d963a43d33f6ca13fd0efa2c4696880e871/scripts/src/worker/publish.mts#L32-L40

JasonVMo: I played with this a bit but this actually ends up a bit messier this way. I'd prefer to keep it as-is for now, we can look at changing this in the future if we don't like it.

Saadnajmi (approved):

PR #4116, opened 2026-06-01
Bump win32 to 0.81

acoates-ms:
Also creating a 0.81 version of the dependency profile.


Saadnajmi:
Do we still need the win32 app or can we use the RNTA managed RNW one?

acoates-ms:
> Do we still need the win32 app or can we use the RNTA managed RNW one?

We still need the win32 one until we drop paper.

JasonVMo (approved):
Looks like the windows app isn't building yet but otherwise it looks good.

PR #4120, opened 2026-06-04
Switch core packages to stricter types

JasonVMo:
### Platforms Impacted

- all / none

### Description of changes

This moves several core packages to the strict tsconfigs. This adds:

- stronger types, particularly around object and style merging
- default values for checked state which shouldn't change behavior unless there was a bug as it makes the pre-initialized state consistent with the unchecked state. Previously it would go undefined -> true -> false, now it alternates between true and false.
- this reapplies the changes from: [PR 4058](https://github.com/microsoft/fluentui-react-native/pull/4058)


packages/components/Button/src/ToggleButton/__snapshots__/ToggleButton.test.tsx.snap
```diff
@@ -14,7 +14,7 @@ exports[`ToggleButton default 1`] = `
   accessibilityState={
     {
       "busy": undefined,
-      "checked": undefined,
+      "checked": false,
```
Saadnajmi: This might actually cause a difference in Screenreaders / VoiceOver fwiw

Saadnajmi (approved):

acoates-ms (approved):

PR #4124, opened 2026-06-11
Finish ESM transition, move to composite builds, detangle type refs, and re-enable macos PR pipeline

JasonVMo:
### Platforms Impacted

- all

### Description of changes

This change has a number of things rolled together into one PR as one thing kind of led to another.

#### Move to ESM

The previous strategy of using module/moduleResolution: "nodenext" doesn't work to esm output on react native because if a module import as an extension (i.e. `import { foo } from "myFile.ts";`) this will stop both metro and typescript from finding myFile.android.ts or other module suffixes. As a result:
- module -> "esnext" to get esm output
- moduleResolution -> "bundler" to enable export maps
- type -> undefined to allow type imports to handle extensionless importing. (module enables stricter handling)

#### Composite builds

To speed up builds, and avoid race conditions for the lage builds where the file wasn't fully committed before the next tsgo instance tried to pick it up, I turned on project references and building with tsgo -b.
- this does an incremental build from any package and when building from the root will build the whole repo at once
- build time has decreased substantially (<8s on my laptop for the whole repo)
- noEmit builds are not permitted directly and must be done as builds from solution files.
- yarn constraints (part of lint-repo) will validate the links are current and update them if necessary
- references will include any workspace dependencies and any tsconfig.* files under a targets/ directory if present. This allows long-term platform specific type checking per-package and other patterns which would yield a large number of tsconfig files.

#### Type defragmentation

The massive parallelism of tsgo highlighted places where we were bringing in multiple versions of @types/react into the same package. The main sources were the icons package bringing in a @types/react-native: 0.68 package and the adapters package including rn, rnw, and rn-macos in the same build, each of which have rn -> react -> react types. Resolution in these cases can be order dependent and is a problem, these were all made more explicit.

#### Re-enable macos pipelines

This also attempts to enable the macos PR builds again as they were turned off in the recent 81 test app changeover.


apps/tester-core/package.json
```diff
@@ -11,22 +11,20 @@
...
-      "require": "./lib-commonjs/index.js",
+      "require": "./lib/index.js",
       "default": "./src/index.ts"
     }
   },
   "scripts": {
-    "build": "fluentui-scripts build",
-    "build-cjs": "tsgo --outDir lib-commonjs",
-    "build-core": "tsgo --outDir lib --module esnext --moduleResolution bundler",
+    "build": "tsgo -b",
```
Saadnajmi, line 26: Was build really nothing e than just ts ?

JasonVMo: It was handling doing both targets as a single command

apps/fluent-tester/tsconfig.json
```diff
@@ -4,7 +4,53 @@
...
+    {
+      "path": "../tester-core/tsconfig.json"
+    },
+    {
+      "path": "../../packages/experimental/Tooltip/tsconfig.json"
+    },
+    {
+      "path": "../../packages/experimental/VibrancyView/tsconfig.json"
+    }
+  ]
```
Saadnajmi, line 55: Where does this list come from?

apps/tester-core/assets/accessible-icon-brands.svg.d.ts
```diff
@@ -0,0 +1,4 @@
+import type * as React from 'react';
```
Saadnajmi, line 1: Can these be generated?

apps/tester-core/package.json
```diff
@@ -82,7 +80,7 @@
     "@fluentui-react-native/vibrancy-view": "workspace:*",
     "@fluentui-react-native/win32-theme": "workspace:*",
     "@fluentui/react-native": "workspace:*",
-    "@fluentui/react-native-icons": "^2.0.316",
+    "@fluentui/react-native-icons": "patch:@fluentui/react-native-icons@npm%3A2.0.322#~/.yarn/patches/@fluentui-react-native-icons-npm-2.0.322-c4bc045837.patch",
```
Saadnajmi: I'm a bit worried it'll be hard to find this patch in the future, but I guess the patches are all stored in one place. Regardless, maybe good to send a PR (or have Copilot do so)?

apps/win32-81/targets/tsconfig.check.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "@fluentui-react-native/scripts/tsconfig",
+  "compilerOptions": {
+    "rootDir": "..",
+    "noEmit": true,
+    "composite": true,
+    "tsBuildInfoFile": "../.cache/tsconfig.tsbuildinfo"
```
Saadnajmi, line 7: Where does this come from?

packages/utils/adapters/src/__drift__/macos.drift.ts
```diff
@@ -0,0 +1,26 @@
...
+ * `enableFocusRing`/`tooltip` on Text — see the comments in `../filters.macos.ts`).
+ */
+import type RN from 'react-native';
+import type { ViewProps, TextProps, ImageProps } from 'react-native-macos';
+import type { ViewPropsMacOS, TextPropsMacOS, ImagePropsMacOS } from '../platformProps';
+import type { AssertNoKeyDrift, MissingKeys, ExtraKeys } from './driftHelpers';
+
+// Props the macOS View mask passes that react-native-macos omits from its *types* (injected at runtime).
+type MacOSViewInjected = 'onPreferredScrollerStyleDidChange' | 'passthroughAllKeyEvents' | 'tooltip' | 'validKeysDown' | 'validKeysUp';
+// Props the macOS Text mask passes that react-native-macos omits from its *types* (injected at runtime).
```
Saadnajmi, line 16: I wonder how to better handle this, assuming we can upstream fixes / edit RN Core, macos, and windows as needed. Historically FURN is the only place tracking these types

scripts/src/tasks/constrain.ts
```diff
@@ -1,18 +1,70 @@
...
+    if (depBuilds.size > 0) {
+      ensureTSConfigReferences(tsCtx, depBuilds);
+    } else {
+      tsCtx.enforce('references', undefined);
+    }
+    targetTSConfigs.add(`${workspace.cwd}/tsconfig.json`);
+    if (!tsCtx.raw.files || tsCtx.raw.files.length > 0) {
+      tsCtx.enforce('compilerOptions.composite', true);
+      tsCtx.enforce('compilerOptions.tsBuildInfoFile', '.cache/tsconfig.tsbuildinfo');
+    }
```
Saadnajmi, line 42: So these things _are_ generated?

Saadnajmi: If constraints are doing this level of work, I feel like we should flip the flag to have constraints run with install. It's basically install-level work now.

scripts/src/utils/tsConfigs.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export function pathsToReferences(paths: Iterable<string>): TsConfigJson['references'] {
+  const refs: TsConfigJson['references'] = [];
+  for (const path of paths) {
+    refs.push({ path });
+  }
+  return refs;
+}
+
+export function areSetsEqual(a: Set<string>, b: Set<string>): boolean {
```
Saadnajmi, line 38: This really doesn't exist in JS / Node?

apps/tester-core/package.json
JasonVMo: Actually this patch should be removed, it was an earlier attempt at fixing the import issue that was eventually fixed with the ignore plugin.

apps/fluent-tester/tsconfig.json
JasonVMo: This list is automatically configured by the code running in yarn constraints. It is effectively setting it to match the package dependencies. If those dependencies change it will be regenerated by running yarn lint-repo-fix at the root or yarn constraints --fix. The list is effectively all the dependencies + any tsconfigs in the package target directory.

apps/tester-core/package.json
JasonVMo: What we need to do is fix it upstream.

packages/utils/adapters/src/__drift__/macos.drift.ts
Saadnajmi: not blocking

Saadnajmi (approved):
Approving to unblock

apps/tester-core/assets/accessible-icon-brands.svg.d.ts
JasonVMo: Yeah, having some tooling based approach to this would be better. The blanket module declaration caused issues with parallelism though so this is a stopgap.

apps/tester-core/package.json
JasonVMo: Removed the patch as it is no longer needed.

apps/win32-81/targets/tsconfig.check.json
JasonVMo: It's hand authored (for now). The tsBuildInfoFile just says to use that file as the incremental build cache and puts it under a .cache directory which is .gitignored.

packages/utils/adapters/src/__drift__/macos.drift.ts
JasonVMo: I'd like to actually extract the types as codegen so we can have distinct named types pulling from upstream without the reference confusion from actually pulling in the libraries.

scripts/src/tasks/constrain.ts
JasonVMo: To do that it would need to be a yarn plugin. I worry about it slowing down install though. It's fast but install runs much more frequently.

scripts/src/utils/tsConfigs.ts
JasonVMo: In es2025 you do if (a.size === b.size && a.isSubsetOf(b)) but I believe we still support node 22 and I don't know if that version of v8 has this functionality.

PR #4134, opened 2026-06-29
Initial agentic-components package + storybook

JasonVMo:
### Platforms Impacted

- none, private only change that doesn't touch components

### Description of changes

This is a staging change that has a few parts:

#### `@fluentui-react-native/agentic-components` package

- This package is private to start as names/structure are TBD
- Adds one rudimentary Button component done in the most basic way possible

#### `@fluentui-react-native/agentic-components-storybook` package

- This is a private sub-package, located in a sub-directory of the components package
- It's done as a sub-package (itself a workspace package) to ensure that the dependency requirements don't pollute the main package dependencies
- This stands up storybook react-native using lite UI mode for macos, though windows is planned.
- Stories are pulled from the component library

### `prebuild` build step

This sets up a `prebuild` build step which a package can use for codegen of various types.

- It is guaranteed to run before build during the root build actions
- NOTE: Individual package build commands don't do this automatically
- It can generate files under src or lib right now (for purposes of caching)
- reworked some of the root scripts to make it work

### `knip` configuration extensions

- Connected the `ignoreDependencies` setting to the furn config so untraceable dependencies can be ignored.
- Fixing the dependency issue in the storybook would create a dependency between the codegen of a file that is .gitignored and repo-linting, so better to use the escape hatch already provided by knip


package.json
```diff
@@ -15,16 +15,18 @@
...
   ],
   "scripts": {
-    "build": "tsgo -b",
-    "clean-all": "node ./scripts/src/preinstall/clean-all.ts",
+    "build": "lage root-build",
     "docs": "yarn workspace fluent-rn-website start",
     "bundle:repo": "lage bundle",
-    "clean": "lage clean",
+    "clean": "tsgo -b --clean && lage clean",
+    "clean-all": "node ./scripts/src/preinstall/clean-all.ts",
```
Saadnajmi, line 22: Why is clean under preinstall?

Saadnajmi (approved):

package.json
JasonVMo: It actually deletes all node_modules so it basically has to be able to run in an uninstalled state.

PR #4140, opened 2026-07-09
Add @fluentui-react-native/design package with some basic codegen for token constants

JasonVMo:
### Platforms Impacted

- none

### Description of changes

This is a non-coding change right now. It adds a new package: `@fluentui-react-native/design` located in packages/agentic-design. The goals of this package are to consolidate design/theme/token/appearance types, concepts, instructions, helpers, and codegenerated constants.

Right now this just adds the package as well as adding some comments to theme-types showing where the theme types are used.

The benefit of the codegen constants is that right now we have a ton of references like:

```ts
// pulls in all globalTokens no matter what
import { globalTokens } from `@fluentui-react-native/tokens`;

const myTokens = {
   rounded: {
      borderRadius: globalTokens.corner.radius40, 
      // minifies to something like g.corner.radius40
   },
   circular: {
     borderRadius: globalTokens.corner.radiusCircular,
   },
   square: {
     borderRadius: globalTokens.corner.radiusNone,
   }
};
```

Usage can be switched to this:

```ts
// imports are statically analyzable for tree-shaking
import { cornerRadius40, cornerRadiusCircular, cornerRadiusNone } from '@fluentui-react-native/design/global';

const myTokens = {
   rounded: {
     borderRadius: cornerRadius40,
     // minifies to something like c1
   }
   // etc...
};
```

The other benefit is the codegen files make it easy to see what values are shared and what values are different.


packages/theming/theme-tokens/package.json
```diff
@@ -33,10 +33,10 @@
...
   },
   "dependencies": {
-    "@fluentui-react-native/design-tokens-android": "^0.53.0",
-    "@fluentui-react-native/design-tokens-ios": "^0.53.0",
-    "@fluentui-react-native/design-tokens-win32": "^0.53.0",
-    "@fluentui-react-native/design-tokens-windows": "^0.53.0",
+    "@fluentui-react-native/design-tokens-android": "catalog:",
+    "@fluentui-react-native/design-tokens-ios": "catalog:",
+    "@fluentui-react-native/design-tokens-win32": "catalog:",
+    "@fluentui-react-native/design-tokens-windows": "catalog:",
```
Saadnajmi, line 39: Feel free to dismiss/resolve, I wonder if it's worth naming our catalogs at some point

JasonVMo: Yeah, it's a good thing to investigate. Given the number of things in the catalog, I think creating some groupings would be good.

Saadnajmi (approved):

PR #4153, opened 2026-07-15
Add basic flex tokens shape and mapping information

JasonVMo:
### Platforms Impacted

- targets all platforms but no runtime change right now

### Description of changes

This is work generated from the design repository. It does the following:
- Create a basic react-native flex tokens shape. This matches the values that will be in newer web-specs or design resources going forward.
- These names/values also map to the token instructions for the various component specs that were merged in recently.

It also defines several mapping files, also encapsulating research and mapping work I did in that repository.
- maps color flex-tokens to fluent tokens (either alias, shadow, or global)
- maps non-color flex tokens to fluent tokens (generally global tokens) 
- maps CSS names to these flex token names

### Verification

Non-coding change


mganandraj (approved):

packages/agentic-design/src/tokens/mappings/fluent-token-map.json
```diff
@@ -0,0 +1,145 @@
+{
+  "_meta": {
+    "description": "Default-value source mapping: each flex React Native token resolves its default value from the referenced FluentUI React Native / Fluent 2 (v9-compatible) token. The value prefix names the source collection.",
+    "prefixes": {
+      "alias.": "Themed semantic alias token — @fluentui-react-native/design-tokens-windows {light,dark}/tokens-aliases.json (theme.colors.alias.color<Name>).",
+      "global.": "Raw global palette token — tokens-global.json, dot-path into the color tree (theme.colors.global.*). Theme-invariant.",
+      "shadow.": "Elevation shadow token — tokens-shadow.json (theme.shadows.*)."
```
Saadnajmi, line 7: Specifically with shadow tokens.. new arch has multiple shadows (what fluent needs) as a first class prop in RN while old arch tried to make a <Shadow /> component that was quite buggy. I am not sure if RNW ever got support for multiple shadows (@acotes-ms ?) but that would be one nuance I'd like the new types to use

packages/agentic-design/src/tokens/mappings/fluent-token-map.json
```diff
@@ -0,0 +1,145 @@
...
+    },
+    "direction": "flexToken <- fluentToken  (key = flex RN token, value = prefixed Fluent token reference)",
+    "states": "Interactive alias-sourced tokens inherit alias interaction states: base uses <alias>.rest, hover.<token> uses <alias>.hover, pressed.<token> uses <alias>.pressed.",
+    "fluentSource": "@fluentui-react-native/design-tokens-windows@0.56.0",
+    "notes": [
+      "global.* references are best-fit matches into the Fluent global palette; expression families map to the matching shared-color ramp step, gray to the grey ramp, alpha/translucent roles to the *Alpha ramps.",
+      "colorMaterialBackground* map to whiteAlpha fallbacks — platform acrylic/vibrancy material (WinUI AcrylicBrush / NSVisualEffectView) is preferred where available.",
```
Saadnajmi, line 14: NSVisualEffectVIew or the Liquid Glass View.. which we don't have a native component for and should add if it matters for the token work.

packages/agentic-design/src/tokens/mappings/gnrc-token-map.json
```diff
@@ -0,0 +1,371 @@
+{
+  "--gnrc-color-surface-neutral-farther-hover": "hover.colorSurfaceNeutralFarther",
```
Saadnajmi, line 2: gnrc?

packages/agentic-design/src/tokens/mappings/fluent-token-map.json
JasonVMo: Definitely something to look at in the future.

packages/agentic-design/src/tokens/mappings/fluent-token-map.json
JasonVMo: If they are relatively easy to produce this makes a lot of sense.

packages/agentic-design/src/tokens/mappings/gnrc-token-map.json
JasonVMo: It's the current naming convention.

PR #4184, opened 2026-08-13
Move Callout to use a primitive component pattern rather than being framework dependent

JasonVMo:
### Platforms Impacted

- all

### Description of changes

These changes are in preparation for building fabric versions of Callout. There are two notable changes here:

#### Dedupe the macos version

The macos version was using `FRNCallout` instead of `RCTCallout` which is what other platforms used. This causes issues when you try to run it through codegen. A secondary issue is that macos variants can't be forked with platform suffixes because macos goes through the ios codegen pipeline and gets confused. The changes here are:
- rename FRNCallout -> RCTCallout, checked OMR and talked to Saad and this should be fine
- get rid of the macos specific version which differed only in default values. This can be handled in JS if necessary

#### Remove the old framework dependency

This component didn't need to be built with either framework, native components aren't really meant to be customized or composed, they should be somewhat dumb layers that are controlled by their callers.
- removed the dependency on the frameworks
- removed the custom adapters usage
- removed unused dependencies
- the callout will be transparent by default, styling can still be set as before or it can be dictated by the contents

### Verification

Tests, manual inspection/debugging


Khalef1 (approved):

acoates-ms (approved):

packages/components/Callout/src/Callout.types.ts
```diff
@@ -24,7 +15,7 @@ export type DirectionalHint =
   | 'rightCenter'
   | 'rightBottomEdge'
   | 'bottomLeftEdge'
-  | 'bottonLeftEdge' // Typo in the original code, should be 'bottomLeftEdge'
+  | 'bottonLeftEdge'
```
acoates-ms, line 18: Can we just remove this?  Anyone with the typo would just have to fix it on FURN upgrade.

acoates-ms (approved):

packages/components/Callout/src/Callout.types.ts
JasonVMo: I'll do this separately but if we can remove it we should definitely do so.

PR #4195, opened 2026-08-17
Release updates + remove unnecessary yarn resolutions

ecraig12345:
## Release updates

- Use the actual `@microsoft/esrp-npm-release` package and remove temporary copy
- Use variable group for certain shared values (open to feedback on whether this is worthwhile)

## Yarn resolutions updates

Update yarn resolutions to remove resolutions that are unnecessary (either the bad version is no longer referenced, or the fixed version is in-range), and scope down the remaining resolutions where possible.

### Recommended procedure for security updates in nested dependencies

In general, to handle security updates in nested dependencies, look at `yarn.lock` and/or use `yarn why`  to determine which of the following cases apply:

- If range is `^` or `~` and fixed version is in range: delete lock file entry, reinstall
- If something pins the dep to a specific version (same major or 0.x-minor as fixed version): set resolution to `"foo@npm:1.2.3": "^1.2.3"` and reinstall
- If fixed version of nested dependency is different major or 0.x-minor from fix:
  - find what depends on it and try to update that
    - direct dep of repo: try updating directly
    - nested dep: same method of deleting from lock file, maybe up to multiple parents
  - for dev or unpublished tool deps only, check the changelog for the new version to see if it's compatible, and if so use a targeted resolution as above


.ado/azure-pipelines.publish.yml
```diff
@@ -180,14 +187,14 @@ extends:
                   PACKED_PACKAGES_PATH: $(packagesArtifactPath)
                   ESRP_PRODUCT_NAME: fluentui-react-native
                   ESRP_NPM_TAG: latest
-                  # TODO should be somebody else
+                  # TODO choose someone else who works on this?
                   ESRP_USER: elcraig@microsoft.com
```
Saadnajmi: You can choose me

Saadnajmi (approved):

